Issue #595, opened 2021-05-30
Open Tickets only counts Tickets for Core Achievements


Tsearo:
Currently the site will only display total core tickets and will not count tickets in unofficial towards a users total. I believe these should be counted now as we are able to see unofficial tickets in the ticket manager now thanks to MrOwnership.

To update the count on the users page and logged in user card (top right) the following change needs to be made:
`AND ach.Flags = '3'` -> `AND ach.Flags IN('3','5')`
https://github.com/RetroAchievements/RAWeb/blob/5b28158ef29497bc9f4a1eeff99013ccca0af3f9/lib/database/ticket.php#L419

The part I haven't been able to find a good solution to is on the DeveloperStats page. It needs to count ach.Flags 3 (Core) and 5 (Unofficial) but it can't simply be added here due to only wanting to count Achievements in core for that total. I'm not too great at SQL so it may be a simple fix but I don't currently have one.
https://github.com/RetroAchievements/RAWeb/blob/5b28158ef29497bc9f4a1eeff99013ccca0af3f9/lib/database/user.php#L1600-L1622

JeffJenkins2:
@Tsearo You probably can query for both flags and use a CASE statement in the COUNT call (e.g. `COUNT(DISTINCT CASE WHEN ach.Flags = 3 THEN ach.ID ELSE NULL END)`. That should get the ticket count to include both flags, but the cheevo count to only include one flag. If that sample DB has some data that I can test that logic against, I can give it a go.

PR #630, opened 2021-07-05
fix: #595 include unofficial achievements in ticket counts, but not achievement counts

JeffJenkins2:
This adjusts the ticket counts to include both core and unofficial flags while not affecting the achievement counts (only includes core).


MagmarFire (approved):

MagmarFire (approved):
I think it would be worth our while to use named constants wherever possible instead of magic numbers, at least for readability's sake. This is standard fare for easy V1 changes, though, so I'll approve this.

Issue #384, opened 2019-12-27
Forum: Code doesn't ignore forum formatting or style tags


ViperZang:
When using a code block in the forums, I would think it should be styled as plain text, and not be able to use the tags within them. I just noticed when testing after backslashes weren't weren't being displayed ~while using monospace ascii art~ inside code.

In the following image, the first spoiler tag probably shouldn't be displayed as an actual spoiler tag, but as the regular text because it is inside a code block:

![code-backslash](https://user-images.githubusercontent.com/31069034/71528713-344beb80-28af-11ea-9450-9daeec2da159.PNG)

Example of how code blocks should probably work:
`![code-backslash](https://user-images.githubusercontent.com/31069034/71528713-344beb80-28af-11ea-9450-9daeec2da159.PNG)`

PR #636, opened 2021-07-06
fix: #384 encode tags within code snippets

JeffJenkins2:
This will convert tags within code snippets so they don't get parsed and replaced, resulting in formatted text (bold, italic, etc.) within preformatted code blocks.

This will not address nested code snippets (i.e. `[code]test [code]...[/code] test[/code]`) due to the way the parser is written.


MagmarFire (approved):

luchaos:
There's a `test/bbcode.php` file where I'd like to see use cases implemented for bbcode/shortcode changes.
Changes in the parser have proven to be problematic in the past.
May I ask you to add examples to said file to make it easier for me/others to reproduce the issue?

Issue #521, opened 2020-11-21
Require users to select Hardcore or Softcore when submitting tickets. 


televandalist:
There are few things more irritating than spending time troubleshooting a ticket only to find out that the user was playing on softcore and hits weren't saving/loading for them. 

This would also help on the manual unlock end.

luchaos:
> Softcore not storing hits hasn't been relevant for some time on anything other than Duckstation standalone.

Still relevant?

luchaos:
nevermind. added in https://github.com/RetroAchievements/RAWeb/pull/673

PR #673, opened 2021-08-06
Force user to choose mode for a ticket (hardcore/softcore) and allow filter tickets by it

Xymjak:
Issue: https://github.com/RetroAchievements/RAWeb/issues/521

Video demo: https://cdn.discordapp.com/attachments/657437326229897226/873181231838277642/1.mp4
(ignore "old" - it's from another PR-testing)

Need to run "Add_Ticket_Hardcore.sql" before.


Xymjak:
I have no idea what php-cs-fixer wants. I manually run it on every file one by one - it said "fixed" without changing anything...

database/20210806_000000_Add_Ticket_Hardcore.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `ticket` ADD `Hardcore` TINYINT(1) AFTER `ReportType`;
```
luchaos, line 1: I see `Hardcore` is later used as `mode` - I think calling it `Mode` here would be more consistent?

luchaos (approved):
Not too happy with the mixed boolean flag for hardcore and 1/2 for mode but the feature itself seems to work just fine.

Xymjak:
Oops, only now noticed it's somehow alive (I had 400+ unread inbox messages on that "spam"-e-mail, can't find time to read. ><)

> and 1/2 for mode

Should had replaced it, it was quickly braindeadly coded in ~1 hour while being sleep deprived without any rechecking later because "guess it's dead".

> seems to work just fine

I thought it became broken by another PR because I tried to mimic the current site's codestyle... which uses explicit numbers  which, when I studied programming ~20 years ago, was always called as a bad code and can easily cause conflicts.

If I coded the way I like, I would delete the existing hundreds lines of code and replaces with automatic bitflag generation inside object, UI-generation via foreach-loop through the filters-field of that object instead of explicit, and filter-request generated through a function of that object again with internal foreach. That way:
1.  multiply people can add new filters at the same time without any chance of breaking others code (which is easy now)
2. no need to manually search "where need to change bitflags" which reduces both time to add new filters and chances to introduce bugs
3. the code is much shorter without dozen of "copypaste"-parts which is easier/faster to read (for me) and again makes it harder to break

But I'm used to more low-level/performance/math-based code and I'm not an experienced PHP-coder and seems like here people are scared of OOP.

PR #674, opened 2021-08-08
Links to supplementary ticket info are now provided after certain fields

### LINKED ISSUE #524: Add info and disclaimers to ticket report page.
There really should be some info on the ticket report page rather than just the fields for the report itself. 

- Information on finding and reporting the correct hash, emulator versions, etc. and why it's important. Not everyone is a supernerd, after all. 
- Disclaimer saying "Doesn't work" is about the most worthless thing you can add to a ticket with a "Doesn't trigger" issue. These are known as the "No shit?" tickets. Maybe word it more politely than that. 

Overall, it should all have a "help us help you" approach. Better tickets lead to less disgruntled devs, which leads to more and better achievements, which leads to happiness and rainbows for everyone.

MagmarFire:
Certain fields now have clickable links that will bring up a modal that gives important information for that given field. I have also added an Emulator Version field to help make tickets more useful, especially in light of recent updates making save states store hit counts.

image

image

I also added in an "Easter egg" of sorts that will prevent a user from submitting the ticket if the description has nothing but something along the lines of, "Doesn't work," prompting them to put in something more substantial, if the description is shorter than 25 characters. This won't prevent someone from making a long-ass description that says absolutely nothing, but I figure it will help educate people inclined to lazily state the obvious to be more thoughtful.

image

Intended to close #524.

### DIFF
--- a/lib/render/layout.php
+++ b/lib/render/layout.php
@@ -32,13 +32,6 @@ function RenderHtmlHead($pageTitle = null)
 
 function RenderSharedHeader()
 {
-    echo "<link rel='stylesheet' href='/css/styles.css?v=" . VERSION . "' media='screen'>\n";
-
-    $customCSS = RA_ReadCookie('RAPrefs_CSS');
-    if ($customCSS !== false && mb_strlen($customCSS) > 2) {
-        echo "<link rel='stylesheet' href='$customCSS?v=" . VERSION . "' media='screen'>\n";
-    }
-
     echo "<link rel='icon' type='image/png' href='/favicon.png'>\n";
     echo "<link rel='image_src' href='/Images/RA_Logo10.png'>\n";
     echo "<meta http-equiv='content-type' content='text/html; charset=UTF-8'>\n";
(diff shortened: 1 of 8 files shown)


public/reportissue.php
```diff
@@ -68,57 +68,65 @@ function displayCore() {
...
                             <?php endforeach ?>
                         </select>
                     </td>
                 </tr>
-                <tr id="core-row" style="display: none">
+                <tr>
+                    <td><label for="version">Emulator Version</label></td>
+                    <td>
+                        <input type="text" name="note[emulatorVersion]" id="version" required />
+                        <a href="/views/versionDescriptionModal.html" rel="modal:open">Why?</a>
```
luchaos, line 116: I'd prefer inlining those modals instead of separate files. Not too sure about that though.

public/reportissue.php
```diff
@@ -68,57 +68,65 @@ function displayCore() {
             <table>
                 <tbody>
                 <tr>
-                    <td>Game:</td>
+                    <td>Game</td>
```
luchaos, line 71: Less noise - I like it

lib/render/layout.php
```diff
@@ -62,11 +55,23 @@ function RenderSharedHeader()
     echo "<script src='https://ajax.googleapis.com/ajax/libs/jquery/1.8.1/jquery.js'></script>\n";
     echo "<script src='https://ajax.googleapis.com/ajax/libs/jqueryui/1.8.23/jquery-ui.js'></script>\n";
 
+    // jQuery Modal
+    echo "<script src='https://cdnjs.cloudflare.com/ajax/libs/jquery-modal/0.9.1/jquery.modal.min.js'></script>";
+    echo "<link rel='stylesheet' href='https://cdnjs.cloudflare.com/ajax/libs/jquery-modal/0.9.1/jquery.modal.min.css' />";
+
+    echo "<script src='https://cdnjs.cloudflare.com/ajax/libs/knockout/3.5.0/knockout-min.js'></script>";
```
luchaos, line 62: Do we need knockout here?

lib/render/layout.php
```diff
@@ -62,11 +55,23 @@ function RenderSharedHeader()
...
+
     //    jQuery, and custom js
     //echo "<script src='//ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js'></script>\n";
     //echo "<script src='/vendor/jquery-ui-1.10.2.custom.min.js'></script>\n";
     echo "<script src='/js/all.js?v=" . VERSION . "'></script>\n";
     echo "<script>window.assetUrl='" . getenv('ASSET_URL') . "'</script>\n";
+
+    echo "<link rel='stylesheet' href='/css/styles.css?v=" . VERSION . "' media='screen'>\n";
```
luchaos, line 70: I there a technical reason for moving the styles to the bottom of the list? web vitals?

luchaos (changes requested):

public/reportissue.php
MagmarFire: Strong disagree there. It's way better to split up views to make them easier to digest for future devs and even for people like me, who like to compartmentalize everything into small files.

public/reportissue.php
MagmarFire: Thank you! That's the plan. :)

lib/render/layout.php
MagmarFire: It's way better to use than jQuery, I can say that much. Leads to much less spaghetti that jQuery typically leads to and is generally less buggy and easier to cover all edge cases.

lib/render/layout.php
MagmarFire: Yep, sure is. We don't want vendor CSS to override site CSS in such a way that requires us to use `!important`, which I consider bad code smell if used willy nilly. The best way to prevent this is to include our site CSS after all vendor CSS.

And in this case, the practical reason is because if I don't do this, the menu bar at the top of the site overlaps the jQuery modals and makes them look *really* funky. The only way to fix it without reorganizing the CSS was to use `!important` on the menu's z-index, but that felt really hacky to me.

televandalist:
Merge wen

PR #681, opened 2021-08-18
clearer login error message

### LINKED ISSUE #678: [Request] Retroarch integration / Provide better feedback to application when login token has expired
When logging in to Retroachievements, Retroach initially uses your username/password, but then saves a token to its configuration file for future logins. What I have noticed, is if you do not log in for some time (turns out to be two weeks), the token expires. At that point, the user would need to log in with their password once again, which in terms of Retroach workflows means they need to delete the data from cheevos_token and re-add their password to cheevos_password in their retroarch.cfg file. From my testing, it looks like Retroarch will actually remove the token from its coniguration file when this occurs.

So the problem is it is not clear to the user that the expired token is the problem, and that the next step would be to confirm their token is cleared from retroarch.cfg and that they now need to re-enter their password.

I would propose that when the token is expired, a better message back is provided (I think it is coming [from here](https://github.com/RetroAchievements/RAWeb/blob/51d72fffe421e65b94616858ce0b10459e6cdee0/public/dorequest.php#L114)) and the message is currently: "Error with login! Please try again."

Thanks! Let me know if you need any more information or testing on my part.

meleu:
closes #678

### DIFF
--- a/public/dorequest.php
+++ b/public/dorequest.php
@@ -111,7 +111,7 @@ switch ($requestType) {
             /**
              * Token invalid or out of date
              */
-            DoRequestError("Error with login! Please try again.");
+            DoRequestError("Login error! Check your credentials and try again.");
         }
         break;
 
--- a/public/dorequest.php
+++ b/public/dorequest.php
@@ -111,7 +111,7 @@ switch ($requestType) {
             /**
              * Token invalid or out of date
              */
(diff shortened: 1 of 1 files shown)


Jamiras:
Maybe "Login failed.". There wasn't really an error.

Jamiras (approved):

PR #682, opened 2021-08-21
Refactor and add filtering to the "Active Players" widget

MagmarFire:
In a random bout of inspiration, I took it upon myself to refactor and improve the "Active Players" widget. In its current state, it is a cool little thing that can give you an idea of who's playing what at what time, which is especially useful information if you want to know at a glance whoever's playing a set you devved. However, the problematic thing is having to sift through games you simply don't care about in order to find the ones you do. My changes aim to eliminate that problem completely.

At first glance, it looks pretty much the same, except for a hamburger button at the right:

image

Clicking on it reveals a hidden submenu (and it can be hidden simply by clicking it again or by refreshing the page):

image

With this, the user can type in a string (case insensitive) and filter every result according to how they want--whether it be the user's name, the game title, the Rich Presence, or even the console the game was released for. Additionally, if the user checks "Remember My Filter", the browser will remember the string the user typed in subsequent sessions, until the user either deletes the filter or unchecks the box.

image

To give an even greater degree of control, users can also add OR operations to their filter, courtesy of the `|` operator.

image

This widget could possibly be improved in the future quite easily, too, due to the refactoring of old code to code that's much easier to maintain.


luchaos:
Very interesting feature! 👍  
I'll try it locally and review in the upcoming days.

Tsearo:
Beautiful <3 This will take care of being able to search for players developing achievements as well which was something I had wanted to try adding at in the past 👍

luchaos (approved):
As discussed on Discord a while ago this is fine.
I'm still reluctant to add more dependencies or stack changes to this codebase than is absolutely necessary but maybe knockout and lodash will come in handy for maintenance after all.
Let's have it.

PR #819, opened 2021-09-06
YouTube URLs placed in [url] tags are now properly rendered as links

### LINKED ISSUE #818: Forum Issue: YouTube videos break BBCode URL tag
currently, I you put a youtube link with no BBCode in the forums, it automatically turns the video into an embed.
However, using the BBCode tag for URL links causes the video to try to use YouTube embed features, which breaks the BBCode. I don't know what the backend stuff looks like, but from a user POV, it seems to be trying to convert the url into YouTube's iframe embed code. 

**Supplemantary information**
---

`[url=https://youtu.be/dMH0bHeiRNg]v2 wen[/url]` (as well as the long `https://www.youtube.com/watch?v=dMH0bHeiRNg` links)
will display as
image

---

YouTube's sharing  feature gives this as embed code:
```
<iframe width="560" height="315" src="https://www.youtube.com/embed/dMH0bHeiRNg" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
```

---

Not sure if it help, but here's the YouTube iframe API stuff
https://developers.google.com/youtube/iframe_api_reference

MagmarFire:
Fixes #818

### DIFF
--- a/lib/util/bbcode.php
+++ b/lib/util/bbcode.php
@@ -172,6 +172,12 @@ function linkifyYouTubeURLs($text)
     // https://www.youtube.com/watch?v=1YiNYWpwn7o
     // www.youtube.com/watch?v=Yjba9rvs4iU
 
+    // Don't linkify this text if it's a shortcode construct.
+    $linkedPattern = '/\[.*=.*\].*\[\/.*\]/';
+    if (preg_match($linkedPattern, $text) == 1) {
+        return $text;
+    }
+
     $pattern = '~
         # Match non-linked youtube URL in the wild. (Rev:20130823)
         (?:https?://)?    # Optional scheme. Either http or https.


lib/util/bbcode.php
```diff
@@ -172,6 +172,12 @@ function linkifyYouTubeURLs($text)
     // https://www.youtube.com/watch?v=1YiNYWpwn7o
     // www.youtube.com/watch?v=Yjba9rvs4iU
 
+    // Don't linkify this text if it's a shortcode construct.
+    $linkedPattern = '/\[.*=.*\].*\[\/.*\]/';
+    if (preg_match($linkedPattern, $text) == 1) {
+        return $text;
+    }
```
luchaos, line 179: I haven't given this a run yet but from how it reads this would omit parsing _any_ youtube urls if there is at least one (url) shortcode construct with an attribute in the whole text?
I assumed that the pattern below has to be extended so that may not be preceded by indication of an enclosing shortcode.
The same should be applied to the other embed media parsers (twitch, imgur) as those would cause a similar problem.

MagmarFire: I'm not sure what you mean. This code exists here because there's no reason to linkify text that exists in a tag. The function seems to me to only be for wild links, so it stands to reason that it should return early if the input is not a wild link.

Are you suggesting we apply the same fix to the other linkify functions, or are you suggesting I modify the original monster regex? Doing the latter is, in my opinion, unnecessary and rather dangerous (after all, [we know the old adage about those who use regular expressions to solve problems](https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/)).

luchaos: Gave this another run to see what I meant there - can be confirmed when opening `/test/bbcode.php` in the browserfor smoke testing. It will not render any YouTube embeds anymore when a shortcode pattern can be found in the whole text that is to be parsed.

It doesn't do that line by line or for each match (globally, greedy).

I'll add the example from https://github.com/RetroAchievements/RAWeb/issues/818 to this PR and see if i can make it work with the existing monster regex.

I agree that it's dangerous to fiddle with regexes but we shouldn't have regex to parse bbcode/shortcode in the first place. Just trying to deal with what we already have :) Going by the simple test page should give at list _some_ confidence that nothing else breaks.

luchaos: Turns out we already have a negative lookahead in that regex which works when it finds a youtube url within an `<a>` tag.
All it's missing is a negative lookahead for shortcodes, too.

https://github.com/RetroAchievements/RAWeb/blob/master/lib/util/bbcode.php#L187-L193

I'll fix that for all the embed options.

luchaos:
Closing in favor of https://github.com/RetroAchievements/RAWeb/pull/870. Was a bit more work after all.

Issue #426, opened 2020-02-14
Notify when exceeding character limit in forum posts


Sanaki:
Currently you find out you exceeded the forum post size limit by hitting submit and having your post get cut off. Adding a notice of some sort would be appreciated.

JeffJenkins2:
It looks like the limit for forum posts is 65000 bytes and the text field for adding a comment has a maxlength attribute of 60000.

I don't see how people are submitting posts that are lengthier than those limits. When I tried pasting a 70000 byte text string of lorem ipsum text into the field, the text was cut off at the 60000 limit and I wasn't able to enter any more content into the text field.

Can you provide an example of content that is being truncated unexpectedly?

luchaos:
Suggestion: add a character counter to textareas

Sanaki:
@JeffJenkins2 Sorry, somehow your message slipped through my notifications. Regardless, this is where I encountered the issue:
https://retroachievements.org/viewtopic.php?t=9973

PR #823, opened 2021-09-08
feat: #426 add textarea character counter

JeffJenkins2:


MagmarFire (commented):
I wonder if we can simplify the implementation somewhat and decouple some of the view logic and business logic if we use Knockout.

But I'm biased, of course, so this can still suffice.

MagmarFire:
I thought that AlpineJS was what was going to be used primarily for V2, not V1. But besides that, your point is sound.

'Twas only a suggestion, so I'll go ahead and approve this.

MagmarFire (approved):

luchaos:
> If the team feels that it is imperative to build a KO ViewModel to manage the view state for these character counters on a deprecated code base, then I will have no problem refactoring the code. I simply don't think that using KO for this situation is going to have any significant impact on performance or maintainability.

not at all. I share your thoughts about knockout, alpine.js, jquery. 
it's perfectly fine to still use jquery in this projects imo as it's been built with it. 

i think i would've preferred alpine.js instead of knockout, too - but that's not important.
to me it doesn't really matter anymore in this codebase tbh. let's keep maintenance of this codebase to a minimum.

PR #825, opened 2021-09-12
fix incorrect key for hardcore check in gamecompare page

### LINKED ISSUE #204: Compare with user doesn't show hardcore indicators for second user
When using the "compare with user" feature, the user's achievements don't have gold borders, the "-= HARDCORE =-" header, and the summary doesn't indicate bonus completion.

image

The user has earned the achievements in hardcore mode:

image

Also note that _my_ completion on the "compare with user" page shows as 118/59 and 800/400, but the other user's profile completion shows 59/59 and 400/400. These should both display the same value. My suggestion is to modify the "compare with user" to also show 59/59 and 400/400, or possibly modify both to show 59/59 and 800/400.

Jamiras:
fixes #204

image

also no longer counts hardcore achievements as two unlocks

image

### DIFF
--- a/public/gamecompare.php
+++ b/public/gamecompare.php
@@ -149,13 +149,12 @@ RenderHtmlHead("Game Compare");
                 $awardedLeft = isset($nextAch['DateEarned']) ? $nextAch['DateEarned'] : null;
                 $awardedRight = isset($nextAch['DateEarnedFriend']) ? $nextAch['DateEarnedFriend'] : null;
                 $awardedHCLeft = isset($nextAch['DateEarnedHardcore']) ? $nextAch['DateEarnedHardcore'] : null;
-                $awardedHCRight = isset($nextAch['DateEarnedHardcoreFriend']) ? $nextAch['DateEarnedHardcoreFriend'] : null;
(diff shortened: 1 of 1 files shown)


public/gamecompare.php
```diff
@@ -189,8 +188,7 @@
                         echo GetAchievementAndTooltipDiv($achID, $achTitle, $achDesc, $achPoints, $gameTitle, $badgeName, true, true, "", $iconSize, "goldimage awardremote");
                         echo "</div>";
                         $rightAwardedCount++;
-                        $rightAwardedCount++;
-                        $leftAwardedPoints += $achPoints;
+                        $rightAwardedPoints += $achPoints;
```
MagmarFire, line 191: Why are we duplicating this line?

MagmarFire (changes requested):

public/gamecompare.php
Jamiras: Hardcore badges are worth double points. This is actually changing one instance of `leftAwardedPoints` to `rightAwardedPoints` as the hardcore points were being attributed to the wrong user. You can see the duplication for the left user on lines 159/160.

MagmarFire: That's a good point! Approved.

MagmarFire (approved):

PR #829, opened 2021-09-22
Validate PSP

televandalist:
Don't deploy till Oct 1st!


ScottFromDerby (commented):
Ooo it's exciting

MagmarFire (approved):
I realized just how many magic constants we have here. :S

luchaos:
> I realized just how many magic constants we have here. :S

oh yes 😅  those console/system IDs have to match the list in the emulators. so one or another mapping has to happen in any case.

luchaos (approved):

PR #835, opened 2021-09-30
Load minified jQuery scripts

coliff:
= smaller file sizes and faster downloads

REF: https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fretroachievements.org%2F


luchaos (approved):

luchaos (approved):
Good catch, thank you!

PR #838, opened 2021-10-04
Devs can now close a ticket as a mistaken report

### LINKED ISSUE #837: [Enhancement] Give developers the ability to close a ticket as a mistaken report, even if they're not the ticket creator
As a developer, I want to be able to close a ticket as a mistaken report, even if I'm not the ticket creator, so that I don't have to close a ticket for "another reason" just to state that it's a mistaken report in the event the reporter doesn't close the ticket themselves or doesn't even know how to.

MagmarFire:
Fixes #837


public/ticketmanager.php
```diff
@@ -742,17 +742,14 @@
                     echo "<select name='action' required>";
                     echo "<option value='' disabled selected hidden>Choose an action...</option>";
                     if ($reportState == 1) {
-                        if ($user == $reportedBy) { // only the reporter can close as a mistaken report
-                            echo "<option value='closed-mistaken'>Close - Mistaken report</option>";
-                        }
```
Jamiras, line 747: Shouldn't the reporter still be able to select this option, even if they're not a dev?

MagmarFire: That's a good catch.

Jamiras:
It'll create some duplication, but I think it's important to put the option back where you had it initially. We don't want it to be the first option in the dev's list, and probably not the last ("other" implies no further options apply). I'd recommend something like:
```
if (dev) {
   options
} else if ($user === $reportedBy) {
   option
}
```

Jamiras (approved):

PR #843, opened 2021-10-17
don't hide forum topics if last post is from an unverified user

Jamiras:
Instead of filtering on forum topics where the last post is unauthorized, this modifies the query to count the number of authorized posts in a forum topic and ignores anything without any authorized posts.

The list will still show the unauthorized user as the user for the last post, but upon opening the topic, the unauthorized posts will remain hidden.
image

The unauthorized links page still shows the full post count:
image

The unauthorized post also does not appear in the Forum Activity list on the main page, or the Forum Post History. No change was necessary for that behavior.

### DIFF
--- a/public/viewtopic.php
+++ b/public/viewtopic.php
@@ -252,7 +252,7 @@ RenderHtmlStart();
 
             if ($nextCommentAuthorised == 0) {
                 // Allow, only if this is MY comment (disclaimer: unofficial), or if I'm admin (disclaimer: unofficial, verify user?)
-                if ($permissions >= Permissions::Developer) {
+                if ($permissions >= Permissions::Admin) {
                     // Allow with disclaimer
                     $showDisclaimer = true;
                     $showAuthoriseTools = true;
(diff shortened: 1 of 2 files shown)


public/viewtopic.php
```diff
@@ -252,7 +252,7 @@
 
             if ($nextCommentAuthorised == 0) {
                 // Allow, only if this is MY comment (disclaimer: unofficial), or if I'm admin (disclaimer: unofficial, verify user?)
```
luchaos, line 254: This comment doesn't seem to align with the implementation, doesn't it?

Jamiras: I'm not sure what the "disclaimer: unofficial" means, but the "MY comment" aligns with line 259, and the "I'm admin" aligns with line 255 (after my change).

Jamiras: I'm guessing it's supposed to be "disclaimer: unverified" to indicate the disclaimer will be shown.

luchaos: Ah, yes - I see it now, my bad

luchaos (approved):

PR #844, opened 2021-10-18
Update viewtopic.php

televandalist:
Changes the forum post placeholder text from `Enter a comment here...` to `Enter a comment here... and please do not share links to ROMs...`

### DIFF
--- a/public/css/styles.css
+++ b/public/css/styles.css
@@ -1002,6 +1002,15 @@ th {
 }
 
 /* Forum */
+
+#commentTextarea {
+  padding: 4px;
+}
+
+#commentTextarea::placeholder {
+  font-size: 1.1em;
+}
+
 .commentpayload {
   padding: 6px 4px;
   vertical-align: top;
--- a/public/editpost.php
+++ b/public/editpost.php
@@ -74,7 +74,7 @@ RenderHtmlHead("Edit post");
 
         RenderPHPBBIcons();
 
-        echo "<textarea id='commentTextarea' class='fullwidth forum' style='height:300px' rows='32' cols='32' name='p' placeholder='Enter a comment here... and please do not share links to ROMs...'>$existingComment</textarea></td></tr>";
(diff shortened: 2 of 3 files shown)


Jamiras:
Conflict needs to be resolved, and it looks like you were planning to make additional changes:
> I'll change it

luchaos:
Not sharing links to copyrighted ROMs is just one of many things that a user should be aware of when posting _anywhere_ on the site. A generalized info line with a link to the terms/code of conduct might serve this purpose better?

televandalist:
> 

You'd think it would, but I'm pretty sure people would still do it. I figured having a reminder in the placeholder comment wouldn't hurt.

luchaos:
I know what you mean - and I'm probably taking the wrong approach to this.
does this mean we have to single out the rule of not being a dick, too? and translate it all to pt-br to make sure most of our userbase understands?
I'd rather hold our userbase to high standards. when signing up they agree to _some_ terms of use. they all better read the damn whole thing and comply to it. this includes one of the better known legal issues of not sharing links to copyright protected ROMs.
we have rules and they have to be enforcable - there are many other things that are a bannable offense. otherwise moderation has nothing.
i skip a lot of terms texts myself for close to 30 years now. doesn't mean i don't expect to be moderated _hard_ when i don't comply by accident.
personally I'm against simplifying our user interface for that single issue but i can see where a reminder for that wouldn't hurt. guess I'm just puzzled by how little we can expect these days.
let's add it then

luchaos (approved):

Jamiras:
This still has a conflict, and I think I agree with luchaos's comments. It's a slippery slope when we start calling out specific rules and not others. And given that it appears as a watermark, I'm guessing most users won't even read it.

FYI, this is what it looks like:
image

I seem to recall a brief discussion about removing the "please" as well. Having that there implies it would be okay to do, just frowned upon.

luchaos:
Screenshot 2021-12-08 at 01 04 09
Screenshot 2021-12-08 at 01 04 44

luchaos (approved):
Changed as discussed in staff meetings.
The simplest solution for now.
To be revisited when we see fit.

PR #847, opened 2021-11-05
Prevent blocked users from sending private messages

### LINKED ISSUE #813: "Instant and permanent friends"
### Discussed in https://github.com/RetroAchievements/RAWeb/discussions/691

<div type='discussions-op-text'>

<sup>Originally posted by **JuliaSalsa** February 17, 2018</sup>
When you are requesting a friend, they are automathically added to friend leaderboard, from which you can't remove them.</div>

Jamiras:
Removes the "Send Private Message" link from a user's page when that user has blocked the logged in user. Also removes the ability to post comments on that user's page.

Additionally, removes blocked users from the "compare with friends" lists on the game page and a couple of API calls.

Fixes #813. Closes #691

### DIFF
--- a/lib/render/error.php
+++ b/lib/render/error.php
@@ -26,6 +26,7 @@ function RenderErrorCodeWarning(?string $errorCode)
         'issue_submitted' => "Your issue ticket has been successfully submitted.",
         'merge_failed' => "Problems encountered while performing merge. These errors have been reported and will be fixed soon... sorry!",
         'merge_success' => "Game merge successful!",
+        'messageblocked' => "Message was not delivered - blocked by recipient.",
         'modify_game_ok' => "Game modify successful!",
         'modify_ok' => "Info: Modified OK!",
         'newadded' => "Friend request sent.",
--- a/lib/render/error.php
+++ b/lib/render/error.php
@@ -26,7 +26,6 @@ function RenderErrorCodeWarning(?string $errorCode)
         'issue_submitted' => "Your issue ticket has been successfully submitted.",
         'merge_failed' => "Problems encountered while performing merge. These errors have been reported and will be fixed soon... sorry!",
         'merge_success' => "Game merge successful!",
-        'messageblocked' => "Message was not delivered - blocked by recipient.",
         'modify_game_ok' => "Game modify successful!",
         'modify_ok' => "Info: Modified OK!",
         'newadded' => "Friend request sent.",
--- a/lib/database/achievement.php
+++ b/lib/database/achievement.php
@@ -1052,7 +1052,7 @@ function getAchievementRecentWinnersData($achID, $offset, $count, $user = null, 
 
     $extraWhere = "";
     if (isset($friendsOnly) && $friendsOnly && isset($user) && $user) {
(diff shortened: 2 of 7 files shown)


ScottFromDerby:
Thank-you for doing this <3

ScottFromDerby (approved):

ScottFromDerby:
checked all changes, I'd be happy to approve but luchaos may want the final say

lib/render/error.php
```diff
@@ -26,6 +26,7 @@ function RenderErrorCodeWarning(?string $errorCode)
         'issue_submitted' => "Your issue ticket has been successfully submitted.",
         'merge_failed' => "Problems encountered while performing merge. These errors have been reported and will be fixed soon... sorry!",
         'merge_success' => "Game merge successful!",
+        'messageblocked' => "Message was not delivered - blocked by recipient.",
```
luchaos, line 29: Not sure of the implications. Do we want to expose that a user blocked another?

Jamiras: I'm not sure what the correct behavior here is. It seems the most common implementation is to pretend it worked, without actually doing anything.

luchaos: Yeah, that's my thought, too. Otherwise it might be seen as a privacy issue (I would).
I'd say we have a a look at other platforms, how they deal with blocked messages.

Jamiras: From a brief trip through Google search results, it seems like there's really only two common implementations:
1) All activity related to the blocking user is removed from the blocked user's visibility.
    - As far as the blocked user knows, the blocking user no longer exists.
2) The blocked user has no way of knowing they've been blocked. 
    - The blocking user still exists, but does not receive any direct communication from the blocked user. 
    - The blocked user can only assume they're being ignored by the blocking user, but in reality they're being ignored by the site.
- In some cases, the blocked user is also scrubbed from the blocking user's visibility. Either fully, or with spoiler-like tags so the blocking user can choose to see the blocked user's content.

The provided implementation doesn't really fit into either of those.

The ability to directly message the user was explicitly removed, making it obvious that they've been blocked. The ability to post on the blocking user's wall was also removed, but the blocked user could attribute that to the user disabling wall comments.

The highlighted point of contention comes from the Inbox implementation. Once you reply to any Inbox message, you can change the recipient (it even updates the user's image when doing so!), so you could reply to another user and still target the blocked user. This change generates an error at the last possible moment in case any other workarounds exist.

I'll change the code to return `sentok` in this situation.

As the "Send Private Message" directs to the same page as the Reply functionality, it could be handled in the same way to hide the fact that the user has been blocked. The only remaining indication would be the lack of an "Add Friend" button.

luchaos: Thanks for going in-depth with this.

> I'll change the code to return `sentok` in this situation.

Seems to be the most sensible solution for now, I concur.

> The only remaining indication would be the lack of an "Add Friend" button.

Still an indicator but i guess that's fine for now, too.
It's mainly the `messageblocked` hint that I'd want to avoid - the reasons for not delivering a message might be varying - exposing this one is very explicit.

luchaos (approved):

PR #849, opened 2021-11-13
Added configuration for docker setup that allow to use Xdebug in your IDE

anegoda1995:
I created it for PhpStorm and tested in 2021.2.3
It will be good if someone have idea how to modify my work to use in different IDE's


MagmarFire (commented):
Have we confirmed that this wouldn't break other people's environments with XDebug already installed (as I myself have installed manually for use in Visual Studio Code)? The pull request comment stating that it would be "nice to modify [your] work" leads me to believe that this may not be tested with different set ups, so I'm nervous to approve this right away.

luchaos (commented):
It looks like this configuration will have xdebug running with each and every request which is not good practice at it will slow down everything.
The on-demand setup with the browser extension is preferred to only have xdebug run when desired.
I'll validate the configuration with my own Docker / PHPStorm setup.

luchaos (approved):
Thanks for the PR!

Reduced the xdebug configuration to the minimum required.

Enabling xdebug is optional and can be enabled via JetBrain's browser extension:
https://www.jetbrains.com/help/phpstorm/2022.1/browser-debugging-extensions.html

Ran docker composer build to have xdebug installed. Works both via docker and bare metal without conflicts (PHPStorm server configuration only takes one, switching requires deleting the old configuration but that's a PHPStorm issue imo).

PR #850, opened 2021-11-14
Fix first place display when an Untracked user is tied

Tsearo:
If the No.1 spot of a leaderboard is tied with an untracked user it will still display the untracked user if they earned it first. 

Adding an extra check to remove untracked users from the leaderboard entry results to prevent this as the untracked user check above in the query is for finding the best tracked user score.

Example - Highlighted user is Untracked but displays in the No.1 spot in the right panel
image


Jamiras (approved):

luchaos:
> I added a `|| ua.User is null` check to the inner WHERE clause to allow leaderboards with no entries 
> 
> That will have to be added to this change as well. I'm not sure if you want to do that here in this PR, or I can add it to mine after yours gets merged.

Should be included here, too.

PR #852, opened 2021-11-18
Added configuration for docker setup that allow to use Mailcatcher for catching every email

anegoda1995:
To test this PR you can just create new account (/createaccount.php) and open http://localhost:64126/


anegoda1995:
To test this PR you can just create new user and open http://localhost:64126/

lib/util/mail.php
```diff
@@ -104,6 +104,42 @@ function mail_utf8($to, $from_user, $from_email, $subject = '(No subject)', $mes
...
+
+        $message = (new Swift_Message($subject))
+            ->setFrom([$from_email => $from_user])
+            ->setTo($to)
+            ->setBody($message);
+
+        return $mailer->send($message);
+    }
+
+    if (getenv('MAIL_DRIVER') === 'mailcatcher') {
```
meleu, line 119: These repeated if blocks are intentional?

anegoda1995: Yes, SMTP Mailcatcher connection only if **mailcatcher** driver is in _.env_

Jamiras: But why three times (107-117, 119-129, 131-141)? And if three times are necessary, why not use a loop?

anegoda1995: Sorry for three times repeated **if** blocks. First message from @meleu  was not understood correctly, because I don't see whole part of code in discussion. Already fixed, thank you!

anegoda1995: Sorry for three times repeated **if** blocks. First message from @meleu  was not understood correctly by me, because I don't see whole part of code in discussion.

luchaos:
> Package swiftmailer/swiftmailer is abandoned, you should avoid using it. Use symfony/mailer instead.

I haven't used mailcatcher yet. We'll use mailhog for the Laravel iteration in the future (see https://github.com/laravel/sail/blob/1.x/stubs/mailhog.stub)

I'd prefer to switch both the dependency and mail tool to those.

lib/util/mail.php
```diff
@@ -104,6 +104,18 @@ function mail_utf8($to, $from_user, $from_email, $subject = '(No subject)', $mes
         return mail_ses($to, $subject, $message);
     }
 
+    if (getenv('MAIL_DRIVER') === 'mailcatcher') {
+        $transport = (new Swift_SmtpTransport('mailcatcher', 1025));
```
MagmarFire, line 108: I'm not sure the outer parentheses for this initialization are really necessary, but I won't hold the review back over it.

MagmarFire (approved):

.env.example
```diff
@@ -36,7 +38,7 @@ DB_PASSWORD=retroachievements
 #SESSION_DOMAIN=.retroachievements.org
 #SESSION_SECURE_COOKIE=true
 
-MAIL_DRIVER=sendmail
+MAIL_DRIVER=sendmail #{sendmail || ses || mailcatcher}
```
luchaos, line 41: Comments in .env files should be on their own line for good practice. There are other .env parsers that will take the comment as part of the value.
```
 # sendmail, ses, mailcatcher
MAIL_DRIVER=sendmail
```

composer.json
```diff
@@ -18,7 +18,8 @@
         "friendsofphp/php-cs-fixer": "^3.0",
         "phpunit/phpunit": "^9.0",
         "phpstan/phpstan": "^0.12.18",
-        "vimeo/psalm": "^4.0"
+        "vimeo/psalm": "^4.0",
+        "swiftmailer/swiftmailer": "^6.3"
```
luchaos, line 22: [Swiftmailer is deprecated](https://github.com/swiftmailer/swiftmailer). We should use [Symfony Mailer](https://symfony.com/doc/current/mailer.html) instead.

lib/util/mail.php
luchaos: Can be removed, agreed.

.env.example
```diff
@@ -5,6 +5,8 @@ HTTP_PORT=64100
 HTTPS_PORT=64101
 MINIO_PORT=64140
 PHPMYADMIN_PORT=64180
+MAILCATCHER_PORT_SMTP=64125
+MAILCATCHER_PORT_HTTP=64126
```
luchaos, line 9: I'd like to use Mailhog instead of Mailcatcher if there are no objections.
 I'm open for suggestions though, are there specific benefits compared to Mailhog?

lib/util/mail.php
```diff
@@ -83,14 +79,12 @@ function mail_ses($to, $subject = '(No subject)', $message = '')
         $promise = $pool->promise();
         // Force the pool to complete synchronously
         $promise->wait();
-        // $timeEnd = microtime(true);
-        // echo sprintf('Operation completed in %s seconds' . PHP_EOL, $timeEnd - $timeStart);
+
         return true;
     } catch (Exception $e) {
-        // echo sprintf('Error: %s' . PHP_EOL, $e->getMessage());
```
MagmarFire, line 90: Is there a reason why we're getting rid of logging here? Sounds counterproductive to me.

MagmarFire (changes requested):

lib/util/mail.php
MagmarFire: I'm all for removing the commented `echo` statement, though.

luchaos:
Closing in favor of #911 
See:
Mailhog: https://github.com/RetroAchievements/RAWeb/blob/master/docker-compose.yml#L58-L64
SMTP mails via Symfony Mailer: https://github.com/RetroAchievements/RAWeb/blob/master/lib/util/mail.php#L41-L62

PR #855, opened 2021-11-21
Fixed sorting by 'Last Updated' field for games list

### LINKED ISSUE #841: Issue with "Last Updated" sorting
Sorting Game List by Last Updated (IDs `6` and `16`) don't seem to follow an order.

Game Boy Color for example:
image

anegoda1995:
Fixes #841

### DIFF
--- a/lib/database/game.php
+++ b/lib/database/game.php
@@ -420,10 +420,10 @@ function getGamesListByDev($dev, $consoleID, &$dataOut, $sortBy, $ticketsFlag = 
             break;
 
         case 6:
-            $query .= "ORDER BY gd.ConsoleID, ach.DateModified DESC, Title ";
+            $query .= "ORDER BY gd.ConsoleID, MAX(ach.DateModified) DESC, Title ";
             break;
         case 16:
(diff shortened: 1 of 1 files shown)


lib/database/game.php
```diff
@@ -420,10 +420,10 @@ function getGamesListByDev($dev, $consoleID, &$dataOut, $sortBy, $ticketsFlag =
             break;
 
         case 6:
-            $query .= "ORDER BY gd.ConsoleID, ach.DateModified DESC, Title ";
+            $query .= "ORDER BY gd.ConsoleID, MAX(ach.DateModified) DESC, Title ";
```
luchaos, line 423: Good find 👍 

As @Jamiras pointed out this could be done by reusing the `DateModified` column alias, too, to keep it simpler.

See https://github.com/RetroAchievements/RAWeb/blob/master/lib/database/game.php#L357

Maybe a better name for that alias would be fitting - like `SetDateModified` as it's about the last change in a game's achievement set rather than a game's own metadata. Not that important, though.

luchaos (changes requested):

luchaos (approved):

PR #861, opened 2021-11-27
eliminate second call to expensive query

Jamiras:
The query used to determine the latest masters of a set and highest scorers for a set is very intensive when a set has a lot of users as it has to calculate every user's score for the set and the datetime of the last achievement earned by the user from the set.

The current implementation runs the query twice, once for High Scores, and once for Latest Masters. This PR runs it once to determine both.

In my dev env, this eliminated 25% of the time spent generating the game page for Super Mario Bros. (6.6s down from 8.8s), which has 5186 unique users for which those values had to be calculated.

I'm not sure if this will translate directly to the live site. The live Super Mario Bros. set has over 17k users, and Super Mario World has over 27k users. Both of these pages took me more than 15 seconds to load. Even shaving 4 seconds (~25%) off those times would be beneficial.

### DIFF
--- a/lib/database/game.php
+++ b/lib/database/game.php
@@ -906,25 +906,26 @@ function getTotalUniquePlayers($gameID, $requestedBy)
  * Gets a game's high scorers or latest masters.
  *
  * @param int $gameID game ID to get high score information for
- * @param int $offset query offset value
- * @param int $count query number of returned rows
  * @param string $requestedBy user requesting the information
- * @param int $type The type of data to return.
- *          0 - High Scores
- *          1 - Latest Masters
  *
  * @return array of user information to display on the High Scores section of a game page
  */
-function getGameTopAchievers($gameID, $offset, $count, $requestedBy, $type = 0)
+function getGameTopAchievers($gameID, $requestedBy)
 {
     sanitize_sql_inputs($gameID, $offset, $count, $requestedBy);
 
-    $retval = [];
-    $havingQuery = "";
-    $order = "ASC";
-    if ($type == 1) {
-        $havingQuery = "HAVING TotalScore = (SELECT SUM(Points * 2) AS Points FROM Achievements WHERE GameID = $gameID AND Flags = 3)";
-        $order = "DESC";
+    $high_scores = [];
+    $masters = [];
+    $mastery_score = 0;
+
+    $query = "SELECT SUM(Points * 2) AS Points FROM Achievements WHERE GameID = $gameID AND Flags = 3";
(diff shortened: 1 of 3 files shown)


lib/database/game.php
```diff
@@ -936,19 +937,31 @@ function getGameTopAchievers($gameID, $offset, $count, $requestedBy, $type = 0)
...
 
     $dbResult = s_mysql_query($query);
     SQL_ASSERT($dbResult);
 
     if ($dbResult !== false) {
         while ($data = mysqli_fetch_assoc($dbResult)) {
-            $retval[] = $data;
+            if (count($high_scores) < 10) {
+                array_push($high_scores, $data);
+            }
```
Jamiras, line 949: the first ten rows will be the high scores

lib/database/game.php
```diff
@@ -936,19 +937,31 @@ function getGameTopAchievers($gameID, $offset, $count, $requestedBy, $type = 0)
...
+
+            if ($data['TotalScore'] == $mastery_score) {
+                if (count($masters) == 10) {
+                    array_shift($masters);
+                }
+                array_push($masters, $data);
+            } elseif (count($high_scores) == 10) {
+                break;
```
Jamiras, line 957: keep the last ten rows matching the mastery score (they'll be the latest masters)

lib/database/game.php
```diff
@@ -906,25 +906,26 @@ function getTotalUniquePlayers($gameID, $requestedBy)
...
-    if ($type == 1) {
-        $havingQuery = "HAVING TotalScore = (SELECT SUM(Points * 2) AS Points FROM Achievements WHERE GameID = $gameID AND Flags = 3)";
-        $order = "DESC";
+    $high_scores = [];
+    $masters = [];
+    $mastery_score = 0;
+
+    $query = "SELECT SUM(Points * 2) AS Points FROM Achievements WHERE GameID = $gameID AND Flags = 3";
+    $dbResult = s_mysql_query($query);
+    SQL_ASSERT($dbResult);
```
luchaos, line 923: This call to `SQL_ASSERT` can be omitted - it only will result in an error log entry which wouldn't serve any meaningful purpose if the query failed for whatever reason.
Not relevant to this PR though - we still have a few of those.

luchaos (approved):

lib/database/game.php
Jamiras: Just copied from a few lines down.

luchaos: Got it - i see that we use this quite often even though we don't need it anymore.
Can be ignored

PR #862, opened 2021-11-27
add uploadleaderboard dorequest api

Jamiras:
Combines the `dorequest.php?r=uploadachievement` functionality and the `request/leaderboard/update.php` code to support creating and editing leaderboards through the dorequest api.

Adds support to the web editor for "Time (Seconds)" and "Time (Minutes)" formats.

Changes the header on the leaderboard entries page from "Date Won" to "Date Submitted".


lib/database/leaderboard.php
```diff
@@ -548,10 +549,21 @@ function GetLeaderboardData($lbID, $user, $numToFetch, $offset, $friendsOnly, $n
     return $retVal;
 }
 
+function _FormatValueSeconds($hours, $mins, $secs)
```
luchaos, line 552: As a general purpose string formatting function I think it'd be fitting to move it to `util/string.php` or similar.
Underscore prefixed functions in PHP are usually reserved for runtime/lang level functions - the non-prefixed camelCase variant (snake case even) would be better practice `formatValueSeconds` imo.
I'm aware that most of our functions are PascalCased, but let that slide because of the legacy character of the code base. Let's stick to the existing code style at least

lib/database/leaderboard.php
```diff
@@ -577,12 +589,29 @@ function GetFormattedLeaderboardEntry($formatType, $scoreIn)
...
 }
 
+function isValidLeaderboardFormat($formatType)
+{
+    return $formatType == 'TIME' ||      // Frames
+           $formatType == 'TIMESECS' ||  // Seconds
+           $formatType == 'MINUTES' ||   // Minutes
+           $formatType == 'MILLISECS' || // Hundredths of seconds
+           $formatType == 'VALUE' ||     // Raw number
+           $formatType == 'SCORE';      // Number padded to six digits
```
luchaos, line 612: Alternatively:
```
return in_array($formatType, [
    'TIME', // Frames
    'TIMESECS', // Seconds
    'MINUTES', // Minutes
    'MILLISECS', // Hundredths of seconds
    'VALUE', // Raw number
    'SCORE', // Number padded to six digits
]);
```
Seeing those "enums" being used more often in code they could be moved to (class) constants instead of comparing strings.
None of the above is necessary - just as a reminder.

lib/database/leaderboard.php
```diff
@@ -810,6 +840,85 @@ function SubmitNewLeaderboard($gameID, &$lbIDOut, $user)
...
+    }
+
+    if (!isValidConsoleId(getGameData($gameID)['ConsoleID'])) {
+        $errorOut = "You cannot promote leaderboards for a game from an unsupported console (console ID: " . getGameData($gameID)['ConsoleID'] . ").";
+        return false;
+    }
+
+    if (!isValidLeaderboardFormat($format)) {
+        $errorOut = "Unknown format: $format";
+        return false;
```
luchaos, line 890: I like those early returns a lot 👍

public/leaderboardinfo.php
```diff
@@ -153,7 +153,7 @@
             echo "<div class='larger'>$lbTitle: $lbDescription</div>";
 
             echo "<table><tbody>";
-            echo "<tr><th>Rank</th><th>User</th><th>Result</th><th>Date Won</th></tr>";
+            echo "<tr><th>Rank</th><th>User</th><th>Result</th><th>Date Submitted</th></tr>";
```
luchaos, line 156: Much better choice imo

luchaos (changes requested):
Requesting changes mainly because of the underscore prefixed function.
Rest looks great imo.

lib/database/leaderboard.php
Jamiras: The intent was to make it a private function. Underscore is the standard way to do that in python. What is the PHP equivalent?

luchaos: Got it - technically it would be a private method within a class but we do not have a lot of classes in this codebase.
It's still common to have global functions in PHP which usually are camel cased - there's no way to declare it as private in this flat/function based style we have here.
I understand that we'll probably only ever need this function in the context of leaderboards, so it's fine to keep it this way if you prefer it. I'm personally used to mark private members and functions as private with underscores, it's just not common in this very codebase. Up to you - I'm fine with either, really.

luchaos (approved):

PR #866, opened 2021-12-04
don't show leaderboards with DisplayOrder<0 on game pages

Jamiras:


Tsearo (approved):
Tested and didn't notice any issues, hidden when value is negative 👍

luchaos (approved):

PR #537, opened 2021-01-15
Escape apostrophes when replying to private messages

Alfex3:
Fixes #535, but for replies.

PR #536, opened 2021-01-15
Escape apostrophes in private message's title when contacting the reporter of an achievement.

Alfex3:
Fixes #535

PR #871, opened 2021-12-05
Fix escaping of apostrophes and newlines in inbox

### LINKED ISSUE #535: Contacting the reporter about a ticket of a game with an apostrophe in the title
When clicking on "Contact the reporter", the title of the PM is cut where the apostrophe is, and the heading of the message doesn't appear.

Jamiras:
Fixes newlines in messages being converted to "rn" and apostrophes prematurely truncating message titles (#535).

image

I chose to reimplement #536 and #537 as I believe a better solution is to url encode the parameter (instead of c escaping it) and then always htmlencode the string when building the new page (instead of only encoding the value read from the database and expecting the url parameter to be encoded).

Closes #536
Closes #537 
Fixes #535


lib/database/message.php
```diff
@@ -2,10 +2,10 @@
 
 function CreateNewMessage($author, $destUser, $messageTitle, $messagePayloadIn)
 {
-    sanitize_sql_inputs($author, $destUser, $messageTitle, $messagePayloadIn);
-
     $messagePayload = nl2br($messagePayloadIn);
 
+    sanitize_sql_inputs($author, $destUser, $messageTitle, $messagePayloadIn);
```
Jamiras, line 7: the important change here is actually moving the `messagePayload` line, so the <br> conversion is performed on a not-sanitized input.

lib/database/ticket.php
```diff
@@ -114,10 +114,10 @@ function submitNewTickets($userSubmitter, $idsCSV, $reportType, $noteIn, &$summa
         return false;
     }
 
-    sanitize_sql_inputs($userSubmitter, $reportType, $noteIn);
+    $note = $noteIn;
+    sanitize_sql_inputs($userSubmitter, $reportType, $note);
```
Jamiras, line 118: sanitize a copy of the message so the unescaped value can be passed to later functions.

lib/database/message.php
luchaos: makes sense 👍

luchaos (approved):
lgtm, tested locally.

PR #874, opened 2021-12-07
Remove aggressive sanitization from achievements; fix escaping for tooltips

Jamiras:
Closes #755

When uploading an achievement containing non-ASCII characters, they're aggressively replaced with underscores, as were slashes:
image

This replaces that sanitization with something more appropriate, and updates all the pages (that I could find) that aren't properly escaping the values for display.

image

The non-ASCII characters do display correctly in RALibretro and RetroArch (at least on Windows):
image

image

I've also fixed the display/escaping for leaderboard titles/descriptions, and game titles, both of which already seem to support non-ASCII characters.

The database (at least my local one) is still all `latin1`. I know there's already a lot of utf-8 data present in the forums, code notes, and rich presence, so I'm hoping this doesn't create any headaches down the line.


lib/database/achievement.php
```diff
@@ -498,15 +496,15 @@ function UploadNewAchievement(
...
-                '$author', NOW(), NOW(),
+                '$progressFmt', $points, $type,
+                '$dbAuthor', NOW(), NOW(),
                 NOW(), 0, 0,
                 '$badge', 0, NULL,
                 0
             )";
         // log_sql($query);
-        if (s_mysql_query($query) !== false) {
-            global $db;
```
Jamiras, line 509: `sanitize_sql_inputs` escapes quotes with backslashes, and `s_mysql_query` rejects anything containing backslashes. calling `mysqli_query` directly after calling `sanitize_sql_inputs` seems common in places where we support apostrophes (messages, comments, forums, etc).

lib/render/achievement.php
```diff
@@ -49,7 +49,7 @@ function GetAchievementAndTooltipDiv(
     }
 
     if ($inclSmallBadge) {
-        $achNameAttr = htmlspecialchars($achName, ENT_QUOTES);
+        $achNameAttr = attributeEscape($achName);
```
Jamiras, line 52: `htmlspecialchars` escapes a lot of things that the tooltip didn't want to unescape (like `&frac12;`). This just escapes quotes and double quotes.

public/gameInfo.php
```diff
@@ -962,8 +963,6 @@ function submitSetRequest(user, gameID) {
 
                             $earnedOnHardcore = isset($nextAch['DateEarnedHardcore']);
 
-                            $achDesc = str_replace('"', '\'', $achDesc);
```
Jamiras, line 965: There's no reason to change double quotes to single quotes for display on the web page. You can see this behavior in the "before" image in the PR description. `"47______"` in the title, but `'47______'` in the description.

lib/database/achievement.php
luchaos: as discussed - this is the preferred way 👍

lib/util/string.php
```diff
@@ -9,6 +9,16 @@ function sanitize_outputs(&...$outputs)
     }
 }
 
+function attributeEscape($input)
+{
+    // htmlspecialchars escapes a bunch of stuff that the tooltip can't handle
+    // (like &rsquo; $frac12; and &deg;). when placed in title or alt fields.
+    // just do the bare minimum.
```
luchaos, line 16: interesting - good solution 👍

luchaos (approved):

PR #875, opened 2021-12-08
Remove Facebook integration

### LINKED ISSUE #23: Facebook Integration broken
I don't know the details but I know it's not working.

luchaos:
Sending Facebook feed notifications hasn't worked as expected for the longest time.
Hides Facebook connect site awards.

Closes #23


lib/render/user.php
```diff
@@ -220,11 +220,10 @@ function RenderSiteAwards($userAwards)
...
                 $linkdest = ''; //    TBD: referrals page?
-            } elseif ($awardType == 5) { //    Signed up for facebook!
-                $tooltip = "Awarded for associating their account with Facebook! Thanks for spreading the word!";
-
-                $imagepath = "/Images/_FBAssoc.png";
-                $linkdest = "/controlpanel.php#facebook";
+            // } elseif ($awardType == 5) { // Signed up for facebook!
+            //     $tooltip = "Awarded for associating their account with Facebook! Thanks for spreading the word!";
+            //     $imagepath = "/Images/_FBAssoc.png";
+            //     $linkdest = "/controlpanel.php#facebook";
```
Jamiras, line 226: why comment out instead of remove?

public/reorderSiteAwards.php
```diff
@@ -119,10 +119,10 @@
...
                     $imagepath = "/Images/00083.png";
                 }
-            } elseif ($awardType == 5) { //    Signed up for facebook!
-                $tooltip = "Awarded for associating their account with Facebook! Thanks for spreading the word!";
-                $awardTitle = "Facebook Association";
-                $imagepath = "/Images/_FBAssoc.png";
+                // } elseif ($awardType == 5) { //    Signed up for facebook!
+                //     $tooltip = "Awarded for associating their account with Facebook! Thanks for spreading the word!";
+                //     $awardTitle = "Facebook Association";
+                //     $imagepath = "/Images/_FBAssoc.png";
```
Jamiras, line 125: why comment out instead of remove?

Jamiras (approved):

lib/render/user.php
luchaos: I kept those for documentation purposes.`SiteAward` entries of that type still exist. 

Usually i'd remove dead code but this is the only reference to the magic number `5` - there are no DB comments on the respective column.

I went ahead and introduced the AwardType class which does the documentation job instead and removed those blocks.

public/reorderSiteAwards.php
luchaos: See other comment

Jamiras: still present

lib/render/user.php
Jamiras: 👍

public/reorderSiteAwards.php
luchaos: Oh my - copied code 🙈  Thanks!

Jamiras (approved):

PR #883, opened 2021-12-16
feat(api): add API_GetAchievementDistribution.php

wescopeland:
Hi! I am the author of https://github.com/achievements-app/retroachievements-js and https://github.com/achievements-app/psn-api. I'm in the process of building a dashboard for my own portfolio site that is connected to all the gaming platforms I play on, and I've hit a snag during the development process.

Currently there is no good way for a RetroAchievements API user to calculate what the mastery % of a game is without manually scraping the site.

This PR:
* Adds an endpoint to retrieve the achievement distribution data from a given game ID.

On the current V1 site, this is represented via this graph:
Screen Shot 2021-12-15 at 7 58 05 PM

After this PR is merged, this data can be retrieved via the _/API/API_GetAchievementDistribution.php_ public route.

### DIFF
--- a/public/API/API_GetAchievementDistribution.php
+++ b/public/API/API_GetAchievementDistribution.php
@@ -10,4 +10,4 @@ $hardcore = requestInputQuery('h', 0, 'integer');
 $requestedBy = requestInputQuery('z');
 $flags = requestInputQuery('f', 3, 'integer');
 
-echo json_encode(getAchievementDistribution($gameId, $hardcore, $requestedBy, $flags));
+echo json_encode(getAchievementDistribution($gameID, $hardcore, $requestedBy, $flags));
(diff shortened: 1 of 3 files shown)


public/API/API_GetAchievementDistribution.php
```diff
@@ -0,0 +1,13 @@
...
+require_once __DIR__ . '/../../lib/bootstrap.php';
+
+runPublicApiMiddleware();
+
+$gameID = requestInputQuery('i');
+$hardcore = requestInputQuery('h', 0, 'integer');
+$requestedBy = requestInputQuery('z');
+$flags = requestInputQuery('f', 3, 'integer');
+
+echo json_encode(getAchievementDistribution($gameId, $hardcore, $requestedBy, $flags));
```
Tsearo, line 13: You define $gameID but call $gameId which phpstan is complaining about

wescopeland: Oops -- should be all set now.

Tsearo (approved):

PR #893, opened 2021-12-27
add lazy loading attribute to profile and game images outputted by JS

Krylan:
This is a small change in JS code to include "loading" attribute, which will greately decrease transfer, when visiting homepage. Active players widget includes a lot of images, and more players are active in games, the more data user needs to download to see main page (and, I guess, that's first contact with RA for a lot of new users).

Implementing this change will let browsers to not load those images, until user explicitly show the need for them (by scrolling down in active players list).

### DIFF
--- a/lib/constants.php
+++ b/lib/constants.php
@@ -1,4 +1,4 @@
 <?php
 
-const VERSION = '1.78.0';
+const VERSION = '1.78.1';
 const MIN_POINTS = 500;
--- a/public/js/all.js
+++ b/public/js/all.js
@@ -341,7 +341,7 @@ function GetGameAndTooltipDiv(gameID, gameTitle, gameIcon, consoleName, imageIns
   if (imageInstead) {
     displayable = '<img alt="started playing ' + gameTitle
       + '" title="Started playing ' + gameTitle + '" src=\'' + gameIcon
-      + '\' width=\'32\' height=\'32\' class=\'badgeimg\' />';
+      + '\' width=\'32\' height=\'32\' class=\'badgeimg\' loading=\'lazy\' />';
   }
   return '<div class=\'bb_inline\' onmouseover="Tip(\'' + tooltip
     + '\')" onmouseout="UnTip()" >'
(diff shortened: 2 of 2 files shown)


MagmarFire (approved):

lib/constants.php
```diff
@@ -1,4 +1,4 @@
 <?php
 
-const VERSION = '1.78.0';
+const VERSION = '1.78.1';
```
luchaos, line 3: Feel free to bump the minor version instead of just the patch version. This is more than a bugfix, a feature rather: `1.79.0`. Bumps the version for the next release which might include more.

luchaos (approved):
Looking at the changes i realized that this only addresses the active players widget which doesn't benefit from the idea as much as achievements on players profiles or game icons on games pages. Might be worth adding those in another PR.

lib/constants.php
Krylan: @luchaos, I just tought that this change is so small (and only concerns preformance improvement), that I felt increasing minor version just for that would be too much.
Addressing above comment about other possibilities for improvement (on player, or game page), I can include it in this PR as well (and then bump the minor version). Just the problem with active players widget was in my opinion so heavily impacting performance, that I focused on that one.

Krylan: @luchaos oh, I just noticed that achievement, game, and user images on other pages are already lazy loaded! You did that last year in april, here: [https://github.com/RetroAchievements/RAWeb/commit/92e4f03ba3899abe733e7c415ae6185fadd1b266](https://github.com/RetroAchievements/RAWeb/commit/92e4f03ba3899abe733e7c415ae6185fadd1b266)

So I guess there is nothing more to do, and active players widget is the thing, which indeed needs optimization.
Should I bump minor version before merging, or it can go like that?

luchaos: oh, you're right! i messed up my search and didn't see those 😅

luchaos: Going for the minor version is fine - it's not about the size but the type of change. I usually bump the minor version before release but you can do so in this PR, too, or just leave it as is. Both is fine imo.

PR #899, opened 2022-01-15
Prevent shortcode parser changing case …

luchaos:
… of text in brackets when it's not a registered tag.

See test file for an example.


src/Shortcode/Shortcode.php
```diff
@@ -50,8 +50,13 @@ private function parse(string $input, array $options = []): string
 
         // case insensitive
         foreach ($this->handlers->getNames() as $tag) {
-            $input = preg_replace("~$tag]~i", "$tag]", $input);
-            $input = preg_replace("~\[$tag~i", "[$tag", $input);
+            if (strlen($tag) === 1) {
```
MagmarFire, line 53: I feel like it would be more concise and clear to use `strtolower` or `strtoupper` on the tag to convert it to a consistent case. There should theoretically be no need to check for length or to use ifs here.

MagmarFire (changes requested):

src/Shortcode/Shortcode.php
MagmarFire: Unless I'm misunderstanding something?

Jamiras: The `strtolower` wouldn't work because this code is trying to normalize the case of the tags in the document. The tags themselves are already all lowercase. i.e. this should change `[ACH=12345]` to `[ach=12345]`

I'm not understanding the need for the if/else either. It seems like the intent was to fix a case where a substring would get matched incorrectly, but is there ever a case where you _do_ want to match a substring? Shouldn't the old code just be replaced with this?
```
        // case insensitive
        foreach ($this->handlers->getNames() as $tag) {
            $input = preg_replace("~\[/$tag]~i", "[/$tag]", $input); // closing tag
            $input = preg_replace("~\[$tag]~i", "[$tag]", $input); // opening tag
            $input = preg_replace("~\[$tag=~i", "[$tag=", $input); // opening tag with value
        }
```
Or possibly (to include whitespace):
```
        // case insensitive
        foreach ($this->handlers->getNames() as $tag) {
            $input = preg_replace("~\[\s*/$tag\s*]~i", "[/$tag]", $input); // closing tag
            $input = preg_replace("~\[\s*$tag\s*]~i", "[$tag]", $input); // opening tag
            $input = preg_replace("~\[\s*$tag\s*=~i", "[$tag=", $input); // opening tag with value
        }
```

Jamiras: I've applied the first set of changes with luchaos's approval:
> your suggestion sounds reasonable - not sure why i did not change the regex to include the closing brackets; much cleaner 👍
> i wouldn‘t include whitespace within/before/after the brackets though as those should remain being invalid usage

luchaos: thanks @Jamiras & @MagmarFire - i haven't spent enough time on that issue, my initial solution wasn't a good fit.

Jamiras (approved):

PR #902, opened 2022-01-31
delay new forum post notifications from unverified users

### LINKED ISSUE #901: Bug: Subscribed Forum thread issues notifications that lead nowhere
At the moment of posting this issue, if I follow the link on the screenshot, there's no message from Gorets, I assume this is **because user and their messages are not verified**, if that's indeed the case - email notification should not be sent if forum post was made by unverified user.

image

Jamiras:
fixes #901 

Prevents sending an email notification when an unverified user adds a comment to a forum topic.
Sends email notifications for the forum topics when the unverified user becomes verified.


luchaos (approved):
Great solution imo 👍

PR #904, opened 2022-02-03
add mature content warning

Jamiras:
When accessing a game that's associated to the `[Theme - Mature]` hub, the user will now be presented with a Mature Content warning:
image

Clicking `Yes` will load the page normally. Clicking `No` will take them back to the games list for the console.

Because the acknowledgment is not stored in the database, the user will have to re-acknowledge the warning each time they return to the game's page.

The user can completely disable the functionality from their settings page (this _is_ stored in the database):
image


public/gameInfo.php
```diff
@@ -56,6 +57,56 @@
...
+<head prefix="og: http://ogp.me/ns# retroachievements: http://ogp.me/ns/apps/retroachievements#">
+    <?php RenderSharedHeader(); ?>
+    <?php RenderTitleTag($pageTitle); ?>
+    <?php RenderGoogleTracking(); ?>
+</head>
+<body>
+<?php RenderTitleBar($user, $points, $truePoints, $unreadMessageCount, $errorCode, $permissions); ?>
+<?php RenderToolbar($user, $permissions);
+            echo "<div id='mainpage'>";
+            echo "<div id='leftcontainer'>";
```
MrOwnership, line 79: This should probably be `fullcontainer` so it takes up the full width of the page rather than just the left portion.

public/gameInfo.php
```diff
@@ -56,6 +57,56 @@
...
+    <form id='consentform' action='/game/<?php echo $gameID ?>' method='get' style='float:left'>
+      <input type='hidden' name='v' value='1'/>
+      <input type='submit' value='Yes. I&apos;m an adult'/>
+    </form>
+    <form id='escapeform' action='/gameList.php' method='get' style='float:left; margin-left:16px'>
+      <input type='hidden' name='c' value='<?php echo $consoleID ?>'/>
+      <input type='submit' value='Not Interested'/>
+    </form>
+  </div>
+</div>
```
MrOwnership, line 101: Missing another `</div>` here which is causing the footer to become distorted.

public/gameInfo.php
Jamiras: I think it flows better as just a left container as clicking Yes will appear to flow into the left container of the actual game page and the rest of the elements will appear around the warning. But it does sort of feel like a waste of space.

public/gameInfo.php
Jamiras: good call. will fix.

public/gameInfo.php
luchaos: let's have it as is and iterate on that later - we should release this feature.

luchaos (approved):

public/gameInfo.php
MrOwnership: Sounds good to me.

PR #910, opened 2022-02-14
Rename consoles

luchaos:


lib/render/layout.php
```diff
@@ -227,10 +227,10 @@ function RenderToolbar($user, $permissions = 0)
     echo "<li><a href='/gameList.php?c=9'>Sega CD</a></li>";
     echo "<li><a href='/gameList.php?c=10'>Sega 32X</a></li>";
     echo "<li><a href='/gameList.php?c=39'>Sega Saturn</a></li>";
-    echo "<li><a href='/gameList.php?c=40'>Sega Dreamcast</a></li>";
+    echo "<li><a href='/gameList.php?c=40'>Dreamcast</a></li>";
     /// Other
     echo "<li class='dropdown-header'>Other</li>";
-    echo "<li><a href='/gameList.php?c=43'>3DO Interactive Multiplayer</a></li>";
+    echo "<li><a href='/gameList.php?c=43'>3DO</a></li>";
```
Jamiras, line 233: I'm not sure what we're trying to standardize to here. Clicking the link takes you to a page that says "3DO Interactive Multiplayer", so it's probably fine the way it is.

Conversely, clicking on "Sega Saturn" or "Sega 32X" takes you to pages labelled "Saturn" and "32X" respectively.

Some of the other labels don't match either. "NES/Famicom" just goes to "NES", "SNES/Super Famicom" goes to "SNES", "Genesis/MegaDrive" goes to "Mega Drive", and "PC Engine/TurboGrafx 16" goes to "PC Engine"

luchaos: You're right - that wouldn't match the naming convention.
I opened this as a quick draft; needs more discussion (or closing).

PR #913, opened 2022-02-22
separate hubs from similar games

Jamiras:
Extracts Hubs from the Related list and turns them into metadata which will be displayed below the game screenshots.

image image

If a Hub name starts with Meta, that's discarded. The remaining Hub name is split on the last hyphen, creating a label and value. Multiple items sharing a label are grouped together, otherwise labels are ordered as they would have been in the Related list.

The metadata items appearing above the screenshots are still populated from the game details (editable in the Dev section), but matching Hub items will be merged. If the label/values are not exact, they will be appended instead of merged:

image

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -55,12 +55,41 @@ $isFullyFeaturedGame = !in_array($consoleName, ['Hubs']);
 
 $pageTitle = "$gameTitle ($consoleName)";
 
-$gameAlts = getGameAlternatives($gameID);
+$relatedGames = getGameAlternatives($gameID);
+$gameAlts = [];
+$tags = [];
+foreach ($relatedGames as $gameAlt) {
+    if ($gameAlt['ConsoleName'] != 'Hubs') {
+        $gameAlts[] = $gameAlt;
+        continue;
+    }
+
+    $parts = explode(' - ', ltrim(rtrim($gameAlt['Title'], ']'), '['));
+    if (count($parts) > 2 && $parts[0] == 'Meta') {
+        unset($parts[0]);
+        $parts = array_values($parts);
+    }
+
+    if (count($parts) > 2) {
+        $value = $parts[-1];
+        unset($parts[-1]);
+        $key = implode(' - ', $parts);
+    } else {
+        $key = $parts[0];
+        $value = $parts[1];
+    }
+
+    if (substr($key, 0, 5) == 'Meta|') {
+        $key = substr($key, 5);
+    }
+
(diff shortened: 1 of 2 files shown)


public/gameInfo.php
```diff
@@ -560,36 +569,57 @@ function submitSetRequest(user, gameID) {
...
+
+                $first = true;
+                if ($gameHubs) {
+                    $hubPrefix = "[$label - ";
+                    foreach ($gameHubs as $hub) {
+                        if (substr($hub['Title'], 0, strlen($hubPrefix)) == $hubPrefix) {
+                            if ($first) {
+                                $first = false;
+                            } else {
+                                $output .= ", ";
```
luchaos, line 584: Minor thing - alternatively, I'd probably collect all the outputs in an array and implode it with a separator after the loop.
Then there's no need for the local `$first` variable and the `if/else` construct.
```
$hubLinks = [];
foreach ($gameHubs as $hub) {
    $hubLinks[] = "<a href=/game/" . $hub['gameIDAlt'] . ">$value</a>";
}
$output .= implode(', ', $hubLinks);
```

Not sure how it affects `$value == $gameDataValue` and the other check below - those might be reduced to filtering the array of hubs beforehand?

public/gameInfo.php
```diff
@@ -560,36 +569,57 @@ function submitSetRequest(user, gameID) {
             $imageIngame = $gameData['ImageIngame'];
             $pageTitleAttr = attributeEscape($pageTitle);
 
+            function addMetadata($label, $gameDataValue, $gameHubs)
```
luchaos, line 572: As this function renders HTML I'd probably put it in the render lib folder - even though this application is only used here atm. Yet, I don't think we keep functions within "endpoint" files anywhere else.
I'd suggest `renderMetadata()` for the function name as it does in fact render, not add metadata?
That'd be mostly in line with the rest of the application, too.

luchaos (commented):
Wonderful idea to add that data to the game details. Some minor code calisthenics I'd personally change - feel free to discard though, nothing important there. It works fine.

luchaos (approved):

public/gameInfo.php
Jamiras: Good idea. I forget about language-specific tricks like this.

public/gameInfo.php
Jamiras: The function was created to reuse code. It's kind of awkward to put that in a different file as it's only generating a single table row, but I did it anyway.

PR #914, opened 2022-02-22
add admin comment block to user page

Jamiras:
Adds a new comment block within the admin tools on a user's page, allowing administrators to note things about the user, such as why they were untracked or banned. It also leverages the virtual audit functionality for tracking changes to settings managed by administrators.

image


luchaos (approved):
tested, lgtm. fantastic improvements imo

PR #920, opened 2022-03-10
Allow labelling hashes

Jamiras:
Replaces the Unlink Hashes page with a Manage Hashes page where additional information about a hash can be noted.

image

This information will be shown on the Linked Hashes page for the game:

image

Labels is a CSV of potential tags to apply to the hash entry. If a .png image exists in `host/Images/labels`, it will be used (i.e. `nointro` will be turned into host/Images/labels/nointro.png`), otherwise, the text will be shown.

This is intended to augment/replace the information posted in the first post of an official forum post for a game:

image

But there's a couple caveats:
* Only one filename (description) can be applied to a hash. You can see in the current official forum topic for Super Mario Bros that both a `nointro` and a `goodtools` filename are provided.
* The actual ROM checksum (md5) and CRC32 values are not captured. These could be added in the future without too much effort.

The descriptions also appear in the hash selector when creating a ticket:

image


ScottFromDerby:
This makes me so happy :) I was so hesitant to do this originally because I didn't want people fighting over filenames, but I honestly you've got the best solution here (with labels). Love it, thank-you Jamiras

luchaos (approved):
lgtm. executed migration queries on production db.

PR #921, opened 2022-03-12
fix edits not being logged for leaderboards with an unknown author

Jamiras:
The logic to prevent adding an "edit" comment when first uploading a new leaderboard was also not adding an "edit" comment when modifying a leaderboard whose author has been NULLed out for whatever reason.


Tsearo (approved):
Confirmed behavior locally, site updates were already creating edit messages but emulator updates wouldn't. This code fixed the issue with edit messages not showing when doing emulator edits.

PR #928, opened 2022-03-26
Use git native core.hooksPath config

luchaos:
Chores:
- Add phpstan dev dependecy & fix new phpstan errors
- Update npm dependencies & run `npm run fix`
- Update composer dependencies
- Addresses recent dependabot security alert


lib/database/achievement.php
```diff
@@ -1126,8 +1126,8 @@ function recalculateTrueRatio($gameID)
 
         foreach ($achData as $nextAch) {
             $achID = $nextAch['ID'];
-            $achPoints = $nextAch['Points'];
-            $numAchieved = $nextAch['NumAchieved'];
+            $achPoints = (int) $nextAch['Points'];
+            $numAchieved = (int) $nextAch['NumAchieved'];
```
luchaos, line 1130: PHPStan issue

lib/database/user.php
```diff
@@ -1012,13 +1012,12 @@ function getUsersRecentAwardedForGames($user, $gameIDsCSV, $numAchievements, &$d
...
-    $gameIDsArray = explode(',', $gameIDsCSV);
-
-    $numIDs = count($gameIDsArray);
-    if ($numIDs == 0) {
+    if (empty($gameIDsCSV)) {
         return;
     }
 
+    $gameIDsArray = explode(',', $gameIDsCSV);
+
```
luchaos, line 1020: PHPStan really didn't like this one for some reason

.php-cs-fixer.php
```diff
@@ -1,61 +1,44 @@
...
             'extra',
             'throw',
             'use',
         ],
     ],
+    'not_operator_with_space' => false,
+    'phpdoc_align' => false,
+    'phpdoc_summary' => false,
+    'single_line_throw' => false,
+    'single_line_comment_spacing' => false,
```
luchaos, line 33: Disabled that new rule as it would unnecessarily touch 90 files:
`//commet` -> `// comment`.
We should consider removing all those debug comments instead.

.hooks/post-merge
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+composer install
```
Jamiras, line 3: shouldn't this be `php composer.phar install` to be consistent with the readme?

> Use composer provided in this repository...
> `$ php composer.phar install`
> ...or your globally installed instance.
> `$ composer install`

.hooks/pre-commit
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+printf "\n✅ pre-commit hook\n\n"
```
Jamiras, line 3: what's the point of this? didn't it used to do a `composer lint` or something similar? I know I've had to run `composer fix` to commit before.

.hooks/pre-push
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+printf "\n⏳ composer lint" && \
+vendor/bin/php-cs-fixer fix --verbose --dry-run --show-progress=none && \
+printf "\n⏳ npm lint" && \
+npm run lint && \
+printf "\n⏳ composer analyse" && \
+vendor/bin/phpstan analyse --memory-limit 512M --no-progress && \
```
Jamiras, line 8: why not run `compser lint` and `composer analyse` for these steps?

.hooks/post-merge
luchaos: I forget that we ship a `composer.phar` in this project - this is not the standard usually. Guess we can make it the phar version here for those who do not have it installed globally 👍

.hooks/pre-commit
luchaos: To have it in place - the linting etc has been moved to `pre-push`.

.hooks/pre-push
luchaos: The composer scripts run the `--ansi` variants which messes with some git GUIs' console output

.hooks/post-merge
luchaos: Updated to use shipped `composer.phar`

luchaos:
Running `composer install` (`composer.phar install`) will put the git config in place. If the previous `cghooks`-managed hooks are still in place it should run `install` automatically when fast forwarding master (after this has been merged, obviously).

Jamiras (approved):

luchaos:
I'll add `npm run lint` to the GitHub ci action in a separate PR.

luchaos:
Please run `composer.phar install` once manually after pulling `master` to make sure the git hooks config gets updated for you. (Just in case the existing `post-merge` git hook fails to do so)

PR #933, opened 2022-03-28
Simplify article comments

luchaos:
- Reload after submit at all times.
- Textareas now take the full width
- The submit button is accessible via keyboard

Chores:
- Remove unused JS
- Class import

### DIFF
--- a/lib/render/comment.php
+++ b/lib/render/comment.php
@@ -142,7 +142,7 @@ function RenderCommentInputRow($user, $articleTypeId, $articleId)
                 <form action="/request/comment/create.php" onsubmit="onSubmitComment(event)">
                     <input type="hidden" name="a" value="$articleId">
                     <input type="hidden" name="t" value="$articleTypeId">
-                    <div class="d-flex">
+                    <div class="d-flex align-items-center">
                         <textarea
                             class="comment-textarea" 
                             name="c"
@@ -153,6 +153,9 @@ function RenderCommentInputRow($user, $articleTypeId, $articleId)
                         <button class="comment-submit-button">
                             <img src="$assetUrl/Images/Submit.png" alt="Submit">
                         </button>
+                        <span class="comment-loading-indicator">
+                            <img src="$assetUrl/Images/loading.gif" alt="Loading">
+                        </span>
                     </div>
                     <div class="textarea-counter" data-textarea-id="comment_textarea_$commentId"></div>
                     <div class="form-error"></div>
--- a/lib/render/comment.php
+++ b/lib/render/comment.php
@@ -18,7 +18,7 @@ function RenderCommentsComponent(
 
     echo "<div class='leftfloat'>";
     if ($numComments == 0) {
-        echo "No comments yet. Will you be the first?<br>";
+        echo "<i>No comments</i><br>";
     } else {
         echo "Recent comment(s):<br>";
     }
@@ -72,8 +72,7 @@ function RenderCommentsComponent(
 
     if (isset($user)) {
         //    User comment input:
-        $commentInputBoxID = "art_{$articleTypeID}_{$articleID}";
-        RenderCommentInputRow($user, $commentInputBoxID, $articleTypeID);
+        RenderCommentInputRow($user, $articleTypeID, $articleID);
     }
 
     echo "</tbody></table>";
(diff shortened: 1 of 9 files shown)


public/managehashes.php
```diff
@@ -45,64 +45,56 @@
...
-                $('#warning').html('Status: Errors...' + data);
-                return;
-            }
+    if (confirm('Are you sure you want to unlink the hash ' + hash + '?') === false) {
+        return;
+    }
+    var $warning = $('#warning');
+    $warning.html('Status: updating...');
+    var posting = $.post('/request/game/modify.php', { u: user, g: gameID, f: 3, v: hash });
+    posting.done(function onUnlinkComplete(data) {
```
Jamiras, line 77: while you're in here, you could remove the extraneous function name

public/request/comment/create.php
```diff
@@ -20,6 +20,7 @@
 $articleType = requestInputPost('t', null, 'integer');
 
 $commentPayload = requestInputPost('c');
+// printable characters only. see https://www.asciitable.com/
```
Jamiras, line 23: Probably something for a separate PR, but is there some reason we shouldn't support non-ascii characters, or at least replace them with a question mark instead of stripping them entirely?

lib/render/comment.php
```diff
@@ -127,25 +126,38 @@ function RenderArticleComment(
...
+            </td>
+            <td colspan="3">
+                <form action="/request/comment/create.php" onsubmit="onSubmitComment(event)">
+                    <input type="hidden" name="a" value="$articleId">
+                    <input type="hidden" name="t" value="$articleTypeId">
+                    <div class="d-flex">
+                        <textarea
+                            class="comment-textarea" 
+                            name="c"
+                            maxlength="2000"
```
Jamiras, line 149: 'Comment.Payload` is only 255 characters.

public/js/all.js
```diff
@@ -63,136 +43,31 @@ function strPad(input, padLength, padString) {
...
-          var posting = $.post('/request/comment/create.php', {
-            a: activityId,
-            c: safeComment,
-            t: articleType,
-          });
-          posting.done(function (data) { onCommentSuccess(data, articleType); });
-          var submitButton = commentRow.find('#submitButton');
-          if (submitButton.length) {
-            submitButton.attr('src', '/Images/loading.gif'); // Change to 'loading' gif
-            submitButton.attr('onclick', ''); // stop being able to click this
```
Jamiras, line 183: I was able to submit the same comment multiple times by accidentally clicking on the spinner

public/managehashes.php
luchaos: absolutely, thanks for reminding me!

public/js/all.js
luchaos: Oh my, good call. The submit button is now removed and fully replaced with the loading spinner.
This has the advantage that the loading image is properly prefixed with the `ASSET_URL`.

lib/render/comment.php
luchaos: Interesting - for me and on production db it's a `text` column.
I think that has been changed a while ago but it's not reflected in the `database` migrations - I updated the SQL file accordingly.

Jamiras: Did that change recently? I was still splitting ticket comments into multiple posts in January: https://retroachievements.org/ticketmanager.php?i=44850

Jamiras: Looks like last summer: #639. I must have split it out of habit.

public/request/comment/create.php
Jamiras: It'll probably mess with the character counting as non-ASCII UTF-8 characters take more than 1 latin1 character each. But we've started allowing UTF-8 in other places, so it seems reasonable to allow it here. As long as users don't try to completely fill the field with non-ASCII characters, it should be okay.

ScottFromDerby:
being reminded of my amateurish js... for shame... :P

public/request/comment/create.php
luchaos: Agreed, let's have that. We already allow it in forum comments where the counter is thrown off by the same issue 👍

Jamiras (approved):

PR #934, opened 2022-03-28
Adopt versioned assets filenames

luchaos:
+ serve fresh assets locally at all times.

### DIFF
--- a/lib/render/layout.php
+++ b/lib/render/layout.php
@@ -34,16 +34,11 @@ function RenderSharedHeader()
     echo "<link rel='image_src' href='/Images/RA_Logo10.png'>\n";
     echo "<meta http-equiv='content-type' content='text/html; charset=UTF-8'>\n";
     echo "<meta name='robots' content='all'>\n";
-    //echo "<meta name='Copyright' content='Copyright 2014'>\n";
     echo "<meta name='description' content='Adding achievements to your favourite retro games since 2012'>\n";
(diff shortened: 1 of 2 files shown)


lib/render/layout.php
```diff
@@ -52,22 +47,22 @@ function RenderSharedHeader()
...
+        echo "<script src='/js/all-" . VERSION . ".js'></script>\n";
+    }
+    if (getenv('APP_ENV') === 'local') {
+        echo "<link rel='stylesheet' href='/css/styles.css?" . rand() . "' media='screen'>\n";
+    } else {
+        echo "<link rel='stylesheet' href='/css/styles-" . VERSION . ".css' media='screen'>\n";
+    }
     $customCSS = RA_ReadCookie('RAPrefs_CSS');
     if ($customCSS !== false && mb_strlen($customCSS) > 2) {
         echo "<link rel='stylesheet' href='$customCSS?v=" . VERSION . "' media='screen'>\n";
```
luchaos, line 68: Those theme files rarely ever change.

Jamiras (approved):
Did you want to include the nginx rule in ra-web-v1.conf for documentary purposes even if it's not used for local installs?

luchaos:
> Did you want to include the nginx rule in ra-web-v1.conf for documentary purposes even if it's not used for local installs?

Very good idea, I completely forgot about those, thanks!
I actually do use those configuration files by running it via docker.

PR #937, opened 2022-04-05
split Site Awards into Game/Event/Site Awards

Jamiras:
If any section would be empty it's not displayed. 

image

image

The only caveat is if all three sections would be empty, an empty Game Awards section is displayed.

image

The reorder site awards page is similarly split up:

image


luchaos (approved):
lgtm

PR #938, opened 2022-04-08
Add HTTP tests for Web & Connect APIs

luchaos:
... to quickly test for PHP 8 compatibility.
I think those are for JetBrains IDEs only, at least I could not find any reference as to whether those can be run by another HTTP client. Added curl notations for reference.

Includes all of Web API but only a few Connect API requests as the latter was acceptance tested with an actual client (latest RALibretro).

### DIFF
--- a/public/API/API_GetAchievementOfTheWeek.php
+++ b/public/API/API_GetAchievementOfTheWeek.php
@@ -15,7 +15,7 @@ if (empty($achievementID)) {
         'Achievement' => ['ID' => null],
         'StartAt' => null,
     ]);
-    return;
+    exit;
 }
 
 $achievementData = GetAchievementMetadataJSON($achievementID);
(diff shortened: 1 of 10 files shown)


tests/http/http-client.env.json
```diff
@@ -0,0 +1,16 @@
+{
+  "dev": {
+    // Override these in http-client.private.env.json
```
luchaos, line 3: PHPStorm HTTP client ignores invalid comments in JSON files

public/API/API_GetAchievementOfTheWeek.php
```diff
@@ -15,7 +15,7 @@
         'Achievement' => ['ID' => null],
         'StartAt' => null,
     ]);
-    return;
+    exit;
```
luchaos, line 18: Prevent any further execution

Jamiras (approved):
If you believe these are JetBrains specific, you should document as such with a comment at the top of the files. I certainly don't recognize the format for any tools I've used.

luchaos:
> If you believe these are JetBrains specific, you should document as such with a comment at the top of the files. I certainly don't recognize the format for any tools I've used.

Good idea - done 👍

PR #942, opened 2022-04-10
Add pagination to games list; replace dev filter with has achievements filter

Jamiras:
Before:
image
image

After:
image
image

The total number of games for the console is now displayed as a subheading. The other totals have been removed as calculating them counteracts the benefits of adding pagination.

Additionally, the "filter by developer" filter has been replaced with a "has achievements" filter. Per developer achievements lists may still be accessed from their user pages.

The user may choose to look only at games with achievements, without achievements, or both.
image

Because most consoles have many more games without achievements than with achievements, pagination was added.


Blazekickn:
ranted on discord already, but when looking at games with achievements on a specific console, would it be possible to add a way to revert it to the old display? Pages has benefits but i still prefer having all the games displayed when actually trying to find something to play.

PR #943, opened 2022-04-10
hide controlpanel sidebar until user is registered

Jamiras:


public/controlpanel.php
```diff
@@ -573,6 +573,7 @@ function onUploadImageComplete(data) {
             <?php endif ?>
         </div>
     </div>
+    <?php if ($permissions >= Permissions::Registered): ?>
```
luchaos, line 576: The affected endpoints should match these permissions:
`/request/user/recalculate-score.php`
`/request/user/update-avatar.php`
`/request/user/remove-avatar.php`

With the change in https://github.com/RetroAchievements/RAWeb/pull/944 users may end up in the situation where they have a custom avatar, change their email address and are set to unregistered. Maybe the last one, remove avatar, should remain visible for that particular case?

luchaos (approved):

public/controlpanel.php
Jamiras: I've updated the other endpoints, as well as the reorderSiteAwards page (also linked in the sidebar).

I have decided not to leave the Remove Avatar functionality visible. The user can either validate their account to access the remove functionality, or have an admin do it for them. The ability to change their email remains present even for unregistered users, so if it's important enough that they want it removed, they could switch to another email they can verify long enough to to the removal.

luchaos: fair enough, sgtm 👍

PR #944, opened 2022-04-10
reset account to unverified when user changes their email address

Jamiras:
this forces the user to verify their new email address.

if the user has elevated privileges, they will be warned about losing them:
image

### DIFF
--- a/lib/util/mail.php
+++ b/lib/util/mail.php
@@ -137,7 +137,7 @@ function sendValidationEmail($user, $email)
 
     // error_log(__FUNCTION__ . " sending mail to $user at address $email");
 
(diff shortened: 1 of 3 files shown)


luchaos (approved):
Will this send out a new email verification prompt mail after changing the address?

Like `sendValidationEmail()` in `/request/auth/send-verification-email.php` it should probably regenerate a validation key to be sent to the new email address?

Jamiras:
> Will this send out a new email verification prompt mail after changing the address?

It will now, but the messaging is a bit weird:

> You or someone using your email address has attempted to sign up for an account at RetroAchievements.org
> If this was you, please click the following link to confirm this email address and complete sign up
> ...

PR #946, opened 2022-04-11
lazy load images for completion list

Jamiras:
Avoids greedily downloading images for the 90% of the games in the Completion List that are not visible unless the user scrolls

image


ScottFromDerby (approved):
Love this <3

PR #947, opened 2022-04-11
Achievement messages

Tsearo:
Add more specific edit messages when making updates to achievement logic, badge, title/description and points.

Reasoning is to eliminate the ambiguity of if someone broke the logic or just did a rescore/badge update.

Example of old vs new
image


lib/database/achievement.php
```diff
@@ -597,11 +600,28 @@ function UploadNewAchievement(
...
+                    }
+
+                    $editString = implode(', ', $stringArray);
+
                     addArticleComment(
                         "Server",
                         \RA\ArticleType::Achievement,
                         $idInOut,
-                        "$author edited this achievement.",
+                        "$author edited this achievements $editString.",
```
Jamiras, line 624: possessive form should be "achievement's".

lib/database/achievement.php
```diff
@@ -527,13 +527,16 @@ function UploadNewAchievement(
...
-        $query = "SELECT Flags, Points, Author FROM Achievements WHERE ID='$idInOut'";
+        $query = "SELECT Flags, MemAddr, Points, Title, Description, BadgeName, Author FROM Achievements WHERE ID='$idInOut'";
         $dbResult = s_mysql_query($query);
         if ($dbResult !== false && mysqli_num_rows($dbResult) == 1) {
             $data = mysqli_fetch_assoc($dbResult);
 
             $changingAchSet = ($data['Flags'] != $type);
             $changingPoints = ($data['Points'] != $points);
+            $changingBadge = ($data['BadgeName'] != $badge);
+            $changingWording = ($data['Title'] != $title | $data['Description'] != $desc);
```
Jamiras, line 538: two bars (`||`) for logical or.

Tsearo: Oof missed that, thanks

lib/database/achievement.php
Tsearo: Thanks, should be correct now

Jamiras (approved):

PR #953, opened 2022-04-15
Only run code quality tools on changed files

luchaos:
- Run eslint in GitHub action as part of CI
- Re-run Rector

### DIFF
--- a/.hooks/pre-push
+++ b/.hooks/pre-push
@@ -8,7 +8,7 @@ if [ ${#modifiedPhpFiles} -gt 0 ]; then
     vendor/bin/phpstan analyse --memory-limit 512M --no-progress
 
     printf "\n⏳ composer lint\n"
-    vendor/bin/php-cs-fixer fix --verbose --dry-run --show-progress=none ${modifiedPhpFiles}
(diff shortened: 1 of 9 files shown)


.hooks/pre-push
```diff
@@ -1,9 +1,29 @@
...
+if [ ${#modifiedPhpFiles} -gt 0 ]; then
+    printf "\n⏳ composer analyse\n"
+    vendor/bin/phpstan analyse --memory-limit 512M --no-progress
+
+    printf "\n⏳ composer lint\n"
+    vendor/bin/php-cs-fixer fix --verbose --dry-run --show-progress=none --config=.php-cs-fixer.php ${modifiedPhpFiles}
+
+#    printf "\n⏳ composer test\n"
+#    phpunit
+fi
```
Jamiras, line 15: Without the "&&" chains, I don't think this will stop when an error occurs. Maybe you should add an `|| exit 1` to each command?

luchaos: The `e` flag should take care of this:
```
# exit on error
set -e
```

> -e  Exit immediately if a command exits with a non-zero status.

It stopped for me when encountering errors.

Jamiras: Ah, missed that. And/or didn't realize that's what it did. :)

Jamiras (approved):

PR #956, opened 2022-04-18
Expose offset and count parameters in GetAchievementUnlocks API

Biendeo:
Just testing around with the API, and I found out that, while the internal `getAchievementWonData` has some pagination implemented, it's currently not exposed to the `API_GetAchievementUnlocks.php` endpoint, preventing APIs from getting more than 50 unlock details for each achievement.

This PR adds two parameters, `c` and `o` for the count and offset respectively, and passes them into the internal method. They default to 50 for the count (max of 500), and 0 for the offset.

### DIFF
--- a/public/API/API_GetAchievementUnlocks.php
+++ b/public/API/API_GetAchievementUnlocks.php
@@ -7,6 +7,8 @@ runPublicApiMiddleware();
 
 $user = null;
 $achievementID = (int) (requestInputQuery('a') ?? null);
+$count = min(requestInputQuery('c', 50), 500);
+$offset = requestInputQuery('o', 0);
 
 if (empty($achievementID)) {
     echo json_encode([
@@ -38,7 +40,7 @@ $console = [
     'Title' => $achievementData['ConsoleName'] ?? null,
 ];
 
-getAchievementWonData($achievementID, $numWinners, $numPossibleWinners, $numRecentWinners, $winnerInfo, $user);
+getAchievementWonData($achievementID, $numWinners, $numPossibleWinners, $numRecentWinners, $winnerInfo, $user, $offset, $count);
 
 usort($winnerInfo, fn ($a, $b) => strtotime($a['DateAwarded']) - strtotime($b['DateAwarded']));
 


Jamiras:
While I approve of these changes from the API standpoint, there's a fundamental flaw in the way `getAchievementWonData` handles `$offset` and `$count`. In all places where the function is currently used, `$offset` is always 0, but `$count` varies.

The problem is that we store separate records for hardcore vs. non-hardcore unlocks, so when you ask for the 5 most recent winners, we query 10 rows and merge the hardcore/non-hardcore unlocks. If all ten rows are five unlocks with both a hardcore and non-hardcore entry, then you get back 5 pieces of data. If any one of those rows is just the non-hardcore entry, then you'll get back 6 pieces of data. If all of them are only non-hardcore unlocks, you'll get back 10 pieces of data.

Similarly, when you ask for offset 5, we only skip the 5 most recent rows. which is most likely two hardcore/non-hardcore unlocks, and half of the third hardcore/non-hardcore unlock. As a result, you get back the other part of the third, then the next 4-5 entries (i.e. entries 3-8 where the 3rd and 8th are only partially correct).

You can see this by going to any achievement page where there are more than 50 earners. We call the function to return the last 50 results, but if there's any non-hardcore unlocks in the list and you actually count the entries displayed, you'll find there's more than 50 shown.

As such, I can't recommend allowing this change as it's not going to function as intended.

Biendeo:
Ah, in that case perhaps we can close this, and come at it from a different angle.

At a high level, I was expecting to effectively be able to go through all the pages and, inevitably, return one entry for each user who has unlocked the achievement including the state and date of each unlock (e.g. with [achievement 342](https://retroachievements.org/achievement/342), there should be 28979 entries after accounting for all pages, each with unique users). But if it'll return the same user twice for example for non-hardcore and hardcore unlocks, then *either* I'll be needing a different SQL query under the hood would need to be called, or the API method needs to do some transformation to smoosh it down. But that'd also could change the behaviour of this endpoint, so it'd be more ideal to introduce a new endpoint to do the intended behaviour.

Also, side question, how does resetting achievement progress persist? Would a user who immediately resets after an unlock still show up in this API method, and could re-achieving something allow the user to show up twice?

luchaos:
Superceded by #973

PR #958, opened 2022-04-19
Ticket filter

Tsearo:
Adds in the following to the Ticket Manager:
Dev Status Filter - Inactive, Active, Junior
Karma Filter - Only show tickets where the Resolver is not the Developer (aka tickets worth karma)
Developer Filter - Added in a search bar to be able to search more easily
Reporter Filter - Search for tickets reported by a specific user
Resolver Filter - Search for tickets resolved by a specific user

So now you can do searches for various new things like
Tickets reported by Infernum
Tickets resolved by televandalist besides his own
Current tickets for Jr Devs
Current tickets for Inactive Devs
And so on

Also added in a Resolver column to the table but only display it if Closed || Resolved are selected

Additional joins required for Resolved by and Karma done on the database are conditional so they will only grab the extra data if the appropriate filters call for it.

image


public/ticketmanager.php
```diff
@@ -215,17 +232,31 @@
...
+            sanitize_outputs(
+                $reportedByUser,
+            );
+            $resolvedByUser = requestInputSanitized('r', null);
+            if (!isValidUsername($resolvedByUser)) {
+                $resolvedByUser = null;
+            }
+            sanitize_outputs(
+                $resolvedByUser,
+            );
```
Jamiras, line 248: `sanitize_outputs` should only be used for things that are getting written to the page. These are going to be passed to `countOpenTickets` in a few lines where they'll be used in queries. That should be calling `sanitize_inputs` on them, and won't expect them to have already been escaped for HTML.

public/ticketmanager.php
```diff
@@ -398,37 +433,71 @@
...
+                if ($devJunior) {
+                    echo "<b><a href='/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&p=$reportedByUser&r=$resolvedByUser&f=$gamesTableFlag&t=" . ($ticketFilters & ~(1 << 16)) . "'>*Junior</a></b>";
+                } else {
+                    echo "<a href='/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&p=$reportedByUser&r=$resolvedByUser&f=$gamesTableFlag&t=" . ($ticketFilters | (1 << 16)) . "'>Junior</a>";
+                }
+                echo "</div>";
+
+                // Karma Filters
+                echo "<div>";
+                echo "<b>Karma:</b> ";
```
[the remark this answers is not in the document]
Tsearo: I wasn't fond of using that term tbh but I find it better than something like Resolved By: Not Self or Other. If you have a simple rename suggestion I can do that but I had some reasoning on not splitting it out into multiple filters at least for this implementation.

I could make it work with something like two flags Author | Other but it kind of complicates the filter requirements to needing either one of these active when filtering Closed | Resolved while also keeping both of them inactive when searching only Open. I could automate the filters to do this but I didn't want to meddle with that and confuse users even more. As a single toggle it works better imo where it works regardless of Open/Closed/Resolved status while off and only requires Closed | Resolved when activated.

public/ticketmanager.php
Tsearo: $assignmedToUser, $resolvedByUser and $reportedByUser actually do get written out after selecting a developer for any of them to notate what value it currently has. So I think in that case it is safe to keep these sanitize_output calls correct?
image

I did clean up the text boxes though which were also displaying the value which I didn't think of until now.

lib/database/ticket.php
```diff
@@ -471,12 +494,45 @@ function countOpenTickets(
...
+    $reporterJoin = "";
+    if ($reportedByUser != null) {
+        $reporterJoin = "LEFT JOIN UserAccounts AS ua ON ua.ID = tick.ReportedByUserID";
+        $reporterCond = " AND ua.User LIKE '$reportedByUser'";
+    }
+
+    // Resolver condition
+    $resolverCond = "";
+    if ($resolvedByUser != null) {
+        $resolverCond = " AND ua2.User LIKE '$resolvedByUser'";
```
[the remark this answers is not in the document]
Tsearo: Yep you're right, I added it to $resolverJoin instead and also swapped that check from the karmaCond check into the $resolverJoin it sets since it used that same join statement.

public/ticketmanager.php
```diff
@@ -294,102 +325,106 @@
...
+                $devActive = ($ticketFilters & (1 << 15));
+                $devJunior = ($ticketFilters & (1 << 16));
+                $karmaStatus = ($ticketFilters & (1 << 17));
 
                 // State Filters
                 echo "<div>";
                 echo "<b>Ticket State:</b> ";
                 if ($openTickets) {
-                    echo "<b><a href='/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&f=$gamesTableFlag&t=" . ($ticketFilters & ~(1 << 0)) . "'>*Open</a></b> | ";
+                    echo "<b><a href='/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&p=$reportedByUser&r=$resolvedByUser&f=$gamesTableFlag&t=" . ($ticketFilters & ~(1 << 0)) . "'>*Open</a></b> | ";
```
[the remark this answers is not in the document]
Tsearo: Yeah you're right, did as suggested to clean up most cases but left a few of the unique ones as is for now.

public/ticketmanager.php
```diff
@@ -294,102 +325,108 @@
...
                 } else {
-                    echo "<a href='/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&f=$gamesTableFlag&t=" . ($ticketFilters | (1 << 10)) . "'>Emulator Unknown</a>";
+                    echo "<a href='$standardFilterURL" . ($ticketFilters | (1 << 10)) . "'>Emulator Unknown</a>";
                 }
                 echo "</div>";
 
                 // Core/Unofficial Filters - These filters are mutually exclusive
                 echo "<div>";
                 echo "<b>Achievement State:</b> ";
+                $gamesFilterUrl = '/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&p=$reportedByUser&r=$resolvedByUser&f=';
```
Jamiras, line 419: this needs to be in double quotes to expand the variables

public/ticketmanager.php
```diff
@@ -294,102 +325,108 @@
...
                 $modeSoftcore = ($ticketFilters & (1 << 13));
+                $devInactive = ($ticketFilters & (1 << 14));
+                $devActive = ($ticketFilters & (1 << 15));
+                $devJunior = ($ticketFilters & (1 << 16));
+                $notAuthorStatus = ($ticketFilters & (1 << 17));
 
                 // State Filters
                 echo "<div>";
                 echo "<b>Ticket State:</b> ";
+                $standardFilterURL = '/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&p=$reportedByUser&r=$resolvedByUser&f=$gamesTableFlag&t=';
```
Jamiras, line 336: this needs to be in double quotes to expand the variables

public/ticketmanager.php
Jamiras: My point is that you're escaping the value before using it in a query. I think `sanitize_outputs` only escapes `<` and `>` because of the flags that we pass to `htmlentities`, but if those were valid username characters, and you tried to match `User<DEV>`, `$resolvedByUser` would be `User&lt;DEV&gt;` going into `countOpenTickets`, and wouldn't match the `User<DEV>` in the database.

You should either sanitize them later (after they've been used in the query), or create a copy to sanitize.

public/ticketmanager.php
Jamiras: You should just use `$standardFilterURL` here. It already contains `f=$gamesTableFlag`, which is the only thing you've extracted, and it immediately gets added back by the next couple of lines. The only difference is the explicit mapping from everything that's not 5 to 3. You could do that before constructing `$standardFilterUrl` so it's correct for all the other links too.

public/ticketmanager.php
```diff
@@ -464,6 +594,9 @@
                 echo "<th>Game</th>";
                 echo "<th>Developer</th>";
                 echo "<th>Reporter</th>";
+                if ($closedTickets || $resolvedTickets) {
+                    echo "<th>Resolver</th>";
+                }
```
Jamiras, line 599: Would it also make sense to hide the "Resolver" search field and "Resolved By: Not Author" selector if neither the closed or resolved subfilters are selected?

public/ticketmanager.php
```diff
@@ -551,14 +693,14 @@
                     if ($prevOffset < 0) {
                         $prevOffset = 0;
                     }
-                    echo "<a href='/ticketmanager.php?g=$gameIDGiven&u=$assignedToUser&f=$gamesTableFlag&t=$ticketFilters'>First</a> - ";
-                    echo "<a href='/ticketmanager.php?o=$prevOffset&g=$gameIDGiven&u=$assignedToUser&f=$gamesTableFlag&t=$ticketFilters'>&lt; Previous $maxCount</a> - ";
+                    echo "<a href='$standardFilterURL" . "$ticketFilters'>First</a> - ";
+                    echo "<a href='/ticketmanager.php?o=$prevOffset&g=$gameIDGiven&u=$assignedToUser&p=$reportedByUser&r=$resolvedByUser&f=$gamesTableFlag&t=$ticketFilters'>&lt; Previous $maxCount</a> - ";
```
Jamiras, line 697: Could use `standardFilterURL` here (and the next couple links) too. Just move the `o=` parameter to the end.

public/ticketmanager.php
Tsearo: I originally did this but these variables are retained so if you uncheck both Closed and Resolved but still have Resolver set it will not be visible but still affect the query. I opted to just always show them to not interfere too much with the users filter settings but the other option I considered was blanking out Resolver and unchecking Resolved By if both Closed and Resolved become unchecked. It would just be doing an extra check when unsetting Closed or Resolved to see if the other is unchecked and then modify the url to blank Resolver and uncheck Resolved By.

public/ticketmanager.php
Tsearo: I've tried to do this but I'm not seeing how it is possible since `$gamesTableFlag` needs to be set before `$standardFilterUrl` in order for it to take effect (which is fine for just changing !=5 to 3) but the `Core | Unofficial` filters have to have different values of `$gamesTableFlag`. All Urls need to follow `$standardFilterUrl` except for the opposite of the current `$gamesTableFlag` filter which can't be done without redefining `$standardFilterUrl` in one of the else statements which would then need to undo and go back to the original value right after. For now I will keep it as is and fix `$gamesFilterUrl` with double quotes.

Jamiras: You're right. For the ones that change the value, you need to construct it with the alternate value, but for the ones where the value isn't changing, you could leverage the `$standardFilterUrl`. It's probably fine to reconstruct those too for readability.

Jamiras, line 419: Still need to convert the single quotes to double quotes so the variables will be expanded.

public/ticketmanager.php
Jamiras: You could just blank it out when reading the variables too.
```
        $resolvedByUser = requestInputSanitized('r', null);
        if ((!$closedTickets && !$resolvedTickets) || !isValidUsername($resolvedByUser)) {
            $resolvedByUser = null;
        }
```

public/ticketmanager.php
Tsearo: Ended up creating a few template Url variables and eliminated almost all other calls now to standardize them as much as I could.

public/ticketmanager.php
Tsearo: Removed `$gamesFilterUrl` but converted to double quotes in the new Url variables created

public/ticketmanager.php
Tsearo: This came down to a similar situation as with the `$gamesTicketFilter` where even if I set `$resolvedByUser` before `$standardFilterUrl` based on closed and resolved status I would still need specific opposing Urls for the specific scenarios. Since these scenarios already take care of setting `$resolvedByUser` I didn't need to add the example provided. It now unsets those values when Closed and Resolved are not checked and hides Resolved By and Resolver sections.

public/ticketmanager.php
Tsearo: Oh wow yeah I misunderstood what you were saying and what I was doing with those, I believe I fixed this now by moving it down to where the html starts building and also cleaned up the sanitize_outputs call to combine the three variables into one call since it supports that.

Tsearo:
Just realized my commit didn't get pushed through the other day... I believe I've covered all of the items you've given feedback on with this one. I greatly appreciate the time and advice given as well

public/ticketmanager.php
```diff
@@ -604,6 +763,7 @@
                 echo "<th>Game</th>";
                 echo "<th>Developer</th>";
                 echo "<th>Reporter</th>";
+                echo "<th>Resolver</th>";
```
Jamiras, line 766: It looks like you forgot to actually populate the column for this header:
image

public/ticketmanager.php
```diff
@@ -398,37 +440,75 @@
...
 
+                // Resolved By Filter
+                if ($closedTickets || $resolvedTickets) {
+                    echo "<div>";
+                    echo "<b>Resolved By:</b> ";
+
+                    if ($notAuthorStatus) {
+                        echo "<b><a href='$standardFilterUrl" . ($ticketFilters & ~(1 << 17)) . "'>*Not Author</a></b> ";
+                    } else {
+                        echo "<a href='$standardFilterUrl" . ($ticketFilters | (1 << 17)) . "'>Not Author</a> ";
```
Jamiras, line 488: I definitely like this better than the previous text, but I'm worried that "Author" is ambiguous between the achievement author and the ticket author. Maybe "Not Developer" to match the column in the list, or "Not Creator" to match the "Created by <user>" shown on the achievement page?

Jamiras (approved):

PR #961, opened 2022-04-21
return failure messages when provided with an invalid file

Jamiras:
Fixes an issue when uploading a file where the extension (implied mime type) didn't match the contents. When uploading game badges, it would return a 500 Internal Error error. On the user profile, it would silently fail (it was probably still 500ing in the ajax request).

### DIFF
--- a/public/doupload.php
+++ b/public/doupload.php
@@ -25,7 +25,16 @@ if (isset($_FILES["file"]) && isset($_FILES["file"]["name"])) {
 
 switch ($requestType) {
     case "uploadbadgeimage":
-        $response['Response'] = UploadBadgeImage($_FILES["file"]);
+        $uploadResponse = UploadBadgeImage($_FILES["file"]);
+        $response['Success'] = $uploadResponse['Success'];
+        unset($uploadResponse['Success']);
+
+        if ($uploadResponse['Error']) {
+            $response['Error'] = $uploadResponse['Error'];
+            unset($uploadResponse['Error']);
+        }
+
+        $response['Response'] = $uploadResponse;
         break;
 
     default:
--- a/public/request/user/update-avatar.php
+++ b/public/request/user/update-avatar.php
@@ -17,7 +17,16 @@ if (!RA_ReadCookieCredentials($user, $points, $truePoints, $unreadMessageCount, 
     exit;
 }
 
-$response['Response'] = UploadUserPic($user, $filename, $rawImage);
-
+$uploadResponse = UploadUserPic($user, $filename, $rawImage);
+$response['Success'] = $uploadResponse['Success'];
+unset($uploadResponse['Success']);
 settype($response['Success'], 'boolean');
+
+if ($uploadResponse['Error']) {
+    $response['Error'] = $uploadResponse['Error'];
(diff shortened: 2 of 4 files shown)


public/doupload.php
```diff
@@ -25,7 +25,16 @@
...
+        $uploadResponse = UploadBadgeImage($_FILES["file"]);
+        $response['Success'] = $uploadResponse['Success'];
+        unset($uploadResponse['Success']);
+
+        if ($uploadResponse['Error']) {
+            $response['Error'] = $uploadResponse['Error'];
+            unset($uploadResponse['Error']);
+        }
+
+        $response['Response'] = $uploadResponse;
```
luchaos, line 37: Is this kept for backwards compatibility in the clients or can we remove the `Response` key?
This would allow to shorten it to `$response = UploadBadgeImage($_FILES["file"])` to be in line with `dorequest` responses where `Response` is used as "data payload". In this case there is no actual data to be returned?

public/request/user/update-avatar.php
```diff
@@ -17,7 +17,16 @@
...
+unset($uploadResponse['Success']);
 settype($response['Success'], 'boolean');
+
+if ($uploadResponse['Error']) {
+    $response['Error'] = $uploadResponse['Error'];
+    unset($uploadResponse['Error']);
+}
+
+$response['Response'] = $uploadResponse;
+
```
luchaos, line 31: As this endpoint is only used internally i'd prefer it to be simplified to `$response = UploadUserPic($user, $filename, $rawImage);` as suggested above.

luchaos (commented):
Apart from the two comments (open for discussion) I'd like to refactor the upload code a bit further in a separate commit if that's ok for you?

Should I state a separate PR for that instead? I'd probably touch the S3 environment check as well to make minio work.

public/doupload.php
luchaos: Looking at `dorequest.php` I can see where this comes from with all those `Passthru`s. I wouldn't use that as a template as it's very inconsistent in `dorequest.php` and the more verbose variant of handling the response. I'd go simple by default in this case?

Jamiras: Yes, this is very much for compatibility with the existing API usages.

https://github.com/RetroAchievements/RAIntegration/blob/552049a1b5d93b3bfed0c88529c3713330f4ca15/src/api/impl/ConnectedServer.cpp#L1388-L1401

The client could be modified to check for either, then we could eliminate it. `r=submitticket` has the same behavior.

public/request/user/update-avatar.php
Jamiras: done

public/doupload.php
Jamiras: > In this case there is no actual data to be returned?

It's returning the ID associated to the new badge:
```
{"Success":true,"Response":{"Filename":"uploadbadgeimage.jpg","Size":4260,"BadgeIter":"61838"}}
```

luchaos: got it - sgtm then 👍

luchaos (approved):

PR #962, opened 2022-04-22
ensure variable type is specified for match clauses

Jamiras:
This fixes the issue where the `(subscribe)` link was not changing to `(unsubcribe)` after subscribing.

#911 changed several `switch` statements to `match` expressions. It turns out that `switch` will coerce the data type, but `match` will not. This led to a situation where a variable that was sanitized for SQL got converted to a string and was no longer being matched by the `match` clauses.

I went through and audited all of the `match` clauses added in #911 and made sure the variable type was well known at the point where the `match` is evaluated. It turns out there was only a couple questionable cases, which I have addressed in this PR.

### DIFF
--- a/lib/render/leaderboard.php
+++ b/lib/render/leaderboard.php
@@ -395,6 +395,7 @@ function getGlobalRankingData($lbType, $sort, $date, $user, $friendsOf = null, $
 {
     $pointRequirement = "";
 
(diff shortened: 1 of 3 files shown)


luchaos (approved):
Awesome, thanks!

PR #963, opened 2022-04-23
Fix RR total players used for calc and remove softcore users from total

Tsearo:
Checking the RR calculation I noticed it was not calculating total users correctly and was instead using the total earners of the achievement with the most earns. You can see an example below

Current
1/2 and 2/2
image

Fixed
1/10 and 2/10
image

The other fix is removing softcore users from the calculation, it now counts the total hardcore users that earned the achievements as well as uses the total unique hardcore players for the total to check against. Example below with 34 hardcore users and 37 total

Current
26/37
image

Fixed
1/34
image

### DIFF
--- a/lib/database/achievement.php
+++ b/lib/database/achievement.php
@@ -1099,7 +1099,9 @@ function recalculateTrueRatio($gameID)
     $query = "SELECT ach.ID, ach.Points, COUNT(*) AS NumAchieved
               FROM Achievements AS ach
               LEFT JOIN Awarded AS aw ON aw.AchievementID = ach.ID
-              WHERE ach.GameID = $gameID AND ach.Flags = 3 AND aw.HardcoreMode = 1
+              LEFT JOIN UserAccounts AS ua ON ua.User = aw.User
+              WHERE ach.GameID = $gameID AND ach.Flags = 3 AND aw.HardcoreMode = 1 
+              AND NOT ua.Untracked
               GROUP BY ach.ID";
 
     $dbResult = s_mysql_query($query);
--- a/lib/database/achievement.php
+++ b/lib/database/achievement.php
@@ -976,7 +976,7 @@ function getAchievementWonData(
     $numWinners = $data['NumEarned'];
     $gameID = $data['GameID'];   // Grab GameID at this point
 
-    $numPossibleWinners = getTotalUniquePlayers($gameID, $user);
+    $numPossibleWinners = getTotalUniquePlayers($gameID, $user, false);
 
     $numRecentWinners = 0;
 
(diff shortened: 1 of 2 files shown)


lib/database/achievement.php
```diff
@@ -1099,41 +1099,32 @@ function recalculateTrueRatio($gameID)
     $query = "SELECT ach.ID, ach.Points, COUNT(*) AS NumAchieved
               FROM Achievements AS ach
               LEFT JOIN Awarded AS aw ON aw.AchievementID = ach.ID
-              WHERE ach.GameID = $gameID AND ach.Flags = 3 AND aw.HardcoreMode = 0
+              WHERE ach.GameID = $gameID AND ach.Flags = 3 AND aw.HardcoreMode = 1
               GROUP BY ach.ID";
```
Jamiras, line 1103: This query doesn't account for Untracked users, so the NumAchieved will sometimes be slightly higher than expected. In my tests, it didn't seem to affect the resulting calculation by more than a fraction of a point, so it's probably okay to ignore it.

Jamiras (commented):
Overall, I think the changes are good, but I want to clarify that in most cases the overall retro ratio is going to go _down_.

In your examples, the numbers go up because the ratio is becoming smaller:
1/2 (0.5) -> 1/10 (0.1)
26/37 (0.70) -> 1/34 (0.03)

However, when I run this against the test database, I see everything going down because the ratio is actually getting larger:

1575/3913 (0.4025)
image

1209/2657 (0.4550)
image

Also note the 1209 doesn't match the 1192 in the image. That's because there are 17 untracked users being included in the calculation. 1192/2657 is still 0.4486, so the final value wouldn't really change.

lib/database/achievement.php
Tsearo: Good callout, it's a minor effect like you said but I'd rather have it be as accurate as possible so I added that to the TrueRatio calc query. I'm sure there could be an edge case on an easy achievement where the retro points could be lower than the actual depending on untracked users.

Tsearo:
Yeah I was debating showing more examples of how RR would change I should have probably made that more clear, the ones I used were mainly to showcase the fact that the current calculation was incorrect.

For some unaltered examples of before and after based on awarded data from 2020

Super Mario World - 556
Before - (3074)
After - (1957)
After with Untracked removed - (1982)

Donkey Kong Country - 386
Before - (1141)
After - (820)
After with Untracked removed - (830)

~Bonus~ Super Mario World - 1340
Before - (68791)
After - (73577)
After with Untracked removed - (73995)

So yeah overall it may be lowering a lot of games (I'm sure we'll be hearing plenty about peoples RR getting nerfed...) but they will be more accurate as they only account for hardcore user. Which is likely something that would need to be done when softcore points are eventually removed in the future.

Jamiras (approved):

PR #964, opened 2022-04-23
add NumAchievements,NumLeaderboards,Points, and Hashes to API_GetGameList

Jamiras:
Previous response:
```
{
  "Title":"Ninja Golf",
  "ID":"13352",
  "ConsoleID":"51",
  "ImageIcon":"\/Images\/019885.png",
  "ConsoleName":"Atari 7800"
}
```

New response:
```
{
  "Title":"Ninja Golf",
  "ID":"13352",
  "ConsoleID":"51",
  "ConsoleName":"Atari 7800",
  "ImageIcon":"\/Images\/019885.png",
  "NumAchievements":31,
  "NumLeaderboards":0,
  "Points":485,
  "Hashes":["bb551483d114dcc5bae7aaa892ee856a","220121f771fc4b98cef97dc040e8d378"]
}
```

Note that the `"Hashes"` subdata is only available if `h=1` is passed to the API. `f=1` will limit the response to games with achievements (existing behavior).

The new fields are returns as integers where appropriate. I did not "fix" the previous fields that should have been returned as integers as that's a potentially breaking change.

### DIFF
--- a/lib/database/game.php
+++ b/lib/database/game.php
@@ -4,27 +4,6 @@ use RA\ActivityType;
 use RA\ArticleType;
 use RA\Permissions;
 
-function getGameFromHash($md5Hash, &$gameIDOut, &$gameTitleOut)
-{
-    sanitize_sql_inputs($md5Hash);
-
-    $query = "SELECT ID, GameName FROM GameData WHERE GameMD5='$md5Hash'";
-    $dbResult = s_mysql_query($query);
-
-    if ($dbResult !== null) {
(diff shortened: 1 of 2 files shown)


luchaos (approved):
IDs as strings is preferred, even if their in-db type is an integer. This is more in line with how JSON-API conform APIs would treat IDs, as they may as well be UUIDs. Not that it matters much in V1 but I think it's fine how it is; IDs are strings, counts and aggregates are ints 👍

PR #970, opened 2022-05-02
add TicketFilters enum

Jamiras:
Also updates the ticket manager filter code to use helper functions for modifying the URL and makes the user search fields and game/achievement filters behave similarly to the bitflags - the "Clear Filter" link has been removed. Instead, clicking on the selected filter will clear it. 

This feels a little weird with the Game/Achievement filters, but is consistent, as clicking on the Game link deselects the game instead of only deselecting the Achievement. The breadcrumb link (at the top of the page) does take the user back to the list for all ticketed achievements for the game. Maybe the Game/Achievement filters should just be removed as they're redundant with the breadcrumbs, and can only be removed.

Before:
image

After:
image

And with selected users:
image


luchaos (approved):

Tsearo:
Sorry I started a review a few days ago but I guess it never submitted. Here are the details I mentioned

I noticed an issue when calling `TicketFilters::All` including `ResolvedByNonAuthor` which causes issues with `$altTicketData = getAllTickets(0, 99, null, null, null, null, $ticketData['AchievementID'], TicketFilters::All);` in `ticketmanager.php`. I noticed the issue when going to a ticket with other related closed and open tickets which were not displaying due to the ResolvedByNonAuthor flag being checked. I believe it should instead be `public const All = (1 << 17) - 1` which would be all bits 0-16 active instead. In doing so I'm not sure if `& ~self::ResolvedByNonAuthor` would need to be included anymore

PR #980, opened 2022-05-11
add warnings for account deletion

Jamiras:
Adds a banner on all pages of the website, and an email when the action is first taken. This should make it obvious to a user when their account is marked for deletion, whether intentional, accidental, or malicious.

image

> Hello TestUser1,
> Your account has been marked for deletion.
> If you do not cancel this request before 2022-05-25, you will no longer be able to access your account.
> Thanks!
> -- Your friends at RetroAchievements.org

With these changes (and the logging added in #914), we should be able to re-enable the cron job that actually does the deleting.

We should probably wait a few days after publishing these changes so users can see the banner (which will report a time in the past) in case they aren't aware there's a pending deletion request and want to cancel it.

### DIFF
--- a/lib/database/auth.php
+++ b/lib/database/auth.php
@@ -89,12 +89,6 @@ function validateUser_app(&$user, $token, &$fbUser, $permissionRequired): bool
     );
 }
 
-function getCookie(&$userOut, &$cookieOut)
-{
-    $userOut = RA_ReadCookie('RA_User');
-    $cookieOut = RA_ReadCookie('RA_Cookie');
-}
-
 function RA_ValidateCookie(
     &$userOut,
     &$permissionsOut,
--- a/lib/database/auth.php
+++ b/lib/database/auth.php
@@ -89,42 +89,6 @@ function validateUser_app(&$user, $token, &$fbUser, $permissionRequired): bool
     );
 }
 
-function validateUser_cookie(&$user, $cookie, $permissionRequired, &$permissions = 0): bool
-{
-    return validateFromCookie($user, $points, $permissions, $permissionRequired);
-}
-
-function validateFromCookie(&$userOut, &$pointsOut, &$permissionsOut, $permissionRequired = 0): bool
-{
-    $userOut = RA_ReadCookie("RA_User");
-    $cookie = RA_ReadCookie("RA_Cookie");
-
-    sanitize_sql_inputs($userOut);
-
-    if (mb_strlen($userOut) < 2 || mb_strlen($cookie) < 2 || !isValidUsername($userOut)) {
-        // There is no cookie
-        return false;
-    } else {
-        // Cookie maybe stale: check it!
-        $query = "SELECT User, cookie, RAPoints, Permissions FROM UserAccounts WHERE User='$userOut'";
-        $dbResult = s_mysql_query($query);
-        if ($dbResult == false) {
-            return false;
-        } else {
-            $data = mysqli_fetch_array($dbResult);
-            if ($data['cookie'] == $cookie) {
-                $pointsOut = $data['RAPoints'];
-                $userOut = $data['User']; // Case correction
-                $permissionsOut = $data['Permissions'];
-
-                return $permissionsOut >= $permissionRequired;
-            } else {
-                return false;
-            }
-        }
-    }
-}
-
 function getCookie(&$userOut, &$cookieOut)
 {
     $userOut = RA_ReadCookie('RA_User');
--- a/lib/database/auth.php
+++ b/lib/database/auth.php
@@ -132,22 +132,27 @@ function getCookie(&$userOut, &$cookieOut)
 }
 
 function RA_ValidateCookie(
+    &$userOut,
+    &$permissionsOut,
     &$userDetailsOut,
     $minPermissions = null
 ): bool {
-    $user = RA_ReadCookie('RA_User');
+    $userOut = RA_ReadCookie('RA_User');
     $cookie = RA_ReadCookie('RA_Cookie');
+    $permissionsOut = Permissions::Unregistered;
 
-    if (mb_strlen($cookie) >= 10 && isValidUsername($user)) {
-        if (getAccountDetails($user, $userDetailsOut)) {
+    if (mb_strlen($cookie) >= 10 && isValidUsername($userOut)) {
+        if (getAccountDetails($userOut, $userDetailsOut)) {
+            $permissionsOut = $userDetailsOut['Permissions'];
+
             if (strcmp($userDetailsOut['cookie'], $cookie) === 0 &&
-                    $userDetailsOut['Permissions'] != Permissions::Banned) {
+                    $permissionsOut != Permissions::Banned) {
                 // valid active account. update the last activity timestamp
-                userActivityPing($user);
+                userActivityPing($userOut);
 
                 // validate permissions for the current page if required
                 if (isset($minPermissions)) {
-                    return $permissionOut >= $minPermissions;
+                    return $permissionsOut >= $minPermissions;
                 }
 
                 // return true meaning 'logged in'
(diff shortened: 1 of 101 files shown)


lib/database/auth.php
```diff
@@ -89,113 +89,42 @@ function validateUser_app(&$user, $token, &$fbUser, $permissionRequired): bool
     );
 }
 
-function validateUser_cookie(&$user, $cookie, $permissionRequired, &$permissions = 0): bool
-{
-    return validateFromCookie($user, $points, $permissions, $permissionRequired);
-}
-
-function validateFromCookie(&$userOut, &$pointsOut, &$permissionsOut, $permissionRequired = 0): bool
```
Jamiras, line 97: `validateUser_cookie` and `validateFromCookie` were effectively the same as `RA_ReadCookieCredentials`, with different parameters. All three have been merged into a singular `RA_ValidateCookie`.

lib/database/user.php
```diff
@@ -1723,6 +1736,8 @@ function deleteRequest($username, $date = null): bool
     if ($dbResult !== false) {
         addArticleComment('Server', ArticleType::UserModeration, $user['ID'],
             $username . ' requested account deletion');
+
+        SendDeleteRequestEmail($username, $user['EmailAddress'], $date);
```
Jamiras, line 1740: Here's where the new call to send the email was added

public/APIDemo.php
```diff
@@ -4,7 +4,7 @@
 require_once __DIR__ . '/../lib/bootstrap.php';
 require_once __DIR__ . "/../src/RetroAchievementsWebApiClient.php";
 
-RA_ReadCookieCredentials($user, $points, $truePoints, $unreadMessageCount, $permissions);
+RA_ValidateCookie($user, $permissions, $userDetails);
```
Jamiras, line 7: Instead of adding another optional parameter to `RA_ReadCookieCredentials` just to pass data along to `RenderTitleBar`, `points`, `truePoints`, and `unreadMessageCount` were replaced with a `userDetails` object that is passed to `RenderHeader`, which renders both the titlebar and menu (there's no where that was only doing one or the other).

Most of the rest of the changes in the PR are just applying those changes to every file.

public/achievementinspector.php
```diff
@@ -5,7 +5,7 @@
 require_once __DIR__ . '/../vendor/autoload.php';
 require_once __DIR__ . '/../lib/bootstrap.php';
 
-if (!RA_ReadCookieCredentials($user, $points, $truePoints, $unreadMessageCount, $permissions, Permissions::Registered)) {
+if (!RA_ValidateCookie($user, $permissions, $userDetails, Permissions::Developer)) {
```
Jamiras, line 8: changed minimum permission level to match menu visibility

public/controlpanel.php
```diff
@@ -6,13 +6,7 @@
...
 require_once __DIR__ . '/../lib/bootstrap.php';
 
-if (RA_ReadCookieCredentials($user, $points, $truePoints, $unreadMessageCount, $permissions)) {
-    if (getAccountDetails($user, $userDetails) == false) {
-        // Immediate redirect if we cannot validate user!
-        header("Location: " . getenv('APP_URL') . "?e=accountissue");
-        exit;
-    }
-} else {
+if (!RA_ValidateCookie($user, $permissions, $userDetails)) {
```
Jamiras, line 9: `RA_ValidateCookie` returns the `$userDetails` without requiring a second call to fetch them.

For the second call to fail after the first succeeded, the account would have to still have a valid cookie, but be marked for deletion. This logic is now included in `RA_ValidateCookie`.

public/controlpanel.php
```diff
@@ -287,8 +280,6 @@ function validateEmail() {
                 echo "<td>";
                 echo "<form method='POST' action='/request/user/update-motto.php'>";
                 echo "<input class='fullwidth' name='m' value=\"$userMottoString\" maxlength='49' id='usermottoinput' placeholder='Add your motto here! (No profanity please!)' />";
-                echo "<input type='hidden' name='u' VALUE='$user'>";
-                echo "<input type='hidden' name='c' VALUE='$cookie'>";
```
Jamiras, line 291: This is manually passing the cookie information to the API page. It doesn't seem to be necessary as the real cookies are passed too. Many API pages rely only on the real cookies, so I think this is a safe change to make. It definitely works on Windows/Chrome.

public/createtopic.php
```diff
@@ -35,15 +29,13 @@
 $thisCategoryID = $forumData['CategoryID'];
 $thisCategoryName = $forumData['CategoryName'];
 
-getCookie($user, $cookieRaw);
```
Jamiras, line 38: This was just fetching the user's cookie to pass the the API pages. See previous comment about my belief that functionality is unnecessary.

public/leaderboardList.php
```diff
@@ -180,7 +176,7 @@ function onUpdateComplete(data) {
     }
 
     if (isset($gameData) && isset($user) && $permissions >= Permissions::JuniorDeveloper) {
-        echo "<div id='warning'>Status: OK!</div>";
+        echo "<div id='warning'></div>";
```
Jamiras, line 179: Don't show message until an API call is made

public/viewtopic.php
```diff
@@ -380,7 +380,7 @@ class="fullwidth"
             echo "</tbody></table></div>";
         } else {
             echo "</tbody></table></div>";
-            RenderLoginComponent($user, $points, $errorCode, true);
+            echo "<br/>You must log in before you can join this conversation.<br/>";
```
Jamiras, line 383: This could have passed `$userDetails['RAPoints']` as a replacement for `$points`, but having a second set of login fields on the page felt awkward.

luchaos: Thanks, that was bothering me for some time now

lib/database/auth.php
luchaos: I like those changes a lot! 

I'd like to rename those `RA_` prefixed functions to be more in line with the rest:
- `ValidateCookie`
- `ValidateToken` (dropping the redundant `Credentials`)
- `ClearCookie`
- `ReadCookie`
- `ApplyCookie` (`storecookie` already exists in php)
- `CookieExists`

`RA_ReadTokenCredentials` looks a lot like `RA_ValidateCookie` used to, where `points`, `truePoints`, and `unreadMessages` are passed by reference. Yet it seems those aren't set in that function nor used as such in `dorequrest` nor `doupload`. Those parameters can be removed if I'm not mistaken.

I can add those changes as an additional commit or PR even - what do you think?

luchaos: The more I look the more i find ... `validateUser()` still has a `fbUser` parameter, 
`validateUser_app()` isn't used anymore and can be removed, too.

luchaos: `authenticateWith...` instead of `validate...` even?

Jamiras: > I can add those changes as an additional commit or PR even - what do you think?

This PR is already touching all the files, so we should get the naming correct here to avoid another PR touching all the files.

You want me to change `RA_ValidateCookie` to `authenticateWithCookie`?
```
authenticateWithCookie($user, $permissions, $userDetails)
```
and
```
if (!authenticateWithCookie($user, $permissions, $userDetails, Permissions::Developer)) {
```

What do you think about `getUserFromCookie`?

Jamiras: > I can add those changes as an additional commit or PR even - what do you think?

I guess you're asking two separate questions here.

I think the `RA_ValidateCookie` name should be finalized in this PR. the changes to clean up `RA_ReadTokenCredentials`, `validateUser`, and `validateUser_app` should be in a separate PR.

luchaos: > I think the RA_ValidateCookie name should be finalized in this PR. the changes to clean up RA_ReadTokenCredentials, validateUser, and validateUser_app should be in a separate PR.

That sounds good 👍 

> What do you think about `getUserFromCookie`?

I think I'd prefer either `validate` or `authenticateWith` as getting it from cookie sounds like user information is stored within cookies/token/password. The reasoning for not using `validate` is that it sounds like validating a value by rules whereas it's a full authentication/authorization routine.

At the end of the day it doesn't really matter that much to me :) I'm fine with whatever you prefer here.

Jamiras: I'm torn because the function is overloaded to authenticate _and_ return the user data. The only actual input parameter is the optional minimum permissions level.

I'm going to go with `authenticateFromCookie`. For some reason, the "with" just doesn't sit well with me.

luchaos: I like that :) I wasn't too happy using `with` either 👍

luchaos (approved):

PR #981, opened 2022-05-11
use paginator for forums

Jamiras:
Replaces the list of pages with a selector (both on the forum list and in an individual topic)

before:
image

after:
image


luchaos (approved):
Great to see the pagination being consolidated into a shared function 👍

Thoooreau:
It's much cleaner now for sure, but I think it's lacking an instant click to go to the first and last pages of the thread.

I suggest "First Page" (or "First", or "<<") before the "<" and "Last Page" (or "Last", or ">>") after the ">".

PR #986, opened 2022-05-20
Add achievement type constants

luchaos:
- Remove development tutorial game checks

### DIFF
--- a/src/AwardType.php
+++ b/src/AwardType.php
@@ -16,10 +16,15 @@ abstract class AwardType
 
     public const PATREON_SUPPORTER = 6;
 
-    public static array $active = [
+    public const ACTIVE = [
         self::MASTERY,
         self::ACHIEVEMENT_UNLOCKS_YIELD,
         self::ACHIEVEMENT_POINTS_YIELD,
         self::PATREON_SUPPORTER,
     ];
+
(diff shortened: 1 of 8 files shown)


src/AwardType.php
```diff
@@ -16,7 +16,7 @@ abstract class AwardType
 
     public const PATREON_SUPPORTER = 6;
 
-    public static $active = [
+    public static array $active = [
```
Jamiras, line 19: What's the difference between "static array" here and "const" for AchievementType::FLAGS? Shouldn't they both use the same syntax?

Jamiras (approved):
I'm guessing there's more places still using 3/5 (the ticket filters page comes to mind), but we can get them in the future.

src/AwardType.php
luchaos: Constants in PHP 8.0 cannot have a type - static properties should.
I was adding the missing type as the IDE complained about it.
Now that you mention it - it should be a const, too, as it's not to be changed during runtime.
I'll change that 👍

luchaos: See https://github.com/RetroAchievements/RAWeb/pull/986/commits/902d3b5b2c5216942025d1b4f5f30cb85017bdc2

PR #987, opened 2022-05-20
don't show promote link for non-validated consoles

Jamiras:


luchaos (approved):

luchaos:
https://github.com/RetroAchievements/RAWeb/pull/988 handles invalid consoles on the server and displays an error alert to the user.

Jamiras:
I think the solution in #988 is better. When the user tries to promote they'll get an error. This implementation hides the functionality without any feedback as to why.

PR #989, opened 2022-05-20
Fix broken dev page when first achievement is 0 pointer

Tsearo:
Fixes cases of developer stat page breaking at the hardest achievement which prevents it from displaying the rest of the info when the developers first achievement (in core) is a 0 pointer.

### DIFF
--- a/public/individualdevstats.php
+++ b/public/individualdevstats.php
@@ -351,7 +351,7 @@ foreach ($userArchInfo as $achievement) {
         $firstAchievement = $achievement;
         $lastAchievement = $achievement;
     } else {
-        if ($hardestAchievement['Points'] && $achievement['Points'] && ($hardestAchievement['TrueRatio'] / $hardestAchievement['Points']) < ($achievement['TrueRatio'] / $achievement['Points'])) {
+        if ($hardestAchievement['Points'] == 0 || $hardestAchievement['Points'] && $achievement['Points'] && ($hardestAchievement['TrueRatio'] / $hardestAchievement['Points']) < ($achievement['TrueRatio'] / $achievement['Points'])) {
             $hardestAchievement = $achievement;
         }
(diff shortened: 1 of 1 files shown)


public/individualdevstats.php
```diff
@@ -351,7 +351,7 @@
         $firstAchievement = $achievement;
         $lastAchievement = $achievement;
     } else {
-        if ($hardestAchievement['Points'] && $achievement['Points'] && ($hardestAchievement['TrueRatio'] / $hardestAchievement['Points']) < ($achievement['TrueRatio'] / $achievement['Points'])) {
+        if ($hardestAchievement['Points'] == 0 || $hardestAchievement['Points'] && $achievement['Points'] && ($hardestAchievement['TrueRatio'] / $hardestAchievement['Points']) < ($achievement['TrueRatio'] / $achievement['Points'])) {
```
Jamiras, line 354: I think this would be clearer if the old logic was in a set of parenthesis, possibly on a separate lin:
```
if ($hardestAchievement['Points'] == 0 || 
    ($hardestAchievement['Points'] && $achievement['Points'] && ($hardestAchievement['TrueRatio'] / $hardestAchievement['Points']) < ($achievement['TrueRatio'] / $achievement['Points']))) {
```
For performance reasons, you could also calculate `$hardestAchievement['TrueRatio']/$hardestAchievement['Points']` outside the loop and put the new logic there.

Jamiras: Nevermind, that won't work. As `hardestAchievement` could change within the loop.

Tsearo: Good point, this was mainly just a hotfix that I pulled out from the other changes I'm making in individualdevstats.php which aren't ready yet.
I can make the formatting suggestion along with the `$easiestAchievement  == 0 ||` version below it in my next PR for this (already added them to it) if that is fine. Otherwise I'll add it here if that is preferred.

Jamiras: Fair enough.

Jamiras (approved):

PR #992, opened 2022-05-21
Fix floats from breaking stored hits in mem explained view

Tsearo:
Fix case where float values broke the stored hits in mem explained and added a case to test/trigger.php
image


lib/util/trigger.php
```diff
@@ -62,7 +62,11 @@ function parseOperand($mem)
                 if ($mem[1] == '+' || $mem[1] == '-') {
                     $count++;
                 }
-                while ($count < $max && (ctype_digit($mem[$count]) || $mem[$count] == '.')) {
+                $periodCount = 0; // Stop after reaching first period used for float
+                while ($count < $max && (ctype_digit($mem[$count]) || $periodCount < 1)) {
+                    if ($mem[$count] == '.') {
+                        $periodCount++;
+                    }
```
Jamiras, line 69: This seems like it might be too greedy. As long as a period hasn't been seen, it will grab everything.

It would probably be easier to match two numeric groups:
```
                while ($count < $max && ctype_digit($mem[$count])) {
                    $count++;
                }
                if ($count < $max && $mem[$count] == '.') {
                    $count++;
                    while ($count < $max && ctype_digit($mem[$count])) {
                        $count++;
                    }
                }
```

Tsearo: Ah yeah your method seems to be more efficient after running a few tests, thanks for the suggestion 👍

Jamiras (approved):

PR #994, opened 2022-05-22
eliminate RA_User cookie

Jamiras:
The old webpage logic matched the `RA_User` cookie to the `RA_Cookie` value in the database. This updates the logic to only look for a match with the `RA_Cookie` value. The length of the cookie has been increased to 96 characters to minimize any chance of someone guessing a valid cookie.


luchaos (approved):

lib/database/auth.php
```diff
@@ -90,37 +90,39 @@ function validateUser_app(&$user, $token, &$fbUser, $permissionRequired): bool
...
+    if (RA_CookieExists('RA_User')) {
+        RA_ClearCookie('RA_User');
+    }
 
-            if (strcmp($userDetailsOut['cookie'], $cookie) === 0 &&
-                    $permissionsOut != Permissions::Banned) {
-                // valid active account. update the last activity timestamp
-                userActivityPing($userOut);
+    $cookie = RA_ReadCookie('RA_Cookie');
+    if (getAccountDetailsFromCookie($cookie, $userDetailsOut)) {
```
luchaos, line 107: This taps into `$userDetailsOut`'s value - given the suggestion below for `getAccountDetailsFromCookie()` is implement:

```suggestion
    if ($userDetailsOut = getAccountDetailsFromCookie($cookie)) {
```

lib/database/user.php
```diff
@@ -378,6 +388,34 @@ function getAccountDetails(&$user, &$dataOut)
...
+                AND Deleted IS NULL";
+
+    $dbResult = s_mysql_query($query);
+    if ($dbResult == false || mysqli_num_rows($dbResult) !== 1) {
+        return false;
+    }
+
+    $dataOut = mysqli_fetch_array($dbResult);
+    return true;
+}
```
luchaos, line 417: To reduce the pass-by-reference usage (it's a overdone in this codebase imo):

```suggestion
function getAccountDetailsFromCookie(?string $cookie): ?array
{
    if (empty($cookie)) {
        return null;
    }

    sanitize_sql_inputs($cookie);

    // ID, User, and Permissions are used to identify the user and their permissions
    // DeleteRequested is used to show the "Your account is marked to be deleted" banner
    // RAPoints, TrueRAPoints, and UnreadMessageCount are used for the logged-in user area
    // websitePrefs allows pages to enable/disable functionality
    $query = "SELECT ID, User, Permissions, DeleteRequested,
                     RAPoints, TrueRAPoints, UnreadMessageCount,
                     websitePrefs
                FROM UserAccounts
                WHERE cookie='$cookie'
                AND Deleted IS NULL";

    $dbResult = s_mysql_query($query);
    if (!$dbResult || mysqli_num_rows($dbResult) !== 1) {
        return null;
    }

    return mysqli_fetch_array($dbResult) ?: null;
}
```

PR #996, opened 2022-05-26
add first/last links to paginator

Jamiras:
image


lib/render/layout.php
```diff
@@ -499,8 +499,10 @@ function RenderThemeSelector()
 function RenderPaginator($numItems, $perPage, $offset, $urlPrefix)
 {
     if ($offset > 0) {
+        echo "<a title='First' href='${urlPrefix}0'>&#x226A;</a>&nbsp;";
```
luchaos, line 502: Alternatively: `&laquo;`

lib/render/layout.php
```diff
@@ -513,6 +515,10 @@ function RenderPaginator($numItems, $perPage, $offset, $urlPrefix)
 
     $nextOffset = $offset + $perPage;
     if ($nextOffset < $numItems) {
-        echo "&nbsp;<a href='$urlPrefix$nextOffset'>&gt;</a>";
+        echo "&nbsp;<a title='Next' href='$urlPrefix$nextOffset'>&gt;</a>";
+
+        $lastOffset = $numItems - 1; // 0-based
+        $lastOffset = $lastOffset - ($lastOffset % $perPage);
+        echo "&nbsp;<a title='Last' href='$urlPrefix$lastOffset'>&#x226B;</a>";
```
luchaos, line 522: Alternatively: `&raquo;`

luchaos (approved):

lib/render/layout.php
Jamiras: `&laquo;` is `&#x00AB;`: «
`&#x226A;` is ≪ (much less than mathematic symbol)
The larger one looks better next to the normal less than symbol. Here's the laquo/raquo:
image

luchaos: Makes sense :) Was pointing it out for completeness' sake as laquo and raquo are already used in other places.
There are `&lsaquo;` and `&lsaquo;` - the single quote variants, too, which would match the double variants.
Then again, they are quite small.

PR #998, opened 2022-05-26
add flag to return ticket details with tickets-for-game query

Jamiras:


public/API/API_GetTicketData.php
```diff
@@ -212,6 +235,16 @@
         );
         $ticketData['URL'] = $baseUrl . "?g=$gameIDGiven";
 
+        $details = requestInputQuery('d', 0, 'integer');
+        if ($details == 1) {
+            $ticketData['Tickets'] = getAllTickets($offset, $count, null, null, null, $gameIDGiven, null, $defaultTicketFilter);
```
luchaos, line 240: With PHP 8 named arguments it's possible to omit all those default `null`s and better describe what is actually passed:
```php
getAllTickets($offset, $count, givenGameID: $gameIDGiven, ticketFilters: $defaultTicketFilter);
```

luchaos (approved):

PR #1001, opened 2022-05-26
Allow tab delimiter in admin tools

luchaos:
- Add `separateList()` which allows to split by common delimiters (tab, comma, semicolon, pipe, space)
- Clean up `public/admin.php` and `lib/util/debug.php`
- Add tests


tests/src/UtilTest.php
```diff
@@ -0,0 +1,21 @@
...
+
+final class UtilTest extends TestCase
+{
+    public function testSeparateList()
+    {
+        $expected = ['x', 10, 20, 30, 40, 50];
+
+        $this->assertEquals($expected, separateList("	x,10;20 30|40	50,	"));
+        $this->assertEquals($expected, separateList('x, 10, 20, 30, 40, 50'));
+        $this->assertEquals($expected, separateList("x	10	20	30	40	50"));
```
Jamiras, line 17: might be better to use escape sequences so the tabs don't accidentally get replaced by spaces by someone's editor

Jamiras (approved):

tests/src/UtilTest.php
luchaos: Good point, changed 👍

PR #1007, opened 2022-05-28
Fix user parameters not being optional for guest requests

luchaos:
+ some minor cleanup


luchaos:
Merged as some pages are broken otherwise.

PR #1008, opened 2022-05-28
Cleanup namespaces

luchaos:
- Remove remaining `fbUser` and `fbPrefs` references (migration to drop columns not included yet)
- Rename `ObjectType` to `RatingType` as that's what it seems to be
- Flatten `src` directory, no need to group things yet
- PascalCase for enum-like constants

### DIFF
--- a/lib/database/subscription.php
+++ b/lib/database/subscription.php
@@ -1,11 +1,11 @@
 <?php
 
+use RA\SubscriptionSubjectType;
+
 /**
  * Update a subscription, i.e, either subscribe or unsubscribe a given user to or from a subject.
  *
  * @param bool $state whether the user is to be subscribed (true) or unsubscribed (false)
- * @return bool whether the update was successful
- * @throws Exception
  */
 function updateSubscription(string $subjectType, int $subjectID, int $userID, bool $state): bool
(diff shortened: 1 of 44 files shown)


lib/database/subscription.php
```diff
@@ -185,3 +180,118 @@ function mergeSubscribers(array $subscribersA, array $subscribersB): array
 
     return $subscribersA;
 }
+
+function getSubscribersOfGameWall($gameID): array
+{
+    return getSubscribersOfArticle(1, $gameID, (1 << 1));
```
Jamiras, line 186: I realize these are just getting moved around, but it would be a good time to replace the constants with enums:
```
return getSubscribersOfArticle(ArticleType::Game, $gameID, 
    (1 << UserPreference::EmailOn_AchievementComment));
```
You could even just pass the UserPreferenceValue and do the shift inside the function.

Jamiras (approved):

lib/database/subscription.php
luchaos: I've looked into that but went down a rabbit hole. I'll open a separate PR to clean up the subscriptions.

PR #1014, opened 2022-06-03
Update dependencies

luchaos:
- Update package-lock.json format using npm 8
- Update composer.phar and dependencies

Various fixes:
- Handle empty emulator ID in dorequest latestclient
- Handle empty input for attributeEscape()
- Handle empty input for getFriendCount()
- Handle games without achievements in gamecompare dropdown
- Sanitize gamecompare outputs
- Update contributing retroachievements link schema to https


public/dorequest.php
```diff
@@ -147,11 +147,16 @@ function DoRequestError($errorMsg): void
         $emulatorId = requestInput('e');
         $consoleId = requestInput('c');
 
-        if ($emulatorId === null && $consoleId !== null) {
+        if (empty($emulatorId) && !empty($consoleId)) {
             DoRequestError("Lookup by Console ID has been deprecated");
             break;
         }
 
+        if (empty($emulatorId)) {
```
Jamiras, line 155: shouldn't this logic be tied to the `integration_id` data in releases.php?

public/gamecompare.php
```diff
@@ -58,6 +58,13 @@
     }
 }
 
+sanitize_outputs(
+    $gameTitle,
+    $consoleName,
+    $gameIcon,
+    $user,
+);
```
Jamiras, line 66: I'm leary of this change. These values are passed to `GetGameAndTooltipDiv` and `GetUserAndTooltipDiv`, which will likely sanitize them again.

public/dorequest.php
luchaos: I thought of it more like validation before passing it along but it might as well be handled in `getEmulatorReleaseByIntegrationId()`, yeah. If I understand correctly that is.

Jamiras: oh, missed that next line. isn't this redundant then?

public/gamecompare.php
luchaos: Good point - looks good so far though? 

Screenshot 2022-06-03 at 20 26 08
Screenshot 2022-06-03 at 20 26 14

public/dorequest.php
luchaos: as `getEmulatorReleaseByIntegrationId(int $integrationId)` requires an input it might be caught before even attempting to look it up. Admittedly, in this case it's not that helpful - it's just an array lookup - yet, we query the database a lot for e.g. empty usernames and rely on the query to not find anything instead of preventing even being run because the input was empty in the first place.

public/dorequest.php
```diff
@@ -223,9 +223,9 @@ function DoRequestError($errorMsg): void
 
     case "lbinfo":
         $lbID = requestInput('i', 0, 'integer');
-        $nearby = true; // Nearby entry behavior has no effect if $user is null
-        $friendsOnly = 0; // TBD
-        $response['LeaderboardData'] = GetLeaderboardData($lbID, $user, $count, $offset, $friendsOnly, $nearby);
+        // Note: Nearby entry behavior has no effect if $user is null
+        // TBD: friendsOnly
+        $response['LeaderboardData'] = GetLeaderboardData($lbID, $user, $count, $offset, friendsOnly: 0, nearby: true);
```
luchaos, line 228: Minor cleanup

public/dorequest.php
```diff
@@ -98,8 +98,8 @@ function DoRequestError($errorMsg): void
         $response['Response'] = GetAllUserProgress($user, $consoleID);
         break;
 
-    // TODO: still used?
     case "badgeiter":
+        // Used by RALibretro achievement editor
```
luchaos, line 102: Verified using RALibretro

Jamiras (approved):

Issue #1009, opened 2022-05-31
Empty shortcode tags cause errors


luchaos:
Empty embeds cause errors because values are expected.

Issue #957, opened 2022-04-18
Unsecure (http://) links being forced into secure (https://) links causing issue.


blendedsea:
A user reported that the Mother 3 v1.3 Translation patch link was no longer working. When I checked it myself I was getting the same issue. When the link is clicked it brings you to > https://mother3.fobby.net/blog/ < which is not correct as it should be > http://mother3.fobby.net/blog/ < 

The unsecure link is what is in the forum post link but when clicked it seems to be forced into a secure link which causes the error. The link was working as normal roughly 2-3 weeks ago when I had previously checked it.

I've created a short video replicating the issue. https://www.youtube.com/watch?v=YH_U5aBf4DE

PR #1015, opened 2022-06-03
Shortcode parser fixes

luchaos:
- Fix empty shortcode tags causing errors -> #1009
- Only fix url protocol for missing scheme and retroachievment URLs -> #957


src/Shortcode.php
```diff
@@ -91,13 +91,18 @@ private function renderLink(ShortcodeInterface $shortcode): string
...
+            return '';
+        }
+
         $scheme = parse_url($href, PHP_URL_SCHEME);
+        $host = parse_url($href, PHP_URL_HOST);
 
         if (empty($scheme)) {
             $href = 'https://' . ltrim($href, '/');
-        } elseif ($scheme === 'http') {
+        } elseif ($scheme === 'http' && str_ends_with($host, 'retroachievements.org')) {
```
luchaos, line 105: Only our own links will be https-only.

src/Shortcode.php
```diff
@@ -177,10 +182,37 @@ private function embedTicket($id): string
...
+    private function autolinkRetroachievementsUrls(string $text): string
+    {
+        // see https://stackoverflow.com/a/2271552/580651:
+        // [...] it's probably safe to assume a semicolon at the end of a URL is meant as sentence punctuation.
+        // The same goes for other sentence-punctuation characters like periods, question marks, quotes, etc..
+        // lookahead: (?<![!,.?;:"\'()-])
+        return (string) preg_replace(
+            '~
+                (?:https?://)?      # Optional scheme. Either http or https.
+                (?:www.)?           # Optional subdomain.
```
luchaos, line 201: www subdomains for retroachievements.org will be dropped

src/Shortcode.php
```diff
@@ -177,10 +182,37 @@ private function embedTicket($id): string
...
+                retroachievements\.        # Host.
+                ([\w!#$%&\'()*+,./:;=?@\[\]-]+
+                (?<![!,.?;:"\'()]))
+                (?!                 # Assert URL is not pre-linked.
+                  [^<>]*>           # Either inside a start tag,
+                  | [^<>]*</a>      # End recognized pre-linked alts.
+                )                   # End negative lookahead assertion.
+            ~ix',
+            '<a href="https://retroachievements.$1">https://retroachievements.$1</a>',
```
Jamiras, line 210: My regex-fu is not very strong, but wouldn't `$1` be the `https` or `www` capture?

luchaos: Same here. Yet, I learned a while ago that `(?` are non-capturing groups; meaning they will be treated as an entity but not exposed as a match. I trusted regex101, stackoverflow and the tests in this case tbh.
The reason for not including `org` in the domain stems from pretty much the same vein of not having a damn clue what regex wants from me. The sentence-punctuation characters clause stopped working as soon as i wanted to include the tld in there for some reason. Chose the easy way out. But it works 🤷

Jamiras (approved):

PR #1017, opened 2022-06-03
Authenticated user in form request only

luchaos:
Remove most occurrences where the username is passed in a hidden form field and used in requests. 

Exceptions are forms where it is an actual input field meant to be provided by the acting user.


public/js/all.js
```diff
@@ -372,22 +372,6 @@ jQuery(document).ready(function onReady($) {
     $('.searchusericon').attr('src', '/UserPic/' + ui.item.id + '.png');
     return false;
   });
-
-  var $resetForm = $('#resetform');
```
luchaos, line 376: Given the age of this codebase, still using jquery (for the better), we somehow happen to use javascript in a quite inflationary fashion. Simple HTML forms are much easier to reason about and easier to respond to in the backend. Let's get rid of that if possible. There's no "ajax fanciness" to be gained here.

public/managehashes.php
```diff
@@ -49,7 +50,7 @@ function UpdateHashDetails(user, hash) {
     $warning.html('Status: updating...');
```
luchaos, line 50: All of this might as well be just HTML forms if I'm not mistaken.

public/request/game/modify.php
```diff
@@ -1,48 +1,53 @@
...
+        echo "OK";
         exit;
-    } else {
-        if ($field == 3) { // Only return status when unlinking hash
-            echo "FAILED!";
-            exit;
-        }
-        header("location: " . getenv('APP_URL') . "/game/$gameID?e=errors_in_modify_game");
+    }
+    header("location: " . getenv('APP_URL') . "/game/$gameID?e=modify_game_ok");
```
luchaos, line 43: Didn't find enough time to untangle this but this request script should _either_ response to ajax requests (with structured json data if possible, not just echo'd strings) _or_ form requests that issue redirects. I'd prefer the latter.
An attempt for simplification will probably follow in a later PR.

public/userInfo.php
```diff
@@ -371,18 +371,18 @@ function resize() {
 
             $newValue = $userIsUntracked ? 0 : 1;
             echo "<tr><td>";
-            echo "<form method='post' action='/request/user/update.php' enctype='multipart/form-data'>";
```
luchaos, line 374: These forms do not post multipart form-data

public/request/game/modify.php
Jamiras: I'm going to be touching this as part of the game audit comments feature. What exactly are you looking for here?

luchaos: https://github.com/RetroAchievements/RAWeb/pull/1017/files#diff-52e0986859da92ea38c57b01ddcee4160047d764052aeb975ffe7b3cc62f0632

Basic forms and location header redirects if possible (_If_ it makes sense to implement it with javascript XHR requests then it should at least respond with JSON. we really shouldn't be using javascript for form requests though imo)

Jamiras (approved):

PR #1022, opened 2022-06-06
Do not open links in new tabs by default

luchaos:
Users may decide for themselves if they want to open links in a new tab.

### DIFF
--- a/lib/database/releases.dist.php
+++ b/lib/database/releases.dist.php
@@ -55,8 +55,8 @@ return [
                 71, // Arduboy
                 72, // WASM-4
             ],
-            'description' => 'Maintained by <a href="https://github.com/libretro/" target="_blank">libretro</a>. Supports a multitude of platforms - including Linux, Mac, Windows, Android.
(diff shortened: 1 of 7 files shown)


Soulsderfuchs:
I don't know about this one, It's a good thing links open in a new tab as is.

luchaos:
> I don't know about this one, It's a good thing links open in a new tab as is.

I agree to some extent. Yet, most links in forums do not open in a new tab. It's inconsistent as is and seen as bad UX to tamper with links/their default behaviour when there is no good reason to.

Jamiras:
I dislike things automatically opening new tabs for me, especially on mobile where I can't see that there's still tabs left open in the background. If I wanted a new tab, I can shift+click or long click.

Google seems to agree. The first three pages that I read suggest that only reasonable excuse to open a new tab/page is to allow media to keep playing. We don't have any media, so there doesn't seem to be any reason to open links in new pages.

Jamiras (approved):

PR #1024, opened 2022-06-06
Introduce achievement action constants

luchaos:
- Do not require/send user in ajax post payloads
- Move scripts from all.js to where they are used
- Minor cleanup

### DIFF
--- a/public/controlpanel.php
+++ b/public/controlpanel.php
@@ -49,7 +49,7 @@ function RenderUserPref($websitePrefs, $userPref, $setIfTrue, $state = null): vo
   function GetAllResettableGamesList() {
     $('#resetgameselector').empty();
 
-    var posting = $.post('/request/user/list-games.php', {u: '<?= $user ?>'});
+    var posting = $.post('/request/user/list-games.php');
     posting.done(function (data) {
       if (data !== 'ERROR3') {
         var htmlToAdd = '<select id=\'resetgameselector\' onchange="ResetFetchAwarded()" >';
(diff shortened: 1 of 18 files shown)


public/controlpanel.php
```diff
@@ -395,22 +395,18 @@ function validateEmail() {
...
+            if ($errorCode == 'badnewpass') {
+                echo "<div id=\"warning\">Info: Errors changing your password, passwords too short!</div>";
+            }
+            if ($errorCode == 'passinequal') {
+                echo "<div id=\"warning\">Info: Errors changing your password, new passwords were not identical!</div>";
+            }
+            if ($errorCode == 'badpassold') {
+                echo "<div id=\"warning\">Info: Errors changing your password, old password was incorrect!</div>";
+            }
+            if ($errorCode == 'changepassok') {
```
Jamiras, line 408: these should probably be `elseif`s - only one will ever match at a time.

src/AchievementAction.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace RA;
+
+abstract class AchievementAction
```
Jamiras, line 5: I think this should be `AchievementUpdateAction` or `Update` should be added to the individual actions. `AchievementAction::DisplayOrder` just doesn't convey the fact that DisplayOrder is being updated.

public/controlpanel.php
luchaos: while that's true I personally dislike `if/elseif/else` chains. i favor clean and easy to ready constructs over minor performance gains.
i think this should be rewritten to a `match` clause where the message is set, and if, display a html element containing the message.
my goal in this case was to untangle the unnecessary indentations for future refactoring.

src/AchievementAction.php
luchaos: Fair point - this was one of the reasons for marking this PR as a draft eventually. The constants should include the verb similar to the other enums, i agree.
Yet, while working on this I found it odd to even specify the kind of update as an enum - these actions may as well be separate request files doing just one kind of update. That was the goal when moving request files in the first place. I'll probably refactor all of those update requests to get rid of the "action" constants where possible and make their usage in scripts obsolete by calling dedicated endpoints explicitly.

luchaos:
Outdated and superceded by upcaoming v2 updates. Closing.

PR #1025, opened 2022-06-06
Refactor preferences & subscriptions

luchaos:
- Add subscriptions created and updated timestamps migration
- Simplify preferences bit shifting
- Use constants for subscriptions


luchaos:
Will be refactored with new settings pages

PR #1033, opened 2022-06-18
Softcore Points Implementation

Tsearo:
This change removes Softcore Points from the Global Leaderboards as well as makes them exclusive to Softcore only earns.

### Database Updates
A new column added to the `UserAccounts` table called `RASoftcorePoints` for tracking them separately. `RAPoints` will now be only for Hardcore Points. No Retro (True) Points will be tracked for Softcore Points.

### Global Leaderboard Changes
The Global Leaderboard will now only track and display Hardcore Achievements, Points and Masteries.
Will now only track users with 250 or more Hardcore points (point requirement to be ranked)
Will now only track users with 1250 or more Retro Points when sorting by Retro Points
image

### User Page/Display Changes
Points will be split out into Hardcore and Softcore and be displayed when appropriate in various sections for users.

Players login section will now display Softcore points in grey:
image

User Page Points will be displayed as follows:
image

Player Cards will show similar changes
image

### Set Requests
As Hardcore points are now effectively halved so will the requirements for Set Requests changing from:
2,500 and 5,000
5,000 to 20,000 +1 per 5k
20,000-200,000 +1 per 10k*
200,000+ +1 per 20k

to 
1,250 and 2,500
2,500 to 10,000 +1 per 2.5k
10,000-100,000 +1 per 5k*
100,000+ +1 per 10k

Up to 10k Softcore points will count towards the users total for requests
*An extra request was also added at 190k/95k to make the tier fill out to an even 200k/100k

### Game Details
High Scores for a game will only count Hardcore Points now

Game comparisons will list softcore unlocks and points separately
image

The dropdown menu will now list achievements once instead of hardcore counting as 2
image

Recent Winners of Achievements will show Hardcore! in blue now
image

### Player History
Total Points Graph - Was capped at 1000 data points staring with the earliest so the total would not be accurate past that. Updated to show all data.
Also Hardcore and Softcore points are separate plots now
image

Score and number of achievements earned will stay the same in Best Days but when clicking on a specific day the red (+X) for hardcore have been removed and now softcore earns will display as grey.
image

### Other Misc Updates
~~Automatic recalc updated from checking players with > 0 Hardcore points to >= 500 points (Min Rank req.) to cut down the time it takes for the global~~

Admin Menu - `Recalc Score Now` will recalculate the users score instead of the admins now


public/historyexamine.php
```diff
@@ -152,7 +161,7 @@ function convertDate() {
             echo "</td>";
 
             echo "<td nowrap>";
-            echo "$achPoints $achPointsNote";
+            echo "$achPointsNote";
```
Jamiras, line 164: The summary (line 182) still shows combined totals.

history list:
image
drilldown summary:
image

lib/database/user.php
```diff
@@ -265,7 +265,8 @@ function getUserPageInfo(&$user, &$libraryOut, $numGames, $numRecentAchievements
     }
     $libraryOut['ContribCount'] = $userInfo['ContribCount'];
     $libraryOut['ContribYield'] = $userInfo['ContribYield'];
-    $libraryOut['TotalPoints'] = $userInfo['RAPoints'];
+    $libraryOut['TotalHardcorePoints'] = $userInfo['RAPoints'];
+    $libraryOut['TotalSoftcorePoints'] = $userInfo['RASoftcorePoints'];
```
Jamiras, line 269: This change will impact the public API:
https://github.com/RetroAchievements/RAWeb/blob/f366b690c3a06565b6b1d1900c644edc0d6b2e0d/public/API/API_GetUserSummary.php#L9-L11

Recommend duplicating `TotalHardcorePoints` into `TotalPoints` there, as is already being done for `Points`:
https://github.com/RetroAchievements/RAWeb/blob/f366b690c3a06565b6b1d1900c644edc0d6b2e0d/public/API/API_GetUserSummary.php#L106

lib/render/leaderboard.php
```diff
@@ -361,14 +363,11 @@ function RenderTopAchieversComponent($user, array $gameTopAchievers, array $game
...
- *            5 - Points
- *            6 - Retro Points
- *            7 - Retro Ratio
- *            8 - Completed Awards
- *            9 - Mastered Awards
+ *            2 - Hardcore Achievements
+ *            3 - Hardcore Points
+ *            4 - Retro Points
+ *            5 - Retro Ratio
+ *            6 - Mastered Awards
```
Jamiras, line 370: Changing these values could break bookmarks (not that I expect many people to bookmark this page). It might be best to just leave them as unsupported:
```
 *            1 - User
 *            2 - Total Achievements (no longer supported)
 *            3 - Softcore Achievements (no longer supported)
 *            4 - Hardcore Achievements
 *            5 - Hardcore Points
 *            6 - Retro Points
 *            7 - Retro Ratio
 *            8 - Completed Awards (no longer supported)
 *            9 - Mastered Awards
```

This is reflected by the fact that the userInfo page has a hardcoded `s=5`:
https://github.com/RetroAchievements/RAWeb/blob/f366b690c3a06565b6b1d1900c644edc0d6b2e0d/public/userInfo.php#L268
which should probably just be removed as it reflects the default sort

lib/render/leaderboard.php
```diff
@@ -534,30 +525,24 @@ function getGlobalRankingData($lbType, $sort, $date, $user, $friendsOf = null, $
...
+              COALESCE(MAX(MasteredAwards), 0) AS MasteredAwards
               FROM
                   (
                       (
                           SELECT aw.User AS User,
-                          COALESCE(SUM(CASE WHEN HardcoreMode = 0 THEN 1 ELSE 0 END), 0) AS SoftcoreCount,
-                          COALESCE(SUM(CASE WHEN HardcoreMode = 1 THEN 1 ELSE 0 END), 0) AS HardcoreCount,
-                          COUNT(*) AS AchievementsObtained,
-                          SUM(ach.Points) AS Points,
+                          Count(ach.ID) AS HardcoreCount,
```
Jamiras, line 537: `COUNT` (all caps) for consistency

public/API/API_GetUserRankAndScore.php
```diff
@@ -18,7 +18,10 @@
 
 $retVal = [];
 
-$retVal['Score'] = getPlayerPoints($user);
+$retVal['Score'] = 0;
+if (getPlayerPoints($user, $userPoints)) {
+    $retVal['Score'] = $userPoints['RAPoints'] + $userPoints['RASoftcorePoints'];
+}
```
Jamiras, line 24: I think `Score` should continue to reflect what's displayed in the login section (i.e. hardcore points). After this PR, there's nowhere else where we display the combined value. You could add the breakdown as additional fields if desired.

public/API/API_GetUserSummary.php
```diff
@@ -103,7 +103,7 @@
 getAccountDetails($user, $userDetails);
 
 $retVal['ID'] = $userDetails['ID'];
-$retVal['Points'] = $userDetails['RAPoints'];
+$retVal['Points'] = $userDetails['RAPoints'] + $userDetails['RASoftcorePoints'];
```
Jamiras, line 106: Same comment as above. Given the other fields that are being copied, this seems like it should just be the hardcore points. The breakdown will already be provided by the `getUserPageInfo` above (see also my comment on that).

public/admin.php
```diff
@@ -87,7 +87,14 @@
                     $message .= "<br>";
                 }
                 recalculatePlayerPoints($validUser);
-                $message .= "- Recalculated Score: " . getPlayerPoints($validUser) . "<br>";
+
+                $hardcorePoints = 0;
+                $softcorePoints = 0;
+                if (getPlayerPoints($user, $userPoints)) {
```
Jamiras, line 93: this is still displaying the points for the admin.

public/dorequest.php
```diff
@@ -213,7 +213,10 @@ function DoRequestError($errorMsg): void
          * Prefer later values, i.e. allow AddEarnedAchievementJSON to overwrite the 'success' key
          */
         $response = array_merge($response, unlockAchievement($user, $achIDToAward, $hardcore));
-        $response['Score'] = getPlayerPoints($user);
+        $response['Score'] = 0;
+        if (getPlayerPoints($user, $userPoints)) {
+            $response['Score'] = $userPoints['RAPoints'] + $userPoints['RASoftcorePoints'];
+        }
```
Jamiras, line 219: This _definitely_ should only be returning the hardcore points. It's used in the client:
image

Which I now realize will likely need supporting changes as well. You'll need to pass `SoftcorePoints` out too. I don't think we need a copy of `Points` in `HardcorePoints`.

Jamiras: Also, whatever you do here will need to be done in `authenticateFromPasswordOrAppToken` so the initial values are known. This just updates the values after an achievement is unlocked.

public/gameInfo.php
```diff
@@ -835,18 +818,31 @@ function submitSetRequest(user, gameID) {
...
+                }
+
+                if ($user !== null && $numAchievements > 0) {
                     if ($numEarnedHardcore > 0) {
-                        echo "<a href='/user/$user'>$user</a> has won <b>$numEarnedHardcore</b> HARDCORE achievements";
-                        if ($totalEarnedHardcore > 0) {
-                            echo ", worth a further <b>$totalEarnedHardcore</b> points";
+                        echo "You've earned <b>$numEarnedHardcore</b> HARDCORE achievements, worth <b>$totalEarnedHardcore</b> <span class='TrueRatio'>($totalEarnedTrueRatio)</span> points.<br>";
+                        if ($numEarnedCasual > 0) { // Some Hardcore earns
+                            echo "You've also earned <b> $numEarnedCasual </b> SOFTCORE achievements worth <b>$totalEarnedCasual</b> points.<br>";
```
Jamiras, line 842: Recommend `You have earned`... Contractions are generally harder for non-native speakers, and make the site feel just a little less professional.

public/request/user/recalculate-score.php
```diff
@@ -5,13 +5,19 @@
...
+if (!authenticateFromCookie($actingUser, $permissions, $actingUserDetails, Permissions::Registered)) {
     header("Location: " . getenv('APP_URL') . "?e=badcredentials");
     exit;
 }
 
-if (recalculatePlayerPoints($user)) {
-    header("Location: " . getenv('APP_URL') . "/controlpanel.php?e=recalc_ok");
+if (recalculatePlayerPoints($targetUser)) {
+    if ($targetUser !== $actingUser && $permissions >= Permissions::Admin) {
+        header("Location: " . getenv('APP_URL') . "/user/$targetUser?e=recalc_ok");
```
Jamiras, line 17: Recommend `e=ok`. `e=recalc_ok` just looks weird when you're redirected to the user's profile:
> Score recalculated! Your new score is shown at the top-right next to your avatar.

public/userInfo.php
```diff
@@ -108,14 +108,17 @@
 
 $pageTitle = "$userPage";
 
-$userPagePoints = getPlayerPoints($userPage);
+$userPagePoints = 0;
+if (getPlayerPoints($userPage, $userPoints)) {
+    $userPagePoints = $userPoints['RAPoints'] + $userPoints['RASoftcorePoints'];
+}
```
Jamiras, line 114: `$userPagePoints` doesn't appear to be used anywhere. I believe you can eliminate this call.

public/API/API_GetUserRankAndScore.php
Jamiras: Or keep `Score` as hardcore points and add a `SoftcoreScore` similar to [my suggestion](https://github.com/RetroAchievements/RAWeb/pull/1033/files#r900964715) for `dorequest`.

cronjobs/cron_1m.php
```diff
@@ -21,7 +21,7 @@ function GetNextHighestGameID($givenID)
 function GetNextHighestUserID($givenID)
 {
     $query = "SELECT MIN(ID) AS NextID FROM UserAccounts
-				  WHERE ID > $givenID AND RAPoints > 0";
+				  WHERE ID > $givenID AND RAPoints >= " . MIN_POINTS;
```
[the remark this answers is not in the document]
Tsearo: Yeah was more of a tradeoff to increase the overall accuracy of the overall global leaderboard by increasing the frequency of the more relevant positions with the exception being that unranked users near but below 500 points may miss out on recalcs from rescores that would bump them up.

Could compromise a bit and change it to `MIN_POINTS / 2` if you think that's a better solution

I completely missed that RASoftcorePoints were getting missed out so I added a check for RAPoints OR RASoftcorePoints >= MIN_POINTS

lib/database/achievement-creator.php
```diff
@@ -91,8 +92,8 @@ function getOwnAchievementsObtained(string $user): bool|array|null
     sanitize_sql_inputs($user);
 
     $query = "SELECT 
-              SUM(CASE WHEN aw.HardcoreMode = 0 THEN 1 ELSE 0 END) AS SoftcoreCount,
-              SUM(CASE WHEN aw.HardcoreMode = 1 THEN 1 ELSE 0 END) AS HardcoreCount
+              SUM(CASE WHEN aw.HardcoreMode = " . AchievementAwardType::Softcore . " THEN 1 ELSE 0 END) AS SoftcoreCount,
+              SUM(CASE WHEN aw.HardcoreMode = " . AchievementAwardType::Hardcore . " THEN 1 ELSE 0 END) AS HardcoreCount
```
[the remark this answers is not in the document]
Tsearo: I was a bit hesitant on whether or not it was needed as you said but figured maybe one day we'll add something like midcore or ultra hardcore haha. I updated the name to AwardedHardcoreMode for now but If it's preferred to remove completely I can

lib/database/friend.php
```diff
@@ -158,6 +160,7 @@ function getAllFriendsProgress($user, $gameID, &$friendScoresOut): int
                     FROM Achievements AS ach
                     WHERE ach.GameID = '$gameID' AND ach.Flags = 3
                 ) AS Inner1 ON Inner1.ID = aw.AchievementID
+                WHERE aw.HardcoreMode = " . AchievementAwardType::Softcore . "
```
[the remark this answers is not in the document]
Tsearo: This is specific to the Friends Ranking on the user page and not the Global Ranking filtered by friends. For some reason I felt that it could include all unlocks as more of a tracking method for friends but it's probably best to change this to hardcore to prevent confusion and inconsistency. Still pending a fix here I'll look further into it

lib/database/player-game.php
```diff
@@ -197,7 +199,7 @@ function getUsersGameList($user, &$dataOut): int
         LEFT JOIN GameData AS gd ON gd.ID = ach.GameID
         LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
         LEFT JOIN ( SELECT ach1.GameID AS GameIDInner, ach1.ID, COUNT(ach1.ID) AS TotalAch FROM Achievements AS ach1 GROUP BY GameID ) AS gt ON gt.GameIDInner = gd.ID
-        WHERE aw.User = '$user'
+        WHERE aw.User = '$user' AND aw.HardcoreMode = " . AchievementAwardType::Softcore . "
```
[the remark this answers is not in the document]
Tsearo: Nice catch!

lib/database/player-history.php
```diff
@@ -115,21 +121,30 @@ function getAwardedList($user, $listOffset, $maxToFetch, $dateFrom = null, $date
...
+                FROM Awarded AS aw
+                LEFT JOIN Achievements AS ach ON ach.ID = aw.AchievementID
+                LEFT JOIN GameData AS gd ON gd.ID = ach.GameID
+                WHERE aw.user = '$user' 
+                AND ach.Flags = " . AchievementType::OfficialCore . "
+
+                $dateCondition
+                GROUP BY YEAR(aw.Date), MONTH(aw.Date), DAY(aw.Date)
+                ORDER BY aw.Date ASC
+                $limitCondition";
```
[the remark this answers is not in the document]
Tsearo: Yeah the intention was to just use the overall total here but after you mentioned the graphing package I looked into it and saw it wasn't too hard to make some changes to split them out which is much more ideal 👍

lib/render/leaderboard.php
Tsearo: Hmm not too concerned about the bookmarks issue but I changed the values back and noted which ones are unsupported like you mentioned. I think the userInfo link is still relevant since it includes `$rankOffset` to take you to the correct page for the user so I'll keep that for now.

public/userInfo.php
Tsearo: Hmm yeah you are right this one wasn't being used at all

public/dorequest.php
Tsearo: Thanks for the insight on this part as well as the API files as well. These were ones I had some issues with and I may need some more assistance with this as I'm seeing issues in RALib as I test earning achievements after I earn one the Total Points displayed go to 0 until I reload RALib so I likely still have something wrong between this or the API. Not sure if there is a good way to get more details from RALib (do I need to build the debug version?) or if this is related to the changes that you mention will need to be done on that end

cronjobs/cron_1m.php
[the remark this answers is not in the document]
Jamiras: The problem would still exist as long as you excluded any point value from the query. Even `>= 2" would still miss anyone who had a single point if that single-point achievement were rescored.

Ideally, recalcs wouldn't be necessary, but the codebase isn't perfect.

It's probably a good idea to put some threshold in place with the immediate release so active players do get recalculated sooner (maybe even raise the threshold higher), but long term, I think we need to go back to recalculating everyone with points.

At 500 points, there's roughly 42000 accounts to process (about 10 days at 3 accounts per minute)
At 1000 points, there's about 28000 accounts to process (about 6 1/2 days)
At 2000 points, there's about 18000 accounts (about 4 days)
At 5000 points, there's about 9000 accounts (about 2 days)
At 10000 points, there's about 5000 accounts (about 27 hours)

Maybe we could commit with the limit at 10000 points, let it run for a day and a half, then change the limit to `MIN_POINTS`, let it run for a couple weeks, then change the limit back to `>0`?

Or, if we can get @luchaos to approve, we could use the stored procedure I sent you to do the whole database in a couple hours, but that might put undue load on the server.  Side note: you should probably include the stored procedure script in the PR (database subfolder) for other devs to mass-update their data.

public/admin.php
```diff
@@ -87,7 +87,14 @@
                     $message .= "<br>";
                 }
                 recalculatePlayerPoints($validUser);
-                $message .= "- Recalculated Score: " . getPlayerPoints($validUser) . "<br>";
+
+                $hardcorePoints = 0;
+                $softcorePoints = 0;
+                if (getPlayerPoints($awardAchievementUser, $userPoints)) {
```
Jamiras, line 93: I think `$awardAchievementUser` is a CSV (see preg_split on line 68). It's probably best to just use `$validUser` as that's whose points were recalculated on line 89.

public/dorequest.php
Jamiras: It's a data typing issue.
Prod:
```
{
 "Success":false,"AchievementsRemaining":12,
 "Error":"User already has this achievement unlocked in hardcore mode.",
 "Score":152872,"AchievementID":3159
}
```
Test:
```
{
 "Success":false,"AchievementsRemaining":18,
 "Error":"User already has this achievement unlocked in hardcore mode.",
 "Score":"75650","SoftcoreScore":"1100","AchievementID":3159
}
```
Make sure to cast `Score` and `SoftcoreScore` to integers.

Tsearo: That was it thanks again! I believe the overall issue was when I changed `getPlayerPoints` I didn't preserve the typing when changing it to instead pass `$dataOut`

cronjobs/cron_1m.php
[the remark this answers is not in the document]
Tsearo: Sorry I see what you mean about asking if this is a permanent change now along with your other points. I'll change the limit to > 0 for either hardcore or softcore with the assumption that we will be able to run the stored procedure which I've also added to the database folder. If that turns out to not be a valid option we can figure out the best way to handle it

lib/database/user.php
Tsearo: I believe I've taken care of this and the other API call now but I'll let you verify that I did it correctly as I'm still not 100% sure.

public/API/API_GetUserSummary.php
```diff
@@ -7,7 +7,8 @@
  *    a : number of recent achievements to return (default: 10)
  *
  *  string     ID                      unique identifier of the user
- *  string     TotalPoints             number of points the user has
+ *  string     TotalPoints             number of hardcore points the user has
+ *  string     TotalSoftcorePoints     number of hardcore points the user has
```
Jamiras, line 11: "number of _softcore_ points the user has"

lib/database/user.php
Jamiras: Looks good.

API_GetUserSummary:
```
   "TotalPoints":"2900",
   "TotalSoftcorePoints":"25",
   "TotalTruePoints":"4268",
...
   "Points":"2900",
   "SoftcorePoints":"25",
```
API_GetUserRankAndScore:
```
    "Score":2900,
    "SoftcoreScore":25
```
User page:
image

Jamiras (approved):
Approving the code changes. We still need to decide how we're going to recalc the points - stored procedure, ramp up, or slowly over a month.

lib/database/player-history.php
```diff
@@ -1,36 +1,41 @@
...
 
-    $query .= "LIMIT $listOffset, $maxDays ";
+    $query = "SELECT YEAR(aw.Date) AS Year, MONTH(aw.Date) AS Month, DAY(aw.Date) AS Day, COUNT(*) AS NumAwarded, SUM(Points) AS TotalPointsEarned 
+                FROM Awarded AS aw 
+                LEFT JOIN Achievements AS ach ON ach.ID = aw.AchievementID
+                WHERE User='$user' 
+                AND aw.HArdcoremode = " . AchievementAwardType::Softcore . "
+                GROUP BY YEAR(aw.Date), MONTH(aw.Date), DAY(aw.Date)
+                $orderCond
+                LIMIT $listOffset, $maxDays";
```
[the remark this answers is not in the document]
luchaos: Fixed. Note that while this works on Windows and macOS just fine as they are case insensititve, the application runs on Linux and thus is case sensitive. The case within queries has to match at all times

cronjobs/cron_1m.php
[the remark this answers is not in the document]
luchaos: Sgtm. I'll give it a try locally to see what kind of load we're expecting.

luchaos:
Running the stored procedure locally would take about an hour on my not so beefy machine without impairing any other database interactions while browsing the site.
The server should take it easily.

PR #1035, opened 2022-06-18
fix mature content user preference not saving

### LINKED ISSUE #1034: Mature Content Warning can't be deactivated
Since a while there is the mature content warning feature that shows an age check before loading game entries that are linked with the [Mature hub](https://retroachievements.org/game/8323).

I recently discovered that it's not possible to disable this warning in the settings of registered users as it should be. When disabling this particular setting a green check symbol appears which indicates that it was saved but it actually doesn't. Opening the settings again shows that it's still activated and loading a qualified game entry also throws the warning.
image

I tried to logout and login again and also cleared the cache but the issue persists.

Jamiras:
fixes #1034 

#1008 [renamed the input fields](https://github.com/RetroAchievements/RAWeb/pull/1008/files#diff-c217b8190030ab7ae176fb7a26083812018fb555b076716a211eb5b041eaec70L183) for the user preference checkboxes, but missed one.

### DIFF
--- a/public/controlpanel.php
+++ b/public/controlpanel.php
@@ -185,7 +185,7 @@ function RenderUserPref($websitePrefs, $userPref, $setIfTrue, $state = null): vo
     }
 
     // 7 is set if unchecked
-    var checkbox = document.getElementById('UserPref7');
+    var checkbox = document.getElementById('UserPreference7');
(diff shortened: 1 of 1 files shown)


luchaos (approved):
Oy, my bad. Thanks!

PR #1039, opened 2022-06-25
convert newlines to <br>s for comment widget

Jamiras:
Fixes a display issue when newlines are used in comments. I noticed it in ticket comments, but it applies to all comment controls.

Before:
image

After:
image

Interestingly enough, I can only find newlines in the 2020 database dump in comments older than Feb 2014. I'm not sure if we were previously stripping them and somehow aren't any more...


luchaos:
Newlines not translating properly doesn't sound right, having a quick look. I'll have to investigate this issue in-depth

Jamiras:
> I can only find newlines in the 2020 database dump in comments older than Feb 2014. I'm not sure if we were previously stripping them and somehow aren't any more...

I think this is the change that matters: 

https://github.com/RetroAchievements/RAWeb/pull/933/files#diff-7ed11b656194a2c674ee61652f95e4cc6b65cfc4c8200ba0b981c5e0624d5a25L23

Prior to that, newlines would have been stripped from comments on submission.

luchaos (approved):

PR #1042, opened 2022-06-30
Replace Friends terminology with Followed Users

Jamiras:
Modifies the Friends list page to show pending friend requests and blocked users, with commands to manage them.

image

Users showing up in the "Friends" list have mutually friended each other. Users in the "Seeking Friendship" group have attempted to friend you, and are waiting for you to confirm. Users in the "Pending" list are requests you've initiated and are waiting for them to confirm. Users in the Blocked list have been blocked by you.

**Additionally, all references to "Friends" in the site now only return users who are mutually friends** (previously, it would return users from the Friends list and the Pending list). This gives the second user a voice in who can monitor their progress.

Declining a friendship request removes the request so they no longer appear as pending to the originating user.

Emails are still only sent when receiving a friendship request or a friendship confirmation. No site messages are sent (consistent with previous behavior, but misleading given that there's a checkbox on the settings page for it).

These changes were made without any additional database changes. There's still two rows in the friends table (one for each direction).


luchaos:
Interesting changes - very nice.

Some thoughts for context: when thinking about how to restructure user connections in V2 i initially ended up dropping this completely - instead of continuing the "mutual friendship" approach I would've translated it into a "passive" follower system.

Thinking about this today I guess it's fine to finish the idea of how the friends system was probably meant to work in the first place and introduce your changes. In V2 we might as well have both - a friends system (ala Steam, linkedin, ...) and a follower system (ala Twitter, Instagram, ...).

My main concern is maintainability and privacy issues (that are both currently not fully addressed and to be rethought in the future).

A passive follower system makes it easier to maintain a blocklist. Following a public profile does not require mutual agreement but blocks can affect it. Whereas a friend system requires both sides to agree.

In both cases blocks should not expose privacy settings with what action is available to the other party.

Listing followers and followins users is easier than determining friendship.

A follower system does not require to send out email notifications with approval actions included. On the other hand, the on-site notifications should have means to display such changes (included in v2 by the framework as a first-party feature).

Jamiras:
Maybe we should ask the community how they are using the feature. The existing behavior is very much a "followers" system (not a "friends" system). But you can't see who's following you, and blocking a user only prevents them from sending you messages. You still appear in _their_ "friends" lists.

I could fairly easily rework this to merge the "Friends/Pending" lists into "Following" and rename the "Seeking Friendship" to "Followers".

luchaos:
Asking the community sounds like a good idea.

Personally, I'd very much welcome the transition to the followers system - the other changes you prepared are still valid and useful. If i understand correctly we could even drop/delete a few things related to the "mutual" nature of the existing system.

Jamiras:
The community poll overwhelming indicates that the feature should just be for following. Of the thirty votes (so far), 40% used it to see other users' activity, 30% of users didn't use the feature, 25% used it to compare themselves against other users, and only 5% used it as a leaderboard amongst friends.

Most users are either not using the feature, or only using it in a "following" fashion. Those who were using it as an actual friend system can still mutually follow each other.

Here's an updated image for a "following" implementation:
image

The wording has also been updated everywhere I could find it:
image
image
image
image

pinguupinguu:
I feel like "Pending" could be "Outgoing" requests

luchaos (approved):
ℹ️ The table in v2 will be called `user_relations` to hold follow and block information

PR #1054, opened 2022-07-20
Allow editing achievement title, description, points and badges on site

Tsearo:
Allows dev's to make edits to achievement Titles, Descriptions, Points and Badges through the site to allow for faster/easier tweaks for games not owned or on mobile as it doesn't require loading the game (bulk updates will likely still be faster through emulators).

Also puts the video URL examples for embed videos in a collapsed link to make up for the extra space taken by the edit boxes.

Normal example for developers
image

Jr's will not be able to make edits to achievements in core which matches their current restrictions
image

Jr's will be able to edit these fields if they are the sole author and the achievement is in unofficial but will still not be able to upload an embed video link
image

Badges have to be uploaded separately from the text edits but can also be done under the above criteria
image

Embed expanded
image

Title and Description:
Text field only allows inputting 64 and 256 characters per their respective limits, Unicode can bypass this but will be calculated and prevent users from attempting to input anything longer than the database can store
Ex. These unicode take up 80 characters after converted to ISO-8859-1 charset which is used by the database
image

Points will also not let you submit values not within the valid tiers [0,1,2,3,4,5,10,25,50,100] which are stored in a constant in case of updates
image

Minor extra - fixed escaping html for Game Info text
Before:
image

After:
image


televandalist:
Tsearo/Searo more like hero

public/achievementInfo.php
```diff
@@ -85,13 +91,71 @@
...
+        if ( titleLength > maxTitle) {
+                showStatusFailure('Error: Title too long');
+                return;
+        } else if ( descLength > maxDescription) {
+                showStatusFailure('Error: Description too long');
+                return;
+        } else if ( points == '' || !validPoints.includes(Number(points))) {
+                showStatusFailure('Error: Invalid Points');
+                return;
+        }
```
meleu, line 115: maybe being picky, but trying to improve readability/clarity...

those `else`s are not necessary when the previous `if`s are `return`ing.

Doing like in the example would have the same result and a reduced cognitive load (fancy words for "less code to read and understand"):
```php
if(condition1) {
  doSomething1();
  return;
}
if(condition2) {
  doSomething2();
  return;
} 
if(condition3) {
  doSomething3();
  return;
}
```

Jamiras:
I'm able to add apostrophes to title and description and they save, but they don't show up correctly on load.
image

public/achievementInfo.php
```diff
@@ -50,6 +51,11 @@
 
 getAchievementUnlocksData($achievementID, $numWinners, $numPossibleWinners, $numRecentWinners, $winnerInfo, $user, 0, 50);
 
+// Determine if the logged in user is the sole author of the set
+if (isset($user)) {
```
Jamiras, line 55: could check for `$permissions > Jr` before doing this lookup

public/achievementInfo.php
```diff
@@ -85,13 +91,71 @@
 
 <body>
 <?php RenderHeader($userDetails); ?>
-<?php if ($permissions >= Permissions::Developer): ?>
+<?php if (($permissions >= Permissions::Developer || ($isSoleAuthor && $permissions >= Permissions::JuniorDeveloper && $achFlags === AchievementType::Unofficial))): ?>
```
Jamiras, line 94: extra level of parenthesis is unnecessary

public/achievementInfo.php
```diff
@@ -85,13 +91,71 @@
...
+    <?php if ($permissions >= Permissions::Developer): ?>
       function PostEmbedUpdate() {
         var url = $('#embedurlinput').val();
         url = replaceAll('http', '_http_', url);
 
         showStatusMessage('Updating...');
+
+        if ( url == '') {
+                showStatusFailure('Error: Blank url');
+                return;
```
Jamiras, line 156: looks like this code should be using two space indents. probably the code added above too.

public/achievementInfo.php
```diff
@@ -237,16 +303,30 @@ function updateAchievementTypeFlag(typeFlag) {
...
-                RenderStatusWidget();
-                echo "<li>Set embedded video URL:</li>";
+            if ($permissions >= Permissions::Developer || ($isSoleAuthor && $permissions >= Permissions::JuniorDeveloper && $achFlags === AchievementType::Unofficial)) {
+                echo "<div>Update achievement details:</div>";
                 echo "<table><tbody>";
-                echo "<input type='hidden' name='a' value='$achievementID' />";
-                echo "<input type='hidden' name='f' value='2' />";
-                echo "<tr><td>Embed:</td><td style='width:100%'><input id='embedurlinput' type='text' name='v' value='$embedVidURL' style='width:100%;'/></td></tr>";
+                echo "<tr><td>Title:</td><td style='width:100%'><input id='titleinput' type='text' name='t' value='$achievementTitle' style='width:100%' maxlength='64';></td></tr>";
+                echo "<tr><td>Description:</td><td style='width:100%'><input id='descriptioninput' type='text' name='d' value='$desc' style='width:100%' maxlength='256';></td></tr>";
```
Jamiras, line 312: should use constants for 64 and 256

public/achievementInfo.php
```diff
@@ -237,16 +303,30 @@ function updateAchievementTypeFlag(typeFlag) {
...
+                echo "<tr><td>Points:</td><td style='width:100%'><input id='pointsinput' type='number' name='p' value='$achPoints' style='width:100%' min='0' max='100';'></td></tr>";
```
Jamiras, line 313: maybe use constants for 0 and 100?

public/request/achievement/update.php
```diff
@@ -18,6 +18,14 @@
 $field = requestInputPost('f', null, 'integer');
 $value = requestInputPost('v');
 
+$achievementTitle = requestInputPost('t');
+$desc = requestInputPost('d');
+$badgeName = requestInputPost('b');
+$achPoints = requestInputPost('p', null, 'integer');
+$achMem = requestInputPost('m');
+$gameID = requestInputPost('g', null, 'integer');
+$achFlags = requestInputPost('l');
```
Jamiras, line 27: as these are all specific to "f=4", they should be in the if block for that (similar to 'g' for "f=1")

public/request/achievement/update.php
```diff
@@ -93,6 +103,31 @@
...
+                idInOut: $achievementId,
+                badge: $badgeName,
+                errorOut: $errorOut
+        )) {
+            echo json_encode(['success' => true, 'message' => 'OK']);
+            exit;
+        } else {
+            http_response_code(400);
+            echo json_encode(['success' => false, 'error' => 'Bad request']);
+            break;
```
Jamiras, line 129: this `break` will cause two `success=false` outputs.

public/achievementInfo.php
Tsearo: Do you feel it would be better to make new constants for min/max points or to access the first/last element in the VALID_POINTS const using something like this `VALID_POINTS[array_key_first(VALID_POINTS)];` and `VALID_POINTS[array_key_last(VALID_POINTS)];`?

Jamiras: I believe php has functions for returning the min/max value from an array.

public/request/achievement/update.php
Tsearo: I tried to split this out the best I could, doing so required splitting the jrDevAllowed check a bit more which I'm not fully happy with how it ended up as well as duplicate calls on gameID and achFlags in case 4 for Jrs but phpstan complains if I put those in if statements

Jamiras (approved):

public/achievementInfo.php
```diff
@@ -85,13 +91,73 @@
 
 <body>
 <?php RenderHeader($userDetails); ?>
-<?php if ($permissions >= Permissions::Developer): ?>
+<?php if (($permissions >= Permissions::Developer || $isSoleAuthor && $permissions >= Permissions::JuniorDeveloper && $achFlags === AchievementType::Unofficial)): ?>
```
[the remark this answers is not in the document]
luchaos: Reverted and removed outer parantheses.

luchaos (approved):
Changed a few things.

I did not like that the frontend can dictate which fields are update. That sounds like and invitation for trouble. 
Reduced it to just the three fields in question and moved it to a separate request file as that's what will happen in v2 (already done, branch will be up soon).

Same for constants - I've already removed that file in v2 and moved the constants to Enum-like classes. Left a comment there to prevent any further additions until v2 is merged.

The validation constants for length were not actually required - the browser recognises the maxlength attribute. If a longer string would be posted, which requires fiddling with the payload, it would just be cut off on the database side. No harm there. V2 will add proper validation on the server side.

luchaos:
Screenshot 2022-08-07 at 23 22 34

Changed points input to a select as there is a defined list of valid values.

Tsearo:
I think that makes sense about the frontend parts.

As for the size checks I did those since using unicode will change the length and if it is too long it throws the error seen below which is why I added in specific checks and error messages to let the user know why.
Ex.
image

And nice I didn't think to use a select for the points

luchaos:
Yep :) I assumed that's why the Blob checks are in place going by your screenshots.
I think it will be enough to validate on the server (which will do properly it with multibyte support).
We don't validate in the frontend - not even the character count in comment areas is correct when using multibyte characters.
I would've waited until we have a proper frontend validation library in place to be used for _all_ inputs. It barely makes a difference atm and complicates things.
I'll add it back.

luchaos:
Added the frontend validation. Additionally, the backend now validates the points value globally.

lib/database/achievement.php
```diff
@@ -248,6 +249,11 @@ function UploadNewAchievement(
         return false;
     }
 
+    if (!AchievementPoints::isValid((int) $points)) {
+        $errorOut = "Invalid points value (" . $points . ").";
+        return false;
+    }
```
luchaos, line 255: @Jamiras haven't tested yet but i hope this play nicely with the toolkit?
It validates the points value globally with that addition.

luchaos: See https://github.com/RetroAchievements/RAWeb/pull/1054/files#diff-2553fd75e71c70fd2914740c5ea8c486a7c21f93496f8b785c94102c36537ea6R7

Tsearo: Just tested it locally in RALib and it seems to give this error when trying to upload an invalid point which seems pretty nice to me
image

luchaos: Awesome, thanks for testing!

PR #1070, opened 2022-08-07
Awards counters for profile pages

EmoonX:
Adds simple counters for showing how many awards a player has gotten in a given category.

award-counter

As for game awards, separate counters are used both for completed and mastered games (as expected, if a player got none of a given type, then no counter is shown). There are also tooltips on hover in case user is not aware of what the numbers actually mean.

award-tooltip

I've used simple emoji for the icons. Can be changed to better images later if the community sees fit.

This should be particularly useful for promptly knowing how many games a player has mastered, especially for e.g those who have a huge roster with 100+ badges.

---

Hey, this is my first PR! Thank you very much @Tsearo for showing me the way around PHP/DB struggles.


luchaos (approved):
Nice job! I might refactor it a little bit down the line but it's working nicely as is 👍

PR #1071, opened 2022-08-08
Cleanup on `Latest Masters` / `High Scores` tables

EmoonX:
General CSS cleanup on said tables. Better padding, centered text, wording, etc.

high-scores

For the Latest Masters one, instead of generic numbers from 1 to 10 we have now an actual rank representing given player was the `n`th to master. Also reduced date/time's font a bit and got rid of seconds (excessive info and no need for so much precision).

latest-masters

Cells have fixed size now; which means that, when transitioning from one table to the other, the content (hopefully) won't dance around.


Jamiras:
"Mastered on" is wrapping for me, which makes the header two lines.

Here's a side-by-side of my before and after:
image

I would also recommend removing the italics from the times; it's not something we do anywhere else that I can think of.

lib/render/leaderboard.php
```diff
@@ -267,15 +267,17 @@ function RenderTopAchieversComponent($user, array $gameTopAchievers, array $game
     // Latest Masters Tab
     echo "<div id='latestmasters' class='tabcontentscores' style=\"display: " . ($numLatestMasters >= $masteryThreshold ? "block" : "none") . "\">";
     echo "<table class='smalltable'><tbody>";
-    echo "<tr><th>Pos</th><th colspan='2' style='max-width:30%'>User</th><th>Mastery Date</th></tr>";
+    echo "<tr><th>#</th><th>User</th><th>Mastered on</th></tr>";
```
luchaos, line 270: I think just "Mastered" would be enough. I planned on reducing most of those table headers to simpler terms a while ago.

public/css/styles.css
```diff
@@ -1608,6 +1608,47 @@ svg > g > g:last-child { pointer-events: none }
   border: 1px solid rgb(44, 151, 250);
 }
 
+.tabcontentscores th {
+  text-align: center;
```
luchaos, line 1612: I'd like to keep table headers left aligned and normal font-weight for now to match the rest and do such changes site-wide eventually with the new stack.
Can stay as is as well though, as you wish.

luchaos (approved):

PR #1072, opened 2022-08-08
Fix achievement change author

luchaos:
I changed one too many things there. The achievement author was showing as change author instead of the acting user.


Tsearo (approved):
Looks like this fixes that issue 👍

PR #1077, opened 2022-08-09
don't require mutual following to display users in 'compare to others' list

Jamiras:
remnant of the original attempt to make friends require mutual acknowledgment


Tsearo (approved):
Looks good

PR #1079, opened 2022-08-09
update max game progress to 100%

Jamiras:
Game progress no longer reports values over 100% (mastering a game is no longer 200%)

Reported progress is now the maximum of (hardcore progress/softcore progress), and if the game has 100% hardcore completion, it's reported as Mastered:
image

If the user has both hardcore and softcore unlocks, the bar gains a tooltip indicating the hardcore progress. The reported progress will be for the softcore progress.

The change affects the recently played games on the user's profile page, and the bar shown on individual game pages (showing tooltip):
image

### DIFF
--- a/lib/render/game.php
+++ b/lib/render/game.php
@@ -259,3 +259,43 @@ function RenderRecentGamePlayers($recentPlayerData): void
     echo "</tbody></table>";
     echo "</div>";
 }
+
+function RenderGameProgress(int $numAchievements, int $numEarnedCasual, int $numEarnedHardcore)
+{
+    $pctComplete = 0;
+    $pctHardcore = 0;
+    $pctHardcoreProportion = 0;
+    $title = '';
+
+    if ($numAchievements) {
+        $pctAwardedCasual = ($numEarnedCasual + $numEarnedHardcore) / $numAchievements;
+        $pctAwardedHardcore = $numEarnedHardcore / $numAchievements;
+        $pctAwardedHardcoreProportion = 0;
+        if ($numEarnedHardcore > 0) {
+            $pctAwardedHardcoreProportion = $numEarnedHardcore / ($numEarnedHardcore + $numEarnedCasual);
+        }
+
+        $pctComplete = sprintf("%01.0f", $pctAwardedCasual * 100.0);
+        $pctHardcore = sprintf("%01.0f", $pctAwardedHardcore * 100.0);
+        $pctHardcoreProportion = sprintf("%01.0f", $pctAwardedHardcoreProportion * 100.0);
+
+        if ($numEarnedCasual && $numEarnedHardcore) {
+            $title = " title='$pctHardcore% hardcore'";
+        }
+    }
+
+    echo "<div class='progressbar'>";
+    echo "<div class='completion' style='width:$pctComplete%'$title>";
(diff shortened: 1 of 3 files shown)


Soulsderfuchs:
Hardcore Progress should probably be the one being the text and the softcore being hidden by the toolttip as to not makes this less confusing that it is, Does this also affect the Completion Progress widget on users profiles?

Jamiras:
> Hardcore Progress should probably be the one being the text and the softcore being hidden by the toolttip as to not makes this more confusing that it is

I disagree. As the tooltip only exists when a user has both hardcore and softcore unlocks, we can assume the only reason the player has softcore unlocks is because they're trying to 100% the set (you still get a completion badge for unlocking everything in softcore). As such, the more important measure is how close they are to that target. The breakdown is less important and can be ushered into the tooltip.

> Does this also affect the Completion Progress widget on users profiles?

No. That still reports the number of unlocks for the game, regardless of whether or not they're hardcore unlocks. The bar still shows yellow/blue to indicate the ratio, but the percentage isn't reported anywhere.

Jamiras:
> What does an example of a 100% softcore only completion look like with no hardcore earns?

100% complete. There's still room for improvement, even though I wouldn't expect someone with no hardcore unlocks to play through the whole set again in hardcore.

lib/render/game.php
```diff
@@ -259,3 +259,43 @@ function RenderRecentGamePlayers($recentPlayerData): void
...
+    echo "<div class='progressbar'>";
+    echo "<div class='completion' style='width:$pctComplete%'$title>";
+    echo "<div class='completionhardcore' style='width:$pctHardcoreProportion%'>";
+    echo "&nbsp;";
+    echo "</div>"; // completionhardcore
+    echo "</div>"; // completion
+
+    if ($pctHardcore >= 100.0) {
+        echo "Mastered<br>";
+    } else {
```
Tsearo, line 296: Everything looks good to me, if you wanted to add Completed as an alternative display for that scenario of only softcore progress at 100% I believe you can just add something along the lines of this here
```
else if ($pctHardcore == 0.0 && $pctComplete >= 100.0) {
    echo "Completed<br>";
}
```

Tsearo (approved):

lib/render/game.php
```diff
@@ -259,3 +259,43 @@ function RenderRecentGamePlayers($recentPlayerData): void
...
+
+    if ($numAchievements) {
+        $pctAwardedCasual = ($numEarnedCasual + $numEarnedHardcore) / $numAchievements;
+        $pctAwardedHardcore = $numEarnedHardcore / $numAchievements;
+        $pctAwardedHardcoreProportion = 0;
+        if ($numEarnedHardcore > 0) {
+            $pctAwardedHardcoreProportion = $numEarnedHardcore / ($numEarnedHardcore + $numEarnedCasual);
+        }
+
+        $pctComplete = sprintf("%01.0f", $pctAwardedCasual * 100.0);
```
luchaos, line 278: I think the fractions can be omitted as it's dynamically typed. Not that important though.

luchaos (approved):

lib/render/game.php
Jamiras: As mentioned above, I think the "100% complete" is fine. If it nags the user into playing hardcore to change the text to Mastered, so be it.

PR #1082, opened 2022-08-11
Fix hub matching for values containing ampersands

Jamiras:
Relocating the `sanitize_outputs` in https://github.com/RetroAchievements/RAWeb/commit/76e5ef9066e1de281d0e487bca22e6c3cd529951#diff-66930c10e82ee4f8d7cf1fd91cb8a68f42f751f9e653b776fbca42c44603cf56R625-R630 caused the value to become escaped, and then it doesn't match the hub name, resulting in
image

Since the previous location of `sanitize_outputs` was before they were being initialized, it doesn't seem like it's necessary. I've removed it, and things appear to be working. The `attribute_escape` is handling the sanitization when it becomes necessary.

### DIFF
--- a/lib/render/game.php
+++ b/lib/render/game.php
@@ -158,9 +158,10 @@ function RenderGameAlts($gameAlts, $headerText = null): void
 function RenderMetadataTableRow($label, $gameDataValue, $gameHubs = null, $altLabels = []): void
 {
     $gameDataValues = !empty($gameDataValue) ? array_map('trim', explode(',', $gameDataValue)) : [];
+    $unmergedKeys = array_keys($gameDataValues);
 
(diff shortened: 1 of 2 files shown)


Tsearo (commented):
I moved the sanitization to where it can be used because it is needed since you can inject html before I made the change and will be able to again if removed.

Before and after
179879885-c5b32516-9e58-4b70-b9e2-246573f826fd
179880540-6eb546e6-d529-43c1-a511-e6c1a376340e

Jamiras:
Makes sense. I moved the sanitization to after the matching logic:
image

Tsearo (approved):
Looks good

PR #1083, opened 2022-08-11
add softcore leaderboard rankings

Jamiras:
Allows players who primarily (or exclusively) use softcore mode to better track their own progress compared to other users.

_NOTE: Softcore removes all restrictions on the emulators. In addition to save states and rewind, you can use things like game genie codes, edit memory, disable layers to see through fog, view tile buffers to see stuff offscreen, or even have a TAS script play the whole game for you. Having a softcore leaderboard requires user to acknowledge that other players on that leaderboard could be using those features without any repercussions. The admin team has already stated that they will not investigate cheat reports submitted against softcore unlocks. So while it might be nice to see yourself go up a rank every few weeks, the rank itself loses its meaning. SomeOtherPlayer isn't better than you, he just uses more tools._

If you have at least one hardcore point, the hardcore information will appear on your profile page. If you have at least one softcore point, the softcore information will appear on your profile page. Softcore and hardcore leaderboards are not exclusive to each other. A player may appear in both. The number of ranked users for each leaderboard is simply the number of users in the database with at least the minimum amount of points for the leaderboard (currently 250 for both).

image

Clicking the softcore rank link takes the user to the softcore global leaderboard. The Community menu still links to the hardcore global leaderboard, but a filter has been added allowing the user to switch between them. Switching the mode resets the offset to 0, showing the top 25 users for the selected mode.

image

The user tooltips have also been updated to better distinguish between the two playstyles. For users who have more hardcore than softcore points, only the hardcore points and rank will be shown in the tooltip:

image

A user with more softcore points than hardcore points will show the softcore points and rank in the tooltip:

image

### DIFF
--- a/database/20220810_000000_Add_UserAccount_SoftcorePoints_Key.sql
+++ b/database/20220810_000000_Add_UserAccount_SoftcorePoints_Key.sql
@@ -1,0 +1,1 @@
+ALTER TABLE UserAccounts ADD KEY `RASoftcorePointsUntracked` (`RASoftcorePoints`,`Untracked`) USING BTREE;
--- a/lib/render/leaderboard.php
+++ b/lib/render/leaderboard.php
@@ -566,8 +566,10 @@ function getGlobalRankingData($lbType, $sort, $date, $user, $friendsOf = null, $
     } else {
         if ($unlockMode == UnlockMode::Hardcore) {
             $achPoints = "CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN ach.Points ELSE 0 END";
+            $achCount = "CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END";
         } else {
             $achPoints = "CASE WHEN HardcoreMode = " . UnlockMode::Softcore . " THEN ach.Points ELSE -ach.Points END";
+            $achCount = "CASE WHEN HardcoreMode = " . UnlockMode::Softcore . " THEN 1 ELSE -1 END";
         }
 
         $query = "SELECT User,
@@ -580,7 +582,7 @@ function getGlobalRankingData($lbType, $sort, $date, $user, $friendsOf = null, $
                   (
                       (
                           SELECT aw.User AS User,
-                          COUNT(ach.ID) AS AchievementCount,
+                          SUM($achCount) AS AchievementCount,
                           SUM($achPoints) as Points,
                           SUM(ach.TrueRatio) AS RetroPoints,
                           NULL AS TotalAwards
--- a/lib/render/leaderboard.php
+++ b/lib/render/leaderboard.php
@@ -1,5 +1,6 @@
 <?php
 
+use RA\AwardType;
 use RA\Rank;
 use RA\UnlockMode;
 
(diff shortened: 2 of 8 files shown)


database/20220810_000000_Add_UserAccount_SoftcorePoints_Key.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE UserAccounts ADD KEY `RASoftcorePointsUntracked` (`RASoftcorePoints`,`Untracked`) USING BTREE;
```
Jamiras, line 1: The softcore rank query was very slow without this. It's basically a clone of the `RAPointsUntracked` KEY.

lib/render/leaderboard.php
```diff
@@ -150,7 +151,7 @@ function RenderScoreLeaderboardComponent(string $user, bool $friendsOnly, int $n
             }
 
             if ($friendsOnly) {
-                $data = getGlobalRankingData($j, 5, $currentDate, null, $user, 0, 0, $friendCount, 1);
+                $data = getGlobalRankingData($j, 5, $currentDate, null, $user, 0, 0, $friendCount + 1, 1);
```
Jamiras, line 154: +1 to account for the user also appearing in the list

lib/render/leaderboard.php
```diff
@@ -471,19 +475,50 @@ function getGlobalRankingData($lbType, $sort, $date, $user, $friendsOf = null, $
...
-                    COALESCE(ROUND(ua.TrueRAPoints/ua.RAPoints, 2), 0) AS RetroRatio ";
+            if ($unlockMode == UnlockMode::Hardcore) {
+                $selectQuery = "SELECT ua.User,
+                        (SELECT COALESCE(SUM(CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END), 0) FROM Awarded AS aw WHERE aw.User = ua.User) AS AchievementCount,
+                        COALESCE(ua.RAPoints, 0) AS Points,
+                        COALESCE(ua.TrueRAPoints, 0) AS RetroPoints,
+                        COALESCE(ROUND(ua.TrueRAPoints/ua.RAPoints, 2), 0) AS RetroRatio ";
+            } else {
+                $selectQuery = "SELECT ua.User,
+                        (SELECT COALESCE(SUM(CASE WHEN HardcoreMode = " . UnlockMode::Softcore . " THEN 1 ELSE -1 END), 0) FROM Awarded AS aw WHERE aw.User = ua.User) AS AchievementCount,
```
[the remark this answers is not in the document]
Jamiras: I think we may have to live with this behavior until we have some of the v2 data structures in place.

We can accurately count hardcore achievements for a date because we only have to look at the hardcore unlock and don't care about the softcore unlock. To accurately count softcore unlocks we have to join back to the Awarded table and only count entries where there isn't a hardcore unlock for the same achievement on _any day_. The v2 data structure will have both the hardcore unlock date and softcore unlock date in the same row so we can do the filters more efficiently.

For now, I'm going to add a filter to ignore Achievement counts less than zero (there's already a filter for Points less than zero for a similar reason).

Tsearo (approved):

lib/render/leaderboard.php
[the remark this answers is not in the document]
Tsearo: Yeah I don't think there is any reasonable way to do this accurately at the moment like you said, we can just make an issue for it to fix once the v2 structure is available.

database/20220810_000000_Add_UserAccount_SoftcorePoints_Key.sql
luchaos: Executed on production

PR #1084, opened 2022-08-11
Fix game award counters overflow

EmoonX:
This is a simple amend for #1070. One of the two counters used to overflow below (at least in Windows) in case there were too many digits to display:

counter-overflow

Fixed by decreasing `<h3>`'s size on the right container (which has a rather positive side-effect since those were too big anyway).  Now it should hold up to 6 digits total.

### DIFF
--- a/public/css/styles.css
+++ b/public/css/styles.css
@@ -1612,47 +1612,6 @@ svg > g > g:last-child { pointer-events: none }
   border: 1px solid rgb(44, 151, 250);
 }
 
-.tabcontentscores th {
-  text-align: center;
-  font-weight: bold;
-}
-
-.tabcontentscores td {
-  text-align: center;  
-  padding: 2px 5px;
-}
-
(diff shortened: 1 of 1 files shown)


Tsearo:
Confirmed it functionally resolves the issue up to 6 total digits on Windows, 
image

May want to future proof it a bit by being able to handle 7 possibly which is probably the realistic max we'd ever see any time soon. Although it's possible with the upcoming V2 UI changes this won't be an issue so not worth putting much effort into now.

I'll leave whether or not changing all right container h3 sizes (game page, achievement inspector, etc) is good or not to the more UI inclined members as I'm by no knowledgeable on good practices there

luchaos:
Heads up: this will be removed with the v2 branch as the size of all headings has been slightly reduced.
Another option will be to increase the sidebar size on larger screens.

luchaos (approved):

PR #1085, opened 2022-08-11
Finished claims widget

Tsearo:
Changes out the New Achievements component with Finished Set Claims component

Old
image

New
image

Reasoning being that a majority of the time New Achievements essentially shows that a single game got a set created. Changing to show newly finished sets/revisions show more useful details and if you want to see specific new achievements you will likely be checking the Achievement List page anyways which shows them all.

Finished Claims "more..." link will take you to the Claim List page filtered on Complete statuses.

### DIFF
--- a/lib/render/set-claim.php
+++ b/lib/render/set-claim.php
@@ -5,28 +5,19 @@ use RA\ClaimSetType;
 use RA\ClaimSorting;
 
 /**
- * Creates the Set Claims component.
+ * Creates the New Set Claims component.
  */
-function renderClaimsComponent(int $count, int $claimFilter = ClaimFilters::Default, int $claimSort = ClaimSorting::ClaimDateDescending): void
+function renderNewClaimsComponent(int $count): void
 {
     echo "<div class='component'>";
-    if ($claimFilter == ClaimFilters::CompletedFilter) {
-        echo "<h3>Finished Set Claims</h3>";
-    } else {
-        echo "<h3>New Set Claims</h3>";
-    }
+    echo "<h3>New Set Claims</h3>";
 
-    $claimData = getFilteredClaimData(0, $claimFilter, $claimSort, false, null, false, 0, $count);
+    $claimData = getFilteredClaimData(0, ClaimFilters::Default, ClaimSorting::ClaimDateDescending, false, null, false, 0, $count);
 
     echo "<tbody><table>";
     echo "<th>User</th>";
     echo "<th>Game</th>";
-    if ($claimFilter == ClaimFilters::CompletedFilter) {
-        echo "<th nowrap>Type</th>";
-        echo "<th nowrap>Finished On</th>";
-    } else {
-        echo "<th nowrap>Claimed On</th>";
-    }
+    echo "<th nowrap>Claimed On</th>";
     foreach ($claimData as $claim) {
         $claimUser = $claim['User'];
         echo "<tr><td class='text-nowrap'>";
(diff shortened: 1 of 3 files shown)


lib/render/set-claim.php
```diff
@@ -1,22 +1,32 @@
...
 use RA\ClaimFilters;
+use RA\ClaimSetType;
 use RA\ClaimSorting;
 
 /**
- * Creates the New Set Claims component.
+ * Creates the Set Claims component.
  */
-function renderNewClaimsComponent(int $count): void
+function renderClaimsComponent(int $count, int $claimFilter = ClaimFilters::Default, int $claimSort = ClaimSorting::ClaimDateDescending): void
```
Jamiras, line 10: I'm really struggling with this change. About half of the function is shared, and the other half is conditionally A or B based on a parameter that can have values other than A or B.

I really think this should probably be two separate functions despite the fact that's there's a decent amount of redundancy between them.

Jamiras (commented):
Great idea. Seems to work as expected. Left some comments about the implementation.

src/ClaimFilters.php
```diff
@@ -33,6 +33,11 @@ abstract class ClaimFilters
     // This should be updated every time a new filter is added so it has all possible filter bits set
     public const AllFilters = (1 << 13) - 1;
 
+    // Filter to show all Complete Primary claims
+    public const CompletedFilter = self::AllFilters & ~self::ActiveClaim
+                                            & ~self::DroppedClaim
+                                            & ~self::CollaborationClaim;
```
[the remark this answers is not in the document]
Tsearo: Thanks that is a much better name for this I agree

lib/render/set-claim.php
Tsearo: I was a bit on the fence about it as well but figured less code was better. I'll split it out though as it's easier to read that way

televandalist:
I find this very enjoyable. Yes.

snowpin:
Ah yea! This makes much more sense to have over a list of achievements

Jamiras (approved):

lib/render/set-claim.php
```diff
@@ -36,3 +37,40 @@ function renderNewClaimsComponent(int $count): void
...
     echo "</div>";
 }
+
+/**
+ * Creates the Completed Set Claims component.
+ */
+function renderFinishedClaimsComponent(int $count): void
+{
+    echo "<div class='component'>";
+    echo "<h3>Finished Set Claims</h3>";
```
luchaos, line 47: "New Sets"?

lib/render/set-claim.php
```diff
@@ -36,3 +37,40 @@ function renderNewClaimsComponent(int $count): void
...
+
+    $claimData = getFilteredClaimData(0, ClaimFilters::AllCompletedPrimaryClaims, ClaimSorting::FinishedDateDescending, false, null, false, 0, $count);
+
+    echo "<tbody><table>";
+    echo "<th>User</th>";
+    echo "<th>Game</th>";
+    echo "<th nowrap>Type</th>";
+    echo "<th nowrap>Finished On</th>";
```
luchaos, line 55: "On" ca be omitted imo.

luchaos (approved):

lib/render/set-claim.php
Jamiras: It's also completed revisions, and it complements the "new set claims" that's further down the page.

luchaos: Was about to write that into a discussion - probably still should. Maybe that's all just a matter of taste.

I find the increasing use of "claims" on the front page problematic.

Claims - while unfortunately necessary - just describe the underlying processes that we had to introduce due to the shortcomings of what the site can do.

Personally, I'd rather focus on what this should become - a more collaborative system. "Claiming" something, to me, implies ownership and that should not have to exist.
Coming to the site as a player "claims" requires more explanation than calling it "new/revised sets" and/or "sets in progress".

Anyways, this probably not the best place to discuss this but wanted to have written down somewhere.

Tsearo: Would something like `Recently Finished Sets / Revisions` work better in your opinion?

Jamiras: "Recently Finished" seems overly wordy. Just "New Sets/Revisions" seems fine.

I'm not sure about better wording for "New Set Claims" though. "Sets in Progress" (as suggested above) seems like the best candidate. Then you'd also want to change "Claimed On" to "Started".

Jamiras: > Was about to write that into a discussion - probably still should.

This probably deserves a discussion and a separate PR as the "claim" nomenclature permeates the site.

At a minimum, the menu item "Community > Claim List" should also be updated to "Sets in Progress". The rest of the references are arguably acceptable (even the "Claim List" page itself) as it does represent a view of the process.

luchaos: I'd see "recently" as implied/superfluous, too. It's the front page so it should be the new stuff only anyways.

Agreed, from a developer's perspective it's all about claims; I'd keep public pages (except for where it's explicitly about claims) targeted at players and go for "sets" as that's what a player interacts with eventually, not underlying claim.

PR #1109, opened 2022-08-27
fix white point calculation for daily/weekly global leaderboards

Jamiras:
white points should only be tallied for hardcore unlocks


luchaos (approved):

Jamiras:
oops. duplicate of #1102

PR #1117, opened 2022-09-17
fix resetting progress from achievement page

### LINKED ISSUE #1116: Unable to reset achievement progress in RAWeb 2.0.0
image

I get this error when trying to reset all progress in a set or individual achievement progress.

Jamiras:
fixes #1116

### DIFF
--- a/public/achievementInfo.php
+++ b/public/achievementInfo.php
@@ -146,6 +146,24 @@ RenderContentStart($achievementTitleRaw);
     }
     </script>
 <?php endif ?>
+
+<?php if ($achievedLocal): ?>
+    <script>
+    function ResetProgress() {
+        if (confirm('Are you sure you want to reset this progress?')) {
+            showStatusMessage('Updating...');
+
+            $.post('/request/user/reset-achievements.php', {
+                achievement: <?= $achievementID ?>
+            })
+                .done(function () {
+                    location.reload();
+                });
+        }
+    }
+    </script>
+<?php endif ?>
+
 <div id="mainpage">
     <div id="leftcontainer">
         <?php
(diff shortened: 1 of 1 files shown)


public/achievementInfo.php
```diff
@@ -231,11 +249,7 @@ function updateAchievementTypeFlag(typeFlag) {
             echo "<div class='devbox mb-3'>";
             echo "<span onclick=\"$('#resetboxcontent').toggle(); return false;\">Reset Progress ▼</span>";
             echo "<div id='resetboxcontent' style='display: none'>";
-            echo "<form action='/request/user/reset-achievements.php' method='post' onsubmit='return confirm(\"Are you sure you want to reset this progress?\")'>";
-            echo csrf_field();
-            echo "<input type='hidden' name='a' value='$achievementID'>";
```
luchaos, line 236: looks like the input name was the culprit there

Jamiras: Yep. But after fixing that, the submission would page turn to the javascript response, which is why I replaced the form with an ajax call.

public/achievementInfo.php
```diff
@@ -146,6 +146,24 @@ function updateAchievementTypeFlag(typeFlag) {
     }
     </script>
 <?php endif ?>
+
+<?php if ($achievedLocal): ?>
```
luchaos, line 150: personally i'd opt for default form post request behaviour instead of emulating it via javascript.
i think adjusting the input name should be enough?

Jamiras: reset-achievements.php returns JSON, not a web page or redirect.

public/achievementInfo.php
luchaos: oh, i see. disregard my other comment then. this mix of json and default form requests is a mess 😅

luchaos (approved):

PR #1124, opened 2022-09-18
Use media asset helper where applicable

luchaos:


luchaos:
Urgent fix - incorrect URLs are filling up the wrong logs. Merging.

PR #1130, opened 2022-09-21
add RenderUserLink function

Jamiras:
Replaces two calls to `GetUserAndTooltipDiv` for every place where a user and their icon was shown. Each call to `GetUserAndTooltipDiv` makes two queries: A simple one to fetch some data from the UserAccounts table, and a more complex one to calculate the user's rank. The first is relatively quick, but the second has measurable impact.

Additionally, for widgets that show the user/icon links, the tooltip can be cached between rows to avoid even more queries. For forum pages and the comments widget, this reduced page load times by 10-20%. For code notes, it reduced the load time by as much as 90%.

### DIFF
--- a/app/Support/Shortcode/Shortcode.php
+++ b/app/Support/Shortcode/Shortcode.php
@@ -2,6 +2,7 @@
 
 namespace App\Support\Shortcode;
 
+use RA\LinkStyle;
 use Thunder\Shortcode\Event\FilterShortcodesEvent;
 use Thunder\Shortcode\EventContainer\EventContainer;
 use Thunder\Shortcode\Events;
@@ -186,7 +187,7 @@ final class Shortcode
             return '';
         }
 
-        return GetUserAndTooltipDiv($username, false);
+        return GetUserLink($username, LinkStyle::Text);
     }
 
     private function autolinkRetroachievementsUrls(string $text): string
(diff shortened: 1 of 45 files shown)


luchaos (changes requested):
Love where this it going ❤️ 

I had something similar in mind - reducing the shortcode/tooltip's rendering and query time will increase performance tremendously. however i'd like to attack this a bit differently.

- we can leverage the framework's caching capabilities instead of passing an array all the way down
- _any_ avatar should be cached, be it rendered text or image or both, not only users' (achievement, game, ticket, leaderboard, ...)
- ephemeral caching can be done using the `array` cache driver
- i'd argue that all of the cards' data can be cached for at least a minute - those do not have to be up to date at all times as most users won't even see them, yet, they affect all users' response times
- eventually, the cards shouldn't even be rendered on page load but asynchronously on hover which allows us to move to full response caching instead of caching individual queries's data

i had something prepared in the v2 prototype for dealing with large amounts of shortcode embeds by collecting all references (`[ach=]`, `[game=]`,`[user=]`, etc) and then fetch data from the db efficiently (deduplicated etc). that comes with a few drawbacks thus i landed on the cache/async approach.

another issue to me is the handling of image size naming and how the `LinkStyle` enum is structured - there is a media config which does something similar. i'd like to see those to be in line.

luchaos:
> eventually, the cards shouldn't even be rendered on page load but asynchronously on hover which allows us to move to full response caching instead of caching individual queries's data

See https://github.com/RetroAchievements/RAWeb/pull/1142

Jamiras:
Closing in favor of #1142. Will revisit at a later time to reintroduce the concept of `LinkStyle` to eliminate duplicated queries for the image link and text link.

PR #1136, opened 2022-09-25
Handle missing referers in wiki edit redirects

luchaos:
Fixes `Undefined array key "HTTP_REFERER"`

### DIFF
--- a/public/wiki-edit-redirect.php
+++ b/public/wiki-edit-redirect.php
@@ -18,7 +18,7 @@ $repoDomainMap = [
     'docs.retroachievements.org' => 'docs',
     'guides.retroachievements.org' => 'guides',
 ];
-$repo = $repoDomainMap[parse_url($_SERVER['HTTP_REFERER'])['host']] ?? null;
+$repo = $repoDomainMap[parse_url($_SERVER['HTTP_REFERER'] ?? null)['host'] ?? null] ?? null;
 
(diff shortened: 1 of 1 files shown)


public/wiki-edit-redirect.php
```diff
@@ -18,7 +18,7 @@
     'docs.retroachievements.org' => 'docs',
     'guides.retroachievements.org' => 'guides',
 ];
-$repo = $repoDomainMap[parse_url($_SERVER['HTTP_REFERER'])['host']] ?? null;
+$repo = $repoDomainMap[parse_url($_SERVER['HTTP_REFERER'] ?? null)['host'] ?? null] ?? null;
```
Jamiras, line 21: Won't this just cause a "cannot index ['host'] from null" error?

luchaos: By using the null-coalescing `??` operator it shouldn't run into an undefined index

luchaos: It does however trigger a deprecation warning - goal wasn't to fully fix it but to prevent it from filling up the logs. A better solution will be introduced when using the controller action from the v2 prototype.

Jamiras: I didn't realize `??` short-circuited the evaluation.

Jamiras (approved):

PR #1139, opened 2022-09-25
Use cachable configuration values instead of environment variables

luchaos:
Allows to cache configuration during deployment again. Improves request response times.


Jamiras (approved):
I don't see anything obviously wrong, but I'd normally want to test this to make sure there weren't any typos in the conversion. I don't think most of this can be tested in the dev environment, so I'll accept your assurances.

luchaos:
> I don't think most of this can be tested in the dev environment

Ah, yes, I should've added a test recipe:
You can cache the configuration running `php artisan config:cache`; everything has to work the same as before.
After running it locally, testing, the resulting cache file should be removed to have configuration evaluated ad-hoc again: `bootstrap/cache/config.php`.
The same will be done with the routes as soon as it's possible to do so. The legacy layer won't allow for that just yet.

Will be available on staging for further smoke testing in a minute: https://stage.retroachievements.org

luchaos:
Outside of the `config/` directory there should be no `getenv()` or `env()` calls anymore. After caching the configuration none of the environment variables are available to to application anymore.

PR #1140, opened 2022-09-25
show decimal tooltip for hex values when viewing trigger

Jamiras:
image


luchaos (approved):
Not directly related; personally I'd prefer using type safe comparisons `===` going forward.
Increasing the PHPStan will mark any unsafe comparisons as an issue.
It will be tricky to introduce that to the existing codebase (much like typed parameters) but I'd still like to attack this at some point.

PR #1142, opened 2022-09-25
Lazily load tooltip cards

luchaos:
Includes multiple layers of caching.

Moves rendering of cards (user, achievement, game, ticket) to a dedicated function that can be invoked asynchronously in a new card render request endpoint.

- Load tooltip cards lazily; best way to improve performance is to not load nor render data that is not needed immediately
- Cache (local/javascript) card responses per page visit
- Cache (server/redis) user rank data

User rank cache is not invalidated but can be. Reflecting the correct rank in the card is not important enough to find all valid invalidation opportunities yet. It would be if cache TTL was way above 15 minutes.

### DIFF
--- a/lib/render/achievement.php
+++ b/lib/render/achievement.php
@@ -13,33 +13,12 @@ function GetAchievementAndTooltipDiv(
     $smallBadgeSize = 32,
     $imgclass = 'badgeimg'
 ): string {
-    $tooltipIconSize = 64; // 96;
-
     sanitize_outputs(
         $achName,
         $consoleName,
         $gameName,
         $achPoints
     );
-
-    $achNameStr = $achName;
-    $achDescStr = $achDesc;
-    $gameNameStr = $gameName;
-
-    $tooltip = "<div id='objtooltip' class='flex items-start' style='max-width: 400px'>";
-    $tooltip .= "<img style='margin-right:5px' src='" . media_asset("Badge/$badgeName.png") . "' width='$tooltipIconSize' height='$tooltipIconSize' />";
-    $tooltip .= "<div>";
-    $tooltip .= "<b>$achNameStr</b><br>";
-    $tooltip .= "$achDescStr<br>";
-    if ($achPoints) {
-        $tooltip .= "<br>$achPoints Points<br>";
-    }
-    $tooltip .= "<i>$gameNameStr</i><br>";
-    $tooltip .= $extraText;
-    $tooltip .= "</div>";
-    $tooltip .= "</div>";
-
-    $tooltip = tipEscape($tooltip);
 
     $smallBadge = '';
     $displayable = "$achName";
@@ -59,12 +38,42 @@ function GetAchievementAndTooltipDiv(
         }
     }
 
-    return "<div class='inline' onmouseover=\"Tip('$tooltip')\" onmouseout=\"UnTip()\" >" .
+    return "<div class='inline' onmouseover=\"Tip(loadCard('achievement', $achID))\" onmouseout=\"UnTip()\" >" .
         "<a href='/achievement/$achID'>" .
         "$smallBadge" .
         "$displayable" .
         "</a>" .
         "</div>";
+}
+
+function renderAchievementCard(int $achievementId): string
+{
+    $achData = [];
+    getAchievementMetadata($achievementId, $achData);
+
+    $badgeName = $achData['BadgeName'];
+    $achNameStr = $achData['AchievementTitle'];
+    $achDescStr = $achData['Description'];
+    $gameNameStr = $achData['GameTitle'];
+    $achPoints = $achData['Points'];
+
+    $tooltip = "<div class='tooltip-body flex items-start gap-2 p-2' style='max-width: 400px'>";
+    $tooltip .= "<img src='" . media_asset("Badge/$badgeName.png") . "' width='64' height='64' />";
+    $tooltip .= "<div>";
+    $tooltip .= "<div><b>$achNameStr</b></div>";
+    $tooltip .= "<div class='mb-1'>$achDescStr</div>";
+    if ($achPoints) {
+        $tooltip .= "<div>$achPoints Points</div>";
+    }
+    $tooltip .= "<div><i>$gameNameStr</i></div>";
+
+    // TODO: extra text should tell if user has unlocked the achievement; use request()->user() for that
+    // $tooltip .= $extraText;
+
+    $tooltip .= "</div>";
+    $tooltip .= "</div>";
+
+    return $tooltip;
 }
 
 function RenderRecentlyUploadedComponent($numToFetch): void
--- a/lib/render/user.php
+++ b/lib/render/user.php
@@ -23,23 +23,32 @@ function GetUserAndTooltipDiv(
 
         $userSanitized = $user;
         sanitize_outputs($userSanitized);
+
         return '<del>' . $userSanitized . '</del>';
     }
 
-    return _GetUserAndTooltipDiv($user, $userCardInfo, $imageInstead, $customLink, $iconSizeDisplayable, $iconClassDisplayable);
+    $linkURL = '/user/' . $userCardInfo['User'];
+    if (!empty($customLink)) {
+        $linkURL = $customLink;
+    }
+
+    $displayable = $userCardInfo['User'];
+    if ($imageInstead) {
+        $displayable = "<img loading='lazy' src='" . media_asset('/UserPic/' . $user . '.png') . "' width='$iconSizeDisplayable' height='$iconSizeDisplayable' alt='$displayable' class='$iconClassDisplayable' />";
+    }
+
+    return "<span class='inline' onmouseover=\"Tip(loadCard('user', '$user'))\" onmouseout=\"UnTip()\">" .
+        "<a href='$linkURL'>" .
+        "$displayable" .
+        "</a>" .
+        "</span>";
 }
 
-function _GetUserAndTooltipDiv(
-    $user,
-    $userCardInfo,
-    $imageInstead = false,
-    $customLink = null,
-    $iconSizeDisplayable = 32,
-    $iconClassDisplayable = 'badgeimg'
-): string {
-    $userSanitized = $user;
-    sanitize_outputs($userSanitized);
+function renderUserCard(string $user): string
+{
+    getUserCardData($user, $userCardInfo);
 
+    $username = $userCardInfo['User'];
     $userMotto = $userCardInfo['Motto'];
     $userHardcorePoints = $userCardInfo['HardcorePoints'];
     $userSoftcorePoints = $userCardInfo['SoftcorePoints'];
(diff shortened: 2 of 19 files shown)


lib/render/user.php
```diff
@@ -23,23 +23,32 @@ function GetUserAndTooltipDiv(
...
+    if (!empty($customLink)) {
+        $linkURL = $customLink;
+    }
+
+    $displayable = $userCardInfo['User'];
+    if ($imageInstead) {
+        $displayable = "<img loading='lazy' src='" . media_asset('/UserPic/' . $user . '.png') . "' width='$iconSizeDisplayable' height='$iconSizeDisplayable' alt='$displayable' class='$iconClassDisplayable' />";
+    }
+
+    return "<span class='inline' onmouseover=\"Tip(loadCard('user', '$user'))\" onmouseout=\"UnTip()\">" .
```
Jamiras, line 40: This successfully displaces the rank query, which is the more expensive part of building the tooltip, but `getUserCardData` is still called on line 17 every time this method is called (once for image, once for text, repeated for each occurrence on the page). That's one of the things I was addressing in #1130.

From viewing an achievement:
* 1130: `148 statements were executed, 5 of which were duplicated, 143 unique 11.49s`
* this: `148 statements were executed, 115 of which were duplicated, 33 unique 9.75s`

I offloaded 115 duplicate queries (redundant between image and text), but replaced them with 115 queries for the user ranks, which was clearly slower.

I'd like to reapply that logic in a separate PR after this is merged.

lib/render/achievement.php
```diff
@@ -59,14 +38,44 @@ function GetAchievementAndTooltipDiv(
...
+    $tooltip .= "<div>";
+    $tooltip .= "<div><b>$achNameStr</b></div>";
+    $tooltip .= "<div class='mb-1'>$achDescStr</div>";
+    if ($achPoints) {
+        $tooltip .= "<div>$achPoints Points</div>";
+    }
+    $tooltip .= "<div><i>$gameNameStr</i></div>";
+
+    // TODO: extra text should tell if user has unlocked the achievement; use request()->user() for that
+    // $tooltip .= $extraText;
```
Jamiras, line 71: Were you planning to address this now? It is a change in behavior for the tooltips on a user profile. In addition to not showing the unlock time, it always shows the unlocked image.

public/js/all.js
```diff
@@ -114,9 +114,33 @@ function replaceAll(find, replace, str) {
   return str.replace(new RegExp(find, 'g'), replace);
 }
 
+var cardsCache = {};
```
Jamiras, line 117: I presume it wouldn't be difficult to populate this from the body of a page? For example, there's no reason not to preload the achievement tooltips on a game page as all the necessary data is already shown on the page itself.

Jamiras (approved):
Very nice.

public/js/all.js
luchaos: > all the necessary data

True, yet sounds like premature optimization to me (ironically). Not all of said data will be available at all times as we might want to move around sections and context of pages.
Making it dependent on the context of the currently viewed page introduces complexity i'd rather get rid of/not introduce. Rewriting it all to have it prepared on page load seems overly complicated to me. I'm open for suggestions however, i cannot claim to have it all thought through.

To me, loading tooltips cards should be a separate aspect of the system at all times; able to be cached and rendered separately from the rest of the site.

I'm aware that there most of the data is already fetched at the time the tooltips could be rendered. Still, not assuming that all data is available, but fetching it when needed seems wiser to me. We should look for loading _less_ data for rendering the initial data instead. This PR opens some opportunities for that.

lib/render/user.php
luchaos: I was about to add caching to `getUserCardData` - but i found it to be a very simple query only fetching account details.
Added ephemeral/per-request caching; does this address the issue? https://github.com/RetroAchievements/RAWeb/pull/1142/commits/4670b78fa17eef6bc9f3dcae114a55c47dcdfe5f

lib/render/achievement.php
luchaos: I feel like i should, yes - it leaves a gap in expected behaviour; yet, I'm not sure if that unlock information grants enough value in all contexts? (It probably does)

lib/render/user.php
Jamiras: Those numbers were from a random achievement page - 82 unique people earned the achievement (50 shown on the page), and 5 users in my all-time followed users ranking.

I tried again with and without the per-request caching, and there's clearly an impact on the number of queries, but no noticeable impact to the overall time elapsed.

* before: `148 statements were executed, 115 of which were duplicated, 33 unique 14.21s`
* after: `93 statements were executed, 5 of which were duplicated, 88 unique 14.36s`

And the math works out. 148-93=55. 55x2=110. 115-110=5. 55 duplicated statements were eliminated. Since they were each duplicated, they were counted twice in the duplicated list. Now they appear in the unique list once. 50 winners + 5 friends = 55.

However, applying the same logic to a code notes page where there were 65 code notes all from the same author:

* before: `75 statements were executed, 67 of which were duplicated, 8 unique 468ms`
* after: `11 statements were executed, 2 of which were duplicated, 9 unique 44.42ms`

The 65 duplicated lookups for the code note author became 1 unique lookup and saved 90% of the page load.

And the "recent forum posts" page (where 15 unique users posted 25 messages):

* before: `59 statements were executed, 52 of which were duplicated, 7 unique 120ms`
* after: `24 statements were executed, 2 of which were duplicated, 22 unique 98.35ms`

public/js/all.js
Jamiras: > Not all of said data will be available at all times

Obviously, if the information isn't available, avoiding the query to fetch it would be desirable. But in cases where it is available, wouldn't it be desirable to avoid the extra query and the extra round trip to the server?

For the game page, the data is available - including the unlock timestamps:
image

I agree that it's unlikely that a user would need all of the tooltips, but if the data is available, why not send it on the first response and avoid the extra round trips.

For the game page, the tooltip don't even seem necessary. They're just redundant, which supports my assertation that all the necessary data is already available. In this case, it might be better to remove the tooltip entirely.

For something like the achievement grids on the profile page, the tooltip becomes much more important as there is no text displayed. The query to build the recent games grid already fetches all the information necessary to generate the tooltip (achievement title, description, points, game title, unlock date):
https://github.com/RetroAchievements/RAWeb/blob/62b1e3216fa076b5cde344be14f02e516db78f34/lib/database/player-achievement.php#L232

So why not use it?

lib/render/achievement.php
Jamiras: I think it would be neat to see whether or not I've unlocked an achievement when achievements are referenced in the forums. Although, I'm not sure that the image should be greyed out if the user hasn't unlocked the achievement in that case.

It may be unnecessary in something like the ticket manager, but I don't think it would detract from the functionality.

Perhaps this could be an argument towards pre-generating those tooltips (with the unlock data) on the profile page, and leave the non-profile page requests without the unlock information?

Jamiras: Conversely, I've considered putting game progress in the game tooltips (once the denormalized data is ready). This solution makes that even more reasonable, and potentially possible without the denormalized data.
image

Jamiras: Also, the unlock state/timestamp information in the tooltip for achievements on user B's profile should be user B's unlock state/timestamp, but the current user's. To do that via the lazyCard would require passing additional context.

public/js/all.js
luchaos: > why not send it on the first response and avoid the extra round trips
> already fetches all the information necessary
> So why not use it?

Normally i'd absolutely agree and we can surely revisit that option at another time. It's an odd situation where the database is not the only bottleneck for response performance.

I think it introduces code for optimizing something that is not that much of an issue. Maybe I'm thinking of the wrong thing there and we should give it a try (in a separate step)-

The round trips currently aren't as expensive as rendering time. Even the suggested combination of icon and text queries aren't as much of a problem in comparison to the excessive pre-rendering of multiple hundreds tooltips on some pages.
We have forum topics that contain an almost uncapped amount of references to games/achievements/users which not only resulted in a lot of (sometimes expensive) queries but also pushed the rendering time to an excessive amount.

The server's cpu load is fairly evenly distributed between PHP and mysql atm - PHP should not take that much time rendering but leave as much resources to the database as possible (whilst we also reduce its own usage).

lib/render/achievement.php
luchaos: Agreed. 
That line is more important than I was aware for achievement unlock contexts where it's not shown elsewhere on the page (user pages as you mentioned).

Adding game progress wouldn't be as expensive anymore with similar caching, indeed.

I'll give it a shot right away.

luchaos: > I'm not sure that the image should be greyed out if the user hasn't unlocked the achievement in that case.

I'd stick with always using the colored version for now?

public/js/all.js
luchaos: > already fetches all the information necessary

I think that we actually could reduce the amount of data fetched, there's certainly some overfetching happening in some areas. I'll try to identify those areas at a later time.

lib/render/achievement.php
Jamiras: > I'd stick with always using the colored version for now?

I think that's fine. It'll be consistent with the dynamically loaded tooltips everywhere else in the code. (i.e. hovering over a locked achievement in the game page will now show the colored version). Although, I'm still leaning towards removing those.

public/js/all.js
Jamiras: > We have forum topics that contain an almost uncapped amount of references to games/achievements/users which not only resulted in a lot of (sometimes expensive) queries but also pushed the rendering time to an excessive amount.

I'm not suggesting we do it everywhere. Forums wouldn't have that data available, and I agree that pre-fetching it there is overhead that should be avoided. 

I'm just suggesting that in places where the data _is_ available, we could use it and avoid the extra server hit(s)/queries.

luchaos: I understand :) Personally, I'd rather take the occasional extra hits (which are only done once for each individual tooltip and most of the time not if no interaction happens) than introducing complexity by conditionally rendering them in some places and not in others. It makes it more rigid, too, as we'd have to _assure_ that all data will be kept available in those places where we have the data available.

I'll just give it a try to see what it could look like.

lib/render/achievement.php
luchaos: You mean removing the tooltips altogether or changing to the grayscale image?

Jamiras: https://github.com/RetroAchievements/RAWeb/pull/1142#discussion_r979516891

The tooltip is redundant on the game page and should probably just be removed.

If it's not removed, it'll be loaded dynamically and show the non-gray image, so showing the non-gray image in the tooltip on the profile page would be consistent.

luchaos: I see, I don't think removing the tooltip is necessary. It's potentially cached anyways and keeps it consistent with how all the other resource avatars work; if there's an icon or resource name it can also hold the trigger for fetching the tooltip card. I'd expect it to load the tooltip nonetheless.

luchaos: Oh, wait, I just realized you were referring to a different part of the game page than what I had in mind 😅 Yeah, the achievement tooltips are redundant there.

The methods should be renamed, really, in v2 they were all called avatars.
Adding a parameter to enable or disable tooltips might work fine then. I'll give that a try.

luchaos: Working on a separate PR for the optional tooltip data.

public/js/all.js
luchaos: Working on a separate PR for the optional tooltip data.

PR #1143, opened 2022-09-25
Cache currently active on front page

luchaos:
Microcaching the currently active request doesn't work anymore for multiple reasons. Cache it in redis (production; whatever you choose to use locally) instead.

### DIFF
--- a/public/request/user/list-currently-active.php
+++ b/public/request/user/list-currently-active.php
@@ -1,13 +1,20 @@
 <?php
 
-$ingameList = getLatestRichPresenceUpdates();
+use Illuminate\Support\Facades\Cache;
 
-$mergedList = [];
+$ttlSeconds = 60 * 15;
 
-foreach ($ingameList as $playerIngame) {
-    // Array merge/overwrite
-    $mergedList[$playerIngame['User']] = $playerIngame;
-    $mergedList[$playerIngame['User']]['InGame'] = true;
-}
+$currentlyActive = Cache::remember('currently-active', $ttlSeconds, function () {
+    $ingameList = getLatestRichPresenceUpdates();
+    $mergedList = [];
 
-return response()->json(array_values($mergedList));
+    foreach ($ingameList as $playerIngame) {
+        // Array merge/overwrite
+        $mergedList[$playerIngame['User']] = $playerIngame;
+        $mergedList[$playerIngame['User']]['InGame'] = true;
+    }
+
+    return array_values($mergedList);
+});
+
+return response()->json($currentlyActive);
--- a/public/request/user/list-currently-active.php
+++ b/public/request/user/list-currently-active.php
@@ -2,7 +2,7 @@
 
 use Illuminate\Support\Facades\Cache;
 
-$ttlSeconds = 60 * 15;
+$ttlSeconds = 60 * 2;
 
 $currentlyActive = Cache::remember('currently-active', $ttlSeconds, function () {
     $ingameList = getLatestRichPresenceUpdates();


public/request/user/list-currently-active.php
```diff
@@ -1,13 +1,20 @@
 <?php
 
-$ingameList = getLatestRichPresenceUpdates();
+use Illuminate\Support\Facades\Cache;
 
-$mergedList = [];
+$ttlSeconds = 60 * 15;
```
Jamiras, line 5: Definitely want to refresh this more than once every fifteen minutes. The client will poll every two. If it's not changing every two, then why poll that frequently? I'd recommend a minute at most. Maybe 30 seconds, 15, or even 10.

luchaos: It used to be 15 minutes when the microcache was still active on the server.

The client polling more frequently will and does serve a different use. Displaying active users on the front page is not as important as letting the platform know when a session should be updated.

The polling interval for ´RefreshActivePlayers` is exaggerated, agreed. It won't have much of an impact as the response is cached anyways. You were referring to the poll interval with the recommendation?

Jamiras: I was referring to this:
https://github.com/RetroAchievements/RAWeb/blob/d154bd9be5101e2e4174b08974efc1c3db328b02/public/js/activePlayersBootstrap.js#L105

Though I was positive that it was 2 minutes. If I just sit on the main page and watch the "last updated at" timestamp, I thought it changed every two minutes.

Jamiras: Verified that the home page does only refresh every 5 minutes. I still assert that that the cached value should be updated at least twice as frequently as that limit.

luchaos: It's one of the most frequently called endpoints going by the logs; seems a lot of users have the front page open idling there. That's why I had a 15 minutes microcache enabled on the server pre-v2 as it could take quite some time to load the data. The interval of the polling wasn't that relevant anymore in that case, it would just retrieve a cached response.

I'd prefer to not go below 5 minutes, 10 minutes even, for the server-side cache; the polling can be adjusted accordingly or just stay at 5 minutes; doesn't matter as long caching is applied.

Jamiras: If the client polls every 5 minutes, but the server returns cached data that's only updated every 15 minutes, that's two calls that effectively do nothing.

> Minute 1: user A opens page
> Minute 3: cached data is expired
> Minute 4: another user causes the cached data to be updated
> Minute 6: user A asks for updated data and gets updated data (generated at minute 4)
> Minute 11: user A asks for updated data and gets data generated at minute 4
> Minute 16: user A asks for updated data and gets data generated at minute 4
> Minute 19: cached data is expired
> Minute 20: another user causes the cached data to be updated
> Minute 21: user A asks for updated data and gets data generated at minute 20

The calls at minute 11 and 16 are pointless. The client already has that data.

Whereas if it expires every two minutes:

> Minute 1: user A opens page
> Minute 3: cached data is expired
> Minute 4: another user causes the cached data to be updated
> Minute 6: cached data is expired; user A asks for updated data and gets updated data (generated at minute 6)
> Minute 8: cached data is expired
> Minute 9: another user causes the cached data to be updated
> Minute 10: cached data is expired; another user causes the cached data to be updated
> Minute 11: user A asks for updated data and gets updated data (generated at minute 10)

I agree that caching the data so it only has to be calculated once for all clients is a great change. I'm just concerned that it should expire at least as frequently as the client polling interval. Preferably more frequently to avoid the rare case where the client interval and cache interval are basically sync'd and the depending on how quickly the cache is updated the client will sometimes get new data and sometimes not.

Assuming there's 20 users idling on the home page that have somehow managed to equally space themselves out so every request would hit an expired cache, you could expire the data every 15 seconds and still break even on the number  of queries being performed. Dropping that to 2 minutes reduces the number of queries by 87%. And I'm guessing that there's more than 20 users idling if "It's one of the most frequently called endpoints going by the logs".

luchaos: The server should decide all of that, not the frontend, answering the question of how up-to-date it has to be whilst reducing databases queries as much as possible. That used to be 15 minutes. The frontend polls too often is not an issue imo, it still would make sure that if the cache was refreshed within the first 5 minutes it would be presented to the user earlier.

luchaos: The cleanest way to improve this in the frontend would be to add an expiry header server-side and let the frontend poll again shortly after expiry the earliest

luchaos: well, the way you laid it out it's still the smarter way to go about it. let's start with a 2 minute cache and keep the polling at 5 minutes. 15 minutes probably was a bit too aggressive in the first place.
if it hits the server too hard we can still increase it

luchaos: Sorry, it seems I'm talking in circles today around my stubbornness :)
2 minutes ttl it is 👍

Jamiras (approved):

PR #1129, opened 2022-09-20
Update release.php

televandalist:
Adds 74 - Arcadia 2001, 75 - Interton VC 4000, and 76 - Elektor TV Games Computer to console list.

PR #1146, opened 2022-09-27
slightly widen menu to prevent wrapping

Jamiras:
image
to
image

Added a bit more spacing than needed to account for upcoming changes in #1129


resources/css/nav.css
```diff
@@ -111,7 +111,7 @@
   display: block;
   color: #000;
   line-height: 1.4em;
-  width: 150px;
+  width: 175px;
```
luchaos, line 114:
```suggestion
  min-width: 175px;
  white-space: nowrap;
```

Only pushes the width if an item exceeds the maximum width whilst assuring none break lines.

luchaos (approved):

resources/css/nav.css
Jamiras: This is certainly more future-proof, but it does create an oddity where each column of the Games menu is a different width:
image

3D0 Interactive Multiplayer appears in the second column, and has a more natural wrapping on the site currently so it didn't stand out.

luchaos: That's fair, fixed 175px makes more sense then to keep that intact.

PR #1147, opened 2022-09-27
Remove unused activity feed code

luchaos:
Will come back at a later time and work a bit differently - none of that code will be needed for that.


public/managehashes.php
```diff
@@ -73,7 +73,7 @@ function UnlinkHash(user, gameID, hash, elem) {
             var date = new Date();
             var dateStr = date.getUTCDate() + ' ' + shortMonths[date.getUTCMonth()] + ' ' + date.getUTCFullYear() + '<br>' + date.getUTCHours() + ':' + ('0' + date.getUTCMinutes()).slice(-2);
 
-            $('.comment-textarea').parents('tr').before('<tr class="feed_comment localuser system"><td class="smalldate">' + dateStr + '</td><td class="iconscommentsingle"></td><td class="commenttext">' + hash + ' unlinked by ' + user + '</td></tr>');
+            $('.comment-textarea').parents('tr').before('<tr class="comment system"><td class="smalldate">' + dateStr + '</td><td></td><td>' + hash + ' unlinked by ' + user + '</td></tr>');
```
Jamiras, line 76: Ah, missed these in #1145

Jamiras (approved):

public/managehashes.php
luchaos: I only looked for remaining references of those feed classes after merging your PR; otherwise I would have added it there :)

PR #1149, opened 2022-09-28
Optional tooltip triggers for resource avatars

luchaos:
Follow-up for discussion in #1142.

- Split of resource avatars and their tooltips.
- Allow to pass avatar/tooltip data to reduce server round trips and/or tooltip rendering


Jamiras:
First obvious problem: non-tooltip images on profile page are not being rendered correctly. Additionally, the game name appears twice: 
image

It looks like the padding changed too, so there's less items per line: 
image

Jamiras:
Forums are doubling up too:
image

Master:
image

Jamiras:
`Unknown named parameter $labelPoints` exception trying to render embedded achievement link in forum

Jamiras:
`Unknown named parameter $labelConsoleName` exception trying to render embedded game link in forum

Jamiras:
`gameAvatar(): Argument #4 ($iconSize) must be of type int, string given, called in /home/brian/source/RAWeb/public/ticketmanager.php on line 453` trying to open ticket manager

Jamiras:
`Named parameter $iconSize overwrites previous argument` viewing followers list

Jamiras:
`Trying to access array offset on value of type null` accessing Community > Developer page. I think it's related to #1151

luchaos:
> locked images are no longer greyed out on profile page
> The "last played" information should be on a separate line
> Deleted developer does not show up on developers list

Should be fixed now

Jamiras:
Locked achievements on game page are not greyed out
image

Jamiras:
Deleted developer now shows crossed out name for image (it doesn't on production)
image

Jamiras:
Name is duplicated on leaderboards sidebar
image

lib/database/achievement.php
```diff
@@ -79,7 +79,7 @@ function getAchievementsList(
     }
 
     $query = "SELECT
-                    ach.ID, ach.Title AS AchievementTitle, ach.Description, ach.Points, ach.TrueRatio, ach.Author, ach.DateCreated, ach.DateModified, ach.BadgeName, ach.GameID,
+                    ach.ID, ach.Title AS Title, ach.Title AS AchievementTitle, ach.Description, ach.Points, ach.TrueRatio, ach.Author, ach.DateCreated, ach.DateModified, ach.BadgeName, ach.GameID,
```
Jamiras, line 82: I have a feeling these changes will affect the public API as they mostly just dump the query objects. I suppose duplicated fields won't hurt anything, unless you're concerned about bandwidth.

lib/database/user.php
```diff
@@ -29,25 +28,23 @@ function getAccountDetails(&$user, &$dataOut): bool
...
+                AND Deleted IS NULL";
+
+    $dbResult = s_mysql_query($query);
+    if (!$dbResult || mysqli_num_rows($dbResult) !== 1) {
+        return false;
+    }
 
-        return mysqli_fetch_array($dbResult);
-    });
+    $dataOut = mysqli_fetch_array($dbResult);
```
Jamiras, line 47: this is going to supercede #1151

luchaos:
> Locked achievements on game page are not greyed out
> Deleted developer now shows crossed out name for image
> Name is duplicated on leaderboards sidebar

Fixed

lib/render/achievement.php
```diff
@@ -1,127 +1,108 @@
...
+
+    $data = [];
+    if (is_array($achievement)) {
+        $data = $achievement;
+    }
+
+    if (empty($data)) {
+        $data = Cache::store('array')->rememberForever('achievement:' . $id . ':card-data', function () use ($id) {
+            $data = [];
+            getAchievementMetadata($id, $data);
```
Jamiras, line 67: Do we really want to cache the entire metadata object? It includes stuff like `memAddr` that can be rather large, and aren't needed by the tooltip. Since it's just for the duration of the request, it's probably okay for now, but it's something we should revisit later.

lib/render/achievement.php
```diff
@@ -1,127 +1,108 @@
...
-    $achDescStr = $achData['Description'];
-    $gameNameStr = $achData['GameTitle'];
-    $achPoints = $achData['Points'];
+    // /** @var ?User $user */
+    // $user = request()->user();
+    // if ($context && $user) {
+    //     getAchievementUnlocksData($id, $numWinners, $numPossibleWinners, $numRecentWinners, winnerInfo: $data, user: $user->User);
+    //
+    //     $data['Unlock'] = 'UNLOCKED';
+    // }
```
Jamiras, line 79: commented code can be removed. looks to be handled below

lib/database/achievement.php
Jamiras, line 82: These changes could impact the public APIs as they mostly just dump query results. Duplicated data shouldn't matter, unless you care about the bandwidth.

lib/render/leaderboard.php
```diff
@@ -301,8 +297,8 @@ function RenderTopAchieversComponent($user, array $gameTopAchievers, array $game
         echo "</td>";
 
         echo "<td>";
-        echo GetUserAndTooltipDiv($nextUser, true, iconSizeDisplayable: 16);
-        echo GetUserAndTooltipDiv($nextUser);
+        echo userAvatar($nextUser, true, iconSize: 16);
+        echo userAvatar($nextUser);
```
Jamiras, line 301: I've noticed in some places you replace two calls to `GetUserAndTooltipDiv` with one call to `userAvatar`. Can't that be done here?

lib/render/news.php
```diff
@@ -58,7 +58,7 @@ function RenderNewsHeader(News $newsData): void
     echo "<div class='newstext shadowoutline absolute' style='width: 90%; top: 40px; left:10px;'>$payload</div>";
 
     // Author
-    echo "<div class='newsauthor shadowoutline absolute' style='width: 470px; top: 196px; left:0; text-align: right'>~~$authorLink, $niceDate</div>";
+    echo "<div class='newsauthor shadowoutline absolute' style='width: 470px; top: 196px; left:0; text-align: right'>$authorLink, $niceDate</div>";
```
Jamiras, line 61: Unrelated to this PR, I think we need to limit the text area of the carousel so the user/date don't overwrite the text.  It's even more awkward with the icon now.

image

public/linkedhashes.php
```diff
@@ -67,7 +67,7 @@
 
             echo '<br/><code> ' . $hash['Hash'] . '</code>';
             if (!empty($hash['User'])) {
-                echo ' linked by ' . GetUserAndTooltipDiv($hash['User']);
+                echo ' linked by ' . userAvatar($hash['User']);
```
Jamiras, line 70: this didn't used to have an image:
image

or possible line 84 (or both)

lib/database/achievement.php
luchaos: Made that change before making the avatar functions more robust, allowing for multiple variants of field names. That one in particular is a remnant of that. I was aware that those additional fields would show up in the web API but I don'T think it would be much of a problem. It's a public change nontheless. I'll revert that.

lib/render/achievement.php
luchaos: It was meant to allow for fetching unlock data in cases where no data is given for rendering the tooltip upfront. The `context` parameter would take care of that, however I did not find time to finish it, and it wasn't needed either as we pass said data along in all areas where it's needed. I'd like to keep it for now and see if i can finish it; passing the user context might come in handy in the future, not only for fetching unlock data for achievements asynchronously but for other resources types as well where user context can be applied.

I had a similar thing prepared in v2 where the user context for e.g. games would show additional user-specific infos.

lib/render/achievement.php
luchaos: Thought about that, too, but I don't think this is an issue for now either.

In the future we might want to think about not overfetching data in the first place by using ORM models that only hold the information required to fulfill the task of presenting data in the tooltip. At that point we'd have relations in place, and probably all moved to the blade templating engine.

lib/database/user.php
luchaos: I presumed so; wasn't sure thought. Good to know!

lib/database/achievement.php
luchaos: Just realized that achievementAvatar() and renderAchievementCard() already can take both variants. Definitely reverting.

lib/render/leaderboard.php
luchaos: Certainly, these seem to escape my eye even after the 20th regex search somehow. Thanks!

public/linkedhashes.php
luchaos: Adjusted accordingly

lib/render/news.php
luchaos: Removed the avatar.

Agreed. Currently it'S in the hands of the news editor to find a suitable length for the post. 

I plan on replacing the carousel with a general news overhaul after our performance hunt is over. Short variants in a new carousel, detail pages with comments, news index, ...

lib/render/achievement.php
luchaos: Removed - I'll revisit that feature another time.

Jamiras (approved):

RALordAndrew:
"Set Exists" text was removed from setRequestList.php. https://retroachievements.org/viewtopic.php?t=18451

PR #1150, opened 2022-09-28
Validate PS2 on site

televandalist:
Deploy on Oct 1st at 00:00 GMT.

- Validates PlayStation 2 on site

- Adds PCSX2 to emulator list. Minimum and latest versions have comment slashes since the exact version number is unknown until launch.

- Removes rollout message at the top of PS2 game entries.


storage/app/releases.dist.php
```diff
@@ -113,6 +113,20 @@
...
+            'name' => 'PCSX2',
+            'handle' => 'PCSX2
+            'active' => true,
+            'integration_id' => Emulators::PCSX2,
+            'link' => 'https://pcsx2.net',
+            'description' => '<strong>NOTE:</strong> The PCSX2 standalone is the only emulator officially supporting achievements for PlayStation 2.',
+            'systems' => [
+                21, // PlayStation 2
+            ],
+        ],
```
Jamiras, line 129: This isn't going to work. Assuming you did define the Emulators::PCSX2 enum, you're also missing a quote on line 121.

However, more importantly, this logic is for clients we maintain (see https://github.com/RetroAchievements/RAInterface/blob/master/RA_Emulators.h), where we can force the user to upgrade.

PCSX2 is using the non-emulatorID API to initialize the DLL, so the version validation code wouldn't be called anyway. Furthermore, when not using the DLL, no validation occurs.

You need to shoehorn something in akin to the RetroArch entry:
```
        101 => [
            'name' => 'PCSX2',
            'handle' => 'PCSX2',
            'active' => true,
            'link' => 'https://pcsx2.net/guides/basic-setup/',
            'description' => '<strong>NOTE:</strong> The PCSX2 standalone is the only emulator officially supporting achievements for PlayStation 2.
Download from <a href="https://pcsx2.net/downloads">pcsx2.net</a>.',
            'systems' => [
                21, // PlayStation 2
            ],
        ],
```

Jamiras: Lastly, do we want this featured as the third item in the list (below RetroArch and RALibretro, and above everything else)? I think it's probably fine there.

lib/render/content.php
```diff
@@ -126,16 +126,4 @@ function RenderAOTWComponent($achID, $forumTopicID): void
...
-    // PS2
-    if ($consoleId === 21) {
-        echo <<<HTML
-            <div class="mb-4">
-                <a href="/viewtopic.php?t=11108">⚠️️ Achievement developers are currently involved in a PlayStation 2 rollout. The planned launch date is October 1st, 2022 at 00:00 GMT! Click for more details. ⚠️</a>
-            </div>
-        HTML;
-    }
-}
+}
```
Jamiras, line 129: The CI is complaining about this (no newline at end of file). I thought the git hooks were supposed to prevent that.

televandalist: This threw me off because all I did was reverse the change from commit #985 - I didn't add anything to it.

https://github.com/RetroAchievements/RAWeb/blob/d74a02e00e4cd3116682858ad26048b6b3863f51/lib/render/content.php#L129

There's not a new line at the end; there's only a bracket 😭 
wat

storage/app/releases.dist.php
```diff
@@ -113,6 +113,16 @@
                 72, // WASM-4
             ],
         ],
+        Emulators::PCSX2 => [
```
Jamiras, line 116: This constant still isn't defined. Nor do we want to define one. Just use `101`.

storage/app/releases.dist.php
```diff
@@ -113,6 +113,16 @@
...
+            'name' => 'PCSX2',
+            'handle' => 'PCSX2
```
Jamiras, line 118: Missing closing quote

Jamiras (approved):

Jamiras:
Deployment note: the block in `releases.dist.php` has to be copied into `releases.php`.

PR #1152, opened 2022-09-30
Unify releases file & wording

luchaos:


src/Emulators.php
```diff
@@ -33,6 +35,12 @@ abstract class Emulators
 
     public const RAMelonDS = 12;
 
+    // Third Party Standalone
+
+    public const RetroArch = 'retro_arch';
```
Jamiras, line 40: I'd probably leave the underscore out.

Jamiras (approved):

src/Emulators.php
luchaos: Done

PR #1155, opened 2022-09-30
Render default icon for deleted user avatar

luchaos:


lib/render/user.php
```diff
@@ -24,14 +24,15 @@ function userAvatar(
...
         $userSanitized = $username;
         sanitize_outputs($userSanitized);
 
-        return '<del>' . $userSanitized . '</del>';
+        $iconLabel = '';
+        if ($icon !== false && ($icon || !$label)) {
+            $iconLabel = "<img src='" . media_asset('/UserPic/_User.png') . "' alt='$userSanitized' width='$iconSize' height='$iconSize'>";
+        }
+
+        return "<span class='inline whitespace-nowrap'>$iconLabel <del>$userSanitized</del></span>";
```
Jamiras, line 35: I don't think you can implicitly put the user name in the text. Suggest putting it in a `title` attribute.
image

luchaos: Just noticed 👍

lib/render/user.php
```diff
@@ -24,14 +24,20 @@ function userAvatar(
...
 
-        $userSanitized = $username;
-        sanitize_outputs($userSanitized);
+        $usernameLabel = '';
+        if ($label !== false && ($label || !$icon)) {
+            $usernameLabel = "<del>$userSanitized</del>";
+        }
 
-        return '<del>' . $userSanitized . '</del>';
+        return "<span class='inline whitespace-nowrap'>" . $iconLabel . $usernameLabel . "</span>";
```
Jamiras, line 40: Not sure where the extra padding is coming from, but the images aren't aligned.
image

lib/render/user.php
```diff
@@ -24,14 +24,20 @@ function userAvatar(
     });
 
     if (!$user) {
+        $userSanitized = $username;
+        sanitize_outputs($userSanitized);
+
+        $iconLabel = '';
         if ($icon !== false && ($icon || !$label)) {
-            return '';
+            $iconLabel = "<img src='" . media_asset('/UserPic/_User.png') . "' alt='$userSanitized' width='$iconSize' height='$iconSize'>";
```
Jamiras, line 32: Recommend adding a `title` tag, as the image has no meaning without an associated label or tooltip (code notes).

lib/render/user.php
luchaos: Thanks, that didn't occur to me.

lib/render/user.php
luchaos: Good call, added

Jamiras (approved):

PR #1165, opened 2022-10-01
improve error handling of duplicate tickets

Jamiras:
prevents "Duplicate entry '245899-142106' for key 'Ticket.AchievementID_ReportedByUserID'" error when trying to submit a ticket for an achievement that you already have a ticket for.

Updates the "Report issue" button and "Create Ticket" pages to redirect to the existing (or newly created) ticket.

If the ticket already exists, an error message is displayed.
image


Jamiras:
NOTE: There's a unique index on `AchievementID_ReportedByUserID`, which leads to the exception. Ideally, that would be relaxed and the code could be updated to allow creating a new ticket if the previous ticket has already been closed or resolved.

luchaos (approved):

PR #1166, opened 2022-10-01
use loadCard for active players

Jamiras:
Updates the active players callback to use the `loadCard` call for the user. Also updates the game card to match the new style (no parentheses on console name).

Before:
image

After:
image


public/js/activePlayersBootstrap.js
```diff
@@ -72,14 +89,15 @@ var ActivePlayersViewModel = function () {
   };
 
   this.ConvertToObservablePlayer = function (player) {
+    CacheCardDiv('game', player.GameID, player.GameTitle, player.ConsoleName, player.GameIcon);
```
Jamiras, line 92: I'm not that familiar with `ko`. Is this going to be problematic?

luchaos: I don't think so. I'm not sure why those are observables looking at it.

public/js/activePlayersBootstrap.js
```diff
@@ -1,3 +1,20 @@
+function CreateCardIconDiv(type, id, title, icon, url) {
```
luchaos, line 1: `avatar()` would match the backend variant - not important here though.

luchaos (approved):

public/js/activePlayersBootstrap.js
luchaos: Seems to work just fine for knockout (`ko`).

PR #1174, opened 2022-10-02
cache top achievers per game

Jamiras:
eliminates an expensive query on the game page. only cached if there are at least 10 masteries (full list), minimizing the expiration logic to: new mastery, achievement promoted/demoted, and reset achievement.

NOTE: the current user is no longer shown in the mastery list if they are untracked. I believe this is consistent with the messaging around being untracked:
> An untracked account **doesn't appear in the global ranking**, its activities do not appear in the feed, it's **no longer listed among all other earners of each achievement** (on the achievement page) and their entries on specific leaderboards will not be publicly visible.

### DIFF
--- a/lib/database/achievement.php
+++ b/lib/database/achievement.php
@@ -374,6 +374,7 @@ function UploadNewAchievement(
                             $author
                         );
                     }
+                    expireGameTopAchievers($gameID);
                 } else {
                     $fields = [];
                     if ($changingPoints) {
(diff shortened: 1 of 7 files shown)


luchaos (approved):
Wonderful. I had a very similar thing prepared for the Web API but scrapped it eventually as the problem seemed to be with the server's configuration (which turned out to be true). 

I think something like this could still very much be applied to some of the more expensive Web API endpoints.

PR #1198, opened 2022-10-14
fix display issue when embedding game and achievement with same ID in same page

Jamiras:
For example, if a forum linked to `[game=111]` and later in the same forum page, there was a link to `[ach=111]`, the second link would be incorrect.

The internal storage for resolved linked was storing both in a bucket for "game:111". The achievement link should have gone into an "achievement:111" bucket.


luchaos (approved):
🤦 thanks!

PR #1200, opened 2022-10-17
Cleanup

luchaos:


lib/util/trigger.php
```diff
@@ -9,39 +9,129 @@ function parseOperand($mem)
...
-            case 'r': case 'R': $size = 'Bit5'; break;
-            case 's': case 'S': $size = 'Bit6'; break;
-            case 't': case 'T': $size = 'Bit7'; break;
-            case 'l': case 'L': $size = 'Lower4'; break;
-            case 'u': case 'U': $size = 'Upper4'; break;
-            case 'k': case 'K': $size = 'BitCount'; break;
-            case 'w': case 'W': $size = '24-bit'; break;
-            case 'g': case 'G': $size = '32-bit BE'; break;
-            case 'i': case 'I': $size = '16-bit BE'; break;
-            case 'j': case 'J': $size = '24-bit BE'; break;
```
Jamiras, line 39: I prefer the grid look of the before (lowercase in first column, uppercase in second, result in third), but I understand the desire for consistency.

Jamiras (commented):
I haven't actually tested any of these changes, but it looks like it's almost all non-functional.

lib/util/trigger.php
luchaos: Same, I think keeping the column formatting in this particular case would be preferable. I'll check if I can exclude this block from the rules.

luchaos: Restored formatting 👍

Jamiras (approved):

PR #1204, opened 2022-10-24
Fix garbage activity

Jamiras:
Fixes `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '@?=,3804,4986)' at line 3` error when non-numeric data is in the activity table.

image


lib/database/user-activity.php
```diff
@@ -472,7 +477,7 @@ function getRecentlyPlayedGames(string $user, int $offset, int $count, ?array &$
             $recentlyPlayedGameIDs[] = $recentlyPlayedGame['GameID'];
         }
 
-        $recentlyPlayedGameIDs = implode(',', $recentlyPlayedGameIDs);
+        $recentlyPlayedGameIDs = implodeInts(',', $recentlyPlayedGameIDs);
```
luchaos, line 480: Alternative to the custom `implodeInts` function using collections:
```php
$recentlyPlayedGameIDs = collect($recentlyPlayedGameIDs)
    ->filter(fn($id) => is_int($id) || is_numeric($id))
    ->implode(',');
```

luchaos (approved):

PR #1206, opened 2022-10-25
prevent creating account with invalid username

Jamiras:
image

https://discord.com/channels/310192285306454017/310195854642249728/1034405380165468180


public/request/auth/register.php
```diff
@@ -12,6 +12,10 @@
 $pass = $input['password'];
 $email = $input['email'];
 
+if (!isValidUsername($username)) {
```
luchaos, line 15: This seemed redundant at first but I just learned that Laravel's `alpha_num` validation rule doesn't prevent usage of "number-like"s.

I'll try to find a better solution for username validation in a separate PR.

luchaos (approved):

public/request/auth/register.php
Jamiras: It also doesn't prevent non-ASCII letters, which was what originally prompted this PR.

PR #1216, opened 2022-11-06
add explain blocks for leaderboards

Jamiras:
implements #1215 

Adds collapsible sections for start/cancel/submit/value triggers to the leaderboard page (not the manage leaderboards page). Includes support for translating legacy value definitions to Measured strings.

image

image

image

image

### DIFF
--- a/lib/util/trigger.php
+++ b/lib/util/trigger.php
@@ -127,7 +127,11 @@ function parseOperand($mem)
             }
         }
 
-        $value = '0x' . str_pad(dechex((int) substr($mem, 0, $count)), 6, '0', STR_PAD_LEFT);
+        $padded = str_pad(dechex((int) substr($mem, 0, $count)), 6, '0', STR_PAD_LEFT);
+        if (strlen($padded) > 8) {
+            $padded = substr($padded, -8);
+        }
+        $value = '0x' . $padded;
         $mem = substr($mem, $count);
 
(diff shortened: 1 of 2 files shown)


lib/util/trigger.php
```diff
@@ -261,16 +265,16 @@ function getAchievementPatchReadableHTML($mem, $memNotes)
...
-      <th>Flag</th>
-      <th>Type</th>
-      <th>Size</th>
-      <th>Memory</th>
-      <th>Cmp</th>
-      <th>Type</th>
-      <th>Size</th>
-      <th>Mem/Val</th>
-      <th>Hits</th>
+      <th nowrap>ID</th>
```
luchaos, line 268: The `nowrap` html attribute is outdated afaik - alternatively use `class="whitespace-nowrap"`.

Jamiras: I've updated everything to use `class='whitespace-nowrap'`.

Single word cells were definitely being wrapped.
image

luchaos: Oh, interesting. That's odd

luchaos (approved):

lib/util/trigger.php
Jamiras: Here's an extreme example when viewed from a phone:
image

No sure what it'll look like with the changes.

PR #1218, opened 2022-11-08
fix display of images on game compare page

### LINKED ISSUE #1217: Game Compare: other user's achievement icons don't show properly
When comparing your scores to another user, all of the icons for THEIR achievements are always greyed out, even if it shows the time and date when they earned them. Here's a screenshot of what it looks like: 
rabug

Jamiras:
fixes #1217

### DIFF
--- a/lib/database/game.php
+++ b/lib/database/game.php
@@ -34,7 +34,7 @@ function getGameData($gameID): ?array
     }
 }
 
-function getGameTitleFromID($gameID, &$gameTitle, &$consoleID, &$consoleName, &$forumTopicID, &$allData): string
+function getGameTitleFromID($gameID, &$gameTitle, &$consoleID, &$consoleName, &$forumTopicID, &$allData): bool
 {
     sanitize_sql_inputs($gameID);
     settype($gameID, "integer");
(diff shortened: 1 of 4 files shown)


public/gamecompare.php
```diff
@@ -196,7 +196,7 @@
                         $rightSoftcoreAwardedPoints += $achPoints;
                         echo "<small class='smalldate'>$awardedRight</small>";
                     }
-                    echo achievementAvatar($nextAch, label: false, icon: $badgeName . "_lock", iconSize: $iconSize, iconClass: 'awardremote', tooltip: false);
+                    echo achievementAvatar($nextAch, label: false, icon: $badgeName, iconSize: $iconSize, iconClass: isset($awardedHCRight) ? 'goldimage' : '', tooltip: false);
```
Jamiras, line 199: `_lock` was incorrectly added by https://github.com/RetroAchievements/RAWeb/pull/1149/files#diff-1b3f2a3b9e72f81134ba1e00c5c39e9206122eb246ed2d1495c7ea923109d36bL199, but `awardremote` was there before. It doesn't seem to be defined, so I've replaced with the standard gold border for hardcore logic.

luchaos: Thanks! That was an oversight on my end

luchaos (approved):

PR #1221, opened 2022-11-10
Consolidate username validation

luchaos:
Fix register page not remembering old input on validation failure.

### DIFF
--- a/app/Support/Rules/CtypeAlnum.php
+++ b/app/Support/Rules/CtypeAlnum.php
@@ -8,6 +8,7 @@ use Illuminate\Translation\PotentiallyTranslatedString;
 
 /**
  * Stricter version of the alpha_num validation rule
+ * Restricts the input to ASCII characters
  */
 class CtypeAlnum implements InvokableRule
 {
@@ -17,7 +18,7 @@ class CtypeAlnum implements InvokableRule
     public function __invoke($attribute, mixed $value, $fail): void
     {
         if (!ctype_alnum($value)) {
-            $fail('validation.alpha_num')->translate();
+            $fail('validation.ctype_alnum')->translate();
         }
     }
 }
--- a/app/Support/Rules/CtypeAlnum.php
+++ b/app/Support/Rules/CtypeAlnum.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace App\Support\Rules;
+
+use Closure;
+use Illuminate\Contracts\Validation\InvokableRule;
+use Illuminate\Translation\PotentiallyTranslatedString;
+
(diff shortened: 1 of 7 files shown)


lib/database/ticket.php
```diff
@@ -392,8 +392,9 @@ function updateTicket($user, $ticketID, $ticketVal, $reason = null): bool
 
     addArticleComment("Server", ArticleType::AchievementTicket, $ticketID, $comment, $user);
 
-    if (!getAccountDetails($userReporter, $reporterData))
+    if (!getAccountDetails($userReporter, $reporterData)) {
         return true;
+    }
```
luchaos, line 397: The `braces` CS fixer rule would've demanded curly braces here.
The rule is disabled however because the indentations in php files in the public directory would be messed up because of it.
It will be enabled as soon as the files in the public directory have been refactored into controllers and views/components.

public/createaccount.php
```diff
@@ -20,19 +20,19 @@
...
                         <td>
-                            <div><input type="text" id="email" name='email' value="<?= old('e') ?>"></div>
+                            <div><input type="text" id="email" name='email' value="<?= old('email') ?>"></div>
                         </td>
                     </tr>
                     <tr>
                         <td class="text-right"><label for="email_confirmation">Confirm Email</label></td>
                         <td>
-                            <div><input type="text" id="email_confirmation" name='email_confirmation' value="<?= old('e_confirmation') ?>"></div>
+                            <div><input type="text" id="email_confirmation" name='email_confirmation' value="<?= old('email_confirmation') ?>"></div>
```
luchaos, line 35: Missed those during the v2 refactoring.

app/Support/Rules/CtypeAlnum.php
```diff
@@ -0,0 +1,20 @@
...
+
+class CtypeAlnum implements InvokableRule
+{
+    /**
+     * @param Closure(string): PotentiallyTranslatedString $fail
+     */
+    public function __invoke($attribute, mixed $value, $fail): void
+    {
+        if (!ctype_alnum($value)) {
+            $fail('validation.ctype_alnum')->translate();
```
luchaos, line 17: Rereading the changes I think it's not necessary to add a custom validation message; reuse `validation.alpha_num` instead?

Jamiras: Makes sense. Unless you want to expand on the message:
`The :attribute must contain only unaccented letters and numbers`

Might also want to include a comment indicating that this differs from the built-in `alpha_num` validation in that it restricts the input to ASCII characters.

PR #1248, opened 2022-12-01
include calling function location in silenced sql failure

Jamiras:
To help track down errors like `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-8, 15' at line 6`

Alternately, we could just throw the exception in production (see line 74) and let the framework log it normally.

### DIFF
--- a/lib/util/database.php
+++ b/lib/util/database.php
@@ -73,6 +73,10 @@ function log_sql_fail(): void
         if (config('app.debug')) {
             throw new Exception($error);
         }
+
+        $calling_function = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1)[0];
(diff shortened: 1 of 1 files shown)


luchaos:
Now I wonder why throwing the exception wasn't the default 🤔 

Would throwing the exception, by removing the debug env var check, serve us as much useful information as your change would?

I'm intrigued to try your solution otherwise.

Jamiras:
I wouldn't say throwing the exception is any more useful. It's just more verbose. Here's two log entries - the first is non-exception, the second it exception (note the function args won't appear in the stack trace on production):

image

luchaos (approved):
As you mentioned, throwing an exception would be appropriate yet it might have side effects in certain places that are hard to find without tests.

Heads up: Laravel would/will throw an exception for any db error, too.

PR #1255, opened 2022-12-05
only fetch tooltip if hovering

Jamiras:
Requires the cursor to be over a image/link for 200ms before fetching the tooltip.

Should minimize the occurrences of the 429 error when scrolling through a page or just moving the cursor around.

### DIFF
--- a/public/js/all.js
+++ b/public/js/all.js
@@ -139,7 +139,7 @@ function useCard(type, id, context = null, html = '') {
   return html;
 }
 
-function loadCard(owner, type, id, context = null) {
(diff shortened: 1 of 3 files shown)


public/js/all.js
```diff
@@ -139,7 +139,7 @@ function useCard(type, id, context = null, html = '') {
   return html;
 }
 
-function loadCard(type, id, context = null) {
+function loadCard(owner, type, id, context = null) {
```
luchaos, line 142: I'd go with `target` or `element` to refer to the reference HTML element.

luchaos (approved):

PR #1257, opened 2022-12-08
adjust contribution tier thresholds

Jamiras:
Implements #775. 

New tier levels (with badges provided by GameZone):

**Achievements Awarded to Others**:
image

**Points Awarded to Others**:
image

Includes a pre-deployment database migration step to recalculate all contribution points for existing users (builds on #1228) and reallocates the badges for each developer. This took about 90 minutes on my VM. I expect it to take less than 15 minutes on production. During the migration, developers may see their contribution points change and their contribution badges downgraded to line up with the new tiers, but the old image/tooltip will still be displayed.

With the recalculated contributions, our highest contributors are in the 1M range for achievements, and 8M range for points.  The last new tiers are at least 5M and 50M respectively, so there's still badges for those users to earn.

As a side note, `php artisan migrate` appears to be commented out in `deploy.sh`. Is that intentional?


Krylan:
Didn't check the code, but in case of artwork, I would have two remarks:
– this border... it doesn't really fit. Even its colors are not matching well with all badge colors. I would be for something more subtle, or for a border suiting with color to each type/level of the badge (so it could be some transparent border with some effects to make it concave/convex).
– noticed, that first red ghost (2500000 points) has different eyes color than the rest. I don't think it was on purpose.

database/migrations/legacy/2022_12_06_000000_adjust_contrib_badges.php
```diff
@@ -0,0 +1,108 @@
+<?php
```
luchaos, line 1: Migrations are usually meant to be "just" schema migrations - personally I'd implement migrations for existing data like calculation or moving data around as commands. 

Comparable to the sync commands, this would keep migrations fast (think running migrations during testing where this data migration is not relevant) and lets more time intensive tasks to be run on-demand and repeatedly.

There are a few commands in the `ra:` namespace already: `php artisan list | grep "ra:"`

Could be something like `ra:server:developer:recalc-contrib-yield` in this case?

luchaos (approved):

PR #1262, opened 2022-12-12
Set up rate limits

### LINKED ISSUE #1055: Achievement Inspector - Sorting Order Bug

### LINKED ISSUE #1227: Too Many Requests error when switching pages
**Describe the bug**
When switching pages quickly, such as jumping around different hubs, shows a Too Many Requests error red bar at the top.
Screenshot_19

**To Reproduce**

just switch pages quickly in RA.

luchaos:
Fix #1227
Fix #1055

Allow to control rate limits in the application' route service providers.

Requires some additional adjustments in the webserver configuration.

### DIFF
--- a/app/Api/RouteServiceProvider.php
+++ b/app/Api/RouteServiceProvider.php
@@ -8,7 +8,10 @@ use App\Api\Controllers\WebApiController;
 use App\Api\Controllers\WebApiV1Controller;
 use App\Api\Middleware\LogApiUsage;
 use App\Support\Http\HandlesPublicFileRequests;
+use Illuminate\Cache\RateLimiting\Limit;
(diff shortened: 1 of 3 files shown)


luchaos:
Closing in favor of having updated the nginx rate limiting for connect api from 3 requests per second to 10 requests per second.

PR #1265, opened 2022-12-18
Change way of breaking text in forum post content preview

### LINKED ISSUE #1224: Forum Activity on front page does not truncate body of posts with long single word causing View button to not display
**Description**
In a situation wherein one of the first words is extremely long in a forum posting, it will not truncate by character count, causing the text to extend the length of its container and the view button to fail to display.

**Repro**
1. Make a forum post that starts with a very long word (for example, a link)
2. Go to the home page

**Expected behavior**
Impacted text is truncated

**Screenshots**
image

 - OS: Win10
 - Browser: Edge/Chrome
 - Version: 107.0.1418.35 (Official build) (64-bit)

Krylan:
This PR resolves #1224 issue.

At first, I just added one of the Tailwind's utility class, "break-all", to div containing post content for forum activity preview.
The problem with it, though, is that normal words are also broken this way. I was searching for other way of breaking words, and got a solution, which breaks words only when it's necessary (word is longer than whole line, otherwise just wrap to the next line).
```
overflow-wrap: anywhere;
word-break: normal;
```
Wanted to get this effect with Tailwind's classes (https://tailwindcss.com/docs/word-break), but none of them provide this combo: `break-normal` doesn't breakes long words, `break-words` as well, `break-all` breakes also normal words.
I felt, that aforementioned combo works good, because it wraps words not fit into the line, while breaks only when it is necessary – so in our cases of trolling/spam (?) or posting a link.

I created a new class `.text-overflow-wrap` on typography.css – I guess it suits there.

### DIFF
--- a/resources/css/typography.css
+++ b/resources/css/typography.css
@@ -32,7 +32,7 @@ pre, code, .code {
   white-space: pre-wrap;
 }
 
-.text-overflow-wrap{
+.text-overflow-wrap {
   overflow-wrap: anywhere;
   word-break: normal;
 }
--- a/resources/css/typography.css
+++ b/resources/css/typography.css
@@ -31,3 +31,8 @@ pre, code, .code {
   word-break: break-word;
   white-space: pre-wrap;
 }
+
+.text-overflow-wrap{
+  overflow-wrap: anywhere;
+  word-break: normal;
+}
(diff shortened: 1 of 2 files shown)


resources/css/typography.css
```diff
@@ -31,3 +31,8 @@ pre, code, .code {
   word-break: break-word;
   white-space: pre-wrap;
 }
+
+.text-overflow-wrap{
```
Jamiras, line 35: Based on the rest of the file, I believe a space should be present between `wrap` and `{`.

Jamiras (commented):
Verified the functionality fixes the reported issue. As it's a styling change, I'll leave approval to luchaos.

resources/css/typography.css
Krylan: You are right, it was a small oversight from my side 😅 
Now it's fixed.

luchaos (approved):

PR #1271, opened 2022-12-20
Award counters: count hidden awards + CSS cleanup

EmoonX:
This PR changes the award counters on profiles so those account for awards the player has hidden (index -1). The actual hidden amount is specified on hover:

awards-hidden

An user requested it some time ago, and the reasoning is: the majority of people hide awards primarily for tidiness. It's not like the actual number is a secret anyway; plus it can be found by other means. A further possibility is having a setting where players can opt for counting the hidden ones or not, but I'm leaving it as-is for now.

Furthermore, did some visual cleanup, as the CSS had been broken since the V2 update.

new-visual


lib/render/site-award.php
```diff
@@ -66,53 +66,61 @@ function RenderSiteAwards($userAwards): void
 function RenderAwardGroup($awards, $title): void
 {
     $numItems = is_countable($awards) ? count($awards) : 0;
-    if ($numItems == 0) {
+    $numHidden = 0;
+    foreach ($awards as $award) {
+        if ($award['DisplayOrder'] != -1) break;
+        $numHidden++;
```
Jamiras, line 72: Any negative value can hide the items.

Also, the break here doesn't really provide any benefits. Recommend inverting the logic:
```
if ($award['DisplayOrder'] < 0) {
    $numHidden++;
}
```

lib/render/site-award.php
```diff
@@ -66,53 +66,61 @@ function RenderSiteAwards($userAwards): void
...
         // Count and show # of completed/mastered games
-        $numGamesCompleted = 0;
+        [$numCompleted, $numCompletedHidden] = [0, 0];
         foreach ($awards as $award) {
             if ($award['AwardDataExtra'] != 1) {
-                $numGamesCompleted++;
+                $numCompleted++;
+                if ($award['DisplayOrder'] == -1) {
+                    $numCompletedHidden++;
+                }
```
Jamiras, line 92: Same comment here. Any negative value should hide the item.

lib/render/site-award.php
```diff
@@ -128,6 +136,20 @@ function RenderAwardGroup($awards, $title): void
     echo "</div>";
 }
 
+function RenderCounter($icon, $text, $numItems, $numHidden): string
+{
+    $tooltip = $icon . "  " . $numItems . " " . $text;
+    if ($numHidden > 0) {
+        $tooltip .= "【" . $numHidden . " hidden】";
```
Jamiras, line 143: Recommend unstylized parentheses.

lib/render/site-award.php
```diff
@@ -128,6 +136,20 @@ function RenderAwardGroup($awards, $title): void
     echo "</div>";
 }
 
+function RenderCounter($icon, $text, $numItems, $numHidden): string
+{
+    $tooltip = $icon . "  " . $numItems . " " . $text;
```
Jamiras, line 141: Personally, I don't think the icon belongs in the tooltip. It's redundant with the icon that the user hovered over to get the tooltip.

lib/render/site-award.php
```diff
@@ -66,53 +66,61 @@ function RenderSiteAwards($userAwards): void
...
+        $numMasteredHidden = $numHidden - $numCompletedHidden;
         $counters = "";
-        if ($numGamesMastered > 0) {
+        if ($numMastered > 0) {
             $icon = mb_substr($icons[$title], 0, 1);
-            $counters .= "
-                <div class='awardcount' title='# of mastered games'>
-                    <span class='icon'>$icon</span><span class='numitems'>$numGamesMastered</span>
-                </div>";
+            $text = ($numMastered > 1 ? "games" : "game") . " MASTERED";
```
Jamiras, line 100: Recommend leaving MASTERED as lowercase

lib/render/site-award.php
```diff
@@ -66,53 +66,61 @@ function RenderSiteAwards($userAwards): void
...
+        }
+        $counters = RenderCounter($icon, $text, $numItems, $numHidden);
     }
 
     echo "<div id='" . strtolower(str_replace(' ', '', $title)) . "'>";
     echo "<h3>$title $counters</h3>";
     echo "<div class='component flex flex-wrap justify-between gap-2'>";
     $imageSize = 48;
     $numCols = 5;
+    $numItems -= $numHidden;
```
Jamiras, line 123: Hidden items aren't actually being hidden:
image

image

Jamiras (changes requested):
Two major points:
* Any negative value should be hidden
* Hidden items aren't actually being hidden anymore

Also left several suggestions.

lib/render/site-award.php
EmoonX: The break was just a minor optimization since `DisplayOrder` is in increasing order. But indeed, the latter way is cleaner.

lib/render/site-award.php
EmoonX: I'm changing it to [square brackets] then, if that's okay with you.

luchaos: I'm for normal paranthesis ;)

EmoonX: So 2-1 for normal parenthesis then 🗳️

EmoonX:
Yeah, I wasn't aware that any negative value worked as hidden. The UI description explicitly states `-1` though; not sure if anybody wants to change that:
https://github.com/RetroAchievements/RAWeb/blob/3fe4650c2856644511f8981d6e66488c024f3b58/public/reorderSiteAwards.php#L33-L34

All in all, hidden items should appear as hidden again now. I've probably made a regression when updating the code and forgot to test it later.

Jamiras (approved):

lib/render/site-award.php
```diff
@@ -66,46 +66,54 @@ function RenderSiteAwards($userAwards): void
...
+    if ($numItems == $numHidden) {
+        // No items to show
         return;
     }
 
     $icons = [
         "Game Awards" => "👑🎖️",
         "Event Awards" => "🌱",
-        "Site Awards" => "⬩",
+        "Site Awards" => "🌐",
```
luchaos, line 83: I'd like to replace all those emojis with svg icons down the line for visual consistency's sake.

luchaos (approved):

PR #1274, opened 2022-12-21
Stylized categories/subsets in game titles

EmoonX:
As of now, special categories are directly imprinted in game titles (`~Hack~`, `~Homebrew~`), which are more like placeholders and not exactly pretty when displayed raw. This PR parses the titles' text, wraps the category string and applies CSS to it, with specific colors for each one.

homebrew
unlicensed
all-categories

In a similar way, the same is done for subsets, although with a different styling. Red is reserved for `Bonus`, blue for `Multi`; and otherwise orange for all the custom ones.

subset-bonus
contra-subsets

I've applied the render in most of the website places (i.e game pages, achievement pages, game lists, game cards...); though I'm not sure if all the relevant corners were covered. Some instances were intentionally left out (e.g the navpath bar, as pure text fits better there).

Eventually, I imagine such categories will be stored in DB using dedicated fields. Even then, whenever that happens the only needed change would be discarding the regex parsing; so everything's good.

I sought to achieve an unobtrusive and fitting color scheme (I hope it worked?); nevertheless, feel free to suggest changes.


Jamiras:
Not sure if there's an easy fix, but the decorators are not aligned with the text.

Actual:
image

Preferred:
image

Jamiras:
Might want to choose a different color for Subset-Multi. It's barely noticable in the game page title:
image

Jamiras:
Only place I found where the change was missed is the achievement tooltip:
image

And mastery tooltip (though that's just a simple text tooltip currently):
image

lib/render/avatar.php
```diff
@@ -11,11 +11,13 @@ function avatar(
...
     $iconClass = 'badgeimg',
     ?string $context = null,
+    bool $sanitize = true,
+    ?string $altText = null,
 ): string {
-    $escapedName = attributeEscape($label);
-    sanitize_outputs(
-        $label,
-    );
+    $escapedName = attributeEscape($altText);
```
Jamiras, line 17: `altText` defaults to null. this should probably be `$altText ?? $label`

lib/render/game.php
```diff
@@ -41,9 +45,38 @@ function gameAvatar(
         iconSize: $iconSize,
         iconClass: $iconClass,
         context: $context,
+        sanitize: $labelIsHtml === false,
```
Jamiras, line 48: The point of sanitization is to escape HTML. It seems like this effectively circumvents that protection. Did you encounter a case where this was problematic? The `~Category~` tags shouldn't be affected by the sanitization process.

luchaos:
Nice intermediate solution until we have tags.

Even though i enjoy the colors I'd ask you to drop them altogether for now. There's no guarantee that all of them will work nicely with the current themes and it makes it harder to introduce the tags system down the line; something grayscale or using the chosen theme.

When that's done we can revisit colors for tags.

resources/css/game-category.css
```diff
@@ -0,0 +1,48 @@
+.game-category {
```
luchaos, line 1: generalize as `.label` to be reusabe in other areas.
or `.achievement-set-category` to match the upcoming schema changes

resources/css/app.css
```diff
@@ -40,6 +40,7 @@
 @import "award-count.css";
 @import "comment.css";
 @import "devbox.css";
+@import "game-category.css";
```
luchaos, line 43: See other comment (label.css, tag.css)

lib/render/game.php
EmoonX: The actual `$label` that gets passed in `gameAvatar(...)`'s call to `avatar(...)` is the already-rendered game title (e.g `<span>~Category~</span> Lorem Ipsum`), so sanitization can't happen in this case. But indeed, suspending it every time there's HTML doesn't make sense either. A simple solution is to just sanitize the title _before_ generating the HTML.

resources/css/game-category.css
EmoonX: Added a generic `.tag` (in `tag.css`) and an `achievement-set.css` file with specific style for categories and subsets.

EmoonX:
> Not sure if there's an easy fix, but the decorators are not aligned with the text.
> 
> Actual: image
> 
> Preferred: image

I think I managed to amend it with some hocus pocus. It's browser-dependent; seems to line up fine on Chrome but still not perfectly on Firefox. Ugh, aligning text with CSS sucks...

@luchaos So minimalist it is. I've reduced all of it to a single style for categories and another for subsets. {White text + dark background + lighter outline} produces a good result both in dark and light themes. Hope this is fine for you.

new-styles

Blazekickn:
i think the version with the colors looks way better, it's much easier to tell between the types of games at a glance. I don't think they would clash with any of our existing themes either, maybe unlicensed on black theme or prototype on white theme but i think the colors are different enough. I don't see why we can't have them.

Jamiras:
In keeping with the minimalist theme, it seems like subsets should have the same coloring.

Also, there's a minor overlapping issue in tooltips for subsets with other modifiers (which might be less noticeable if subset shares the same color):
image

Jamiras (approved):
Seems fine functionally. Will let @luchaos have the final say on colors.

EmoonX:
>In keeping with the minimalist theme, it seems like subsets should have the same coloring.

You mean using a single style for all tags? As I see it, they have different purposes:
- category tags are more like "flavors" to describe games (derived or not, and there can be more than one tag at once)
- subsets tags, on the other hand, designate alternate cheevo content for the base game (aka subsets), which can only exist as a branch of the main set

Having a different style for both serves as a _visual cue_ so users can quickly tell apart they represent distinct concepts (especially useful for newcomers).

>Also, there's a minor overlapping issue in tooltips for subsets with other modifiers

Those game tooltips could use some CSS overhaul. I guess this issue is better left for a PR which improves the tooltips' styling as a whole.

luchaos (approved):
Let's leave it at that for now 👍

PR #1277, opened 2022-12-25
explicitly return error to prevent logging exception

Jamiras:
prevents logging of the following errors (while still reporting them to the client):

> Invalid file type {"userId":354831,"exception":"[object] (Exception(code: 0): Invalid file type at /srv/web/releases/2022-12-22T225652-2.4.1-ae5bf5e/lib/util/media.php:36)

>imagecreatefromjpeg(): gd-jpeg: JPEG library reports unrecoverable error: Not a JPEG file: starts with 0x52 0x49 {"userId":352525,"exception":"[object] (ErrorException(code: 0): imagecreatefromjpeg(): gd-jpeg: JPEG library reports unrecoverable error: Not a JPEG file: starts with 0x52 0x49 at /srv/web/releases/2022-12-22T225652-2.4.1-ae5bf5e/lib/util/media.php:53)

> imagecreatefrompng(): &quot;/tmp/data-url3Fhk3J&quot; is not a valid PNG file {"userId":354508,"exception":"[object] (ErrorException(code: 0): imagecreatefrompng(): &quot;/tmp/data-url3Fhk3J&quot; is not a valid PNG file at /srv/web/releases/2022-12-22T225652-2.4.1-ae5bf5e/lib/util/media.php:52)

### DIFF
--- a/public/request/user/update-avatar.php
+++ b/public/request/user/update-avatar.php
@@ -6,6 +6,10 @@ if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::Regi
     return back()->withErrors(__('legacy.error.permissions'));
 }
 
-UploadAvatar($user, request()->post('imageData'));
+try {
+    UploadAvatar($user, request()->post('imageData'));
+} catch (Exception $ex) {
+    return response()->json(['message' => $ex->getMessage()], 400);
+}
 
 return response()->json(['message' => __('legacy.success.ok')]);
--- a/public/request/user/update-avatar.php
+++ b/public/request/user/update-avatar.php
@@ -1,5 +1,6 @@
 <?php
 
+use Illuminate\Support\Facades\Log;
 use RA\Permissions;
 
 if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::Registered)) {
@@ -9,7 +10,17 @@ if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::Regi
 try {
     UploadAvatar($user, request()->post('imageData'));
 } catch (Exception $ex) {
-    return response()->json(['message' => $ex->getMessage()], 400);
+    $error = $ex->getMessage();
+    if ($error == 'Invalid file type' || $error == 'File too large') {
+        return response()->json(['message' => $error], 400);
+    }
+
+    if (preg_match('/(not a .* file)/i', $ex->getMessage(), $match)) {
+        return response()->json(['message' => ucfirst($match[0])], 400);
+    }
+
+    Log::error($ex);
+    abort(500);
 }
 
 return response()->json(['message' => __('legacy.success.ok')]);


public/request/user/update-avatar.php
```diff
@@ -6,6 +6,10 @@
     return back()->withErrors(__('legacy.error.permissions'));
 }
 
-UploadAvatar($user, request()->post('imageData'));
+try {
+    UploadAvatar($user, request()->post('imageData'));
+} catch (Exception $ex) {
+    return response()->json(['message' => $ex->getMessage()], 400);
```
luchaos, line 12: Very nice, missed that when refactoring/integrating the v2 prototype.

To not expose internals of the server we might want to respond with a generic error message in the frontend instead, similar to the [news image update request](https://github.com/RetroAchievements/RAWeb/blob/master/public/request/news/update-image.php#L14-L18)?

```php
try {
    $imagePath = UploadNewsImage($input['image']);
} catch (Exception) {
    abort(400);
}
```

luchaos (changes requested):

public/request/user/update-avatar.php
Jamiras: We need to return some sort of error message or the red bar shows up blank. The actual error returned to the client is just the exception message and doesn't contain the stack trace. The errors listed above came from the error log.

image

All of these come from uploading images with the wrong extension (either unsupported or a png file that's really a bmp). I think "not a valid PNG file" and "Invalid file type" are valid error messages to display to the user. I would prefer a solution where those weren't exceptions, but I'm just working with what's available.

luchaos: Using [this bitmap file](https://www.w3.org/People/mimasa/test/imgformat/img/w3c_home.bmp) and turning off `APP_DEBUG` in my .env file i do get valid "Server Error" responses even when renamed to be a png file. They do get logged though as mentioned.

Using `abort(400)` will show "Bad Request" in the red bar instead and not log it.

Jamiras: While most of the word soup in the error message is unimportant for the user, they might be able to figure out what they did wrong if they see the "Not a JPEG file" portion.

Additionally, "Invalid file type" and "File too large" are errors the user could probably correct. Saying "Bad Request" is no better than just saying "Failed".

luchaos: > Saying "Bad Request" is no better than just saying "Failed".

While I'm fine with that I can see what you're aiming for.

The underlying problem for me is how this upload works in the first place. JavaScript is base64-serializing image data and sending it to the server. This should be be an actual file upload. Only then would the server be allowed to validate the data properly and inspect file type and size.

The user does not gain much from the error message as the file was invalid in the first place. Most valid file types still work as intended. Even exposing which functions are used to process input data is too much exposure imo.

Do we have examples where the server responded with errors even though the uploaded image data was valid?

luchaos: Meaning that I'd rather have it logged than served to the user.

Usually this means that there has to be better validation to tell the user what exatly went wrong; what's wrong with the data they provided.

This would be the only place where we tell a user what went wrong internally. I'd rather try to hide the underlying system and improve the interface instead.

Jamiras: All three of the log messages captured in the initial description can be easily reproduced and potentially corrected by the user.

> Invalid file type {"userId":354831,"exception":"[object] (Exception(code: 0): Invalid file type at /srv/web/releases/2022-12-22T225652-2.4.1-ae5bf5e/lib/util/media.php:36)

Try to upload a bmp file.

> imagecreatefromjpeg(): gd-jpeg: JPEG library reports unrecoverable error: Not a JPEG file: starts with 0x52 0x49 {"userId":352525,"exception":"[object] (ErrorException(code: 0): imagecreatefromjpeg(): gd-jpeg: JPEG library reports unrecoverable error: Not a JPEG file: starts with 0x52 0x49 at /srv/web/releases/2022-12-22T225652-2.4.1-ae5bf5e/lib/util/media.php:53)

Change the extension of a png file to be jpg. Try to upload

> imagecreatefrompng(): "/tmp/data-url3Fhk3J" is not a valid PNG file {"userId":354508,"exception":"[object] (ErrorException(code: 0): imagecreatefrompng(): "/tmp/data-url3Fhk3J" is not a valid PNG file at /srv/web/releases/2022-12-22T225652-2.4.1-ae5bf5e/lib/util/media.php:52)

Change the extension of a jpg file to be png. Try to upload

In all three of these cases, the exception would be handled by the framework and reported to the user after logging. They're not actually server problems, so I don't want them appearing in the error log. My initial changes simply returned the error message (as they would have seen before) without logging.

I have made further changes based on your feedback to only return the error message for three specific errors:
1) Invalid file type - user tried to upload a file with an unsupported extension
2) File too large - user tried to upload a super huge image
3) Not a XXX file - user tried to upload a file with a supported extension but the decoder couldn't decode it.

All other errors will be logged and returned to the client simply as "Internal server error".

luchaos: > They're not actually server problems, so I don't want them appearing in the error log

Much agreed; those are good changes to better serve the users' needs. I think we'll be able to improve and simplify how asynchronous media uploads work in the future based on that. I'll probably rewrite most of it to basic file uploads so the framework's validation rules can be applied more easily.

This works for me.

luchaos (approved):

PR #1280, opened 2022-12-28
add user game activity page

Jamiras:
Mimics the Game Mastery calculations provided by RATools. 

Currently limiting access to Administrators as the queries against the Activity table are expensive (which is why we lost the feed long ago). Using the Activity table provides more data (including 'Started Playing' timestamps for more accurately determining the distance to the first unlocked achievement for a session, and softcore unlocks obscured by a later hardcore unlock).

Adds an Admin section to the game compare page:
image

Clicking the `View User Game Activity` link goes to a new page that groups achievement unlocks into sessions and summarizes the estimated playtime for the user.

image

If only one session is captured, the time doesn't have to be estimated.

image


public/usergameactivity.php
```diff
@@ -0,0 +1,106 @@
...
+
+use RA\AchievementType;
+use RA\Permissions;
+
+if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::Admin)) {
+    abort(401);
+}
+
+$gameID = requestInputSanitized('ID', null, 'integer');
+$user2 = requestInputSanitized('f');
```
luchaos, line 11: I'd probably go with `$targetUser`, or `$targetUsername` even, instead of `$user2`.

The parameter can be `u` instead of `f` as it's not used here and seems more fitting?

```php
$gameID = (int) request()->input('ID');
$targetUser = request()->input('u');
```

lib/database/user-activity.php
```diff
@@ -711,3 +712,210 @@ function GetMostPopularTitles($daysRange = 7, $offset = 0, $count = 10): array
...
+                }
+                $lastAchievementTime = $achievement['When'];
+            }
+        }
+    }
+
+    // assume every achievement took roughly the same amount of time to earn. divide the
+    // user's total known playtime by the number of achievements they've earned to get the
+    // approximate time per achievement earned. add this value to each session to account
+    // for time played after getting the last achievement of the session.
```
luchaos, line 815: That's a fitting approach! Very cool

luchaos (approved):

public/usergameactivity.php
luchaos: Heads up: eventually the route might be mapped to something like `user/{username}/game/{gameId}/activity`. Improving the routing & redirecting legacy URLs to pretty URLs is on my list for after the mysqli refactoring.

Jamiras: That was just copy/pasted from gamecompare.php:
https://github.com/RetroAchievements/RAWeb/blob/564e3bc7b1eeb1407a234a401894b4bed936bad9/public/gamecompare.php#L9-L10

luchaos: That's what I assumed :) Can stay as is imo.

lib/util/date.php
```diff
@@ -1,5 +1,10 @@
 <?php
 
+function formatHMS(int $seconds): string
+{
+    return sprintf("%d:%02d:%02d", $seconds / 3600, ($seconds / 60) % 60, $seconds % 60);
```
luchaos, line 5: Haven't tested it but Carbon could do that:

```php
CarbonInterval::seconds($seconds)->cascade()->forHumans();
```

luchaos (approved):

lib/util/date.php
Jamiras: That particular command generates something too verbose for the placement in the grid:

image

luchaos: Checking the docs it seem `->forHumans(['short' => true])` can make it less verbose.

However, sticking with our own formatting is just fine imo. Just wanted to point out that carbon can do a lot of those things :)

PR #1281, opened 2022-12-28
Completion/mastery icons for progress bars

EmoonX:
Simple emoji icons for when a player has 100% completion in a set.

-  `Completion Progress` list:

completion-progress

- Game pages:

completion
mastery


Jamiras:
I think the icon would look better placed next to the text below the bar:
image

image

Also, the completed icon shows up for games with no achievements:
image

lib/render/game.php
```diff
@@ -288,3 +290,31 @@ function RenderGameProgress(int $numAchievements, int $numEarnedCasual, int $num
...
+    float|string $hardcoreRatio,
+    bool $tooltip = false
+): string {
+    if ($awardedCount < $totalCount) {
+        return "<div class='completion-icon'></div>";
+    }
+    $icon = $hardcoreRatio == 100.0 ? '👑' : '🎖️';
+    $class = 'completion-icon active';
+    if ($tooltip) {
+        $tooltip = $hardcoreRatio == 100.0 ? 'Mastered (hardcore)' : 'Completed';
```
Jamiras, line 314: This is repurposing the `$tooltip` variable. Suggest an alternate variable name (i.e. `$tooltipText`)

EmoonX: I've seen this (bad) practice somewhere else in the code. Good to know is isn't recommended, then.

EmoonX:
After a lot of toying around, I think I've managed to arrive at something even better.

enhanced-icons

Bar fills up to the right and "pops" the icon upon reaching the max. Gives that "mission accomplished" touch, I hope. Furthermore, if/whenever those proposed _set tiers_ come to life, the idea can be further enhanced by diving the bar and adding more bubbles.

luchaos:
> After a lot of toying around, I think I've managed to arrive at something even better.

this looks great!

Jamiras:
Looks great when the bar is full or almost full, but when it's empty or almost empty, the border on the bar seems to stand out against not having a border on the circle. It's more noticeable in the recently played list.
image

Also, progress seems to max out at about 90% the game page:
image
image

image
image

EmoonX:
Inverted border/background color for empty icons and added an outline so it stands better against the rest. Ended up doing something similar for all icons.

icons-v3

Positioning and bar length should be displayed correctly everywhere now.

Jamiras:
I honestly like the [larger circles](https://github.com/RetroAchievements/RAWeb/pull/1281#issuecomment-1367690830) better, even if it does have the weird border issue. These icons are just too small:
image

Verified the fix on the game page.

EmoonX:
Getting everything right is a chore. Depends on browser, OS, zoom level... Honestly, the font being "too small" seems to be a site-wide issue for me; that's why I browse pages at 125% zoom here. Thus the emoji can look strange depending on each user's display settings. Perhaps it can become less of an issue if/when those are replaced by png icons.

For now, this is the result (also decided to move labels to the left so those center in the bar itself and not the whole thing):

image

luchaos:
> Perhaps it can become less of an issue if/when those are replaced by png icons.

Eventually there will be svg icons. I've started migrating a lot of view rendering code to blade templates already which will make it easy to use a wide variety of existing icons: https://blade-ui-kit.com/blade-icons

Until then the emojis can stay - they aren't aligned well vertically for me atm (Chrome/macOS) but as you said that might be tricky to get right across browsers.

resources/css/progress.css
```diff
@@ -25,22 +53,41 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 19px;
+  width: 19px;
+  border: solid 2px;
+  border-radius: 50%;
+  outline: solid 2px;
+  font-family: emoji;
+  font-size: 7pt;
```
luchaos, line 69: Use of `pt` here instead of `px/em/rem` is intentional?

luchaos (approved):

resources/css/progress.css
EmoonX: I've arrived at it after some trial-and-error. `9px` or `10px` produces similar results.

luchaos: Gotcha. Personally I'd prefer sticking to `px` for consistency but it doesn't matter too much in this case as the icons will most likely be replaced anyways.

Jamiras (approved):
The alignment still seems a bit off on the game page:
image

vs the list:
image

You've commented several times that getting the alignment right has been difficult. Maybe we can merge this as-is and play with that more later.

EmoonX:
Lesson learned: working with emoji is a mess. Client-dependent fonts, anti-aliasing, unpredictable subpixel rendering... They're fine when used alongside text, but as icons or figures... much better to count on the reliability of `svg` or `png` images instead.

PR #1282, opened 2022-12-28
Better ordering for `Completion Progress` list

EmoonX:
As of now, the set ordering for the aforementioned list doesn't make a distinction between hardcore/softcore-earned cheevos, something especially in disagreement with the new point system (where softcore ones now have essentially value 0). This PR introduces a new ordering:

1. Mastered sets
2. Completed sets
5. Incomplete sets, in the following order:
    1. Hardcore %
    2. Overall % (hardcore + softcore)
    3. Number of cheevos in set

ordering

(which means e.g softcore-only sets will appear at the bottom of the list)

softcore-only

As denoted, non-mastered completed sets (regardless of hardcore ratio) are nevertheless displayed _above_ any incomplete ones; the rationale being that the former still give a badge and people overall value those more than raw points in uncompleted games.

Code-wise, changes were done by modifying the DB query. There was quite a bunch of code that seemingly has no use anymore, so I removed it. Apologies in advance in case it was something important 🫣


lib/database/player-game.php
```diff
@@ -275,17 +275,21 @@ function getUsersCompletedGamesAndMax($user): array
...
         LEFT JOIN
             ( SELECT COUNT(*) AS MaxPossible, ach1.GameID FROM Achievements AS ach1 WHERE Flags = $requiredFlags GROUP BY GameID )
             AS inner1 ON inner1.GameID = ach.GameID AND inner1.MaxPossible > $minAchievementsForCompletion
         LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
-        WHERE aw.User='$user' AND ach.Flags = $requiredFlags
-        GROUP BY ach.GameID, aw.HardcoreMode, gd.Title
-        ORDER BY PctWon DESC, inner1.MaxPossible DESC, gd.Title ";
+        WHERE aw.User = '$user' AND ach.Flags = $requiredFlags
+        GROUP BY ach.GameID, gd.Title
+        ORDER BY PctWon = 1.0 DESC, PctWonHC DESC, PctWon DESC, inner1.MaxPossible DESC, gd.Title";
```
Jamiras, line 292: Something about this syntax is not supported by the xampp enviroment.
```
MariaDB [retroachievements-web]> SELECT gd.ID AS GameID, c.Name AS ConsoleName, c.ID AS ConsoleID, gd.ImageIcon, gd.Title, inner1.MaxPossible, SUM(aw.HardcoreMode = 0) AS NumAwarded, SUM(aw.HardcoreMode = 1) AS NumAwardedHC, (SUM(aw.HardcoreMode = 0) / inner1.MaxPossible) AS PctWon, (SUM(aw.HardcoreMode = 1) / inner1.MaxPossible) AS PctWonHC, aw.HardcoreMode FROM Awarded AS aw LEFT JOIN Achievements AS ach ON ach.ID = aw.AchievementID LEFT JOIN GameData AS gd ON gd.ID = ach.GameID LEFT JOIN ( SELECT COUNT(*) AS MaxPossible, ach1.GameID FROM Achievements AS ach1 WHERE Flags = 3 GROUP BY GameID ) AS inner1 ON inner1.GameID = ach.GameID AND inner1.MaxPossible > 5 LEFT JOIN Console AS c ON c.ID = gd.ConsoleID WHERE aw.User = 'Jamiras' AND ach.Flags = 3 GROUP BY ach.GameID, gd.Title ORDER BY PctWon = 1.0 DESC, PctWonHC DESC, PctWon DESC, inner1.MaxPossible DESC, gd.Title;
ERROR 1247 (42S22): Reference 'PctWon' not supported (reference to group function)
```

EmoonX: From what I could gather here, this seems to be a specific DB/engine limitation caused by an alias expression being used in `ORDER BY`. Try replacing `PctWon = 1.0` for `(SUM(aw.HardcoreMode = 0) / inner1.MaxPossible) = 1.0` and see if the issue is solved for you.

Jamiras: That does seem to work. Thanks.

Jamiras:
I think these image match your described behavior (masteries, completions, then by hardcore %), but they just feel off when I look at them.

Here's the transition from masteries through completions into incomplete by hardcore %:
image

And the end of the list where the hardcore % approaches zero:
image

Maybe having the overall percent first, then the hardcore percent second would be appropriate for the entire list? Masteries would still be first (100% overall, 100% hardcore), then completions (100% overall, <100% hardcore), then everything else ordered by overall percent.

lib/database/player-game.php
```diff
@@ -275,19 +275,24 @@ function getUsersCompletedGamesAndMax($user): array
...
-        WHERE aw.User='$user' AND ach.Flags = $requiredFlags
-        GROUP BY ach.GameID, aw.HardcoreMode, gd.Title
-        ORDER BY PctWon DESC, inner1.MaxPossible DESC, gd.Title ";
+        WHERE aw.User = '$user' AND ach.Flags = $requiredFlags
+        GROUP BY ach.GameID, gd.Title
+        ORDER BY (SUM(aw.HardcoreMode = 0) / inner1.MaxPossible) = 1.0 DESC,
+            PctWonHC DESC, PctWon DESC, inner1.MaxPossible DESC, gd.Title";
 
     $db = getMysqliConnection();
+    mysqli_query($db, "SET SESSION innodb_strict_mode=ON");
```
luchaos, line 295: Is this required for this change to work as intended?

The whole legacy mysqli database layer will be replaced in the near future.

Changing the innodb session behaviour in mid-flight seems risky to me. I might be missing an important detail here though. If absolutely required please add a comment line or to to explain why it is required in this particular case.

EmoonX: Ah no, sorry... That was just a leftover from my previous tests.

EmoonX:
@Jamiras I think the transition from completed to incomplete games won't look that off with #1281, as the table will have two easily-discernible parts: the topmost with icons and the rest without them.

Regarding those <100%, the question is basically whether we should prioritize hardcore (as the site does) or just general visual candy. I agree that using overall first produces a nice and single "waterfall"; but then incomplete softcore-only sets are bound to appear above hardcore-only ones. Maybe there are people better suited than me to decide on that.

Jamiras:
> I think the transition from completed to incomplete games won't look that off with https://github.com/RetroAchievements/RAWeb/pull/1281, as the table will have two easily-discernible parts: the topmost with icons and the rest without them.

This would still be true if you ordered by overall completion first. The ones with icons would still appear before all the ones without them.

> the question is basically whether we should prioritize hardcore (as the site does) or just general visual candy.

I still think sorting by overall completion would be better. The widget is "Completion Progress" and doesn't specify hardcore/softcore. Additionally, most people who are really focused on hardcore will purge any softcore unlocks from their account. Therefore, I propose focusing on the softcore players. If a softcore player has one game with a hardcore unlock, it'll appear before all the games where they have almost everything unlocked.

EmoonX:
@Jamiras Fair enough to me. The most important bit is adding hardcore criteria to the ordering. The rest is a matter of opinion; but I agree with your rationale either way.

New ordering:

new-ordering

Jamiras:
Query is failing again on the xampp install:
```
MariaDB [retroachievements-web]> SELECT gd.ID AS GameID, c.Name AS ConsoleName, c.ID AS ConsoleID, gd.ImageIcon, gd.Title, inner1.MaxPossible, ANY_VALUE(aw.HardcoreMode), SUM(aw.HardcoreMode = 0) AS NumAwarded, SUM(aw.HardcoreMode = 1) AS NumAwardedHC, (SUM(aw.HardcoreMode = 0) / inner1.MaxPossible) AS PctWon, (SUM(aw.HardcoreMode = 1) / inner1.MaxPossible) AS PctWonHC FROM Awarded AS aw LEFT JOIN Achievements AS ach ON ach.ID = aw.AchievementID LEFT JOIN GameData AS gd ON gd.ID = ach.GameID LEFT JOIN ( SELECT COUNT(*) AS MaxPossible, ach1.GameID FROM Achievements AS ach1 WHERE Flags = 3 GROUP BY GameID ) AS inner1 ON inner1.GameID = ach.GameID AND inner1.MaxPossible > 5 LEFT JOIN Console AS c ON c.ID = gd.ConsoleID WHERE aw.User = 'Jamiras' AND ach.Flags = 3 GROUP BY ach.GameID, gd.Title ORDER BY PctWon DESC, PctWonHC DESC, inner1.MaxPossible DESC, gd.Title;
ERROR 1305 (42000): FUNCTION retroachievements-web.ANY_VALUE does not exist
```

Just removing the `ANY_VALUE` wrapper from the `aw.Hardcore` field seems to fix the issue, and the widget appears to display correctly. `MAX` also works. What is the need for using `ANY_VALUE`?

EmoonX:
>What is the need for using ANY_VALUE?

Doesn't actually make a difference in the code itself. However, if you try running the same query in e.g DBeaver, strictness ensues and an error is thrown due to listing `aw.HardcoreMode` raw in SELECT:



Thankfully, using `MAX` instead solves both issues.

Jamiras (approved):

PR #1283, opened 2022-12-29
Add deptrac and isolate domains and namespaces

luchaos:
This is in preparation for the upcoming "replace mysqli" PR.

The Legacy namespace is meant to be subsequently replaced as the new data structures are synced and used. To prevent legacy structures bleeding into the new structures deptrac helps with maintaining a clean architecture and enforce usage rules.


app/Legacy/Models/PlayerBadge.php
```diff
@@ -5,8 +5,85 @@
 namespace App\Legacy\Models;
 
 use App\Legacy\Database\Eloquent\BaseModel;
+use RA\AwardType;
 
 class PlayerBadge extends BaseModel
 {
+    // Note: will be renamed and split into Community/UserBadge and Platform/PlayerBadge
```
luchaos, line 12: It has yet to be decided if a split, of what currently is known as "SiteAwards", into two distinct types of Badges is applicable. The addition of `Community/Models/UserBadge` to the existing `Platform/Models/PlayerBadge` is a mere preparation for that for now.

app/Platform/Commands/NoIntroImport.php
```diff
@@ -13,7 +13,7 @@
 
 class NoIntroImport extends Command
 {
-    protected $signature = 'ra:server:rom:no-intro:import {jsonDatFile} {systemId?} {--seed} {--ignore-system-mismatch}';
+    protected $signature = 'ra:platform:rom:no-intro:import {jsonDatFile} {systemId?} {--seed} {--ignore-system-mismatch}';
```
luchaos, line 16: The `ra:server` namespace for commands is a leftover from before it was called `platform`.

app/Site/AuthServiceProvider.php
```diff
@@ -40,12 +39,8 @@ public function boot(): void
          * don't drink and root
          */
         Gate::before(
-            function (User|LegacyUser $user, $ability) {
```
luchaos, line 43: Adding deptrac i realized that i already broke some rules previously.

lib/database/achievement-creator.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-use App\Platform\Models\Badge;
+use App\Legacy\Models\PlayerBadge;
```
luchaos, line 3: Only use legacy models within the legacy domain for now to keep it clean.

deptrac.yaml
```diff
@@ -0,0 +1,158 @@
...
+          value: public/.*
+        - type: directory
+          value: src/.*
+        - type: directory
+          value: database/factories/Legacy/.*
+        - type: directory
+          value: database/migrations/legacy/.*
+        - type: directory
+          value: database/seeders/Legacy/.*
+  ruleset:
```
luchaos, line 111: These rules define what is allowed to be used where. Up for discussion but i think it's a sensible set in the current situation.

.hooks/pre-push
```diff
@@ -12,6 +12,9 @@ if [ ${#modifiedPhpFiles} -gt 0 ]; then
 
     printf "\n⏳ composer analyse\n"
     vendor/bin/phpstan analyse --memory-limit 512M --no-progress
+
+    printf "\n⏳ composer deptrac\n"
+    vendor/bin/deptrac analyse
```
luchaos, line 17: Run it both on push and within the GitHub phpstan/analyse action.

deptrac.yaml
```diff
@@ -0,0 +1,158 @@
...
+      - Community
+      - Support
+      - Site
+    Site:
+      - Community
+      - Connect
+      - Platform
+      - Support
+    Support:
+      - Site
```
luchaos, line 137: This mishmash of usage among "v2" domains will eventually be cleaned up, too.

deptrac.yaml
```diff
@@ -0,0 +1,158 @@
...
+      - Platform
+      - Site
+      - Support
+    Views:
+      - Helpers
+      - Legacy
+    # Legacy
+    Legacy:
+      - Helpers
+      - Support
```
luchaos, line 158: Limit the `Legacy` domain to itself; it is not allowed to be used in "v2 structured" domains yet either to keep complexity low during the transition phase.

Jamiras (approved):
Upon first glance, these appear to be sensible. Until I actually run into a conflict, I'll trust your design. `composer deptrac` passes when run locally.

PR #1285, opened 2022-12-31
Consolidate legacy code

luchaos:
Followup to the previous deptrac PR, extracted from the upcoming "replace mysqli" PR.

Apart from a ton of moved files those changes are fairly trivial; slowly becoming a proper Laravel application. I'd wait for the other open PRs to be merged first. I'll take care of rebasing then.

- Move root folders into dedicated legacy application directory:
  - Classes in `src` moved to `Enum` namespaces within respective domain
  - Helper functions in `lib` moved to `Helpers` namespace
- Apply same domains split as in v2
- Implement MustVerifyEmail in legacy User model
- Fix View contract return types

Tests for future refactorings will be included in the next PR where the database layer will be replaced.

Cronjobs will be moved to commands within legacy app domains in a separate PR.


app_legacy/Platform/Commands/Developer/RecalculateContributionYield.php
```diff
@@ -2,26 +2,21 @@
...
-use RA\AwardType;
+use LegacyApp\Community\Enums\AwardType;
+use LegacyApp\Platform\Enums\AchievementType;
+use LegacyApp\Platform\Models\PlayerBadge;
+use LegacyApp\Site\Models\User;
 
 class RecalculateContributionYield extends Command
 {
-    protected $signature = 'ra-legacy:community:developer:recalculate-contribution-yield {username?}';
+    protected $signature = 'ra-legacy:platform:developer:recalculate-contribution-yield {username?}';
```
luchaos, line 16: Accidentally changed this to `:community:` in the previous PR.

app_legacy/Site/Models/User.php
```diff
@@ -99,4 +88,25 @@ public function getAvatarUrlAttribute(): string
...
+    }
+
+    public function sendEmailVerificationNotification(): void
+    {
+    }
+
+    public function getEmailForVerification(): string
+    {
+        return $this->EmailAddress;
+    }
```
luchaos, line 111: Not relevant for now but must be implemented by contract/interface.

app_legacy/Helpers/database/release.php
```diff
@@ -110,7 +110,12 @@ function getIntegrationRelease(): ?array
 
 function getReleasesFromFile(): ?array
 {
-    return file_exists(storage_path('app/releases.php')) ? require_once storage_path('app/releases.php') : null;
+    try {
+        return file_exists(storage_path('app/releases.php')) ? require_once storage_path('app/releases.php') : null;
+    } catch (Throwable) {
```
Jamiras, line 115: Do we really want to just swallow this exception? Shouldn't it at least get logged?

It tripped me up for about 15 minutes because the `use` path in my `releases.php` hadn't been updated and things relying on the releases information were just misbehaving.

Jamiras: Also, as updating that is a manual step, we must remember to do it when we deploy, and need to somehow inform all devs after this gets merged.

Jamiras (approved):
It's going to take some time to retrain myself to look in `app_legacy/Helpers` instead of `lib`, but I didn't encounter any obvious issues with some spot testing (including the dorequest and public APIs).

app_legacy/Helpers/database/release.php
luchaos: Very good point. Sorry that I missed that in the self-review - I wanted to add a comment about that initially.

See https://github.com/RetroAchievements/RAWeb/pull/1285/commits/c0ef78bcfdbba642de5ef4edf3e84e7be07c2d96

PR #1290, opened 2023-01-04
Hide banned/spam accounts' user pages

luchaos:
Return 404 to non-admin/mods when trying to access a banned/spam account's profile page.

Update composer & npm dependencies while at it and ran rector on some newer changes.

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -25,7 +25,7 @@ if (empty($userMassData)) {
     abort(404);
 }
 
-if ((int) $userMassData['Permissions'] < Permissions::Registered && $permissions < Permissions::Admin) {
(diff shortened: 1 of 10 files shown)


public/userInfo.php
```diff
@@ -25,6 +25,10 @@
     abort(404);
 }
 
+if ((int) $userMassData['Permissions'] < Permissions::Registered && $permissions < Permissions::Admin) {
+    abort(404);
+}
```
Jamiras, line 30: An Unregistered user should be allowed to access their own page.

Jamiras (changes requested):

public/userInfo.php
luchaos: Oh my ... thanks for spotting that!

Jamiras (approved):

PR #1291, opened 2023-01-06
Fix claim tables on front page & cleanup

luchaos:
Introducing the wrapper for game titles makes them break between icon and label when too long:
Screenshot 2023-01-06 at 16 56 43

Move icons to a separate column like in other places:
Screenshot 2023-01-06 at 16 56 07

Includes minor cleanups:
- Docblocks
- Only keep the last release instead of five

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -78,8 +78,15 @@ ${PHP_BIN} artisan config:cache
 
 ### Cleanup ###
 
-# cleanup releases - keep the last five (pass +6 to tail)
+# cleanup releases
+
+# keep the last five (pass +1 to tail)
+#cd "${BASEDIR}/${RELEASES_DIR}"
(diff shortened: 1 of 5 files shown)


Jamiras (commented):
This is definitely an improvement, but it really feels like the game image is just floating in a limbo between the two text fields. Would it be possible to have the headers span both columns (also affects User) and adjust the column width so the text is closer to the image?

image

luchaos:
Moving the header to the left would affect all tables. In general I prefer the header columns to be in line with texts that it potentially sorts (see game and achievement tables) - similar to https://completionist.me/retro/achievements. 

However, I agree that the icons are a bit too far away. The horizontal padding for all table cells came up as an issue before - decreasing that would make it look slightly less detached:
Screenshot 2023-01-06 at 18 23 11

Jamiras (approved):

PR #1294, opened 2023-01-06
fix padding when awards are hidden

Jamiras:
The calculation for how many blank items being added to the end of the list was not accounting for hidden items, resulting in displays like this:
image


luchaos:
Addressed in https://github.com/RetroAchievements/RAWeb/pull/1293

The columns per line calculation doesn't hold up on small screens.

PR #1296, opened 2023-01-06
Fix progress bar completion icon alignment

luchaos:
Simplify progress CSS & tailwind class usage. Same result.

Revert the `.progress` container; a relative container is not needed when aligning progress bar a completion icon using flex. The global and player variants only set the width which can be replaced by reusable utility classes.

Progress bar labels can be aligned using utility classes, too, where needed.

All of this will be moved to a normalized view component eventually so we do not have to maintain it in three different places.

Fix tailwind content paths config.

### DIFF
--- a/resources/css/progress.css
+++ b/resources/css/progress.css
@@ -3,6 +3,7 @@
   height: 6px;
   border: 1px solid var(--embed-highlight-color);
   border-radius: .5em 0 0 .5em;
+  border-right: 0;
   background-color: var(--embed-color);
 }
 
@@ -34,6 +35,7 @@
   border-radius: 50%;
   font-size: 10px;
   line-height: 1em;
+  margin-left: -1px;
 }
 
 .completion-icon.tooltip {
@@ -41,12 +43,12 @@
 }
 
 .completion-icon.mastered {
-  border-color: #c90;
+  border-color: rgb(204, 153, 0);
   background-color: #606;
 }
 
 .completion-icon.completed {
-  border-color: #0b71c1;
+  border-color: rgb(11, 113, 193);
   background-color: #083171;
 }
 
--- a/resources/css/progress.css
+++ b/resources/css/progress.css
@@ -3,7 +3,7 @@
   height: 6px;
   border: 1px solid var(--embed-highlight-color);
   border-radius: .5em 0 0 .5em;
-  border-right: 0;
+  border-right: 0 none;
   background-color: var(--embed-color);
 }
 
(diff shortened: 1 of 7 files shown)


EmoonX:
Good to know that mess could be cleaned up. Flexboxes make magic happens indeed 🪄

Only caveat is that there's a little gap between bar and icon due to the wrapping border on bar. One way to amend it is to set `border-right: none` for player progressbars.

image

luchaos:
> Only caveat is that there's a little gap between bar and icon due to the wrapping border on bar.

Was thinking about that gap, too, but found it to be ok actually as it tells where the bar actually ends?

Removing the right border would work, but would look weird with the border radius. Which curiously is missing in your screenshot. I'll check that again and see if it's even needed. Brb.

resources/css/progress.css
```diff
@@ -59,32 +26,29 @@
...
+  height: 20px;
+  width: 20px;
   border: solid 2px;
   border-radius: 50%;
-  font-family: emoji;
-  font-size: 7pt;
-  cursor: default;
+  font-size: 10px;
+  line-height: 1em;
+  margin-left: -1px;
```
luchaos, line 38: Just realized that the border radius was already removed, nice :) 
Removed the gap by removing the border and moving the completion icon one pixel to the left.

luchaos: Looks like the negative margin was only needed when zoomed in (which also is the cause of the apparently missing bottom left radius in your screenshot i assume?)

luchaos:
Award counter alignment

Screenshot 2023-01-07 at 02 18 10

resources/css/progress.css
EmoonX: >(which also is the cause of the apparently missing bottom left radius in your screenshot i assume?)

Now that you mention it... you're right. They constantly appear off to me. Zooming in seemingly messes the rendering for corners.

luchaos:
Still can't upload images to discord...

About those tags/labels: how about a more unified inverted version that does not have too many variants and uses existing color variables?

I know you wanted to make a distinction between types but i feel it's still to early to introduce such changes before the tag system is in place. Tags will have to then emulate _every_ behaviour that's added until then. Hence why i wanted those colors to be gone previously. Keep it simple for now.

Screenshot 2023-01-07 at 02 42 03
Screenshot 2023-01-07 at 02 42 13
Screenshot 2023-01-07 at 02 42 23

Jamiras (approved):

PR #1298, opened 2023-01-07
Revoke mastery badge when resetting achievements for a set

Jamiras:
Supports #658 

When resetting a published (core) achievement, or an entire achievement set, the associated mastery/completion badge will now be revoked.

Additionally, a manual upgrade command has been added that will audit existing mastery badges. This can be run after deployment. If a user has a mastery badge for a game and _no_ achievements unlocked for that game, and the game still has published achievements, then the badge will be revoked. If the game has no published achievements, it is assumed the set has been demoted for repairs. If the user has at least one achievement unlocked for the game, it is assumed that a revision occurred and the badge will not be revoked. If the user wishes to revoke the badge themselves, they may reset any or all achievements for the game.

### DIFF
--- a/app_legacy/Platform/Commands/UpdatePlayerMasteries.php
+++ b/app_legacy/Platform/Commands/UpdatePlayerMasteries.php
@@ -1,0 +1,140 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LegacyApp\Platform\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+use LegacyApp\Community\Enums\AwardType;
+use LegacyApp\Platform\Enums\AchievementType;
+use LegacyApp\Platform\Enums\UnlockMode;
+use LegacyApp\Site\Models\User;
+
+class UpdatePlayerMasteries extends Command
+{
+    protected $signature = 'ra-legacy:platform:update-player-masteries {username?}';
+
+    protected $description = 'Recalculates masteries for a user';
+
+    private $gameAchievements = [];
+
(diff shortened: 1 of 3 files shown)


app_legacy/Platform/Commands/UpdatePlayerMasteries.php
```diff
@@ -0,0 +1,140 @@
...
+        $username = $this->argument('username');
+        if (!empty($username)) {
+            $this->recalculate($username);
+
+            return;
+        }
+
+        $users = DB::connection('mysql_legacy')
+            ->table('SiteAwards')
+            ->where('AwardType', '=', AwardType::Mastery)
```
luchaos, line 33: `DB::connection('mysql_legacy')>table('SiteAwards')` can usually be replaced by using the `PlayerBadge` model's facade directly (which is bound to the `SiteAwards` table), e.g.:

```php
use LegacyApp\Platform\Models\PlayerBadge;

PlayerBadge::where('AwardType', '=', AwardType::Mastery)
    ->where('AwardType', '=', AwardType::Mastery)
    ->distinct()
    ->get(['User'])
```

I will however touch that in the mysql refactoring PR - so no need to do that right now.

luchaos (approved):
I think the outlined approach to handling existing badges is very good.

luchaos:
Tested locally: resetting one achievement of a previously mastered game will not remove the badge, resetting all of them does remove the badge. As I understood it, resetting a single achievement manually should have removed the badge, too?

luchaos:
Tested a few more cases using the command - works all just fine 👍

PR #1300, opened 2023-01-07
use cards for game award tooltips

Jamiras:
Updates the tooltips on the game awards to use the game card.

Before:
image

After:
image

This does affect the event awards as they're game entries. It does not affect the contribution or patreon badges.

### DIFF
--- a/app_legacy/Helpers/render/site-award.php
+++ b/app_legacy/Helpers/render/site-award.php
@@ -174,16 +174,21 @@ function RenderAward($award, $imageSize, $clickable = true): void
 
     if ($awardType == AwardType::Mastery) {
         if ($awardDataExtra == '1') {
-            $tooltip = "MASTERED $awardGameTitle ($awardGameConsole)";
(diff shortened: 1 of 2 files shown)


luchaos (approved):
Nice

app_legacy/Helpers/render/site-award.php
```diff
@@ -174,16 +174,21 @@ function RenderAward($award, $imageSize, $clickable = true): void
 
     if ($awardType == AwardType::Mastery) {
         if ($awardDataExtra == '1') {
-            $tooltip = "MASTERED $awardGameTitle ($awardGameConsole)";
+            $awarded = "MASTERED on $awardDate";
```
luchaos, line 177: Just a thought - should we case this normally (`Mastered`) as we did that in other places, too?

Jamiras: Yea, will do.

PR #1301, opened 2023-01-07
hide game progress if game has no achievements

Jamiras:
The new progress widget really stands out on game pages with no achievements compared to before.

Current:
image

Before:
image

But there's no reason for it to be there. The Set Request widget really should just replace it. So that's what I did.

After:
image


luchaos (approved):
Much better, agreed 👍

PR #1302, opened 2023-01-07
include game name in page titles for game subpages

Jamiras:
Affects the following game subpages:
* Code Notes
* Manage Leaderboards
* Linked Hashes
* Manage Hashes
* Manage Claims

Helps to differentiate subpages when creating bookmarks and linking in Discord

### DIFF
--- a/public/leaderboardList.php
+++ b/public/leaderboardList.php
@@ -21,18 +21,18 @@ $sortBy = requestInputSanitized('s', empty($gameID) ? 3 : 0, 'integer');
 
 $lbCount = getLeaderboardsList($consoleIDInput, $gameID, $sortBy, $count, $offset, $lbData);
 
-$pageTitle = "Leaderboard List";
+$pageTitle = "Leaderboards - ";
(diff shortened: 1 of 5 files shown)


public/leaderboardList.php
```diff
@@ -21,16 +21,18 @@
 
 $lbCount = getLeaderboardsList($consoleIDInput, $gameID, $sortBy, $count, $offset, $lbData);
 
+$pageTitle = "Leaderboard List";
```
luchaos, line 24: I'd go with "Leaderboards" like on game pages; "List" seems redundant?

Just noticed that it's "Achievements List" in `achievementList.php`; could be just "Achievements", too.

luchaos (approved):

public/leaderboardList.php
Jamiras: That's unchanged (just move from line 48), but the change makes sense. I'll update it to match the breadcrumb since I changed that too.

PR #1304, opened 2023-01-08
Matching color tags and unique subset styling

EmoonX:
Follow up to #1297.
- Change tags so their background always match the adjacent text color
- Add the bright-arrow-label styling to subset tags

Maybe this can still use some visual improvements. Gonna think a bit about it.


Jamiras (approved):

resources/css/tag.css
```diff
@@ -1,13 +1,38 @@
...
+  border-right-width: 0;
+}
+
+.tag .inner {
+  padding: .15em 0;
+  color: black;
+}
+
+.tag .label ~ .inner {
+  margin-left: .2em;
```
luchaos, line 37: It doesn't need two classes for both label and inner; just a class for the highlighted part (and the arrow). Just adding a span without the class can then inherit default behaviour. 

The color could be theme aware using a css var.

```css
.tag > * {
  color: var(--bg-color);
}

.tag-sub {
  display: inline-block;
  margin: -.1em .4em -.1em -.4em;
  padding: .1em .4em;
  background: rgba(var(--highlight-color-values), 50%);
}
```

`.tag-sub` would be the same as `.label` here - could be `.tag-label` to be more specific in this case.

luchaos (approved):

resources/css/tag.css
```diff
@@ -1,13 +1,38 @@
 .tag {
-  display: inline-block;
-  opacity: .9;
+  display: inline-flex;
+  align-items: center;
+  opacity: 0.9;
   margin: 0 .15em;
```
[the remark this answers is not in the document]
EmoonX: >and the margin here removed as it can be just an inline element implicitly 

Main reason margin is there is to make the tags more spaced out. They look a bit too close to text without it, IMO.

>I think the achievement-set classes are not needed anymore at this point so could be removed altogether? The margin of .achievement-set.title should not be applied globally.

Indeed, considering how barren `achievement-set.css` looks now, that class is just a leftover at this point. I think some game avatars could use a larger gap between icon and title, but there are better ways to solve that.

>If variants of tags should exist they should be just tag variants themselves?

Makes the most sense to me. They can generally just go to `tag.css`.

resources/css/tag.css
EmoonX: >It doesn't need two classes for both label and inner

I recall I merely created `.inner` because tags in general needed the inner span (for `currentColor` to work) and I didn't want to leave just a generic `<span>$category</span>` there.

EmoonX:
```css
background: rgba(var(--highlight-color-values), 50%);
```
Wow. Had no idea this was even possible.

resources/css/tag.css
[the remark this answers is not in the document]
luchaos: Added the margin back 👍

resources/css/tag.css
luchaos: Ad inner class: I can understand. Yet, I removed so many confusing and unused classes in this project already that I'm careful to add any new 😅 I'm personally fine with a structural element without a class.

luchaos: > ```css
> background: rgba(var(--highlight-color-values), 50%);
> ```

Yeah, that was a new one for me, too. However, didn't end up using it in the other commits :) white with just opacity as you did it works much better.

EmoonX: >I'm personally fine with a structural element without a class

I think it's better your way. Sadly webdev-driven OCD leads us to want a class in every `<span>` and `<div>` we can find. 🤓

PR #1314, opened 2023-01-12
Simplify breadcrumbs and separate main/subset links

EmoonX:
Since we're in the process of semantically separating tags from the actual game titles, and since there have been concerns about the visual clutter in breadcrumbs (either in the old or new formats), here are some changes I've implemented:

- Category tags won't be displayed (in general) when a game has one. 99.9% of the time its title is already unique for a given console, and the full tagged title usually appears right below in most pages anyway

void

- In the really rare case of a pair of games sharing the same base title, the category tag for the derived one is shown -- to avoid ambiguity. Unless I'm mistaken about my searches, this one below is currently the sole actual instance in... _5792_ sets

1-in-5792

- Subsets get more interesting. I've separated the main title and subset tag into different crumbs. Clicking the first link sends you to the main set, while clicking the second one takes you to the subset page

subsets

Once again, I'm not sure if I managed to catch every page that employs the navbar. Feel free to report missing ones.

### DIFF
--- a/app_legacy/Helpers/render/game.php
+++ b/app_legacy/Helpers/render/game.php
@@ -93,46 +93,42 @@ function renderGameTitle(?string $title, bool $tags = true): string
  */
 function renderGameBreadcrumb(array $data, bool $addLinkToLastCrumb = true)
 {
+    [$consoleID, $consoleName] = [$data['ConsoleID'], $data['ConsoleName']];
+
     // Return next crumb (i.e `» text`), either as a link or not
     $nextCrumb = function ($text, $href = ''): string {
         return " &raquo; " . ($href ? "<a href='$href'>$text</a>" : "<b>$text</b>");
     };
 
     // Retrieve separete IDs and titles for main game and subset (if any)
-    $getSplitData = function ($data): array {
+    $getSplitData = function ($data) use ($consoleID): array {
         $gameID = $data['GameID'] ?? $data['ID'];
         $gameTitle = $data['GameTitle'] ?? $data['Title'];
 
+        // Match and possibly split main title and subset
+        [$mainID, $mainTitle] = [$gameID, $gameTitle];
         $matches = [];
         if (preg_match('/(.+)(\[Subset - .+\])/', $gameTitle, $matches)) {
-            [$gameTitle, $subset] = [trim($matches[1]), $matches[2]];
+            [$mainTitle, $subset] = [trim($matches[1]), $matches[2]];
+            $mainID = getGameIDFromTitle($mainTitle, $consoleID);
             $subsetID = $gameID;
             $renderedSubset = renderGameTitle($subset);
-
-            // Retrieve main game ID
-            sanitize_sql_inputs($gameID);
-            settype($gameID, 'integer');
-            $query = "SELECT ga.gameIDAlt FROM GameAlternatives ga
-                WHERE ga.gameID = $gameID";
-            $result = s_mysql_query($query);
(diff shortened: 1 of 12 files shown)


app_legacy/Helpers/render/game.php
```diff
@@ -56,25 +56,93 @@ function gameAvatar(
...
+            $subsetID = $gameID;
+            $renderedSubset = renderGameTitle($subset);
+
+            // Retrieve main game ID
+            sanitize_sql_inputs($gameID);
+            settype($gameID, 'integer');
+            $query = "SELECT ga.gameIDAlt FROM GameAlternatives ga
+                WHERE ga.gameID = $gameID";
+            $result = s_mysql_query($query);
+            $gameID = $result->fetch_array()[0];
```
Jamiras, line 115: This query probably needs some adjustments. Several subsets are linked to more than one game alternative - especially if you include the hubs. It's just happenstance that the first linked game is usually the parent. I recommend stripping the [Subset] moniker and searching for the game title. This could be combined with the logic below for determining if a base title is shared.

public/reportissue.php
```diff
@@ -52,10 +52,7 @@ function displayCore() {
 <div id="mainpage">
     <div id="fullcontainer">
         <div class="navpath">
-            <a href="/gameList.php">All Games</a>
-            &raquo; <a href="/gameList.php?c=<?= $consoleName ?>"><?= $consoleName ?></a>
-            &raquo; <a href="/game/<?= $gameID ?>"><?= renderGameTitle($gameTitle) ?></a>
-            &raquo; <a href="/achievement/<?= $achievementID ?>"><?= $achievementTitle ?></a>
```
Jamiras, line 58: the achievement crumb has been lost

Jamiras:
>  I'm not sure if I managed to catch every page that employs the navbar. 

I think I touched them all in #1307. You should be able to grep for `renderGameTitle`. It looks like, at a minimum, you missed the Manage Hashes page.

app_legacy/Helpers/render/game.php
```diff
@@ -56,25 +56,93 @@ function gameAvatar(
...
+        // with another one, include category to solve ambiguity
+        $renderedMain = renderGameTitle($gameTitle, tags: false);
+        if ($renderedMain !== $gameTitle) {
+            sanitize_sql_inputs($gameTitle);
+            $query = "SELECT Title FROM GameData
+                WHERE Title = TRIM(SUBSTRING_INDEX('$gameTitle', '~', -1))";
+            $result = s_mysql_query($query);
+            if (!$result) {
+                $renderedMain = renderGameTitle($gameTitle);
+            }
```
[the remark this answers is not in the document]
EmoonX: Aside from the original query not taking consoles into account, I noticed the check I did made no sense at all. It merely worked for `~Hack~ Rock n' Roll Racing` by a false positive 🤦

Using `getGameIDFromTitle` works wonders. Thanks.

app_legacy/Helpers/render/game.php
EmoonX: > It's just happenstance that the first linked game is usually the parent

Yep, I (wrongly) assumed that was the rule and not merely usual.

EmoonX:
>From your images above, there appears to be an issue where an apostrophe in the game title is being escaped unnecessarily

Happened due to sanitizing game's title for the query. Using `getGameIDFromTitle` instead fix it.

Jamiras (approved):

PR #1315, opened 2023-01-12
Render `[m]` in achievement titles as a tag

EmoonX:
Self-explanatory. Should be able to be eventually expanded to other kinds of achievement tags.

normal
hover
tooltip

### DIFF
--- a/app_legacy/Helpers/render/achievement.php
+++ b/app_legacy/Helpers/render/achievement.php
@@ -64,12 +64,11 @@ function renderAchievementTitle(string $title, bool $tags = true): string
     $span = '';
     if ($tags) {
         $span = "<span class='tag missable' title='Missable'>"
-            . "<div>[<span>m</span>]</div>"
+            . "<abbr>[<b>m</b>]</abbr>"
             . "</span>";
     }
-    $html = trim(str_replace('[m]', $span, $title));
 
-    return $html;
+    return trim(str_replace('[m]', $span, $title));
 }
 
 function renderAchievementCard(int|string|array $achievement, ?string $context = null): string
--- a/app_legacy/Helpers/render/achievement.php
+++ b/app_legacy/Helpers/render/achievement.php
@@ -53,13 +53,16 @@ function achievementAvatar(
     );
 }
 
-function renderAchievementTitle(string $title): string
-{   
+/**
+ * Render achievement title, parsing `[m]` (missable) as a tag
+ */
+function renderAchievementTitle(string $title, bool $tags = true): string
+{
     if (!str_contains($title, '[m]')) {
         return $title;
     }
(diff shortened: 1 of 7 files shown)


Jamiras (commented):
While visually pleasant, this makes it impossible to search for missables on a page. I've seen users mention using "Ctrl+F" "[m]" to find missables when viewing a set. 

It might be worth waiting on this change.

EmoonX:
> this makes it impossible to search for missables on a page. I've seen users mention using "Ctrl+F" "[m]" to find missables when viewing a set

Well, this actually doesn't need to be an issue. We can just make the brackets invisible instead:

invisible
ctrl+f

And to make it less confusing to new users, the brackets can be shown on hover:

image

Jamiras (approved):

app_legacy/Helpers/render/achievement.php
```diff
@@ -45,9 +48,30 @@ function achievementAvatar(
...
+ */
+function renderAchievementTitle(string $title, bool $tags = true): string
+{
+    if (!str_contains($title, '[m]')) {
+        return $title;
+    }
+    $span = '';
+    if ($tags) {
+        $span = "<span class='tag missable' title='Missable'>"
+            . "<div>[<span>m</span>]</div>"
```
luchaos, line 67: As it's an inline element this should probably be a `span` instead of a `div`?

app_legacy/Helpers/render/achievement.php
```diff
@@ -45,9 +48,30 @@ function achievementAvatar(
...
+            . "</span>";
+    }
+    $html = trim(str_replace('[m]', $span, $title));
+
+    return $html;
```
luchaos, line 72: Can omit the local variable here:

```php
return trim(str_replace('[m]', $span, $title));
```

luchaos (approved):

luchaos:
I think the tag should be written out as `missable` instead of `[m]` but let's keep it as is for now

app_legacy/Helpers/render/achievement.php
EmoonX: Hmmm yeah, it seems it isn't a good convention putting `div`s inside `span`s. And the context for tags is inline since they flow alongside text. To be honest, I'm just not fond of having something like `<span><span>...<span>...</span>...</span></span>`. I'll see if I can find a semantically good way of addressing this.

EmoonX: P.S: the reason we need an intermediate `<span>` is just because `.tag` is `inline-flex`, which breaks the inner text when searching with `Ctrl+F`.

EmoonX: Okay, I think this feels better now:
```html
<span class='tag missable' title='Missable'>
   <abbr>[<b>m</b>]</abbr>
</span>
```

PR #1323, opened 2023-01-17
Highlight list rows (lightly) on hover

EmoonX:
A small & simple visual feedback. Overall this is a common element in UI lists (e.g GitHub/GitLab use it) and it (hopefully) helps users focusing on the content they're hovering at the moment.

[hover](https://user-images.githubusercontent.com/22218549/212938673-e5c53192-6789-438c-96f3-4eb1e91d5a00.webm)

Applied to ~all~ most tables site-wide.


resources/css/table.css
```diff
@@ -10,6 +10,21 @@ tbody tr:nth-child(odd) {
   background-color: rgba(80, 80, 80, .09);
 }
 
+tbody tr:not(.forum-post) {
+  outline: 1px solid transparent;
```
EmoonX, line 14: `outline` shouldn't be an issue here, as `border-radius` isn't being used and it isn't supposed to break anything anyway.

Jamiras:
In general, this works well. But there are several places with nested tables that this makes weird.

The biggest issue is viewing a ticket. Every row of the summary can be highlighted individually.
image
(This also occurs in the game summary, but it feels less distracting there).

Additionally, there's nested tables in the ticket page for Player History and Comments.
image
image

Similar behavior occurs in the dev box on the achievement info page.
image

And the header can be highlighted in the inbox
image

There may be some other outliers - I didn't look at all of the pages.

EmoonX:
Ideally it should style just lists (e.g games, cheevos, wall comments). Sadly the legacy HTML for the site follows the trend "everything is a `<table>`". Gonna work on figuring out and restricting the cases.

luchaos:
> Ideally it should style just lists (e.g games, cheevos, wall comments). Sadly the legacy HTML for the site follows the trend "everything is a `<table>`". Gonna work on figuring out and restricting the cases.

I'd be fine with `table-highlight` class which specifically enables those hover styles

EmoonX:
I've searched far and wide and analyzed over ***50*** php pages one-by-one.

Where highlight was applied (several may pertain at once):
- Content naturally organized as a list (games, achievements, users, comments)
- Content that follows a pattern (e.g some setting sections in the control panel)
- Places where highlighting the row helps user focusing better (e.g the large rows in the developer stats page)

Where highlight _wasn't_ applied:
- Table headers (i.e first row when it has `th`s)
- List-like content in which rows don't really follow a pattern
- Isolated fields (like comment input)

All the relevant tables were marked with `.table-highlight` and all the excluding rows with `.do-not-highlight`.

public/gameSearch.php
```diff
@@ -60,14 +60,16 @@
...
+        echo "<table class='table-highlight'><tbody>";
 
+        echo "<tr class='do-not-highlight'>";
         echo "<th>Rank</th>";
         echo "<th>Game</th>";
         echo "<th>Genre</th>";
         echo "<th>Publisher</th>";
         echo "<th>Developer</th>";
         echo "<th>Total Retro Ratio</th>";
+        echo "</tr>";
```
EmoonX, line 72: In occurrences like these (the majority of them, actually), styling the first row as `.do-not-highlight` could be avoided if we just had the `th`s nicely wrapped in a `thead`. Sadly, `thead` pretty much isn't used anywhere in the legacy code...

luchaos: I'll put in the work and add the `thead` elements so we can remove the `.do-not-highlight` class in this PR if that's ok for you?

EmoonX: Sadly, we can't just remove `.do-not-highlight` altogether due to the need of excluding other `tr` instances besides the topmost ones in tables (like comment fields).

That's why I opted for using it as an intermediate solution instead of toying with layout -- since we're going to move to blade templates and I imagine most of the "phony tables" will be transformed in generic containers instead.

luchaos: Sounds good, let's have it this way in the meantime 👍

luchaos (approved):

PR #468, opened 2020-05-02
Use GameID from submitgametitle request if available

juherr:
and improve game title management

Fix #289

PR #1325, opened 2023-01-21
Follow-up fixes for breadcrumb tags

EmoonX:
Follow-up to #1314, as per @RALordAndrew's report.

- Fix breadcrumb links not working for game titles with special characters (e.g `/`, `'` and `\`). Previously, sanitization was converting titles like `.hack//Infection` to `.hack--Infection`, which caused queries to fail. Solved by modifying `sanitizeTitle` to slash-escape text instead; then, `s_mysql_sanitized_query` is called (which is fine, I hope, since the dangerous parts have already been sanitized).

- Add tag rendering to `comments.php` breadcrumb + header:

image

---

Furthermore, we have some breaking cases revolving around base game and subset naming; more precisely, when `$title` isn't actually the base game for the subset `$title [Subset - XYZ]`. However, I could find only two instances of it, which are ironically opposite in nature:
1. [This subset](https://retroachievements.org/game/16084) being derived from the Red part of [Pokemon Red / Blue](https://retroachievements.org/game/724), which is a merged set
2. [This other subset](https://retroachievements.org/game/12399) being derived both from [Zelda Oracle of Ages](https://retroachievements.org/game/710) and [Zelda Oracle of Seasons](https://retroachievements.org/game/676), which are split.

I couldn't think of a clever way to solve it, so I left the issue untouched. Ideas?

### DIFF
--- a/app_legacy/Helpers/database/game.php
+++ b/app_legacy/Helpers/database/game.php
@@ -485,7 +485,7 @@ function getGameIDFromTitle($gameTitleIn, $consoleID): int
               FROM GameData AS gd
               WHERE gd.Title='$gameTitle' AND gd.ConsoleID='$consoleID'";
 
-    $dbResult = s_mysql_query($query);
+    $dbResult = s_mysql_sanitized_query($query);
     if ($retVal = mysqli_fetch_assoc($dbResult)) {
         settype($retVal['ID'], 'integer');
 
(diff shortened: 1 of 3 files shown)


app_legacy/Helpers/database/game.php
```diff
@@ -817,13 +817,18 @@ function submitNewGameTitleJSON($user, $md5, $gameIDin, $titleIn, $consoleID, $d
...
-    $title = str_replace("'", "''", $titleIn);
-    $title = str_replace("/", "-", $title);
+    $title = $titleIn;
+    $tokens = ["\\", "'", "/"];
+    foreach ($tokens as $token) {
+        $title = str_replace($token, "\\$token", $title);
+    }
 
-    return str_replace("\\", "-", $title);
+    return $title;
```
Jamiras, line 831: `sanitizeTitle` should be discarded in favor of just passing `$gameTitle` to `sanitize_sql_inputs`:
```
function getGameIDFromTitle($gameTitle, $consoleID): int
{
    sanitize_sql_inputs($consoleID, $gameTitle);
    settype($consoleID, 'integer');
```

It looks like it was added in #468. I'm not sure why `sanitize_sql_inputs` was not used there.

Jamiras: Note that it's also called from `createNewGame`, so a similar change would need to be made there.

EmoonX: Well, there goes my solution... at least it's nice to be able to remove extra code.

Should I just delete `sanitizeTitle` then? It's not used anywhere else besides those two functions.

EmoonX:
I actually found out two even worse breaking instances...
- https://retroachievements.org/game/5214
- https://retroachievements.org/game/15689

Those ones refer to the JP version of Pokemon Red / Blue, so it's not like there's even a main set for them.

I don't think this is worthy of being dealt with in general until we have something in the DB side of things to mark derived games. For now, I'll just _not_ show the links whenever the code is unable of finding correspondence with a base game.

Jamiras (approved):

PR #1327, opened 2023-01-22
Visual fixes for Game Info page

Krylan:
This PR provides some visual changes to Game Info page.
It includes:

- do not display "Compare" section, if there are no achievements (there is nothing to compare) or when user is not logged in (there is no primary profile to compare, using this function in this state leads to "Unauthorized" error),
- do not display "Achievement Distribution" if there are no achievements (showing empty graph is not necessary),
- fix wrong bottom margin for "Create Forum Topic" button,
- changed headings for whole page, because of accessibility and SEO reasons, but still keeping their visuals unchanged by additional CSS classes (these can be used on other pages as well, to fix their hierarchy),
- added heading for "Set Requests" section,
- "No Active Claims" text doesn't show, if game already has achievements (in any other situation it displays as always did),
- fix achievement list progress bars, as right side border was different than left side (changes applied only there),
- reordered elements within "Achievements" section. Moved "Game Rating" (and "Set Requests" as well) section to the top-left of "Achievements", brought authors and claim info at the beginning of the section, and achievement stats along with user progress below. There is a screenshot how it looks like:
image
Mobile view for these sections are fixed, too.


Jamiras (approved):
> do not display "Achievement Distribution" if there are no achievements (showing empty graph is not necessary),

The "High Scores" section is also not applicable if there are no achievements.

I'd suggest also hiding the "Leaderboards" section if there are no achievements _and_ no leaderboards (there are several games in the DB that have leaderboards without achievements - like 4977 and 18490).

EmoonX:
>do not display "Compare" section, if there are no achievements (there is nothing to compare) or when user is not logged in (there is no primary profile to compare, using this function in this state leads to "Unauthorized" error)

I like the idea in general, but I think users should still be able to "compare" even when not logged in. Currently, a logged user X can compare with user Y even if 1) X haven't played the game, or 2) Y haven't played the game, or 3) neither of them played it at all. Thus the compare page ends up being just a "see which cheevos this other user has" page in some cases.

Ideally, we could have some updates to `gamecompare.php` to handle access when not logged in; and maybe a special messages for when an user has zero achievements in a game.

EmoonX:
>fix wrong bottom margin for "Create Forum Topic" button,

Not directly related to the PR, but I've noticed `Create Forum Topic` is the only list button stylized as `btn-link` (which is kinda ironic considering it's the only one which isn't a link but a POST action instead...). I think having all of them looking the same would make more sense.

(also, maybe a separate emoji to distinguish it from `Official Forum Topic`? I'd suggest ✨)

create-topic

EmoonX:
Not a thing for now, but eventually this part should undergo a total revamp. Way too verbose and the same can be achieved in a much more concise way by other UI means instead of text.

verbose

Krylan:
@Jamiras Yeah, "High Scores" can be hidden too, I guess – I will commit it. In case of leaderboards, I think it serves the same purpose as an empty "Achievements" section – to show that there could be leaderboards, but for now there is nothing, so it encourages devs with some text to make it. I would keep this section this way as it is.

@EmoonX but if not logged user uses "Compare" and it shows data only for requested user, then we could still call it "Compare"? :D 
I don't mind if this would have other purpose, but at this state it makes more sense to hide in situations, when you can't really use it.

I think that "Create Forum Topic" is made this way to not click on it by mistake, thinking it is a link to existing topic (as they would look exactly the same, except the small difference in text). While this button doesn't have background and border, it stands out from other links, so it is more clear that this is a different action.

And about revamps, there is quite a lot of things which could definetely look better. In my free time I try first to tackle things, which really stings my eyes, but I know that in UX/UI there is some what to do :D

EmoonX:
>but if not logged user uses "Compare" and it shows data only for requested user, then we could still call it "Compare"? :D

Semantically, nope. That's why it'd better to call it something else in this case.

Even so, I don't recall the site offering a way to show individual user achievements for a game, if not by going to the compare page itself. Ideally, we should have either a separate page to handle the single user case, or modify `gamecompare.php`'s code (and the filename, possibly) to allow either 1 or 2 users as parameters. And then add some site-wide links where it's fitting. Of course, this is outside of the scope of this PR.

>I think that "Create Forum Topic" is made this way to not click on it by mistake, thinking it is a link to existing topic (as they would look exactly the same, except the small difference in text). While this button doesn't have background and border, it stands out from other links, so it is more clear that this is a different action.

This doesn't really make sense. Having it styled as a normal link makes users actually _more_ prone to think of it as a link — especially on the main container where it appears alone above the comment section. Also, there's an after-click confirmation dialog as safeguard. If we're not going to comply with the `btn` / `btn-link` semantics, better to at least be consistent then (i.e every item as a `btn`)

public/gameInfo.php
```diff
@@ -1085,25 +1070,54 @@ function ResetProgress() {
...
+                    if ($numEarnedCasual > 0 || $numEarnedHardcore > 0) {
+                        echo "<div class='devbox mb-4'>";
+                        echo "<span onclick=\"$('#resetboxcontent').toggle(); return false;\">Reset Progress ▼</span>";
+                        echo "<div id='resetboxcontent' style='display: none'>";
+                        echo "<button class='btn btn-danger' type='button' onclick='ResetProgress()'>Reset your progress for this game</button>";
+                        echo "</div></div>";
+                    }
+                }
+
+                echo "<br>";
```
luchaos, line 1119: I'd use tailwind's margin classes instead of breaks. I was trying to reduce that `<br>` usage until they're gone.

app_legacy/Helpers/render/game.php
```diff
@@ -195,7 +195,7 @@ function RenderGameAlts($gameAlts, $headerText = null): void
 {
     echo "<div class='component gamealts'>";
     if ($headerText) {
-        echo "<h3>$headerText</h3>";
+        echo "<h2 class='like-h3'>$headerText</h2>";
```
luchaos, line 198: To be in line with tailwind's naming I'd go with `text-h3`.

luchaos (approved):

public/gameInfo.php
Krylan: Done! Removed a few br tags and used divs with margins to achieve the same effect.

app_legacy/Helpers/render/game.php
Krylan: You are right, `text-h3` would sound better and fit nice with rest of Tailwind's class names (I'm just used to name it `like-h3` as this is what I do often at work, where is no Tawilwind). Changed the names everywhere I used them.

PR #1328, opened 2023-01-22
show shared rank in global listing

Jamiras:
https://discord.com/channels/310192285306454017/645777658319208448/1065718996697096284

The rank shown in the list was just the row number and didsn't account for when multiple users had the same score. I've updated the logic to account for that.

image


EmoonX:
Perhaps we can also use ordering tiers? Just my two cents, as it barely makes a difference most of the time.

Something like:
1. Points
2. \# of masteries/completions
3. Retro points (if hardcore ranking)
4. \# of achievements

So list is ordered first based on whatever the user choose and then tier by tier. Finally, if everything is still tied, rank is shared.

luchaos (approved):

luchaos:
> Perhaps we can also use ordering tiers? Just my two cents, as it barely makes a difference most of the time.
> 
> Something like:
> 
> 1. Points
> 2. # of masteries/completions
> 3. Retro points (if hardcore ranking)
> 4. # of achievements
> 
> So list is ordered first based on whatever the user chooses and then tier by tier. Finally, if everything is still tied, rank is shared.

Good input - maintaining the point rankings will undergo an overhaul as soon as the database layer has been revamped. We can revisit your suggestion then 👍

PR #1331, opened 2023-01-26
Add `RAPatches` tag to games with mandatory patching

EmoonX:
Following [this suggestion](https://discord.com/channels/310192285306454017/1048299730670735421/1067468989887950888).

Adds a rendered tag to games which require unique patches for achievements to be enabled on them. This is done by checking the linked hashes table; in case _all_ hashes for a certain game have the `rapatches` label, it is assumed said game needs mandatory patching.

game
gamelist

The actual image used is an upscaled, monochromatic version of `rapatches.png` label:

image


app_legacy/Helpers/database/hash.php
```diff
@@ -36,6 +36,9 @@ function getHashListByGameID($gameID): array
 {
     sanitize_sql_inputs($gameID);
     settype($gameID, 'integer');
+    if ($hashList = Cache::get("hashList:$gameID")) {
+        return (array) $hashList;
+    }
```
EmoonX, line 41: Had to resort to cache since queries take forever when e.g. loading game lists. Is this the best way to proceed? I still need to apply cache updates to most instances that need it (hashes being added, removed, edited, etc); I need a greenlight before going ahead with it, though.

Jamiras: I would recommend just not showing it on the game list, or in tooltips, or anywhere other than the game page itself. It's not really part of the title, and should probably just be displayed in the sidebar:
image
image

Or as part of the game data:
image

Jamiras: Ideally, we'd want some way to link directly to the patch (similar to the suggestion for guides in #1051)

redwizard42: For games with mandatory patches (ie: strip out a debug mode that doesn't latch), I would kinda want the games list/searches (whereever the user might first see the game exists on the site/has a set) to show the requirement. Rather than the user grab a wrong hash and wonder what's up and have to go to the game page and notice it.  a nice friendly required patch list at the time the user first sees the game is a nice QoL, IMO

Jamiras: For people who don't plan to ever play the game, the additional flair is distracting. How many more tags do we want to put into titles before the whole site is just a bunch of bubbles? What about games that require a specific core/emulator to play? Should we put that information everywhere the game is referenced? Maybe we should put the publisher, or the genre up there too?

The fact that a game requires a patch is not something you need to know unless you're going to play the game. In that case, it makes sense to have it be prominent on the game page, but I don't see any reason I'd want to see it just looking through a list of available games.

Soulsderfuchs: We should be encouraging people to use the already existing linked patched page instead.

RALordAndrew: If used in game lists, the RAPatches tag would often appear alongside `Hack` and `Subset` tags, which already imply a patch requirement. The information conveyed is redundant in those cases.

redwizard42: I get the site clutter argument, but I think that unless you have _some other way_ to direct people to linked hashes that this sort of thing is going to be problematic. I guarantee you 95% or more of people just go and find roms and try them until one works and give up if they can't, possibly leaving a wall comment if we're lucky and probably never really scrutinize the game page part from maybe the achievement list.

Regardless, the other (and better, probably) way I could see of solving this is to be able to link a hash as unsupported, but belonging to a game, and tag it with useful tags so that when a user loads it, they receive a notice that achievements are unsupported in that hash, but also be able to note that a patch is required, or something else based on tag.  That would be a satisfactory, but non-trivial implementation that would direct people to what they need (patches, hash list, etc) at the point of failure where I suspect a lot of users give up.

redwizard42: > If used in game lists, the RAPatches tag would often appear alongside `Hack` and `Subset` tags, which already imply a patch requirement. The information conveyed is redundant in those cases.

Yeah, this should be displayed in non-obvious cases should it be a thing implemented.

EmoonX: > If used in game lists, the RAPatches tag would often appear alongside `Hack` and `Subset` tags, which already imply a patch requirement. The information conveyed is redundant in those cases.

The way I coded things, tags are already not applied to any subsets.

And I don't think \~Hack\~ sets in general should receive the tag `Requires RAPatches`, because when those need patches it isn't for complying with RA cheevo rules (after all, you can get such ROMs without manual patching from you-know-what means). Unless, of course, it is a hacked game that supposedly  _also_ needs a patch-fix. Do we have those around, at all?

EmoonX: > Regardless, the other (and better, probably) way I could see of solving this is to be able to link a hash as unsupported, but belonging to a game

I think this idea is great, all-around. I could even use the change to guarantee the code only picks up games which indeed need patch-fixes.

Of course, it would need some manual work to mark all the appropriate hashes, but per my counts here (the ones in the `Fix`/`Fixes` folders in RAPatches repo) we only have about 60 games or so. Pretty doable, as I see it.

Jamiras: #789 (and #717 to some extent) discuss tracking per-hash compatibility.

RALordAndrew: Super Mario Unlimited Deluxe might need a special RetroAchievements patch, since it includes built-in cheats. That would warrant a Patch Required tag, while other hacks hosted on RAPatches would not.

Jamiras: #708 proposes having a developer blurb on the game page where the developer could list important nuances. "This game requires applying a patch to disable the in-game debug menu" seems like a good example of a game-specific nuance.

EmoonX:
Considering all opinions presented, I suppose the best way to address this is a mix of #708 and RetroAchievements/RAIntegration#934 (two birds at once):
- You have a blurb where important info is presented (as a list) to the player
- Content can (and ideally should) also be displayed in in-game notifications
- For required games, we would simply have a list item like "This game requires a patch to be played (...)"

With that, the "warning" lays in a non-distracting place, and can be expanded to even subsets (which is useful to further avoid confusion to newcomers).

redwizard42:
> 

Those are all good.  Though I still maintain that a non-obvious patch being required is at *least* as important an up-front presentation as telling the user that a game is a hack, prototype, or homebrew (probably more important than the latter to, IMO). I know Soulsderfuchs is worried that it'll lead to a slippery slope of 'tag all the things and show them with the name!' but I personally think there's a very limited amount of things that warrant that level of visibility and that this is one of them.

PR #1332, opened 2023-01-26
fix 'Route [login] not defined' exception

Jamiras:
When trying to access pages guarded by `auth` middleware without being logged in (like the log viewer), an `AuthenticationException` is raised. The default handler for that tries to `redirect()->guest(route('login'))`, and since we don't provide a `login` route, another exception is raised and logged. This PR adds a clause to just return a 401 response instead of trying to redirect to the non-existent login page.


luchaos:
I have implementation of the login page on my todo list - I'd like to keep the exception handler as genereic as possible. If the log entries aren't to bothersome until then I'd like to skip this step?

Jamiras:
It's not that the log entries are bothersome; it's that they're telling us something is wrong. I finally figured out where this error was coming from and thought it would be good to clean it up. The only errors logged on 2/1 were this error.

As for keeping the exception handler generic - it's already handling four exception types for JSON requests. This just adds one exception handler for non-JSON requests.

This is a relatively small change that you can easily revert when you do implement the login page. I'd really like to remove the noise from the error logs so any outstanding issues can more easily be identified.

However, as the production exception handler just shows Cheevo with a generic "Server Error" message and this changes it to show Cheevo with a "Not Authorized" message, the user experience doesn't really change, so I'm not opposed to leaving things as-is if you feel that is the appropriate framework decision.

luchaos (approved):

luchaos:
That's fair - I'll remove it when I have the login page ready 👍

PR #1336, opened 2023-01-28
Simplify vite server config detection

luchaos:
Default to localhost if it's not a secured valet host. 
Update composer and npm dependencies.


luchaos:
@Krylan @EmoonX @Jamiras can you please try/review this change on your machines?

```shell
npm run dev
```

This should start the local vite watcher regardless of your local web server setup and allow to load assets in your browser from `http://localhost:5173` (or `http://localhost:64009` if you set the `VITE_PORT` environment variable accordingly).

Jamiras:
Fails spectacularly:
```
Error: ENOSPC: System limit for number of file watchers reached, watch '/home/brian/source/RAWeb/storage/app/media/Images/064857.png'
```
I think it's trying to monitor all of my downloaded badges
```
~/source/RAWeb/storage/app/media$ find . -name *.png | wc -l
38582
```

Jamiras:
Doesn't seem to work quite right if I temporarily remove that directory either.

First, it never runs to completion (I suspect that's expected as it has to have a process running to deal with the changes).

Second, I tried to change a color in one of the css files and this was the result:
image

Reverting the change didn't fix it. I had to abort the `npm run dev` and use `npm run build` to fix it.

luchaos:
> System limit for number of file watchers reached

Interesting! I didn't know it tries to watch all of it. I'll look into that.

> First, it never runs to completion (I suspect that's expected as it has to have a process running to deal with the changes).

Yup, that's how it should be - it watches for file changes and pushes atomic changes to the page via websockets.

> Reverting the change didn't fix it. I had to abort the `npm run dev` and use `npm run build` to fix it.

The idea is to have `npm run dev` up and running whilst working on the frontend (css, js). Just running `npm run build` once/after each pull should be enough otherwise.

luchaos:
Large volume directories are now ignored - can you please try again @Jamiras ?

Looking at the browser's console there might be errors regarding connecting to the vite host, it might require a refresh of the tab after restarting it.

In case that does not resolve it please let me know of the specific error message.

Jamiras:
Seems to not crash with the large media folders.

As for the refresh - the console error is:
```
GET http://localhost:5173/@vite/client net::ERR_SOCKET_NOT_CONNECTED
```
It's something with exposing the port through my VM. I tried within the VM and it does update appropriately.

luchaos:
> It's something with exposing the port through my VM. I tried within the VM and it does update appropriately.

Ah, yes - that makes perfectly sense then.

> Seems to not crash with the large media folders.

Awesome - so it's all working for you? I'll wait for the others to test it on their machines.

EmoonX:
I face the same issues as Jamiras's screenshot. Same happened when I tried to run `npm run dev` on the master branch, so it seems to be an error that persisted from before the update. I need to run `npm run build` individually to be able to correctly work with CSS.

Krylan:
@luchaos I pulled the branch and checked if `npm run dev` works: I had no problems with it, no interruptions with my localhost setup, resources were built on the fly, website was also rendered correctly.

luchaos:
@Krylan fantastic, thanks! Can you please try again with the latest change which does not explicitly set the host - should technically work just fine.

@EmoonX Thanks for the hint, I forgot to check Sail again. I pushed another fix for the Sail/Docker setup. Additionally, please check your `.env` file...

```env
APP_PORT=64000
VITE_PORT=64009
# ...
APP_URL=http://localhost:${APP_PORT}
# ASSET_URL is not set or set to the same as APP_URL
```

... then restart the containers:

```shell
sail up -d
```

Make sure npm dependencies are installed for the docker environment and restart Vite:

```shell
sail npm ci
sail npm run dev
```

Works via Sail for me now.

EmoonX:
@Jamiras It works for me now! 🥳
Setting up `VITE_PORT` and restarting Sail did the trick.

luchaos:
Jamira's setup runs in a dedicated VM, not a docker container. It's a similar yet different problem with port forwarding there if i understood correctly.

EmoonX:
@luchaos Aaah no 🫣 The ping was actually meant for you. There's something in my head that causes me to mix the names "luchaos" and "Jamiras". I knew it was bound to happen someday (sorry for the ping, @𝕵amiras).

Jamiras (approved):

PR #1343, opened 2023-02-04
build: unpin engines and adjust dockerfile node version

wescopeland:
Currently when following the readme exactly, npm will always throw a version mismatch warning when running `sail npm install`. This is mostly due to a lack of engine permissiveness as designated in the _package.json_ file.

This PR:
* Adjusts the Sail _Dockerfile_ to install Node 18 instead of incorrectly installing Node 16.
* Adjusts _package.json_ engine pinning to only require major semantic releases, not the exact minor or patch version. _npm_ now uses ">=" because v9 is shipping with Node 18 and, for the most part, is backwards compatible with v8.


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "raweb",
+    "name": "html",
```
luchaos, line 2: Interesting, where does `html` come from?

wescopeland: Odd I did not check the package-lock.json diff -- this seems to have been autogenerated when running `sail npm install`.

wescopeland: Figured this out - it's the default value because package.json does not have a name key (it must have at one point and was removed). I am pushing a fix.

luchaos (approved):

package-lock.json
luchaos: Ah, yes. I think that's fine to not have a name in there

wescopeland: I went ahead and added it, as someone else will no doubt encounter this in the future when any new dependency is added to the package.json file.

luchaos (approved):

package-lock.json
luchaos: I'll probably remove it again in a year because i find it superfluous 😅 
I'll try to remember why it was added again.

PR #1344, opened 2023-02-04
Fix logout request

luchaos:
Strict model rules apply (https://github.com/RetroAchievements/RAWeb/blob/master/app_legacy/Site/AppServiceProvider.php#L24).

Laravel tries to access `remember_token` on the legacy User model which does not exist.
Setting it to `null` resolves this for now.

### DIFF
--- a/app_legacy/Site/Models/User.php
+++ b/app_legacy/Site/Models/User.php
@@ -44,6 +44,11 @@ class User extends BaseModel implements AuthenticatableContract, AuthorizableCon
         'SaltedPass',
     ];
 
+    public function getRememberTokenName(): ?string
+    {
+        return null;
+    }
+
     public function getAuthPassword()
     {
         return $this->Password;


Jamiras (approved):
Not sure I understand the need for the change, but it does fix the problem.

luchaos:
> Not sure I understand the need for the change, but it does fix the problem.

To better explain: 
- Laravel assumes a `remember_token` column to be present in a users table by default (https://github.com/laravel/laravel/blob/9.x/database/migrations/2014_10_12_000000_create_users_table.php#L22).
- Laravel nukes the `remember_token` after a logout
- Laravel introduced a "strict" mode for models recently which, in it's strictest variant (which I enabled here https://github.com/RetroAchievements/RAWeb/blob/master/app_legacy/Site/AppServiceProvider.php#L24), requires all columns to be fetched explicitly
- Telling the framework that our legacy user does not have this field solves the problem

The remember token is used to allow for the usual "remember me" feature during login which allows to create an "infinetly" valid session cookie. We will tap into this feature as soon as we move to the new schema, disabling it for the legacy layer is appropriate.

PR #1350, opened 2023-02-06
Require 0-point achievements for `Latest Masters` table

EmoonX:
Address #1349.

So far, the "Latest Masters" table only checked for 100% points, not accounting for cases where the player missed one or more 0-point achievements (thus not 100% cheevos). This is in disagreement with other instances for showing mastery, e.g the "Completion Progress" table. As a side effect, one could possibly even repeatedly "update" their table spot by earning 0-point cheevos.

Fixed by simply changing the check from _total points_ to _amount of achievements_.

### DIFF
--- a/app_legacy/Helpers/database/player-game.php
+++ b/app_legacy/Helpers/database/player-game.php
@@ -393,18 +393,20 @@ function getGameTopAchievers(int $gameID): array
 
     $high_scores = [];
     $masters = [];
-    $mastery_score = 0;
+    $numAchievementsInSet = 0;
 
-    $query = "SELECT SUM(Points) AS Points FROM Achievements WHERE GameID = $gameID AND Flags = " . AchievementType::OfficialCore;
+    $query = "SELECT COUNT(*) AS NumAchievementsInSet
+        FROM Achievements
+        WHERE GameID = $gameID AND Flags = " . AchievementType::OfficialCore;
     $dbResult = s_mysql_query($query);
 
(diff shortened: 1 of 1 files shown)


Jamiras (approved):

luchaos (approved):

redwizard42:
Something with this is not working properly. Evidence at: https://retroachievements.org/game/15947

Jamiras:
There's some caching going on. It'll update sometime in the next 30 days (depends on when it was generated), or you can force it to update by demoting/repromoting an achievement, which I did with Vet Dev IX here: https://stage.retroachievements.org/game/15947

luchaos:
Cleared the whole application cache.

EmoonX:
Something is still not right... there are people who mastered DevQuest 004, yet the list is blank.

EmoonX:
I get it now. It sometimes breaks out of the loop before all masters can be added, since the query is currently being ordered just by points and not # of achievements.

```php
elseif (count($high_scores) == 10) {
    break;
}
```

redwizard42:
Maybe a dumb question, but was high scores also being affected?

EmoonX:
>Maybe a dumb question, but was high scores also being affected?

No, but it stops checking for masters on the first non-master following the high score table being filled up (10 entries). So taking DevQuest 004 as an example: since none of the first 10 players is actually a master, it results in an empty Latest Masters list.

PR #1353, opened 2023-02-10
docs(README): add encryption key section

wescopeland:
Currently, after doing a fresh clone of the project, running the setup instructions, and then running the command `sail up`, the web server will crash in an infinite loop with the following error:

```
raweb-laravel.test-1  |    Illuminate\Encryption\MissingAppKeyException 
raweb-laravel.test-1  | 
raweb-laravel.test-1  |   No application encryption key has been specified.
raweb-laravel.test-1  | 
raweb-laravel.test-1  |   at vendor/laravel/framework/src/Illuminate/Encryption/EncryptionServiceProvider.php:79
raweb-laravel.test-1  |      75▕     protected function key(array $config)
raweb-laravel.test-1  |      76▕     {
raweb-laravel.test-1  |      77▕         return tap($config['key'], function ($key) {
raweb-laravel.test-1  |      78▕             if (empty($key)) {
raweb-laravel.test-1  |   ➜  79▕                 throw new MissingAppKeyException;
raweb-laravel.test-1  |      80▕             }
raweb-laravel.test-1  |      81▕         });
raweb-laravel.test-1  |      82▕     }
raweb-laravel.test-1  |      83▕ }
```

This is resolved by generating a new key via `artisan key:generate`.

I considered hardcoding a key in _.env.example_, but considering this is directly related to encryption I figured that might be unwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,12 @@ It is a Laravel application ([TALL stack](https://tallstack.dev/)) including ind
 
 ## Requirements
 
--   Local web server
--   [PHP 8.0](http://php.net/manual/en/)
--   [Composer v2](https://getcomposer.org/) PHP dependency manager
--   [MySQL 8](https://dev.mysql.com/doc/refman/8.0/en/)
--   [Node.js 18](https://nodejs.org/)
-
+- Local web server
+- [PHP 8.0](http://php.net/manual/en/)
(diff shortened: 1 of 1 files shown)


luchaos:
The readme suggest to run `composer setup` during installation which takes care of generating the application key 👍

PR #1354, opened 2023-02-11
feat(home): redesign Statistics section

wescopeland:
This PR redesigns the Statistics section on the home page.

**Why?**
The section provides interesting data, but I have always found it difficult to read. Originally, I was just going to add number separators, but I pushed a bit further to use this as an exercise to get more comfortable with the dev experience available on RAWeb.

Notable changes:
- Rather than a paragraph, content is displayed in a small 4x4 grid view.
- Numbers now have separators (ie: 1,000,000 instead of 1000000).
- The latest registered user now is using a timeago-style relative date, so instead of showing a hard date like "11 Feb 02:31", something more approachable like "5 minutes ago" is displayed.

**BEFORE**
Screenshot 2023-02-10 at 9 52 36 PM

**AFTER**
Screenshot 2023-02-10 at 9 52 26 PM

### DIFF
--- a/resources/views/content/static-data.blade.php
+++ b/resources/views/content/static-data.blade.php
@@ -1,7 +1,7 @@
 <?php
 
-use LegacyApp\Site\Models\StaticData;
 use Illuminate\Support\Carbon;
+use LegacyApp\Site\Models\StaticData;
 
 /** @var ?StaticData $staticData */
 $staticData = StaticData::with('lastRegisteredUser')
--- a/resources/views/content/static-data.blade.php
+++ b/resources/views/content/static-data.blade.php
@@ -24,7 +24,7 @@ $lastRegisteredUser = $staticData['LastRegisteredUser'];
 $lastRegisteredUserAt = $staticData['LastRegisteredUserAt'];
 $totalPointsEarned = $staticData['TotalPointsEarned'];
 
-$lastRegisteredUserAtDate = new DateTime('2020-05-21 06:05:10');
+$lastRegisteredUserAtDate = new DateTime($lastRegisteredUserAt);
 $now = new DateTime();
 $interval = $lastRegisteredUserAtDate->diff($now);
 
@@ -38,7 +38,7 @@ $timeUnits = [
 ];
 foreach ($timeUnits as $format => $unit) {
     if ($interval->$format >= 1) {
-        $unit = $interval->$format > 1 ? "$unit"."s" : $unit;
+        $unit = $interval->$format > 1 ? $unit . "s" : $unit;
         // "%$format $unit ago" --> "5 minutes ago"
         $lastRegisteredUserTimeAgo = $interval->format("%$format $unit ago");
         break;
--- a/resources/views/content/static-data.blade.php
+++ b/resources/views/content/static-data.blade.php
@@ -49,30 +49,37 @@ if ($lastRegisteredUser == null) {
     $lastRegisteredUser = 'unknown';
 }
 ?>
-<div class="component statistics">
+<div class="component statistics !mb-0">
     <h3>Statistics</h3>
 
-    <div class="infobox mb-4">
+    <div class="infobox">
         <div class="w-full">
-            <div class="w-full h-16 mb-2 flex flex-col justify-center items-center">
-                <span class="text-2xl">{{ number_format($totalPointsEarned) }}</span>
-                <span>Points earned since 2nd March 2013</span>
-            </div>
-
-            <div class="grid grid-cols-2 gap-px">
+            <div class="grid grid-cols-2 gap-px mb-2">
                 <x-home-stats-embed label="Games" :count="$numGames" href="/gameList.php?s=1" />
                 <x-home-stats-embed label="Achievements" :count="$numAchievements" href="/gameList.php?s=2" />
                 <x-home-stats-embed label="Achievement Unlocks" :count="$numAwarded" href="/achievementList.php" />
                 <x-home-stats-embed label="Registered Players" :count="$numRegisteredPlayers" href="/userList.php" />
+            </div>
+
+            <div class="w-full h-16 flex flex-col justify-center items-center">
+                <span>Points earned since 2nd March 2013</span>
+                <span class="text-2xl">{{ number_format($totalPointsEarned) }}</span>
             </div>
         </div>
     </div>
 
     <div>
         @if($staticData->lastRegisteredUser)
-            The newest registered user is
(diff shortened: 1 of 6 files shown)


vite.config.js
```diff
@@ -29,6 +29,9 @@ export default defineConfig(({ mode }) => {
           'resources/css/app.css',
           'resources/js/app.js',
         ],
+        refresh: [
+          'resources/views/**'
+        ]
```
wescopeland, line 34: Vite was not reloading on edits to blades, making working with Tailwind kind of frustrating. This fixed the issue.

EmoonX:
I like the overall look of it from the screenshot. I have two suggestions:
- Put # of games as the topmost counter. Game support is a major "selling point" for the site, and as I see it, a number far more important than all the other ones
- Registered user text could be maybe reduced. Something like `Newest user: [user] (xx minutes/hours) ago`. This way it may even fit in one line

televandalist:
It's worth noting that the numbers are actually kinda misleading because it includes stuff in unofficial and events. It says 6410 games at this moment, but we're still about 55-60 sets away from reaching 6000.

wescopeland:
> I like the overall look of it from the screenshot. I have two suggestions:
> 
> - Put # of games as the topmost counter. Game support is a major "selling point" for the site, and as I see it, a number far more important than all the other ones
> 
> - Registered user text could be maybe reduced. Something like `Newest user: [user] (xx minutes/hours) ago`. This way it may even fit in one line

Counter proposal for # of games as topmost counter suggestion: how about I move number of points earned below the boxes instead of above. The stat boxes would now be the top element of the section. 

Motivation: points earned cannot be a clickable link and cannot fit in one of the boxes. This solution keeps Games as the first data point in the section without breaking the design. 

I like your idea for getting the recent player stat down to one line. I will implement that suggested change.

vite.config.js
luchaos: This used to be in there but was removed to some performance problems with chokidar which tested too many files. Has to be tested within a Linux VM again to make sure the problem didn't come back

wescopeland: I can look into setting this to be enabled if an env var is truthy, and set that new var to be falsy by default. That way there's an easy way to toggle it, and it's documented why it's disabled by default.

luchaos: I don't think an additional env var is necessary - given that it just works in a VM. @Jamiras if you find time, could you please check if the change in the vite config causes any of the previous problems in your VM?

If it doesn't then it can stay imo. Unfortunately I haven't found time to set up a VM on my windows machine yet.

luchaos:
None of those numbers are accurate iirc. Emphasizing them might be misleading in general. Then again, the visual change is good - those numbers still need fixing though.

Jamiras:
The prototype put the text above the data for attributes.
image


Is that something we'd want to do here?
image

wescopeland:
All great feedback here. It is much appreciated!

After some tweaking, here is where things currently stand:
Screenshot 2023-02-11 at 2 52 11 PM

* @EmoonX Since # of the games is the most important thing, I have moved points earned back to the bottom of the stats.
* @EmoonX I fought a lot to get newest user down to a single row but could not come up with a way to do it that gave me any confidence. The problem primarily comes from usernames being an unpredictable length. Even with a tiny font size, if a user's name has more than ~16 characters, it overflows to the next row. As a happy medium, I was able to condense this down a bit though.
* @Jamiras I have swapped the label and number positions.
* @televandalist @luchaos I have opened another issue with regards to the data being inaccurate. Fixing this extends beyond my present knowledge of the system - if possible I would like to keep the scope of this PR as visual in nature.

EmoonX:
>@EmoonX Since # of the games is the most important thing, I have moved points earned back to the bottom of the stats

Looking great. Though I'd suggest moving the `Achievements Unlocks` counter last. It's the biggest number and the one that ultimately depends on all the other three entities (games, achievements and players).

>@EmoonX I fought a lot to get newest user down to a single row but could not come up with a way to do it that gave me any confidence. The problem primarily comes from usernames being an unpredictable length. Even with a tiny font size, if a user's name has more than ~16 characters, it overflows to the next row. As a happy medium, I was able to condense this down a bit though.

Honestly, this is even better. With the bonus of being responsive to different-length usernames.

wescopeland:
> Looking great. Though I'd suggest moving the `Achievements Unlocks` counter last (in the grid). It's the biggest number and the one that ultimately depends on all the other three entities (games, achievements and players).

Good point! I have made this change.

Jamiras:
The Achievements link goes to the Games list (sorted by achievements), and the Achievements Unlocked link just goes to the Achievement list. I realize that's consistent with the current links, but I think it would make more sense to have the Achievements link go to the Achievement list, and the Achievements Unlocked link go to the Recent Masteries page.

wescopeland:
> The Achievements link goes to the Games list (sorted by achievements), and the Achievements Unlocked link just goes to the Achievement list. I realize that's consistent with the current links, but I think it would make more sense to have the Achievements link go to the Achievement list, and the Achievements Unlocked link go to the Recent Masteries page.

Agreed - updated.

resources/views/content/static-data.blade.php
```diff
@@ -24,32 +24,62 @@
...
+    "s" => "second",
+];
+foreach ($timeUnits as $format => $unit) {
+    if ($interval->$format >= 1) {
+        $unit = $interval->$format > 1 ? $unit . "s" : $unit;
+        // "%$format $unit ago" --> "5 minutes ago"
+        $lastRegisteredUserTimeAgo = $interval->format("%$format $unit ago");
+        break;
+    }
+}
```
Jamiras, line 46: Carbon has a way to generate these nicely formatted elapsed times:
https://github.com/RetroAchievements/RAWeb/pull/1321/files#diff-81cb00a9d43fff3bd73a0f74426506b9ddeec2eab6d9d72d210d26a9381fa39eR13

vite.config.js
Jamiras: Seems to be okay. There's only 124 files in `resources/views`. My problem was files in storage, where I have over 60000.

luchaos: Perfect, thanks for checking!

resources/views/content/static-data.blade.php
wescopeland: That saves a lot of code, thanks. Fixed.

I had to add a truthiness check in the blade to keep a test case from failing.

Jamiras (approved):

PR #1357, opened 2023-02-11
Add tests for GetActiveClaims & GetUserClaims

luchaos:


deptrac.yaml
```diff
@@ -238,6 +239,7 @@ deptrac:
       - Support
     # Legacy Database
     LegacyDatabase:
+      - LegacyCommunity
```
Jamiras, line 242: This reads like a circular reference.

database/factories/Legacy/AchievementSetClaimFactory.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        return [
+            'User' => $this->fakeUsername(),
+            'GameID' => 0,
+            'ClaimType' => ClaimType::Primary,
+            'SetType' => ClaimSetType::NewSet,
+            'Status' => ClaimStatus::Active,
+            'Extension' => ClaimStatus::Active,
+            'Special' => ClaimSpecial::None,
+            'Finished' => Carbon::now(),
```
Jamiras, line 36: `insertClaim` initializes Finished to "now + 3 months".

database/factories/Legacy/AchievementSetClaimFactory.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public function definition(): array
+    {
+        return [
+            'User' => $this->fakeUsername(),
+            'GameID' => 0,
+            'ClaimType' => ClaimType::Primary,
+            'SetType' => ClaimSetType::NewSet,
+            'Status' => ClaimStatus::Active,
+            'Extension' => ClaimStatus::Active,
```
Jamiras, line 34: Extension is just a number, it's a count of how many times the claim has been extended. The default should be 0.

tests/Feature/Api/V1/AchievementSetClaimTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $claim = AchievementSetClaim::factory()->create([
+            'GameID' => $game->ID,
+            'User' => $this->user->User,
+        ]);
+
+        $this->get($this->apiUrl('GetUserClaims', ['u' => $this->user->User]))
+            ->assertSuccessful()
+            ->assertExactJson([
+                [
+                    'ClaimType' => ClaimType::Primary,
```
Jamiras, line 52: I would assume these asserts should use the expected integers, rather than the enums, to catch if the enums gets reordered. However, as the values are stored in the database, that seems unlikely.

app_legacy/Community/Enums/ClaimFilters.php
```diff
@@ -34,13 +34,10 @@ abstract class ClaimFilters
...
     // Filter to show all Complete Primary claims
-    public const AllCompletedPrimaryClaims = self::AllFilters & ~self::ActiveClaim
-                                            & ~self::DroppedClaim
-                                            & ~self::CollaborationClaim;
+    public const AllCompletedPrimaryClaims = self::AllFilters & ~self::ActiveClaim & ~self::DroppedClaim & ~self::CollaborationClaim;
 
     // Default filter is everything except Complete and Dropped claims
-    public const Default = self::AllFilters & ~self::CompleteClaim
-                                            & ~self::DroppedClaim;
+    public const Open = self::AllFilters & ~self::CompleteClaim & ~self::DroppedClaim;
```
Jamiras, line 40: `AllOpenClaims` to be consistent with `AllCompletedPrimaryClaims`? Or maybe `AllActiveClaims` to also be consistent with the enum.

app_legacy/Helpers/database/set-claim.php
```diff
@@ -232,13 +232,13 @@ function getClaimData(int $gameID, bool $getFullData = true): array
...
-    int $gameID = 0,
+    int $gameID = null,
     int $claimFilter = ClaimFilters::AllFilters,
     int $sortType = ClaimSorting::ClaimDateDescending,
     bool $getExpiringOnly = false,
     ?string $username = null,
-    int $offset = 0,
-    int $limit = 50
+    int $offset = null,
+    int $limit = null
```
Jamiras, line 241: Shouldn't these be `?int`?

deptrac.yaml
luchaos: True. It allows both layers to talk to each other, which technically is bad design. Not much of an issue in this case though.

The legacy database layer may talk to its community layer while the community layer may also define stuff in the database layer.

database/factories/Legacy/AchievementSetClaimFactory.php
luchaos: Thanks! Missed that.

database/factories/Legacy/AchievementSetClaimFactory.php
luchaos: Thanks! Missed that, too.

tests/Feature/Api/V1/AchievementSetClaimTest.php
luchaos: Yeah, a similar thing could be said about the point values etc in the achievement endpoints where the same value of the models is tested again which makes those not the most robust.
I'd say it's more important for now to check if the schema, the keys in particular, are present. That caused the biggest problem before (removal of the user ID) plus it seems with PHP 8.0 we cannot be too sure about the proper casting within the sqlite driver.
We can revisit this as soon as we're running on PHP 8.2 and be more strict about the actual values?

app_legacy/Helpers/database/set-claim.php
luchaos: Good catch, thanks!
PHPStan in level 5 (current) does not catch those. I'd love to increase the level to at least 6/7 at some point which would highlight such inconsistencies, too.

app_legacy/Community/Enums/ClaimFilters.php
luchaos: Changed to `AllActiveClaims` in https://github.com/RetroAchievements/RAWeb/pull/1357/commits/ea2e99e57e2d9c9f2502dcc9f69bcc1cd212d022

tests/Feature/Api/V1/AchievementSetClaimTest.php
```diff
@@ -0,0 +1,74 @@
...
+                    'ConsoleName' => $system->Name,
+                    'Created' => $claim->Created->__toString(),
+                    'DoneTime' => $claim->Finished->__toString(),
+                    'Extension' => 0,
+                    'GameID' => $game->ID,
+                    'GameIcon' => '/Images/000001.png',
+                    'GameTitle' => $game->Title,
+                    'ID' => $claim->ID,
+                    'MinutesLeft' => Carbon::now()->diffInMinutes(Carbon::now()->addMonths(3)),
```
Jamiras, line 65: It seems like this should be `Carbon::now()->diffInMinutes($game->Finished)`, but that errors:
```
  -    'MinutesLeft' => 129599,
  +    'MinutesLeft' => '128160.0',
```

Jamiras (approved):

tests/Feature/Api/V1/AchievementSetClaimTest.php
luchaos: Ah, yes. `Carbon::now()->diffInRealMinutes()` does the trick.

app_legacy/Helpers/database/set-claim.php
```diff
@@ -15,39 +16,36 @@
...
             return false;
         }
 
-        $query = "
-            INSERT INTO
-                SetClaim (`User`, `GameID`, `ClaimType`, `SetType`, `Status`, `Extension`, `Special`, `Created`, `Finished` ,`Updated`)
-            VALUES
-                ('$user', '$gameID', '$claimType', '$setType', '" . ClaimStatus::Active . "', '0', '$special', NOW(), DATE_ADD(NOW(), INTERVAL 3 MONTH), NOW())";
+        $now = Carbon::now();
+        $finished = Carbon::now()->addMonths(3);
```
luchaos, line 27: Using Carbon for timestamps makes it mysql/sqlite agnostic; sqlite doesn't know `DATE_ADD(NOW(), INTERVAL 3 MONTH)`

app_legacy/Helpers/database/set-claim.php
```diff
@@ -15,39 +16,36 @@
  */
 function insertClaim(string $user, int $gameID, int $claimType, int $setType, int $special, int $permissions): bool
```
luchaos, line 17: Refactored `insertClaim` and `getActiveClaimCount` while at it because I thought there's an issue in those because of the minutes remaining calculation.

tests/Feature/Api/V1/AchievementSetClaimTest.php
```diff
@@ -43,11 +44,16 @@ public function testGetUserClaims(): void
...
+
+        insertClaim(
+            $this->user->User,
+            $game->ID,
+            ClaimType::Primary,
+            ClaimSetType::NewSet,
+            ClaimSpecial::None,
+            Permissions::Developer
+        );
+        $claim = AchievementSetClaim::first();
```
luchaos, line 56: Testing the `insertClaim()` function specifically was not the goal here, the `AchievementSetClaim::factory()` would suffice, however, since I refactored this method already i guess it should stay like this for now until the public request files get their own tests.

app_legacy/Helpers/database/set-claim.php
```diff
@@ -15,39 +16,36 @@
...
-            return true;
-        }
-    } else { // Collaboration claim
-        // For a collaboration claim we want to use the same Finished time as the primary claim
         $query = "
             INSERT INTO
                 SetClaim (`User`, `GameID`, `ClaimType`, `SetType`, `Status`, `Extension`, `Special`, `Created`, `Finished` ,`Updated`)
             VALUES
-                ('$user', '$gameID', '$claimType', '$setType', '" . ClaimStatus::Active . "', '0', '" . ClaimSpecial::None . "', NOW(),
-                (SELECT Finished FROM (SELECT Finished FROM SetClaim WHERE GameID = '$gameID' AND Status = " . ClaimStatus::Active . " AND ClaimType = " . ClaimType::Primary . ") AS sc),
```
Jamiras, line 42: I believe this is intended to share the Finished date with the primary claim when making a collaboration claim. Collaborating two months into a claim shouldn't give you any extra time.

Jamiras: Nevermind, see it below. Diff confused me.

PR #1359, opened 2023-02-12
fix: remediate large layout shift on legacy pages

wescopeland:
This PR fixes an issue that I believe was introduced as part of the V2 rollout. There is now a substantial cumulative layout shift on legacy pages that have leftcolumn + rightcolumn layouts. 

I am still not totally clear on why this happens, much to my disappointment. I believe the right column renders in the DOM after the left column, but that is just my guess. Regardless, this should fix the issue while still preserving full-width layouts when they occur.

Note that this does not fix the issue on modern pages. I think only the home page and TOS fall under this umbrella presently. However since [layout shift is a core web vital](https://web.dev/cls/), this feels essential to fix for the sake of SEO.

**BEFORE**
Broken3

**AFTER**
Fixed2

### DIFF
--- a/resources/css/legacy.css
+++ b/resources/css/legacy.css
@@ -3,8 +3,9 @@
   padding: 0;
   position: relative;
   min-height: 450px;
-  display: flex;
-  grid-gap: 1rem;
+  display: grid;
+  column-gap: 1rem;
+  grid-template-columns: 1.329fr 0.671fr;
 }
 
 @media only screen and (max-width: 1023px) {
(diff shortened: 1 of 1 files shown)


resources/css/legacy.css
```diff
@@ -3,8 +3,9 @@
   padding: 0;
   position: relative;
   min-height: 450px;
-  display: flex;
-  grid-gap: 1rem;
+  display: grid;
+  column-gap: 1rem;
+  grid-template-columns: 1.329fr 0.671fr;
```
luchaos, line 8: It used to be grid, but caused trouble when widening the overall container width.
While content shift/SEO is important, this site is not quite in a state where this is of the highest priority imo.

luchaos (approved):

PR #1352, opened 2023-02-10
Navbar Update

televandalist:
A couple things were bugging me and one thing led to another... 🤷‍♂️ 

Changes to the `Achievements` dropdown:
  - Adds `New Sets & Revisions` which leads [here](https://retroachievements.org/claimlist.php?s=9&f=8109).
  - Adds `Sets in Progress` and removes `Claim List` from the `Community` dropdown.
  - Adds `Most Requested Sets` and removes it from the `Manage` dropdown.

Changes to the `Community` dropdown:
  - Adds `RANews Home` and `RANews Issues List`
  - Adds `RAGuides Wiki`
  - Adds `Community Events` which links to [the hub](https://retroachievements.org/game/3105).
  - Adds `Developer Events` which links to [the hub](https://retroachievements.org/game/3273).
  - Adds `Hub List` which is [this page](https://retroachievements.org/gameList.php?s=6&c=100&f=1)
  - Adds `Central Hub` and links to [the hub](https://retroachievements.org/game/6914).
  - Removes `- Community` since it led to the same page as `Forums`
  - Removes `- Developers` since the Developer forum is rarely used for anything and shows up in the forum index anyway.
  - Renames `Forums` to `Forum Index`
  - Renames `Recent Posts` to `Recent Forum Posts`
  - Renames the tacky `+- Competitions` link to `Event Forums`
  - Adds a second column to the dropdown since it was getting crowded.

Adds a `Contact` dropdown with links to send a message to the following accounts:
  - [RAdmin](https://retroachievements.org/createmessage.php?t=RAdmin)
  - [DevCompliance](https://retroachievements.org/createmessage.php?t=DevCompliance)
  - [QATeam](https://retroachievements.org/createmessage.php?t=QATeam)
  - [RANews](https://retroachievements.org/createmessage.php?t=RANews)
  - [RAEvents](https://retroachievements.org/createmessage.php?t=RAEvents)
  - [DevQuest](https://retroachievements.org/createmessage.php?t=DevQuest)
  - [TheUnwanted](https://retroachievements.org/createmessage.php?t=TheUnwanted)
  - [QualityQuest](https://retroachievements.org/createmessage.php?t=QualityQuest)

PR #1362, opened 2023-02-12
add Contact Us page

Jamiras:
Provides a more detailed interaction with the unique accounts over the proposed dropdown menu in #1352.

Removes the "RetroAchievements" section from the footer, as it just linked to the "All Games", "All Achievements", and "Downloads" pages, which are readily accessible from the main menu.

Replaces the "About" header for the Legal/Privacy Policy links with "RetroAchievements", and adds a "Contact Us" link there.

image

Clicking on the "Contact Us" link goes to a page with links to contact each of the unique accounts, and reasons why those individual accounts should be contacted.

image

### DIFF
--- a/resources/views/contacts.blade.php
+++ b/resources/views/contacts.blade.php
@@ -1,0 +1,100 @@
+<x-app-layout :page-title="__('Contact Us')">
+    <div class="lg:grid grid-cols-2 gap-5">
+        <div>
+            <x-section>
+                <h2>Admins and Moderators</h2>
+
+                <p class="mb-3">
+                    <a href='/createmessage.php?t=RAdmin'>Send a message to RAdmin</a> for:
+                    <ul>
+                        <li>&#x2022; Reporting offensive behavior.</li>
+                        <li>&#x2022; Reporting copyrighted material.</li>
+                        <li>&#x2022; Reporting cheating to be investigated.</li>
+                        <li>&#x2022; Requesting to be untracked.</li>
+                    </ul>
+                </p>
+            </x-section>
+
+            <x-section>
+                <h2>Developer Compliance</h2>
+
+                <p class="mb-3">
+                    <a href='/createmessage.php?t=DevCompliance'>Send a message to DevCompliance</a> for:
+                    <ul>
+                        <li>&#x2022; Requesting set approval or early set release.</li>
+                        <li>&#x2022; Reporting achievements or sets with unwelcome concepts.</li>
+                        <li>&#x2022; Reporting sets failing to cover basic progression.</li>
+                    </ul>
+                </p>
+            </x-section>
+
+            <x-section>
+                <h2>Quality Assurance</h2>
+
+                <p class="mb-3">
+                    <a href='/createmessage.php?t=QATeam'>Send a message to QATeam</a> for:
+                    <ul>
+                        <li>&#x2022; Reporting a broken set, leaderboard, or rich presence.</li>
+                        <li>&#x2022; Reporting achievements with grammatical mistakes.</li>
+                        <li>&#x2022; Requesting a set be playtested.</li>
+                        <li>&#x2022; Hash compatibility questions.</li>
+                        <li>&#x2022; Hub organizational questions.</li>
+                        <li>&#x2022; Getting involved in a QA sub-team.</li>
+                    </ul>
+                </p>
+            </x-section>
+        </div>
+
+        <div>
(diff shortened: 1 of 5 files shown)


resources/views/contacts.blade.php
```diff
@@ -0,0 +1,100 @@
...
+                <p class="mb-3">
+                    <a href='/createmessage.php?t=QATeam'>Send a message to QATeam</a> for:
+                    <ul>
+                        <li>&#x2022; Reporting a broken set, leaderboard, or rich presence.</li>
+                        <li>&#x2022; Reporting achievements with grammatical mistakes.</li>
+                        <li>&#x2022; Requesting a set be playtested.</li>
+                        <li>&#x2022; Hash compatibility questions.</li>
+                        <li>&#x2022; Hub organizational questions.</li>
+                        <li>&#x2022; Getting involved in a QA sub-team.</li>
+                    </ul>
```
wescopeland, line 43: nit: Embedding unicode `&#x2022;` characters on each list item should be achievable automatically with CSS & Tailwind at the UL level.

```html
<ul class="list-disc marker:text-[color:var(--variable-name)]>
```

wescopeland: > I was surprised it was necessary to add those in the first place.

Yeah Tailwind's built-in reset does this automatically; not sure if it's being applied or not.

> Your suggestion puts the bullets in the margin

Oops - I forgot that is the default behavior of bullets inside a UL. I bet this fixes it:

```html
<ul class="list-disc list-inside"> // probably don't need the marker declaration
</ul>
```

resources/views/contacts.blade.php
```diff
@@ -0,0 +1,100 @@
+<x-app-layout :page-title="__('Contact Us')">
+    <div class="lg:grid grid-cols-2 gap-5">
+        <div>
+            <x-section>
+                <h2>Admins and Moderators</h2>
+
+                <p class="mb-3">
```
wescopeland, line 7: Sorry, one final comment nit on this that I just noticed.

Technically this violates the HTML5 spec, as it's inserting a block-level element into a P tag. I recommend changing these wrapper `<p>` tags to divs.

ref: https://www.w3.org/TR/html401/struct/text.html#h-9.3.1
> The [P](https://www.w3.org/TR/html401/struct/text.html#edef-P) element represents a paragraph. It cannot contain [block-level elements](https://www.w3.org/TR/html401/struct/global.html#block-inline) (including [P](https://www.w3.org/TR/html401/struct/text.html#edef-P) itself).

Jamiras: That was copied from `terms.blade.php`, which I used as a template.

https://github.com/RetroAchievements/RAWeb/blob/5a76467744180fc87cedc54abbb6bf04cdf858bc/resources/views/terms.blade.php#L4-L8

wescopeland: Though the browser interprets it okay, It is not valid HTML and should be fixed there too. Maybe it makes sense to keep this as-is and update in a follow-up PR. I don't mind taking this on once this PR is merged 👍

luchaos: it's separate over there ;)

https://github.com/RetroAchievements/RAWeb/blob/5a76467744180fc87cedc54abbb6bf04cdf858bc/resources/views/terms.blade.php#L52-L63

Jamiras: Oh, I misread the problem. I just saw "change `<p>` to `<div>`" and missed the complaint about having the list in the `<p>`.

Should the `<ul>` in terms be attributed with "list-disc list-inside" too?

luchaos: > Should the <ul> in terms be attributed with "list-disc list-inside" too?

Ah, yes, good catch. I missed that those should be displayed as lists, too.

luchaos:
> Removes the "RetroAchievements" section from the footer, as it just linked to the "All Games", "All Achievements", and "Downloads" pages, which are readily accessible from the main menu.

It's usually fine to have redundant links in the doormat/footer navigation; to act as some sort of quick sitemap. It's fine to remove them though since our top navigation is sticky and scrolls to the bottom.

> Replaces the "About" header for the Legal/Privacy Policy links with "RetroAchievements", and adds a "Contact Us" link there.

I think I'd prefer the RetroAchievements section to be the first one. Then again, reading the documentation should come before contacting us.

luchaos (approved):

PR #1363, opened 2023-02-12
feat(top-card): redesign guest login

wescopeland:
This PR makes minor design and a11y changes to the guest login component displayed on the top right of the site. 

The component now responds better to all screen widths. 

A longstanding a11y bug in the existing component was also fixed: the input fields were confusing to screen readers. Up to now, screen readers have only been able to determine that the password field was, in fact, a password field. The username field was undeterminable, so easily confused for username / email address / etc. To fix this, I have attached invisible semantic labels to both fields.

**BEFORE**
_Mobile_
Mobile Before

_Desktop_
Desktop Before

**AFTER**
_Mobile_
Mobile After

_Desktop_
Desktop After

### DIFF
--- a/resources/views/components/user/top-card.blade.php
+++ b/resources/views/components/user/top-card.blade.php
@@ -31,10 +31,6 @@ $user = request()->user();
                 <button type="submit" name="submit" class="lg:col-start-2 lg:row-span-2 flex items-center justify-center p-2 w-full lg:w-24">Log In</button>
             </div>
         </form>
-
-        <!-- <div class="grid grid-cols-2 lg:flex lg:justify-between text-center">
-            <a href="/createaccount.php">Register</a>
-        </div> -->
     @endguest
     @auth
         <div class="flex justify-between items-start gap-2">
--- a/resources/views/components/user/top-card.blade.php
+++ b/resources/views/components/user/top-card.blade.php
@@ -26,7 +26,7 @@ $user = request()->user();
             </div>
         </form>
 
-        <div class="grid grid-cols-2 text-center">
+        <div class="grid grid-cols-2 lg:flex lg:justify-between text-center">
             <a href="/createaccount.php">Register</a>
             <a href="/resetPassword.php">Forgot Password?</a>
         </div>
--- a/resources/views/components/user/top-card.blade.php
+++ b/resources/views/components/user/top-card.blade.php
@@ -12,24 +12,29 @@ $user = request()->user();
     @guest
         <form class="mb-2" action="/request/auth/login.php" method="post">
             @csrf
-            <div class="grid lg:flex gap-2 mb-2">
+            <div class="grid gap-1 mb-2">
                 <div>
                     <label for="username-input" class="sr-only">Username</label>
-                    <input class="w-full p-2" type="text" placeholder="Username" id="username-input" name="u">
+                    <input class="w-full py-[6px]" type="text" placeholder="Username" id="username-input" name="u">
                 </div>
-                <div>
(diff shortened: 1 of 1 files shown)


Jamiras:
I honestly like the original layout better (at least for desktop). The username and password fields are bigger, and the "forget password" link is more closely associated to the password field.

redwizard42:
> I honestly like the original layout better (at least for desktop). The username and password fields are bigger, and the "forget password" link is more closely associated to the password field.

Yeah, I like the new mobile layout fine, but new desktop looks off from my expectations of a login form.

wescopeland:
> ... new desktop looks off from my expectations of a login form.

I could say the same for what's there right now; it is helpful to elaborate more on these expectations. Currently, this component is a two-column grid with a random missing space to the right of the Username field. Login is a button with text floated to the left that spans the full width of its cell. On the MD (iPad / netbook) breakpoint the button grows extraordinarily large and spans the page.

> I honestly like the original layout better (at least for desktop). The username and password fields are bigger, and the "forget password" link is more closely associated to the password field.

This makes sense. I am going to keep chewing on this one but for the time being revert back to the original for desktop and keep the new layout for mobile.

redwizard42:
> I could say the same for what's there right now; it is helpful to elaborate more on these expectations. Currently, this component is a two-column grid with a random missing space to the right of the Username field. Login is a button with text floated to the left that spans the full width of its cell. On the MD (iPad / netbook) breakpoint the button grows extraordinarily large and spans the page.

I think it's probably just the location of the forgot password link. If it were under the password textbox, I bet my brain wouldn't be flagging the layout as 'wrong'.

wescopeland:
> I think it's probably just the location of the forgot password link. If it were under the password textbox, I bet my brain wouldn't be flagging the layout as 'wrong'.

This makes sense & seems to echo Jamiras' feedback too. Valid point - I'm going to think more carefully on how to shape the desktop layout for the component up.

wescopeland:
@Jamiras @redwizard42 

I've taken another crack at designing for desktop based on your feedback:

Screenshot 2023-02-14 at 4 39 51 PM

Let me know if this is good, bad, or getting closer.

redwizard42:
> @Jamiras @redwizard42
> 
> I've taken another crack at designing for desktop based on your feedback:
> 
> Screenshot 2023-02-14 at 4 39 51 PM
> 
> Let me know if this is good, bad, or getting closer.

Layout overall seems acceptable.

wescopeland:
Been collaborating a lot with @EmoonX on this, we paired & landed on another design revision. The previous version seemed to be moving in the right direction, but had a slightly higher px height than what is currently on the site. This revision keeps the same height as what's in prod today.

Screenshot 2023-02-14 at 6 32 16 PM

Screenshot 2023-02-14 at 6 31 59 PM

Jamiras (approved):

luchaos:
I'm still not sure what the exact goal was 😅 I think the original layout was fine.
The login button acting as form submit button now is mixed in with non-form elements and places it in weird position where it's not on the same vertical position or below the last input. Feels odd to me but I guess it works just fine like this, too.

My hope is that some day the whole top auth/user card will be removed and replaced with a solution that makes better use of the navigation bar. I have a few things prepared for that with the permissions update (after database refactoring and tests) as this will have an effect on a few things in the navigation.

EmoonX:
@luchaos While not necessarily the main goal, container's height is now shorter (100px including padding) and consistent (matches the profile card which replaces it after logged in). It wouldn't be doable by having more than two rows to display.

Issue #1365, opened 2023-02-14
Hover-driven tooltips produce a poor user experience on iOS


wescopeland:
**Describe the bug**
Game and achievement tooltips appear when hovering over a game or achievement on the site. While these can show interesting data for desktop users, the results on mobile can be frustrating and sometimes broken.

For iOS Safari, most links on the site actually require a double tap. A single tap opens a tooltip, [the 2nd tap closes the tooltip and opens the link](https://imgur.com/3yXoD7P).

It is also possible to enter a state where the tooltip doesn't close and [it follows you around on the screen as you scroll the page](https://imgur.com/OP1VAOO).

Rather than try to "fix" the tooltip behavior, I think it makes more sense to disable it altogether on touch events. They are driven by hover behavior which rarely exists on mobile at the XS breakpoint anyway.

I have some ideas for possible fixes:
* Check with JavaScript using `window.matchMedia()` if the user is on the XS breakpoint and if so, disable tooltips.
* Differentiate in the tooltip between hover events and touch events, and disable the tooltip on touch events. I believe this is the preferable approach but is probably more involved.
* Implement something like [FastClick](https://github.com/ftlabs/fastclick).
* `ontouchstart="this.removeEventListener('mouseover', Tip)"`

Further discussion: https://stackoverflow.com/questions/2741816/is-it-possible-to-force-ignore-the-hover-pseudoclass-for-iphone-ipad-users

**To Reproduce**
On mobile iOS Safari, attempt to open any game page through a direct link to the game on the RA site, such as from the home page.

**Expected behavior**
The tooltips should not display on iOS when there is a touch event.

**Videos**
https://imgur.com/3yXoD7P
https://imgur.com/OP1VAOO

**Smartphone**
 - Device: iPhone 12
 - OS: iOS 16 (though all are valid)
 - All browsers supported by iOS

PR #1367, opened 2023-02-15
fix: turn off tooltips for touch events

wescopeland:
This PR resolves https://github.com/RetroAchievements/RAWeb/issues/1365. There are various issues with tooltips on mobile because taps are interpreted as hover events. I have confirmed these issues occur not only on iOS, but occur on some Android browsers as well.

A new event listener, `ontouchstart`, is attached to DOM elements responsible for opening tooltips. This listener will always fire before `onmouseover`. If it does fire, a flag is set and tooltips will not open on tap.

This solution has the benefit of not breaking mobile devices that are using trackpads (which is increasingly common with iPads).

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -1006,7 +1006,7 @@ sanitize_outputs(
                     echo "</div>";
 
                     echo "<script>var {$containername}tooltip = \"$tooltip\";</script>";
-                    echo "<div style='float: left; clear: left' onmouseover=\"Tip({$containername}tooltip)\" onmouseout=\"UnTip()\">";
+                    echo "<div style='float: left; clear: left' ontouchstart=\"mobileSafeTipEvents.touchStart()\" onmouseover=\"mobileSafeTipEvents.mouseOver({$containername}tooltip)\" onmouseout=\"UnTip()\">";
                     echo "<span class='$labelname'>$labelcontent</span>";
                     echo "</div>";
 
--- a/public/js/all.js
+++ b/public/js/all.js
@@ -121,6 +121,31 @@ function replaceAll(find, replace, str) {
   return str.replace(new RegExp(find, 'g'), replace);
 }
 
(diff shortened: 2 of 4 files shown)


public/gameInfo.php
```diff
@@ -1006,7 +1006,7 @@ function ResetProgress() {
                     echo "</div>";
 
                     echo "<script>var {$containername}tooltip = \"$tooltip\";</script>";
-                    echo "<div style='float: left; clear: left' onmouseover=\"Tip({$containername}tooltip)\" onmouseout=\"UnTip()\">";
+                    echo "<div style='float: left; clear: left' ontouchstart=\"mobileSafeTipEvents.touchStart()\" onmouseover=\"mobileSafeTipEvents.mouseOver({$containername}tooltip)\" onmouseout=\"UnTip()\">";
```
Jamiras, line 1009: A similar change also needs to be made in activePlayersBootstrap.js

public/js/all.js
```diff
@@ -121,6 +121,31 @@ function replaceAll(find, replace, str) {
   return str.replace(new RegExp(find, 'g'), replace);
 }
 
+let wasTouchDeviceDetected = false;
+function mobileSafeTipTouchStart() {
+  wasTouchDeviceDetected = true;
+}
```
Jamiras, line 127: this function appears unused.

wescopeland: Fixed

Jamiras (approved):

luchaos (approved):
Eventually all of the very old wz_tooltip implementation should be replaced with https://popper.js.org/ which would also resolve a lot of the mobile issues (events, positioning)

wescopeland:
@luchaos Floating UI is probably worth looking into as a more modern choice: https://opencollective.com/floating-ui/updates/popper-is-evolving-into-floating-ui

PR #1368, opened 2023-02-15
fix(forum): remediate an issue where long strings dont wrap

wescopeland:
This PR remediates an issue where a long string, such as a very long URL, does not break properly on forum posts and instead overflows the x-axis of the container.

This can be reproduced by going here and scrolling horizontally:
https://retroachievements.org/viewtopic.php?t=13871&c=154476#154476

The post that is overflowing is by a user named ExTermi.

After this PR, the same post wraps correctly:
Screenshot 2023-02-14 at 9 47 09 PM

### DIFF
--- a/public/viewtopic.php
+++ b/public/viewtopic.php
@@ -207,7 +207,7 @@ RenderContentStart($pageTitle);
             echo userAvatar($nextCommentAuthor, label: false, iconSize: 64);
             echo "</td>";
 
-            echo "<td class='w-full py-3 break-normal' id='$nextCommentID'>";
+            echo "<td class='w-full py-3 break-all' id='$nextCommentID'>";
 
             echo "<div class='flex justify-between mb-2'>";
             echo "<div>";
--- a/public/viewtopic.php
+++ b/public/viewtopic.php
@@ -207,7 +207,7 @@ RenderContentStart($pageTitle);
             echo userAvatar($nextCommentAuthor, label: false, iconSize: 64);
             echo "</td>";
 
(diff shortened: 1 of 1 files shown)


public/viewtopic.php
```diff
@@ -207,7 +207,7 @@
             echo userAvatar($nextCommentAuthor, label: false, iconSize: 64);
             echo "</td>";
 
-            echo "<td class='w-full py-3' id='$nextCommentID'>";
+            echo "<td class='w-full py-3 break-normal' id='$nextCommentID'>";
```
Jamiras, line 210: This doesn't seem to have any effect in Chrome or Edge. It does seem to work in Firefox.

`break-all` does appear to work in all three, but I don't now the implications of using it, or if there's a better middle ground.

wescopeland: `break-all` should be fine, although I am seeing `break-normal` work on Chrome and Edge. Is it possible on your local machine that Tailwind isn't picking up & adding the class name to the compiled CSS?

Jamiras: Anything is possible, but running `npm run build` doesn't have any noticeable effect when using `break-normal`, but was required when using `break-all`.

Jamiras: And to echo my previous statement, it did work in Firefox, so that seems to have picked it up.

wescopeland: Confirmed this was an issue with Tailwind & Vite on my local setup. Fixed.

Jamiras (approved):

PR #1370, opened 2023-02-15
feat(game): add mature content permanent bypass option

wescopeland:
This PR adds a 3rd option to the mature content warning. When clicked, this option disables the notification setting for mature content, then redirects the user to the game page.

This change was motivated by some discussion in the #site-feedback channel.

Screenshot 2023-02-15 at 4 43 49 PM

Also, "No" is now in the first position. If this content warning is continuously displayed we can now assume going forward this is no accident. Therefore it makes sense for "No" to be the primary action.

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -96,8 +96,9 @@ if ($v != 1 && $isFullyFeaturedGame) {
          * @param {number} currentPreferenceValue
          * @param {number} gameId
          */
-        function disableMatureContentWarningPreference(currentPreferenceValue, gameId) {
+        function disableMatureContentWarningPreference(currentPreferenceValue) {
             const newPreferencesValue = <?= $userDetails['websitePrefs'] | (1 << $matureContentPref) ?>;
+            const gameId = <?= $gameID ?>;
 
             fetch('/request/user/update-notification.php', {
                 method: 'POST',
@@ -139,10 +140,7 @@ if ($v != 1 && $isFullyFeaturedGame) {
                     <?php if ($userWebsitePrefs): ?>
                         <button 
                             class='break-words whitespace-normal leading-normal' 
-                            onclick='disableMatureContentWarningPreference(
-                                <?= $userWebsitePrefs ?>,
-                                <?= $gameID ?>
-                            )'
+                            onclick='disableMatureContentWarningPreference(<?= $userWebsitePrefs ?>)'
                         >
                             Yes. Never ask me again for games with mature content.
                         </button>
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -130,17 +130,24 @@ if ($v != 1 && $isFullyFeaturedGame) {
                 <div class="flex flex-col sm:flex-row gap-4 sm:gap-2">
                     <form id='escapeform' action='/gameList.php'>
                         <input type='hidden' name='c' value='<?= $consoleID ?>'/>
-                        <input type='submit' value='No, get me out of here.'/>
+                        <input type='submit' class='leading-normal' value='No, get me out of here.'/>
                     </form>
 
                     <form id='consentform' action='/game/<?= $gameID ?>'>
                         <input type='hidden' name='v' value='1'/>
-                        <input type='submit' value='Yes, I&apos;m an adult.'/>
+                        <input type='submit' class='leading-normal' value='Yes, I&apos;m an adult.'/>
                     </form>
 
                     <?php
                     if ($userWebsitePrefs) {
-                        echo "<button class='w-max' onclick='disableMatureContentWarningPreference({$userWebsitePrefs}, {$matureContentPref}, {$gameID})'>Yes, and never ask me again for games with mature content.</button>";
+                        echo "
+                            <button 
+                                class='break-words whitespace-normal leading-normal' 
+                                onclick='disableMatureContentWarningPreference({$userWebsitePrefs}, {$matureContentPref}, {$gameID})'
+                            >
+                                Yes, and never ask me again for games with mature content.
+                            </button>
+                        ";
                     }
                     ?>
                 </div>
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -94,10 +94,9 @@ if ($v != 1 && $isFullyFeaturedGame) {
     <script>
         /**
          * @param {number} currentPreferenceValue
-         * @param {number} matureContentEnumValue
          * @param {number} gameId
          */
-        function disableMatureContentWarningPreference(currentPreferenceValue, matureContentEnumValue, gameId) {
+        function disableMatureContentWarningPreference(currentPreferenceValue, gameId) {
             const newPreferencesValue = <?= $userDetails['websitePrefs'] | (1 << $matureContentPref) ?>;
 
             fetch('/request/user/update-notification.php', {
@@ -142,7 +141,6 @@ if ($v != 1 && $isFullyFeaturedGame) {
                             class='break-words whitespace-normal leading-normal' 
                             onclick='disableMatureContentWarningPreference(
                                 <?= $userWebsitePrefs ?>,
-                                <?= $matureContentPref ?>,
                                 <?= $gameID ?>
                             )'
                         >
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -27,7 +27,6 @@ $userID = $userDetails['ID'] ?? 0;
 $userWebsitePrefs = $userDetails['websitePrefs'] ?? null;
 $matureContentPref = UserPreference::SiteMsgOff_MatureContent;
 
-
 $officialFlag = AchievementType::OfficialCore;
 $unofficialFlag = AchievementType::Unofficial;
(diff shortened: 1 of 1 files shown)


EmoonX:
"Yes, and never ask me again" still seems a bit ambiguous to me in whether it refers to _all_ mature games or just the game in question.

Maybe "Yes, and never ask me again for games with mature content." would be clearer.

wescopeland:
> "Yes, and never ask me again" still seems a bit ambiguous to me in whether it refers to _all_ mature games or just the game in question.
> 
> Maybe "Yes, and never ask me again for games with mature content." would be clearer.

Sounds good. Updated to the suggested copy 👍

wescopeland:
After some adjustments to line height and word break issues on mobile:

Screenshot 2023-02-16 at 8 28 47 AM

Screenshot 2023-02-16 at 8 28 58 AM

public/gameInfo.php
```diff
@@ -101,14 +126,31 @@
...
+
+                <div class="flex flex-col sm:flex-row gap-4 sm:gap-2">
+                    <form id='escapeform' action='/gameList.php'>
+                        <input type='hidden' name='c' value='<?= $consoleID ?>'/>
+                        <input type='submit' class='leading-normal' value='No, get me out of here.'/>
+                    </form>
+
+                    <form id='consentform' action='/game/<?= $gameID ?>'>
+                        <input type='hidden' name='v' value='1'/>
+                        <input type='submit' class='leading-normal' value='Yes, I&apos;m an adult.'/>
```
Jamiras, line 138: Google's [grammar suggestions](https://www.margieholdscourt.com/after-yes-and-no/#:~:text=Let's%20keep%20it%20simple%3A,the%20question%2C%20use%20a%20comma.) say that you should use a period unless you're echoing the question. 
`Yes, I want to view it` vs `Yes. I'm an adult.`

I think that a period is appropriate for all three of these lines.

public/gameInfo.php
```diff
@@ -89,6 +91,29 @@
...
     <?php RenderContentStart($pageTitle) ?>
+    <script>
+        /**
+         * @param {number} currentPreferenceValue
+         * @param {number} matureContentEnumValue
+         * @param {number} gameId
+         */
+        function disableMatureContentWarningPreference(currentPreferenceValue, matureContentEnumValue, gameId) {
+            // eslint-disable-next-line no-bitwise -- this must be calculated via bitwise operations
+            const newPreferencesValue = currentPreferenceValue + (1 << matureContentEnumValue);
```
Jamiras, line 102: Could you do the bitwise operation at script time (instead of even passing it as parameters)?
```
const newPreferencesValue = <?= $userDetails['websitePrefs'] | (1 << $matureContentPref) ?>;
...
```

Could also inline the `gameId` value as well.

public/gameInfo.php
```diff
@@ -101,14 +126,31 @@
...
+                    </form>
+
+                    <form id='consentform' action='/game/<?= $gameID ?>'>
+                        <input type='hidden' name='v' value='1'/>
+                        <input type='submit' class='leading-normal' value='Yes, I&apos;m an adult.'/>
+                    </form>
+
+                    <?php
+                    if ($userWebsitePrefs) {
+                        echo "
```
Jamiras, line 143: consider `<?php if ($userWebsitePrefs): ?>` ... `<?php endif ?>`

public/gameInfo.php
wescopeland: Good idea, I will give this a try.

public/gameInfo.php
EmoonX: Sounds so strange and unnatural to me. Like a robot talking.

Jamiras: The comma connects statements. "Yes, I'm an adult." is a confirmation to a question that wasn't asked. "Yes. I'm an adult." is one statement answering the question that was asked, and another statement providing additional context to the response.

EmoonX: Grammar-wise, it makes sense. It's just that the period-induced pause makes the reply feel more mechanic/blunt and less "friendly", at least to me.

Some mature sites [do indeed](https://user-images.githubusercontent.com/22218549/219454967-9bc784a6-0137-444a-a2e2-cfd3f6d43d1b.png) use commas for dialogs like these. The majority, however, simply [drop](https://user-images.githubusercontent.com/22218549/219456371-aff7f922-f647-4dcf-8d70-c40f469523b9.png)
 the "yes/no" part altogether, or just don't use punctuation at all. I guess in the end it merely depends on what impression we want to leave in the user.

public/gameInfo.php
wescopeland: This worked and is much cleaner, thanks.

public/gameInfo.php
wescopeland: Fixed.

public/gameInfo.php
wescopeland: I've adjusted this to follow "[Yes|No]. [phrase]".

public/gameInfo.php
```diff
@@ -89,6 +91,27 @@
 ?>
 <?php if ($gate): ?>
     <?php RenderContentStart($pageTitle) ?>
+    <script>
+        /**
+         * @param {number} currentPreferenceValue
+         * @param {number} gameId
+         */
+        function disableMatureContentWarningPreference(currentPreferenceValue, gameId) {
```
Jamiras: currentPreferenceValue is no longer used, and gameId could be eliminated if you change line 111 to:
```
window.location = `/game/<?= $gameId ?>`;
```

public/gameInfo.php
```diff
@@ -101,14 +124,30 @@
...
+
+                    <?php if ($userWebsitePrefs): ?>
+                        <button 
+                            class='break-words whitespace-normal leading-normal' 
+                            onclick='disableMatureContentWarningPreference(
+                                <?= $userWebsitePrefs ?>,
+                                <?= $gameID ?>
+                            )'
+                        >
+                            Yes. Never ask me again for games with mature content.
```
Jamiras, line 145: Going back to my [previously referenced web page](https://www.margieholdscourt.com/after-yes-and-no/#:~:text=Let's%20keep%20it%20simple%3A,the%20question%2C%20use%20a%20comma.), I believe it's acceptable to say "Yes. And never..."

public/gameInfo.php
```diff
@@ -89,6 +91,27 @@
 ?>
 <?php if ($gate): ?>
     <?php RenderContentStart($pageTitle) ?>
+    <script>
+        /**
+         * @param {number} currentPreferenceValue
+         * @param {number} gameId
+         */
+        function disableMatureContentWarningPreference(currentPreferenceValue, gameId) {
+            const newPreferencesValue = <?= $userDetails['websitePrefs'] | (1 << $matureContentPref) ?>;
```
Jamiras, line 100: This either needs to check for the presence of `$userDetails`, or the entire function should be suppressed when a user is not logged in.

public/gameInfo.php
wescopeland: Done

public/gameInfo.php
wescopeland: Fixed

public/gameInfo.php
wescopeland: You're right - should be fixed now.

Jamiras (commented):
I think this is acceptable as-is, but I'm wondering if we shouldn't still require the user to state that they're an adult in the "never ask again" clause.

It's not like we actually require proof, but there's probably some legal loophole where completely disabling the warning without forcing the user to lie about their age could get us into trouble. Though, I suppose the same could be said for the check box on the profile page.

Jamiras (approved):

wescopeland:
> I think this is acceptable as-is, but I'm wondering if we shouldn't still require the user to state that they're an adult in the "never ask again" clause.
> 
> It's not like we actually require proof, but there's probably some legal loophole where completely disabling the warning without forcing the user to lie about their age could get us into trouble. Though, I suppose the same could be said for the check box on the profile page.

I will do some additional research on this behind the scenes as I am also unsure.

Also, thank you for your attention to detail in this PR review. It was quite educational for me.

public/gameInfo.php
```diff
@@ -89,6 +91,29 @@
...
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/x-www-form-urlencoded',
+                    'X-CSRF-TOKEN': document.querySelector('meta[name="csrf-token"]').getAttribute('content')
+                },
+                body: `preferences=${newPreferencesValue}`,
+                credentials: 'same-origin'
+            }).then(() => {
+                window.location = `/game/<?= $gameID ?>`;
+            })
```
luchaos, line 114: While I appreciate the modern approach, note that this escapes the prepared ajax http interceptor setup: https://github.com/RetroAchievements/RAWeb/blob/master/public/js/all.js#L9-L45

I have not yet found the time to further improve the frontend setup; it would include generalizing `fetch` usage in the application so we can replace jquery down the line, and move all.js to resources to be included in the build step.

luchaos (approved):

PR #1371, opened 2023-02-16
Add `100-Point Achievements` award section

EmoonX:
Add a new award section for displaying 100-point achievements unlocked by given player.

100-point-achievements
tooltip

**Why?**
Because such feats represent outstanding display of skill and are very rare in nature. Unlike masteries - than can easily go to the hundreds figures - 100-point achievements are far and few between. Milestones for those are also featured in _RANews_, giving yet another reason for them to be featured in the website.

For coherence and preventing abuse, only achievements unlocked on _hardcore_ are taken into account.

Since data for this can be easily fetched from the `Achievements` and `Awarded` tables, no interaction to `SiteAwards` is done. This, however, doesn't directly allow such awards to be reordered; which is why I opted for using a dummy display order (9999) and showing them below all others. I plan to tackle such issues in a future PR.

### DIFF
--- a/app_legacy/Community/Enums/AwardType.php
+++ b/app_legacy/Community/Enums/AwardType.php
@@ -18,6 +18,8 @@ abstract class AwardType
 
     public const CertifiedLegend = 7;
 
+    public const HundredPointAchievement = 8;
+
     public static function cases(): array
     {
         return [
@@ -26,6 +28,7 @@ abstract class AwardType
             self::AchievementPointsYield,
             self::PatreonSupporter,
             self::CertifiedLegend,
+            self::HundredPointAchievement,
         ];
     }
 
--- a/app_legacy/Helpers/database/site-award.php
+++ b/app_legacy/Helpers/database/site-award.php
@@ -125,9 +125,12 @@ function getUsersSiteAwards($user, $showHidden = false): array
         $retVal = array_values(array_filter($retVal));
     }
 
-    $query = "SELECT ach.ID, ach.GameID, ach.Title, ach.BadgeName, UNIX_TIMESTAMP(aw.Date) AS AwardedAt
+    // Get unlocked 100-point achievements
+    $query = "SELECT ach.ID, ach.Title, ach.Description, ach.Points, ach.BadgeName,
+            gd.Title AS GameTitle, UNIX_TIMESTAMP(aw.Date) AS AwardedAt
         FROM Awarded aw
         LEFT JOIN Achievements ach ON aw.AchievementID = ach.ID
+        LEFT JOIN GameData gd ON ach.GameID = gd.ID
         WHERE aw.User = '$user' AND aw.HardcoreMode = 1 AND ach.Points = 100
         GROUP BY ach.ID";
 
--- a/app_legacy/Helpers/database/site-award.php
+++ b/app_legacy/Helpers/database/site-award.php
@@ -125,6 +125,22 @@ function getUsersSiteAwards($user, $showHidden = false): array
         $retVal = array_values(array_filter($retVal));
     }
 
+    $query = "SELECT ach.ID, ach.GameID, ach.Title, ach.BadgeName,
+            MAX(aw.HardcoreMode) AS HardcoreMode, UNIX_TIMESTAMP(aw.Date) AS AwardedAt
+        FROM Awarded aw
+        LEFT JOIN Achievements ach ON aw.AchievementID = ach.ID
+        WHERE aw.User = '$user' AND ach.Points = 100
+        GROUP BY ach.ID";
+    
+    $dbResult = mysqli_query($db, $query);
+    if ($dbResult !== false) {
+        while ($db_entry = mysqli_fetch_assoc($dbResult)) {
+            $db_entry['AwardType'] = AwardType::HundredPointAchievement;
+            $db_entry['DisplayOrder'] = 9999;
+            array_push($retVal, $db_entry);
+        }
+    }
+
     return $retVal;
 }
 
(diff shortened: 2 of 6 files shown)


Soulsderfuchs:
I fail to see the point in displaying these achievements themselves if more likely than not they will also get the mastery itself , making this not only redundant but just spammy.

Would be better if it was an actual custom display section for any achievement the user would want to put on their profiles.

wescopeland:
> I fail to see the point in displaying these achievements themselves if more likely than not they will also get the mastery itself , making this not only redundant but just spammy.
> 
> Would be better if it was an actual custom display section for any achievement the user would want to put on their profiles.

Couldn't disagree more.

A user may have 100 masteries but only 1 100-point achievement. That makes the 100-point achievement quite special.

EmoonX:
While it is almost a tautology that _100-point cheevo implies mastery_ in a set, the 100-point/mastery ratio is so absurdly low on average (most good-ranked players are bound to not have even a single one) it barely makes a difference most of the time. And I'm pretty sure the majority of people would like to show or see extremely rare feats like those by default.

Again, it is well-known that the sidebar awards section is pretty crowded. But I think we should focus on fixing the direst parts first (like the neverending giant wall of mastery badges, which is in the plans) before sticking to minutiae.

Krylan:
I think that better indicator of "extremely rare feats" would be achievements with highest Retro Ratio (which takes into account assigned point value, but also how many players did really achieve it). I would see it as a left-side section of few (5-10) the most rare achievements a player obtained (on hardcore, of course).
It would be available for more players (if they obtained any achievements, then those would be the most rare for them), but the display would be limited to few achievements (like on steam profile sections), so we are sure it won't be crowded.
In case of another "neverending" section, we could have tens or even hundreds achievements (images) displayed there – and there are still added more. Players with the most masteries, who already have a really long list of badges, would probably have a lot of 100-points achievements, too, so it would make the problem worse.

EmoonX:
> achievements with highest Retro Ratio (which takes into account assigned point value, but also how many players did really achieve it)

You got the concepts mixed up. Ratio is the number derived from win%, Retro _Points_ is the one calculated using it and normal points (doesn't help that the achievement list page got it wrong too). Either way, RP is an essentially unreliable measure and even more in this case; bonus sets, in particular, _love_ to have 100-pointers with low RP (usually due to only pr0s attempting them).

> Players with the most masteries, who already have a really long list of badges, would probably have a lot of 100-points achievements

Define "a lot". From the outdated dump, the one with the most 100s is HippopotamusRex (26) and that doesn't account for not even 10% of their amount of masteries. Whenever we shall start considering "hundreds" of 100-pointers, people will have over 1000 mastery badges. And then what? There's just no sense in worrying about clutter from a minor part while an elephant is taking the most space.

Jamiras:
> I think that better indicator of "extremely rare feats" would be achievements with highest Retro Ratio (which takes into account assigned point value, but also how many players did really achieve it)

There is an item in the backlog (no issue/discussion yet) to showcase the rarest achievements that a player has earned (i.e. "only 0.5% of players did this"), but anything dealing with rarity is currently waiting for denormalized unlock metrics to simplify the queries. White points could serve as a denormalized value, but a "common" 100 pointer would invariably have more white points than the rarest 10 pointer.

Additionally, using unlock rates to derive rarity (either directly or through white points) is notably broken currently. Achievements in subsets have a disproportionally high earn rate because most people who would not earn the achievements won't play the subset. Late game achievements have a lower earn rate due to player attrition (not difficulty), and achievements added as part of a revision also have low earn rates simply because of the number of people who played the game before the new achievements were added.

wescopeland:
If it is not currently possible to accurately query by rarity, my perception is this change is a step in the right direction and is a good stopgap solution until such queries are possible and/or reliable.

EmoonX:
> There are currently 1958 hardcore players (about 5.5%) with at least one hardcore 100-pointer. 171 with more than 5, (...)

So for 99.5% of ranked players, the section would either not be shown at all or have only one row. As for allowing ordering ordering/hiding of such awards, it can/will be done in another PR.

Metric-wise, I see 100-pointers as perfectly fine. There's a reason PSN trophies are so popular, and it's basically about _simplicity_. People tend to react better to easily-recognizable, discrete feats than arbitrary thresholds as percentages. Sure, it's not perfect; it can be argued that we still have some 50-pointers harder than 100 ones, but those are exceptions and honestly more an issue with cheevo design than the rule.

As for what to show in the sidebar section... I recall @luchaos mentioning at some point something in the lines of the current view being just a _concession_ at the moment. Ideally, we should have a dedicated, separate page where such exhaustive info (full mastery and achievements lists) would be displayed and toyed with. Then, the sidebar content would restrict itself just to a fraction of the full content (like selectable achievements - AND masteries). However, until we get to that... I see no issue in using a minor part of the current model as a lab to new features.

Either way, I'm planning with @wescopeland to push some updates to the sidebar award section soon. I recommend waiting to see that; then, we can revisit the tidbits.

Krylan:
> You got the concepts mixed up. Ratio is the number derived from win%, Retro Points is the one calculated using it and normal points (doesn't help that the achievement list page got it wrong too).

@EmoonX yeah, I know the difference, that was just a mistake on my writting - of course, I meant Retro Points (was writting this early on the morning and looking at achievement list page also empowered that error).

I wanted to refer here to that elephant comparison, but I saw this part of your latest comment:
>Either way, I'm planning with @wescopeland to push some updates to the sidebar award section soon. I recommend waiting to see that; then, we can revisit the tidbits.

I decided to wait (that's why I didn't respond) and today I saw proposal on Discord's site-feedback channel. Now I can see how this wall issue will be addressed, so if these sections would work this way, then I'm a bit more calm about this.

PR #1334, opened 2023-01-28
Replace mysqli with Eloquent/PDO

luchaos:
- Drop mysqli related functions, use query bindings or models instead
- DB engine agnostic time diff statements (mysql/sqlite)
- Add tests for legacy database layer, home page, web api v1
- Move cron scripts to commands
- Remove IDE-only http tests
- Remove unused `lib/util/cookie.php`
- Refactor some render functions into blade views

PR #1372, opened 2023-02-16
Grid-like section for top links

EmoonX:
This is a follow-up to a [previous discussion](https://github.com/RetroAchievements/RAWeb/pull/1334#discussion_r1089792205).

In place of having everything in a list or square grid, I opted for a mix of both; should feel less monotone and more heterogeneous this way. Bottom links are ideal for being displayed in a single row, as labels are small there. 

**Old:**

before

**New:**

https://user-images.githubusercontent.com/22218549/219261010-29fdecaf-6cd3-46a0-a582-ff621b069dae.mp4

---
Gets rid of the margins between buttons in exchange for some more padding. Changes most icon colors too. Instead of every `btn` having round corners, those are applied in a way that styles the container itself.

### DIFF
--- a/resources/views/content/top-links.blade.php
+++ b/resources/views/content/top-links.blade.php
@@ -1,39 +1,39 @@
-<x-section class="mb-3 flex flex-col">
-    <a class="btn flex justify-center gap-1.5 p-3 text-xs" href="/globalRanking.php?s=5&t=2">
+<x-section class="top-links">
+    <a href="/globalRanking.php?s=5&t=2">
         <span class="text-yellow-500"><x-fas-medal/></span>
         Global Ranking
     </a>
-    <a class="btn flex justify-center gap-1.5 p-3 text-xs" href="{{ route('download.index') }}">
+    <a href="{{ route('download.index') }}">
         <span class="text-purple-400"><x-fas-gamepad/></span>
         Download Emulator
     </a>
     @if(config('services.discord.invite_id'))
-        <a class="btn flex justify-center gap-1.5 p-3 text-xs" href="https://discord.gg/{{ config('services.discord.invite_id') }}">
+        <a href="https://discord.gg/{{ config('services.discord.invite_id') }}">
             <span class="text-discord"><x-fab-discord/></span>
             Join us on Discord
         </a>
     @endif
     @if(config('services.patreon.user_id'))
-        <a class="btn flex justify-center gap-1.5 p-3 text-xs" href="https://www.patreon.com/bePatron?u={{ config('services.patreon.user_id') }}">
+        <a href="https://www.patreon.com/bePatron?u={{ config('services.patreon.user_id') }}">
             <span class="text-patreon"><x-fab-patreon/></span>
             Become a Patron
         </a>
     @endif
     <div class="grid grid-cols-4">
-        <a class="btn flex flex-col items-center gap-3 p-3 text-xs" href="https://news.retroachievements.org/">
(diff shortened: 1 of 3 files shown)


wescopeland:
Maybe it makes sense for me to round the corners in https://github.com/RetroAchievements/RAWeb/pull/1354 too

resources/views/content/top-links.blade.php
```diff
@@ -1,41 +1,39 @@
...
-        <span class="text-yellow-400"><x-fas-medal/></span>
+<x-section class="top-links">
+    <a href="/globalRanking.php?s=5&t=2">
+        <span class="text-yellow-500"><x-fas-medal/></span>
         Global Ranking
     </a>
-    <a class="btn text-center py-2" href="{{ route('download.index') }}">
-        <span class="text-heading"><x-fas-gamepad/></span>
+    <a href="{{ route('download.index') }}">
+        <span class="text-purple-400"><x-fas-gamepad/></span>
```
Jamiras, line 7: The blue color was more on-brand than this pinkish hue.

Jamiras (commented):
Seems functional. Want @luchaos's approval on the styling choices.

resources/views/content/top-links.blade.php
EmoonX: Blue was just the theme-based color being applied though. I think one reason I've changed it is that the Discord icon (not shown on ss) is also blue.

luchaos: The Discord and Patreon icons use their respective brand color. I'd prefer the other icons to stay in the theme color.
The only other icon to have a different color than the default theme color should be the ranking as that's a bit of an exception. Other resources like games do not have any color assigned.

luchaos:
I'm not sure if removing all the gaps is preferable. There should be a minimum gap between any interactive item imo.

> Maybe it makes sense for me to round the outer corners in #1354 too

@wescopeland Yes, I forgot about that in the stats PR - they should have rounded corners just like any other button.

resources/views/content/top-links.blade.php
```diff
@@ -1,41 +1,39 @@
...
-        <a class="btn text-center py-2 grow" href="https://news.retroachievements.org/">
-            <span class="text-heading"><x-fas-newspaper/></span>
+    <div class="grid grid-cols-4">
+        <a href="https://news.retroachievements.org/">
+            <span><x-fas-newspaper/></span>
             RANews
         </a>
-        <a class="btn text-center py-2 grow" href="https://www.youtube.com/channel/UCIGdJGxrzmNYMaAGPsk2sIA">
-            <span class="text-heading"><x-fas-microphone/></span>
+        <a href="https://www.youtube.com/channel/UCIGdJGxrzmNYMaAGPsk2sIA">
```
Jamiras, line 27: It has been suggested that this link should be updated to `https://www.youtube.com/@RAPodcast`: https://discord.com/channels/310192285306454017/490333108621672448/1080659700967813180

EmoonX:
I'll be converting this to draft as I figured out I'm not a big fan of my proposed design either.

resources/views/content/top-links.blade.php
ScottFromDerby: I would be really grateful if we could go ahead with this change please, whenever is convenient

wescopeland: PR to adjust the link: https://github.com/RetroAchievements/RAWeb/pull/1426

PR #1374, opened 2023-02-16
Small fixes/adjustments for Inbox/Outbox pages

### LINKED ISSUE #1373: Inbox/Outbox messages list bug
I have been noticing this bug months ago, but only tried to report just now when found a solution after looking Inbox.php code...

This kind of bug only happens if you have all your messages listed whereas Total Messages is divisible by 10 (Example: 20, 50, 100, 150, etc.). If you have 24 messages, then, next page will list the last 4 ones correctly.

**Try for yourself:**
1. Send a lot of messages to yourself at the way that you fill any page with 10 messages
2. Click "Next 10 >" link
3. You will see a blank list there.

**Screenshots:**

Last page with last message shown (Page 2):

Next page leading to nowhere... (Page 3):

**Computer Information:**
 - OS: Windows 10 Pro 20H2
 - Browser: Microsoft Edge

**Solution:**
After looking on [Inbox.php](https://github.com/RetroAchievements/RAWeb/blob/master/public/inbox.php) file here on RAWeb repository, it seems that Line 186 is the culprit as it makes to show that link "Next 10 >" everytime when exactly 10 messages are loaded. This also applies to Outbox case:

`if ($totalMsgs == $maxCount)`

Since $maxCount is always 10 as expressed on Line 5, $totalMsgs will only be 10 if all 10 messages are loaded. However, it doesn't block the case if all messages ($totalMessageCount) have already been loaded. Looking on next lines, you can see that $offset variable is being incremented by 10 as new offset to next page on Line 187 inside Echo command. So, why not checking if this sum is less than $totalMessageCount before replicating that "Next 10 >" link?

What I mean is something like this:

`if (($totalMsgs == $maxCount) && (($offset + $maxCount) < $totalMessageCount))`

For example above, I am on second page, so $offset value is 10 since it will list from IDs 10 to 19 on this page. On next page is expected to list from IDs 20 to 29 with $offset being 20. However ($offset + $maxCount) is 20 on second page which is equal to $totalMessageCount. So, clicking "Next 10 >" will lead to nowhere...

On another hand, if you have 52 messages to display in total ($totalMessageCount == 52), then, you will have 6 pages to display whereas Page 5 will display 10 messages with $offset == 40 and ($offset + $maxCount == 50). Next page will exhibit the last 2 messages: one message with ID 50 and other with ID 51. Since ($offset + $maxCount) will be 60, only "< Previous 10" will be displayed.

Hope that it helps!

**Disclaimer:**
This is my first report here on Github. Sorry if I posted this on wrong place.

EmoonX:
Fix #1373, plus a few more changes.

- Don't show `Next ... >` if there are no more messages ahead (corner case handling)
- Allow passing custom `c` parameter for messages-per-page amount instead of previously hardcoded 10 (no UI way of utilizing this yet; must be URL-modified)
- Fix server error due to occasional negative offset after clicking `< Previous` (defaults to 0; not an expected error, only happens when toying with URL)
- Shows true number of messages next if below max amount (e.g `Next 5 >` instead of `Next 10 >` when at offset 30 with 45 total, 10 per page)

### DIFF
--- a/public/inbox.php
+++ b/public/inbox.php
@@ -173,15 +173,10 @@ function MarkAsUnread(msgID) {
 
             echo "</tbody></table>";
 
-            // Get message count and build URL from current GET parameters
+            // Get message count and build paginator URL from current GET parameters
             $messageCount = $unreadOnly ? $unreadMessageCount : $totalMessageCount;
-            $urlPrefix = '/inbox.php?';
-            foreach ($_GET as $parameter => $value) {
-                if ($parameter !== 'o') {
-                    $urlPrefix .= "$parameter=$value&";
-                }
-            }
-            $urlPrefix .= 'o=';
+            unset($_GET['o']);
+            $urlPrefix = '/inbox.php?' . http_build_query($_GET) . '&o=';
 
             echo "<div class='float-right'>";
             RenderPaginator($messageCount, $displayCount, $offset, $urlPrefix);
--- a/public/inbox.php
+++ b/public/inbox.php
@@ -2,10 +2,10 @@
 
 use App\Support\Shortcode\Shortcode;
 
+$outbox = requestInputSanitized('s', 0, 'integer');
+$unreadOnly = requestInputSanitized('u', 0, 'integer');
 $count = requestInputSanitized('c', 10, 'integer');
 $offset = requestInputSanitized('o', 0, 'integer');
-$unreadOnly = requestInputSanitized('u', 0, 'integer');
-$outbox = requestInputSanitized('s', 0, 'integer');
 
 if (!authenticateFromCookie($user, $permissions, $userDetails)) {
     abort(401);
@@ -76,7 +76,7 @@ function MarkAsUnread(msgID) {
                 echo "<a class='btn btn-link' href='/inbox.php?s=1'>Outbox</a>";
                 echo "<div class='flex gap-2'>";
                 if ($unreadOnly) {
-                    echo "<a class='btn btn-link' href='/inbox.php?u=0'>View All Messages</a>";
+                    echo "<a class='btn btn-link' href='/inbox.php'>View All Messages</a>";
                 } else {
                     echo "<a class='btn btn-link' href='/inbox.php?u=1'>View Unread Only</a>";
                 }
@@ -177,7 +177,7 @@ function MarkAsUnread(msgID) {
 
             if ($offset > 0) {
                 $newOffset = max($offset - $count, 0);
-                echo "<a class='btn btn-link' href='/inbox.php?c=$count&o=$newOffset&u=$unreadOnly&s=$outbox'>";
+                echo "<a class='btn btn-link' href='/inbox.php?s=$outbox&u=$unreadOnly&c=$count&o=$newOffset'>";
                 echo "&lt; Previous $count";
                 echo "</a>";
             }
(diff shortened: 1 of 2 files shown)


Alexdatadestroyer:
> - Shows true number of messages next if below max amount (e.g Next 5 > instead of Next 10 > when at offset 30 with 45 total, 10 per page)

I think I already had a solution for this!

I will use what I proposed on #1373 

```
if (($totalMsgs == $maxCount) && (($offset + $maxCount) < $totalMessageCount))
{
   $customOffset = $totalMessageCount - ($offset + $maxCount);
   if (($customOffset % 10) != 0)
   {
      echo "<a class='btn btn-link' href='/inbox.php?o=" . ($offset + $customOffset) . "&amp;u=$unreadOnly&amp;s=$outbox'>";
      echo "Next $customOffset &gt;";
      echo "</a>";
   }
}
```
$customOffset will create a custom Offset at last page for cases when messages left is less than 10!

Alexdatadestroyer:
Good to know that my Inbox/Outbox solution was useful enough to started a fix revolution regarding this feature! That was a simple insight for a small exception that I could notice. I couldn't notice others though.

I was attempting to fix your last item: `Next 5 >` instead of `Next 10 >` when at offset 30 with 45 total, 10 per page, but I realized later that this was already fixed and removed the code posted here.

public/inbox.php
```diff
@@ -178,14 +176,18 @@ function MarkAsUnread(msgID) {
...
-            if ($totalMsgs == $maxCount) {
-                echo "<a class='btn btn-link' href='/inbox.php?o=" . ($offset + $maxCount) . "&amp;u=$unreadOnly&amp;s=$outbox'>";
-                echo "Next $maxCount &gt;";
+            $messagesLeft = ($unreadOnly ? $unreadMessageCount : $totalMessageCount) - $offset;
+            if ($messagesLeft > $count) {
+                $newOffset = $offset + $count;
+                $messagesNext = min($count, $messagesLeft - $count);
+                echo "<a class='btn btn-link' href='/inbox.php?s=$outbox&u=$unreadOnly&c=$count&o=$newOffset'>";
+                echo "Next $messagesNext &gt;";
                 echo "</a>";
```
Jamiras, line 191: Consider swapping out the prev/next links for a call to `RenderPaginator`. It would improve visual consistency with other parts of the product and assist with some of the edge cases (like attempted negative navigation).

EmoonX: Nice. It didn't cross my mind `inbox.php` was still using the old nav solution.

Jamiras (approved):

public/inbox.php
```diff
@@ -175,24 +173,15 @@ function MarkAsUnread(msgID) {
...
-                echo "Next $maxCount &gt;";
-                echo "</a>";
-            }
+            // Get message count and build paginator URL from current GET parameters
+            $messageCount = $unreadOnly ? $unreadMessageCount : $totalMessageCount;
+            unset($_GET['o']);
+            $urlPrefix = '/inbox.php?' . http_build_query($_GET) . '&o=';
 
+            echo "<div class='float-right'>";
+            RenderPaginator($messageCount, $displayCount, $offset, $urlPrefix);
```
Jamiras, line 182: Noticed a small issue with the paginator when there are no items (unread only, no unread messages). Could you add this to layout.php?
```
     $pages = floor(($numItems + $perPage - 1) / $perPage);
+    if ($pages < 1) {
+        $pages = 1;
+    }
     for ($i = 1; $i <= $pages; $i++) {
```

EmoonX: Okay, done. I have mixed feelings about whether we should show the paginator at all in the single page case, but... it's outside the scope of the PR.

PR #1375, opened 2023-02-16
fix(home): remediate issue where active players table overflows

wescopeland:
This PR fixes an issue reported in #site-feedback where long strings don't properly break in the table, resulting in a horizontal scrollbar and weird formatting for rich presence if a single one of these strings overflows.

chrome_QvCMy9A7TZ

### DIFF
--- a/app_legacy/Helpers/render/content.php
+++ b/app_legacy/Helpers/render/content.php
@@ -26,7 +26,7 @@ function RenderActivePlayersComponent(): void
                         <tr>
                             <td data-bind='html: playerHtml'></td>
                             <td data-bind='html: gameHtml'></td>
-                            <td data-bind='text: richPresence' class="w-full"></td>
+                            <td data-bind='text: richPresence' class="w-full break-words"></td>
                         </tr>
                         <!-- /ko -->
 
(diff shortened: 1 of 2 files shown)


app_legacy/Helpers/render/content.php
```diff
@@ -26,7 +26,7 @@ function RenderActivePlayersComponent(): void
                         <tr>
                             <td data-bind='html: playerHtml'></td>
                             <td data-bind='html: gameHtml'></td>
-                            <td data-bind='text: richPresence' class="w-full"></td>
+                            <td data-bind='text: richPresence' class="w-full break-words"></td>
```
Jamiras, line 29: Similar to https://github.com/RetroAchievements/RAWeb/pull/1368#discussion_r1108776953, I had to use `break-all` to make this work in Chrome (did not test Edge or Firefox)

Jamiras: In this case, it might have been my choice of test data.
image

wescopeland: Assuming it's working here for you on Chrome, it's interesting that it's an issue on another page. Can you confirm `break-words` is functioning correctly for you here?

Jamiras: It is not.
> I had to use `break-all` to make this work in Chrome

wescopeland: Thank you for the confirmation, I got confused by the follow-up comment. I will make the same adjustment to this PR as I make in the other one later today.

wescopeland: I can now repro what you're seeing. I think this was a problem with Tailwind & Vite on my end. Fixed.

Jamiras (approved):

PR #1376, opened 2023-02-17
feat(game): add scrolling for long lists of awards

wescopeland:
This is a collaborative PR between myself, @EmoonX, and @drisc.

This PR changes the behavior of long awards lists on user profiles. Previously, awards would infinitely populate the page in a single unrestricted div. Now, the div is given a max-height based on the user's screen size. An expand button is now available, and visual cues are also shown indicating the container can be scrolled.

Screenshot 2023-03-02 at 8 37 42 AM

**This screenshot is purely for illustrative purposes. On a 1440p display, roughly 85 awards are shown before the scroll kicks in.**

Additionally, based on Discord feedback, a new temporary setting has been created.

Screenshot 2023-03-02 at 8 39 13 AM

If enabled, a cookie is set which basically disables all this new logic.

### Things to note

* Per discussions with @luchaos, `vitest` has been added for coverage under _resources/js_. Additionally, I have pulled in some `@testing-library` dependencies and `jsdom`.
* The new front-end code is all written in TypeScript and has as much test coverage as can happen in JSDOM. Unfortunately, JSDOM is not able to simulate scrolling, so our coverage ends there.
* There is a thin scrollbar matching the user's theme to provide another visual cue that the list is scrollable. It was discussed also adding this to the "Completion Progress" section underneath, but decided not to for the time being to not further increase the scope of this PR.
* The new setting is entirely cookie-based and is not persisted to the DB. This decision follows some Discord discussion [here](https://discord.com/channels/310192285306454017/1048299730670735421/1076853639315992636).


public/js/all.js
```diff
@@ -376,3 +376,21 @@ function initializeTextareaCounter() {
 }
 
 window.addEventListener('load', initializeTextareaCounter);
+
+let shouldShowTopFade = false;
+let shouldShowBottomFade = true;
```
wescopeland, line 381: @EmoonX Right now this is global I think. Not sure if that matters but something to consider.

public/js/all.js
```diff
@@ -376,3 +376,21 @@ function initializeTextareaCounter() {
...
+function foo(event) {
+  const minimumScrollPosition = event.target.offsetHeight;
+  const scrollPosition = event.target.scrollTop + event.target.offsetHeight;
+
+  shouldShowBottomFade = scrollPosition < event.target.scrollHeight;
+  shouldShowTopFade = scrollPosition > minimumScrollPosition + 0;
+
+  event.target.classList.toggle('fade-hidden-top', !shouldShowTopFade);
+  event.target.classList.toggle('fade-hidden-bottom', !shouldShowBottomFade);
+}
```
wescopeland, line 396: For perf we should cache `event.target` and reduce down to a single `event.target.classList.toggle()` call

package.json
```diff
@@ -27,6 +27,7 @@
         "lodash": "^4.17.19",
         "postcss": "^8.4.14",
         "pusher-js": "^8.0.0",
+        "tailwind-scrollbar": "^2.1.0",
```
wescopeland, line 30: This package is not mandatory, but it does feel well-supported. The alternative is doing this with vanilla CSS, which seems like a reimplementation of what this package does anyway.

https://openbase.com/js/tailwind-scrollbar

app_legacy/Helpers/render/site-award.php
```diff
@@ -132,15 +136,20 @@ function RenderAwardGroup($awards, $title): void
         $counters = RenderCounter($icon, $text, $numItems, $numHidden);
     }
 
+    $fadeHiddenBottom = count($awards) <= 50 ? 'fade-hidden-bottom' : '';
     echo "<div id='" . strtolower(str_replace(' ', '', $title)) . "'>";
     echo "<h3 class='flex justify-between gap-2'><span class='grow'>$title</span>$counters</h3>";
-    echo "<div class='component flex flex-wrap justify-start gap-2'>";
+    echo "<div class='relative'>";
+    echo "<div onscroll='foo(event)' class='awards-fade fade-hidden-top $fadeHiddenBottom component flex flex-wrap justify-start gap-2 max-h-[340px] lg:max-h-[640px] lg:scrollbar-none overflow-y-auto transition'>";
```
wescopeland, line 143: We should double check if the `transition` classname is actually needed here

public/js/all.js
```diff
@@ -376,3 +376,21 @@ function initializeTextareaCounter() {
...
 window.addEventListener('load', initializeTextareaCounter);
+
+let shouldShowTopFade = false;
+let shouldShowBottomFade = true;
+
+// FIXME: Rename me.
+/**
+ * @param {Event} event
+ */
+function foo(event) {
```
wescopeland, line 387: It may be worth debouncing this so it doesn't run on every scroll frame

app_legacy/Helpers/render/avatar.php
```diff
@@ -20,7 +20,7 @@ function avatar(
     }
 
     if ($iconUrl) {
-        $iconLabel = "<img loading='lazy' width='$iconSize' height='$iconSize' src='$iconUrl' alt='$escapedName' class='$iconClass'>";
+        $iconLabel = "<img loading='lazy' decoding='async' fetchpriority='low' width='$iconSize' height='$iconSize' src='$iconUrl' alt='$escapedName' class='$iconClass'>";
```
EmoonX, line 23: Gotta make sure this snippet of code doesn't have unintended side-effects, given that it's used to generate most of the game/cheevo images on the site.

app_legacy/Helpers/render/site-award.php
EmoonX: Also a good idea to move those dozens tailwind classes to `.awards-fade` in the CSS file.

wescopeland:
The fade effect should not block user scrolling. There is a Tailwind class for this, just need to find it

app_legacy/Helpers/render/site-award.php
```diff
@@ -132,15 +136,22 @@ function RenderAwardGroup($awards, $title): void
         $counters = RenderCounter($icon, $text, $numItems, $numHidden);
     }
 
+    $fadeEnabled = count($awards) > 50 ? 'fade-enabled' : '';
     echo "<div id='" . strtolower(str_replace(' ', '', $title)) . "'>";
     echo "<h3 class='flex justify-between gap-2'><span class='grow'>$title</span>$counters</h3>";
-    echo "<div class='component flex flex-wrap justify-start gap-2'>";
+    echo "<div class='relative'>";
+    echo "<div onscroll='foo(event)' class='component flex flex-wrap justify-start gap-2 max-h-[340px] lg:max-h-[640px] lg:scrollbar-none overflow-y-auto transition'>";
```
wescopeland, line 143: The fade effect should not block user scrolling. There is a Tailwind class for this, just need to find it

public/js/all.js
```diff
@@ -376,3 +376,19 @@ function initializeTextareaCounter() {
...
+  const minimumScrollPosition = event.target.offsetHeight;
+  const scrollPosition = event.target.scrollTop + event.target.offsetHeight;
+
+  const topFadeOpacity = Math.min((scrollPosition - minimumScrollPosition) / 180, 1.0);
+  const bottomFadeOpacity = Math.min((event.target.scrollHeight - scrollPosition) / 180, 1.0);
+
+  event.target.querySelector('.awards-fade-top').style.opacity = topFadeOpacity;
+  event.target.querySelector('.awards-fade-bottom').style.opacity = bottomFadeOpacity;
```
wescopeland, line 392: We need to cache these rather than querying them on every frame. This could be problematic if there are 2 awards sections that both need to be scrolled. We probably need a different query, maybe ID > child class

public/js/all.js
```diff
@@ -376,3 +376,33 @@ function initializeTextareaCounter() {
...
+  const userCurrentScrollPosition = event.target.scrollTop + event.target.offsetHeight;
+
+  const newTopFadeOpacity = Math.min((userCurrentScrollPosition - minimumContainerScrollPosition) / 180, 1.0);
+  const newBottomFadeOpacity = Math.min((event.target.scrollHeight - userCurrentScrollPosition) / 180, 1.0);
+
+  // Cache the returned elements so we're not querying constantly.
+  if (!cachedAwardsScrollElements[containerKind]) {
+    cachedAwardsScrollElements[containerKind] = {
+      top: event.target.querySelector('#gameawards .awards-fade-top'),
+      bottom: event.target.querySelector('#gameawards .awards-fade-bottom')
```
wescopeland, line 400: Don't hardcode ID here

app_legacy/Helpers/render/avatar.php
wescopeland: I have reverted this change

EmoonX: Would it be useful to have an extra parameter so those attributes are applied just in this particular case and not generally?

wescopeland:
TODOs:

@EmoonX 

- This should probably be made relative to the element somehow, sibling selector?
```js
const seeAllButton = document.querySelector('.awards-see-all'); // FIXME
```

- The fade doesn't disappear from the bottom after clicking the See All button.

wescopeland:
@drisc will be investigating:
> See if still having the scrollbar be visible but very thin is possible with pure CSS

wescopeland:
I am a bit uneasy with how much JavaScript is being added to the _all.js_ file.
It might be a better approach to move this into _resources/js_, and possibly convert it to TypeScript pending https://github.com/RetroAchievements/RAWeb/pull/1390.

wescopeland:
@EmoonX @drisc I have migrated this to TypeScript and added some tests. Please give things a final look and if all is good we can move out of draft.

package.json
```diff
@@ -21,6 +23,9 @@
         "@tailwindcss/forms": "^0.5.2",
         "@tailwindcss/line-clamp": "^0.4.0",
         "@tailwindcss/typography": "^0.5.0",
+        "@testing-library/dom": "^9.0.0",
+        "@testing-library/jest-dom": "^5.16.5",
+        "@testing-library/user-event": "^14.4.3",
```
wescopeland, line 28: These are all testing deps for poking and prodding at DOM elements. `jsdom` lets us thinly simulate a document headlessly in vitest.

wescopeland:
@luchaos @Jamiras This is now ready for review

app_legacy/Helpers/render/site-award.php
luchaos: Personally, I still prefer dozens of tailwind classes over the @apply strategy as it requires building the frontend.
As discussed with @wescopeland there are SEO reasons for grouping css in classes, yet, I worked really hard to get rid of superfluous custom css classes in the past, I might be overly cautious with the recent trend of re-introducing classes.

resources/js/utils/throttle.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Creates a throttled version of a function that can be called at most
+ * once per `waitMs` milliseconds.
```
luchaos, line 3: Could this be lodash's debounce instead of a custom implementation?

public/controlpanel.php
```diff
@@ -148,6 +159,17 @@ function confirmEmailChange(event) {
...
+                echo "<td>Always expand User Awards on profiles</td>";
+                echo "<td>";
+                echo "<div class='flex flex-col'>";
+                echo "<div>";
+                echo "<input onclick='handleExpandedAwardsCheckboxClick(event)' class='mr-2' type='checkbox' " . ($doesPreferAlwaysExpandedRewards ? 'checked' : '') . ">";
+                echo "<button class='btn mr-4' onclick='expandableAwards.saveExpandableAwardsPreference(\"$expandedAwardsCookieName\", isExpandedAwardsCheckboxChecked)'>Save</button>";
+                echo "</div>";
+                echo "<span class='text-xs'>This only applies to your current device.</span>";
+                echo "</div>";
+                echo "</td>";
```
luchaos, line 172: I'd prefer there to be an implicit way of remembering this setting.
Instead of adding a setting to the controlpanel, there could be a 'Collapse' button that resets the preference. No need for a use setting here imo.

Regardless of that, if visual feedback is given it should re-use existing features, I removed a lot of alerts in the past, confirm dialogs will be replaced with dialogs in the future.
Screenshot 2023-03-02 at 19 58 13

luchaos (changes requested):

app_legacy/Helpers/render/site-award.php
```diff
@@ -132,16 +136,42 @@ function RenderAwardGroup($awards, $title): void
...
+    /**
+     * It is possible for the user to specify they want awards to
+     * always be expanded. This is set via a temporary cookie value.
+     * If it's set, we must never conceal an awards group.
+     * @see public/controlpanel.php
+     */
+    $expandedAwardsCookieName = 'prefers_always_expanded_rewards';
+    $doesPreferAlwaysExpandedRewards =
+        isset($_COOKIE[$expandedAwardsCookieName])
+        && $_COOKIE[$expandedAwardsCookieName] === 'true';
```
luchaos, line 150: Instead of using the `$_COOKIE` superglobal (which would result in warnings in some code quality tools) Laravel's `cookie()` helper can be used.

To not encrypt this cookie, which is the default, it has to be excluded explicitly here: https://github.com/RetroAchievements/RAWeb/blob/master/app/Site/Middleware/EncryptCookies.php#L14-L18

tailwind.config.js
```diff
@@ -46,6 +46,14 @@ module.exports = {
       lg: '1024px',
       xl: '1280px',
       '2xl': '1536px',
+    },
+    extend: {
+      colors: {
+        raTheme: {
+          scrollthumb: 'var(--link-color)',
```
luchaos, line 53: Something more muted would be preferable, as custom styled scrollbars might be used more in the future? Link colors are usually very bright; maybe just the text color?

tailwind.config.js
```diff
@@ -46,6 +46,14 @@ module.exports = {
...
+          scrolltrack: 'var(--text-color-muted)'
```
luchaos, line 54: Scrolltrack seems to be overly exposed on light scheme

resources/js/types/global.d.ts
```diff
@@ -1,8 +1,11 @@
 import type { Alpine } from 'alpinejs';
 
+import type { expandableAwards as ExpandableAwards } from '../utils/expandableAwards';
```
luchaos, line 3: It would be great if this could be generalized in some way, not including `awards` in the name if feasible? (Those will be just "badges" in the future) We might want to use this in other places too?

wescopeland: `badgeGroup`, perhaps?

resources/js/utils/throttle.ts
wescopeland: `lodash.throttle` might achieve the same thing - I will check. It's a subtle difference from `debounce` because we are not trying to delay the call, only rate limit it.

public/controlpanel.php
wescopeland: Your proposed approach definitely seems better. I was unclear how to get the "Changed" message to appear, so I naively went with `alert()` 😈

I will move this persistence entirely to the expand button itself and add make it more like a toggle-button, allowing the user to collapse the list after expansion.

app_legacy/Helpers/render/site-award.php
wescopeland: I am able to set the cookie with `cookie()`, but I am lost on how to read its current value using `cookie()`.

```php
$doesPreferAlwaysExpandedRewards = cookie($expandedAwardsCookieName) === 'true';
```

The above always seems to fail, even if the cookie is set as "true".

wescopeland: Screenshot 2023-03-02 at 6 39 20 PM

When I do `echo cookie(...)`, the value comes out as being deleted?

app_legacy/Helpers/render/site-award.php
```diff
@@ -24,9 +24,13 @@ function SeparateAwards($userAwards): array
...
     $eventAwards = array_values(array_filter($eventAwards, fn ($award) => !in_array($award, $devEventAwards)));
 
-    $siteAwards = array_values(array_filter($userAwards, fn ($award) => ($award['AwardType'] != AwardType::Mastery && AwardType::isActive((int) $award['AwardType'])) ||
-        in_array($award, $devEventAwards)
-    ));
+    $siteAwards = array_values(
+        array_filter(
+            $userAwards,
+            fn ($award) => ($award['AwardType'] != AwardType::Mastery && AwardType::isActive((int) $award['AwardType'])) ||
+            in_array($award, $devEventAwards)
```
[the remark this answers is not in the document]
wescopeland: Fixed

resources/js/utils/throttle.ts
wescopeland: I have kept the wrapper in case we want to swap this dep at some point in the future (happy to remove if you'd prefer). `lodash.throttle` works fine here.

public/controlpanel.php
wescopeland: I've reverted all changes on _controlpanel.php_. Per your suggestion, this is now totally persisted by the button itself. When clicking the button to Expand, rather than it being removed from the DOM, it becomes a "Collapse" button. Clicking it again adjusts the persisted setting to the user's preference.

I've updated the tests to ensure this behavior is working correctly.

tailwind.config.js
wescopeland: I agree that `--text-color` is a better fit here. Fixed.

tailwind.config.js
wescopeland: Agreed, after changing `scrollthumb` to use `--text-color`, this seems a bit better, however I am very finicky about my monitor's color calibration and it is entirely possible we are not seeing the same thing. Attached is a screenshot. Do you still find this difficult to see?

Screenshot 2023-03-02 at 6 56 32 PM

Generally, _everything_ using this combination of scheme and color settings appears a bit washed to my eyes.

resources/js/types/global.d.ts
wescopeland: I've changed `expandableAwards` to `badgeGroup`.

wescopeland:
@luchaos I've attempted to address all feedback left so far. The only thing lingering that I'm struggling with is the `cookie()` suggestion. Writing seems fine, reading always seems to fail.

wescopeland:
I am closing this for the time being. It was an excellent teamwork effort. I am certain we will revisit what the best UX solution for this is at some point; perhaps a more appropriate timing would be when the awards component is refactored to Blade.

PR #1389, opened 2023-02-25
Update systems seeder

luchaos:
Keep seeder in sync with the releases file.
Prevents errors on the downloads page after a fresh install without db import.


config/systems.php
```diff
@@ -508,6 +508,41 @@
...
+        'manufacturer' => '',
+        'active' => false,
+    ],
+    77 => [
+        'name' => 'Atari Jaguar CD',
+        'name_full' => 'Atari Jaguar CD',
+        'name_short' => 'Atari Jaguar CD',
+        'manufacturer' => 'Atari',
+        'active' => false,
+    ],
```
Jamiras, line 545: Should probably toss 78 (Nintendo DSi) in here too.

'active' could be audited too. PS2, CPC, MegaDuck, Channel F, maybe others.

luchaos: Thanks! Addressed feedback; not sure about the other systems but I think it's good enough for now? It's mainly about not getting errors when systems are referenced in the releases file that are not yet seeded.

Jamiras (approved):

PR #1390, opened 2023-02-26
refactor: migrate all new vite js code to typescript

wescopeland:
Per our conversation yesterday @luchaos, this PR migrates all new code living under _resources/js_ to TypeScript.

Additionally, it knocks out a TODO in _cookie.js_ for using a proper cookies library. I've chosen [js-cookie](https://bundlephobia.com/package/js-cookie@3.0.1), which is strongly-typed, tiny, and very widely used.

I have also refactored the code for setting and persisting themes to improve readability.

ESLint has been configured to run on both legacy JS code and new TS code.

_vite.config.js_ has also been migrated to _vite.config.ts_.


package.json
```diff
@@ -4,30 +4,37 @@
...
-        "fix:eslint": "eslint public/js resources/js postcss.config.js tailwind.config.js vite.config.js --fix",
+        "fix:eslint": "eslint public/js resources/js postcss.config.js tailwind.config.js vite.config.ts --fix",
         "analyze": "cross-env ANALYZE=true npm run prod",
         "apidoc": "npm run apidoc:connect",
         "apidoc:connect": "apidoc -i src/Connect/ -c docs/src/connect-api -t docs/src/connect-api/template -o docs/dist/api/connect"
     },
+    "dependencies": {
+        "alpinejs": "^3.0.6",
+        "js-cookie": "^3.0.1"
+    },
```
wescopeland, line 17: Anything shipping directly to the browser I've moved into `"dependencies"`. This notably excludes tailwind because Tailwind is just a PostCSS plugin.

I am happy to undo this but this is within alignment of conventions with _package.json_ files.

package.json
```diff
@@ -4,30 +4,37 @@
...
+        "@types/js-cookie": "^3.0.3",
+        "@typescript-eslint/eslint-plugin": "^5.53.0",
+        "@typescript-eslint/parser": "^5.53.0",
         "autoprefixer": "^10.4.7",
         "eslint": "^8.4.1",
         "eslint-config-airbnb-base": "^15.0.0",
         "eslint-plugin-import": "^2.20.2",
         "laravel-echo": "^1.4.0",
         "laravel-vite-plugin": "^0.7.0",
-        "lodash": "^4.17.19",
```
wescopeland, line 27: This appeared to be unused so I took the opportunity to delete it. I am not sure when we would ever need `lodash`, but in the event we do, using a pure module for the single needed function is generally a better choice, eg: [lodash.merge](https://www.npmjs.com/package/lodash.merge). Lodash is not tree-shakeable, so a single import from lodash massively inflates the bundle size.

resources/js/utils/helpers.ts
```diff
@@ -0,0 +1,11 @@
+export const config = (key: string) => window?.cfg?.[key];
+
+export const asset = (uri: string) => `${config('assetsUrl')}${uri}`;
+
+export const clipboard = (text: string) => {
+  navigator.clipboard.writeText(text).then(() => {
+    if (window.showStatusSuccess) {
+      window.showStatusSuccess('Copied!');
+    }
+  });
```
wescopeland, line 10: This previously was using the deprecated `document.execCommand` technique for feeding content into the clipboard. I have migrated this to use the [navigator.clipboard API](https://caniuse.com/mdn-api_navigator_clipboard).

package.json
luchaos: It's a default dependency in the Laravel skeleton. It's fine to have it removed imo

vite.config.ts
```diff
@@ -27,7 +27,7 @@ export default defineConfig(({ mode }) => {
       laravel({
         input: [
           'resources/css/app.css',
-          'resources/js/app.js',
+          'resources/js/index.ts',
```
luchaos, line 30: I'd go with `app.ts` as that's the common entry file in Laravel frontends (see `app.css`) and to not confuse it with "simple" index files. Doesn't make that much of a difference to me though.

luchaos (approved):

vite.config.ts
wescopeland: No problem, was wondering about this one. Should be fixed now.

luchaos:
Wonderful - thanks for taking care of this!

PR #1393, opened 2023-02-28
Strict types, refactor database helpers, rector & tests

luchaos:
Declare strict types.

Increase PHPStan level & fix errors.

Rector is configured to help with identifying type issues and some code style matters.


app_legacy/Helpers/database/search.php
```diff
@@ -130,7 +138,7 @@ function performSearch(int|array $searchType, string $searchQuery, int $offset,
         }
     }
 
-    if (count($articleTypes) > 0) {
```
luchaos, line 133: Using `count()` is expensive as it will actually count the items in the array. There are often better ways to check for "emptiness".

app_legacy/Helpers/database/user-auth.php
```diff
@@ -306,24 +311,3 @@ function generateAPIKey($user): string
 
     return $newKey;
 }
-
-function GetAPIKey($user): ?string
```
luchaos, line 310: Not used anymore

app_legacy/Helpers/render/game.php
```diff
@@ -147,9 +150,9 @@ function renderGameBreadcrumb(array|int $data, bool $addLinkToLastCrumb = true):
     return $html;
 }
 
-function renderGameCard(int|string|array $game): string
```
luchaos, line 150: `string` would've been the ID as string.

app_legacy/Helpers/render/leaderboard.php
```diff
@@ -102,7 +104,8 @@ function RenderScoreLeaderboardComponent(string $user, bool $friendsOnly, int $n
         // Create a tab for each of the 5 leaderboard types
         $id = uniqid();
         echo "<div class='tab'>";
-        for ($i = 0; $i < count($lbTypes); $i++) {
+        $lbTypesCount = count($lbTypes);
+        for ($i = 0; $i < $lbTypesCount; $i++) {
```
luchaos, line 108: (Rector) This makes a difference in performance

composer.json
```diff
@@ -148,7 +149,7 @@
         "mfs": "php artisan migrate:refresh --seed",
         "octane": "php artisan octane:start --watch --port=64888",
         "paratest": "php artisan test --parallel",
-        "rector": "vendor/bin/rector process && composer fix",
```
luchaos, line 151: `composer fix` would not run if rector does not finish successfully

database/factories/AchievementFactory.php
```diff
@@ -14,9 +14,9 @@ class AchievementFactory extends Factory
     public function definition(): array
     {
         return [
-            'title' => ucwords($this->faker->words(2, true)),
+            'title' => ucwords(fake()->words(2, true)),
```
luchaos, line 17: Suggested by Laravel code style rector rule

phpstan.neon
```diff
@@ -5,9 +5,9 @@ parameters:
         # - app
         - app_legacy
         - public
+        - resources/views
         - tests
-    # TODO v2 was at level 6
-    level: 5
+    level: 6
```
luchaos, line 10: Level 7 would produce additional ~400 errors

public/API/API_GetAchievementOfTheWeek.php
```diff
@@ -78,7 +78,7 @@
     'ID' => $staticData['Event_AOTW_ForumID'] ?? null,
 ];
 
-$unlocks = getAchievementUnlocksData($achievementID, $numWinners, $numPossibleWinners, $user, 0, 500);
+$unlocks = getAchievementUnlocksData((int) $achievementID, $numWinners, $numPossibleWinners, null, 0, 500);
```
luchaos, line 81: No $user here.

app_legacy/Helpers/database/achievement.php
```diff
@@ -35,32 +40,27 @@ function getAchievementsList(
                 LEFT JOIN GameData AS gd ON gd.ID = ach.GameID
                 LEFT JOIN Console AS c ON c.ID = gd.ConsoleID ";
 
-    if (isset($achievementType)) {
```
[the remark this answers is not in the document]
Jamiras: Looks like it did for about three months in 2019: https://github.com/RetroAchievements/RAWeb/commit/7e227e8cd0669452a652ac9727ddadb3951bc2de

public/leaderboardList.php
```diff
@@ -435,11 +431,12 @@ function UpdateLeaderboard(lbID) {
...
-        // Max number fetched, i.e. there are more. Can goto next 25.
-        $nextOffset = $offset + $maxCount;
-        echo "<a href='/achievementList.php?s=$sortBy&amp;o=$nextOffset'>Next $maxCount &gt;</a>";
-    }
+    // TODO fix pagination - $listCount is not being increased
+    // if ($listCount == $maxCount) {
+    //     // Max number fetched, i.e. there are more. Can goto next 25.
+    //     $nextOffset = $offset + $maxCount;
+    //     echo "<a href='/achievementList.php?s=$sortBy&amp;o=$nextOffset'>Next $maxCount &gt;</a>";
+    // }
```
Jamiras, line 439: TODO needs to be addressed

app_legacy/Helpers/database/user-account-deletion.php
```diff
@@ -61,76 +68,43 @@ function deleteRequest($username, $date = null): bool
...
+    $threshold = Carbon::today()->setTime(8, 0)->subWeeks(2);
 
-    $dbResult = s_mysql_query($query);
-    if (!$dbResult) {
-        return;
-    }
+    /** @var Collection<int, User> $users */
+    $users = User::where('DeleteRequested', '<=', $threshold)
+        ->orderBy('DeleteRequested')
+        ->get();
```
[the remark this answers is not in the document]
luchaos: Laravel adds this clause by default. To included deleted models it requires `withTrashed()`

app_legacy/Helpers/database/achievement.php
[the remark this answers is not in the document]
luchaos: I see - thanks for looking it up! Guess it's fine to have that check removed then.

phpstan.neon
luchaos: Level 7/8 will be addressed in a separate, smaller PR

app_legacy/Helpers/database/user-account-deletion.php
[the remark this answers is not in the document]
luchaos: So it was intentional, yes :) Thanks for pointing it out, i had that on my self-review list initially to be mentioned explicitly.
Using `withTrashed` to get all soft-deleted models is often the cause for confusion and is important to keep in mind; e.g. when building console commands for data migration like here https://github.com/RetroAchievements/RAWeb/blob/master/app_legacy/Site/Commands/UpdateUserTimestamps.php#L83

app_legacy/Helpers/database/user.php
```diff
@@ -310,38 +310,30 @@ function getControlPanelUserInfo($user, &$libraryOut): bool
...
-    sanitize_sql_inputs($offset, $count, $requestedBy, $perms);
-    settype($offset, 'integer');
-    settype($count, 'integer');
-    settype($showUntracked, 'boolean');
-
     $whereQuery = null;
     $permsFilter = null;
 
-    settype($perms, 'integer');
+    $perms = (int) $perms;
```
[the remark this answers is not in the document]
luchaos: Good point, it receives an `int` at all times, too. I'll make it not-nullable, set it to Permissions::Unregistered (0) by default and remove the superfluous pass-by-reference.

app_legacy/Helpers/database/user.php
```diff
@@ -49,8 +40,12 @@ function getAccountDetails(&$username, &$dataOut): bool
     return true;
 }
 
-function getUserIDFromUser($user): int
+function getUserIDFromUser(?string $user = null): int
```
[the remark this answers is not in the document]
luchaos: ℹ️ Ah, it's the other way around, I remembered incorrectly when adding those defaults.

When defining the parameter with a null default without making it a nullable type PHP will allow it but some code quality tools will complain, e.g.:
```
function getUserIDFromUser(string $user = null): int
```
which is cleaner like this:
```
function getUserIDFromUser(?string $user = null): int
```

public/leaderboardList.php
luchaos: Doesn't look like this worked before.
https://retroachievements.org/leaderboardList.php?g=3680
It incorrectly links to `achievementList.php`, the listCount is not calculated properly either. Nice to see PHPStan finding bugs like this :)
I'll fix that.

app_legacy/Helpers/database/leaderboard.php
```diff
@@ -101,110 +90,9 @@ function SubmitLeaderboardEntryJSON($user, $lbID, $newEntry, $validation): array
     return $retVal;
 }
 
-function submitLeaderboardEntry($user, $lbID, $newEntry, $validation, &$dataOut): bool
```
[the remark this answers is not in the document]
luchaos: Renamed `SubmitLeaderboardEntryJSON` to `SubmitLeaderboardEntry`

public/leaderboardList.php
luchaos: Addressed in https://github.com/RetroAchievements/RAWeb/pull/1404

app_legacy/Helpers/database/leaderboard.php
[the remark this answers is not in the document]
luchaos: Addressed in https://github.com/RetroAchievements/RAWeb/pull/1404.

app_legacy/Helpers/database/player-rank.php
```diff
@@ -162,4 +153,6 @@ function getUserRank(string $username, int $type = RankType::Hardcore): ?int
 
         return (int) legacyDbFetch($query)['UserRank'];
     });
+
+    return $cachedRank ?? null;
```
Jamiras, line 157: isn't `?? null` redundant?

app_legacy/Helpers/database/subscription.php
```diff
@@ -9,22 +9,19 @@
  */
 function updateSubscription(string $subjectType, int $subjectID, int $userID, bool $state): bool
 {
-    sanitize_sql_inputs($subjectType, $subjectID, $userID, $state);
-    settype($state, 'integer');
+    sanitize_sql_inputs($subjectType);
+    $state = (int) $state;
```
Jamiras, line 13: I think this line can be eliminated because of the logic on line 15.

luchaos: Ah yes. I updated most of the bool to int casts to work the same.
This works just fine:
```php
echo (int) true;
// 1
```

app_legacy/Helpers/database/player-rank.php
luchaos: Yeah, that makes no sense anymore :D 

That was a leftover from when I tried to make sure that any cached string value is cast to int, too, for `declare(strict_types=1)` not causing `TypeError`s.

Reverted to what you initially had written to directly return the cached value.

app_legacy/Helpers/database/user.php
```diff
@@ -3,27 +3,16 @@
...
-
-    if (!$dbResult || mysqli_num_rows($dbResult) != 1) {
-        log_sql_fail();
-
-        // failed: Achievement $id doesn't exist!
-        return null;
-    }
-
-    return mysqli_fetch_assoc($dbResult);
+    return User::find($username)?->toArray();
```
Jamiras, line 10: This doesn't seem to work.
```
$ curl "localhost/dorequest.php?r=awardachievement&u=Jamiras&a=40403&h=1&m=8e3630186e35d477231bf8fd50e54cdd&t=xxxxxxxxxxxxxxxx"
{"Success":false,"Error":"User data cannot be found for Jamiras","Score":84514,"SoftcoreScore":1100,"AchievementID":40403}
```
Changing it to `firstWhere('User', $username)` does.

luchaos: Good catch, thanks! I missed that the username is not the primary key there - `find()` only works with the primary key.

I'll fix that and check that i did not add this mistake anywhere else.

app_legacy/Helpers/database/leaderboard.php
```diff
@@ -22,34 +26,19 @@ function SubmitLeaderboardEntry($user, $lbID, $newEntry, $validation): array
...
         $retVal['LBData'] = $lbData;
-        settype($retVal['LBData']['LeaderboardID'], 'integer');
-        settype($retVal['LBData']['GameID'], 'integer');
-        settype($retVal['LBData']['LowerIsBetter'], 'integer');
+        $retVal['LBData']['LeaderboardID'] = (int) $retVal['LBData']['LeaderboardID'];
+        $retVal['LBData']['GameID'] = (int) $retVal['LBData']['GameID'];
+        $retVal['LBData']['LowerIsBetter'] = (int) $retVal['LBData']['LowerIsBetter'];
 
         $retVal['Score'] = $newEntry;
-        settype($retVal['Score'], 'integer');
```
Jamiras, line 39: `$retVal['Score']` needs to be an integer. The simple parser in the rcheevos code gets confused by the quotation mark.

luchaos: I'm not sure how this can be a non-integer at this point. When trying to cast it to an int explicitly my IDE complains that it's already an integer:

`$retVal['Score'] = (int) $newEntry;`

I'm missing something.

Jamiras: `sanitize_sql_inputs` (line 16) is changing it.

app_legacy/Helpers/database/user.php
luchaos: Fixed in https://github.com/RetroAchievements/RAWeb/pull/1393/commits/c534351ad4379613b3f5c1e224771c680108b453

app_legacy/Helpers/database/leaderboard.php
luchaos: oh 🤦 thank you! forgot to remove the integers from the sql sanitization in this case.
Fixed in https://github.com/RetroAchievements/RAWeb/pull/1393/commits/08d2381ca160369cce6cce76c808ed8151abcdf0

app_legacy/Helpers/database/user.php
```diff
@@ -3,27 +3,16 @@
...
-
-    if (!$dbResult || mysqli_num_rows($dbResult) != 1) {
-        log_sql_fail();
-
-        // failed: Achievement $id doesn't exist!
-        return null;
-    }
-
-    return mysqli_fetch_assoc($dbResult);
+    return User::firstWhere($username)?->toArray();
```
Jamiras, line 10: This still doesn't work.
```
Unknown column 'Jamiras' in 'where clause'
```
You need `firstWhere('User', $username)`

luchaos: Oh my, thanks! Fixed. I should get more sleep ...

Jamiras (approved):

PR #1394, opened 2023-02-28
feat(game): show author names on individual unofficial achievements

wescopeland:
This PR makes a slight change when viewing a game's list of unofficial achievements.

If the currently logged in user is a Junior Developer or greater, authorship metadata is shown for each achievement in the list. If not, nothing changes with the list.

This change is motivated by some feedback I've received: 

> for code reviewers, when viewing the list of unofficial achievements it's difficult to tell which Junior dev (or otherwise) authored which achievement. When a Junior picks up a set that had achievements in unofficial limbo from a dev that gave up (or a set that got demoted for being broken) and they don't ask to have them renamed to [void], it's hard to tell which ones to check when reviewing

Screenshot 2023-02-28 at 4 08 12 PM


public/gameInfo.php
```diff
@@ -1295,6 +1296,9 @@ function ResetProgress() {
                             echo " <span class='TrueRatio'>($achTrueRatio)</span>";
                             echo "</div>";
                             echo "<div class='mb-2'>$achDesc</div>";
+                            if ($flags != $officialFlag && isset($user) && $permissions >= Permissions::JuniorDeveloper) {
+                                echo "<p class='text-2xs'>Author: <a href='/user/$achAuthor/'>$achAuthor</a></p>";
+                            }
```
Jamiras, line 1301: Why not use `userAvatar`? These links won't have hover tooltips, whereas the author links at the top of the page do.

Jamiras, line 1301: Why not use `userAvatar`? These links won't have hover tooltips, whereas the author links at the top of the page do.

wescopeland: Fixed

Jamiras (approved):

PR #1402, opened 2023-03-02
Add validation to web api user summary and recently played games

luchaos:
This PR acts as a template for other web api endpoint validation.

It applies Laravel's validation to `API_GetUserSummary` and `API_GetUserRecentlyPlayedGames`, comparable to request files in `public/request/`, without making too many assumptions about how our endpoints are used. Adding basic validation to them however might reduce resource usage caused by invalid requests (cpu, db hits, ...) and logging useless errors.

Calling `/API/API_GetUserSummary.php?y=API_KEY&u=&g=nope&a=-1` results in:

```
{
  "message": "The u field is required. (and 2 more errors)",
  "errors": {
    "u": [
      "The u field is required."
    ],
    "g": [
      "The g must be an integer."
    ],
    "a": [
      "The a must be at least 0."
    ]
  }
}
```

Fixes 
```
getUserPageInfo(): Argument #1 ($user) must be of type string, null given, called in public/API/API_GetUserSummary.php on line 85 {"userId":24699,"exception":"[object] (TypeError(code: 0): getUserPageInfo(): Argument #1 ($user) must be of type string, null given, called in public/API/API_GetUserSummary.php on line 85 at app_legacy/Helpers/database/user.php:226)
```

### DIFF
--- a/public/API/API_GetUserRecentlyPlayedGames.php
+++ b/public/API/API_GetUserRecentlyPlayedGames.php
@@ -22,11 +22,12 @@
  *    int        ScoreAchievedHardcore    number of points earned by the user in hardcore
  */
 
+use App\Support\Rules\CtypeAlnum;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
 
 $input = Validator::validate(Arr::wrap(request()->query()), [
-    'u' => 'required',
+    'u' => ['required', 'min:2', 'max:20', new CtypeAlnum()],
     'c' => 'nullable|integer|min:0',
     'o' => 'nullable|integer|min:0',
 ]);
--- a/public/API/API_GetUserRecentlyPlayedGames.php
+++ b/public/API/API_GetUserRecentlyPlayedGames.php
@@ -22,15 +22,21 @@
  *    int        ScoreAchievedHardcore    number of points earned by the user in hardcore
  */
 
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'u' => 'required',
+    'c' => 'nullable|integer|min:0',
+    'o' => 'nullable|integer|min:0',
+]);
+
 $user = request()->query('u');
 $count = min((int) request()->query('c', '10'), 50);
 $offset = (int) request()->query('o');
 
 $recentlyPlayedData = [];
-$numRecentlyPlayed = 0;
-if (!empty($user)) {
-    $numRecentlyPlayed = getRecentlyPlayedGames($user, $offset, $count, $recentlyPlayedData);
-}
+$numRecentlyPlayed = getRecentlyPlayedGames($user, $offset, $count, $recentlyPlayedData);
 
 if (!empty($recentlyPlayedData)) {
     $gameIDsCSV = $recentlyPlayedData[0]['GameID'];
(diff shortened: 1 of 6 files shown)


public/API/API_GetUserRecentlyPlayedGames.php
```diff
@@ -22,15 +22,21 @@
  *    int        ScoreAchievedHardcore    number of points earned by the user in hardcore
  */
 
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'u' => 'required',
```
luchaos, line 29: Note that this could include a lookup:
```
'u' => 'required|exists:mysql_legacy.UserAccounts,User',
```
I opted to not do that to reduce db hits as user data is fetched after validation anyways if needed.

Jamiras: I think it's reasonable to not validate the presence of the data at this point.

I've already noticed discrepancies between the various APIs when a non-existant user is queried. Most return an empty object, some return a 404 error, and one returns a fully populated (but zeroed out) response.

tests/Feature/Api/V1/UserRecentlyPlayedGamesTest.php
```diff
@@ -20,6 +20,20 @@ class UserRecentlyPlayedGamesTest extends TestCase
     use RefreshDatabase;
     use BootstrapsApiV1;
 
+    public function testItValidates(): void
+    {
+        $this->get($this->apiUrl('GetUserRecentlyPlayedGames', [
+            'u' => null,
```
Jamiras, line 26: Is there a difference between explicitly stating null and not providing the parameter?

Jamiras (approved):

tests/Feature/Api/V1/UserRecentlyPlayedGamesTest.php
luchaos: Not in this case, no - added it for brevity. It could also be `''` and should result in the same validation error.

luchaos: Removed it as it seems superfluous now that you mentioned it :)

public/API/API_GetUserRecentlyPlayedGames.php
```diff
@@ -22,11 +22,12 @@
  *    int        ScoreAchievedHardcore    number of points earned by the user in hardcore
  */
 
+use App\Support\Rules\CtypeAlnum;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
 
 $input = Validator::validate(Arr::wrap(request()->query()), [
-    'u' => 'required',
+    'u' => ['required', 'min:2', 'max:20', new CtypeAlnum()],
```
luchaos, line 30: Improved validation slightly to match the isValidUsername helper:
https://github.com/RetroAchievements/RAWeb/blob/master/app_legacy/Helpers/util/string.php#L29-L37
I'll look into creating a custom validation rule that combines those in a serparate PR.

tests/Feature/Api/V1/UserCompletedGamesTest.php
```diff
@@ -17,6 +17,14 @@ class UserCompletedGamesTest extends TestCase
     use RefreshDatabase;
     use BootstrapsApiV1;
 
+    public function testItValidates(): void
+    {
+        $this->get($this->apiUrl('GetUserSummary'))
```
luchaos: GetUserSummary came up in the logs, too.

Jamiras (approved):

PR #1404, opened 2023-03-02
Remove leaderboards pagination & other unused code

luchaos:
Leaderboard list were changed in the past to require a user, game ID to be set, not allowing for console filter anymore. Code that still assumes otherwise is removed in this PR.


public/leaderboardList.php
```diff
@@ -206,7 +126,7 @@ function UpdateLeaderboard(lbID) {
     $sort6 = ($sortBy == 6) ? 16 : 6;
     $sort7 = ($sortBy == 7) ? 17 : 7;
 
-    if (isset($gameData) && isset($user) && $permissions >= Permissions::JuniorDeveloper) {
+    if ($permissions >= Permissions::JuniorDeveloper) {
```
Jamiras, line 129: The `else` makes the titles clickable for sorting, but without the `g` parameter, doing so generates 404 errors. Recommend either adding the `g` parameter to the links, or removing the else block. Removing the else block would also eliminate the need for the `s` parameter and all of its related code.

I don't think a non-dev can get to this page without being handed a URL, so maybe the whole page should be restricted to jr-dev+.

luchaos: I find it tempting to remove all of that.
However, I think there's some value in having a list page for a game's leaderboards for standard users. It should be linked on a game's page, too, maybe even replacing the long lists in the sidebar and just keeping most recent leaderboard changes in a much more condensed form and a "more" button below.
Adding the game ID is an easy change. Additionally, non devs should be presented with the table in full width. Done in https://github.com/RetroAchievements/RAWeb/pull/1404/commits/9e852eca643b239a1d2f0ae8fe496ebf3add136e

Jamiras: This list definitely needs some additional work, especially as it's only accessible by non-developers.
image

* Game is completely unnecessary. All leaderboards will be for the current game.
* I can't think of any reason a general user would care about the ID, or ordering by the ID.
* Similarly, Type has very little value for general users. We don't even display it on the individual leaderboard details page.
* As a user, I'd probably expect to see the top entry (just like is already shown in the sidebar).

After all of those changes, the only real difference between this and the sidebar is swapping Entries for top entry and the sortable headers. 

I'd also argue that Display Order should be an available sort order, but then we'd have to discard the headers in favor of something like we use for the achievements list:
image

Jamiras: Adding the game id to the links at least makes the page functional. If further changes are desired, they can always be done later.

Jamiras (approved):

public/leaderboardList.php
luchaos: Yeah, I agree; all those items you mentioned are welcome improvements for this page. I'll leave those for another improvement PR? Making this functional again and a first pass of removing excess code was the goal for this one.

PR #1408, opened 2023-03-03
feat(game): add support for guide URLs

wescopeland:
This PR adds support for entry and display of RAGuide URLs to the game page.

Related discussion: https://github.com/RetroAchievements/RAWeb/discussions/1051
There is a lot of dialogue in here on how this can be approached and/or implemented. My perception is there is a sentiment that it's better to support linking to our existing RAGuides as opposed to doing nothing or waiting until we have a grand design for this feature. I believe many users may not even be aware the RAGuides exist, which is a shame because they are very well written.

**How?**

From a back-end POV, a new column has been added to the `GameData` table called `GuideURL`. The existing _update-meta.php_ call has been leveraged to update this data. The legacy create_base_tables migration has been updated to insert this column.

**UI/UX**

If an unregistered user visits a game page with no associated RAGuide, their experience is unchanged from what's live today.

If a registered user visits a game page with no associated RAGuide, they see a new link on the right-hand side. When clicked, this takes them to a page on how to contribute an RAGuide.

Screenshot 2023-03-02 at 10 30 29 PM

If a registered user visits a game page with an associated RAGuide, they see a new link that when clicked takes them to the RAGuide.

Screenshot 2023-03-02 at 10 31 01 PM

If a developer goes to the dev menu, they see a new field for the RAGuide Link. This field is protected by `active_url` validation.

Screenshot 2023-03-02 at 10 37 03 PM

**Things I am unsure of**

- Did I handle the migration correctly? This is my first time writing a migration like this.
- Is `GameData` the appropriate table for this? This feels like metadata, so it felt like the right fit, but it's possible I'm missing context about the system.


app_legacy/Helpers/database/game.php
```diff
@@ -12,7 +12,7 @@ function getGameData(int $gameID): ?array
         return null;
     }
 
-    $query = "SELECT gd.ID, gd.Title, gd.ConsoleID, gd.ForumTopicID, IFNULL( gd.Flags, 0 ) AS Flags, gd.ImageIcon, gd.ImageTitle, gd.ImageIngame, gd.ImageBoxArt, gd.Publisher, gd.Developer, gd.Genre, gd.Released, gd.IsFinal, c.Name AS ConsoleName, c.ID AS ConsoleID, gd.RichPresencePatch
+    $query = "SELECT gd.ID, gd.Title, gd.ConsoleID, gd.ForumTopicID, IFNULL( gd.Flags, 0 ) AS Flags, gd.ImageIcon, gd.ImageTitle, gd.ImageIngame, gd.ImageBoxArt, gd.Publisher, gd.Developer, gd.Genre, gd.Released, gd.IsFinal, gd.RAGuideURL, c.Name AS ConsoleName, c.ID AS ConsoleID, gd.RichPresencePatch
```
wescopeland, line 15: This does cause `RAGuideURL` to be attached to `API_GetGameExtended.php` and `API_GetGameInfoAndUserProgress.php`. This is additive so it should not be a breaking change. The value will either be `null` or a string.

wescopeland:
Per https://discord.com/channels/310192285306454017/1048299730670735421/1081166912408535110, language for "RAGuide" has been genericized to "Guide".

database/migrations/legacy/2012_10_03_133633_create_base_tables.php
```diff
@@ -375,6 +376,12 @@ public function up()
             // UPDATE `GameData` SET `Created` = NULL WHERE `Created` IS NOT NULL;
         }
 
+        if (!Schema::hasColumns('GameData', ['GuideURL'])) {
+            Schema::table('GameData', function (Blueprint $table) {
+                $table->string('GuideURL')->nullable();
+            });
+        }
```
Jamiras, line 383: This should be its own step, or databases that have already run this step will not apply the change. I don't think we have any for adding a column yet, but here's one for adding an index: https://github.com/RetroAchievements/RAWeb/blob/master/database/migrations/legacy/2022_10_01_000000_update_ticket_index.php.

Basically, you'll end up with these five lines in your `up()` and a `dropColumn` call in your `down()` function.

Running `php artisan migrate` should add the column and `php artisan migrate:rollback` should remove it.

wescopeland: Cool, that makes perfect sense. I was wondering about this. Will revise later today.

wescopeland: All set

public/gameInfo.php
```diff
@@ -864,6 +865,8 @@ function ResetProgress() {
                     echo "<div class='text-right'><button class='btn'>Submit</button></div>";
                     echo "<label for='game_release'>First Released</label><input type='text' name='release' id='game_release' value='" . attributeEscape($released) . "' class='w-full'>";
                     echo "<div class='text-right'><button class='btn'>Submit</button></div>";
+                    echo "<label for='guide_url'>Guide URL</label><input type='url' name='guide_url' id='guide_url' value='" . attributeEscape($guideURL) . "' class='w-full'>";
+                    echo "<div class='text-right'><button class='btn'>Submit</button></div>";
```
Jamiras, line 869: Recommend moving this down near Forum Topic to be more consistent with the page layout

public/gameInfo.php
```diff
@@ -1384,8 +1390,12 @@ function ResetProgress() {
                 if ($numAchievements == 0) {
                     echo "<li><a class='btn py-2 mb-2 block' href='/setRequestors.php?g=$gameID'><span class='icon icon-md ml-1 mr-3'>📜</span>Set Requestors</a></li>";
                 }
+                if ($numAchievements >= 1 && (!isset($guideURL) || (empty($guideURL)))) {
+                    echo "<a class='btn py-2 mb-2 block' href='https://github.com/RetroAchievements/guides/wiki' target='_blank' rel='noreferrer'><span class='icon icon-md ml-1 mr-3'>📖</span>Create a Guide</a>";
+                }
```
Jamiras, line 1395: I'm not sold on having a "Create" button on every page. Most users will never use it, and [#1352](https://github.com/RetroAchievements/RAWeb/pull/1352/files#diff-8e708c2e13ee510e7add867723475fd6944bcb7661604d8456d720e0f9c540a7R153) adds this same link to the Community menu.

public/gameInfo.php
wescopeland: Done

public/gameInfo.php
wescopeland: Fair enough, removed

wescopeland:
@Jamiras @luchaos One thing I'm concerned about after stepping away from this for a bit and coming back to it: the language on the link could possibly be ambiguous.

Screenshot 2023-03-05 at 10 31 14 AM

I am afraid this implies the link leads to an official guide for the game (eg: a Nintendo Power guide), rather than our in-house RAGuides.

Perhaps "Official Set Guide" or "Official Achievements Guide" would be more fitting?

luchaos:
> Perhaps "Official Set Guide" or "Official Achievements Guide" would be more fitting?

I'd go for just "Guide" as mentioned in #site-feedback.

Jamiras (approved):

PR #1409, opened 2023-03-03
exclude sets for rollout consoles from new sets widget on main page

Jamiras:
Claims for games for rollout consoles can be completed from Unofficial, but not played by users yet. This filters completed claims for rollout consoles from the main page. They'll still appear in the page that loads when clicking the "more..." link, and users who have requested the set will still be notified that the claim was completed.

https://discord.com/channels/310192285306454017/1067638638919307414/1081271680833429604

### DIFF
--- a/app_legacy/Helpers/render/set-claim.php
+++ b/app_legacy/Helpers/render/set-claim.php
@@ -60,8 +60,6 @@ function renderFinishedClaimsComponent(int $count): void
     echo "<div class='component'>";
     echo "<h3>New Sets/Revisions</h3>";
 
-    $claimData = getFilteredClaims(null, ClaimFilters::AllCompletedPrimaryClaims, ClaimSorting::FinishedDateDescending, false, null, 0, $count);
-
     echo "<table class='table-highlight mb-1'>";
     echo "<thead>";
     echo "<tr class='do-not-highlight'>";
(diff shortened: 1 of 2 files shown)


app_legacy/Helpers/render/set-claim.php
```diff
@@ -74,25 +72,47 @@ function renderFinishedClaimsComponent(int $count): void
...
-        echo gameAvatar($claim, icon: false);
-        echo "</td>";
-        echo "<td>" . ($claim['SetType'] == ClaimSetType::NewSet ? ClaimSetType::toString(ClaimSetType::NewSet) : ClaimSetType::toString(ClaimSetType::Revision)) . "</td>";
-        echo "<td class='smalldate'>" . getNiceDate(strtotime($claim['DoneTime'])) . "</td>";
-        echo "</tr>";
-    }
+
+    $remaining = $count;
+    $offset = 0;
+    do {
```
luchaos, line 78: Wow, I just realized that I haven't seen a `do {} while()` loop in a while :)
Regardless, I assume this this can be simplified in the future to not pass the offset and count as soon as the changes of https://github.com/RetroAchievements/RAWeb/pull/1393 are in?

luchaos (approved):

app_legacy/Helpers/render/set-claim.php
Jamiras: I don't think #1393 will have any impact on this. The problem being addressed is that the query is returning items where the console ID has not yet been validated (console is rollout, claim was completed in unofficial, so players can't actually see or play it yet). This solution queries the first $count items, and then applies the filter locally. If any items were filtered, it asks for me.

I realize it's not the most efficient way to fetch the additional records, but there's no easy way to turn `isValidConsoleId` into a CSV for use in the query, and the second fetch is going to be unnecessary most of the time.

luchaos: wait, that was in a different PR.
`getFilteredClaims()` already allows to retrieve all entries by not passing the `$limit` parameter?

luchaos: i understand now 👍

luchaos (approved):

PR #1411, opened 2023-03-04
fix: enable mobile tooltips on Android only

wescopeland:
This PR addresses some discussion that recently came up in #site-feedback regarding recent changes to mobile tooltips, specifically in the case of Android users.

It seems Android and iOS handle "hover" gestures differently, and things were not broken on Android as they are on iOS for tooltips.

```
Hey yall, Jamiras directed me here. Just wanted to voice that I liked / used site tooltips on mobile quite a bit (android chrome). not sure if it can be disabled for one browser and not another, or via user profile settings, but it never seemed to cause any issues on my platform. minor thing but wanted to pass it on, thanks for all the hard work :pray~1:

Same here, I was trying to pull them up the other day while looking at a ticket to see the description of an achievement by tap holding it but can't do that now. I used it fairly often for other things as well :pray~1:

I'll throw my hat in for having it reenabled on Android also. It is t perfect since it usually pops up the menu when I press, but it still allows me to see info that is useful
```

Changes in this PR:
* The `mobileSafeTipEvents` code has been migrated from _all.js_ to a TypeScript util.
* `onTouchStart` is no longer needed, as we are no longer basing this conditional on the presence of user touch. Therefore, this code has been deleted.
* Mobile tooltips now gracefully degrade if iOS and mobile are both detected. Otherwise, tooltips will always display. This is handled by user agent sniffing. I have chosen `ua-parser-js` to take care of this lift, as my impression is we don't want to be rolling our own method of sniffing the user agent.

I have tested this on desktop, the iOS simulator, emulated iOS, and emulated Android. This seems to achieve the effect we want. Tooltips are enabled for desktop and Android, but they are disabled on iOS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     },
     "dependencies": {
         "alpinejs": "^3.0.6",
-        "js-cookie": "^3.0.1"
+        "js-cookie": "^3.0.1",
+        "ua-parser-js": "^1.0.33"
     },
     "devDependencies": {
         "@defstudio/vite-livewire-plugin": "^1.0.0",
@@ -23,6 +24,7 @@
         "@tailwindcss/typography": "^0.5.0",
         "@types/alpinejs": "^3.7.1",
         "@types/js-cookie": "^3.0.3",
+        "@types/ua-parser-js": "^0.7.36",
         "@typescript-eslint/eslint-plugin": "^5.53.0",
         "@typescript-eslint/parser": "^5.53.0",
         "autoprefixer": "^10.4.7",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -13,7 +13,8 @@
     },
     "dependencies": {
         "alpinejs": "^3.0.6",
-        "js-cookie": "^3.0.1"
+        "js-cookie": "^3.0.1",
+        "ua-parser-js": "^1.0.33"
```
wescopeland, line 17: https://bundlephobia.com/package/ua-parser-js@1.0.33

luchaos (approved):

PR #1412, opened 2023-03-04
fix: add basic point pluralization to achievement tooltips

wescopeland:
This PR makes a very minor change to the "Points" label on achievement tooltips. If Points === 1, the label now displays "1 Point" instead of "1 Points".

Screenshot 2023-03-04 at 4 11 37 PM

### DIFF
--- a/app_legacy/Helpers/render/achievement.php
+++ b/app_legacy/Helpers/render/achievement.php
@@ -111,7 +111,7 @@ function renderAchievementCard(int|string|array $achievement, ?string $context =
     $tooltip .= "<div><b>$title</b></div>";
     $tooltip .= "<div class='mb-1'>$description</div>";
     if ($achPoints) {
-        $tooltip .= "<div>$achPoints " . ($achPoints == 1 ? "Point" : "Points") . "</div>";
+        $tooltip .= "<div>$achPoints " . __res('point', $achPoints) . "</div>";
     }
     if ($gameTitle) {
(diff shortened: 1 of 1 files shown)


app_legacy/Helpers/render/achievement.php
```diff
@@ -111,7 +111,7 @@ function renderAchievementCard(int|string|array $achievement, ?string $context =
     $tooltip .= "<div><b>$title</b></div>";
     $tooltip .= "<div class='mb-1'>$description</div>";
     if ($achPoints) {
-        $tooltip .= "<div>$achPoints Points</div>";
+        $tooltip .= "<div>$achPoints " . ($achPoints == 1 ? "Point" : "Points") . "</div>";
```
Jamiras, line 114: While this is perfectly fine, you could use Laravel's helper function:
```
       $tooltip .= "<div>$achPoints " . Str::plural('Point', $achPoints) . "</div>";
```
You can do it [from the lang](https://laravel.com/docs/10.x/localization#pluralization) files too, but I don't know if we're ready for that yet.

luchaos: yup, I wrote the `__res()` helper for that which will allow us to have everything translated in the future.
This should work fine anywhere: `__res('point', 1)`.

wescopeland: Cool, good to know. This has been adjusted to use `__res()`.

Jamiras (approved):

PR #1420, opened 2023-03-07
refactor(all.js): begin migration to TypeScript

wescopeland:
This PR migrates a small chunk of _all.js_ to TypeScript. I'll break this into multiple PRs so it isn't overwhelming- this is the first.

The existing `clipboard()` function has been renamed `copyToClipboard()`. Functionally it is the same.

NOTE: _twitch.php_ appears to be completely dead code. I took the liberty of deleting it in this PR (it referenced code being included in _all.js_). I am happy to revert this if we think it's worth keeping around.


luchaos:
> NOTE: _twitch.php_ appears to be completely dead code. I took the liberty of deleting it in this PR (it referenced code being included in _all.js_). I am happy to revert this if we think it's worth keeping around.

I have included that change/removal in https://github.com/RetroAchievements/RAWeb/pull/1393, too. Without the respective js code though. A very welcome change :)

luchaos (approved):

Issue #1421, opened 2023-03-08
[Bug] My Settings Page


StingX2:
https://retroachievements.org/controlpanel.php
Anytime I go to the My Settings page the entire page freezes until I get a non responsive update. My thought is because of the amount of games I have played it wants to load all that as the reset options are there. Unfortunately even if I load it and wait it still crashes my browser tab.

The only solution is to play this cat and mouse game with it closing the tab and reopening and after about 10 tries it will briefly let me hit a button (whether thats change my motto or recalc my score). Even after a succesful push of the button it is a guess if it works or not as the tab then freezes forever.

**Desktop (please complete the following information):**
 - Windows 11
 - Chrome

Blazekickn:
If you wait long enough it does load, but its not ideal

Jamiras:
It's definitely the game list. It took about 1 second to retrieve the data, but the web page was unresponsive for over a minute (There are about 1200 games in your list).

I suspect it's something in the way the list is being populated from the response. Maybe something similar to the changes made in #1391 could help.

StingX2:
> It's definitely the game list. It took about 1 second to retrieve the data, but the web page was unresponsive for over a minute (There are about 1200 games in your list).
> 
> I suspect it's something in the way the list is being populated from the response. Maybe something similar to the changes made in #1391 could help.

The rescore project had me load many many games sadly.

wescopeland:
@StingX2 Good news! I have found a nearly 100x performance optimization for this and will be opening a PR to remediate the issue later this evening.

StingX2:
> @StingX2 Good news! I have found a nearly 100x performance optimization for this and will be opening a PR to remediate the issue later this evening.

my hero :D

PR #1422, opened 2023-03-08
perf(settings): remediate freeze from reset progress list

wescopeland:
This PR optimizes the `GetAllResettableGamesList` function to improve loading performance on the Settings page for users with hundreds of games containing progress. This PR should resolve https://github.com/RetroAchievements/RAWeb/issues/1421.

Changes made:
- Instead of repeatedly updating `innerHTML`, a document fragment is constructed and then inserted into the DOM once.
- The code has been simplified using modern ES6 features.
- Comments have been added to improve readability.

The original implementation of this code caused the page to freeze up for multiple seconds when populating an HTML select field with options for users with large game lists. Now, the code is optimized to avoid the performance issues caused by updating the HTML inside a loop and accessing the DOM numerous times to do so. The result is a significant performance improvement.

I did a naive performance benchmark before and after the changes. With a user who had 400 games and on a very fast machine, the DOM insertion code took ~700ms to finish. After these changes it takes less than 5ms.

The time complexity of the original implementation was O(n^2), where `n` is the number of games retrieved from the DB. This new implementation is O(n).


Jamiras (approved):
This works as described in Chrome (I assume the implementation will work on other browsers, but did not check). There doesn't appear to be any noticeable irresponsiveness while the drop down is populated.

luchaos (approved):
Wonderful, thanks!

PR #1423, opened 2023-03-09
feat(game): require game icons to be 96x96 pixels

wescopeland:
This PR adds validation to uploading game icons to game pages. An error message is now shown if the user attempts to upload an image that is not exactly 96x96.

Screenshot 2023-03-09 at 5 42 39 PM

https://discord.com/channels/310192285306454017/1082859674002522192

### DIFF
--- a/public/request/game/update-image.php
+++ b/public/request/game/update-image.php
@@ -21,7 +21,7 @@ if ($input['type'] === ImageType::GameIcon) {
     Validator::make(
         request()->all(),
         ['file' => ['dimensions:width=96,height=96']],
-        ['file.dimensions' => 'The game icon must be exactly 96x96 pixels.']
+        ['file.dimensions' => 'Game icons are required to have dimensions of 96x96 pixels.']
     )->validate();
 }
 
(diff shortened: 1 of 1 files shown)


public/request/game/update-image.php
```diff
@@ -14,9 +14,17 @@
...
     'type' => ['required', 'string', Rule::in(ImageType::cases())],
-    'file' => 'image',
+    'file' => ['image'],
 ]);
 
+if ($input['type'] === ImageType::GameIcon) {
+    Validator::make(
+        request()->all(),
+        ['file' => ['dimensions:width=96,height=96']],
+        ['file.dimensions' => 'Game icons are required to have dimensions of 96x96 pixels.']
```
luchaos, line 24: For future reference: we'd like to have any user-facing text to be translatable.
The default, translatable, validation error would read `The :attribute has invalid image dimensions.` if i'm not mistaken.
I think it's fine to go with that and add a (translatable) message below the input field that tells that images should be of said dimensions, file type, etc?
I tend to not add to many variants of custom validation errors as it becomes hard to maintain quickly.

luchaos (approved):

PR #1424, opened 2023-03-11
fix(layout): update sitenav Hardest Achievements url

wescopeland:
Currently, the "Hardest Achievements" URL does not link to a list of achievements. It takes the user to a game list sorted by retro ratio. This could be more accurately labeled as "Hardest Games", but then it wouldn't belong under the "Achievements" menu.

This PR updates the URL to instead go to the site's global achievement list, pre-sorted by highest retro ratio.

https://discord.com/channels/310192285306454017/511718348178849792/1083779531065806929


wescopeland:
> Would it be fine to add the hardest games link under Games - Miscellaneous as well so that we don't lose access to that link?

Good feedback - fixed

luchaos (approved):

PR #1425, opened 2023-03-11
refactor: migrate homepage claims tables to Blade

wescopeland:
This PR deletes the legacy _set-claim.php_ file and migrates the "New Sets/Revisions" and "Sets in Progress" tables to Blade. There are also a few QoL improvements to the tables.

This PR gets us halfway to the components themselves being redesigned away from tables, but that isn't quite ready yet. This seems like the logical first step.

QoL changes:
* "Game" is now the first column of the tables instead of "User", since the game is the actual primary datapoint of both tables.
* "Game" and "User" are now on top of the avatar and have colspan="2" instead of starting on top of the text content.
* Empty states have been added for both tables (which will probably only appear in local instances of RAWeb).
* "User" column has been renamed to "Dev" to reduce ambiguity.
* Tooltips have been removed from Game columns on both tables, as the tooltips only render content that is already on the screen.
* Dates in both tables have been changed to TimeAgo format.
* The "New Sets/Revisions" table now has 6 rows instead of 5 to slightly dampen the more-recent effect of sets rolling off the home page within 24 hours.

---

**BEFORE**
Screenshot 2023-03-11 at 12 10 24 PM

Screenshot 2023-03-11 at 12 10 18 PM

---

**AFTER**
Screenshot 2023-03-11 at 12 09 46 PM

Screenshot 2023-03-11 at 12 09 54 PM

Ultimately, the intent is to redesign both of these sections completely (discussion is ongoing on Discord). However it seems reasonable to first get them migrated to Blade and away from legacy render functions.


resources/views/components/claim-new-table-row.blade.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+use Illuminate\Support\Carbon;
+
+$startedTimeAgo = Carbon::createFromFormat("Y-m-d H:i:s", $claim['DoneTime'])->diffForHumans();
```
Jamiras, line 5: This should use the claim created date, not the expiration date.

resources/views/community/components/claims/new-claims.blade.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+use LegacyApp\Community\Enums\ClaimFilters;
+
+$claimData = getFilteredClaims(
+    claimFilter: ClaimFilters::AllActiveClaims,
```
Jamiras, line 6: Recommend excluding collaboration claims while you're tinkering in here.
```
    claimFilter: ClaimFilters::AllActiveClaims & ~ClaimFilters::CollaborationClaim,
```
We don't need to see multiple entries for the same game.

Jamiras: Not sure if it's worth defining an `AllActivePrimaryClaims` constant for it (similar to `AllCompletedPrimaryClaims`)

resources/views/community/components/claims/finished-claims.blade.php
```diff
@@ -0,0 +1,53 @@
...
+use LegacyApp\Community\Enums\ClaimSorting;
+
+$claimData = getFilteredClaims(
+    null,
+    ClaimFilters::AllCompletedPrimaryClaims,
+    ClaimSorting::FinishedDateDescending,
+    false,
+    null,
+    0,
+    $count
```
Jamiras, line 13: recommend using named parameters as was done for new-claims.blade.php

resources/views/home.blade.php
```diff
@@ -2,19 +2,24 @@
...
-    RenderActivePlayersComponent();
-    ?>
-    <x-user.online-count-chart/>
-    <?php
-    renderNewClaimsComponent(5);
-    RenderRecentForumPostsComponent();
-    ?>
+
+    <x-news.carousel />
+    <x-claims.finished-claims count="6" />
```
Jamiras, line 7: I'm still only seeing 5 rows.

resources/views/components/claim-new-table-row.blade.php
wescopeland: 🤦‍♂️ D'oh. Addressed in https://github.com/RetroAchievements/RAWeb/pull/1425/commits/c2825edbcf8c4be73c349a67619991b666c00390.

resources/views/community/components/claims/new-claims.blade.php
wescopeland: > Recommend excluding collaboration claims while you're tinkering in here.

Good idea. I've made this adjustment in https://github.com/RetroAchievements/RAWeb/commit/c2825edbcf8c4be73c349a67619991b666c00390.

> Not sure if it's worth defining an AllActivePrimaryClaims constant for it (similar to AllCompletedPrimaryClaims)

I decided against this for the time being as this would be the only consumer of such a constant; I'm not sure if this would be a premature abstraction.

resources/views/community/components/claims/finished-claims.blade.php
wescopeland: Adjusted in https://github.com/RetroAchievements/RAWeb/commit/c2825edbcf8c4be73c349a67619991b666c00390.

As an aside, what is your opinion on best practice for this? Do you generally reach for named parameters when the count of args exceeds a certain number?

resources/views/home.blade.php
wescopeland: I am unable to reproduce this and continue to see 6 on my local.

Screenshot 2023-03-20 at 5 47 35 PM

Is it possible something is awry with your local environment / build?

Jamiras: I've got one item being hidden by #1409. Your implementation doesn't fetch another to fill the void.

resources/views/community/components/claims/finished-claims.blade.php
Jamiras: I generally used named parameters whenever a default parameter can be skipped. In this case, all the parameters have defaults and most can be skipped.

I sometimes also use it when it's unclear what a parameter is doing. In this case, it's not obvious what the two nulls are for without looking at the parameter names. I wouldn't use that alone as an argument to use named parameters though.

resources/views/community/components/claims/finished-claims.blade.php
```diff
@@ -0,0 +1,53 @@
...
+                        <th>Type</th>
+                        <th>Finished</th>
+                    </tr>
+                </thead>
+
+                <tbody>
+                    @foreach($claimData as $claim)
+                        @if (isValidConsoleId($claim['ConsoleID']))
+                            <x-claim-finished-table-row :claim="$claim" />
+                        @endif
```
Jamiras, line 43: this `if` is what's causing my 6th items to disappear.

resources/views/home.blade.php
wescopeland: Ah! I understand now.

In https://github.com/RetroAchievements/RAWeb/pull/1425/commits/14f4b5d1023eb834da4600e580cd8031b498a3b9, I've reimplemented the original do/while logic from _set-claim.php_ and have done a small refactor to get it down to a simple `while()` loop. This should resolve the issue.

resources/views/community/components/claims/finished-claims.blade.php
wescopeland: Thank you for this additional context - the root issue should be resolved via https://github.com/RetroAchievements/RAWeb/commit/14f4b5d1023eb834da4600e580cd8031b498a3b9

Jamiras (approved):

resources/views/community/components/claims/finished-claims.blade.php
wescopeland: Understood! That is helpful. With this in mind, I modified the function call to exclude the defaults ([ref](https://github.com/RetroAchievements/RAWeb/pull/1425/files#diff-6f464c0feeee2496593b4431164aa1fac3563f32f57ebb5e505cf60d9fc700f5R14)).

resources/views/components/claim-finished-table-row.blade.php
luchaos, line 1: Most components in /components are mostly base components and layout stuff.

As Set claims are a community feature (atm) I'd put those components in the community directory, too.

resources/views/home.blade.php
```diff
@@ -2,19 +2,24 @@
...
+
+    @php
```
luchaos, line 9: Personally I prefer sticking with the `<?php ?>` tag because the `@php` blade directive causes problems in some IDEs with formatting.
Can stay though imo.

resources/views/components/claim-finished-table-row.blade.php
luchaos: I haven't gotten around to adding all the previous v2 view files yet which would give a better idea how the directories are laid out. We'll get to that at

wescopeland: No worries at all, this makes perfect sense! I've addressed this in https://github.com/RetroAchievements/RAWeb/pull/1425/commits/8cd548615949816bd06e748a3a3faaa2ed0b8f3e.

resources/views/home.blade.php
wescopeland: Understood, for consistency I have reverted these back to `<?php ?>`. I was not aware of the formatting issues caused by the blade directive. This has been changed in https://github.com/RetroAchievements/RAWeb/pull/1425/commits/8cd548615949816bd06e748a3a3faaa2ed0b8f3e.

PR #1431, opened 2023-03-13
fix(hashes): add pluralization logic

wescopeland:
This fixes an issue on the game hashes page where there is no pluralization support for the revised copy released in 2.11.0.

Screenshot 2023-03-12 at 9 00 14 PM

I had a lot of trouble getting this to work with the `__res()` helper. Definitely interested if there is a better solution for this that I can learn from.


public/linkedhashes.php
```diff
@@ -41,7 +41,7 @@
              "<a href='https://docs.retroachievements.org/Game-Identification/'>here</a>." .
              "</b></p>";
 
-        echo "<p class='mt-4 mb-1'>There are currently <span class='font-bold'>" . count($hashes) . "</span> supported game file hashes registered for this game.</p>";
+        echo "<p class='mt-4 mb-1'>There " . (count($hashes) === 1 ? "is" : "are") . " currently <span class='font-bold'>" . count($hashes) . "</span> supported game file " . Str::plural('hash', count($hashes)) . " registered for this game.</p>";
```
Jamiras, line 44: Recommend calculating `count($hashes)` once.

wescopeland: Addressed in https://github.com/RetroAchievements/RAWeb/pull/1431/commits/d9d96adac1040ceeb5cce766f2205b91e09e6b26. I have also changed `Str::plural()` to instead use `__res()`.

PR #1432, opened 2023-03-13
Console / System icons in main menu and game lists

oddgoo:
**As per discussion**: #719

This PR adds system icons to both the main Dropdown menu:
image

And system game lists:

image

For this, I've added 67 icons to '\public\assets\images\system'. These include future systems like the DSi, and were sourced from:
http://yspixel.jpn.org/icon/game/index.htm, https://github.com/Tatohead/Console-Iconset, and a couple made by myself. The included readme from yspixel allows usage on websites.

**Further thoughts and considerations:**

- I changed the static rendering of system names in the main dropdown of layout.php to an array-based loop.

- I was asked to make the back-end define the icons instead of letting the front-end guess/fallback. I did this with the array on layout.php, but left the front-end fallback handler in gameList.php. This is because whatever definition we did there would end up clashing with layout.php, as that currently has hardcoded menu items, and refers to systems with names that are not in the database. _eg. NES/Famicom_.  

  In the future, once layout.php becomes a proper blade, we can revise uniting the icon logic / source of truth, and decide where that would best be declared. Currently the icon names simply follow the names in the console table:

  image

- I originally considered polling the community to decide between pixel icons and svg/vectors. But a genuine search did not yield any satisfactory vector icon packs. They were simply inconsistent and lacking. On the contrary, pixel icons were easier to source and complement (and look better to me, in the context of "Retro"achievements).


public/gameList.php
```diff
@@ -198,7 +198,13 @@ function ListGames(
                     echo "<br/>";
                 }
             } else {
-                echo "<h2>$consoleName</h2>";
+                $consoleIconName = str_replace("/", "", $consoleName); // Clean "Mega Drive / Genesis", etc.
+                $fallBackConsoleIcon = "assets/images/system/Unknown.png";
+                echo "<h2 class='flex'>";
+                echo " <img src='assets/images/system/$consoleIconName.png' class='mr-2' alt='Console icon' width='32' height='32'";
```
wescopeland, line 204: nit: Given this is a `flex` layout, spacing can be controlled with a `gap` utility rather than explicit margin.

```
<h2 class="flex gap-x-2">
```

public/gameList.php
```diff
@@ -198,7 +198,13 @@ function ListGames(
...
+                echo " onerror='this.src=\"$fallBackConsoleIcon\"'></img>"; // fallback
+                echo " <div class=''>$consoleName</div>";
```
wescopeland, line 206: nit: It is more semantically appropriate to use `<span>` instead of `<div>`, due to `<span>` being inline-level as opposed to block-level. Since the tag also doesn't have any class names, the `class` attribute should be safe to remove.

app_legacy/Helpers/render/layout.php
```diff
@@ -38,73 +38,90 @@ function RenderToolbar(): void
...
+        foreach ($column as $brand => $systems) {
+            echo "<li class='dropdown-header'>$brand</li>";
+            foreach ($systems as $system){
+                $systemName = $system['systemName'];
+                $listId = $system['listID'];
+                $iconName = $system['iconName'];
+                echo "<li><a href='/gameList.php?c=$listId'> <div class='flex items-center'>";
+                echo " <img src='assets/images/system/$iconName.png' class='mr-2' width='16' height='16' alt='mini console icon'>";
+                echo " <div>$systemName</div>";
+                echo "</div></a></li>";
```
wescopeland, line 120: nit:

This is the current HTML structure:

```html
<li>
  <a>
    <div class="flex items-center">
      <img class="mr-2" />
      <div>system name</div>
    </div>
  </a>
</li>
```

In this case, it's probably okay to get rid of the immediate child <div> of the <a> tag.

```html
<a class="flex items-center">...</a>
```

From there, a gap utility can be used to control the spacing:

```html
<a class="flex items-center gap-x-2">...</a>
```

app_legacy/Helpers/render/layout.php
```diff
@@ -38,73 +38,90 @@ function RenderToolbar(): void
...
+        echo "<ul>";
+        foreach ($column as $brand => $systems) {
+            echo "<li class='dropdown-header'>$brand</li>";
+            foreach ($systems as $system){
+                $systemName = $system['systemName'];
+                $listId = $system['listID'];
+                $iconName = $system['iconName'];
+                echo "<li><a href='/gameList.php?c=$listId'> <div class='flex items-center'>";
+                echo " <img src='assets/images/system/$iconName.png' class='mr-2' width='16' height='16' alt='mini console icon'>";
+                echo " <div>$systemName</div>";
```
wescopeland, line 119: nit: It is more semantically appropriate here to use `<span>` instead of `<div>`, as span is inline-level and div is block-level.

public/gameList.php
oddgoo: Amended!

public/gameList.php
oddgoo: Amended!

app_legacy/Helpers/render/layout.php
oddgoo: Amended! But "flex" needed to become "!flex" due to the class not being picked up when menu is closed.

app_legacy/Helpers/render/layout.php
oddgoo: Amended!

app_legacy/Helpers/render/layout.php
```diff
@@ -38,73 +38,90 @@ function RenderToolbar(): void
...
+                ["systemName" => "Nintendo 64", "listID" => 2, "iconName" => "Nintendo 64"],
+                ["systemName" => "Nintendo DS", "listID" => 18, "iconName" => "DS"],
+                // ["systemName" => "Nintendo DSi", "listID"=>78, "iconName" => "Nintendo DSi"],
+                ["systemName" => "Pokemon Mini", "listID" => 24, "iconName" => "Pokemon Mini"],
+                ["systemName" => "Virtual Boy", "listID" => 28, "iconName" => "Virtual Boy"],
+            ],
+            "Sony" => [
+                ["systemName" => "PlayStation", "listID" => 12, "iconName" => "PlayStation"],
+                ["systemName" => "PlayStation 2", "listID" => 21, "iconName" => "PlayStation 2"],
+                ["systemName" => "PlayStation Portable", "listID" => 41, "iconName" => "Playstation Portable"],
```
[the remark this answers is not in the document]
oddgoo: Fixed!

Jamiras:
All icons are showing in the menu now, but some aren't showing on the individual pages:
image
image
image
image
image
image
image
image
image
image

oddgoo:
@Jamiras The mismatch between the local names  and production has been fixed! Both the gameList and layout.php menu now use the unified config from system.php as recommended by @luchaos .

I have renamed all icons and re-checked all systems :)

Jamiras:
You have to make the filename cases match what it's expecting. Half my icons are missing because it's looking for lowercase names:
image

image
image

oddgoo:
> You have to make the filename cases match what it's expecting. Half my icons are missing because it's looking for lowercase names: image
> 
> image image

@Jamiras , my goodness, this was a case of git not picking up the casing changes. I've forced it to pick them all up now :).
Sorry this has taken so many reviews!

image

Jamiras (approved):

app_legacy/Helpers/render/layout.php
```diff
@@ -38,73 +38,92 @@ function RenderToolbar(): void
...
-    echo "<li><a href='/gameList.php?c=72'>WASM-4</a></li>";
-    echo "<li><a href='/gameList.php?c=63'>Watara Supervision</a></li>";
-    echo "<li><a href='/gameList.php?c=53'>WonderSwan</a></li>";
-    echo "</ul>";
+                $systemName = $system['systemName'];
+                $listId = $system['listID'];
+                $cleanSystemShortName = Str::lower(str_replace("/", "", config("systems.$listId.name_short")));
+                $iconName = Str::kebab($cleanSystemShortName);
+                echo "<li><a href='/gameList.php?c=$listId' class='!flex items-center gap-x-2' >"; // the flex class needs to be forced here
+                echo " <img src='assets/images/system/$iconName.png' width='16' height='16' alt=''>";
```
luchaos, line 120: This should be `asset("assets/images/system/$iconName.png")` to make sure image URLs are prefixed with the configured static host.

PR #1351, opened 2023-02-10
Refactor GetAchievementsEarnedBetween & GetAchievementsEarnedOnDay

luchaos:

PR #1435, opened 2023-03-14
feat(api): add API_GetClaims endpoint

wescopeland:
This PR adds a new endpoint: `API_GetClaims.php`, which allows consumers to fetch for different types of filtered claims: completed, dropped, or expired.

Additionally, claims endpoints now return a new `UserRole` property. This is valuable because users want to know if a set is under development by a junior dev.

**Why?**
Per https://discord.com/channels/310192285306454017/1084229773087223918, there is an extensive Set Claims spreadsheet that has been providing achievement developers quite a bit of value. However, now that the number of active claims has soared to above 400, it is becoming extremely time-intensive to maintain by hand.

The spreadsheet can be found here: https://docs.google.com/spreadsheets/d/15O4yL6CdGvCs_7THeAlb5fwI0qoDdbDiEZShiXY6E9c/edit?usp=sharing

At first, I considered it would just make sense to have this in RAWeb itself. That is probably still the case. However, I have been looking for a good demo project for https://github.com/RetroAchievements/retroachievements-api-js, and this seems like the perfect fit for something that can be built extremely quickly and can provide immediate value and relief.

Unfortunately, the web API only currently supports fetching active claims. The spreadsheet displays much more data than just active claims. We need additional fetching capabilities for all the functionality to be automated in a tiny web app.

Once this endpoint is deployed, the web API demo app can be completed. Here is a small preview of what it looks like now:
https://cdn.discordapp.com/attachments/1084229773087223918/1085314201293164614/New_Recording_Mar_14_2023_0530_PM.gif


tests/Feature/Api/V1/AchievementSetClaimTest.php
```diff
@@ -34,6 +34,151 @@ public function testGetActiveClaims(): void
...
+            ClaimSpecial::None,
+            Permissions::Developer
+        );
+        Carbon::setTestNow(Carbon::now()->addYears(2));
+
+        $claim = AchievementSetClaim::first();
+
+        $this->get($this->apiUrl('GetClaims', ['k' => '3']))
+            ->assertSuccessful()
+            ->assertJsonFragment(
```
wescopeland, line 161: No matter what I try, when I simulate the advancement of time, I cannot get `MinutesLeft` to work properly in this test. I believe this is an issue with SQLite.

public/API/API_GetClaims.php
```diff
@@ -0,0 +1,59 @@
...
+ *    string     GameTitle          title of the claimed game
+ *    string     GameIcon           site-relative path to the game's icon image
+ *    string     ConsoleName        console name of the claimed game
+ *    string     ClaimType          claim type: 0 - primary, 1 - collaboration
+ *    string     SetType            set type claimed: 0 - new set, 1 - revision
+ *    string     Status             claim status: 0 - active, 1 - complete, 2 - dropped
+ *    string     Extension          number of thes the claim has been extended
+ *    string     Special            flag indicating a special type of claim
+ *    string     Created            date the claim was made
+ *    string     DoneTime           date the claim is done
```
wescopeland, line 25: This documentation is mostly pasted from `API_GetActiveClaims.php`. As a result, _roughly half of these documented types are incorrect._ 

In the case of ID, UserRole, GameID, ConsoleID, ClaimType, SetType, Status, Extension, Special, and MinutesLeft, the type is actually `int`, not `string`.

I am happy to update these new docs, I just want to be sure we're also comfortable updating the ones in `API_GetActiveClaims.php` at the same time.

app_legacy/Helpers/database/set-claim.php
```diff
@@ -358,6 +356,7 @@ function getFilteredClaims(
     $selectCondition = "
         sc.ID AS ID,
         sc.User AS User,
+        ua.Permissions AS UserRole,
```
Jamiras, line 359: What is the need for exposing `ua.Permissions`? That probably won't translate easily into the new permissions matrix logic. Looking at the spreadsheet being replaced/augmented, it seems like it would be sufficient to have a `bool UserIsJrDev` property.

public/API/API_GetClaims.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+use Illuminate\Support\Facades\Validator;
+use Illuminate\Validation\Rule;
+use LegacyApp\Community\Enums\ClaimFilters;
+
+/*
+ *  API_GetClaims - returns information about 1000 max set claims.
```
Jamiras, line 8: Which 1000? I assume the latest 1000?

public/API/API_GetClaims.php
```diff
@@ -0,0 +1,60 @@
...
+$droppedClaims = '2';
+$expiredClaims = '3';
+
+$claimKind = request()->query('k', $completedClaims);
+
+$claimFilter = ClaimFilters::AllCompletedPrimaryClaims;
+if ($claimKind === $droppedClaims) {
+    $claimFilter = ClaimFilters::AllDroppedClaims;
+} elseif ($claimKind === $expiredClaims) {
+    $claimFilter = ClaimFilters::AllActiveClaims;
```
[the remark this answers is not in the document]
Jamiras: Would it be beneficial to have an `$activeClaims='4'` to return the unfiltered active claims, or is that too redundant with `API_GetActiveClaims`?

app_legacy/Helpers/database/set-claim.php
wescopeland: I understand your concern. Apologies for my weakness with SQL. Is something like this in the query what you're describing?:

```sql
CASE WHEN ua.Permissions = 2 THEN true ELSE false END AS UserIsJrDev
```

public/API/API_GetClaims.php
wescopeland: Addressed in https://github.com/RetroAchievements/RAWeb/pull/1435/commits/b3e44d98130302069814d21d7e7cdc790c808ab0

public/API/API_GetClaims.php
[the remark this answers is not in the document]
wescopeland: > I would expect "expired" claims to only return active claims past their Finished date. It would probably be easiest to do that as a post-processing step here (`MinutesLeft < 0`), but `getFilteredClaims` has a `getExpiringOnly` flag that could be expanded to an enum and then the filtering could be done at the query level.

Good point. I see where we would do this in the query itself. It appears `getFilteredClaims()` has several consumers. 
To reduce risk, I've opted for the post-processing approach in the API endpoint itself (https://github.com/RetroAchievements/RAWeb/commit/b3e44d98130302069814d21d7e7cdc790c808ab0).

> Would it be beneficial to have an `$activeClaims='4'` to return the unfiltered active claims, or is that too redundant with `API_GetActiveClaims`?

I considered the same thing and decided it was redundant and should not be added. I am happy to defer to your judgment here - if you believe we should have a `4` argument for active, I will add it.

public/API/API_GetClaims.php
Jamiras: We discussed that [here](https://github.com/RetroAchievements/RAWeb/pull/1351#discussion_r1103632477). The general feeling is that it should be okay.

We should update the comments to reflect the changes though.

app_legacy/Helpers/database/set-claim.php
Jamiras: Yes.

I'm not sure if it would be better to use "<= 2" instead of "= 2", as users with 0 or 1 Permissions shouldn't have claims, but if they did, they should be treated as Jr. Devs.

public/API/API_GetClaims.php
wescopeland: In https://github.com/RetroAchievements/RAWeb/pull/1435/commits/343f3c8b1241c7e1c2d17e9c517e620168f10885 I have updated the documentation for both endpoints.

app_legacy/Helpers/database/set-claim.php
wescopeland: I have added the above clause to the query in https://github.com/RetroAchievements/RAWeb/pull/1435/commits/343f3c8b1241c7e1c2d17e9c517e620168f10885.

I went with `<= 2`, since it does indeed seem possible a user with less than `2` can have a claim if they are demoted or banned.

public/API/API_GetClaims.php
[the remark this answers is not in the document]
wescopeland: Sounds good. I will not add a `4` option and will leave the existing endpoint alone. Agreed, we can always add it later if there's any demand for it.

app_legacy/Helpers/database/set-claim.php
Jamiras: You're still exposing `UserRole`, which may not translate well once we move to a permission matrix.

wescopeland: Oversight on my part, forgot to delete the line 🤦‍♂️
Should be resolved now.

Jamiras (approved):

PR #1439, opened 2023-03-16
feat: add drag and drop support for reordering site awards

wescopeland:
This PR makes numerous UX improvements to reorderSiteAwards.php, including the addition of drag and drop functionality.

**Changes in this PR:**
* _update-site-award.php_ has been removed.
* A new endpoint, _update-site-award**s**.php_ (plural) has been added.
* To mitigate rate-limiting issues, award reordering no longer automatically saves on any change.
* The page now has an empty state.
* A Save All Changes button has been added to the page. When clicked, the entire list of award orderings is saved on a single POST.
* Refresh is no longer required to see updates. The new endpoint returns the updated HTML for the right column in its response, and it is automatically injected into the DOM.
* Drag and drop is now available on the table rows.
* Hiding is controlled via a checkbox.
* Sections can be reordered by select boxes near the table heading labels.
* Manual move is available via buttons for mobile users. Depending on how many rows the user has in the table, three possible variations are displayed. For users with hundreds of awards, they see buttons to move up and down 1 row, move up and down 10 rows, and move to top/bottom.
* The UI code is written in TypeScript and lazy-loaded onto the page.

**Basic Award Reordering**

https://user-images.githubusercontent.com/3984985/227674581-60295932-38ae-4921-9afa-06b997b9e997.mp4

---

**Mobile Functionality (+ Manual Moves)**

https://user-images.githubusercontent.com/3984985/227674561-4e7cf437-1e93-4a8e-9432-d76c098ca5d8.mp4

---

**Hiding and Reordering Sections**

https://user-images.githubusercontent.com/3984985/227674538-c8e3f1b4-b893-40d6-8565-b8fa6dbfc11e.mp4


public/reorderSiteAwards.php
```diff
@@ -10,46 +10,171 @@
...
+
+function updateDisplayOrderValues(dropTableId, startingValue = 0) {
+    const tableRows = document.querySelectorAll(`#${dropTableId} tbody tr`);
+
+    tableRows.forEach((row,index) => {
+        const displayOrderInput = row.querySelector('.displayorderedit');
+        const currentValue = parseInt(displayOrderInput.value, 10);
+
+        if (currentValue !== -1) {
+            displayOrderInput.value = startingValue + (10 * index);
```
Jamiras, line 128: I'm not sure how much I like this solution. If I'm just swapping two badges, it seems unnecessary to reassign values for all the rest of them.

It also overwrites any manual changes I've made but not committed. i.e. given:
```
A   1
B   2
C   10
D   15
E   20
```
If I explicitly set D to 9, then drag A below B, this is what I end up with:
```
B   1
A   11
C   21
D   31
E   41
```
The fact that I wanted D before C is lost (and should maybe have been updated when I manually modified it).

It seems like having drag/drop + editable fields is undesirable. Maybe the editable field should be hidden and just reflect the indices? But then we have to decide how to handle hiding badges (negative indices) and group ordering (groups are ordered by the lowest visible badge index between groups).

Regarding hidden items: If I set my first item to be -99 (or and item to -99 and refresh), then any dragging operations automatically make the second item -89 and the third -79, etc., causing 9 additional items to be hidden.

public/reorderSiteAwards.php
```diff
@@ -61,30 +186,39 @@ function RenderAwardOrderTable(string $title, array $awards, int &$counter): voi
...
+                echo "<input type='hidden' name='data' value='$awardData'>";
+                echo "<input type='hidden' name='extra' value='$awardDataExtra'>";
 
                 echo "</tr>\n";
                 $counter++;
             }
             echo "</tbody></table>\n";
+
+            echo "<div class='flex w-full justify-end'>";
+            echo "<button onclick='updateAllAwardsDisplayOrder()'>Save</button>";
```
Jamiras, line 220: Having a Save button in each section is misleading. I can make changes to my Game Awards, but not Save, then go hide a Site Award and when I Save that, the Game Awards get saved too.

public/reorderSiteAwards.php
```diff
@@ -10,46 +10,171 @@
...
-            "The right panel represents how the site awards will look on your user page.</p>";
+        echo <<<HTML
+            <div class="embedded grid gap-y-4">
+                <p>
+                    To rearrange your site awards, modify the 'Display Order' value in the rightmost 
+                    column of each award row. The awards will be displayed on your user page in 
+                    ascending order based on these values. To hide an award, change its 'Display Order' 
+                    value to -1. Remember to save your changes by clicking the 'Save' button, and 
+                    your updates will appear on your user page immediately.
+                </p>
```
Jamiras, line 146: You may want to mention that the rows can be dragged to reorder them, as that's the feature being introduced.

public/reorderSiteAwards.php
```diff
@@ -10,46 +10,171 @@
...
+            $('#rightcontainer').html(response.updatedAwardsHTML);
+        })
+        .fail(function () {
+            showStatusMessage('Error updating awards');
+        });
+}
+
+function handleRowDragStart(event) {
+    currentGrabbedRow = event.target;
+    event.target.style.opacity = '0.3';
```
Jamiras, line 47: I'm not sure if there's an easy solution, but trying to highlight an existing order value by click/dragging in the editable field starts a drag action without highlighting the text.

wescopeland: I experienced this as well and tried multiple approaches to mitigate it, but was unable to land on a solution.

wescopeland: Based on the other feedback here I am going to experiment with getting rid of the Display Order field altogether. We may be able to use something like a checkbox for if the user wants to hide an award.

For group ordering, we can have something on the page for the user to manually determine which order they'd like their groups in. This doesn't seem too bad given there's only three of them.

wescopeland:
It seems like our destination is:
* Remove the Display Order field.
* Use checkboxes to designate awards which should be hidden (which will then float to the top of the list on a refresh).
* Have some solution for the user to manually sort their award groups.

The award indices are an implementation detail and should be completely hidden from the user.

Tomorrow I will start reworking this PR to implement these outlined changes. I will be moving the PR into Draft until the changes are ready.

public/reorderSiteAwards.php
Tsearo: If we are getting rid of the display order field would it be possible to change the way the awards are displayed in the re-ordering section to be more like what they show in the right column so that users can see exactly what it will look like as they re-order. This also seems like it would be easier to update as a user vs a long column list of single achievements per line, users with hundreds of awards would take a while to scroll up but with 5 per row would be 5x less scrolling and 5x more visibility. Not sure how hiding awards would work in this case though, possibly clicking on the award toggles it's visibility or moving it to a separate hidden section above the current group?

wescopeland: I agree that's where we ultimately want to be. I've considered what it would take to get us there on a first pass and, to your point, there's enough nuance with hiding awards that this becomes a much more complex change. 

I think what you're describing is ultimately where we want to land: a WYSIWYG for award reordering. But there is enough to hash out with award hiding that I'd rather make that adjustment in a subsequent PR. I think we're more likely to be successful if we get drag and drop working first, then perfect it.

public/reorderSiteAwards.php
```diff
@@ -7,97 +7,275 @@
...
+): string {
+    $arrowUp = '↑';
+    $arrowDown = '↓';
+
+    if ($arrowCount === 2) {
+        $arrowUp = '↑↑';
+        $arrowDown = '↓↓';
+    } elseif ($arrowCount === 3) {
+        $arrowUp = '↑↑↑';
+        $arrowDown = '↓↓↓';
```
Jamiras, line 73: These should have tooltips. It's unclear what the difference is between them. Maybe it would be better to just use appropriate text labels (i.e. `↑10` or `↑Top`)

Jamiras: Also, wouldn't it be easier to just pass the label in? `arrowCount` doesn't seem to be used for anything else.

public/reorderSiteAwards.php
Jamiras: This text very much no longer aligns with the implementation.

resources/js/app.ts
```diff
@@ -2,10 +2,17 @@ import Alpine from 'alpinejs';
...
+import { lazyLoadModuleOnIdFound } from './lazyLoadModuleOnIdFound';
 import { clipboard, mobileSafeTipEvents, themeChange } from './utils';
 
 // livewire_hot_reload();
 
+lazyLoadModuleOnIdFound({
+  elementId: 'reorder-site-awards-header',
+  codePath: './reorderSiteAwards',
+  moduleName: 'reorderSiteAwards',
+});
```
Jamiras, line 14: I'm assuming this is failing for me. Any changes I attempt to make generate an error similar to this:
```
Uncaught ReferenceError: reorderSiteAwards is not defined
    at HTMLButtonElement.onclick
```

public/reorderSiteAwards.php
wescopeland: I have tested several different treatments for these button labels. I am interested in which you think works best:

A
Screenshot 2023-03-24 at 8 13 52 PM

B
Screenshot 2023-03-24 at 8 15 03 PM

C
Screenshot 2023-03-24 at 8 17 20 PM

resources/js/app.ts
wescopeland: This lazy loading functionality should now be working properly and respond to changes while in dev mode.

wescopeland:
I would still like to add tests to the TypeScript code, but functionally it is in a place where I think it makes sense to comprehensively review. Therefore, I will go ahead and promote this from Draft status.

public/js/all.js
```diff
@@ -311,10 +311,12 @@ function showStatusSuccess(message) {
...
-  var status = $('#status');
-  status.addClass('failure');
-  status.html(message);
-  status.show();
+  const status = document.getElementById('status');
+  if (status) {
+    status.classList.add('failure');
+    status.innerHTML = message;
+    status.style.display = 'block';
+  }
```
wescopeland, line 319: Without this change, I was unable to actually get the `message` value to properly display in `showStatusFailure()`.

public/reorderSiteAwards.php
wescopeland: Where I ultimately landed is "Display Order" is an implementation detail and is now hidden from the user.

public/reorderSiteAwards.php
wescopeland: This text has been rewritten.

I have heard a saying that good UX is like a good joke, it doesn't have to be explained to the person for them to understand it. Ideally I would like to get to the point that this text can be removed altogether, but I'm just not sure if we're there yet.

public/reorderSiteAwards.php
wescopeland: This should now be resolved given that the input fields have been removed altogether. Initiating a drag from the dragging on the buttons is still possible, but I don't think it will cause confusion.

Bryan1150:
Could this implementation also be used for reordering achievements?

wescopeland:
> Could this implementation also be used for reordering achievements? As they both had a similar functionality

With mild adjustments, yes.

public/reorderSiteAwards.php
Jamiras: I like option C, though I'm struggling with the "up/down 50" items. The first time I clicked on the down 50, I couldn't find where the item went because it had moved down like four screens (and I was looking for it down two and a half screens or so). If we're going to keep that functionality, I'd recommend scrolling to the new location, or including numbers on the row so the user has an idea of what N+50 is.

Jamiras (approved):
Seem to be functioning, but I have some concerns about the functionality of the "up/down 50" buttons. 

Works great for small lists, but long lists are hard to rearrange. When you earn a new badge and it appears at the bottom of the list, you have to drag it up one screen at a time to get it where you want, or press the "up 50" and then go try to find it before doing the one screen at a time thing to get it exactly where you want it.

wescopeland:
I discovered while stress testing logged in as the user MaxMilyin I could not save awards due to `max_input_vars` being set to 1000 by default. I have made two adjustments to combat this issue:

- I have increased `max_input_vars` from 1000 to 5000.
- I am now encoding the award entities sent in the POST to occupy one variable of space instead of six.

I have also re-enabled autoscrolling on 50-space moves. Demo:

https://user-images.githubusercontent.com/3984985/228691988-e2ba7296-c334-4e1f-9cfd-ceabcc174074.mp4

I believe these changes are significant enough it is appropriate to request another review.

Jamiras:
Not sure what's happening, but when I press "down 50", it appears to go down 100. I have to press "up 50" twice to get it back where it started.

Jamiras:
I get the following error trying to save after hiding an item:
```
The awards.2 must be 4 comma-separated integers.
```

Jamiras:
The Top/Bottom buttons don't autoscroll to the new location.

Neither do the single up/down buttons (if you're moving the topmost visible item up or the bottommost visible item down), though that's probably less important.

wescopeland:
> The Top/Bottom buttons don't autoscroll to the new location.

This is intentional. I am happy to change it, but figured it was better for usability to keep it as-is. I will defer to your opinion on what you think is best.

wescopeland:
> Not sure what's happening, but when I press "down 50", it appears to go down 100. I have to press "up 50" twice to get it back where it started.

Good catch. I resolved this via simplification of the code for handling manual moves here https://github.com/RetroAchievements/RAWeb/pull/1439/commits/b8bfe2f57cc430a43e496ac63273f138e7af0c51#diff-4f90c4d5e8f31f2becea6e52298b9a6b89901a07afe15c4bbd09be1b6f229365. Previously there was a bunch of logic responsible for skipping over hidden rows during large moves, but after thinking this through I believe this functionality actually makes things more confusing. Deleting it also fixed this bug.

> I get the following error trying to save after hiding an item:

Good catch. This was an issue with the validation regex not accepting negative numbers and has now been resolved.

> The Top/Bottom buttons don't autoscroll to the new location.

I initially did not enable autoscroll on this because I thought it was better UX to leave it disabled. After playing with it some more I agree that it's probably better to have it enabled. This should now be resolved.

Jamiras:
The autoscroll on "Up 1" isn't working right. If I position an item at the top of the screen and press "up 1", it seems to jump down one, not up. It's still not working at all for "Down 1". Probably best to just disable it for those.

The up/down 50 and top/bottom seem to be working good now, and I can hide items.

wescopeland:
> The autoscroll on "Up 1" isn't working right. If I position an item at the top of the screen and press "up 1", it seems to jump down one, not up. It's still not working at all for "Down 1". Probably best to just disable it for those.

It should already be disabled for both. The screen shouldn't scroll at all for "Up 1" and/or "Down 1" moves. When I position an item at the top of the screen and press "Up 1" it remains fixed in place. 

Here is a recording of what I'm seeing, are you seeing something different?

https://user-images.githubusercontent.com/3984985/229363547-376ed813-5e64-4c50-a6f1-4cc3e71ededa.mp4

Jamiras:
I wasn't referring to the first/last item in the list.

I scroll down a bit so one of the badges is partially on screen, then press it's "Up" button. Notice that the list jumps to the item _below_ the one I moved and the moved item (and the item it displaced) are offscreen.

Then I move to an item at the bottom of the screen and press Down. The list doesn't move at all and the item is now offscreen.

Then I move one at the top again to illustrate the first bug.

https://user-images.githubusercontent.com/32680403/229364711-437b5ab0-725c-452e-a892-f348e17e6753.mp4

wescopeland:
> I scroll down a bit so one of the badges is partially on screen, then press it's "Up" button. Notice that the list jumps to the item _below_ the one I moved and the moved item (and the item it displaced) are offscreen.

Apologies for the misunderstanding, thank you for the clarification.

I believe I've tracked this down. It appears the unintended scroll issue was occurring because when a row near the top of the viewport was moved, the browser would automatically adjust the scroll position to accommodate the changes in the table layout. This was causing a jump in the scrollbar's current position. 

To solve this, I cache the current scroll position before moving the row, and then restore the scroll position after the row is moved. This effectively pins the scrollbar for the frame that the row move is occurring and prevents the browser from automatically adjusting the scroll position on its own.

Jamiras (approved):

Issue #1383, opened 2023-02-22
[Achievements won by / Points awarded to] others can incorrectly decrement


wescopeland:
**Describe the bug**
When a user earns achievements they've created, "Achievements won by others" and "Points awarded to others" stats don't increment. This is correct behavior.

However, when I reset these achievements, the numbers decrement.

**To Reproduce**
Earn an achievement you've created, then reset it.

**Expected behavior**
"Achievements won by others" and "Points awarded to others" should not decrement in this circumstance.

**Additional context**
Reported in #site-feedback: https://discord.com/channels/310192285306454017/1048299730670735421/1077770311853944842

PR #1447, opened 2023-03-24
add ResetPlayerAchievementAction and tests

Jamiras:
includes fix for #1383


.hooks/pre-push
```diff
@@ -7,14 +7,14 @@ if [ ${#modifiedPhpFiles} -gt 0 ]; then
     printf "\n⏳ composer lint\n"
     PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --verbose --dry-run --show-progress=none --config=.php-cs-fixer.php ${modifiedPhpFiles}
 
-    printf "\n⏳ composer test\n"
-    vendor/bin/phpunit
-
     printf "\n⏳ composer analyse\n"
-    vendor/bin/phpstan analyse --memory-limit 512M --no-progress
+    vendor/bin/phpstan analyse --memory-limit 512M
```
Jamiras, line 11: deptrac shows a progress bar, so why shouldn't this?

.hooks/pre-push
```diff
@@ -7,14 +7,14 @@ if [ ${#modifiedPhpFiles} -gt 0 ]; then
...
 
     printf "\n⏳ composer deptrac\n"
     vendor/bin/deptrac analyse
+
+    printf "\n⏳ composer test\n"
+    vendor/bin/phpunit
```
Jamiras, line 17: Moved this after the static analysis, as this is already taking a significant amount of time and fixing the static analysis is a much more iterative process

app_legacy/Helpers/database/user-account-deletion.php
```diff
@@ -80,6 +81,20 @@ function deleteOverdueUserAccounts(): void
...
+        u.SaltedPass = '',
+        u.cookie = null,
+        u.appToken = null,
+        u.APIKey = null
+        WHERE u.ID = :userId", ['userId' => $user->ID]
+    );
+
+    // reset all achievements earned by the player
+    $action = new ResetPlayerAchievementAction();
+    $action->execute($user);
```
Jamiras, line 96: This ensures contribution yields are updated correctly for any achievements the deleted user had earned.

It also updates their score to 0 and deletes any game-related badges they may have had, which are redundantly handled later.

app_legacy/Platform/Actions/ResetPlayerAchievementAction.php
```diff
@@ -0,0 +1,74 @@
...
+                         COUNT(ach.ID) AS Count, SUM(ach.Points) AS Points,
+                         SUM(ach.TrueRatio) AS TruePoints
+                  FROM (
+                     SELECT aw.AchievementID, MAX(aw.HardcoreMode) as HardcoreMode FROM
+                     Awarded aw LEFT JOIN Achievements ach ON ach.ID=aw.AchievementID
+                     WHERE aw.User = :username $clause GROUP BY aw.AchievementID
+                  ) as aw_ach
+                  LEFT JOIN Achievements ach ON ach.ID = aw_ach.AchievementID
+                  WHERE ach.Flags = " . AchievementType::OfficialCore . "
+                  GROUP BY ach.Author, ach.GameID, aw_ach.HardcoreMode";
```
Jamiras, line 30: This should probably be converted to Eloquent, but I couldn't figure out the correct syntax to make it work the way I needed. Should be easier once we flatten the Awarded table.

luchaos: Makes sense 👍 The transition to Eloquent will require a bit more (performance) testing anyways, as the ORM fetches way too much data by default. So limiting the Eloquent calls to just the fields that we require will be necessary to get the best performance out of it and not decrease it by accident.

tests/Feature/Platform/TestsPlayerAchievements.php
```diff
@@ -0,0 +1,126 @@
...
+use LegacyApp\Platform\Enums\UnlockMode;
+use LegacyApp\Platform\Models\Achievement;
+use LegacyApp\Platform\Models\PlayerAchievement;
+use LegacyApp\Site\Models\User;
+
+trait TestsPlayerAchievements
+{
+    protected function addPlayerAchievement(User $user, Achievement $achievement,
+        ?Carbon $hardcoreUnlockTime, Carbon $softcoreUnlockTime): void
+    {
```
luchaos, line 17:
```suggestion
    protected function addPlayerAchievement(
        User $user, 
        Achievement $achievement,
        ?Carbon $hardcoreUnlockTime, 
        Carbon $softcoreUnlockTime
    ): void {
```

Personally I prefer to go all with chopping down parameters when introducing new lines - not very important though.

app_legacy/Site/Models/User.php
```diff
@@ -88,6 +90,26 @@ protected function getHashIdAttribute(): int
...
     }
 
+    /**
+     * Return unlocks separated by unlock mode; both softcore and hardcore in "raw" form
+     *
+     * @return HasMany<PlayerAchievement>
+     */
+    public function playerAchievements(): HasMany
+    {
+        return $this->hasMany(PlayerAchievement::class, 'User', 'User');
```
luchaos, line 100: Heads up: this will most likely be replaced by the `ActsAsPlayer` trait as soon as the legacy merge is done in V3. I just thought if that isn't already in there but remembered that we still have two different user models in V2 😅

luchaos (approved):

PR #1448, opened 2023-03-26
Adding sort to hubs

mosquitobird11:
I thought it would be nice if we could sort hub pages by points, as this is something I have wanted myself a lot. I implemented it and abstracted the sorting into a render helper to be used by both the achievements rendering and the hubs GameAlts rendering.

I've never contributed to this repo before, so let me know if there is any process that I am doing wrong :). I love RA so I am happy to work on it some!

image


app_legacy/Helpers/database/game.php
```diff
@@ -191,8 +191,18 @@ function getGameMetadata(
     return $numAchievements;
 }
 
-function getGameAlternatives(int $gameID): array
+function getGameAlternatives(int $gameID, $sortBy = null): array
 {
+    $orderBy = match ($sortBy) {
```
mosquitobird11, line 196: This code was adapted from line 37 where we sort achievements, with the SQL changed to match the query elements for game alternatives

app_legacy/Helpers/render/game.php
```diff
@@ -190,6 +190,41 @@ function renderGameCard(int|array $game): string
...
+    $sort5 = ($sortBy == 5) ? 15 : 5;
+
+    $mark1 = ($sortBy % 10 == 1) ? "&nbsp;$sortType" : "";
+    $mark2 = ($sortBy % 10 == 2) ? "&nbsp;$sortType" : "";
+    $mark3 = ($sortBy % 10 == 3) ? "&nbsp;$sortType" : "";
+    $mark4 = ($sortBy % 10 == 4) ? "&nbsp;$sortType" : "";
+    $mark5 = ($sortBy % 10 == 5) ? "&nbsp;$sortType" : "";
+
+    echo "<a href='/game/$gameID?$flagParam&s=$sort1'>Normal$mark1</a> - ";
+    if ($isFullyFeaturedGame) {
```
mosquitobird11, line 215: This is the only logical change from when this lived in `gameInfo.php` which was done so that this same code could be reused for sorting altgames or for sorting achievements

mosquitobird11: Thanks to hubs and games having the same url structure we can just reuse the query param in the same way

app_legacy/Helpers/render/game.php
```diff
@@ -190,6 +190,41 @@ function renderGameCard(int|array $game): string
...
+    $sort4 = ($sortBy == 4) ? 14 : 4;
+    $sort5 = ($sortBy == 5) ? 15 : 5;
+
+    $mark1 = ($sortBy % 10 == 1) ? "&nbsp;$sortType" : "";
+    $mark2 = ($sortBy % 10 == 2) ? "&nbsp;$sortType" : "";
+    $mark3 = ($sortBy % 10 == 3) ? "&nbsp;$sortType" : "";
+    $mark4 = ($sortBy % 10 == 4) ? "&nbsp;$sortType" : "";
+    $mark5 = ($sortBy % 10 == 5) ? "&nbsp;$sortType" : "";
+
+    echo "<a href='/game/$gameID?$flagParam&s=$sort1'>Normal$mark1</a> - ";
```
wescopeland, line 214: In a Hubs context (and arguably an Achievements list context too), it is ambiguous and unclear what a "Normal" sort means.

EmoonX:
Wouldn't it be better to simply expand `gameList.php` to deal with hubs instead (I think this was discussed at some point)? This way we can just get rid of the hub-specific code and unite features with the ones from console lists. After all, hub pages are essentially game lists, just with a different query condition.

luchaos:
> Wouldn't it be better to simply expand `gameList.php` to deal with hubs instead (I think this was discussed at some point)? This way we can just get rid of the hub-specific code and unite features with the ones from console lists. After all, hub pages are essentially game lists, just with a different query condition.

Heads up: game hubs will be detached from the game resource, as it's just a bastardized game alternatives feature.

I wouldn't spend too much time improving the current hubs handling, some simple adjustments might do the job for now.

See for reference: https://github.com/RetroAchievements/RAWeb/blob/master/database/migrations/platform/2018_10_02_000002_create_games_table.php#L71-L95

mosquitobird11:
Thanks for the info all! 

I think I will remove "normal" sorting from hubs pages and let title sorting do what normal sorting is doing, I don't think there is much use in sorting the entire list by Title. Seems low effort enough to get some functionality before we move to using game sets

Will also fix the one phpstan error.

mosquitobird11:
Ended up keeping the "normal" sort and removing "title" sort, because the default sorting doesn't sort by title so it's weird when you reverse the sort.

app_legacy/Helpers/render/game.php
wescopeland: To help this new functionality be the best it can be, I propose we change the label for "Normal" while a user is in the context of a hub page. I am confused by the current link copy, as it does not provide a clear indication of the sorting criteria being applied to the list.

From my perspective, here is the current user journey:

- I land on a hub page and see a list sorted by points (primary) and game title (secondary).
- I click the "Normal" link. The page refreshes but nothing in the list changes.
- I click the "Normal" link again. The page refreshes, now the list inverts showing 0 to N points by Z-A game title.

To improve the user experience and mitigate confusion, I propose updating the label with a more descriptive name that reflects the sorting mechanism being applied.

Here are some possible alternatives I could think of:
* Points (A-Z)
* Points & Title
* Points, then Title
* Points, Alphabetical

wescopeland: In case we are seeing two different things, here is how I repro'd the 2 click issue mentioned earlier. I am indeed finding I have to click it twice to make any change. This is all good, just sharing so we are on the same page. I don't think this issue is critical at all.

https://user-images.githubusercontent.com/3984985/227798909-1dfb7730-86f4-4dd2-ad22-8bfe01145f1e.mp4

wescopeland: I don't have the ability to mark this comment as resolved, but it's now resolved from my POV.

mosquitobird11: I just realized something about my local setup is actually evaluating differently from production for that, even on the achievements page it thinks normal is sorted already with no query parameters. I think it must be a slight evaluation difference with a null modulus or something because I am on php 8.2

mosquitobird11: Updated the PR again though :)

mosquitobird11:
After discussing with @wescopeland , changed the sorting to "Achievements, Title" and "Points" for hubs, and corrected it so that it reverses title sort for descending sort.

app_legacy/Helpers/render/game.php
luchaos: Heads up: we'll switch to PHP 8.2 in the upcoming weeks, to Laravel 10 after that.
I have to finish a larger task first.

wescopeland:
Try `composer fix` in the root RAWeb dir to fix the linting issue. I have to do this pretty religiously as VSCode seems to be fairly uncooperative when editing PHP.

Otherwise LGTM - @luchaos and/or @Jamiras will leave a final review

luchaos:
> Try `composer fix` in the root RAWeb dir to fix the linting issue. I have to do this pretty religiously as VSCode seems to be fairly uncooperative when editing PHP.

I wonder, are those all pushed with `--no-verify`? The checks should prevent pushing without running those checks locally.

mosquitobird11:
These are pushed with --no-verify because im having issues with the test db driver, I will try to get it fixed. I did run composer fix to catch the uncaught line_endings issue though

luchaos:
> issues with the test db driver, I will try to get it fixed.

Got it! Hit us up on Discord, I'm sure we can figure it out :)

app_legacy/Helpers/database/game.php
```diff
@@ -191,8 +191,16 @@ function getGameMetadata(
     return $numAchievements;
 }
 
-function getGameAlternatives(int $gameID): array
+function getGameAlternatives(int $gameID, ?int $sortBy = null): array
 {
+    $orderBy = match ($sortBy) {
+        11 => "ORDER BY HasAchievements ASC, gd.Title DESC",
+        2 => "ORDER BY gd.TotalTruePoints, gd.ID ASC ",
+        12 => "ORDER BY gd.TotalTruePoints DESC, gd.ID DESC ",
```
Jamiras, line 199: Why wouldn't you want the secondary sort to be on title too?

public/gameInfo.php
```diff
@@ -60,7 +60,7 @@
 
 $pageTitle = "$gameTitle ($consoleName)";
 
-$relatedGames = getGameAlternatives($gameID);
+$relatedGames = getGameAlternatives($gameID, $sortBy);
```
Jamiras, line 63: You probably don't want hubs to be rearranged on the game page when changing the sort order of achievements.  This needs to be behind a `isFullyFeaturedGame` check.

app_legacy/Helpers/render/game.php
```diff
@@ -190,6 +190,45 @@ function renderGameCard(int|array $game): string
...
+        echo "<a href='/game/$gameID?$flagParam&s=$sort2'>Won By$mark2</a> - ";
+        // TODO sorting by "date won" isn't implemented yet.
+        // if(isset($user)) {
+        //    echo "<a href='/game/$gameID?$flagParam&s=$sort3'>Date Won$mark3</a> - ";
+        // }
+        echo "<a href='/game/$gameID?$flagParam&s=$sort4'>Points$mark4</a> - ";
+        echo "<a href='/game/$gameID?$flagParam&s=$sort5'>Title$mark5</a>";
+    }
+    else {
+        echo "<a href='/game/$gameID?$flagParam&s=$sort1'>Achievements, Title$mark1</a> - ";
```
Jamiras, line 225: No need to mention the secondary sort - that will just confuse users.

Though I would recommend changing is to "Has Achievements" or something similar as the implication is that it's sorted by number of achievements.

app_legacy/Helpers/database/game.php
Jamiras: Recommend swapping 2 to be DESC as I would expect that clicking on Points would immediately show me the games with most points.

app_legacy/Helpers/render/game.php
```diff
@@ -190,6 +190,45 @@ function renderGameCard(int|array $game): string
...
+        echo "<a href='/game/$gameID?$flagParam&s=$sort2'>Points$mark2</a>";
```
Jamiras, line 226: The implementation is "Retro Points", not "Points".
image

Either update the label or the filter. While "Points" would be preferred, I understand that would be harder due to not having the denormalized data (it's coming).

app_legacy/Helpers/render/game.php
mosquitobird11: What do you think of "Default" instead of "Normal" and "Has Achievements"? I have been struggling for something that looks right

app_legacy/Helpers/database/game.php
mosquitobird11: Yeah I agree! It makes the meanings different than the achievement sorting but that doesn't really matter

Jamiras: Now you need to make the arrow reflect the the sort order:
image

Up arrow implies ascending sort, but values are descending.

app_legacy/Helpers/render/game.php
Jamiras: "Normal" would at least be consistent with the game page.

wescopeland: For full disclosure, I had previously given feedback that "Normal" was ambiguous and confusing wording for this sort. If I were to give an opinion here, I think "Default" probably works better.

app_legacy/Helpers/render/game.php
```diff
@@ -190,6 +190,53 @@ function renderGameCard(int|array $game): string
...
+    $mark2 = ($sortBy % 10 == 2) ? "&nbsp;$sortType" : "";
+    $mark3 = ($sortBy % 10 == 3) ? "&nbsp;$sortType" : "";
+    $mark4 = ($sortBy % 10 == 4) ? "&nbsp;$sortType" : "";
+    $mark5 = ($sortBy % 10 == 5) ? "&nbsp;$sortType" : "";
+
+    $reverseMark1 = ($sortBy % 10 == 1) ? "&nbsp;$sortType" : "";
+    $reverseMark2 = ($sortBy % 10 == 2) ? "&nbsp;$sortType" : "";
+    $reverseMark3 = ($sortBy % 10 == 3) ? "&nbsp;$sortType" : "";
+    $reverseMark4 = ($sortBy % 10 == 4) ? "&nbsp;$sortType" : "";
+    $reverseMark5 = ($sortBy % 10 == 5) ? "&nbsp;$sortType" : "";
```
Jamiras, line 220: This logic is identical to `markX`. Therefore it's not actually reversing anything. I suspect you want to replace`$sortType` with `$sortReverseType`.

mosquitobird11: LOL thank you, updated! When I was testing I was confusing myself which way was which 💀

Jamiras (approved):

PR #1449, opened 2023-03-27
fix(gameSearch.php): sort console list at the top of the page

wescopeland:
This PR makes a small QoL adjustment to /gameSearch.php (the "Hardest Games" list). Right now, the console filters are not sorted alphabetically. This PR changes the console list to be sorted A-Z, with the exception of Hubs and Events which still remain at the end.

**BEFORE**
Screenshot 2023-03-27 at 5 04 35 PM

**AFTER**
Screenshot 2023-03-27 at 5 04 26 PM


Jamiras (approved):

Jamiras:
I wonder if it wouldn't make more sense to switch to a dropdown, like on the Most Requested Sets page.
image

wescopeland:
> I wonder if it wouldn't make more sense to switch to a dropdown, like on the Most Requested Sets page. image

One other thought I had on this -- do you think it would make sense to also remove consoles that haven't been rolled out yet? (thinking back to the `isValidConsole` thing from the home page Blade PR)

Jamiras:
> do you think it would make sense to also remove consoles that haven't been rolled out yet?

As far as I can tell, this is only used for the Hardest Games link. In that case, it definitely does make sense.

wescopeland:
> As far as I can tell, this is only used for the Hardest Games link. In that case, it definitely does make sense.

Sounds good. I will make both of these proposed changes today.

wescopeland:
Per our discussion, here is the current state of things:

Screenshot 2023-04-08 at 1 55 01 PM

Note that because this page has been repurposed from a "game search" to a "hardest games" page, I have updated the heading and SEO meta accordingly.

The dropdown contains a list of all the consoles that have undergone rollout via the `isValidConsoleId` function, in addition to the default "All Consoles" option.

luchaos:
> this page has been repurposed from a "game search" to a "hardest games" page

This kinda conflicts with the planned redirect to the general page list: 
https://github.com/RetroAchievements/RAWeb/blob/master/config/missing-page-redirector.php#L70

I guess we can adjust the redirect target to include the sorting when we get around to it.

wescopeland:
> > this page has been repurposed from a "game search" to a "hardest games" page
> 
> This kinda conflicts with the planned redirect to the general games list page: https://github.com/RetroAchievements/RAWeb/blob/master/config/missing-page-redirector.php#L70
> 
> I guess we can adjust the redirect target to include the sorting when we get around to it.

Ah, my interpretation of this is it sounds like the ultimate plan is to delete this page and redirect it to the games list, which could possibly sort hardest games by default when the "Hardest Games" link is clicked. I like that plan, thank you for providing this additional context.

Based on that context, is there any preparatory work I should do in this PR to more gracefully handle that future planned redirect?

luchaos (approved):
> Based on that context, is there any preparatory work I should do in this PR to more gracefully handle that future planned redirect?

I don't think anything has to be adjusted 👍 
Just giving more context for some preparations that I did years ago 😅 
Nothing is set in stone there btw. - adding the default sort will be fairly easy to do as soon as we decide to redirect to the new routes.

public/gameSearch.php
```diff
@@ -7,47 +7,57 @@
...
+$eventsKey = $consoleList->search('Events');
+$eventsElement = $consoleList->pull($eventsKey);
+$hubsElement = $consoleList->pull($hubsKey);
+
+$consoleList = $consoleList->sort();
 $consoleList->prepend('All Consoles', 0);
 
+// Add 'Hubs' and 'Events' back to the end of the list.
+$consoleList->put($hubsKey, $hubsElement);
+$consoleList->put($eventsKey, $eventsElement);
```
Jamiras, line 22: Do we really even need Hubs and Events in the list?

wescopeland: Looking at the functionality of these options more critically, I don't believe they add any value. The links in the navbar for these items are probably more helpful than what a user would find on this page. I will remove these options.

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1449/commits/355ab528d0f36fd02217468bfb2b7601fdfbe2e2.

Jamiras: I think you removed too much. Now they're in the middle of the list.

wescopeland: You're right, I got a bit overzealous. Should be resolved now.

PR #1450, opened 2023-03-28
Add tests for set requests

Jamiras:


app_legacy/Community/Controllers/UserGameListController.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace LegacyApp\Community\Controllers;
+
+use Illuminate\Support\Carbon;
+use LegacyApp\Http\Controller;
+use LegacyApp\Site\Models\User;
+
+class UserGameListController extends Controller
+{
+    public static function getUserSetRequestsInformation(User $user): array
```
Jamiras, line 13: Not sure if this should be on a Controller. Do you have any better suggestions?

app_legacy/Community/Models/UserGameListEntry.php
```diff
@@ -0,0 +1,66 @@
...
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use LegacyApp\Platform\Models\Game;
+use LegacyApp\Site\Models\User;
+use LegacyApp\Support\Database\Eloquent\BaseModel;
+
+class UserGameListEntry extends BaseModel
+{
+    use HasFactory;
+
+    protected $table = 'SetRequest';
```
Jamiras, line 19: I plan to replace this with `UserGameListEntries` in a future PR (taking suggestions on name). It would be the same schema, but change `User` to `UserID` and add a `Type` column (see `UserGameListType` enum above).

A first PR would copy the data to the new table and update both tables. After deploying that, a later PR would delete the `SetRequest` table.

deptrac.yaml
```diff
@@ -212,6 +212,7 @@ deptrac:
     LegacyCommunity:
       - LegacyDatabase
       - LegacyPlatform
+      - LegacySite
```
Jamiras, line 215: For the `User` model. Similar relationship exists in non-legacy folders

app_legacy/Community/Models/UserGameListEntry.php
luchaos: I had something similar in mind for V3 which already does a lot of those schema changes, so I added it as a note: https://github.com/RetroAchievements/RAWeb/blob/v3/app/Community/Models/AchievementSetRequest.php#L11-L12

The table currently only holds about 50k of rows so I can include your suggested changes there.
It will add `id` as primary for consistency, keep `User,GameID` as unique index, add `user_id`, `created_at`, and `type`.

app_legacy/Community/Controllers/UserGameListController.php
luchaos: It should not be on a controller, those ideally only should contain slim action methods. Also, we should avoid static methods whenever possible (we might want to switch to Octane at some point which does not play well with static methods and properties).

A common pattern would be to introduce `Service` classes or `Repositories` for such things, I guess moving it just an `Action` class or a global helper method would suffice for now?

app_legacy/Community/Enums/UserGameListType.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LegacyApp\Community\Enums;
+
+abstract class UserGameListType
+{
+    public const SetRequest = 1;
```
luchaos, line 9: For new Enums I'd prefer storing those as strings.
This will also allow for custom lists should we ever want to support that. The values in this enum would just be the default system-provided lists.

app_legacy/Community/Controllers/UserGameListController.php
luchaos: Oh, or put it on a Model directly.´

app_legacy/Community/Models/UserGameListEntry.php
luchaos: Went ahead and adjusted V3 migrations and model accordingly:

https://github.com/RetroAchievements/RAWeb/blob/v3/app/Community/Models/UserGameListEntry.php

https://github.com/RetroAchievements/RAWeb/blob/v3/database/migrations/community/2023_04_08_000001_update_set_requests_table.php

app_legacy/Community/Enums/UserGameListType.php
Jamiras: Not sure I understand this. You want this?
```
public const SetRequest = "SetRequest"
```

Aren't enums supposed to be integers for performance reasons? There's nothing saying we can't assign custom IDs at some fixed offset (100, 256, etc).

app_legacy/Community/Models/UserGameListEntry.php
Jamiras: I'm concerned about the migration route here. We can't really rename a table without downtime. My plan was to copy this data into the new table and maintain both while we transition the queries.

app_legacy/Community/Controllers/UserGameListController.php
Jamiras: This would make sense on a UserGameList model, but there is no list-level model. It's just a collection of UserGameListEntry models. I'll move it there for now.

app_legacy/Community/Models/UserGameListEntry.php
luchaos: The tables won't be renamed just yet, there's a different migration just for that (after v3): https://github.com/RetroAchievements/RAWeb/blob/v3/database/migrations/upcoming/2023_05_01_000001_rename_tables.php

In the (yet to be posted) v3 PR notes I'm mentioning a possible downtime of just very few minutes; to make sure no requests are lost the connect calls will be stored in a paused queue.

```
Instead of implementing sync/double-write strategies most schemas are changed in-place.
However, some tables are too large for that, like the Activity, Awarded, and LeaderboardEntry tables, which have to be synced to a new schema instead.
Activity can be ignored as it's meant to be ephemeral (keep 1-3 months of data, to be defined).
Awarded has to be de-duped anyways (hardcore/softcore) so setting up sync jobs for this table makes much more sense.
```

```
**Implications for deployment**

To make sure the database can do its job without running into bottlenecks and/or timing out active connections a planned downtime of a few minutes should be taken into account.

It could be brute-forced, too, (running migrations while the site takes requests) but might just cause it to run longer.

Running the migrations during the least active day/time of the week is desirable, around 8:00am, Monday-Thursday.

Almost all tables are migrated/updated in-place; migrations that include index/primary key changes for large tables have been excluded:
- Activity (~62m rows) -> partial sync to user_activities (about 3 months, old entries will be dropped in the future, too)
- Awarded (~60m rows) -> sync & dedupe softcore/hardcore to user_achievements
- LeaderboardEntry (~1.5m rows) -> sync to leaderboard_entries

The Comment table holds 1.8m rows but can be altered in-place because there are no major index changes, it already has an integer ID column.

Taking the application down is done using Laravel's `php artisan down` command; affects the site, connect/integration API and Web API.

This downtime should be planned ahead and well communication. The migration will be done during a day and time where we expect the least amount of traffic going by our traffic monitoring tools.
```

luchaos: I'd personally just take the minor downtime at this point and get the schema migration over with. Yet, maybe we can just run the migrations while keeping the site up. I'll test this on staging first.

app_legacy/Community/Enums/UserGameListType.php
luchaos: Performance is not an issue when there's a proper index on the column.

Integers are harder to reason about and not great to maintain, as with most of the existing enums.

I'd go with snake-case (slugged) for the enum values:
```php
public const SetRequest = 'set-request';
public const WantToPlay = 'want-to-play';
public const WantToDev = 'want-to-dev';
```

app_legacy/Community/Controllers/UserGameListController.php
luchaos: Sgtm!

app_legacy/Community/Enums/UserGameListType.php
Jamiras: After spending some time with Google, I can see how this will have minimal impact in the database. The individual values won't be on every row. Instead, each row will point to the same value in the index, effectively using the pointer as a numeric enum.

Logic in the code will still have to do string comparisons, but should be much more limited than what would be required to filter the database records.

luchaos: While i'd still use integers for primary keys for consistency's sake, there is nothing preventing to use strings (like uuids, ulids) as primary key to be indexed.
It used to be problematic in the past with some db engines, these days it's just about index size iirc. MySQL got a lot better since then.

app_legacy/Community/Models/UserGameListEntry.php
Jamiras: > The tables won't be renamed just yet, there's a different migration just for that (after v3) https://github.com/RetroAchievements/RAWeb/blob/v3/database/migrations/upcoming/2023_05_01_000001_rename_tables.php

Renaming all of these tables at once seems dangerous given how many raw queries are still in the system. But it would eliminate redundant validation if we took a more iterative approach.

In any case, that's probably a discussion for the upcoming PR.

luchaos: Note, that this migration in particular is not yet part of v3. The goal was to rewrite all the existing schemas so the current code and its queries do not have to be adjusted, rewritten and tested. That'd be an even larger scope than it already is.
The first step is to get those tables and column changes in there with the least amount of disruption, then we can iteratively populate those and change query by query.

luchaos (approved):

PR #1452, opened 2023-03-30
feat: add api docs landing page

wescopeland:
This PR adds a URL to [the JavaScript API VitePress site](https://retroachievements-api-js.vercel.app) to the site footer.

I have chatted with some folks who are completely unaware that RA has an official JS/TS API wrapper.

I know there is a plan to build a similar wrapper written in PHP. At the time that releases, perhaps this link could change from "JavaScript API" to "API", letting the user decide which flavor they'd like. For the time being though, this seems like an optimal solution to get the SEO juice going for search queries regarding the RA API.

Screenshot 2023-03-30 at 5 21 50 PM


luchaos:
I like the idea, I'd put it on a dedicated page where we can gather all the mentioned future clients as seen here: https://dev.retroachievements.org/docs/api

wescopeland:
> I like the idea! I'd put it on a dedicated page where we can gather all the mentioned future clients

I am happy to do that. Before I go down this path, I want to be sure my gut feeling is the right approach:

In _RouteServiceProvider.php_, add:

```php
Route::get('docs/api', [ContentController::class, 'apiDocs'])->name('docs.api');
```

Update _ContentController.php_ to include something like:

```php
public function apiDocs(): View
{
    return view('api.docs');
}
```

Add an _api.docs.blade.php_ file in _resources/views_. Does this sound right?

luchaos:
> > I like the idea! I'd put it on a dedicated page where we can gather all the mentioned future clients
> 
> I am happy to do that. Before I go down this path, I want to be sure my gut feeling is the right approach:
> 
> In _RouteServiceProvider.php_, add:
> 
> ```
> Route::get('docs/api', [ContentController::class, 'docsApi'])->name('docs.api');
> ```
> 
> Update _ContentController.php_ to include something like:
> 
> ```
> public function docsApi(): View
> {
>     return view('docs.api');
> }
> ```
> 
> Add a _docs.api.blade.php_ file in _resources/views_. Does this sound right?

All of that already exists, just has to be uncommented and the blade view put into place :)

https://github.com/RetroAchievements/RAWeb/blob/master/app/Platform/Controllers/ApiDocsController.php#L14

https://github.com/RetroAchievements/RAWeb/blob/master/app/Platform/RouteServiceProvider.php#L59

wescopeland:
Ah I somehow missed all of this. Good to know I was on the right track though 😆

luchaos:
Here's what used to be in `views/platform/docs/api.blade.php` - might need some adjustments:

```php
@extends('layouts.app', [
    'pageTitle' => __('RetroAchievements API'),
])

@section('header')
    <x-page-header>
        <x-slot name="title"><h2>RetroAchievements API</h2></x-slot>
        <x-slot name="subTitle">
            <p class="lead">
                Resources for Connect & Web APIs.<br>
                Make sure that your application uses the latest stable version that is currently available.
            </p>
        </x-slot>
    </x-page-header>
@endsection

@section('main')
    <div class="md:grid grid-flow-col gap-4">
        <div class="lg:col-6">
            <x-section>
                <x-section-header>
                    <x-slot name="title"><h3>Web API</h3></x-slot>
                </x-section-header>
                <p>
                    You can find your personal API token to access public data in your
                    <a href="{{ route('settings') }}">settings</a>.<br>
                    This is a beta offering and only supports read-only access to data that can already be found on the site.<br>
                    No personal/user data can be accessed beyond what is already publicly accessible for
                    every user (username, avatar, motto and activity).
                </p>
                <x-link class="btn btn-primary" link="https://github.com/retroachievements/web-api-client-php" target="_blank">PHP Client</x-link>
            </x-section>
        </div>
        <div class="lg:col-6">
            <x-section>
                <x-section-header>
                    <x-slot name="title"><h3>Connect API</h3></x-slot>
                </x-section-header>
                <p>
                    RPC APIs to be consumed by emulators, toolkits and other third party applications. Used by various clients.
                </p>
                <x-link class="btn btn-primary" :link="asset('docs/api/connect/index.html')" target="_blank">{{ __('Documentation') }}</x-link>
            </x-section>
        </div>
    </div>
@endsection
```

wescopeland:
@luchaos I have added the /docs/api page. Currently, this has some copy about finding an API key in the settings & only exposing data that is publicly accessible. Currently there is only a link to the JavaScript API docs.

Here is a high-level of what things look like:

https://user-images.githubusercontent.com/3984985/230741865-3bd1a99a-7503-4634-8dcc-5681bc3d4f7a.mp4

luchaos:
> @luchaos I have added the /docs/api page. Currently, this has some copy about finding an API key in the settings & only exposing data that is publicly accessible. Currently there is only a link to the JavaScript API docs.
> 
> 
> 
> Here is a high-level of what things look like:
> 
> 
> 
> https://user-images.githubusercontent.com/3984985/230741865-3bd1a99a-7503-4634-8dcc-5681bc3d4f7a.mp4

I think that's great! 👍

Minor nit: can we drop the "official"? If it isn't official it shouldn't be there imo. Or explicitly marked non-official if we link to other resources.

wescopeland:
> Minor nit: can we drop the "official"? If it isn't official it shouldn't be there imo. Or explicitly marked non-official if we link to other resources.

Sure thing! Should be addressed in the latest commit.

resources/views/components/menu/footer.blade.php
```diff
@@ -21,13 +21,11 @@
...
-            // [
-            //     'active' => Route::is('docs.api'),
-            //     'label' => __('API'),
-            //     'url' => url('APIDemo.php'),
-            //     // 'url' => route('docs.api'),
-            //     'visible' => true,
-            // ],
+            [
+                'label' => __('API'),
+                'url' => url('docs/api'),
```
luchaos, line 26: This can be `'url' => route('docs.api'),` now because the named route is actually usable.
The `active` check could stay, too - not sure if it actually does something atm 🤔

wescopeland: Makes sense. I've made these adjustments. The `active` check doesn't seem to be doing anything presently unless I'm mistaken.

luchaos (approved):

app/Platform/RouteServiceProvider.php
```diff
@@ -56,7 +57,7 @@ protected function mapWebRoutes(): void
             // Route::get('create', CreateController::class)->name('create');
             // Route::resource('developers', DeveloperController::class)->only('index');
 
-            // Route::get('docs/api', [ApiDocsController::class, 'index'])->name('docs.api');
+            Route::get('docs/api', [ApiDocsController::class, 'index'])->name('docs.api');
```
Jamiras, line 60: Is there something I need to run to refresh this? I get a 403 Forbidden error when I click on the API link in the footer.
image

I've already manually run the post-merge commands.
```
php composer.phar install
npm install
npm run build
```

wescopeland: With Sail it automagically worked, however I am unsure about other setups. @luchaos do you have any ideas?

luchaos: There might be a conflict with the existing `public/docs` symlink - probably something that has to be fixed in `public/.htaccess` for Apache? 🤔

If that turns out to be a problem on staging, too, then we should probably remove the symlink for now to make sure that URL is routed to the framework.

Jamiras: Looks like a broken symlink:
```
Symbolic link not allowed or link target not accessible: /home/brian/source/RAWeb/public/docs
```
```
public/docs -> ../storage/app/static/docs/
```
```
storage/app/static/docs -> ../../../docs/dist/
```
There is no `docs/dist` at the root level.

Running `php artisan ra:storage:link --relative` as mentioned in the README.md did not fix the symlink chain.

Removing the public/docs symlink appears to fix the issue.

Running `php artisan ra:storage:link --relative` again restores the link, thus breaking it again.

See https://github.com/RetroAchievements/RAWeb/blob/42023be338b7251f05bc272bfce9ee3a01e20a9f/app/Support/Filesystem/LinkStorage.php#L38

Jamiras: > If that turns out to be a problem with the server's nginx setup, too, then we should probably remove the symlink for now

It doesn't look like there's a docs symlink in production:
```
jamiras@retroachievements:/srv/web/current$ ls -l public/d*
-rw-r--r-- 1 deploy-web www-data  5227 Mar 12 18:23 public/developerstats.php
-rw-r--r-- 1 deploy-web www-data 12619 Mar 12 18:23 public/dorequest.php
-rw-r--r-- 1 deploy-web www-data  1174 Mar 12 18:23 public/doupload.php
-rw-r--r-- 1 deploy-web www-data  3251 Mar 12 18:23 public/download.php
```
```
$ ls -l public/d*
-rw-rw-r-- 1 brian brian  5227 Feb  8 14:34 public/developerstats.php
lrwxrwxrwx 1 brian brian    27 Apr  8 20:37 public/docs -> ../storage/app/static/docs/
-rw-rw-r-- 1 brian brian 12419 Mar 24 10:14 public/dorequest.php
-rw-rw-r-- 1 brian brian  1174 Oct  2  2022 public/doupload.php
-rw-rw-r-- 1 brian brian  3251 Mar  5 10:19 public/download.php
```

Jamiras: I have a workaround, and it doesn't seem like the server should be affected (we'll find out shortly). I'm going to proceed with this as-is and the symlink can be fixed in a future PR.

PR #1455, opened 2023-04-02
ignore demoted unlocks for best days query

Jamiras:


wescopeland (approved):
I've tested this and verified it works as intended.

PR #1456, opened 2023-04-07
fix(active-players): hide untracked users

wescopeland:
This PR fixes an issue where untracked players are still ordered by their would-be rank in the Active Players list on the home page.


app_legacy/Helpers/database/user-activity.php
```diff
@@ -506,8 +506,9 @@ function getLatestRichPresenceUpdates(): array
               LEFT JOIN GameData AS gd ON gd.ID = ua.LastGameID
               LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
               WHERE ua.RichPresenceMsgDate > TIMESTAMPADD( MINUTE, -$recentMinutes, NOW() )
-                AND ua.LastGameID !=0
+                AND ua.LastGameID != 0
                 AND ua.Permissions >= $permissionsCutoff
+                AND NOT ua.Untracked
```
Jamiras, line 511: I feel that untracked users should still appear in the active players list. Untracking just removes them from the leaderboards, they still appear in recent activity for individual games.

Could you wrap RAPoints and RASoftcorePoints such that they return 0 for untracked users?

wescopeland: Good idea, I've removed the `AND NOT` clause and wrapped the `RAPoints` and `RASoftcorePoints` in `IF()`s such that if the user is untracked, these values will return 0.

app_legacy/Helpers/database/user-activity.php
```diff
@@ -500,15 +500,15 @@ function getLatestRichPresenceUpdates(): array
...
+                     ua.RichPresenceMsg, gd.ID AS GameID, gd.Title AS GameTitle, gd.ImageIcon AS GameIcon, c.Name AS ConsoleName
               FROM UserAccounts AS ua
               LEFT JOIN GameData AS gd ON gd.ID = ua.LastGameID
               LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
               WHERE ua.RichPresenceMsgDate > TIMESTAMPADD( MINUTE, -$recentMinutes, NOW() )
-                AND ua.LastGameID !=0
+                AND ua.LastGameID != 0
                 AND ua.Permissions >= $permissionsCutoff
-              ORDER BY ua.RAPoints DESC";
+              ORDER BY RAPoints DESC";
```
Jamiras, line 511: I think we need to modify the sort here. The current sort just prioritizes hardcore users, then everyone else is just kind of randomly at the bottom.

At a minimum, there should probably be a secondary sort on ua.RASoftcorePoints so the untracked users don't appear in the middle of the softcore user list. I don't know if we need an explicit third-level sort for the untracked users.

wescopeland: To be sure I am completely following, do you think a change like:

```sql
ORDER BY RAPoints DESC, RASoftcorePoints DESC
```

would satisfy this and make the sorting more meaningful? 

I am not sure if we need an explicit third-level sort for the untracked users because the `IF(ua.Untracked, 0, ua.RAPoints)` and `IF(ua.Untracked, 0, ua.RASoftcorePoints) `conditions in the query already handle untracked users by setting their RAPoints and RASoftcorePoints to 0. 

This means the ORDER BY in the snippet above will already place untracked users at the bottom of the list, causing them to naturally appear after all the tracked users regardless of whether they are hardcore or softcore players.

Jamiras: Yes to `ORDER BY RAPoints DESC, RASoftcorePoints DESC`. That would guarantee a separation between softcore players and untracked users, _and_ prioritize softcore players with more points in a similar fashion to the way the main sort prioritizes hardcore players with more points.

My thought is that all untracked users would have 0/0 points, so they would just be at the end of the list with no guaranteed sort. I don't think it's a big deal, just noting the behavior.

wescopeland: I understand; I've opted to do:

```sql
ORDER BY RAPoints DESC, RASoftcorePoints DESC, ua.User ASC
```

Now those users floating around the bottom should at least be alphabetically sorted A-Z by their usernames, giving us a bit more determinism in the ultimate ordering of these entities. This change is reflected in my latest commit.

Jamiras (approved):

PR #1457, opened 2023-04-08
Remove all RSS feeds except news

luchaos:


wescopeland (approved):
I've tested this locally and verified the changes are working. 

The removed feeds are returning 404s and not redirecting anywhere. I am not sure if they should explicitly inform the consumer the feed has been removed. It probably doesn't make a big difference either way.

luchaos:
> I've tested this locally and verified the changes are working.
> 
> The removed feeds are returning 404s and not redirecting anywhere. I am not sure if they should explicitly inform the consumer the feed has been removed. It probably doesn't make a big difference either way.

They return a hard 404 on the web server level and don't even hit the application.
As they were abused by bots quite a bit I don't think redirects are needed, they might even come back some day.

Jamiras (approved):

PR #1460, opened 2023-04-08
feat(game): improve leaderboards list readability

wescopeland:
This PR refactors the leaderboard list to use stacked divs instead of a table layout. Each alternate div is then given a different background color.

**Why?**
The current leaderboards list uses a table layout where each row is given an alternate background color. This is disorienting, as each single leaderboard entry in the list has two rows. When taking a cursory glance at the leaderboards list, it can be difficult to immediately tell where one entry ends and another begins.

**BEFORE**
Screenshot 2023-04-08 at 7 28 37 PM

**AFTER**
Screenshot 2023-04-08 at 7 28 27 PM

https://user-images.githubusercontent.com/3984985/230800134-5863f13c-da0d-4b6f-ac15-ccae731a0f35.mp4


wescopeland:
On second thought, I am not entirely sold on this change.

wescopeland:
Reopening after findings from https://discord.com/channels/310192285306454017/511718348178849792/1094431735741157467. This is a more popular proposed change than I initially realized.

Jamiras (approved):

PR #1465, opened 2023-04-11
feat: add new shortcode renderer for non-forum post pages

wescopeland:
This PR creates a new `ExternalShortcode` class for rendering shortcodes specifically on non-forum pages. This renderer has been applied to the Forum Activity widget, the Forum Post History page, and DM content sent in emails.

**Why?**
Currently, shortcodes are leaking on non-post pages. I believe this even includes [spoiler] tags, which render as plain text.

**How?**
`ExternalShortcode` renders thin versions of the content produced by each shortcode. For example, rather than rendering a game avatar, a link to the game like "Game Name (Console Name)" is rendered.

**BEFORE**
Screenshot 2023-04-11 at 7 09 17 PM
  
Screenshot 2023-04-11 at 7 09 22 PM

Screenshot 2023-04-11 at 7 08 56 PM

**AFTER**
Screenshot 2023-04-11 at 7 08 24 PM

Screenshot 2023-04-11 at 5 09 52 PM

Screenshot 2023-04-11 at 7 09 03 PM

**Additional thoughts**
1. If a shortcode is truncated in the content like "Hey this is my content [url=https://goog...", the truncated shortcode will be removed from the string altogether.
2. I deliberately did not add this to post preview content in viewforum.php. I found this just made things look even weirder. We probably need a different kind of visual treatment here, and I will tackle that in a different PR.
3. Preferably, please do not release this with 2.12.0. I think this should be left to bake on our local envs and in stage for a little while.


app/Support/Shortcode/ExternalShortcode.php
```diff
@@ -0,0 +1,152 @@
...
+        $achievementData = Cache::store('array')->rememberForever('achievement:' . $id . ':card-data', fn () => GetAchievementData($id));
+
+        if (empty($achievementData)) {
+            return "<a href='/achievement/$id'>Achievement</a>";
+        }
+
+        $achievementName = $achievementData['Title'];
+        $achievementPoints = $achievementData['Points'];
+
+        return "<a href='/achievement/$id'>$achievementName ($achievementPoints)</a>";
```
Jamiras, line 122: I can't help but notice that a large portion of this file is duplicated from Shortcode.php. Would it be possible to move the bits that need to be different into functions and override them in a subclass?

Jamiras (commented):
It would be a good idea to expand on the tests in tests/Unit/ShortcodeTest.php and create and validate the expected output for both the basic parser and this external parser as a hint that there's two parsers that need to be maintained any time Shortcode changes are made in the future.

luchaos:
As mentioned in Discord, personally, I'm not sure that those tags even have to be rendered in previews? Wouldn't it be cleaner to just remove all tags for the short version?

Maintaining both does not sound convenient.

wescopeland:
Not a problem at all, I close will this PR and explore stripping the tags from the previews as a separate change.

PR #1473, opened 2023-04-15
feat(game): only show expiration timeago copy to jr devs and up

wescopeland:
This PR:

* Sets the game page's expiration timeago copy to visible only to Junior Developers and up.
* Slightly refactors the code to generate the copy for that area of the page.

This change is motivated by https://discord.com/channels/310192285306454017/310195377993416714/1096534286456340572.

Specifically, this is the area of the page affected:
Screenshot 2023-04-15 at 7 51 23 AM

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -1146,16 +1146,16 @@ sanitize_outputs(
                         }
 
                         if ($claimExpiration) {
-                            $claimFormattedDate = $claimExpiration->format('d M Y, H:i');
                             $isAlreadyExpired = Carbon::parse($claimExpiration)->isPast();
+                            $isAlreadyExpiredCopy = $isAlreadyExpired ? "Expired" : "Expires";
 
-                            echo "<p>";
-                            if ($isAlreadyExpired) {
-                                echo "Expired on: $claimFormattedDate (" . $claimExpiration->diffForHumans() . ")";
-                            } else {
-                                echo "Expires on: $claimFormattedDate (" . $claimExpiration->diffForHumans() . ")";
-                            }
-                            echo "</p>";
+                            $claimFormattedDate = $claimExpiration->format('d M Y, H:i');
+                            $claimTimeAgoDate = $permissions >= Permissions::JuniorDeveloper 
+                                ? "(" . $claimExpiration->diffForHumans() . ")" 
+                                : null;
+
+                            // "Expires on: 12 Jun 2023, 01:28 (1 month from now)"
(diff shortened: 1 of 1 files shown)


public/gameInfo.php
```diff
@@ -1146,16 +1146,16 @@ function ResetProgress() {
...
-                            if ($isAlreadyExpired) {
-                                echo "Expired on: $claimFormattedDate (" . $claimExpiration->diffForHumans() . ")";
-                            } else {
-                                echo "Expires on: $claimFormattedDate (" . $claimExpiration->diffForHumans() . ")";
-                            }
-                            echo "</p>";
+                            $claimFormattedDate = $claimExpiration->format('d M Y, H:i');
+                            $claimTimeAgoDate = $permissions >= Permissions::JuniorDeveloper
+                                ? "(" . $claimExpiration->diffForHumans() . ")"
+                                : null;
```
Jamiras, line 1155: Should the false result be empty string instead of null?

public/gameInfo.php
```diff
@@ -1146,16 +1146,16 @@ function ResetProgress() {
                         }
 
                         if ($claimExpiration) {
-                            $claimFormattedDate = $claimExpiration->format('d M Y, H:i');
                             $isAlreadyExpired = Carbon::parse($claimExpiration)->isPast();
+                            $isAlreadyExpiredCopy = $isAlreadyExpired ? "Expired" : "Expires";
```
Jamiras, line 1150: I'd just merge the two lines. `$isAlreadyExpiredCopy` is confusing.
```
$isAlreadyExpired = Carbon::parse($claimExpiration)->isPast() ? "Expired" : "Expires";
```

public/gameInfo.php
wescopeland: `null` seems to function correctly. Stylistically in PHP does this make more sense as an empty string?

public/gameInfo.php
wescopeland: Good idea, resolved in https://github.com/RetroAchievements/RAWeb/pull/1473/commits/4c9b5f8c299a109339a54e213fed620bc2a319b7

public/gameInfo.php
Jamiras: It's being used as a string below, which means it has to be cast to a string, and `(string)null` is empty string, so why not do it yourself?

I know php is pretty lenient, but a ternary operator should return the same type of variable from both paths.

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1473/commits/45bb321a9dea191606bde2485ce23fa2380d07ee

Jamiras (approved):

PR #1478, opened 2023-04-15
add bizhawk and duckstation to downloads page

Jamiras:
image

### DIFF
--- a/storage/app/releases.dist.php
+++ b/storage/app/releases.dist.php
@@ -6,7 +6,7 @@ use LegacyApp\Platform\Enums\Emulators;
 return [
     'integration' => [
         'minimum_version' => '1.0.4',
-        'latest_version' => '1.1.0',
+        'latest_version' => '1.2.0',
         'latest_version_url' => 'bin/RA_Integration.dll',
         'latest_version_url_x64' => 'bin/RA_Integration-x64.dll',
     ],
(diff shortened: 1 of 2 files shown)


storage/app/releases.dist.php
```diff
@@ -130,9 +168,19 @@
                 21, // PlayStation 2
             ],
         ],
+        Emulators::DuckStation => [
+            'name' => 'DuckStation',
+            'handle' => 'DuckStation',
+            'active' => true,
+            'link' => 'https://www.duckstation.org/wiki/Main_Page',
```
wescopeland, line 175: This link is correct, but seems to be returning a hard 404 from DuckStation's web server.

Jamiras: I noticed that too, but it's the link right off their page, and the only link available. The same thing happens if you take the `Main_Page` off - it can't find an index page either.

image

wescopeland: Yeah, I double-checked this as well to be sure.

Taking a look at archive.org, it seems this link has been broken for at least over a month:
https://web.archive.org/web/20230315020146/https://www.duckstation.org/wiki/Main_Page

Feels like a not great experience for us to direct users to a 404 page.

LiquifiedSnow: https://github.com/stenzek/duckstation/wiki There's the github wiki for it. At least it's something.

PR #1479, opened 2023-04-15
Update layout on download page

Jamiras:
With #1478, 4/12 emulators listed on the download page are not managed by RetroAchievements.

This PR includes several changes to better integrate the external emulators.

1) External download links have been removed from the description field and appear on the right side where the internal download links would be displayed.
2) Documentation link has also been moved to the right side.
3) The GitHub footer text has been removed. Links to individual GitHub repositories also appear on the right side.
4) External emulators no longer indicate they're a fork of themselves (repeated title eliminated)
5) The list is now sorted alphabetically

image


app_legacy/Platform/Enums/Emulators.php
```diff
@@ -35,14 +35,16 @@ abstract class Emulators
...
-
     // Third Party Standalone
 
     public const RetroArch = 'retroarch';
 
     public const PCSX2 = 'pcsx2';
 
+    public const DuckStation = 'duckstation';
+
+    public const Bizhawk = 'bizhawk';
```
Jamiras, line 46: This includes the changes in #1478. This will disappear after that is merged.

public/download.php
```diff
@@ -9,64 +14,71 @@
...
-                            sanitize_outputs($system);
-                            ?>
-                            <li>- <?= $system ?></li>
-                        <?php endforeach ?>
-                        </ul>
+            </h2>
+            <div class="flex flex-col lg:flex-row justify-between items-start">
+                <div class="mb-3">
+                    <?php if ($emulator['description'] ?? false): ?>
+                        <div class="mb-2"><?= nl2br($emulator['description']) ?></div>
```
Jamiras, line 29: For some reason, moving the description inside the left column (instead of across the top of both column) breaks the horizontal fill on the bullet list. Any help to restore that would be appreciated.

Before:
image

After:
image

public/download.php
```diff
@@ -9,64 +14,71 @@
...
 
-        <h2>Downloads</h2>
-
-        <p class="embedded mb-5">
-            <b>Upgrade:</b> If you were brought to this page because the emulator told you
-            that a new version is available, download the new version and
-            extract it over the existing folder. This way you won't lose
-            any save files that you may have.
-        </p>
+        <h2>Emulators supporting RetroAchievements</h2>
```
Jamiras, line 17: Not sure this is a good title for the page. The menu item still says "Download.

public/download.php
```diff
@@ -9,64 +14,71 @@
...
-                        <?php foreach ($emulator['systems'] as $system): ?>
-                            <?php
-                            sanitize_outputs($system);
-                            ?>
-                            <li>- <?= $system ?></li>
-                        <?php endforeach ?>
-                        </ul>
+            </h2>
+            <div class="flex flex-col lg:flex-row justify-between items-start">
+                <div class="mb-3">
```
wescopeland, line 27:
```suggestion
                <div class="mb-3 w-full">
```

I suspect this will fix the horizontal fill on the bullet list.

public/download.php
wescopeland: "Download an Emulator" is the only alternative I can come up with, but it loses the context that these are special emulators that support RetroAchievements. That may be implied given this is on the RA website.

This page probably doesn't matter as much for mobile, but "Emulators supporting RetroAchievements" is going to wrap to two lines on most phones, whereas "Download an Emulator" doesn't, at least for an iPhone X. "Download an Emulator" also aligns with the menu item copy.

wescopeland, line 17:
```suggestion
        <h2 class="mb-6">Emulators supporting RetroAchievements</h2>
```

nit: My eyes immediately jumped to how close the H1 is to the first emulator item in the list. I think this is due to the two elements having the same font size. This change is subtle, but gives just a pinch of breathing room.

**BEFORE**
Screenshot 2023-04-15 at 5 56 31 PM

**AFTER**
Screenshot 2023-04-15 at 5 56 22 PM

public/download.php
```diff
@@ -9,64 +14,71 @@
...
-                        <ul style="column-count: 3">
-                        <?php foreach ($emulator['systems'] as $system): ?>
-                            <?php
-                            sanitize_outputs($system);
-                            ?>
-                            <li>- <?= $system ?></li>
-                        <?php endforeach ?>
-                        </ul>
+            </h2>
+            <div class="flex flex-col lg:flex-row justify-between items-start">
```
wescopeland, line 26:
```suggestion
            <div class="flex flex-col lg:flex-row justify-between items-start mb-6">
```

When the layout is MD or less and in `flex-col` mode, the spacing between each emulator item seems to be fairly tight. This change adds a bit of gap between the elements to improve readability.

**BEFORE**
Screenshot 2023-04-15 at 6 04 47 PM

**AFTER**
Screenshot 2023-04-15 at 6 05 55 PM

luchaos:
If you want more horizontal space to make it two columns on desktop, you can add this line in at the top:
`view()->share('wide', true);`

PR #1482, opened 2023-04-16
feat: show outdated emulator notice if RP contains 'Unknown macro'

wescopeland:
This PR alters RP messages on game pages and the active players widget which contain "Unknown macro" anywhere within them. The idea is to filter out RP messages that are broken by outdated emulator use.

**BEFORE**
Screenshot 2023-04-16 at 3 03 55 PM

Screenshot 2023-04-16 at 3 16 25 PM

---

**AFTER**
Screenshot 2023-04-16 at 3 06 23 PM

Screenshot 2023-04-16 at 3 15 47 PM


app_legacy/Helpers/render/game.php
```diff
@@ -385,6 +385,18 @@ function RenderRecentGamePlayers(array $recentPlayerData): void
...
+
+        // Check if $activity contains a message about an "Unknown macro", and
+        // if so, strip the RP and replace it with an outdated emulator warning.
+        if (mb_strpos($activity, 'Unknown macro') !== false) {
+            $activity = <<<HTML
+                <div class="flex items-center gap-x-1" title="Outdated emulator detected!">
+                    <span>⚠️</span>
+                    <span>Playing $gameTitle</span>
+                </div>
+            HTML;
```
wescopeland, line 397: I am not sure if this should also say "Playing $gameTitle with an outdated emulator" just like the homepage active players widget, or if just using a tooltip around the whole div is good enough.

Here's what it looks like with the additional copy, kind of gross with the emoji floating top left imo, but I could be being too nitpicky:

Screenshot 2023-04-16 at 3 26 33 PM

Jamiras: You don't even need "with an outdated emulator" to make it wrap, though mine seems to be wrapping separately from the icon.
image

app_legacy/Helpers/render/game.php
```diff
@@ -385,6 +385,18 @@ function RenderRecentGamePlayers(array $recentPlayerData): void
             $userName,
             $activity
         );
+
+        // Check if $activity contains a message about an "Unknown macro", and
+        // if so, strip the RP and replace it with an outdated emulator warning.
```
Jamiras, line 390: Note that 'Unknown macro' could also indicate an actual problem with the script, though in that case all users should be affected.

wescopeland: Noted, though this does make me wonder if we should even explicitly state the user is using an outdated emulator, as the message could be erroneous.

Two potential ideas that may be better:

1. Show "⚠️ Playing $gameTitle", and wrap this content in a <span> with a `title` attribute saying there was a problem executing the RP script, likely caused by an outdated emulator.

2. Show "⚠️ Playing $gameTitle (There was a problem generating rich presence)"

Or maybe a mix of these two approaches. What do you think?

Jamiras: What about just putting the broken string in the tooltip?

 "⚠️ Playing $gameTitle" with "🚩1-1 🚶[Unknown macro]Number(0xh4ed) 💰[Unknown macro]Number(0xh62b)" in the tooltip?

That way the bad text doesn't affect the layout, but it's still visible for diagnosing.

wescopeland: I think that's a great idea. I've made this change in https://github.com/RetroAchievements/RAWeb/pull/1482/commits/08d46fee9492458cb9d382dcaae9702313eaeae4.

app_legacy/Helpers/render/game.php
wescopeland: I've addressed the wrapping issue in https://github.com/RetroAchievements/RAWeb/pull/1482/commits/408191873ff019f06cab20525699f1facdf54c3b.

Jamiras (approved):

Issue #1333, opened 2023-01-27
Retro Points no longer calculate unearned achievements as if they had 1 earner.


redwizard42:
Looks like one of the recent retro points calculation fixes reverted to old behavior, where unearned achievements are calculated as 0 retro points, rather than as having 1 earner.

wescopeland:
This will be resolved by https://github.com/RetroAchievements/RAWeb/pull/1484.

PR #1484, opened 2023-04-16
fix: calculate ratio points correctly when there are 0 hardcore earners

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1333.

There is currently an issue when ratio points are recalculated for games. When an achievement has zero hardcore earners, the white points for an achievement do not update, and if white points is already at 0 it remains fixed to 0. My understanding is these achievements should always be handled as if there is 1 earner. That is currently not the case.

**Quick Repro**
* Visit https://retroachievements.org/game/15020, which is a newer set no one has yet to master. 
* Force a true ratio recalc by pressing the button in the devbox.
* Observe that achievements with no hardcore earners have their white points still set to 0.

**Root Cause**
The original query fails to include achievements with 0 hardcore earners in the list, thus the `while()` loop inside the `run()` function never iterates over these achievements.

**Solution**
I have resolved the issue by slightly modifying the query. The `SUM()` function now includes a `CASE()` statement to handle NULL values resulting from the `LEFT JOIN` with the `Awarded` table for achievements with 0 hardcore earners.

The `CASE` statement also ensures untracked users are correctly excluded from the calculation.

By adding the conditional logic inside the `SUM()` function, the query now includes achievements with 0 hardcore earners in the list, and ratio points for these achievements can recalculate correctly.


app_legacy/Platform/Actions/UpdateGameWeightedPoints.php
```diff
@@ -10,11 +10,11 @@ class UpdateGameWeightedPoints
 {
     public function run(int $gameId): bool
     {
-        $query = "SELECT ach.ID, ach.Points, SUM(aw.HardcoreMode) AS NumAchieved
+        $query = "SELECT ach.ID, ach.Points, SUM(CASE WHEN ua.Untracked IS NULL OR NOT ua.Untracked THEN aw.HardcoreMode ELSE 0 END) AS NumAchieved
```
Jamiras, line 13: Untracked should [not be nullable](https://github.com/RetroAchievements/RAWeb/blob/47f3e56f4933b89f5f22e990d9e0e21e960339a7/database/migrations/legacy/2012_10_03_133633_create_base_tables.php#L714).
```
show create table UserAccounts
```
```
  `Untracked` tinyint(1) NOT NULL DEFAULT '0',
```

wescopeland: Resolved by https://github.com/RetroAchievements/RAWeb/pull/1484/commits/6297ca3920cb5a7261cf249a0fbe9d824447171b

Jamiras (approved):

PR #1487, opened 2023-04-18
feat(game): add hover color fade for locked achievements

wescopeland:
This PR changes locked achievement badges on game pages and achievement info pages to fade to the fully-colorized version on hover.

Instead of loading the `_lock` version of the badge, the colorized version is loaded and the grayscale, blur, and contrast adjustments are applied via CSS. They are then transitioned away on hover.

https://user-images.githubusercontent.com/3984985/232656439-8517fa3c-6e0e-47b3-a812-e6c5f25c2761.mp4


wescopeland:
There is a badge auto-colorization script that is floating around in userland. If this PR is merged, the existing script(s) will break. I have created a version that works with these new changes: https://gist.github.com/wescopeland/03518db84128d092818768904655b7d7

public/gameInfo.php
```diff
@@ -1266,8 +1264,11 @@ function ResetProgress() {
                                 $completionPctHardcore = sprintf("%01.2f", ($wonByHardcore / $numDistinctPlayersCasual) * 100);
                             }
 
+                            $imgFadeClasses = 'grayscale blur-[0.8px] contrast-[0.65] hover:grayscale-0 hover:contrast-100 hover:blur-0 transition duration-300';
```
Jamiras, line 1267: This is too much blur.

image

Here's how the `_lock` file is created:
https://github.com/RetroAchievements/RAWeb/blob/47f3e56f4933b89f5f22e990d9e0e21e960339a7/app_legacy/Helpers/util/media.php#L109-L113

Jamiras: Also note that the icons for game 16839 were specifically designed to leverage the greyscale algorithm, and don't appear to be greyscaling (or contrasting) exactly the same.

> When possible, I've used cyan and orange(mostly because a chart seemed to indicate that these colors contrast each other even when someone is colorblind) with the exact same shade-Value to create badges that are "invisible" when they are locked, but become visible when unlocked. Hopefully, this will allow people who may have never played MYST before to play mostly-blind without the achievements or badges spoiling what to do or what clues are necessary.

image

Jamiras (commented):
I'm not sure how much I like this. Getting to see the colorized version of the badge should be a reward.

Having said that, I realize that tooltips always show the colorized version and tooltips appear for achievements in the recently played games list on a user's profile, so this isn't unprecedented.

wescopeland:
> Also note that the icons for game 16839 were specifically designed to leverage the greyscale algorithm, and don't appear to be greyscaling (or contrasting) exactly the same.

That is impressively clever.

I have spent about an hour today messing with many combinations of CSS filters trying to get it to match what the back-end image processor is doing. I can get the blur to be identical, but the contrast is always slightly off. The only other alternative I can think of is to load _both_ images, overlay the `_lock` image on top of the non-lock image, and transition to opacity 0 on hover. This sounds bad for performance though.

Philosophically, maybe it's better to go the other way around. If seeing the colorized version of awards is a reward in and of itself, then it sounds like "tooltips always show the colorized version and tooltips appear for achievements in the recently played games list on a user's profile" is actually a bug which this change would codify into a feature.

Regardless, I don't think a pure CSS solution is going to be possible here, and so I'm going to close this PR.

Issue #1468, opened 2023-04-13
[Bug] [Minor] Ghost Unread Message bug


Alexdatadestroyer:
Noticed this suddenly these days after I deleted a message on my Inbox on RA Website: **The mail letter icon was still showing 1 instead of 0** after deleted an unread message.

**Try for yourself:**

1. Send a message to you or take any message that you received as example for doing this task;
2. Then, click to read the message and open the 3-option box (Delete | Mark as Unread | Reply);
3. Click on "Mark as Unread" and delete the message;
4. Take a look at mail letter icon after page reload with "Deleted" green message on top. Even if you reload the page again, this bug remains.

**Expected behavior:**

The mail letter icon was expected to go from 1 to 0 even if you marked that message as unread after reloading the page because you won't have that message pending to be read again, right? xD

**Screenshots:**

![](https://i.imgur.com/g8TWNja.png)

![](https://i.imgur.com/kqBtC0I.png)

However, this bug can be mitigated by 2 ways:

**a)** By creating another message to you, read it and delete without marking that as unread as usual;
**b)** By taking another message as example, marking as unread and unmark it and reload the page. Since that message was marked as unread once, if you reload the page, that message will be highlighted. If you click on it, it will be marked as read once again and mail letter icon bug will be fixed on next page reload.

**Exception Level: Low** -- Does NOT happen so much...

PR #1489, opened 2023-04-19
fix(messages): remediate deleted unread messages not affecting unread message count

wescopeland:
This PR resolves https://github.com/RetroAchievements/RAWeb/issues/1468.

**Issue**
Unread messages that are deleted by the user do not update the user's unread message count.

**Root Cause**
When a message is marked as read, a function `UpdateCachedUnreadTotals()` is called. This does not seem to be called on delete, resulting in the cached unread total not being updated if a delete occurs for an unread message.

**Solution**
`markMessageAsRead()` is now immediately invoked when a delete occurs, causing the cached value to update correctly.


app_legacy/Helpers/database/message.php
```diff
@@ -227,6 +227,10 @@ function DeleteMessage(string $user, int $messageID): bool
         return false;
     }
 
+    // If an unread message is deleted by the user, the total unread
+    // messages count displayed on the site will not be updated correctly.
+    markMessageAsRead($user, $messageID);
```
Jamiras, line 232: This marks the message as read, then calls `UpdateCachedUnreadTotals`. Couldn't you just call that after deleting the message?

app_legacy/Helpers/database/message.php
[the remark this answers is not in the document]
Jamiras: `UpdateCachedUnreadTotals` relies on the data in the Messages table. This is just creating an audit message.

app_legacy/Helpers/database/message.php
wescopeland: I think maybe I was going crazy this morning. Invoking `UpdateCachedUnreadTotals()` seems to work. Resolved in https://github.com/RetroAchievements/RAWeb/pull/1489/commits/0213a998097e645dbd5d798ccbf9d45242c689db.

Jamiras (approved):

PR #1491, opened 2023-04-19
add links to new comments in notifications

Jamiras:
https://discord.com/channels/310192285306454017/1084948245953908776/1084948245953908776

Works for Game, Achievement, User Wall, and Ticket comment notifications.


wescopeland:
I have tested each scenario and verified they are working.

I am noticing a strange behavior on the achievement deep links. It seems on the auto-scroll to the hash, the browser is not accounting for the sticky navbar height, and the navbar is concealing the linked content. I am not able to repro this on mobile, only on desktop. I only observed this on achievement deep links, but I think that's because this specific achievement has a lot of earners (thus increasing the page height).

https://user-images.githubusercontent.com/3984985/233220613-88c00eea-0201-405c-ade9-00d252b331bc.mp4

PR #1498, opened 2023-04-21
fix(user): force locked achievement tooltips on user pages to grayscale

wescopeland:
This PR addresses an issue raised in https://github.com/RetroAchievements/RAWeb/pull/1487 where locked achievement badges shown in tooltips on user profile pages are not in grayscale.

I have come to believe this behavior is a bug, mostly due to badge artwork like what was created for [Myst on 3DO](http://retroachievements.org/game/16839).

This PR restores these images to be grayscale until the achievements are unlocked by the user.

https://user-images.githubusercontent.com/3984985/233530759-9cbc2a91-4ccc-4e4f-99a9-1ce4cd431aac.mp4

### DIFF
--- a/app_legacy/Helpers/render/achievement.php
+++ b/app_legacy/Helpers/render/achievement.php
@@ -101,10 +101,11 @@ function renderAchievementCard(int|string|array $achievement, ?string $context =
     $achPoints = $data['Points'] ?? null;
     $badgeName = $data['BadgeName'] ?? null;
     $unlock = $data['Unlock'] ?? null;
+    $badgeImgSrc = media_asset("Badge/{$badgeName}" . ($unlock ? '' : '_lock') . ".png");
     $gameTitle = renderGameTitle($data['GameTitle'] ?? null);
 
     $tooltip = "<div class='tooltip-body flex items-start gap-2 p-2' style='max-width: 400px'>";
-    $tooltip .= "<img src='" . media_asset("Badge/$badgeName.png") . "' width='64' height='64' />";
+    $tooltip .= "<img src='$badgeImgSrc' width='64' height='64' />";
     $tooltip .= "<div>";
     $tooltip .= "<div><b>$title</b></div>";
     $tooltip .= "<div class='mb-1'>$description</div>";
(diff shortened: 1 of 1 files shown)


Jamiras:
This has the side effect of making all embedded achievement tooltips disabled, regardless of whether or not the player has unlocked them.

image

I think it's desirable to show the colorized icon in the embed itself. For now, the tooltip for an embedded achievement should also be colorized. Maybe in the future, we can include the user unlock state there as well.

This is most likely why the tooltips on the profile page always show colored.

wescopeland:
> I think it's desirable to show the colorized icon in the embed itself. For now, the tooltip for an embedded achievement should also be colorized. Maybe in the future, we can include the user unlock state there as well.

@Jamiras I agree. I made a mistake in my original implementation.

I have now updated this implementation such that only the embeds on the user profile are greyscale.

app_legacy/Helpers/render/achievement.php
```diff
@@ -101,10 +103,11 @@ function renderAchievementCard(int|string|array $achievement, ?string $context =
     $achPoints = $data['Points'] ?? null;
     $badgeName = $data['BadgeName'] ?? null;
     $unlock = $data['Unlock'] ?? null;
+    $badgeImgSrc = media_asset("Badge/{$badgeName}" . ($shouldForceLockedImage ? '_lock' : '') . ".png");
```
Jamiras, line 106: Wouldn't it be easier to just pass `$icon` through and change this to:
```
$badgeImageSrc = $icon ?? media_asset("Badge/{$badgeName}.png");
```

wescopeland: Quite a bit easier 😊

I've made this change in https://github.com/RetroAchievements/RAWeb/pull/1498/commits/425d56b5f0fb895b3146c6812934964ce997f226.

I have kept the variable name as `$badgeImgSrc` to slightly imply it's being attached to the `src` attribute of an `` tag.

The argument name I've set as `$iconUrl`.

Jamiras (approved):

PR #1499, opened 2023-04-21
"Hubs" , "Fallback" and "Events"  Icon Update

oddgoo:
-Gave systems "Hubs" and "Events" a short name
-Created Hubs Icon
-Copied over Events icon
-Updated Fallback/Unknown Icon with a better generic "RA" logo
-Increased contrast of the WASM-4 Icon
-renamed PICO-8 icon to lower-case

image

image

image


Jamiras:
I think the greyscale RA logo is fine for "All Games", but I don't really like it on the individual pages.

image

vs the non-description console box:

image

oddgoo:
> I think the greyscale RA logo is fine for "All Games", but I don't really like it on the individual pages.
> 
> image
> 
> vs the non-description console box:
> 
> image
> 

The best solution to this is to complete / add icons to all the non-supported systems on the page that still don't have one. I'm happy to scope that out soon :)

oddgoo:
@Jamiras As requested, added placeholders for system so that they can continue using the non-descript black box:

image

Also added a couple more icons that existed in the pack.

image

PR #1500, opened 2023-04-21
Console renders for game lists, game pages and avatars

EmoonX:
Parses and stylizes console names (e.g `(Game Boy)`, `(PlayStation 2)` etc) so those are rendered in a separate line and with the respective icons.

image
image

A similar style is also reflected now in game/achievement pages, for the sake of consistency:

image
image

### DIFF
--- a/app_legacy/Helpers/render/game.php
+++ b/app_legacy/Helpers/render/game.php
@@ -22,7 +22,14 @@ function gameAvatar(
             $title = $game['GameTitle'] ?? $game['Title'] ?? null;
             $consoleName = $game['Console'] ?? $game['ConsoleName'] ?? null;
             sanitize_outputs($title);   // sanitize before rendering HTML
-            $label = renderGameTitle($title, console: $consoleName);
+
+            $label = "<div class='inline-flex flex-col gap-0.5'>";
+            $label .= renderGameTitle($title);
+            if ($consoleName) {
+                $consoleID = $game['ConsoleID'] ?? 1;
+                $label .= renderGameConsole($consoleName, $consoleID);
+            }
+            $label .= "</div>";
         }
 
         if ($icon === null) {
@@ -53,9 +60,7 @@ function gameAvatar(
 /**
  * Render game title, wrapping categories for styling
  */
-function renderGameTitle(
-    ?string $title = null, bool $tags = true, ?string $console = null
-): string
+function renderGameTitle(?string $title = null, bool $tags = true): string
 {
     // Update $html by appending text
     $updateHtml = function (&$html, $text, $append) {
@@ -64,15 +69,6 @@ function renderGameTitle(
 
     $title ??= '';
     $html = $title;
-
-    if ($console) {
-        $iconSrc = asset("assets/images/system/nes.png");
-        $span = "<span class='tag console'>"
-            . "<img src='$iconSrc' alt=''>"
-            . "<span>$console</span>"
-            . "</span>";
-        $html .= " $span";
-    }
 
     $matches = [];
     preg_match_all('/~([^~]+)~/', $title, $matches);
@@ -92,7 +88,17 @@ function renderGameTitle(
         $updateHtml($html, $matches[0], $tags ? " $span" : '');
     }
 
-    return $html;
+    return "<div>$html</div>";
+}
+
+function renderGameConsole(string $consoleName, int $consoleID): string
+{
+    $iconSrc = getConsoleIconSrc($consoleID);
+
+    return "<div class='tag console'>"
+        . "<img src='$iconSrc' width='22' height='22' alt=''>"
+        . "<span>$consoleName</span>"
+        . "</div>";
 }
 
 /**
@@ -483,4 +489,12 @@ function renderCompletionIcon(
     }
 
     return "<div class='$class' title='$tooltipText'>$icon</div>";
+}
+
+function getConsoleIconSrc(int $consoleID): string
+{
+    $cleanConsoleShortName = Str::lower(str_replace("/", "", config("systems.$consoleID.name_short")));
+    $iconName = Str::kebab($cleanConsoleShortName);
+
+    return asset("assets/images/system/$iconName.png");
 }
--- a/app_legacy/Helpers/render/game.php
+++ b/app_legacy/Helpers/render/game.php
@@ -99,7 +99,16 @@ function renderConsoleIcon(string $consoleName, int $consoleID): string {
     $iconName = Str::kebab($cleanSystemShortName);
     $iconSrc = asset("assets/images/system/$iconName.png");
 
-    return "<img class='console-icon' src='$iconSrc' alt='$consoleName icon' onerror='this.src=\"$fallBackConsoleIcon\"'>";
+    return <<<HTML
+        <img class='console-icon'
+            src='$iconSrc'
+            alt='$consoleName icon'
+            onerror='this.src="$fallBackConsoleIcon"'>
+    HTML;
+}
+
+function renderGameTitleAndConsole(string $gameTitle, string $consoleName, int $consoleID): string {
+
 }
 
 /**
--- a/app_legacy/Helpers/render/game.php
+++ b/app_legacy/Helpers/render/game.php
@@ -107,10 +107,6 @@ function renderConsoleIcon(string $consoleName, int $consoleID): string {
     HTML;
 }
 
-function renderGameTitleAndConsole(string $gameTitle, string $consoleName, int $consoleID): string {
-
-}
-
 /**
  * Render game breadcrumb prefix, with optional link on last crumb
  *
--- a/app_legacy/Helpers/render/game.php
+++ b/app_legacy/Helpers/render/game.php
@@ -27,7 +27,7 @@ function gameAvatar(
             $label .= renderGameTitle($title);
             if ($consoleName) {
                 $consoleID = $game['ConsoleID'] ?? 1;
-                $label .= renderGameConsole($consoleName, $consoleID, size: 22, avatar: true);
+                $label .= renderGameConsole($consoleName, $consoleID, size: 15, avatar: true);
             }
             $label .= "</div>";
         }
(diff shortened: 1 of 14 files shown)


Soulsderfuchs:
Will this also affect the relations on the sidebar of a game page?, I feel like this would make the console text almost unreadable at that size.

EmoonX:
>Will this also affect the relations on the sidebar of a game page?

These ones? Font size is essentially the same and looks good to me, but increasing is fine either way.

image

Soulsderfuchs:
> > Will this also affect the relations on the sidebar of a game page?
> 
> These ones? Font size is essentially the same and looks good to me, but increasing is fine either way.
> 
> image

Yeah the size of the text of the console is definetly smaller than the text of the game in this images

EmoonX:
>Yeah the size of the text of the console is definetly smaller than the text of the game in this images

That was somewhat intentional, to balance the fact we're highlighting the metadata with CSS (same was done previously to tags, subsets etc).

I increased the font a little bit.

image

wescopeland:
The x-axis margin is extremely harsh on game pages. On these pages, I believe the left margin should be entirely removed and the right margin should be reduced.

Unless the left margin is removed, the icons no longer align properly with the game title. Unless the right margin is reduced, the console text is far away from the icon itself.

**Production**
Screenshot 2023-04-21 at 8 28 54 AM

**This PR**
Screenshot 2023-04-21 at 8 27 44 AM

**This PR With Reduced Margin**
Screenshot 2023-04-21 at 8 34 19 AM

app_legacy/Helpers/render/game.php
```diff
@@ -84,7 +88,33 @@ function renderGameTitle(?string $title = null, bool $tags = true): string
...
+
+    $fallBackConsoleIcon = asset("assets/images/system/unknown.png");
+    $iconSrc = $getConsoleIconSrc($consoleID);
+    $class = $avatar ? ' avatar' : '';
+    $textSize = round(($avatar ? .75 : .6) * $size);
+
+    return "<div class='console$class'>"
+        . "<img src='$iconSrc' width='$size' height='$size' alt='' onerror='this.src=\"$fallBackConsoleIcon\"'>"
+        . "<span style='font-size: {$textSize}px'>$consoleName</span>"
+        . "</div>";
```
wescopeland, line 117: nit: Heredoc syntax can be used to avoid the string concatenation and quote escapes.

```suggestion
    return <<<HTML
        <div class='console$class'>
            <img src='$iconSrc' width='$size' height='$size' alt='' onerror='this.src="$fallBackConsoleIcon"'>
            <span style='font-size: {$textSize}px'>$consoleName</span>
        </div>
    HTML;
```

wescopeland:
The readability of the console names on similar game lists improves if the font size is increased from 11px to 12px. This is very likely operating-system based. However, it seems on macOS that subpixel font rendering is not working properly unless these fonts are scaled up to 12px. I assume the same will be true for iOS.

**Current (11px)**
Screenshot 2023-04-21 at 8 50 45 AM

**Proposed (12px) - console text is a bit sharper**
Screenshot 2023-04-21 at 8 50 57 AM

Jamiras:
I'm not really a fan of the bar thing to the left of the icon. Plus, my icons seems smaller than yours.
image

Additionally (not a big deal), the sidebar padding seems to have increased.
image

I like the idea of separating out the console name, but I'm not sold on including the icon.
image

EmoonX:
>Additionally (not a big deal), the sidebar padding seems to have increased.

Vertical padding is larger now to accommodate game avatars taking at least two lines instead of just one. Benefits are more noticeable in wider game lists.

>I like the idea of separating out the console name, but I'm not sold on including the icon.

Looks decent without it. Having a border + an icon + a name can result in some visual clutter indeed. I'm leaving the icon outside lists only then.

EmoonX:
Consoles are displayed in lists this way now:

image

As for game pages, this should look nicer and solve the alignment issues:

image
image

wescopeland:
Game pages and game lists both definitely look better now. I've marked all my existing feedback as resolved. The only other thing I'm noticing is console names are the same size as the H1 (game title) on the page. I propose slightly reducing the console name so the H1 remains visually as the principally important element:

**Current**
Current

**Proposed**
Proposed

Jamiras:
Profile page icons should be enlarged to accommodate this change:
image

And the AotW widget:
image

I'm not sure what to recommend for forum embeds - maybe just leave the console off entirely? It'll still be in the tooltip.
image

Maybe that can be generic logic across the board: if image size < 32, don't show console?

Jamiras:
I don't like the large console icons on the game page. They almost overshadow the game icon itself.
image

vs. current(master):
image

Maybe it's time to rearrange some stuff?
image

EmoonX:
>Profile page icons should be enlarged to accommodate this change:
>
>And the AotW widget:
>
>I'm not sure what to recommend for forum embeds - maybe just leave the console off entirely? It'll still be in the tooltip

I've ended increasing most of the 22/24 px icons to 28/32 px. Even if the icon doesn't cover the whole text to its right, I don't think it's going to be an issue; bound to happen on tighter places anyway, like the sidebar game lists. Also added a fallback for really small icons (like the followers list that uses 16px for those there).

As for the forums, I opted for keeping the rendered style. I find it to be consistent and produces a visual feedback for quick assimilation of the rendered shortcode as a game. They also take less horizontal space now, which IMO makes reading more comfortable.

Jamiras:
I still find the forum rendering harder to read.

Before:
image

After:
image

The alignments in the "table" at the top of the image get my attention more than the data itself.

This is what my eyes are doing on the last line (I don't think I'm even registering the console as that would require me to zigzah back to read the second line before continuing):
image

Also note that the embedded achievement appears to be one pixel lower as a result of the embedded game.

Jamiras:
I think the recent games widget would look better at 48:
image

I'm really struggling with the two-line thing for pretty much anywhere where they're joined to text.
image

Note the position of the asterisk:
image

Or maybe it's just anywhere where the icon is smaller than the two lines of text?

Whereas I really like it in tables:
image

Even this looks good:
image

I think we need to get some other users to look at it.

Soulsderfuchs:
How would this look in a hub?

Jamiras:
> How would this look in a hub?

Like any other list:
image

image

image

image

EmoonX:
> This is what my eyes are doing on the last line (I don't think I'm even registering the console as that would require me to zigzag back to read the second line before continuing):
> image

This is a good argument. It really can disrupt the reading flow.

We can agree the new format looks great on lists (or list-like), yet on inline contexts... not so much. I don't think it's worth the trouble trying to adequate things like forum posts to it. So I'm opting for the following idea as a general fallback (< 32 px):

image
image
image

EmoonX:
>I don't like the large console icons on the game page. They almost overshadow the game icon itself.

Some of them look too large indeed (especially the NES one). It's hard to have an one-size-fits-all since they all have different colors and aspect ratios. I've reduced the icon size from 36 to 26 px and applied @wescopeland's suggestion on the text:

image

Overall, I'm for keeping the idea because:
- Game and console text look better when vertically aligned like that
- I've received positive feedback about it

Another option would be simply dropping the console icon, but I think the header would lose then a bit of its charm...?

---

> Maybe it's time to rearrange some stuff?
> [image]

I imagine this would be hard to get in a good state without breaking things on mobile.

Plus there are also the other pages which need a similar rendering as the game info one (achievement pages, leaderboards, etc), except it's not ideal merely using the full-sized game badge there.

app_legacy/Helpers/render/game.php
```diff
@@ -99,16 +99,19 @@ function renderGameTitle(?string $title = null, bool $tags = true): string
...
-    $iconSrc = asset("assets/images/system/$iconName.png");
+    $iconPath = public_path("assets/images/system/$iconName.png");
+    $iconUrl = file_exists($iconPath) ? asset("assets/images/system/$iconName.png") : $fallBackConsoleIcon;
 
     return <<<HTML
         <img class='console-icon'
-            src='$iconSrc'
+            src='$iconUrl'
             alt='$consoleName icon'
             onerror='this.src="$fallBackConsoleIcon"'>
```
luchaos, line 116: `onerror` can be removed now?

EmoonX: Ah yes. No need for JS now 🙂

app_legacy/Helpers/render/game.php
```diff
@@ -21,11 +21,21 @@ function gameAvatar(
         if ($label !== false) {
             $title = $game['GameTitle'] ?? $game['Title'] ?? null;
             $consoleName = $game['Console'] ?? $game['ConsoleName'] ?? null;
+            sanitize_outputs($title);   // sanitize before rendering HTML
+
+            $flexDirection = $iconSize >= 32 ? 'col' : 'row';
```
wescopeland, line 26: This refactor fixes a bug and seems a bit easier to reason about:

```php
$flexDirection = $iconSize >= 32 ? 'flex-col' : 'flex-row';
$alignItems = $flexDirection === 'flex-col' ? 'items-start' : 'items-center';

$consoleDiv = '';
if ($consoleName) {
    $consoleClassNames = 'console-name';
    if ($flexDirection === 'flex-row') {
        $consoleClassNames .= " mx-0.5 border-l-2 border-white/[.25] pl-1.5";
    }
    $consoleDiv = "<div class='$consoleClassNames'>$consoleName</div>";
}

$gameTitle = renderGameTitle($title);

$label = <<<HTML
    <div class='flex $flexDirection $alignItems gap-1'>
        $gameTitle
        $consoleDiv
    </div>
HTML;
```

Tailwind tokenizes by the entire name of the string being present in a file. It isn't able to detect "col", "or", "start", or "center". If the page doesn't already have those class names elsewhere, the game title will be broken.

app_legacy/Helpers/render/game.php
```diff
@@ -84,7 +94,22 @@ function renderGameTitle(?string $title = null, bool $tags = true): string
...
+ */
+function renderConsoleIcon(string $consoleName, int $consoleID): string {
+    $fallBackConsoleIcon = asset("assets/images/system/unknown.png");
+    $cleanSystemShortName = Str::lower(str_replace("/", "", config("systems.$consoleID.name_short")));
+    $iconName = Str::kebab($cleanSystemShortName);
+    $iconPath = public_path("assets/images/system/$iconName.png");
+    $iconUrl = file_exists($iconPath) ? asset("assets/images/system/$iconName.png") : $fallBackConsoleIcon;
+
+    return "<img class='console-icon' src='$iconUrl' alt='$consoleName icon'>";
 }
```
wescopeland, line 113: Suggestion: Let's turn this into a Blade component and not add another render function to legacy. resources/views/components might be a good fit as this is fairly generic and could exist on any page throughout the site:

```
<x-console-icon consoleName="Mega Drive" consoleId="1" />
```

It seems everywhere this function is being used, it's being `echo`'d out. A swap to using `Blade::render()` should work nicely and prevent us from having to migrate this later.

PR #1502, opened 2023-04-21
feat(forum): autoinject shortcodes for RA site links

wescopeland:
This PR makes a small quality of life improvement to submitting and editing posts on the site forum.

If the content of the post contains a link to a game, achievement, user, or ticket, the link is now automatically replaced with the corresponding shortcode.

This change was inspired and motivated by a similar script released by @Xymjak. 

https://user-images.githubusercontent.com/3984985/233749854-8b30528b-d06f-4839-be23-de4b75ca147c.mp4


app_legacy/Helpers/database/forum.php
```diff
@@ -246,12 +246,35 @@ function setLatestCommentInForumTopic(int $topicID, int $commentID): bool
...
+    /**
+     * Mapping of URL patterns to their corresponding forum shortcodes
+     * @see app/Support/Shortcode/Shortcode.php Implementation for shortcodes.
+     */
+    $urlToShortcodeMap = [
+        '#https://retroachievements\.org/user/(\w+)#' => '[user=$1]',
+        '#https://retroachievements\.org/game/(\d+)#' => '[game=$1]',
+        '#https://retroachievements\.org/achievement/(\d+)#' => '[ach=$1]',
+        '#https://retroachievements\.org/ticketmanager\.php\?i=(\d+)#' => '[ticket=$1]',
+    ];
```
Jamiras, line 261: Is there any way you could make this work for any host (i.e. local, stage. etc)?

Jamiras (commented):
It looks like there's some proto-code for doing this in the non-legacy shortcode file: https://github.com/RetroAchievements/RAWeb/blob/master/app/Helpers/shortcode.php#L229-L300

Maybe you could call/expand that?

app_legacy/Helpers/database/forum.php
wescopeland: > It looks like there's some proto-code for doing this in the non-legacy shortcode file: [`master`/app/Helpers/shortcode.php#L229-L300](https://github.com/RetroAchievements/RAWeb/blob/master/app/Helpers/shortcode.php?rgh-link-date=2023-04-22T14%3A19%3A39Z#L229-L300)
> 
> Maybe you could call/expand that?

Thanks, I was not aware of this code.

I have opted to expand it by adding the following functions:

- `normalize_game_shortcodes()`
- `normalize_achievement_shortcodes()`
- `normalize_ticket_shortcodes()`

These functions, along with the original `normalize_user_shortcodes()` can all be called via a `normalize_shortcodes()` function I have also added.

I've chosen not to mess with `url2shortcode()`, as this seems to be special-purpose for any URL tag and it is specifically used in the home page news component.

I have removed the auto-inject function I previously created, and am now calling `normalize_shortcodes()`.

Additionally, I have updated the regex patterns list to support URLs from any environment.

Jamiras (approved):

PR #1503, opened 2023-04-23
feat(user): use dashed border on vintage/outdated game awards

wescopeland:
This PR adds a visual indication to user awards if a set has received a revision since the award was granted.

**How?**
It seems this functionality existed at one point, as I did find a "but more achievements have been added!" message that is not currently possible to enable.

I assume this was disabled because the query to retrieve this data is expensive. As an alternative to adding a query, I have opted to re-use data on the user profile to derive this information, namely the data currently used in the "Completed Games" widget.

We can tell from the values fetched for the Completed Games widget whether or not a site award is outdated. This prevents us from needing to add new DB calls to the user profile page.

**Other Notes**
* Vintage "Completions" also get a border, albeit it shows the blue global softcore color. We should probably define this color in our Tailwind palette, but I felt that was probably beyond the scope of this PR.
* I've also brought this functionality over to the reorder site awards page.

Resolves #692.

https://user-images.githubusercontent.com/3984985/233813529-afa4e15d-b22e-4fda-bc4b-92c09962622d.mp4


Jamiras (approved):
I approve the implementation, but be prepared for some backlash. If not for this feature, then for towards the revision process in general. This change will disrupt the cleanliness of a user's wall.

For example, about 10% of my masteries are showing this behavior. Now I suddenly have 15 games to revisit or hide to make everything uniform again. I was only aware of one or two of the revisions, and was purposefully ignoring going back to them because I didn't want to replay the game and the changes didn't interest me.

Jamiras:
Also note: this addresses #692, though not with any of the solutions proposed there.

wescopeland:
> I approve the implementation, but be prepared for some backlash. If not for this feature, then for towards the revision process in general. This change will disrupt the cleanliness of a user's wall.
> 
> For example, about 10% of my masteries are showing this behavior. Now I suddenly have 15 games to revisit or hide to make everything uniform again. I was only aware of one or two of the revisions, and was purposefully ignoring going back to them because I didn't want to replay the game and the changes didn't interest me.

Thank you for this feedback. I agree that this will likely cause some backlash. My perception is not knowing about the "vintage" masteries also causes backlash; I am struggling to grasp how we can please everyone here 🤔

Do you think it would be appropriate to have some kind of setting that allows users to disable this feature from their profile? If so, would `websitePrefs` be the appropriate place for it to live?

wescopeland:
I'm going to assume the answer to whether this should be setting-driven is "yes". That's probably a less disruptive experience for the end user. I'm going to migrate this to a draft state until V3 hits prod, then rework this to support a new user setting which allows people to disable this functionality.

PR #1504, opened 2023-04-23
feat: add a sortable 'Retro Ratio' column to games list

amine4567:
This PR adds a "Retro Ratio" column to the games lists with sorting ability, this is a feature that has been asked for many times.

The "Retro Ratio" per game is calculated by dividing the retro points (white points) by the regular points (blue points), just like how the per player Retro Ratio is calculated.

https://user-images.githubusercontent.com/29414494/233844645-67c05034-77da-4185-a729-e3f9f2640e71.mp4


app_legacy/Helpers/database/game.php
```diff
@@ -289,8 +289,12 @@ function getGamesListByDev(
                 $whereCond
                 GROUP BY gd.ID
                 $havingCond";
+    $query = "SELECT *, 
+              CASE WHEN gl.MaxPointsAvailable > 0 THEN ROUND(gl.TotalTruePoints/gl.MaxPointsAvailable, 2) ELSE 0.00 END AS RetroRatio
+              FROM ($query) as gl
+             ";
```
Tsearo, line 295: From a functionality perspective everything works great 👍

Though I did some quick testing and I believe you'll get better performance if you just put this part in the existing query instead of doing an extra select from the original query
`CASE WHEN SUM(ach.Points) > 0 THEN ROUND(gd.TotalTruePoints/SUM(ach.Points), 2) ELSE 0.00 END AS RetroRatio,`

Jamiras: I agree on putting it in the main select, but doing so doesn't seem to have any impact on performance.

amine4567: I didn't want to put it in he main query at first because I wasn't sure if the database was gonna compute the `SUM` 2 additional times or not. 

But I did some testing on my part too, the current query implementation takes 0.5 seconds while the change proposed by @Tsearo takes only 0.06 seconds on average. All this done on the test data I have on my local database (which is up to March 2023).

I made and pushed the change.

Jamiras: You have to be careful when measuring query performance. The database may be caching results between runs. My testing showed ~100ms both before and after.

If you haven't already, I highly recommend enabling the debug bar. Edit your `.env` file and change `DEBUGBAR_ENABLED` to `true`. This lets you see all the queries used by the page and timings for each individual query.

Jamiras (approved):

Tsearo (approved):

Tsearo (approved):

app_legacy/Helpers/database/game.php
amine4567: Yeah you are right, the debug bar shows that it didn't impace performance much.

PR #1511, opened 2023-04-24
Show count of tickets for dev in game list

Tsearo:
Similar update as the Leaderboard change to now show the number of tickets from the dev vs the total tickets for the set. As with that change the total reflected at the bottom will be just for the dev though these totals could potentially also show dev vs total if desired

image

The value on the left will filter tickets to just the dev while the one on the right will show all tickets for the game.
Not fully set on this method vs just showing the full `X of Y` as a single hyperlink or taking off the dev hyperlink when the dev has `0 of Y` tickets so I'm open to updating if this implementation is not ideal


app_legacy/Helpers/database/game.php
```diff
@@ -266,7 +267,8 @@ function getGamesListByDev(
                            SUM(CASE WHEN ach.Author = :myPtsDev THEN ach.Points ELSE 0 END) AS MyPoints,
                            SUM(CASE WHEN ach.Author = :myRRDev THEN ach.TrueRatio ELSE 0 END) AS MyTrueRatio,
                            SUM(CASE WHEN ach.Author != :notMyAchDev THEN 1 ELSE 0 END) AS NotMyAchievements,
-                           lbdi.MyLBs,";
+                           lbdi.MyLBs,
+                           ticks.MyOpenTickets,";
```
Jamiras, line 271: This clause only exists if `$ticketsFlag`. You can see the problem by trying to open the individual dev stats page.

Jamiras:
> Not fully set on this method vs just showing the full X of Y as a single hyperlink or taking off the dev hyperlink when the dev has 0 of Y tickets so I'm open to updating if this implementation is not ideal

I think it should be a single link. I'm undecided whether it should link to the user's tickets for the game, or just all tickets for the game. I'm leaning towards the second

Jamiras (approved):

PR #1512, opened 2023-04-24
Add link to dev stats page under My Pages dropdown

Tsearo:
Small change to add the `View detailed stats for X` link to the My Pages dropdown, currently the only way to access the page is from the devs user page but I frequently find myself wanting a direct link there without needing to check my profile

Open to naming it differently such as `My Stats` or `My Detailed Stats` etc

image

Alternative way to access the link found here on profile
image


app_legacy/Helpers/render/layout.php
```diff
@@ -185,6 +185,7 @@ function RenderToolbar(): void
         echo "<ul>";
         echo "<li><a href='/user/$username'>Profile</a></li>";
         echo "<li><a href='/gameList.php?d=$username'>My Sets</a></li>";
+        echo "<li><a href='/individualdevstats.php?u=$username'>My Dev Stats</a></li>";
```
wescopeland, line 188: Will this appear for all users, including non-devs?

Jamiras: The dev stats page will redirect back to the user page if the user doesn't have any achievements created.
https://github.com/RetroAchievements/RAWeb/blob/073ed89421841be34d468ed56e83f346942ea593/public/individualdevstats.php#L21-L26

But I don't want to do that query to populate the menu as it would occur on every page.

I think it's acceptable to hide that link for "retired" devs.

Alternately, the user page relies on the user's ContribCount being non-zero for showing the link. That information might be available here.

Tsearo: Added in the checks mentioned above as they make sense (contribution count was easy to pull from the user here), I did leave My Sets and Dev Stats for retired devs since imo even if you are no longer a dev you still have sets and stats but can remove them if still desired

Jamiras (approved):

PR #1513, opened 2023-04-24
feat(news): rewrite the carousel and improve ux

wescopeland:
This PR rewrites the news carousel on the home page in an effort to make numerous UX improvements and bug fixes.

**Why?**
The current news carousel is implemented using a legacy jQuery `rcarousel` component. `rcarousel` has not received an update in almost 10 years. It has a large bundle size, is tethered to jQuery, is not responsive, and is not mobile-friendly. It was an impressive library for its time, but it unfortunately does not leverage features of the modern web.

There are numerous bugs in its implementation. Two that come to mind:

1. Because `rcarousel` requires the component's structure to be managed by client-side code, there is a flash of unstructured content that appears on the home page on first load for every user with a modest connection speed. The duration of this flash is at the mercy of the user's network speed.

2. Content is truncated, and sometimes unreadable, on devices utilizing the XS, SM, and MD breakpoints. This notably includes all mobile phone users and most tablet users. This is likely because `rcarousel` was written before mobile browsers were all that capable.

**I know at some point there is a desire to move away from a carousel for news entirely.** However until that can happen, it makes sense to enable this feature to be the best it can possibly be.

---

**How?**

The carousel has been rewritten to be fully implemented in Blade, Tailwind, Alpine.js, and TypeScript. Additionally, the implementation has integration test coverage.

High level UX highlights:
* There is no longer a flash of unstyled content in the carousel itself. A separate flash _does_ remain from a bug in the page layout, but the carousel is no longer exacerbating the issue.
* The giant navigation buttons have been shrunk down to hang along the carousel sides.
* The background image is now full-width and responsive.
* Hovering on a news item darkens and slightly blurs the image to improve readability.
* Hovering on a news item pins the carousel in place - it will not autoscroll while hovered.
* To improve perceived performance, the first item in the carousel no longer has a text animation.
* Text animations are completely disabled on mobile.
* Mobile users can now swipe to advance through the carousel (!)
* Content is no longer truncated on mobile devices, including down to the iPhone SE.
* Touch events will now pin the carousel in place, preventing auto-scrolls from occurring.
* The carousel will no longer auto-scroll when the user tabs away from the page.

Implementation highlights:
* The structure of the carousel is pure CSS and driven by the scroll snap API.
* I have renamed the component in the _news_ folder to _carousel-2.blade.php_. This was done because I figured if the filename stayed the same, the diff would make no sense. I am happy to change it back though if that is preferred.

---

https://user-images.githubusercontent.com/3984985/233879744-73017b1b-0d33-4bce-a8c7-840f85bbe54b.mp4

https://user-images.githubusercontent.com/3984985/233879764-0ed14a78-30b2-4d43-a7ef-33aa48929507.mp4

https://user-images.githubusercontent.com/3984985/233879771-87816790-4100-49b5-b300-ae9726fefaf8.mp4

https://user-images.githubusercontent.com/3984985/233879780-9b9b2e50-ba44-40da-8fbc-0caca8fbfd62.mp4


resources/js/carousel/newsCarousel.ts
wescopeland, line 1: For reference, Vite compiles all this down to about 1Kb.

Jamiras:
Minor: Clicking on a specific page causes the indicator for that page to remain selected during autoscroll:
image

Minor: Resizing the window resets the carousel to the first page.

wescopeland:
> Minor: Clicking on a specific page causes the indicator for that page to remain selected during autoscroll: image

Nice catch, addressed in https://github.com/RetroAchievements/RAWeb/pull/1513/commits/77ace74451c7302d9d5f91aae61dab4999ed41d9.

> Minor: Resizing the window resets the carousel to the first page.

Intentional -- I can fix this if it's deemed a showstopper, but this behavior was intentionally coded in the TypeScript implementation because of the nature of the CSS scroll snap API. Because the entire carousel lives on a dynamically sized "track" (div), the user's current slide is easily lost when resizing the browser window. To avoid this, it's reset to the first index on resize.

The problem is solvable, but not without adding a lot of client-side complexity.

Jamiras:
At unreasonably small window sizes, a horizontal scrollbar appears that you can use to jump between slides without updating the selected slide.

image

Jamiras:
> > Minor: Resizing the window resets the carousel to the first page.

> The problem is solvable, but not without adding a lot of client-side complexity.

That's fine. I was just noting the change in behavior.

Jamiras (approved):
Code is functional. Noted a couple of minor things. 

Would like to get @luchaos's approval on the implementation.

wescopeland:
@luchaos @Jamiras This branch has been updated to resolve all outstanding conflicts with V3.

Jamiras (approved):

PR #1514, opened 2023-04-24
Add date selector to History page

Tsearo:
Adds a date selector to the History page to allow easy selection defaulting to the current days achievements as well as being able to specify any other date since when you have a large history graph it is difficult to pinpoint a specific day if there is a lot of activity

image


public/history.php
```diff
@@ -230,8 +241,13 @@ function resize() {
         $memberSince = !empty($userSignedUp) ? getNiceDate(strtotime($userSignedUp), true) : '';
         echo "Member since: " . $memberSince . "<br>";
         echo "<br>";
-        echo "<br>";
-        echo "<br>";
+        echo "<form name='gotodateform' action='/historyexamine.php' onsubmit='convertDate()'>";
+        echo "<label for='d'><b>Jump to Date: </b></label>";
```
wescopeland, line 245:
```suggestion
        echo "<label for='d'><span class='font-bold'>Jump to Date: </span></label>";
```

I believe there is a push to get away from using these `<b>` tags in favor of spans with Tailwind classes. (also a similar push elsewhere to mitigate `<br>` usage)

ref: https://tailwindcss.com/docs/font-weight

public/history.php
```diff
@@ -230,8 +241,13 @@ function resize() {
...
+        echo "<input type='date' id='dateinput' value='" . strftime("%Y-%m-%d", $dateUnix) . "' />";
+        echo "<input type='hidden' name='d' value='$dateUnix' />";
+        echo "<input type='hidden' name='u' value='$userPage' />";
+        echo "<input type='submit' value='Goto Date'/>";
```
wescopeland, line 249:
```suggestion
        echo "<input type='submit' value='Goto Date' class='ml-1' />";
```

Unless I do this, the submit button slightly overlays the input field on macOS.

ref: https://tailwindcss.com/docs/margin

Screenshot 2023-04-24 at 5 44 16 AM

public/history.php
```diff
@@ -187,6 +189,15 @@ function resize() {
   }
 </script>
 
+<script>
+  function convertDate() {
+    var field = document.gotodateform.dateinput;
+    var timestamp = new Date(field.value).getTime() / 1000;
+    document.gotodateform.d.value = timestamp;
+    return true;
+ }
```
wescopeland, line 198: nit: Seems there is some missing white space here, I think maybe ESLint is not running on these PHP files 😅

```suggestion
  }
```

public/history.php
```diff
@@ -187,6 +189,15 @@ function resize() {
   }
 </script>
 
+<script>
+  function convertDate() {
+    var field = document.gotodateform.dateinput;
+    var timestamp = new Date(field.value).getTime() / 1000;
+    document.gotodateform.d.value = timestamp;
```
wescopeland, line 196:
```suggestion
    const { dateinput, d } = document.gotodateform;
    const timestamp = new Date(dateinput.value).getTime() / 1000;
    d.value = timestamp;
```

nit: Functionally equivalent, but slightly safer for changes later down the road thanks to immutability.

wescopeland (commented):
Looks great, just a few nit comments here

Jamiras:
@wescopeland - not sure if this impacts your review at all, but all this PR is doing is copying code from historyexamine.php (where the dropdown already exists) to this file so you can go straight to a date without first clicking on a random date in the graph.

wescopeland:
@Jamiras No worries at all, I didn't realize this was a carbon copy of existing code. My stuff was mostly just nits anyway - I've marked everything as resolved.

Tsearo:
I think it's good feedback and I plan to apply it to the historyexamine page where I copied it from 👍

Jamiras (approved):

PR #1517, opened 2023-04-25
feat(leaderboard): localize `VALUE` output

wescopeland:
This PR localizes leaderboard `VALUE` outputs to include number separators. `POINTS` outputs are unaffected and still use zero-padded numbers.

**Before**
Screenshot 2023-04-25 at 4 27 53 PM

Screenshot 2023-04-25 at 4 27 58 PM

**After**
Screenshot 2023-04-25 at 4 27 32 PM

Screenshot 2023-04-25 at 4 27 38 PM


app_legacy/Helpers/database/leaderboard.php
```diff
@@ -492,7 +492,7 @@ function GetFormattedLeaderboardEntry(string $formatType, int $scoreIn): string
     }
```
luchaos, line 493: I guess this could be removed then, too?
It'S what makes it padded numbers in the first place.

luchaos (approved):

app_legacy/Helpers/database/leaderboard.php
wescopeland: Personally, I don't think it should be without a plurality of the achievement devs in agreement. `SCORE` is deliberately picked by the dev as a stylistic choice and results in the zero-padded numbers (though it appears to be a very unpopular choice 😅).

wescopeland: Same 😶

Jamiras: > though it appears to be a very unpopular choice

It's actually fairly common:
```
mysql> select Format, count(*) from LeaderboardDef group by 1 order by 2 desc;
+-----------+----------+
| Format    | count(*) |
+-----------+----------+
| VALUE     |    16675 |
| TIME      |    15106 |
| SCORE     |    10131 |
| MILLISECS |     6744 |
| TIMESECS  |     2262 |
| MINUTES   |       57 |
+-----------+----------+
6 rows in set (0.04 sec)
```
I can definitely see it as the preferred display format for old arcade-style games as they zero-padded on the screen, but for our purposes it doesn't seem necessary.

Jamiras (approved):

app_legacy/Helpers/database/leaderboard.php
wescopeland: Sorry, I was more referring to: https://discord.com/channels/310192285306454017/511718348178849792/1100181080427986984

Although the sample of voters here is hardly representative of the userbase as a whole.

Jamiras: Also note that this might have undesired effects as some leaderboards scale their values. For example, the Pokemon Pinball score leaderboard measures thousands of points as the in-game value can exceed 32-bit. It looks like it's using SCORE right now, so it's not affected, but I can't say there aren't leaderboards that might be.

Jamiras (approved):

PR #1518, opened 2023-04-25
fix(user): remediate mobile css issue with user following notice

wescopeland:
This PR resolves an issue visible when opening a user page on mobile if the user follows you. Right now, the "Follows you" text is clipping into the user avatar.

Additionally, if the user follows you and you don't follow them, the "Follow" button now says "Follow Back".

**Before**
Screenshot 2023-04-25 at 4 55 28 PM

**After**
Screenshot 2023-04-25 at 4 55 06 PM

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -250,11 +250,14 @@ RenderContentStart($userPage);
             if ($areTheyFollowingMe) {
                 echo "<span class='hidden sm:block px-3'>Follows you</span>";
             }
-            echo "</div>";
-        }
 
-        if ($areTheyFollowingMe) {
-            echo "<p class='sm:hidden'>Follows you</p>";
+            echo "</div>";
+
+            // This needs to be a sibling to the user relationship div to prevent spacing
+            // issues with the user avatar.
+            if ($areTheyFollowingMe) {
+                echo "<p class='sm:hidden'>Follows you</p>";
+            }
         }
 
         echo "<br>";
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -206,15 +206,18 @@ RenderContentStart($userPage);
         }
 
         if (isset($user) && ($user !== $userPage)) {
-            echo "<div class='flex items-center gap-1'>";
-            $friendshipType = GetFriendship($user, $userPage);
-            switch ($friendshipType) {
+            echo "<div class='flex items-center sm:gap-1'>";
+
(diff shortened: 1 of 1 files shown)


public/userInfo.php
```diff
@@ -244,10 +247,17 @@ function resize() {
...
+                echo "<span class='hidden sm:block px-3'>Follows you</span>";
             }
+
             echo "</div>";
+
+            // This needs to be a sibling to the user relationship div to prevent spacing
+            // issues with the user avatar.
+            if ($areTheyFollowingMe) {
+                echo "<p class='sm:hidden'>Follows you</p>";
+            }
```
Jamiras, line 260: Why does this need to be present twice? (line 251 + line 259)

wescopeland: A bug in our Tailwind version was causing the class name `sm:justify-start` to not work properly, which was prohibiting this paragraph tag from fitting into a single flex container.

I have bumped our version to the latest minor release which fixes this issue, allowing me to use a single flex container and a single paragraph tag.

Jamiras:
I get the following error running `npm rub build` on this branch now:
```
failed to load config from /home/brian/source/RAWeb/vite.config.ts
error during build:
Error: The package "@esbuild/linux-x64" could not be found, and is needed by esbuild.
```
I've done the full composer/npm rebuild per `.hooks/post-merge`

wescopeland:
If you run `npm install` and then do `npm run build`, do you see the same error?

wescopeland:
> I get the following error running `npm rub build` on this branch now:
> 
> ```
> failed to load config from /home/brian/source/RAWeb/vite.config.ts
> error during build:
> Error: The package "@esbuild/linux-x64" could not be found, and is needed by esbuild.
> ```
> 
> I've done the full composer/npm rebuild per `.hooks/post-merge`

I believe this could be caused by our version of Vite also being outdated. I have bumped Vite to the latest minor release (which caused a ton of lockfile changes, but also made some adjustments to the subdependency you're seeing an issue with).

Please let me know if the issue still persists. If it does, I'll go with a manual CSS approach.

Jamiras:
> Please let me know if the issue still persists. If it does, I'll go with a manual CSS approach.

It's happy now.

Jamiras (approved):
Leaving for now in case @luchaos has any concerns about the package changes

luchaos (approved):

PR #1523, opened 2023-04-27
Small adjustments for missable `[m]` tags

EmoonX:
Two small adjustments:
- [m] tags now aren't bold nor too wide (as per some criticism back then; were particularly not fine on Windows)
- Moves [m] tag always to the end of achievement's title, preceded by a space. This in line with what we're currently doing with tags overall and also as a way to standardize display so it's not bound to devs

Before / after:
image
image


app_legacy/Helpers/render/achievement.php
```diff
@@ -69,16 +69,18 @@ function renderAchievementTitle(?string $title, bool $tags = true): string
...
+    $missableTag = '';
     if ($tags) {
-        $span = '<span class=\'tag missable\' title=\'Missable\'><abbr>[<b>m</b>]</abbr></span>';
+        $missableTag = "<span class='tag missable' title='Missable'><abbr>[<span>m</span>]</abbr></span>";
     }
+    $title = str_replace('[m]', '', $title);
+    $title = str_replace('  ', ' ', $title);
 
-    return trim(str_replace('[m]', $span, $title));
+    return "$title $missableTag";
```
Jamiras, line 83: The space should be part of `$missableTag`. It won't be necessary for most achievements.

app_legacy/Helpers/render/achievement.php
```diff
@@ -69,16 +69,18 @@ function renderAchievementTitle(?string $title, bool $tags = true): string
...
     }
-    $span = '';
+
+    $missableTag = '';
     if ($tags) {
-        $span = '<span class=\'tag missable\' title=\'Missable\'><abbr>[<b>m</b>]</abbr></span>';
+        $missableTag = "<span class='tag missable' title='Missable'><abbr>[<span>m</span>]</abbr></span>";
     }
+    $title = str_replace('[m]', '', $title);
+    $title = str_replace('  ', ' ', $title);
```
Jamiras, line 81: Why is this necessary? The browser should automatically collapse whitespace for us, so we can offload that processing to the client.

Does this also collapse three or more spaces? My local DB only has ~500/280k achievements with at least two spaces next to each other, 90 with at least three, and 15 with four!

EmoonX: Good point. I think my intention was merely to collapse the extra spaces resulting from misplaced `[m]`s. But indeed, browsers ignore any extra, preceding or succeeding whitespace, so we don't have to bother.

Jamiras (approved):

PR #1526, opened 2023-04-27
feat(achievement): allow jr devs to embed videos

wescopeland:
This PR modifies the embed input on the achievement page's devbox to allow Jr Devs to add videos to their achievements. Currently, this is a dev-only feature. However, it is desirable for Jr Devs to be able to do this under the following criteria:

* The achievement is not Official/Core.
* The Jr Dev is the author of the achievement.
* The update is logged in the comments.

ref: https://discord.com/channels/476211979464343552/1002688256309211257/1101263124331515944

Summary of changes:
* Logging is now enabled for embeds. A different logging message is produced for adding an embed, changing an embed, and deleting an embed.
* Jr Devs can now add embeds to achievements based on the above criteria.

Screenshot 2023-04-27 at 7 19 55 PM


public/request/achievement/update-video.php
```diff
@@ -2,18 +2,52 @@
...
 ]);
 
-if (updateAchievementEmbedVideo((int) $input['achievement'], $input['video'])) {
+$achievementId = (int) $input['achievement'];
+$embedUrl = $input['video'];
+
+$achievement = Achievement::find($achievementId);
+if (!$achievement) {
+    abort(400);
+}
```
Jamiras, line 25: Missing record should be 404. Also, doesn't the `exists` validation check ensure this will never happen?

wescopeland: Good point, resolved in https://github.com/RetroAchievements/RAWeb/pull/1526/commits/7f73a9c84c80d856ead8fa30952ada086a51f44e.

public/request/achievement/update-video.php
```diff
@@ -2,18 +2,52 @@
...
+    ($user != $achievement['Author'] || $achievement['Flags'] == AchievementType::OfficialCore)
+) {
+    abort(401);
+}
+
+function buildAuditLogMessage(string $user, ?string $currentVideoUrl, ?string $embedUrl): string {
+    if ($currentVideoUrl && $embedUrl) {
+        return "$user changed this achievement's embed URL from $currentVideoUrl to $embedUrl.";
+    } elseif ($currentVideoUrl && !$embedUrl) {
+        return "$user changed this achievement's embed URL from $currentVideoUrl to no embed URL.";
```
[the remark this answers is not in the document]
wescopeland: That is an astute observation, I've cleaned this up in https://github.com/RetroAchievements/RAWeb/pull/1526/commits/7f73a9c84c80d856ead8fa30952ada086a51f44e.

public/request/achievement/update-video.php
```diff
@@ -2,18 +2,41 @@
...
+    ($user != $achievement['Author'] || $achievement['Flags'] == AchievementType::OfficialCore)
+) {
+    abort(401);
+}
+
+if (updateAchievementEmbedVideo($achievementId, $embedUrl)) {
+    $embedUrlText = $embedUrl === null ? 'null' : $embedUrl;
+    $auditLog = "$user set this achievement's embed URL to $embedUrlText.";
+
+    addArticleComment('Server', ArticleType::Achievement, $achievementId, $auditLog, $user);
```
Jamiras, line 38: I'm going to leave this open for @luchaos to review. The achievement audit log is visible to all, even when not logged in. Having old URLs visible might be undesired if we explicitly censure a video for whatever reason (linked video show some form of cheating, or an unintended glitch to make it easier, or something similar). 

Given that it's a dev posting the video, the level of trust might be acceptable.

Maybe we could modify the comment widget to remove URLs from system comments for non-admin users?

public/request/achievement/update-video.php
```diff
@@ -1,19 +1,42 @@
...
 use Illuminate\Support\Facades\Validator;
 
-if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::Developer)) {
+if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::JuniorDeveloper)) {
     abort(401);
 }
 
 $input = Validator::validate(Arr::wrap(request()->post()), [
-    'achievement' => 'required|integer',
+    'achievement' => 'required|integer|exists:mysql_legacy.Achievements,ID',
```
luchaos, line 15:
```suggestion
    'achievement' => 'required|integer|exists:mysql.Achievements,ID',
```



ℹ️ The `mysql_legacy` db connection has been replaced by the default `mysql` connection.
When trying to embed a video it will result in `Database connection [mysql_legacy] not configured`.

public/request/achievement/update-video.php
luchaos: I too think that the outdated URL should not be present in the audit message publicly.

We don't include values for other audit messages either as far as I'm aware?

ℹ️ A new audit system will be introduced at some point which will include all the changes with values but gated behind certain roles.

luchaos (changes requested):

public/request/achievement/update-video.php
wescopeland: Thanks, I would've missed this. Resolved in https://github.com/RetroAchievements/RAWeb/pull/1526/commits/f1c62c77d3af1025004dcdb0a0634c7de7c8f5ef.

public/request/achievement/update-video.php
wescopeland: Re-reading the requirements from QA and DevC, it sounds like it's indeed safe to omit these links and keep the text just saying the embed URL was updated.

I've made this adjustment in https://github.com/RetroAchievements/RAWeb/pull/1526/commits/f1c62c77d3af1025004dcdb0a0634c7de7c8f5ef.

Screenshot 2023-07-15 at 8 25 39 AM

luchaos (approved):

PR #1528, opened 2023-04-28
always run all analysis tools before pushing

Jamiras:
When trying to push from master, find the _server_ master branch for detecting PHP modifications


luchaos:
As discussed, to keep it simple, I'd go for something like this:
```
#!/bin/sh
# exit on error
set -e

printf "\n⏳ composer lint\n"
PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --verbose --dry-run

vendor/bin/phpstan analyse --memory-limit 512M

printf "\n⏳ composer deptrac\n"
vendor/bin/deptrac analyse

printf "\n⏳ composer test\n"
vendor/bin/phpunit

printf "\n⏳ npm run lint:eslint\n"
#npx eslint --quiet ${modifiedJsFiles}
npx eslint --quiet

#printf "\n⏳ npm run lint:stylelint\n"
#npx stylelint --syntax scss ${modifiedCssFiles}

printf "\n✅ pre-push OK\n\n"
```

wescopeland:
It may also be worth executing `npm run test:ci` as it is now part of the GitHub Actions pipeline.

luchaos:
> It may also be worth executing `npm run test:ci` as it is now part of the GitHub Actions pipeline.

That said, I wanted to ask: what's the difference between `vitest` and `vitest run`?
I usually like to drop the `:ci` suffix from any tasks in composer.json and package.json as I feel that there shouldn't be any task that's only specific to CI pipelines, they all have to work locally, too imo.

wescopeland:
> That said, I wanted to ask: what's the difference between `vitest` and `vitest run`? I usually like to drop the `:ci` suffix from any tasks in composer.json and package.json as I feel that there shouldn't be any task that's only specific to CI pipelines, they all have to work locally, too imo.

`vitest run` forcibly disables watch mode. However, [per the docs](https://vitest.dev/guide/cli.html#vitest), it seems `vitest` now has the capability of detecting whether or not it is running in a CI context. It is possible we can totally remove `test:ci`.

luchaos:
> > That said, I wanted to ask: what's the difference between `vitest` and `vitest run`? I usually like to drop the `:ci` suffix from any tasks in composer.json and package.json as I feel that there shouldn't be any task that's only specific to CI pipelines, they all have to work locally, too imo.
> 
> `vitest run` forcibly disables watch mode. However, [per the docs](https://vitest.dev/guide/cli.html#vitest), it seems `vitest` now has the capability of detecting whether or not it is running in a CI context. It is possible we can totally remove `test:ci`.

Sounds good. For watch commands I'd expect that one to be explicit, like `composer test:watch` or `npm run test:watch`. The default would be be I'd assume to be run in a pipeline, like `composer test` and `npm run test`

wescopeland:
> Sounds good. For watch commands I'd expect those to be explicit, like `composer test:watch` or `npm run test:watch`. The default would be be I'd assume to be run in a pipeline, like `composer test` and `npm run test`

That is very reasonable - addressed in a tiny PR here https://github.com/RetroAchievements/RAWeb/pull/1535

Jamiras:
I've made the suggested changes to always run all of the existing analysis tools.

I'm unclear which additional command(s) are being asked for. `npm run test`, `npm run test:ci`, `vitest run`?

wescopeland:
`npm run test` should be sufficient now that https://github.com/RetroAchievements/RAWeb/pull/1535 has been merged.

luchaos (approved):

.hooks/pre-push
```diff
@@ -2,31 +2,22 @@
...
+printf "\n⏳ composer analyse\n"
+vendor/bin/phpstan analyse --memory-limit 512M
 
-    printf "\n⏳ composer deptrac\n"
-    vendor/bin/deptrac analyse
+printf "\n⏳ composer deptrac\n"
+vendor/bin/deptrac analyse
 
-    printf "\n⏳ composer test\n"
-    vendor/bin/phpunit
```
luchaos, line 17: oh my, just relaized that the phpunit step will be missing 😅

luchaos: can you please add it back?

luchaos: nevermind, noticed it was added at the end, all good 🤦

Jamiras: It's there. I just moved it to the end. It's the slowest step, so that lets everything else have a chance to fail quickly first.

luchaos: perfect, thanks!

PR #1529, opened 2023-04-28
Icons for claim set type (New/Revision)

EmoonX:
Icons for distinction between "New" and "Revision" claim set types.

image


resources/views/community/components/claims/finished-claim-table-row.blade.php
```diff
@@ -5,7 +5,9 @@
 
 $claimSetTypeCopy = ClaimSetType::toString(ClaimSetType::NewSet);
 if ($claim['SetType'] !== ClaimSetType::NewSet) {
-    $claimSetTypeCopy = ClaimSetType::toString(ClaimSetType::Revision);
+    $claimSetTypeCopy = '🗒️ ' . ClaimSetType::toString(ClaimSetType::Revision);
+} else {
+    $claimSetTypeCopy = "💡 <b>$claimSetTypeCopy</b>";
```
wescopeland, line 10:
```suggestion
    $claimSetTypeCopy = "💡 <span class='font-bold'>$claimSetTypeCopy</span>";
```

`<span>` instead of `<b>` so we are at the mercy of Tailwind's defaults, not the browser's. It is also slightly better for accessibility.

But I'm not sure we should be highlighting new sets in this way. There are some incredible revisions being released nowadays, and bolding New but not bolding Revision may send the wrong message.

public/claimlist.php
```diff
@@ -192,7 +192,11 @@
             echo gameAvatar($claim);
             echo "</td>";
             echo "<td>" . ($claim['ClaimType'] == ClaimType::Primary ? ClaimType::toString(ClaimType::Primary) : ClaimType::toString(ClaimType::Collaboration)) . "</td>";
-            echo "<td>" . ($claim['SetType'] == ClaimSetType::NewSet ? ClaimSetType::toString(ClaimSetType::NewSet) : ClaimSetType::toString(ClaimSetType::Revision)) . "</td>";
+            echo "<td class='text-2xs whitespace-nowrap [word-spacing:.2em]'>"
+                . ($claim['SetType'] == ClaimSetType::NewSet ?
+                    ('💡 <b>' . ClaimSetType::toString(ClaimSetType::NewSet) . '</b>') :
```
wescopeland, line 197: Similar to my other comment, if we are indeed going to make these bold we should use `<span>` tags for better flexibility and accessibility, however I'm not sure this is the message we want to send from a UX POV.

resources/views/community/components/claims/finished-claim-table-row.blade.php
wescopeland, line 10: In https://github.com/RetroAchievements/RAWeb/pull/1503 a lightbulb emoji is used to convey sets that have received revisions with unearned achievements. This lightbulb emoji may conflict with the intent of that one if both changes are merged.

Jamiras:
I'm generally against this change for many of the same reasons I don't like emojis in rich presence.

Emoji meanings are generally not widely known, each browser renders them differently, and it adds an extra step for my brain to have to translate while reading. "New" or "Revision" are dissimilar enough that I don't need any additional information to help me understand what they are.

In this case, you're including the text with the emoji instead of just outright replacing it, so the translation is provided in context, but then why have the emoji in the first place? And what makes revision so special? Shouldn't we then also have emojis for Primary/Collaboration? 

Emojis (and images in general) draw the user's attention. Why does the user's attention need to be drawn to that column?

EmoonX:
The emoji part was indeed an outdated slip of my mind. I'm already in the process of changing those for minimalist icons (thanks @wescopeland for the suggestion).

As for the change itself... A lot of pages and tables are just overflowing with text and more text; one of the aims is to change that a bit. It's inoffensive and can make things look nicer. I'm not saying I'm in a crusade to inject icons everywhere in the site. It's just that the **New** vs **Revision** distinction can be considered well-established and important enough to be worthy of such a distinction.

Furthermore, let's not forget the `New Sets/Revisions` is kind of a concession: two tables into one, new sets mixed with revisions... Having a visual feedback to help users identifying what type of row they're looking at sounds like a positive point for the update in my book.

EmoonX:
I've updated the first post with the new design.

Emoji have been replaced with icons (from _pixelarticons_). No more bold on New either.

PR #1530, opened 2023-04-29
feat(game): have subsets inherit the player count of their parent games

wescopeland:
A choice has been made that subsets should inherit the player counts of their parent games. This PR is an implementation of that choice.

**Current Behavior**
A subset's player count is based exclusively on the unique number of players who have played the subset.

**New Behavior**
A subset's player count is derived from the combined count of distinct players who have played both the base set and the subset. Duplicate players are only counted once.

**End-user UX Concerns**
White points on subsets will, for some time, be tremendously inflated. Super Mario World [Subset - Bonus] approaches 460,000 white points after this change. I have been informed that the opinion is it's better for white points to be inflated and subsets to get traffic+interest, as opposed to the present situation.

Screenshot 2023-04-28 at 1 56 24 PM

Total player counts are shown in the achievement list based on a plurality of community feedback [from here](https://discord.com/channels/310192285306454017/511718348178849792/1095688899071840296).

---

**Before**
Screenshot 2023-04-28 at 9 25 37 PM

**After**
Screenshot 2023-04-28 at 9 25 57 PM


app_legacy/Helpers/database/game.php
```diff
@@ -137,24 +150,34 @@ function getGameMetadata(
     }
 
     if ($metrics) {
```
Tsearo, line 152: It looks like you are are calling the same query twice and doing a union on game and parent but you could just do a single query `where GameID IN (gameID, parentGameID)` though I don't see any performance difference it is mainly just to save some space. Below is how I was able to run it by making a conditional check for the GameID check
```
    if ($metrics) {
        $parentGameId = getParentGameIdFromGameTitle($gameDataOut['Title']);

        $bindings = [
            'gameId' => $gameID,
            'parentGameId' => $parentGameId,
            'achievementType' => $flags,
        ];

    $query = "SELECT aw.HardcoreMode, COUNT(DISTINCT aw.User) as Users
            FROM (
              SELECT awb.User, awb.HardcoreMode
              FROM Awarded AS awb
              LEFT JOIN Achievements AS ach ON ach.ID = awb.AchievementID
              LEFT JOIN UserAccounts AS uab ON uab.User = awb.User
              WHERE ach.GameID IN (:gameId , :parentGameId) AND ach.Flags = :achievementType
              AND (NOT uab.Untracked)
            ) AS aw
            GROUP BY aw.HardcoreMode";

        $gameMetaData = legacyDbFetchAll($query, $bindings);

        $numDistinctPlayersCasual = 0;
        $numDistinctPlayersHardcore = 0;
        foreach ($gameMetaData as $data) {
            if ($data['HardcoreMode'] == UnlockMode::Hardcore) {
                $numDistinctPlayersHardcore = $data['Users'];
            } else {
                $numDistinctPlayersCasual = $data['Users'];
            }
        }

        $gameDataOut['ParentGameID'] = $parentGameId;
        $gameDataOut['NumDistinctPlayersCasual'] = $numDistinctPlayersCasual;
        $gameDataOut['NumDistinctPlayersHardcore'] = $numDistinctPlayersHardcore;
    }

    $gameDataOut['NumAchievements'] = $numAchievements;
```

app_legacy/Helpers/database/player-game.php
```diff
@@ -412,38 +412,65 @@ function getUsersCompletedGamesAndMax(string $user): array
...
     if ($requestedBy) {
         $bindings['requestedBy'] = $requestedBy;
-        $requestedByStatement = 'OR ua.User = :requestedBy';
+        $requestedByStatement = 'OR uah.User = :requestedBy';
+    }
+
+    $parentGameIdStatement = '';
+    $parentUnlockModeStatement = '';
+    $parentAchievementTypeStatement = '';
+    if ($parentGameID !== null) {
```
Tsearo, line 442: I believe you can do the same kind of cleanup here not needing much of the extra parent variables or the union if you do `where gameid in (gameid, parentid)`

Tsearo (commented):
Using the example you posted:
image

The achievement page doesn't reflect these new player counts unless that is deliberate though it could cause some confusion if so
image

Potentially could show both the combined and subset only player counts though I'm not sure a good way to do that intuitively but it might be nice to retain that info

app_legacy/Helpers/database/game.php
wescopeland: Excellent feedback, this is much more concise. Addressed in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/f367e723499103b84452ce04e5087506bb829f7c

app_legacy/Helpers/database/player-game.php
wescopeland: Thanks to your suggestion I was able to trim this down significantly in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/f367e723499103b84452ce04e5087506bb829f7c.

I am working through the achievement page issue now.

wescopeland:
> The achievement page doesn't reflect these new player counts unless that is deliberate though it could cause some confusion if so

An interesting discovery I've just made is these player counts are already incorrect, even on prod. This is surely a bug.

For example: https://retroachievements.org/game/373

**Game Page**
Screenshot 2023-04-29 at 8 30 57 AM

**Achievement Page**
Screenshot 2023-04-29 at 8 31 22 AM

I believe this is because the player counts are calculated using a different function on the two pages. The game page uses `NumDistinctPlayersCasual` from `getGameMetadata()`. The achievement page uses `getTotalUniquePlayers()`, which allows for advanced filtering options, but the unfiltered result is lower than what `getGameMetadata()` is returning which leads me to be inclined to believe that's where the bug lies.

wescopeland:
I have resolved the player counts issue for base sets and subsets in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/29918a8256c3b5f502b2fcbe1ad17cbc2d20f472.

The updated tests should also help prevent it from resurfacing.

app_legacy/Helpers/database/player-game.php
```diff
@@ -436,13 +436,19 @@ function getTotalUniquePlayers(int $gameID, ?string $requestedBy = null, bool $h
         $requestedByStatement = 'OR ua.User = :requestedBy';
     }
 
+    if ($parentGameID !== null) {
+        $bindings['parentGameId'] = $parentGameID;
+    }
+
+    $gameIdStatement = $parentGameID === null ? 'ach.GameID = :gameId' : 'ach.GameID IN (:gameId, :parentGameId)';
```
Tsearo, line 443: This check is much better than my above suggestion in game.php for `WHERE ach.GameID IN (:gameId , :parentGameId) AND ach.Flags = :achievementType` so it should probably be applied there as well.

app_legacy/Helpers/database/player-game.php
```diff
@@ -412,22 +412,22 @@ function getUsersCompletedGamesAndMax(string $user): array
...
 }
 
-function getTotalUniquePlayers(int $gameID, ?string $requestedBy = null, bool $hardcoreOnly = false, ?int $achievementType = null): int
+function getTotalUniquePlayers(int $gameID, ?int $parentGameID = null, ?string $requestedBy = null, bool $hardcoreOnly = false, ?int $achievementType = null): int
 {
     $bindings = [
         'gameId' => $gameID,
     ];
 
-    $hardcoreStatement = '';
```
Tsearo, line 421: I'm not sure these all need to be renamed to `base` since the mode/flags are not exclusive to just the base set when checking a subset

app_legacy/Helpers/database/player-game.php
wescopeland: I've massaged a similar check into _game.php_. Unfortunately, I was not able to do a one-liner due to a potential parameters count mismatch in `$bindings`. Addressed in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/e5b5a5c1e1bffe9a4fe9f3ff80f811756dc66590.

app_legacy/Helpers/database/player-game.php
wescopeland: You're right, it is certainly redundant now. Adjusted in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/e5b5a5c1e1bffe9a4fe9f3ff80f811756dc66590.

Tsearo (approved):

tests/Feature/Api/V1/AchievementOfTheWeekTest.php
```diff
@@ -37,7 +38,7 @@ public function testGetAchievementOfTheWeek(): void
         /** @var Game $game */
         $game = Game::factory()->create(['ConsoleID' => $system->ID]);
         /** @var Achievement $achievement */
-        $achievement = Achievement::factory()->create(['GameID' => $game->ID]);
+        $achievement = Achievement::factory()->create(['GameID' => $game->ID, 'Flags' => AchievementType::OfficialCore]);
```
Jamiras, line 41: There's a `published` scope for this.
```
$achievement = Achievement::factory()->published()->create(['GameID' => $game->ID]);
```

public/gameInfo.php
```diff
@@ -787,7 +788,8 @@ function ResetProgress() {
                     if ($permissions >= Permissions::Developer) {
                         echo "<form action='/request/game/recalculate-points-ratio.php' method='post'>";
                         echo csrf_field();
-                        echo "<input type='hidden' name='game' value='$gameID'>";
+                        echo "<input type='hidden' name='gameId' value='$gameID'>";
+                        echo "<input type='hidden' name='parentGameId' value='$parentGameID'>";
```
Jamiras, line 792: I don't think it's worth it to pass parent ID all the way through. Just look it up again in the `UpdateGameWeightedPoints`. That will also handle the case where the scheduled task recalculates the game.

https://github.com/RetroAchievements/RAWeb/blob/8360cffc54dfafd1c9500268803034812bfc0d6a/app_legacy/Platform/Commands/UpdateGameWeightedPoints.php#L36

It also eliminates a possible issue where the URL could be called directly with a mismatched `parentGameId`.

app_legacy/Helpers/database/player-game.php
Jamiras, line 443: suggest merging this into the above if:
```
$gameIdStatement = 'ach.GameID = :gameId';
if ($parentGameID !== null) {
    $gameIdStatement = 'ach.GameID IN (:gameId, :parentGameId)';
    $bindings['parentGameId'] = $parentGameID;
}

File: public/API/API_GetGameExtended.php [hunk 7]
```diff
@@ -8,6 +8,7 @@
  *  string     Title                      name of the game
  *  int        ConsoleID                  unique identifier of the console associated to the game
  *  string     ConsoleName                name of the console associated to the game
+ *  string     ParentGameID               unique identifier of the parent game if this is a subset
```
Jamiras, line 11: shouldn't this be `int?`

app_legacy/Helpers/database/game.php
```diff
@@ -148,13 +163,22 @@ function getGameMetadata(
...
-              GROUP BY aw.HardcoreMode";
+                FROM (
+                  SELECT aw.User, aw.HardcoreMode
+                  FROM Awarded AS aw
+                  LEFT JOIN Achievements AS ach ON ach.ID = aw.AchievementID
+                  LEFT JOIN UserAccounts AS ua ON ua.User = aw.User
+                  WHERE $gameIdStatement AND ach.Flags = :achievementType
+                  AND (NOT ua.Untracked $requestedByStatement)
+                ) AS aw
+                GROUP BY aw.HardcoreMode";
```
Jamiras, line 181: There's something happening with the DW3(GBC) set:

Parent:
image

Subset:
image

Without digging into the data, I assume this is because the relationship is not reciprocal. A player who has only played the subset won't show up in the total number of players for the parent game query.

There probably isn't a good solution for this, as the parent would have to count players from all subsets, and that information would have to be propagated to all subsets.

tests/Feature/Api/V1/AchievementOfTheWeekTest.php
wescopeland: Addressed in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/c384be10eec370633b6d37d8f2c33714547b872f

public/gameInfo.php
wescopeland: Good points, I've addressed this in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/c384be10eec370633b6d37d8f2c33714547b872f. The `run()` function now tries to detect if the given game ID has a parent.

app_legacy/Helpers/database/player-game.php
wescopeland: Done in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/c384be10eec370633b6d37d8f2c33714547b872f

public/API/API_GetGameExtended.php
wescopeland: Yes, it should be. Fixed in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/c384be10eec370633b6d37d8f2c33714547b872f.

app_legacy/Helpers/database/game.php
wescopeland: You are correct, the relationship is indeed not reciprocal. I came to the same conclusion that there wouldn't be a good solution for this.

tests/Feature/Api/V1Test.php
```diff
@@ -255,7 +255,7 @@ public function testGetAchievementUnlocks(): void
         /** @var Game $game */
         $game = Game::factory()->create(['ConsoleID' => $system->ID]);
         /** @var Achievement $achievement */
-        $achievement = Achievement::factory()->create(['GameID' => $game->ID]);
+        $achievement = Achievement::factory()->create(['GameID' => $game->ID, 'Flags' => AchievementType::OfficialCore]);
```
Jamiras, line 258: two more places where `published()` should be used.

wescopeland: Yep, good catch. Addressed in https://github.com/RetroAchievements/RAWeb/pull/1530/commits/11fa9a2f079927caf2caf08407b924813a66754c

Jamiras (approved):

PR #1536, opened 2023-04-30
fix(home): dont show duplicate user+game pairs in new sets/revisions

wescopeland:
This PR makes an adjustment to how the list of New Sets/Revisions are built on the home page.

Currently, it is possible for the same user+game pair to appear N times in the list. After this change, the "oldest" of these user+game pair duplicates will be the winner and the others will be ignored.

**Before**
Screenshot 2023-04-30 at 5 18 50 PM

**After**
Screenshot 2023-04-30 at 5 19 09 PM


Jamiras (approved):
This feels like something that should be done at the query level, but it's probably not that common to worry about.

If the two duplicate records were the last visible record and the first offscreen record, only the newer information would be presented.

PR #1538, opened 2023-04-30
add startsession dorequest api

Jamiras:
The existing "start session" API is just a call that posts a StartedPlaying entry into the Activity table (via `dorequest?r=postactivity`). It's always immediately followed with one or two calls to fetch the user's unlocks for the game (via `dorequest?r=unlocks`).

This creates a new `dorequest?r=startsession` API that combines all three of those server calls into one. It posts the StartedPlaying entry into the Activity table and returns both the hardcore and softcore unlocks that the user has achieved in the game.

Tests have been added for all three dorequest endpoints (postactivity, unlocks, and startsession).

This implements #821.


app/Helpers/database/user-activity.php
```diff
@@ -177,87 +161,46 @@ function postActivity(string $userIn, int $type, ?int $data = null, ?int $data2
...
-    }
-
-    return legacyDbStatement(
-        'UPDATE UserAccounts AS ua SET ua.LastLogin = NOW() WHERE ua.User = :username',
-        ['username' => $username]
-    );
-}
-
-function UpdateUserRichPresence(string $user, int $gameID, string $presenceMsg): bool
+function UpdateUserRichPresence(User &$user, int $gameID, string $presenceMsg): void
```
luchaos, line 199: I'd go with returning the user model after saving the changes instead of pass-by-reference.
Saving here would be the expected side effect going by `update` in the function name.

public/dorequest.php
```diff
@@ -30,12 +34,14 @@
     $validLogin = authenticateFromAppToken($user, $token, $permissions);
 }
 
-function DoRequestError(string $error): JsonResponse
-{
-    return response()->json([
-        'Success' => false,
-        'Error' => $error,
-    ]);
+if (!function_exists('DoRequestError')) {
```
luchaos, line 37: Out of curiosity - did we ever run into this function being defined more than once?

public/dorequest.php
```diff
@@ -287,8 +325,15 @@ function DoRequestError(string $error): JsonResponse
         break;
 
     case "unlocks":
-        $hardcoreMode = (bool) request()->input('h', 0);
-        $response['UserUnlocks'] = GetUserUnlocksData($user, $gameID, $hardcoreMode);
+        $hardcoreMode = request()->input('h', 0) == UnlockMode::Hardcore;
```
luchaos, line 328: To default to type safe comparison this could be
```php
(int) request()->input('h', 0) === UnlockMode::Hardcore
```

luchaos (approved):

app/Helpers/database/user-activity.php
Jamiras: This isn't actually saving. The save occurs in the two places that call it as they make additional changes to the User object.

PHP documentation indicates that all objects are already passed by reference, so the & is redundant. I'll remove it.

public/dorequest.php
Jamiras: I added this logic as part of this PR. The unit tests fail without it:
```
PHP Fatal error:  Cannot redeclare DoRequestError() (previously declared in
/home/brian/source/RAWeb/public/dorequest.php:38) in /home/brian/source/RAWeb/public/dorequest.php on line 38
```

I couldn't figure out where the extra reference is coming from, so I just wrapped it in this logic.

luchaos: Might have to do how the tests are executed in combination with the include logic for script files in `HandlesPublicFileRequests`.
We can ignore that for now as those includes shall be rewritten to controller actions at some point.

luchaos (approved):

PR #1539, opened 2023-05-01
fix(reorderSiteAwards): disable invisible manual move buttons

wescopeland:
This PR addresses a bug on the Reorder Site Awards page. 

When a row is checked to be hidden, the buttons are given an opacity of 0 but are not disabled, so they are still clickable. The changes in this PR disable them when they are hidden, and reenable them when they are unhidden.


wescopeland:
Superseded by https://github.com/RetroAchievements/RAWeb/pull/1543.

Issue #1522, opened 2023-04-27
Global Ranking Achievement Count includes Unofficial and Event Achievements


Tsearo:
Easy example showcasing both issues where the user has 152 achievements but because of doing the stimulus package event worth 1000 achievements even though it was demoted to unofficial and is an event they count towards the total in the ranking
![image](https://user-images.githubusercontent.com/4047771/234739449-365d2d7b-902d-40d7-aeec-56dfc20dbfd9.png)
![image](https://user-images.githubusercontent.com/4047771/234739541-b3b5d7cd-4909-4527-a3ff-8823eb4e59df.png)

A fix to this would be to add checks for only core achievements and exclude the event ID like so but doing so tripled the time to run the query so either a more optimized query is needed or we will need to wait for denormalization to correct this issue

Current query for achievement count
```
(SELECT COALESCE(SUM(CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END), 0) 
                        FROM Awarded AS aw 
                        WHERE aw.User = ua.User) AS AchievementCount,
```

Accurate query (~3x slower)
```
(SELECT COALESCE(SUM(CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END), 0) 
                        FROM Awarded AS aw 
                        LEFT JOIN Achievements ach ON ach.ID = aw.AchievementID 
                        LEFT JOIN GameData AS gd ON gd.ID = ach.GameID
                        WHERE aw.User = ua.User
                        AND ach.Flags = 3
                        AND gd.ConsoleID <> 101) AS AchievementCount,
```

wescopeland:
Resolved by https://github.com/RetroAchievements/RAWeb/pull/1617.

PR #1540, opened 2023-05-01
Add earned achievement totals to user profile

Tsearo:
Adds the users total achievement totals for hardcore/softcore
image

These totals will not match what shows in the Global Ranking counts however as that counts achievements in Unofficial and from Events currently and it is not very feasible to correct those counts without heavily increasing the length of the queries as seen in #1522


app_legacy/Helpers/database/player-rank.php
```diff
@@ -59,6 +59,33 @@ function recalculatePlayerPoints(string $user): bool
...
+
+    $query = "SELECT User, 
+                SUM(CASE WHEN aw.HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END) AS RAAchievements,
+                SUM(CASE WHEN aw.HardcoreMode = " . UnlockMode::Softcore . " THEN 1 ELSE 0 END) AS TotalAchievements
+                FROM Awarded AS aw
+                LEFT JOIN Achievements AS ach ON ach.ID = aw.AchievementID
+                LEFT JOIN GameData AS gd ON gd.ID = ach.GameID
+                WHERE aw.User =:username 
+                AND ach.Flags = " . AchievementType::OfficialCore . "
+                AND gd.ConsoleID NOT IN (100, 101)";
```
Jamiras, line 76: The first time I tried to load Maxmilyn's page on my dev server, this query took 2.4 seconds. Repeated refreshing dropped it to ~300ms.

Can this feature wait for the denormalized data? If an absolute total isn't available, it should still be faster to sum per-game totals.

Tsearo: Oh my I didn't realize how slow it was on the initial run, this led me to realize there was a much simpler way to just increment the count in the existing loop through the game data pulled for the completion widget so it shouldn't impact performance at all now

Jamiras (approved):

PR #1542, opened 2023-05-01
DeveloperStats - Replace Ticket Ratio and add links

Tsearo:
Some improvements to the Developer Stats page:

New
- Show the total number of filtered developers (can see how many active or jr devs there are now)
- Replace Ticket Ratio (arguably useless stat) with Tickets Resolved For Others to be able to see and sort on devs that help out the most with tickets

Links added
- Open Tickets - Self explanatory
- Achievements - Links the the developers sets page (more usable than developers achievements list page)
- Tickets Resolved For Others - Ticket Manager filtered on tickets the dev resolved for others
- Active Claims - Self explanatory

- I wanted to add a link to the developers individual stats page somewhere but wasn't sure where as the name and pfp wouldn't be intuitive and I didn't feel comfortable adding a new column just for that

Before:
image

After:
image

Minor Caveats:
- Open Tickets - Count includes those in unofficial but when viewing open tickets you can only see Official or Unofficial exclusively so the count may seem misleading. Unsure if the count should include those in unofficial or not
- Tickets Resolved For Others - The count matches what is displayed on the IndividualDevStats page which does not include tickets that you are the Reporter of. There is no way to filter on this in the Ticket Manager so the counts shown there can be higher. Unsure if tickets you reported and resolved should count towards the total as that would be too open for abuse and is likely why it isn't counted


public/developerstats.php
```diff
@@ -57,14 +72,13 @@
...
-        echo "<th class='text-right'><a href='/developerstats.php?t=1&f=$devFilter' title='Points gained by others through achievement unlocks'>Yielded Points</a>" . ($type == 1 ? "*" : "") . "</th>";
-        echo "<th class='text-right'><a href='/developerstats.php?t=7&f=$devFilter' title='Set claims currently active'>Active Claims</a>" . ($type == 7 ? "*" : "") . "</th>";
-        // echo "<th class='text-right whitespace-nowrap'><a href='/developerstats.php?t=5'>Last Login</a>" . ($type == AchievementType::UNOFFICIAL ? "*" : "") . "</th>";
+        echo "<th><a href='/developerstats.php?t=6&f=$devFilter'>Name</a>$orderedByName</th>";
+        echo "<th class='text-right whitespace-nowrap'><a href='/developerstats.php?t=3&f=$devFilter'>Open Tickets</a>$orderedByOpenTickets</th>";
+        echo "<th class='text-right whitespace-nowrap'><a href='/developerstats.php?f=$devFilter'>Achievements</a>$orderedByAchievements</th>";
+        echo "<th class='text-right whitespace-nowrap'><a href='/developerstats.php?t=4&f=$devFilter' title='Ticket Resolved for Others'>Tickets Resolved For Others</a>$orderedByResolvedTickets</th>";
```
wescopeland, line 78: Wondering if there are any other headings we can use (though keep `title` as the same). This heading (and all its cells) seems to stick out from all the others because of its character count.

Possible alternatives:

Assisted Resolves
Collab Ticket Fixes
External Ticket Fixes
Helped Tickets

Tsearo: Yeah I wasn't happy with the heading being so long but couldn't think of a clear short alternative. 
I still want to call it `Ticket Karma` but that used to be a thing previously on the individual stats page and was seen negatively thus removed so I think when I asked a long time ago it was advised against but maybe now it would be okay

Jamiras: What about putting a newline/<br> in the middle of it?

app_legacy/Helpers/database/user.php
```diff
@@ -379,6 +379,17 @@ function GetDeveloperStatsFull(int $count, int $sortBy, int $devFilter = 7): arr
...
+        SELECT ua2.User,
+        SUM(CASE WHEN t.ReportState = 2 THEN 1 ELSE 0 END) AS total
+        FROM Ticket AS t
+        LEFT JOIN UserAccounts as ua ON ua.ID = t.ReportedByUserID
+        LEFT JOIN UserAccounts as ua2 ON ua2.ID = t.ResolvedByUserID
+        LEFT JOIN Achievements as a ON a.ID = t.AchievementID
+        WHERE ua.User NOT LIKE ua2.User
+        AND a.Author NOT LIKE ua2.User
+        AND a.Flags = '3' 
+        GROUP BY ua2.User) resolved ON resolved.User = ua.User
```
Jamiras, line 392: Not sure how much it matters, but this query doesn't result in the same count as the query performed on the linked page.

For example, in my DB, MGNS8M shows 43 in the developer stats page, but 79 when clicked.

public/developerstats.php
wescopeland: Screenshot 2023-05-05 at 12 54 21 PM

max-width of 120px and the whitespace-nowrap classname removed.

app_legacy/Helpers/database/user.php
Tsearo: I listed that as a caveat since the existing individual stats metric excludes tickets you report and the ticket manager didn't have a filter for that but I went ahead and added the filter now since it probably should be added if I'm going to link to to the ticket manager for that

app_legacy/Community/Enums/TicketFilters.php
```diff
@@ -42,18 +42,21 @@ abstract class TicketFilters
 
     public const ResolvedByNonAuthor = 1 << 17;
 
-    public const StateRequest = 1 << 18;
+    public const ResolvedByNonReporter = 1 << 18;
 
-    public const EmulatorOther = 1 << 19;
+    public const StateRequest = 1 << 19;
```
Jamiras, line 47: It's generally unsafe to renumber enums. While unlikely, any users with bookmarks for a list with the EmulatorOther filter selected would now get StateRequest matches.

Tsearo: Ah I recall you mentioning that in the past thanks for reminding me 👍

Jamiras (approved):

Issue #1509, opened 2023-04-23
Double Space in Game-Tag referencing Subsets


SporyTike:
As the title mentions the [game=] tag for forums has a double space when referencing to a subset. It's between the game title and the console name, pretty sure it's there since normally the subset tag is between them which gets put at the end since recently.

To reproduce you can just create a forum post using that tag or check out an existing post like here where the tag references a subset. A post like this can be found [here](https://retroachievements.org/viewtopic.php?t=18082&c=132017#132017).

**Screenshot showing that there are two spaces in the tag between the game name and console name**
![image](https://user-images.githubusercontent.com/45672834/233862402-fc4d2631-fc9d-48b0-b717-f6e08c9021fb.png)

There should be some logic for such cases that forces a single space instead of multiple spaces in a row.

amine4567:
The problem seems to come from putting a [game=] tag inside a [code=] tag, if you try a [game=] tag alone, you won't get the double space as you can see from my reply to the linked post (sorry for polluting your post, I will remove my comment later on).

This still needs to be fixed.

wescopeland:
This will be resolved by https://github.com/RetroAchievements/RAWeb/pull/1544.

PR #1544, opened 2023-05-02
fix: remediate erroneous white space in game and ach shortcode rendering

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1509.

This is caused by the browser not trimming the extra whitespace when the shortcodes are rendered in a `<pre>` tag.

**Before**
(notice the space between the console name and game name on the first entry)
Screenshot 2023-05-01 at 11 11 01 PM

Screenshot 2023-05-02 at 1 54 50 PM

**After**
Screenshot 2023-05-01 at 11 10 24 PM

Screenshot 2023-05-02 at 1 54 41 PM


app_legacy/Helpers/render/game.php
```diff
@@ -81,7 +81,7 @@ function renderGameTitle(?string $title = null, bool $tags = true): string
             . "<span class='tag-arrow'></span>"
             . "<span>$subset</span>"
             . "</span>";
-        $updateHtml($html, $matches[0], $tags ? " $span" : '');
+        $updateHtml($html, " $matches[0]", $tags ? " $span" : '');
```
Jamiras, line 84: shouldn't the space be part of the match? what if the name doesn't have a preceding space?

(something maybe like this [untested])
`/\[Subset - (.+)\]/` => `/\s?\[Subset - (.+)\]/`

app_legacy/Helpers/render/achievement.php
```diff
@@ -79,7 +79,11 @@ function renderAchievementTitle(?string $title, bool $tags = true): string
     }
     $title = str_replace('[m]', '', $title);
 
-    return "$title$missableTag";
+    // If we don't strip consecutive spaces, the
+    // browser doesn't collapse them in forum <pre> tags.
+    $title = preg_replace('/\s+/', ' ', $title);
```
Jamiras, line 84: Does this actually do anything? Won't the whitespace just be a single preceding or trailing space (depending on where the [m] was)? If so, wouldn't `trim()` be easier?

wescopeland: With the line removed:
Screenshot 2023-05-05 at 2 20 51 PM

With the line changed to `$title = trim($title);`:
Screenshot 2023-05-05 at 2 21 12 PM

With the line restored:
Screenshot 2023-05-05 at 2 21 23 PM

Jamiras: Are the points already part of the title at this point? That's unfortunate.

wescopeland: Based on my interpretation of what's happening in `achievementsAvatar()`, yes that appears to be the case.

Jamiras (approved):

PR #1546, opened 2023-05-02
Small improvements to Claim List table

EmoonX:
A few small improvements to the aforementioned table.

image

- Font size is a bit smaller, to handle the copious amount of data
- "User" now reads "Dev"
- Game column has fixed, larger width
- "Claim Status" header is just "Status"
- Special shows just "--" when "None"
- Dates use the _smalldate_ styling and don't wrap
- Rightmost columns are centered instead of left-aligned (while the latter is used site-wide, I noticed having unaligned/misplaced text/headers didn't work well for those thin columns; thus the change)


EmoonX:
How's it supposed to look like later with the console and icon updates:

image

public/claimlist.php
```diff
@@ -163,13 +165,14 @@
         }
         echo "<br style='clear:both'>";
 
-        echo "<div class='table-wrapper'><table class='table-highlight'><tbody>";
+        echo "<div class='table-wrapper'><table class='table-highlight text-xs'><tbody>";
 
-        echo "<tr class='do-not-highlight'>";
-        echo "<th></th>";
         // Sortable table headers
+        echo "<tr class='do-not-highlight'>";
```
Jamiras, line 171: Suggest swapping the first two columns while you're in here to be consistent with the change in #1425.

Jamiras (approved):

PR #1547, opened 2023-05-02
feat(game): add hide unlocked achievements checkbox

wescopeland:
This PR adds a new checkbox to game pages that if appears if the user is signed in, has unlocked at least one achievement, and has at least one achievement that is not unlocked on either softcore or hardcore.

When the checkbox is checked, achievements that have already been unlocked will be hidden from the page. Now that many sets on RA have well over 100 achievements, this kind of functionality is growing potentially more useful.

There is a lot of noise in this PR from a Tailwind dependency update to resolve a bug that was discovered in https://github.com/RetroAchievements/RAWeb/pull/1518. Once that PR is merged, the wild numbers for adds+deletes should go away.

**Implementation Details**
- This is implemented using Blade and Alpine.
- The new front-end code has tests associated with it similar to the checkbox that was recently added to user pages.
- This setting is intentionally not persisted.

---

https://user-images.githubusercontent.com/3984985/235769230-ed180e69-61da-4948-a209-c4d1b34e22ab.mp4


Tsearo:
Is there potentially a valid use case for showing the checkbox still when the game is fully completed in order to scroll down to the other info such as RP and Comments more easily?

wescopeland:
> Is there potentially a valid use case for showing the checkbox still when the game is fully completed in order to scroll down to the other info such as RP and Comments more easily?

Honestly, I am not quite sure. On one hand, I think that makes sense. On the other, I wonder if it creates a roundabout UX that should instead be built as a way to quickly scroll to RP/Comments regardless of achievements unlocked status.

Tsearo:
> > Is there potentially a valid use case for showing the checkbox still when the game is fully completed in order to scroll down to the other info such as RP and Comments more easily?
> 
> Honestly, I am not quite sure. On one hand, I think that makes sense. On the other, I wonder if it creates a roundabout UX that should instead be built as a way to quickly scroll to RP/Comments regardless of achievements unlocked status.

That's true my other thought was maybe it would be better to have a button to scroll down past the list of that is the desire (in that scenario)

hb-robo:
What about the edge case where a player has mastered a game that is revised with more achievements? Will the check for mastery/completion hide the checkbox despite the presence of unearned achievements?

Tsearo:
> What about the edge case where a player has mastered a game that is revised with more achievements? Will the check for mastery/completion hide the checkbox despite the presence of unearned achievements?

It is checking your current earned achievements for the game against the current total so yes it will show the checkbox in that case

hb-robo:
> It is checking your current earned achievements for the game against the current total so yes it will show the checkbox in that case

Ah, yeah I just saw the logic for it. Definitely the way to go about it

Jamiras (approved):

PR #1548, opened 2023-05-02
fix(debug): don't render unless DEBUGBAR_ENABLED is true

wescopeland:
This PR makes a minor adjustment to the debug element which shows env, breakpoint, locale, and language. Now, if the Laravel debug bar is disabled, the debug element is also disabled.

This is helpful for showing demo screenshots/videos on tiny displays, like the iPhone SE.

Screenshot 2023-05-02 at 4 27 56 PM

### DIFF
--- a/resources/views/layouts/components/body-end.blade.php
+++ b/resources/views/layouts/components/body-end.blade.php
@@ -1,15 +1,6 @@
 {{-- TODO add livewire--}}
 {{--<livewire:scripts />--}}
-@if(!app()->environment('production'))
-    <script>
-        function hideDebugBar() {
-            const debugBarEl = document.querySelector('#debug');
-            if (debugBarEl) {
-                debugBarEl.classList.add('hidden');
-            }
-        }
-    </script>
-
+@if(!app()->environment('production') && env('DEBUGBAR_ENABLED'))
     <style nonce="{{ csp_nonce() }}">
         pre.xdebug-var-dump {
             background: #FFFFFF;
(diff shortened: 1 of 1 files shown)


resources/views/layouts/components/body-end.blade.php
```diff
@@ -1,6 +1,6 @@
 {{-- TODO add livewire--}}
 {{--<livewire:scripts />--}}
-@if(!app()->environment('production'))
+@if(!app()->environment('production') && env('DEBUGBAR_ENABLED'))
```
Jamiras, line 3: I think this was intended to minimize accidental changes to prod (see also #1519).

Note that unless the debugbar is completely collapsed, you'll never see this. So it almost seems more desirable to have when the debugbar is disabled.

Also note that the debugbar is not enabled on the stage host.

Maybe the information could be moved up into the header, where it would be offscreen for most screenshots?

image

wescopeland: My original implementation actually caused this debug component to just hide on click. What do you think of that approach?

Jamiras: Seems like a reasonable solution.

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1548/commits/863027ddcd630b2c14dde15145e7d2ae5b425510.

Jamiras (approved):

PR #1549, opened 2023-05-04
Add a Derived Hacks section to game alternatives in the sidebar

EmoonX:
Add a new section to go along with the other ones. Inspired by [this suggestion](https://github.com/RetroAchievements/RAWeb/discussions/1453).

image

Results are fetched by a query for the publisher `Hack - $gameTitle`. **EDIT**: and also by searching for `[Hacks - ...` hubs among game alternatives.

A few points of notice:
- Empty sets aren't shown (due to cases like e.g Super Mario World, which has 30% of 0-pointer hacks)
- "Similar Games" doesn't include hacks anymore, to avoid redundancy
- A scrollbar was added to when content gets past a certain height, so sidebar isn't polluted with endless lists.

### DIFF
--- a/app_legacy/Helpers/database/game.php
+++ b/app_legacy/Helpers/database/game.php
@@ -232,6 +232,7 @@ function getGameHacks(int $gameID, string $gameTitle): array
         FROM GameData AS gd
         LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
         WHERE gd.Publisher = "Hack - $gameTitle" AND gd.Title LIKE "~Hack~%"
+            AND gd.ID IN (SELECT gameID FROM Achievements AS ach WHERE ach.flags = $officialCore)
         GROUP BY gd.ID, gd.Title
     SQL;
 
--- a/app_legacy/Helpers/database/game.php
+++ b/app_legacy/Helpers/database/game.php
@@ -220,13 +220,10 @@ function getGameAlternatives(int $gameID, ?int $sortBy = null): array
 function getGameHacks(int $gameID, string $gameTitle): array
 {
     $officialCore = AchievementType::OfficialCore;
-    $alternativeTitleQuery = '';
     if (str_contains($gameTitle, ': ')) {
-        $alternativeTitle = str_replace(': ', ' - ', $gameTitle);
-        sanitize_outputs($alternativeTitle);
-        $alternativeTitleQuery = "OR gd.Title = '[Hacks - $alternativeTitle]'";
+        $gameTitle = str_replace(': ', ' - ', $gameTitle);
     }
-    sanitize_outputs($gameTitle);
+    sanitize_outputs($gameTitle);    
 
     $query = <<<SQL
         SELECT gd.ID gameIDAlt, gd.Title, gd.ImageIcon, c.Name ConsoleName,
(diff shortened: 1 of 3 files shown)


Soulsderfuchs:
We use hubs for this, none such game page will get this many relations or at all manually.

The hacks hub relation is directly hotlinked in the game page info itself in the publisher field of the hacks gamepages.

EmoonX:
> We use hubs for this, none such game page will get this many relations or at all manually.
> 
> The hacks hub relation is directly hotlinked in the game page info itself in the publisher field of the hacks gamepages.

Sorry, but I'm really struggling to get what you mean. Are you saying there's a better (or more foolproof) way of fetching hacks for a given game? Publisher field was all I could find (though I can be wrong).

Soulsderfuchs:
> > We use hubs for this, none such game page will get this many relations or at all manually.
> > The hacks hub relation is directly hotlinked in the game page info itself in the publisher field of the hacks gamepages.
> 
> Sorry, but I'm really struggling to get what you mean. Are you saying there's a better (or more foolproof) way of fetching hacks for a given game? Publisher field was all I could find (though I can be wrong).
> 
> edit: I think I get it now. It seems most games have the Publisher field written directly as "Hack - [base game]" in the DB... but in others the field is simply blank yet RA still fetches the base game from somewhere. Gonna research further into it.

Yeah I wholy believe that clicking that field which brings up the entire game hacks hub view it's a better way to glance at the hacks themselves.

Tsearo:
To be fair there is nothing associated to the base SMW page linking to anything related to it's hacks and none of the linked hubs can get you to them either at least not in a few jumps. I can see at least some reference to it being helpful regardless of if it is just a Hacks section linking to [Hacks - Super Mario World] or something else

Soulsderfuchs:
> To be fair there is nothing associated to the base SMW page linking to anything related to it's hacks and none of the linked hubs can get you to them either at least not in a few jumps. I can see at least some reference to it being helpful regardless of if it is just a Hacks section linking to [Hacks - Super Mario World] or something else

Could probably be just linked to it's hacks hub if somehow it can detect it as being it's base game and thus not overwriting in the publisher field like a regular hack page does.

EmoonX:
If having a Hacks section is counterproductive, then perhaps the same can be said about the Subsets and Similar Games sections. Should they be replaced by separate pages too? Certainly not, because that would be moving backwards.

One important point is that most games are far from behemoths like SMW or Sonic 1. Searching the DB here (considering only games with nonempty Publisher fields), I found:
- **88** titles with just one hack
- **45** with between 2 and 5 hacks
- only **24** with more than that

Which means the vast majority of base games have indeed _few_ hacks. If an user finds a game with just, say, one or two hacks, he won't want to visit a separate page just for viewing those. And we can't force them to; it's simply bad from an UX standpoint.

I'm all up for adding a link to the hub in the section header though, regardless of how many hacks are displayed. Then the user is free to decide if they want quick & condensed info or click for detailed (even though the hub pages are pretty barebones at the moment).

Soulsderfuchs:
Those game pages with few hacks already have them linked in similar games but big games like SMW and SM64 would not be practical to have every single hack linked to it even if you implement a scrollbar , players can go to the hub itself then.

wescopeland:
Possible solution: if a game contains more than N hacks, a certain number could be shown and some kind of "see more" link could be displayed to link to a page listing the rest.

EmoonX:
>Those game pages with few hacks already have them linked in similar games

There are many instances where this isn't respected. Not to mention several cases where some \~Hack\~s aren't actually hacks for the game on the page, but rather a different one (like [here](https://retroachievements.org/game/510)). This PR aims for a dev-agnostic way of displaying derived hacks.

>Possible solution: if a game contains more than N hacks, a certain number could be shown and some kind of "see more" link could be displayed to link to a page listing the rest.

This is a decent solution. Fits the condensed vs detailed schema too.

EmoonX:
- Hubless:
image

- Hub and few hacks:
image

- Many hacks (maybe a shadow effect here later):
image

Jamiras:
I would much rather just have a sidebar link to the hub page. 

image

As a singular link to the hub.

image

Or even just linking to the hub normally (even though that would list it in the hub too):

image

EmoonX:
>I would much rather just have a sidebar link to the hub page.

The main issue is... not every hack is in a dedicated hub. Sometimes they're in a `[Series Hack - ...]` instead, and other times they're not linked anywhere at all. Some of those are one-time game hacks that would make little sense being catalogued anyway.

suspect15:
My suggestion was based on my thoughts that the hubs are out of hand in some cases (see Zelda II for example):
image

I think several of the suggestions here that are dedicated ways are great solutions to identify hacks that don't require sifting through an unsorted and often heavily watered down list of hubs.

PR #1550, opened 2023-05-05
feat(gameList): Show the count of leaderboards owned by the dev even for games the dev hasn't done cheevos for

cIymax:
Include dev's leaderboards for games even when the dev didn't do any cheevos for that game.

Test case: https://retroachievements.org/gameList.php?d=clymax

Note that the following entry isn't included, even though the expected behavior could be that it should be included.
Battle of Olympus (GB) | Achievements: 0 of 24 | Leaderboards: 2 of 2
(source: https://retroachievements.org/game/6433 )

Related to #1474; see also its topic in the site-feedback channel in the Discord server.

Note, because my RAWeb test server is not yet fully set up, I'm submitting this untested on my part; any feedback/assistance is welcome.

PR #1551, opened 2023-05-05
feat(devStats): Add Leaderboard section to individual dev stats page

cIymax:
**Highlights**
- This PR updates the individual dev stats page to add a new section showing statistics on leaderboards owned by the dev.
- Can be considered as a companion to [#1474](https://github.com/RetroAchievements/RAWeb/pull/1474) and [#1550](https://github.com/RetroAchievements/RAWeb/pull/1550); those two reflect the same kind of info but on the dev game listing rather than on the dev stats page.

With this PR, the sections on the dev stats [page](https://retroachievements.org/individualdevstats.php?u=clymax) will be as follows.
- Games
- Achievements
- Leaderboards ← _new section_ to house leaderboard-related metrics
- Code Notes
- Tickets

**Considerations**
- There are a lot more leaderboard stats I can try to add later on, but I wanted to make this PR as simple to review as possible.
- Feedback can also be provided via the Discord [topic](https://discord.com/channels/310192285306454017/1103919950516723742).
- Because my RAWeb test server is not fully set up, I'm submitting this untested; any feedback is welcome.


public/individualdevstats.php
```diff
@@ -1192,6 +1201,20 @@ function drawChart() {
             echo "</table></tbody>";
             echo "</br></br>";
 
+            /*
+             * Leaderboards
+             */
+            echo "<H1>Tickets</H1>";
```
wescopeland, line 1207: Is the intent to duplicate this H1?

cIymax: Fixed!

app_legacy/Helpers/database/leaderboard.php
```diff
@@ -854,3 +854,30 @@ function GetLBPatch(int $gameID): array
...
+              GROUP BY GameID, GameTitle
+              HAVING LeaderboardCount > 0
+              ORDER BY LeaderboardCount DESC, GameTitle";
+
+    $dbResult = s_mysql_query($query);
+    if ($dbResult !== false) {
+        while ($db_entry = mysqli_fetch_assoc($dbResult)) {
+            $retVal[] = $db_entry;
+        }
+    }
```
wescopeland, line 880: `mysqli_fetch_assoc` is deprecated in favor of `legacyDbFetchAll()`.

wescopeland:
I can't open my dev stats page.

Screenshot 2023-05-05 at 9 28 33 AM

wescopeland (changes requested):
`s_mysql_query` and `mysqli_fetch_assoc` are both deprecated as of V2. We should try to mitigate their surface impact when possible. Throughout the codebase you will find instances of `legacyDbFetch()` and `legacyDbFetchAll()` which queries are being migrated towards.

public/individualdevstats.php
EmoonX: Shouldn't it be just lowercase also (`<h1>`)?

wescopeland: It seems to be mirroring the other H1 tags on the page, which is probably okay for the purposes of the PR but is a major semantic bug with the page itself as it should only have a single canonical H1. All the existing H1 tags are probably more well-suited as H2 tags.

EmoonX: Ah, you're right. Better just leave it for some major cleanup later.

app_legacy/Helpers/database/leaderboard.php
cIymax: Updated to use legacyDbFetchAll!

wescopeland:
It seems the query is trying to select from a table that doesn't exist.

Screenshot 2023-05-05 at 4 00 49 PM

cIymax:
@wescopeland Updated!

wescopeland:
Sorry, this still isn't working.

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'Scott' in 'field list' (SQL: SELECT gd.Title as GameTitle, gd.ImageIcon as GameIcon, c.Name as ConsoleName, lb.GameID as GameID, COUNT(lb.GameID) as TotalLeaderboards, SUM(CASE WHEN lb.Author = Scott THEN 1 ELSE 0 END) AS LeaderboardCount FROM LeaderboardDef AS lb LEFT JOIN GameData AS gd ON gd.ID = lb.GameID LEFT JOIN Console AS c ON c.ID = gd.ConsoleID WHERE gd.ID IN (SELECT GameID from LeaderboardDef WHERE Author = Scott GROUP BY GameID) AND gd.Title IS NOT NULL GROUP BY GameID, GameTitle HAVING LeaderboardCount > 0 ORDER BY LeaderboardCount DESC, GameTitle)
```

cIymax:
@wescopeland OK, I've added quotation marks in the query variable to account for the LeaderboardDef table not having a userID field but only a username string.

app_legacy/Helpers/database/leaderboard.php
```diff
@@ -854,3 +854,22 @@ function GetLBPatch(int $gameID): array
...
+              FROM LeaderboardDef AS lb
+              LEFT JOIN GameData AS gd ON gd.ID = lb.GameID
+              LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
+              WHERE gd.ID IN (SELECT GameID from LeaderboardDef WHERE Author = '$username' GROUP BY GameID)
+              AND gd.Title IS NOT NULL
+              GROUP BY GameID, GameTitle
+              HAVING LeaderboardCount > 0
+              ORDER BY LeaderboardCount DESC, GameTitle";
+
+    return legacyDbFetchAll($query)->toArray();
```
wescopeland, line 874:
```suggestion
    return legacyDbFetchAll($query, $bindings)->toArray();
```

I think the reason the query is working correctly is there aren't any bindings for the two `$username` values being injected into it.

```php
$bindings = [
  'user' => $username,
  'user2' => $username,
];

```
WHEN lb.Author = :user

WHERE Author = :user2
```

**Comment by wescopeland** [comment]
When the query does work, it seems there are some structural HTML issues. I am not too familiar with this particular page, but this is what I see when I open Scott's:

Screenshot 2023-05-05 at 4 19 59 PM

File: public/individualdevstats.php [hunk 4]
```diff
@@ -1192,6 +1201,20 @@ function drawChart() {
...
+             */
+            echo "<H1>Leaderboards</H1>";
+            echo "<table class='table-highlight'><tbody>";
+
+            // Leaderboards created
+            echo "<tr><td width='50%'>Leaderboards Created:</td><td>" . $userLeaderboardCount . "</td></tr>";
+
+            echo "</tbody></table>";
+            echo "</div>";
+            echo "</table></tbody>";
```
wescopeland, line 1215:
```suggestion
```

I am able to fix the structural HTML issues by deleting these two lines.

wescopeland:
I think all that's left now is to create a `$bindings` variable in the function and update the query to use the bindings.

```php
$bindings = [
  'user' => $username,
  'user2' => $username,
];
```

In your query, replace $username with :user and :user2 (bindings cannot be duplicated, kind of annoying).

wescopeland (dismissed):
The page seems to work now. I will leave the rest to @Jamiras and/or @luchaos.

public/individualdevstats.php
```diff
@@ -412,6 +412,15 @@
...
 $recentlyObtainedAchievements = getRecentUnlocks($achievementIDs, 0, $maxRecentAchievements);
 
+// Initialize leaderboard variables
+$userLeaderboardCount = 0;
+
+// Get leaderboard information for user
+$leaderboards = getLeaderboardCounts($dev);
+foreach ($leaderboards as $game) {
+    $userLeaderboardCount += $game['LeaderboardCount'];
+}
```
Jamiras, line 422: With #1474, you should be able to tally these up from $gamesList (~line 83) to avoid adding the new function and extra query.

luchaos: This sounds like a good improvement - @cIymax would you be willing to make this change?
Then we could merge this PR.

cIymax: The reason why I added the new function even with #1474 is so that we pull games for which the dev made zero achievements but some leaderboards for.  My understanding of #1474 is that it still wouldn't pull such games, hence the added function. @wescopeland, as the author of #1474, can you confirm my understanding is correct?

wescopeland: To the best of my recollection, that understanding is correct.

https://github.com/RetroAchievements/RAWeb/blob/cf5040e0b4206d27f8017f7fa235364f15ffb3ba/app/Helpers/database/game.php#L294C5-L294C5

cIymax: I see.  Let me know if/how I can assist.  I see a "Resolve conflicts" button below, but it's grayed out for me.

wescopeland:
It appears this PR is based against V2. I am going to convert this PR to a draft until such time that it is ready to be reviewed.

PR #1553, opened 2023-05-06
feat(game): require guide links to be from RAGuides repo

wescopeland:
This PR implements the outcome of this community poll: https://discord.com/channels/310192285306454017/511718348178849792/1104087071460704286

Validation has been added to the "Guide URL" field in the devbox requiring URLs to come from the RAGuides repo. This will ensure content linked to via the "Guide" link is content directly under RA's control, which will make a future on-site migration easier.

Screenshot 2023-05-06 at 9 52 19 AM

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -963,7 +963,7 @@ sanitize_outputs(
                     echo "<input type='hidden' name='genre' value='" . attributeEscape($genre) . "'>";
                     echo "<input type='hidden' name='release' value='" . attributeEscape($released) . "'>";
                     echo "<div class='md:grid grid-cols-[180px_1fr_100px] gap-1 items-center mb-1'>";
-                    echo "<label for='guide_url'>Guide URL</label><input type='url' name='guide_url' id='guide_url' value='" . attributeEscape($guideURL) . "' class='w-full'>";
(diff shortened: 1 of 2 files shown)


public/gameInfo.php
```diff
@@ -963,7 +963,7 @@ function ResetProgress() {
                     echo "<input type='hidden' name='genre' value='" . attributeEscape($genre) . "'>";
                     echo "<input type='hidden' name='release' value='" . attributeEscape($released) . "'>";
                     echo "<div class='md:grid grid-cols-[180px_1fr_100px] gap-1 items-center mb-1'>";
-                    echo "<label for='guide_url'>Guide URL</label><input type='url' name='guide_url' id='guide_url' value='" . attributeEscape($guideURL) . "' class='w-full'>";
+                    echo "<label for='guide_url' class='cursor-help italic' title='Must be from https://github.com/RetroAchievements/guides'>Guide URL<sup>*</sup></label><input type='url' name='guide_url' id='guide_url' value='" . attributeEscape($guideURL) . "' class='w-full'>";
```
Jamiras, line 966: Why does this get to be italic? None of the other fields are.

wescopeland: I was hoping to draw some kind of connection that there's a tooltip attached to the label.

Jamiras: The cursor-help is probably sufficient.

The only other place I can think of where we do something like this is the report ticket page, but those pop up a window instead of just a tooltip.

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1553/commits/4700b3d33325f06fb2ed051461cda9132606a936

Jamiras (approved):

PR #1556, opened 2023-05-07
feat(devStats): Show leaderboard counts on the global dev stats page

cIymax:
This PR adds a new column to the global dev stats [page](https://retroachievements.org/developerstats.php) to show leaderboard counts.

image

**Considerations**
- The new column can be the rightmost column to minimize impact on users visually accustomed to the existing layout, although an alternate position is also depicted.
- Feedback on column inclusion/ordering can be provided in this thread or via the Discord [topic](https://discord.com/channels/310192285306454017/1104581162178977864).  If it turns out that we don't want to add this new column to that page for now, to keep the page more focused, no worries.
- Because my RAWeb test server is not fully set up, I'm submitting this untested on my part. Any feedback/assistance would be greatly appreciated.


EmoonX:
Ideally, I think that table could use a better column ordering. My personal suggestion:
- Achievements (default page sort)
- Ticket info derived from achievements (Open Tickets, Tickets Ratio)
- Achievement statistics (Yielded Unlocks, Yielded Points)
- Leaderboards (correlated with achievements)
- Active Claims

EmoonX:
Heads-up though: your SS is out-of-date actually, as per #1542 : the column `Ticket Ratio` now reads `Tickets Resolved For Others`, and field links have been added. The commit content should be fine, though.

Jamiras (commented):
> The new column can be the rightmost column to minimize impact on users visually accustomed to the existing layout, although an alternate position is also depicted.

I doubt anyone is "accustomed to the existing layout" of this particular page. I think I've visited it three or four times in the many years I've been with the site.

And you move the Open Tickets column anyway (which I think is a reasonable change).

I'd put leaderboards over by achievements.

PR #1559, opened 2023-05-07
feat: throw an error on forum/comment double posts

wescopeland:
This PR adjusts the function calls for submitting comments and forum topic replies.

If the user submits a duplicate payload on the same article/comment ID, the function now returns an error and the duplicate post is blocked.

This issue can occur when the site doesn't instantly respond to a post submit and the user repeatedly spams the button click. After this change, only the first post will be preserved and all subsequent duplicates are blocked.

Screenshot 2023-05-06 at 9 49 03 PM

### DIFF
--- a/app_legacy/Helpers/database/forum.php
+++ b/app_legacy/Helpers/database/forum.php
@@ -301,7 +301,8 @@ function submitTopicComment(
     $userID = getUserIDFromUser($user);
 
     if (getIsForumDoublePost($userID, $topicID, $commentPayload)) {
-        return false;
+        // Fail silently.
+        return true;
     }
 
     // Replace inverted commas, Remove HTML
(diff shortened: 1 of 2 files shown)


app_legacy/Helpers/database/forum.php
```diff
@@ -278,6 +300,10 @@ function submitTopicComment(
     sanitize_sql_inputs($user);
     $userID = getUserIDFromUser($user);
 
+    if (getIsForumDoublePost($userID, $topicID, $commentPayload)) {
+        return false;
```
Jamiras, line 304: I'm not sure what this would look like for an actual double post, but manually entering the exact test from the previous post results in an error banner. Maybe it should silently fail?

wescopeland: I agree it would be better to silently fail. Resolved in https://github.com/RetroAchievements/RAWeb/pull/1559/commits/e6b2669d350f032bf00e601c4fc5d98fe8ca0634.

Jamiras (approved):

Issue #1088, opened 2022-08-13
Global Ranking issues


Tsearo:
Mastered (and soon Completed) columns do not report accurate data for the Daily/Weekly filters and will at most display 1
![image](https://user-images.githubusercontent.com/4047771/184506167-74578459-723f-4a49-8e55-f2dd366a5615.png)

![image](https://user-images.githubusercontent.com/4047771/184506173-594b0efc-db62-4450-a173-7d4ea65eed37.png)

Self rank among followed users is not correctly displayed when you move to 
![image](https://user-images.githubusercontent.com/4047771/184506112-c95f7168-f20f-4675-961f-9e9ea9086993.png)

MSGoodman:
Sorry for being so noisy, I'm just making my way through all of the issues in the backlog to resolve as many of them as possible. This one looks like it can be marked as resolved, right? Testing this myself it looks correct, the merged PRs definitely seem to have done their job.

wescopeland:
This appears to now be resolved.

PR #1561, opened 2023-05-07
fix(globalRanking): Mastered column only displays a value of 1 or 0 when using Daily/Weekly filters

cIymax:
**Highlights**
- This PR attempts to fix an issue with the Mastered column on the global ranking [page](https://retroachievements.org/globalRanking.php?s=9&t=1&d=2023-05-07&f=0) when certain filters are used.
- The issue was first described in [#1088](https://github.com/RetroAchievements/RAWeb/issues/1088) as quoted below.
> Mastered (and soon Completed) columns do not report accurate data for the Daily/Weekly filters and will at most display 1

image

**Considerations**
- Because my RAWeb test server is not fully set up, I'm submitting this untested on my part.  Any feedback/assistance would be greatly appreciated.
- As for the second issue described in [#1088](https://github.com/RetroAchievements/RAWeb/issues/1088) and relating to Friends, I haven't looked into it yet, because I haven't started using the Follow/Friend features yet.

### DIFF
--- a/app_legacy/Helpers/render/leaderboard.php
+++ b/app_legacy/Helpers/render/leaderboard.php
@@ -580,7 +580,8 @@ function getGlobalRankingData(
                           $friendCondAward
                           $singleUserAwardCond
                           $untrackedCond
-                          GROUP BY sa.User, sa.AwardData, sa.AwardDate
+                          AND AwardType = " . AwardType::Mastery .
+                          "GROUP BY sa.User
                       )
                   ) AS Query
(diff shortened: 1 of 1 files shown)


Tsearo (approved):
Looks good to me, checked both hardcore and softcore, all vs followed users and they all seem to check out and sorting works as well

Since you can't see it locally here is an example from the original issue
Before:
image

After:
image

Jamiras (approved):

PR #1562, opened 2023-05-07
fix(globalRanking): Rank column shows your rank as being 0 when using the Followed Users filter and browsing pages past your actual rank

cIymax:
**Highlights**
- This PR attempts to fix an issue with the Rank column on the global ranking [page](https://retroachievements.org/globalRanking.php?s=5&t=2&d=2023-05-07&f=1&o=25) using the Followed Users filter and when browsing pages past your actual rank.
- The issue was first described in the latter part of [#1088](https://github.com/RetroAchievements/RAWeb/issues/1088) as quoted below.
> Self rank among followed users is not correctly displayed when you move to

image

**Considerations**
- Because my RAWeb test server is not fully set up, I'm submitting this untested on my part.  Any feedback/assistance would be greatly appreciated.
- The other issue described in the former part of [#1088](https://github.com/RetroAchievements/RAWeb/issues/1088) is addressed by [#1561](https://github.com/RetroAchievements/RAWeb/pull/1561).

### DIFF
--- a/public/globalRanking.php
+++ b/public/globalRanking.php
@@ -45,7 +45,7 @@ $lbUsers = match ($friends) {
 if ($friends == 1) {
     // We do a maxCount + 1 so that if we get maxCount + 1 rows returned we know
     // there are more row to get and we can add a "Next X" link for page traversal
-    $data = getGlobalRankingData($type, $sort, $date, null, $user, $untracked, $offset, getFriendCount($user) + 1, 0);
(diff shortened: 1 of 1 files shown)


Jamiras:
> Because my RAWeb test server is not fully set up, I'm submitting this untested on my part. Any feedback/assistance would be greatly appreciated.

Please do not submit **untested** PRs. If you need help getting your test server set up, seek it.

PR #1571, opened 2023-05-08
Update `Completion Progress` table layout

EmoonX:
Following this [Discord topic](https://discord.com/channels/310192285306454017/1086195554872217650).

Changes table layout so two columns are used instead of three. Progress bar and counter are put directly under the game's title. This increases the average row height, so icons are a bit larger now to compensate.

Before/after:

image
image

Layout will eventually need some adjustments to deal with the console rendering PR, but, as per testing, the overall structure apparently can be kept without major drawbacks.


Jamiras:
I'm seeing a variance in alignment between entries with 100+ achievements and those with less:
image

The same thing doesn't happen going from double digits to single digits.
image

PR #1572, opened 2023-05-09
refactor: migrate Active Players component from Knockout.js to Livewire

wescopeland:
This PR makes a significant refactor of the Active Players component.

**Why?**
The current implementation of this component relies heavily on Knockout.js, a legacy framework that we've been gradually phasing out in favor of the TALL stack (Tailwind, Alpine.js, Laravel, and Livewire).

There are several benefits to this refactor:
* **Consistency.** By using the TALL stack, we align the Active Players component with the architectural direction of the application, leading to a more unified codebase. This consistency is beneficial for both onboarding new developers and maintenance.

* **Performance.** Livewire offers a more efficient data-binding model than Knockout.js, leading to better performance.

* **Readability and Maintainability.** Laravel's Blade templating language, which we're using in conjunction with Livewire, offers cleaner and more readable syntax than Knockout.js. This improves maintainability and makes future changes and enhancements easier.

---

https://user-images.githubusercontent.com/3984985/236978188-0e265b7f-9652-488e-a503-7fa66f8356c3.mp4


resources/css/utilities.css
```diff
@@ -1,3 +1,7 @@
+[x-cloak] {
```
wescopeland, line 1: `x-cloak` is currently not configured correctly in the project.

ref: https://alpinejs.dev/directives/cloak

Jamiras:
The list doesn't seem to be updating when the count is refreshed:
image
image

I also noticed this going from 0 to populated. I have not verified going from populated to differently populated.

wescopeland:
The observed issue is going to be slightly complex to resolve. It appears the list is not updating on the 1st re-render but is updating on every subsequent re-render, likely due to the state interplay between Livewire and Alpine.js. I will be moving this PR temporarily to draft in the meantime.

wescopeland:
I am going to inevitably take another pass at this, but completely from scratch.

PR #1586, opened 2023-05-12
fix forced score recalculation targeting logged in user

Jamiras:
https://discord.com/channels/476211979464343552/1026595325038833725/1106420549309186048


wescopeland (approved):

wescopeland:
Verified both admin and self recalculation.

PR #1593, opened 2023-05-14
fix(shortcode): support timestamps in youtube embeds

wescopeland:
This PR remediates an issue where timestamps are ignored by YouTube video embeds, such as on achievement pages.

**Root Cause**
The privacy-enhanced YouTube embed URL at `youtube-nocookie.com` does not support the `&t` query parameter. Instead, `&t` must be converted to `&start` which an integer of seconds representing the timestamp.

For example, `&t=1m30s` must be converted to `&start=90`.

### DIFF
--- a/tests/Unit/ShortcodeTest.php
+++ b/tests/Unit/ShortcodeTest.php
@@ -49,4 +49,42 @@ final class ShortcodeTest extends TestCase
             Shortcode::render('[code]test')
         );
     }
+
+    /**
+     * @dataProvider youtubeUrlProvider
+     */
+    public function testAutoEmbedYoutube(string $url, string $expected): void
+    {
+        $this->assertStringContainsString(
+            $expected,
+            Shortcode::render($url)
+        );
+    }
+
+    public function youtubeUrlProvider(): array
+    {
+        // ["given", "expected"]
+        return [
+            // A vanilla URL
+            ['https://www.youtube.com/v/YYOKMUTTDdA', '//www.youtube-nocookie.com/embed/YYOKMUTTDdA'],
+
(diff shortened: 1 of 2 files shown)


tests/Unit/ShortcodeTest.php
```diff
@@ -49,4 +49,42 @@ public function testUnclosedCodeTag(): void
...
+    public function youtubeUrlProvider(): array
+    {
+        // ["given", "expected"]
+        return [
+            // A vanilla URL
+            ['https://www.youtube.com/v/YYOKMUTTDdA', '//www.youtube-nocookie.com/embed/YYOKMUTTDdA'],
+
+            // With explicit video ID "v"
+            ['https://www.youtube.com/watch?v=W3igStIUp8Y', '//www.youtube-nocookie.com/embed/W3igStIUp8Y'],
+            ['https://www.youtube.com/watch?v=SXuIABMlZKc&t=1m30s', '//www.youtube-nocookie.com/embed/SXuIABMlZKc?start=90'],
```
Jamiras, line 73: This doesn't appear to work if I actually paste the URL into a forum comment. I think it's escaping the ampersand to `&amp;` and tripping up the regex.

wescopeland: Excellent find, you're correct on the root cause. I've added a single line to explicitly restore the ampersands in the event that they are escaped - addressed in https://github.com/RetroAchievements/RAWeb/pull/1593/commits/6256f62f0b9f8dac2f5133a5295e1b156cd2a8df.

Jamiras (approved):

PR #1594, opened 2023-05-16
fix(downloads): add missing bizhawk supported systems

wescopeland:
This PR adds three missing systems to Bizhawk on the downloads page:

* Apple II
* Game Gear
* PC Engine CD

All three of these systems have been confirmed as supporting RA, but are currently absent from the list.


Jamiras (approved):
Note that this doesn't actually affect the server. It's just the default values.

I've applied the changes to the server.

PR #1595, opened 2023-05-22
fix(createmessage): remediate html structure issue

wescopeland:
This PR remediates an issue reported here: https://discord.com/channels/310192285306454017/1106029608408723486

**Root Cause**
The HTML structure for createmessage.php is invalid. Right now, the entire form renders outside its `<form>` tag.

### DIFF
--- a/public/createmessage.php
+++ b/public/createmessage.php
@@ -65,11 +65,12 @@ $(document).ready(onUserChange);
             echo "<div class='comment'>$messageContextPayload</div>";
         }
 
+        echo "<form action='/request/message/send.php' method='post'>";
+        echo csrf_field();
+
         echo "<table>";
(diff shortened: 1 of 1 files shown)


Jamiras (approved):
I don't understand why it fixes the problem, but I can confirm that it does.

It's not like the `<form>` and `</form>` were nested at different levels.

wescopeland:
You are correct, they were not nested at different levels. There is something going on in PHP's/Laravel's server-side renderer that I do not understand. The bug can be directly observed in the browser dev tools:

Screenshot 2023-05-23 at 10 13 49 AM

Nesting a form tag within a table tag is perfectly valid HTML, but the output structure by the server is invalid.

PR #1598, opened 2023-05-29
fix(navbar): have Global Ranking link default to All Time

wescopeland:
This seems to be happening everywhere else on the site (footer, home page) except the navbar.

ref: https://discord.com/channels/310192285306454017/1112652985328218173


app_legacy/Helpers/render/layout.php
```diff
@@ -165,7 +165,7 @@ function RenderToolbar(): void
     echo "<li><a href='/forumposthistory.php'>Recent Forum Posts</a></li>";
     echo "<li class='divider'></li>";
     echo "<li><a href='/userList.php'>User List</a></li>";
-    echo "<li><a href='/globalRanking.php'>Global Ranking</a></li>";
+    echo "<li><a href='/globalRanking.php?s=5&t=2'>Global Ranking</a></li>";
```
Jamiras, line 168: It would probably be more effective to change the default value for `t` in globalRanking.php to 2.

https://github.com/RetroAchievements/RAWeb/blob/664b00fed323b0a79c108231314dba7d9f4c408d/public/globalRanking.php#L12-L13

wescopeland: Good point. I've addressed this and updated all these existing links on the site in https://github.com/RetroAchievements/RAWeb/pull/1598/commits/a281234cc0247249a8092904f6167b5c37c19011.

public/userInfo.php
```diff
@@ -294,7 +294,7 @@ function resize() {
                 $userRank = $userMassData['Rank'];
                 $rankPct = sprintf("%1.2f", ($userRank / $countRankedUsers) * 100.0);
                 $rankOffset = (int) (($userRank - 1) / 25) * 25;
-                echo "<a href='/globalRanking.php?s=5&t=2&o=$rankOffset'>$userRank</a> / $countRankedUsers ranked users (Top $rankPct%)";
```
Jamiras, line 297: `s` could also be eliminated here, but it's probably best to leave the two links on the profile with explicit parameters.

wescopeland: I've restored the explicit parameters on this page in https://github.com/RetroAchievements/RAWeb/pull/1598/commits/6735af1497c47b674df7727a2357a9de39143328. For my own intellectual curiosity, could you help me better understand the motivation behind keeping them on the user profile page?

Jamiras: These are explicit links to show the user's position in the list. We want to ensure they always go to the user even if the defaults change (not that I'm expecting them to change).

Jamiras (approved):

PR #1604, opened 2023-06-07
fix(game): remediate escape characters issue with titles

wescopeland:
This PR resolves an issue with character escaping on game pages.

**Root Cause**
Currently, game titles are using Laravel's `e()` function to escape `$gameTitle` in the Blade template before passing it to `renderGameTitle()`. However, this leads to a problem where game titles containing special HTML characters (like "Mario & Luigi: Bowser's Inside Story") are incorrectly displayed due to the HTML encoding of certain characters (& becoming &amp;, ' becoming &#039;, etc.).

**Solution**
`renderGameTitle()` has been modified to escape only the specific dynamic parts of the HTML it generates (i.e., `$category` and `$subset`) and not the entire `$title`. This was achieved by using the `htmlspecialchars()` function on these variables before inserting them into the generated HTML.

This change prevents potential XSS attacks while ensuring that game titles display correctly in the HTML. With this modification, it was no longer necessary to use `e()` on `$gameTitle` in the Blade template.

This approach also ensures that `renderGameTitle()` remains compatible with non-Laravel PHP pages, as it doesn't rely on Laravel-specific functions or classes.

**Before**
Screenshot 2023-06-07 at 6 02 10 PM

**After**
Screenshot 2023-06-07 at 6 02 03 PM

Current prod issue: https://retroachievements.org/game/5323


app_legacy/Helpers/render/game.php
```diff
@@ -69,13 +69,13 @@ function renderGameTitle(?string $title = null, bool $tags = true): string
...
     foreach ($matches[0] as $i => $match) {
-        $category = $matches[1][$i];
+        $category = htmlspecialchars($matches[1][$i], ENT_QUOTES, 'UTF-8');
         $span = "<span class='tag'><span>$category</span></span>";
         $updateHtml($html, $match, $tags ? " $span" : '');
     }
     $matches = [];
     if (preg_match('/\s?\[Subset - (.+)\]/', $title, $matches)) {
-        $subset = $matches[1];
+        $subset = htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
```
Jamiras, line 78: Seems like this is undesirable:
image

wescopeland: Very much so, good catch. Should be resolved in https://github.com/RetroAchievements/RAWeb/pull/1604/commits/2c9f673e4500ea0d51665c0ee3617d61db751a0e.

Jamiras (approved):
I'm going to merge this as-is, but I'm concerned about the  need to unescape the value just so you can re-escape it. That implies something upstream is doing some unnecessary sanitization. We should expect this function to be called with the DB value.

PR #1605, opened 2023-06-07
feat: render console icons in more places

wescopeland:
This PR adds our console icons to various places the icons are missing on the site, namely:
* Achievement info pages.
* Leaderboard pages.
* A user's dev stats.

Additionally, it creates a common util in _app/helpers_ for getting the icon URL. It also eliminates the client-side `onerror` handling for icons on console game lists.

**Achievement Info Pages**
Screenshot 2023-06-07 at 7 25 41 PM

**User Dev Stats Pages**
Screenshot 2023-06-07 at 7 25 59 PM

**Leaderboard Pages**
Screenshot 2023-06-07 at 7 27 51 PM


wescopeland:
These are all the places I'm aware of where the icons are missing. I'm sure there are more -- I'd be happy to add it anywhere else in this PR as well.

public/gameList.php
```diff
@@ -196,6 +196,18 @@ function ListGames(
...
+
+function renderConsoleHeading(int $consoleID, string $consoleName): string
+{
+    $systemIconUrl = getSystemIconUrl($consoleID);
+
+    return <<<HTML
+        <h2 class="flex gap-x-2 items-center">
+            <img src="$systemIconUrl" alt="Console icon" width="32" height="32">
+            <span>$consoleName</span>
+        </h2>
```
Jamiras, line 208: I feel like 32px header images are too similar to 32px game images:
image

What do you think about 24px images and h3 text?
image

Jamiras: Just realized this also affects the all-games-for-a-console list too. And that looks okay at 32px. Probably because there's a singular header. It might be the transition from list to new header that makes it seem less desirable on the dev games list.

wescopeland: I agree with your assessment that 24px images and h3 text would probably look better on those dev-specific game lists.

In https://github.com/RetroAchievements/RAWeb/pull/1605/commits/381f64985e4a0bae69d6f539e804bb70c4e74572, I've modified the function so that this change is applied on those pages, but not the console-specific game list pages.

Jamiras (approved):

PR #1607, opened 2023-06-08
feat(game): add bucketing to achievement distribution chart

wescopeland:
This PR alters the behavior of the achievement distribution bar chart on game pages to enable bucketing when a set has more than 60 achievements. When bucketing is enabled, each bar represents a range of achievements, with the exception of the final bar on the chart which still represents the completions/masteries. The bucket sizes are generated dynamically based on the count of achievements in the set.

**Why?**
Currently, due to the width constraints of the chart, it becomes exceptionally difficult to use when a set has many achievements. Now that modern systems with lots of content are getting sets, this is becoming more and more commonplace.

**Before and After**
Current production is on the left, this PR is on the right.

_Final Fantasy II (NES) - Game 5527_
5527_compare

_Gran Turismo 4 (PS2) - Game 11301_
11301_compare

_Sonic the Hedgehog (MD) - Game 1_
There is no change here.
1_compare

Screenshot 2023-06-08 at 8 53 43 AM


public/gameInfo.php
```diff
@@ -300,32 +300,133 @@ function drawCharts() {
...
+                        echo ", ";
+                    }
+
+                    // Is this the last bucket? If so, we only want it to include
+                    // players who have earned all the achievements, not a range.
+                    if ($index == count($buckets) - 1) {
+                        $label = "Earned $numAchievements achievements";
+                        printBucketIteration($bucketIteration, $numAchievements, $bucket, $label);
+                        $bucketLabels[] = $label;
+                        $hAxisValues[] = $numAchievements;
```
Jamiras, line 385: Separating the last item out into its own bucket confused me a bit in the graphs, but I understand the reasoning. From your FFII example:

image

56-59=4
60-63=3
64-64=1

This is probably related to my other comment about integer-sized buckets.

public/gameInfo.php
```diff
@@ -300,32 +300,133 @@ function drawCharts() {
...
+            function calculateBuckets(int $numAchievements, array $achDist, array $achDistHardcore): array {
+                // Enable bucketing based on the number of achievements in the set.
+                // This number was picked arbitrarily, but generally reflects when we start seeing
+                // width constraints in the Achievements Distribution bar chart.
+                $isBucketingEnabled = $numAchievements > 60;
+
+                // Bucket size is determined based on the total number of achievements in the set.
+                // If bucketing is enabled, we aim for roughly 20 buckets (hence dividing by 20).
+                // If bucketing is not enabled, each achievement gets its own bucket (bucket size is 1).
+                $bucketSize = $isBucketingEnabled ? ceil($numAchievements / 20) : 1;
```
Jamiras, line 312: Using integer sized buckets leads to some awkward distributions. Here's what the buckets look like for a game with 78 achievements (Snake's Revenge-NES):
image

The last few buckets are:
65-68=4
69-72=4
73-76=4
77-77=1
78-78=1

78/4 = 3.9, so the buckets should be 0, 3.9, 7.8, 11.7, 15.6, 19.5, 23.4, 27.3, 31.2, 35.1, 39, 42.9, 46.8, 50.7, 54.6, 58.5, 62.4, 66.3, 70.2, 74.1, 78. You would round those values to get a more fluid distribution of buckets.

Additionally, are you're explicitly keeping a separate bucket for "all achievements", the calculation should be `($numAchievements - 1) / 19`, which would yield the following 20 buckets: 4, 8, 12, 16, 20, 24, 28, 32, 36, 41, 45, 49, 53, 57, 61, 65, 69, 73, 77, 78. The first 19 buckets are evenly distributed using rounded values (note the small jump between buckets 36 and 41 as the fractional portion transitions), and the 20th bucket is reserved for the "all achievements".

public/gameInfo.php
```diff
@@ -300,32 +300,133 @@ function drawCharts() {
...
-                    $plural = 's';
-                    echo ", ";
+            function calculateBuckets(int $numAchievements, array $achDist, array $achDistHardcore): array {
+                // Enable bucketing based on the number of achievements in the set.
+                // This number was picked arbitrarily, but generally reflects when we start seeing
+                // width constraints in the Achievements Distribution bar chart.
+                $isBucketingEnabled = $numAchievements > 60;
+
+                // Bucket size is determined based on the total number of achievements in the set.
+                // If bucketing is enabled, we aim for roughly 20 buckets (hence dividing by 20).
```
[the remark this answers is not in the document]
wescopeland: For sure. I've set this to 44, which seems closer to being in a sweet spot. Happy to adjust it further.

public/gameInfo.php
wescopeland: Good catch, this was difficult indeed to work through, but I may have landed on a resolution in https://github.com/RetroAchievements/RAWeb/pull/1607/commits/1f430f4964997718468b477db45824746e9fb8fb.

First of all, thank you for providing the Snake's Revenge example. This was an excellent test case.

I've adjusted how "smoothing" between the buckets works for hopefully better distributions to be presented to the user:

Screenshot 2023-06-11 at 1 53 39 PM

Screenshot 2023-06-11 at 1 53 45 PM


I _think_ I've got a better handle on this now, but always possible I've missed something. Let me know if you spot anything that looks bizarre.

Jamiras: The X axis labels go from 1-20 now. Can we get it to show start values instead (since the last bucket will always be N, it would be better to have the second-to-last bucket not be N-1, and then the first bucket would be 1, providing the even spacing all the way to the final bucket.

wescopeland: Oops, I deleted something I shouldn't have. As of https://github.com/RetroAchievements/RAWeb/pull/1607/commits/aff4a526ecf6ffbf47d7e973e60009bcf5672607 this functionality should be restored.

Jamiras: I still think it would be better to use the bucket start as the X axis labels (particularly for the bookend entries [4/77 in this case would become 1/74]). It would be a smoother distribution with the last bucket (78).
image

Also, it might be beneficial to use an odd number of buckets so the last one gets labelled.

wescopeland: Both of these pieces of feedback should now be applied.

For the odd number of buckets, I increased the bucket count by 1 rather than decreased. When decreasing by 1, every single bar got a label on the x-axis which felt kind of cramped.

Jamiras (approved):

PR #1608, opened 2023-06-09
fix: hide zero-value RetroPoints display if Points value is zero

wescopeland:
This PR fixes an issue on game pages and achievement info pages where if points is equal to zero, a retropoints value of zero is always shown when it should actually be hidden.

Currently:
On game pages, the zero points value is currently hidden but retro points is always shown.
On achievement info pages, both zero-point values are shown.

After this PR:
On game pages, if an achievement is worth zero points, a retro points value of zero does not appear.
On achievement info pages, if both the achievement and its retro points are zero, neither will be displayed.

**BEFORE**
Screenshot 2023-06-09 at 6 34 43 PM

Screenshot 2023-06-09 at 6 34 55 PM

**AFTER**
Screenshot 2023-06-09 at 6 32 10 PM

Screenshot 2023-06-09 at 6 32 20 PM

### DIFF
--- a/public/achievementInfo.php
+++ b/public/achievementInfo.php
@@ -193,7 +193,7 @@ RenderContentStart($pageTitle);
         echo "<div class='flex justify-between'>";
         echo "<div>";
         echo "<a href='/achievement/$achievementID'><strong>$renderedTitle</strong></a>";
-        if ($achPoints !== 0 && $achTruePoints !== 0) {
+        if ($achPoints !== 0) {
             echo " ($achPoints)<span class='TrueRatio'> ($achTruePoints)</span>";
         }
         echo "<br>";
--- a/public/achievementInfo.php
+++ b/public/achievementInfo.php
@@ -189,9 +189,14 @@ RenderContentStart($pageTitle);
         echo "<div id='achievemententry'>";
 
         $renderedTitle = renderAchievementTitle($achievementTitle);
+
         echo "<div class='flex justify-between'>";
         echo "<div>";
-        echo "<a href='/achievement/$achievementID'><strong>$renderedTitle</strong></a> ($achPoints)<span class='TrueRatio'> ($achTruePoints)</span><br>";
+        echo "<a href='/achievement/$achievementID'><strong>$renderedTitle</strong></a>";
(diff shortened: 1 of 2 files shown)


public/achievementInfo.php
```diff
@@ -189,9 +189,14 @@ function ResetProgress() {
         echo "<div id='achievemententry'>";
 
         $renderedTitle = renderAchievementTitle($achievementTitle);
+
         echo "<div class='flex justify-between'>";
         echo "<div>";
-        echo "<a href='/achievement/$achievementID'><strong>$renderedTitle</strong></a> ($achPoints)<span class='TrueRatio'> ($achTruePoints)</span><br>";
+        echo "<a href='/achievement/$achievementID'><strong>$renderedTitle</strong></a>";
+        if ($achPoints !== 0 && $achTruePoints !== 0) {
```
Jamiras, line 196: This will also hide points on items that have points but true points haven't been calculated yet.

public/gameInfo.php
```diff
@@ -1360,7 +1360,9 @@ function ResetProgress() {
                                 icon: false,
                                 tooltip: false,
                             );
-                            echo " <span class='TrueRatio'>($achTrueRatio)</span>";
+                            if ($achPoints !== 0) {
```
Jamiras, line 1363: this will still show "(0)" for items that have points but true points haven't been calculated yet.

Jamiras (commented):
I can see wanting to hide both values if the achievement is worth 0 points, but an achievement's retro points are 0 when it's uploaded until the first recalculation for the game occurs. During that time, with this change, the points and retro points will be visible in the game list, but neither points nor retro points will be visible when viewing the achievement.

At a minimum, they should behave the same. I'd recommend only looking at the achievement points when determining the visibility of both items.

wescopeland:
Thanks, Jamiras. Both conditionals are now the same and are only checking for the achievement points value, as of https://github.com/RetroAchievements/RAWeb/pull/1608/commits/ab23ed7e98b248b529ac5b26f7efb65cb5d8b37b.

Jamiras (approved):

PR #1614, opened 2023-06-16
add admin tool for finding accounts with the same email address as another account

Jamiras:


public/admin.php
```diff
@@ -101,6 +102,62 @@
...
+    if ($forUser == null) {
+        $message = "Unknown user: $altsForUser";
+    } else {
+        $altsForUser = $forUser->User;
+
+        $emailAddresses = [];
+        if (!empty($forUser->EmailAddress)) {
+            $emailAddresses[] = $forUser->EmailAddress;
+        }
+        if (!empty($forUser->email_backup && $forUser->email_backup != $forUser->EmailAddress)) {
```
wescopeland, line 118: I seem to be missing this `email_backup` column in my `UserAccounts` table, but I'm wondering if there is potentially a logic error here due to a typo:

```suggestion
        if (!empty($forUser->email_backup) && $forUser->email_backup != $forUser->EmailAddress) {
```

rzumer:
Looks good! I suggest including a screenshot of the new tool in the PR to help visualize the UI as this is harder to read than the logic :)

luchaos (approved):

PR #1616, opened 2023-06-18
feat(code-notes): add delete functionality to the page

wescopeland:
This PR gives achievement developers the ability to delete code notes from the code notes page.

**Implementation Details**
* The delete functionality is gated behind the "Edit" button so JrDev/Dev privileges are still honored.
* The delete button is spaced far away from the "Save" button to prevent accidental misclicks.
* On delete, the row is deleted and the note count is decremented by 1. The row in the DB is changed to an empty string.

https://github.com/RetroAchievements/RAWeb/assets/3984985/200b5c1c-d5ac-4a0d-aa94-585e8f15bfb3


public/codenotes.php
```diff
@@ -148,21 +162,30 @@ function saveCodeNote(rowIndex) {
         // Before bailing from edit mode, make sure all new values for
         // the note (the text content and author avatar) are properly
         // displayed in the UI.
-        const noteValueWithLineBreaks = noteEditEl.value.replace(/\n/g, '<br />');
-        noteDisplayEl.innerHTML = noteValueWithLineBreaks;
+        if (isDeleting) {
```
rzumer, line 165: Can this be moved up somehow? Guarding every previous conditional expression with !isDeleting etc. is a bit awkward.

wescopeland:
@Jamiras This one should be all set now for V3.

Jamiras:
If I bump my permissions down to 2 (Jr Dev), I get an error trying to delete a note (after confirmation dialog):
image

wescopeland:
> If I bump my permissions down to 2 (Jr Dev), I get an error trying to delete a note (after confirmation dialog)

Good catch, I think I've narrowed this down to something that changed in `submitCodeNote2()` as a result of V3. Will have a fix pushed up shortly.

app/Helpers/database/code-note.php
```diff
@@ -82,7 +82,7 @@ function submitCodeNote2(string $username, int $gameID, int $address, string $no
     if (
         $i !== false
         && $permissions <= Permissions::JuniorDeveloper
-        && $currentNotes[$i]['User'] !== $user
+        && $currentNotes[$i]['User'] !== $user->User
```
wescopeland, line 85: @Jamiras I believe your find is the result of a bug in this conditional that emerged in the V3 branch. Beforehand, I think `$user` was a string containing the username, now it's an object. Should be fixed now.

Jamiras: I agree with this assessment.

Jamiras (approved):

PR #1617, opened 2023-06-18
fix(globalRanking): don't include unofficial achievement count in total

wescopeland:
This PR resolves an issue on the global ranking page where unofficial achievements are included in the earned achievement counts in the table.

The points totals already don't include unofficial achievements - it stands to reason the counts shouldn't either.

**Before**
Screenshot 2023-06-18 at 10 23 28 AM

**After**
Screenshot 2023-06-18 at 10 23 13 AM


app/Helpers/render/leaderboard.php
```diff
@@ -468,13 +469,13 @@ function getGlobalRankingData(
         if ($info == 0) {
             if ($unlockMode == UnlockMode::Hardcore) {
                 $selectQuery = "SELECT ua.User,
-                        (SELECT COALESCE(SUM(CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END), 0) FROM Awarded AS aw WHERE aw.User = ua.User) AS AchievementCount,
+                        (SELECT COALESCE(SUM(CASE WHEN aw.HardcoreMode = " . UnlockMode::Hardcore . " AND ach.Flags = " . AchievementType::OfficialCore . " THEN 1 ELSE 0 END), 0) FROM Awarded AS aw JOIN Achievements AS ach ON aw.AchievementID = ach.ID WHERE aw.User = ua.User) AS AchievementCount,
```
Jamiras, line 472: This still doesn't match the value on the profile page because of the Event achievements.
https://github.com/RetroAchievements/RAWeb/blob/2ec66530e445a32b03e9b4505b6562c91260d1b3/public/userInfo.php#L57-L64

You'll want to join to GameData and exclude ConsoleIDs 100 and 101.

wescopeland: Thanks, I wasn't sure what the culprit was. I've updated those two lines and the counts are now matching. 

The line length is starting to bug me, but not sure what the best practice is on breaking it up into multiple lines, it seems as soon as I start doing it the syntax highlighting in VSCode stops functioning.

Jamiras: Just use indents. I'd probably do this:
```
                $selectQuery = "SELECT ua.User,
                        (SELECT COALESCE(SUM(CASE WHEN aw.HardcoreMode = " . UnlockMode::Hardcore . "
                                                   AND ach.Flags = " . AchievementType::OfficialCore . "
                                                  THEN 1 ELSE 0 END), 0)
                            FROM Awarded AS aw
                            JOIN Achievements AS ach ON aw.AchievementID = ach.ID
                            JOIN GameData as gd ON ach.GameID = gd.ID
                            WHERE aw.User = ua.User AND gd.ConsoleID NOT IN (100, 101)
                        ) AS AchievementCount,
                        COALESCE(ua.RAPoints, 0) AS Points,
                        COALESCE(ua.TrueRAPoints, 0) AS RetroPoints,
                        COALESCE(ROUND(ua.TrueRAPoints/ua.RAPoints, 2), 0) AS RetroRatio ";
```

wescopeland: Cool, I've made this adjustment in both places.

One thing I noticed while double checking to be sure the indents didn't break anything- it looks like the softcore achievements count on the user page and softcore leaderboard are also not matching. I am seeing the same thing in both Preview and Stage, so I'm assuming it's nothing new caused by this PR or V3.

Jamiras: Yep. Definitely something off with that calculation:
image

Although it was closer before:
image

(User page shows 81).

Jamiras: Got it. 
```
CASE WHEN aw.HardcoreMode = 0 AND ach.Flags = 3 THEN 1 ELSE -1 END
```
Is returning `-1`s for all unofficial achievements. The `AND ach.Flags=3` needs to be moved out of the case, and just put as a filter on the outer query:
```
                        (SELECT COALESCE(SUM(CASE WHEN aw.HardcoreMode = " . UnlockMode::Softcore . " 
                                                  THEN 1 ELSE -1 END), 0) 
                            FROM Awarded AS aw 
                            JOIN Achievements AS ach ON aw.AchievementID = ach.ID 
                            JOIN GameData as gd ON ach.GameID = gd.ID 
                            WHERE aw.User = ua.User AND gd.ConsoleID NOT IN (100, 101)
                            AND ach.Flags = " . AchievementType::OfficialCore . " 
                        ) AS AchievementCount,
```

wescopeland: Great work, Jamiras. It would've taken me forever to figure this out. I've pushed this change and set you as coauthor on that commit.

app/Helpers/render/leaderboard.php
```diff
@@ -468,13 +469,27 @@ function getGlobalRankingData(
         if ($info == 0) {
             if ($unlockMode == UnlockMode::Hardcore) {
                 $selectQuery = "SELECT ua.User,
-                        (SELECT COALESCE(SUM(CASE WHEN HardcoreMode = " . UnlockMode::Hardcore . " THEN 1 ELSE 0 END), 0) FROM Awarded AS aw WHERE aw.User = ua.User) AS AchievementCount,
+                        (SELECT COALESCE(SUM(CASE WHEN aw.HardcoreMode = " . UnlockMode::Hardcore . " 
+                                                   AND ach.Flags = " . AchievementType::OfficialCore . "
```
Jamiras, line 473: You should move this Flags out of the CASE for consistency.

wescopeland: 👍 Change applied in https://github.com/RetroAchievements/RAWeb/pull/1617/commits/75d1bb3c4f430f8e2f34fb58997b55d8dcba73c6

Jamiras (approved):

PR #1618, opened 2023-06-18
feat(game): allow devs to reassign game topic OPs on their claims

wescopeland:
This PR gives achievement developers the ability to reassign the game's forum topic original poster to themselves if they have an active claim on a given game.

**Why?**
Currently when something like this needs to happen, it goes to #admin-site-cleanup who then performs a thread transfer. However, this seems like more of a workaround, and now it is possible that a game's discussion is split between two forum threads instead of isolated into one.

**How?**
A new button has been added to the devbox. The button appears if the user has the Developer role and has an active claim on the game. When pressed, a confirmation dialog appears that if accepted, reassigns the game's forum topic OP to the active user.

https://github.com/RetroAchievements/RAWeb/assets/3984985/df9226cc-34bf-4c8a-9025-35a6f3861233

### DIFF
--- a/app_legacy/Helpers/database/forum.php
+++ b/app_legacy/Helpers/database/forum.php
@@ -216,7 +216,7 @@ function updateTopicOriginalPoster(string $newOriginalPoster, int $existingTopic
         ]);
     }
 
-    // Did the write operation work? 
+    // Did the write operation work?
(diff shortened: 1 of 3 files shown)


wescopeland:
I am going to rework this to be a new tool on the site admin panel.

Issue #1619, opened 2023-06-19
[Bug] [Problem] Text Handling exceptions: User Motto & Textarea inaccurate character limit


Alexdatadestroyer:
While writting such a long commentary on a ticket to give a feedback regarding an Achievement Bug, I found an exception on RA Website regarding text handling: It's caused when you put some Latin Characters on a Textarea on a Ticket Page or another page (probably) and when you put some Latin Characters on User Motto.

**How to Reproduce it?**
- Textarea: Just put any Latin Character like "á, ã, â, ü, ç, ñ, à, Á, Ã, Â, Ü, Ç, Ñ, À" and type a message up to 2000 characters to notice a mismatch between Character Counter and Message Submission being denied with an exception claiming that more than 2000 characters were used.
- User Motto: Type any 50-character message with Latin Characters. NO exception message is thrown: just the message is presented lacking some characters due to LCs being used. Regardless this issue, a "Changed!" message is shown after.

**1. Screenshots: Textarea**

- Character Counter showing that message is still valid to be sent (1992/2000):

![](https://i.imgur.com/LhcGnxS.png)

- Textarea being highlighted on Code Inspector on Edge Browser:

![](https://i.imgur.com/tNEdLg1.png)

- Textarea Message Submit Exception: More than 2000 characters used, but only 1992 was sent:

![](https://i.imgur.com/d8Ib5ZF.png)

**2. Screenshots: User Motto**

- Original message to be posted on User Motto:

![](https://i.imgur.com/SY9yzJQ.png)

- Message being cut due to 3 Latin Characters (Browser showed the entire message as a suggestion for autocompletion though):

![](https://i.imgur.com/K3GuzfE.png)

- Original message accepted ONLY with Latin Characters removed:

![](https://i.imgur.com/PD6aYFD.png)

**Further Investigation:**
[...]

PR #1620, opened 2023-06-19
fix: calculate correct input lengths when using diacritics

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1619.

**Root Cause**
The issue linked occurs due to how JavaScript counts characters in a string when using `String.prototype.length`. Normally, a character in JavaScript is assumed to be a single 16-bit unit. However, some characters such as diacritics and emojis actually use two of these units. In JavaScript, a diacritic like "ã" or an emoji is counted as one character, but in reality, they are composed of two or more 16-bit units.

When the JavaScript code interacts with the backend, which counts the bytes rather than the 16-bit JavaScript characters, these multi-unit characters cause validation discrepancies. The backend interprets these characters as multiple bytes, leading to a count that's higher than what's shown on the frontend. This mismatch can cause issues like hitting the maximum character limit prematurely or having a different count shown on the UI compared to what is actually stored in the MySQL database.

In the case of this specific issue, when a user includes diacritics or emojis in their input, the displayed count in the frontend doesn't match the actual count in the backend, leading to unexpected behavior and surprising error messages for the user.

**Summary of Changes**
* `initializeTextareaCounter()` has been migrated to TypeScript. I opted not to write tests for this migration because this should ultimately land in Blade/Alpine.js anyway. The tests would be throwaway code.
* `initializeTextareaCounter()` now properly counts the length of characters that use more than a single 16-bit unit. Diacritics are now properly counted as being two characters instead of one.
* I have fixed a bug in `initializeTextareaCounter()` where emoji inputs from Apple devices were ignored unless the user blurred away from the field.
* Per the feedback in the ticket, I've added a character counter to the user motto field on the settings page.
* Per the feedback in the ticket, I've slightly adjusted the validation message the backend throws when this error occurs.

Screenshot 2023-06-19 at 9 57 39 AM


Alexdatadestroyer:
That was fast, thanks for taking that seriously! xD

While reading this post, I found a [related case to this incident](https://stackoverflow.com/questions/18578207/strange-length-of-accent-as-%C3%A9-string-return-2). What is curious is: The more you search for it, you would likely find support telling how to treat these characters by removing them instead of counting properly as 2.

I didn't know that Javascript strings uses UTF-16 encode as default. So, this was, probably, a conflict between UTF-16 and UTF-8 from MySQL (or PHP)?

Alexdatadestroyer:
[Minor] Also, I found this part of the fix curious:

`'string' => 'Your :attribute is too long. It should be no more than :max characters.'`

For **User Motto**, it would show: "Your Motto is too long. It should be no more than 50 characters."
But for **Textarea**, it would show: "Your Body is too long. It should be no more than 2000 characters."

The user will understand what "Motto" means, but maybe not what "Body" means - unless if Textarea name attribute gets changed.

**NOTE:** This is just a matter of user-friendly case. Just a minor thing, but it deserves to be mentioned. I am already glad that the Exception was resolved overall!

lang/en_US/validation.php
```diff
@@ -86,7 +86,7 @@
         'array' => 'The :attribute must not have more than :max items.',
         'file' => 'The :attribute must not be greater than :max kilobytes.',
         'numeric' => 'The :attribute must not be greater than :max.',
-        'string' => 'The :attribute must not be greater than :max characters.',
+        'string' => 'Your :attribute is too long. It should be no more than :max characters.',
```
Jamiras, line 89: Why change this? It's inconsistent with the other messages.

Jamiras (commented):
Verified functionality. Left comment about validation message.

lang/en_US/validation.php
wescopeland: Agree- mostly at the request of the corresponding ticket:

> To finish this report, IF I am not asking too much, may we have the exception message a bit less technical, please? That "The Body must not be greater than 2000 characters" is ok for me, but it would be better (and more user-friendly) for just showing a message like "Your message is too LONG! It must not be greater than 2000 characters!" or similar. 

I have no strong opinion.

Jamiras: "Your Body is too long. It should be no more than 2000 characters." is not any clearer. I think it's the `:attribute` that is the most confusing.

wescopeland: Let's punt then; I'm not sure what side effects tweaking this further could have, and it may wind up growing to justify its own PR. I'm going to revert this change.

Jamiras (approved):

lang/en_US/validation.php
Alexdatadestroyer: In matter of changing validation message inconsistency, that's why I said that above:

> The user will understand what "Motto" means, but maybe not what "Body" means - unless if Textarea name attribute gets changed.

If I recall correctly, `:attribute` takes the `name` attribute of HTML element that was used on. So, just changing the Textarea name would suffice as the user will understand better what is going on.

Alexdatadestroyer:
In matter of changing validation message inconsistency, that's why I said that above:

> The user will understand what "Motto" means, but maybe not what "Body" means - unless if Textarea name attribute gets changed.

If I recall correctly, `:attribute` takes the `name` attribute of HTML element that was used on. So, just changing the Textarea name would suffice as the user will understand better what is going on. For Motto, it can stays as it is.

**(Posted this comment just now outside of a Reviewer Conversation because Github marked me as "Reviewer" which is I am not...)**

PR #1621, opened 2023-06-19
fix(achievementInfo): restyle unlocked date label for better mobile support

wescopeland:
This PR fixes a minor styling issue on the achievement info page that is most noticeable on mobile devices.

If a user has unlocked an achievement, the "Unlocked on..." label will now mirror what appears on the game page. For future state, it probably makes sense for both the game page and achievement info page to re-use the same "achievement entity (?)" component. This slight adjustment may make such a migration a pinch easier.

**Before**
Screenshot 2023-06-19 at 10 19 50 AM

**After**
Screenshot 2023-06-19 at 10 17 16 AM


wescopeland:
Not sure why the ESLint job refuses to run, but all should be good since there are no JS changes here.

Jamiras (approved):

Issue #1625, opened 2023-06-20
[Bug] [Problem] Text Handling exceptions #3: Profile Name + Account Status character explosion on User Card


Alexdatadestroyer:
Motivated by User Motto & Textarea bug report (#1619) which was resolved already (#1620) and a related issue reported later (#1624). Now, it's time to finish this Text Handling exception Trilogy: another old problem was noticed. However, this became reality when V2 was implemented in September 2022 compared to other 2 aforementioned cases.

Since when V2 was implemented, I noticed a character explosion between Profile Name and Account Status. This also happens when User Motto is really big: On majority of cases that I ever seen, UM sentence is get wrapped inside UC - which doesn't seem to be a problem. However, in cases where a User puts a big URL, it gets clipped and bugs the UC. Sometimes, these issues are fused together.

**How to Reproduce this?**

Sorry, there is NO such a way to reproduce this issue...

Unless if you create an Account with a Name than surpasses 12 or more... This also dependents of the Account Status of the person (Registered, Unregistered, Banned, Spam, Junior Developer, Developer, Site Moderator, Invalid Permission...). So the exact Profile Name size is impossible to guess. Maybe this is possible [to be reproduced here?](https://stage.retroachievements.org/)

**Screenshots taken from various Users (Some aren't bugged):**

- [ThatAmericanSlacker](https://retroachievements.org/user/ThatAmericanSlacker) UC:

![](https://i.imgur.com/xThnoFk.png)

- [BigWeedSmokerMan](https://retroachievements.org/user/BigWeedSmokerMan) UC (UM sentence wrapped):

![](https://i.imgur.com/UPym5Cd.png)

- [AngeloLeonhart](https://retroachievements.org/user/AngeloLeonhart) UC (UM sentence wrapped):

![](https://i.imgur.com/pLFmk5o.png)

- [AllMyHomiesHateJord](https://retroachievements.org/user/AllMyHomiesHateJord) UC:

![](https://i.imgur.com/kPU2eCI.png)

- [hardcoregamertop](https://retroachievements.org/user/hardcoregamertop) UC:

![](https://i.imgur.com/VnvsmhE.png)

- [Alexdatadestroyer](https://retroachievements.org/user/Alexdatadestroyer) UC Examples:

Normal:
![](https://i.imgur.com/FlW4m9d.png)

UM clipped due to URL as UM sentence:
![](https://i.imgur.com/uvgbWmz.png)

[...]

PR #1429, opened 2023-03-12
refactor(tooltip): migrate from wz_tooltip to Floating UI

wescopeland:
This PR refactors the site's tooltip implementation away from wz_tooltip. Tooltips are now implemented using Floating UI ([the successor to popperjs](https://opencollective.com/floating-ui/updates/popper-is-evolving-into-floating-ui)). 

[Floating UI Docs](https://floating-ui.com/)

**Why?**
1. **Modern and updated.** Floating UI is a more modern and more actively maintained library compared to wz_tooltip, which is an older library with limited updates. wz_tooltip has not received an update since the Bush Administration was in power. Migrating to Floating UI ensures better compatibility with modern browsers, new features, and performance improvements.
2. **Better documentation and community support.** Floating UI has a more extensive and well-maintained documentation, which makes our integration of it easier to maintain. Moreover, it has a larger community, meaning it's more likely to find support and resources online when new tooltip requirements arise.
3. **Accessible.** Floating UI has a built-in focus on accessibility, ensuring that tooltips and popovers are easily accessible to users with assistive technologies.
4. **Customizable and extensible.** Thanks to the library's headless nature and its middleware support, Floating UI provides more options for customization and extensibility than wz_tooltip, allowing us to easily create tooltips and popovers that match RAWeb's look and feel and achieve a consistent user experience.

**How?**
All wz_tooltip code, including its plugins, has been removed.

Floating UI tooltip code is implemented with TypeScript in _resources/js/tooltip_. I have also added numerous tests for our code that uses the library. Most of the code additions in this PR is actually just for test cases.

The intent is for tooltips to remain as close to functionally equivalent as possible to the existing implementation. This is meant to be a refactor, not an enhancement.

Therefore:
* Tooltips are still disabled on mobile iOS.
* User tooltips still experience weird spacing issues if there are too many characters with the username and role.
* There are still weird "skips" that happen on avatar elements, where if you move the cursor between the text and image elements of the avatar, the tooltip momentarily disappears and then reappears.
* Tooltip content can still be fetched either statically or dynamically.
* Tooltips are now attached to elements using an Alpine `x-init` directive.

Bundle size difference:
_app.js_ before: ~23Kb gzipped
_app.js_ now: ~29Kb gzipped

The bundle now appears larger, however this metric is misleading. wz_tooltip is currently being loaded directly in the document body on prod, and it is ~15Kb gzipped. In total, this change actually achieves a small payload savings.

[...]

PR #1629, opened 2023-06-21
Fix username/role cutoff in user cards

rzumer:
* Set 400px as the minimum width instead of fixed width
* Add a margin between the username and user role

Higher fixed width would work too but since roles are variable-length it seems saner to just give it a variable width, in practice it can't be longer than ~500px due to the username length cap and current role lengths.

Before
image
After
image

Tested only in the inspector, not a local environment because WSL v1 doesn't support Docker, Sail couldn't run and I gave up when populating the database failed probably because of some missing init step.

### DIFF
--- a/app_legacy/Helpers/render/user.php
+++ b/app_legacy/Helpers/render/user.php
@@ -94,7 +94,7 @@ function renderUserCard(string|array $user): string
     $lastLogin = $data['LastLogin'] ? getNiceDate(strtotime($data['LastLogin'])) : null;
     $memberSince = $data['Created'] ? getNiceDate(strtotime($data['Created']), true) : null;
 
-    $tooltip = "<div class='tooltip-body flex items-start gap-2 p-2' style='width: 400px'>";
+    $tooltip = "<div class='tooltip-body flex items-start gap-2 p-2' style='min-width: 400px'>";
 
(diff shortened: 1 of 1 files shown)


resources/css/card.css
```diff
@@ -8,6 +8,10 @@
   font-size: theme('fontSize.lg');
 }
 
+.usercardaccounttype {
+  margin-left: 1em;
+}
```
wescopeland, line 13: Please use a Tailwind class such as `ml-4` instead of adding a new class name.

wescopeland (changes requested):

resources/css/card.css
rzumer: OK, done

wescopeland (approved):

Alexdatadestroyer:
So, is this already a solution for #1625 that I reported one day earlier the date of this Pull Request? Amazing!

However, User Motto will give some issues when link is posted on it. I gave more details on my issue earlier that day...

wescopeland:
Hi @Alexdatadestroyer,

It is unlikely this change will be accepted into the project. Due to org permissions, I do not have the ability to dismiss my approval, but it's safe to disregard.

https://github.com/RetroAchievements/RAWeb/pull/1429 is an active PR containing a significant rework of site tooltip functionality. Once it is merged, I will have a subsequent PR that fixes numerous tooltip issues (including https://github.com/RetroAchievements/RAWeb/issues/1625)

PR #1631, opened 2023-06-23
feat: sort user achievements lists by awarded date

wescopeland:
This PR adds new sorting functionality to a user's "Achievements" page.

Achievements are now sorted by default in the order they were unlocked by the user. When that particular sort is applied (which can only happen organically by clicking on the "My Pages > Achievements" link), the "Added" and "Modified" columns are removed from the page to save space.

https://github.com/RetroAchievements/RAWeb/assets/3984985/fc9094c0-94f6-42c4-aee8-c33048a70624


wescopeland:
@Jamiras Conflicts on this PR have been resolved.

public/achievementList.php
```diff
@@ -109,6 +112,7 @@
         $sort6 = ($sortBy == 6) ? 16 : 6;
         $sort7 = ($sortBy == 17) ? 7 : 17;
         $sort8 = ($sortBy == 18) ? 8 : 18;
+        $sort9 = ($sortBy == 19) ? 20 : 19;
```
Jamiras, line 115: This should be 9 and 19 or the modulus 10 logic doesn't work, which causes the `*` to not appear when sorting by awarded ascending.

public/achievementList.php
```diff
@@ -196,13 +205,19 @@
...
-            if (!$mobileBrowser) {
+                if (!$mobileBrowser) {
+                    echo "<td>";
+                    echo "<span class='smalldate'>" . getNiceDate(strtotime($achDateModified)) . "</span>";
+                    echo "</td>";
+                }
+            } else {
                 echo "<td>";
-                echo "<span class='smalldate'>" . getNiceDate(strtotime($achDateModified)) . "</span>";
+                echo "<span class='smalldate'>" . getNiceDate(strtotime($achAwardedDate)) . "</span>";
```
Jamiras, line 220: This generates `01 Jan 1970, 00:00` if `$achEntry['AwardedDate']` is null (Achievements in Core Set, Awarded ascending).

public/achievementList.php
```diff
@@ -32,6 +32,9 @@
 $sortBy = (int) request()->input('s', 17);
 $achData = getAchievementsList($user, $sortBy, $params, $count, $offset, $flags, $dev);
 
+// Is the user looking at their own achievements list?
+$isOwnEarnedAchievementsList = $user !== null;
```
Jamiras, line 36: The only way $user would be null is if they're not logged in, and I don't think including the Created/Modified timestamps for that scenario gets us anything. Recommend removing them there too.

public/achievementList.php
wescopeland: Thanks, I didn't recognize this pattern but that totally makes sense. Should now be addressed.

public/achievementList.php
wescopeland: Yep! I didn't realize `AwardedDate` is actually nullable. Now these values will display "Unknown".

public/achievementList.php
wescopeland: Upon closer inspection it looks like I screwed up a bit of the conditional logic in my first pass of this PR. Hopefully it is buttoned up now.

The intent is for all achievement lists that are not the user's "My Unlocked Achievements" to be functioning as they do in prod (we can of course change this if you'd like). "Awarded" now only appears when the `p` query param is set to `1`.

wescopeland:
Something I wanted to get your opinion on because I am unsure: should the nav link label be changed? Part of me feels it should to be consistent with the others, part of me feels it shouldn't just because it's so long:

Screenshot 2023-07-02 at 9 02 57 AM

public/achievementList.php
```diff
@@ -216,8 +216,13 @@
                     echo "</td>";
                 }
             } else {
+                $renderAwardedDate = "Unknown";
```
Jamiras, line 219: A blank awarded date implies the achievement was not earned. Recommend using blank instead of Unknown.

Although, I can't get these to show up any more. The other changes seem to have made it impossible to list awarded dates for unearned achievements.

Jamiras (approved):

public/achievementList.php
wescopeland: I do have a tiny handful in my database where the `Date` value is set to `NULL`, so at some point it must have been possible (or just something is wrong with the DB dump):

Screenshot 2023-07-02 at 10 51 36 AM

Jamiras:
> should the nav link label be changed? Part of me feels it should to be consistent with the others, part of me feels it shouldn't just because it's so long

The "My" links are all related to dev'ing. The Achievements list is more related to the History link. If I would change anything, I'd just say "Unlocked Achievements". And maybe swap History and Following so History is closer to achievements and Following is closer to Messages.

wescopeland:
> Recommend using blank instead of Unknown.

This is now changed for the rare cases it might appear for earned achievements with a null Date value.

> If I would change anything, I'd just say "Unlocked Achievements". And maybe swap History and Following so History is closer to achievements and Following is closer to Messages.

I went ahead and made these adjustments too in my latest commit.

public/achievementList.php
Jamiras: I think you're counter-point has swayed my decision.

I can't make the list show unearned achievements, but I can make it show earned achievements with a null date.

wescopeland: Got it- just to confirm before I make a commit, should I set this string back to display "Unknown"?

Jamiras: Yes, please.

wescopeland: No problem, I've set this to display "Unknown" again in https://github.com/RetroAchievements/RAWeb/pull/1631/commits/c18a9897267ad67148c86a351ebe94f66edfaeed.

PR #1636, opened 2023-06-30
don't return locked achievements in API_GetUserSummary.RecentAchievements

### LINKED ISSUE #1615: RecentAchievements of API_GetUserSummary.php returns all achievements and number of recent achievements not working
**Describe the bug**

The `RecentAchievements` field of the endpoint `API_GetUserSummary.php` returns a list of recently played games including all achievements. Both unlocked and not unlocked achievements appear. Here also appear the games where I have no Achievements yet.

In addition, the parameter `a` seems to be ignored. At least I get significantly more Achievements back.

Is this the way it is supposed to be? 

**To Reproduce**
1. API call: `API_GetUserSummary.php?y={{key}}&u={{user}}`

**Expected behavior**

In the list of recent achievements, only those should appear that I actually received recently. Also, only as many as I specify via the parameter.

Jamiras:
fixes #1615


app/Helpers/database/player-game.php
```diff
@@ -187,6 +187,10 @@ function getUserProgress(string $user, array $gameIDs, int $numRecentAchievement
         $libraryOut['GameInfo'] = $gameInfo;
     }
 
+    // magic numbers!
+    // -1 = don't populate RecentAchievements field
+    // 0 = return all achievements for each game, with the unlocked achievements first ordered by unlock date
+    // >0 = return the N most recent unlocks across all games queried, grouped by game
```
Jamiras, line 193: -1 when called from API_GetUserProgress and API_GetUserRecentlyPlayedGames
0 when called from the user profile page, which needs the unlocked achievements
>0 when called from API_GetUserSummary, which shouldn't return the unlocked achievements

PR #1641, opened 2023-07-03
fix(viewtopic): add 64px offset for deep links [V3]

wescopeland:
This PR fixes an issue with opening deep links in forum threads where the top of the message is cut off by the navbar. To resolve this, I've set the `id` attributed element to an invisible div that is floating 64px away from the linked post.

The issue can be repro'd by opening (in a new tab) a deep linked forum post that is somewhere halfway through the thread, eg: viewtopic.php?t=20260&c=153328#153328

**Before**
Screenshot 2023-07-03 at 12 55 30 PM

**After**
Screenshot 2023-07-03 at 12 52 50 PM

### DIFF
--- a/app/Helpers/render/comment.php
+++ b/app/Helpers/render/comment.php
@@ -122,7 +122,7 @@ function RenderArticleComment(
         $class .= ' system';
     }
 
-    echo "<tr class='comment$class' id='comment_$commentID'>";
+    echo "<tr class='comment$class' id='comment_" . $commentID . "_highlight'>";
 
     $niceDate = date("j M Y ", $submittedDate);
     $niceDate .= date("H:i", $submittedDate);
@@ -131,6 +131,13 @@ function RenderArticleComment(
     $comment = nl2br($comment);
 
     echo "<td class='align-top py-2'>";
+
+    echo <<<HTML
+        <div class="relative">
+            <div class="absolute h-px w-px left-0" style="top: -74px;" id="comment_$commentID"></div>
+        </div>
+    HTML;
+
     if ($user !== 'Server') {
         echo userAvatar($user, label: false);
     }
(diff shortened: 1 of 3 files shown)


Jamiras:
I've seen this behavior on achievement comment links too: https://retroachievements.org/achievement/50651#comment_2238642

wescopeland:
Me too, let me see if I can fix there as well.

wescopeland:
Fixing this for game/achievement comments definitely took some doing due to them living in table layouts, but I think they're in a better place now with the changes I just committed.

It would be ideal if, like forum posts, the highlight could be applied server-side rather than relying on JS. I think the distinction is, for the purpose of deep links, comments are using hash params and forum posts are using query params. This is probably worth changing at some point in the future.

Jamiras (approved):

Jamiras:
> It would be ideal if, like forum posts, the highlight classname could be applied server-side rather than relying on JS.

Using Javascript eliminates the need to pass the anchor to the server as an additional parameter.

Forum with anchor: http://localhost/viewtopic.php?t=8398&c=169359#169359 (note `c=X` and `#X`)
Comment with anchor: http://localhost/achievement/9439#comment_2187868 (note only `#X`)

wescopeland:
> Using Javascript eliminates the need to pass the anchor to the server as an additional parameter.

I was thinking maybe it could be parsed from the hash parameter and the `c=X` could be eliminated. I am spitballing and haven't explored what a potential implementation of that might look like. I fear that handling this sort of thing purely client-side can result in a weird lag on slow connections for when the highlight is actually displayed on screen (due to the delay in downloading the JS bundle). 

Since we can probably already know ahead of time which tr/div to highlight based on the value of `#X`, applying the highlight ahead of time probably results in better perceived performance than doing it just-in-time.

PR #1642, opened 2023-07-04
feat: migrate user tooltips to blade with minor ux improvements

wescopeland:
This PR migrates the user tooltip logic and template to an anonymous Blade component, and prepares a base component (_tooltip-card.blade.php_) for the same to happen later for the other tooltip kinds.

Additionally, there are several tiny QoL improvements made to the user tooltips that fix a few longstanding bugs.

**UX Overview**
* Similar to the forum topic viewing changes introduced in V3, the role now only appears if the user is >= Junior Developer. This also means we are no longer broadcasting if a user is banned.
* The user role display has been reworked so the role label is no longer "crushed" by long usernames or lack of space.
* Points now has localized number separators.
* Site rank now has localized number separators, is prefixed with "#", and if the user is outside the Top 100, the "Top N%" is shown similar to user profiles.
* Last Activity now uses a timeago-style date.

**Technical Implementation**
* A new `<x-tooltip-card />` component has been built that has a slot for tooltip card content.
* A new `<x-cards.user />` component has been built which now houses all the logic and template for user tooltip cards.
* `renderUserCard()` now simply returns a `Blade::render()` call to the new `<x-cards.user />` component.

**Before & After Comparisons**
Upper is before, lower is after.

Screenshot 2023-07-04 at 4 15 29 PM

Screenshot 2023-07-04 at 4 15 52 PM

Screenshot 2023-07-04 at 4 16 01 PM

Screenshot 2023-07-04 at 4 16 08 PM

Screenshot 2023-07-04 at 4 16 12 PM

Resolves https://github.com/RetroAchievements/RAWeb/issues/1625.


luchaos (approved):
Nice, I like that you chose "card" for the component's name 👍

PR #1649, opened 2023-07-07
fix(forum): visual tweaks

wescopeland:
https://discord.com/channels/476211979464343552/1026595325038833725/1126985746293735524

### DIFF
--- a/resources/views/community/components/forum/post-author-box.blade.php
+++ b/resources/views/community/components/forum/post-author-box.blade.php
@@ -8,12 +8,10 @@ use Illuminate\Support\Carbon;
     'authorUserName',
     'authorJoinDate' => '',
     'authorPermissions' => Permissions::Unregistered,
+    'isAuthorDeleted' => false,
 ])
 
 <?php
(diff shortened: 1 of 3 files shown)


resources/views/community/components/forum/post-author-box.blade.php
```diff
@@ -1,5 +1,4 @@
 <?php
-use App\Site\Enums\Permissions;
```
Jamiras, line 2: This is still used on line 28

wescopeland: Definitely did not intend to do that.

It looks like php-cs-fixer autodeleted it, and [there's no way to stop it from doing that](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/3846#issuecomment-401324204) unless I use `Permissions` somehow in a `<?php ?>` block. The latest commit to the branch does that, but seems like a fairly annoying workaround.

Jamiras (approved):

Issue #1626, opened 2023-06-20
[Minor] [Duplicate] Which of these do you think you can get? (This link should point to a Random Set originally)


Alexdatadestroyer:
This idea came to my mind back in 2021-2022, but I could make it materialize just now (19-20 Jun 2023): The link on Welcome Area at Main Page only points to [Sonic the Hedgehog 1 MD set](https://retroachievements.org/game/1). I don't think this was intended since the own sentence and entire paragraph that Scott wrote years ago gives a hint of a user clicking on this link and accessing a Random Set. This is a long time pendent feature!

This feature already exists, but it can be found on dropdown "Games" menu only. However, I suppose that it should be also available on Welcome Area to preserve original intention and for being available for non-logged users. When user is logged in, Welcome Area is not visible. So, the alternative of it being available on "Games" dropdown menu comes in handy.

**Screenshot of a non-logged user: This shows that the link still points to Sonic 1 MD set since 2013:**

![](https://i.imgur.com/DVpo9Ui.png)

**Exception Level: N/A** -- Just asking for a cloning an existing feature to such a different place on RA Website.

Alexdatadestroyer:
At this point with this already implemented, it useless to post the following message on main post... However, I would like to share what I imagined in past for a feature like this: The idea was to find a Random Set which follow these criteria:

- Non-Mature sets. Out of [[Theme - Mature] hub](https://retroachievements.org/game/8323);
- NO [Hubs (Console ID: 100)](https://retroachievements.org/gameList.php?c=100) and [Event (Console ID: 101)](https://retroachievements.org/gameList.php?c=101) sets;
- Console ID must be valid: Console must have been already released on RA Web;
- The Random Set selected must have valid Achievements online (Core; NOT Unofficial ones);
- The Random Set selected must not be any Subset. Subsets are custom challenges which would, probably, require the user to have their experience on Main Set first before dealing with them (and preparation on how to play them - RAPatches, Floating IPS, etc...)
- The Random Set selected must provide a Mastery Badge (Achievements >= 6) to avoid discouraging users from enlisting on RetroAchievements. A newcomer user wouldn't understand at first why a Mastery Badge wasn't provided after completed a Set with less than 6 Achievements - which can be disheartening.

wescopeland:
For the welcome/splash message, I do think it makes sense for the URL to point to one of the most popular sets on the site. Perhaps it could pick one randomly from the top 20? But I do think it should point to a game users will generally recognize.

[...]

PR #1653, opened 2023-07-08
feat(home): update welcome and faq/docs journey

wescopeland:
This PR makes slight modifications to the "Welcome!" text that appears if a user is not logged in. Additionally, a new component is shown if the user is logged in but their account is less than 1 month old. This new component guides the user towards the user docs FAQ page.

**Before**
Screenshot 2023-07-08 at 4 31 29 PM

**After**
Screenshot 2023-07-08 at 4 28 11 PM

Screenshot 2023-07-08 at 4 27 53 PM

### DIFF
--- a/resources/views/content/getting-started.blade.php
+++ b/resources/views/content/getting-started.blade.php
@@ -1,0 +1,15 @@
+<x-section>
+    <div class="bg-embed p-4 border-text rounded">
+        <h2>{{ __('Getting Started') }}</h2>
+
+        <p class="mb-2">
+            We're excited to have you here! We know you might have some questions about
+            hardcore mode, RetroPoints (white points), or which emulators to use. Don't worry,
+            we've got you covered! Check out our
+            <a href="https://docs.retroachievements.org/FAQ/" target="_blank" rel="noreferrer">
+                comprehensive FAQ
+            </a>
+            to get started. Happy gaming!
+        </p>
+    </div>
+</x-section>
--- a/resources/views/content/welcome.blade.php
+++ b/resources/views/content/welcome.blade.php
@@ -1,15 +1,17 @@
 <x-section>
-    <x-section-header>
-        <x-slot name="title"><h2>{{ __('Welcome!') }}</h2></x-slot>
-    </x-section-header>
-    <p class="mb-2">
-        Were you the greatest in your day at Mega Drive or SNES games? Wanna prove it? Use our
-        <a href="{{ route('download.index') }}">modified emulators</a> and you will obtain achievements and badges as
-        you play! Your progress will be tracked so you can compete with your friends to complete all your favourite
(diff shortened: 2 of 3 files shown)


wescopeland:
I've made a slight alteration to the "Which of these do you think you can get?" hyperlink. Now, a random game from the following array will be picked. I have no strong feeling or opinion about any of these games and would be happy to substitute any of them at the web team's discretion (or remove this functionality altogether). I aimed for two games on the most popular consoles that are likely to be well-known by a large subset of incoming users.

```php
$randomGameIds = [
    1, // Sonic the Hedgehog (Mega Drive)
    3, // Streets of Rage 2 (Mega Drive)
    1451, // Mega Man 2 (NES)
    1995, // Super Mario Bros. 3 (NES)
    335, // Legend of Zelda, The: A Link to the Past (SNES)
    446, // Donkey Kong Country 2: Diddy's Kong Quest (SNES)
    724, // Pokemon Red | Pokemon Blue (GB/C)
    5371, // Legend of Zelda, The: Link's Awakening DX (GB/C)
    10078, // Mario Kart 64 (N64)
    10210, // Banjo-Kazooie (N64)
    10434, // Crash Bandicoot (PSX)
    11242, // Final Fantasy VII (PSX)
    11191, // Pitfall! (Atari 2600)
    2831, // Metal Gear Solid 3: Subsistence (PS2)
    2721, // Dragon Quest VIII: Journey of the Cursed King (PS2)
];
```

Resolves https://github.com/RetroAchievements/RAWeb/issues/1626.

Jamiras (approved):

Alexdatadestroyer:
This is already aproved, but I think it's worth a commentary:

I don't know if there is already a way to check which games are being more accessed or played on RA Website or not. If not, I think we should provide a way because this will make this Array dynamic and random selection more interesting - which would, also, mean that we wouldn't need to update that Array manually over and over and neither having the same games listed there all time. On other side of the coin, there are more overrated games like Shadow of the Colossus, Guitar Hero 3: Legends of Rock, Okami, Sonic Adventure and others that aren't being too much accessed like it was expected to... - these ones might be inserted into Array manually in order to give them proper respect and audience.

As the main idea is to create an Array filled with games that users will recognize easily, it should be contain ONLY original games. I forgot to [mention here before](https://github.com/RetroAchievements/RAWeb/issues/1626#issuecomment-1598108105), but no Unlicensed games and neither Hacks.

For now, as a side effect, it's expected for Sonic the Hedgehog 1 (MD) page not being too excessively accessed like before. Sonic 1 (MD) and Super Mario 64 (N64) used to be pretty heavier on website due to excessive amount of users accessing those pages. Super Mario World was another one. Those pages used to take around 6-11 seconds to load here instead of 2-4 like others.

15 random games to choose. This is a nice start! xD

PR #1656, opened 2023-07-09
feat: add preference for showing absolute dates on forum posts

wescopeland:
This PR adds a new `websitePrefs` value for always showing absolute UTC timestamps on forum posts.

**UX Highlights**
* A small new section has been added on the user settings page which has a table containing a single row housing this new preference, titled "Show absolute dates on forum posts".
* When checked, timeago-style timestamps are no longer used on forum posts or recent forum activity displays. The UTC timestamp is shown instead.

**Technical Implementation**
* `Forum_ShowAbsoluteDates` has been added to the `UserPreference` enum.
* _update-notification.php_ has been renamed to _update-preferences.php_.
* Anywhere forum content is shown with a timestamp, a conditional executes to determine if that timestamp should be timeago-style or absolute-style.

**Screenshots**
Screenshot 2023-07-09 at 9 51 06 AM

Screenshot 2023-07-09 at 9 50 48 AM

Screenshot 2023-07-09 at 9 50 57 AM

Screenshot 2023-07-09 at 9 50 42 AM


public/controlpanel.php
```diff
@@ -225,7 +231,24 @@ function confirmEmailChange(event) {
                     </tr>
                     </tbody>
                 </table>
-                <img id='loadingicon' style='opacity: 0; float: right;' src='<?= asset('assets/images/icon/loading.gif') ?>' width='16' height='16' alt='loading icon'/>
+                <img id='loadingicon-1' style='opacity: 0; float: right;' src='<?= asset('assets/images/icon/loading.gif') ?>' width='16' height='16' alt='loading icon'/>
+            </div>
+
+            <div class='component'>
+                <h3>Settings</h3>
```
Jamiras, line 238: Since you're adding a new section here, you should probably move the Mature Content warning into it.

resources/views/community/components/forum/post-comment-meta.blade.php
```diff
@@ -6,9 +6,18 @@
...
 use Illuminate\Support\Carbon;
 
-$formatMetaTimestamp = function (string $rawDate): string {
+/** @var ?User $user */
+$user = auth()->user();
+$preferences = $user?->websitePrefs ?? 0;
+
+$formatMetaTimestamp = function (string $rawDate, int $preferences = 0): string {
+    if ($preferences && BitSet($preferences, UserPreference::Forum_ShowAbsoluteDates)) {
+        return $rawDate;
```
Jamiras, line 18: It would probably be better to show the non-relative, but still reformatted date.

Setting on:
image

Setting off:
image

If implemented, the other occurrences should also be updated. Consider using the `getNiceDate` helper function.

image

public/controlpanel.php
wescopeland: Agreed, should be addressed in the latest change.

Jamiras: Sounds reasonable. Maybe `Site_SuppressMatureContentWarning` to match the text you used on the settings page?

wescopeland: Handled in https://github.com/RetroAchievements/RAWeb/pull/1656/commits/dfebc91447b224d96acc164d736ea0df5fec0ea3. 

Note that in this commit I removed the value from _NotificationPreferences.php_ altogether [here](https://github.com/RetroAchievements/RAWeb/pull/1656/files#diff-90a5d2802a1107b7dd3334575f3a229d0347615bcc109d483c6aef5bb748154f). I don't believe this is something we'd want to send out user notifications for, though I am happy to back out that change.

public/controlpanel.php
```diff
@@ -27,10 +27,11 @@ function RenderUserPref(
     int $websitePrefs,
     int $userPref,
     bool $setIfTrue,
-    ?string $state = null
+    ?string $state = null,
+    ?int $targetLoadingIcon = 1
```
Jamiras, line 31: I don't believe this is nullable

wescopeland: You're right, this value should never be null. This was me getting the syntax for nullable and optional confused. Resolved in https://github.com/RetroAchievements/RAWeb/pull/1656/commits/c1a9d2e7df133b3896b5a27223795903087f6845.

Jamiras (approved):

PR #1658, opened 2023-07-09
feat(home): redesign achievement of the week

wescopeland:
This PR redesigns the Achievement of the Week component on the home page, and makes some changes to `HomeController` to better facilitate supporting different kinds of event widgets on the home page in the future.

**Before**
Screenshot 2023-07-09 at 4 47 00 PM

**After**
Screenshot 2023-07-09 at 4 46 08 PM

Thanks @Jamiras for design feedback.

---

https://github.com/RetroAchievements/RAWeb/assets/3984985/2c603396-739b-4bbf-9715-e9987cf8248f


resources/views/community/components/event/aotw.blade.php
```diff
@@ -1,31 +1,49 @@
...
+        <a href="{{ route('game.show', $achievement->GameID) }}" class="group flex gap-x-2 text-text transition-transform hover:translate-x-0.5">
+            <img src="{{ media_asset($game->ImageIcon) }}" alt="Achievement of the week game badge" width="32" height="32" class="w-8 h-8">
+            <div class="-mt-1">
+                <p class="font-semibold mb-0.5 text-xs text-link group-hover:text-link-hover">{!! $renderedGameTitle !!}</p>
+                <div class="flex items-center gap-x-1">
+                    <img src="{{ $gameSystemIconUrl }}" width="18" height="18" alt="Console icon">
+                    <span class="block text-xs tracking-tighter">{{ $consoleName }}</span>
+                </div>
+            </div>
+        </a>
```
wescopeland, line 45: Something like this may eventually be worth moving into `achievementAvatar()` as an optional rendering mode, almost in the vein of some of the work happening in https://github.com/RetroAchievements/RAWeb/pull/1500. There are a few places elsewhere on the site this design (console on 2nd line with icon) makes sense, though it doesn't make sense universally.

resources/views/community/components/event/aotw.blade.php
```diff
@@ -1,31 +1,49 @@
...
 
-$staticData = StaticData::first();
+<div class="component">
+    <h3>Achievement of the Week</h3>
 
-if ($staticData === null) {
-    return;
-}
+    <div class="bg-embed p-4 rounded border border-embed-highlight">
+        <a href="{{ route('achievement.show', $achievement->ID) }}" class="group flex gap-x-2 mb-3 text-text transition-transform hover:-translate-y-0.5">
```
Jamiras, line 25: I don't like things moving when I hover over them. I assumed this was a bug until I saw the code intentionally doing it.

https://github.com/RetroAchievements/RAWeb/assets/32680403/88c72ff7-7bb6-4a56-b8d5-25fa5a1f07a1

resources/views/community/components/event/aotw.blade.php
```diff
@@ -1,31 +1,49 @@
...
+            <img src="{{ $achievementIconSrc }}" alt="Achievement of the week badge" width="64" height="64" class="w-16 h-16">
+            <div>
+                <p class="font-semibold leading-4 text-link group-hover:text-link-hover">{{ $achievement->Title }}</p>
+                <p class="text-xs mb-1">5 <span class="TrueRatio">(12)</span> Points</p>
+                <p class="text-xs">{{ $achievement->Description }}</p>
```
Jamiras, line 30: I'm not convinced this behavior is desirable either. Yellow text is a link. Blue text is supplemental. I wouldn't expect to be able to open a new set by clicking on the Type data. Would it make sense to limit the hyperlink to just the Title?

resources/views/community/components/event/aotw.blade.php
```diff
@@ -1,31 +1,49 @@
...
+    <h3>Achievement of the Week</h3>
 
-if ($staticData === null) {
-    return;
-}
+    <div class="bg-embed p-4 rounded border border-embed-highlight">
+        <a href="{{ route('achievement.show', $achievement->ID) }}" class="group flex gap-x-2 mb-3 text-text transition-transform hover:-translate-y-0.5">
+            <img src="{{ $achievementIconSrc }}" alt="Achievement of the week badge" width="64" height="64" class="w-16 h-16">
+            <div>
+                <p class="font-semibold leading-4 text-link group-hover:text-link-hover">{{ $achievement->Title }}</p>
```
Jamiras, line 28: Do we want to process tags on the title?
image

resources/views/community/components/event/aotw.blade.php
wescopeland: No worries at all, this should now be addressed.

resources/views/community/components/event/aotw.blade.php
wescopeland: My (loosely held) opinion is the bigger hyperlink click/tap target is a better experience, however I will defer to your judgment. As of the latest commit, only the title and image are hyperlinked.

resources/views/community/components/event/aotw.blade.php
wescopeland: We do indeed, good catch. This should also be addressed now.

resources/views/community/components/event/aotw.blade.php
Jamiras: Having the console text be linked, but not the achievement description seems inconsistent. I'd almost expect clicking on the console name to take me to a list of games for the console.

I don't use the site on mobile. Is it really that hard to click on the text? I'm going to defer judgment to @luchaos.

luchaos: I don't have that strong of an opinion on this.

My guts too tell me that with the current design the icon and title should be linked to the game while the console name should be linked to the console page.
Whereas if it should act as a button where everything is a label (we don't have those yet afaik) then it should look like a button, too? But maybe that's going too far as it would break the current look.

luchaos: Thinking ahead, avatars for our resources are in need of an overhaul in general. In this case I probably would've gone with a much smaller icon and inline game title + console variant as it's important but still just metadata for the achievement.

wescopeland: >Thinking ahead, avatars for our resources are in need of an overhaul in general.

I agree. Mentioned here: https://github.com/RetroAchievements/RAWeb/pull/1658#discussion_r1258121986 This sort of design is dogfooding a potential alternative for game avatar UX.

It sounds like the path forward on this PR is:
* The console name should link to the console game list page.

wescopeland: Feedback is now addressed in https://github.com/RetroAchievements/RAWeb/pull/1658/commits/33ccf23bc61440d968b9433974714096ef1277f8.

The console name now links to the system page. Additionally, points and retro points are no longer hardcoded 😵‍💫

luchaos (approved):

Issue #1610, opened 2023-06-10
Web API case sensitive for username in user related queries


gorocz:
**Describe the bug**
When I try to retrieve user related data from the Web API (e.g. from API_GetGameInfoAndUserProgress.php), using different capitalization of my username retrieves data several days old (e.g. my username "gorocz" returns current data but "Gorocz" retrieves data from 4 days ago).
Authorization doesn't have a problem with capitalization of username though.

**To Reproduce**
1. Send a Web API request to API_GetGameInfoAndUserProgress.php for game 3528, user "Gorocz"
2. Do the same with user "gorocz"
3. Retrieved data is different
(This reproduction is obviously not consistent forever, since it relies on having recent progress in a specific game, but the screenshots below should be self-explainatory)

**Expected behavior**
Either the username parameter for the query should be case insensitive, or an error should be returned for incorrect username.

**Screenshots**
![picture](https://github.com/RetroAchievements/RAWeb/assets/33903218/55a163d6-e5d2-4376-89c3-de5da515703c)
![picture](https://github.com/RetroAchievements/RAWeb/assets/33903218/968fee71-5532-4fde-87b5-207e13eb628d)

**Additional context**
I know this is a trivial issue to solve from the side of the API's user, but since data was returned, just old, it was not easy to figure out where the issue was, so a fix would help any people in the future who could have this same problem.

Jamiras:
This is most likely the result of the caching added in #1486. Per-game user unlocks are cached for 7 days at a time. Unlocking a new achievement will expire the cached record, but only for the correctly cased username. The secondary cache entry generated by querying the incorrectly cased username will still contain the old data until it expires after 7 days.

The fix should be fairly trivial - just case-correct the username before doing the cache lookup/store, but that will add another query to the call. Another option would be to normalize the case for the cache key.

PR #1659, opened 2023-07-10
fix(api): lowercase username before using as cache key

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1610 with the suggested solution:

>Another option would be to normalize the case for the cache key.


app/Helpers/database/player-game.php
```diff
@@ -262,7 +262,9 @@ function expireUserAchievementUnlocksForGame(string $user, int $gameID): void
 
 function getUserAchievementUnlocksForGame(string $user, int $gameID, int $flags = AchievementType::OfficialCore): array
 {
-    return Cache::remember("user:$user:gameUnlocks:$gameID:$flags",
+    $normalizedUser = strtolower($user);
```
Jamiras, line 265: you also need to normalize user when expiring the entries (just a few lines up)

Jamiras (commented):
Do we also want to normalize the username for the avatar tooltip (`user:$user:card-data`), and/or the recently played games (`user:$user:recentGames`)?

I assume it would be better to case-correct the input coming into the API. That would also allow us to validate the existence of the user.

wescopeland:
>I assume it would be better to case-correct the input coming into the API. That would also allow us to validate the existence of the user.

I agree that case-correcting user input at the API level would be a reasonable approach to ensure consistency and also to validate user existence. However, I'd like to raise a concern related to potential unintended consequences.

Currently, several code paths lead to `getUserAchievementUnlocksForGame()`, including `getGameMetadata()`. Assuming I am interpreting correctly, I believe this proposed change could inadvertently lead to the creation of superfluous cache entries. To maintain consistent cache keys, I believe we would need to case-correct the user input at every single point leading to a user-driven `Cache::remember()` call.

Example:

```php
function generateGameForumTopic(string $user, int $gameID, ?int &$forumTopicID): bool
{
  // ...
  getGameMetadata($gameID, strtolower($user), $achievementData, $gameData);
}

expireUserAchievementUnlocksForGame(strtolower($user->User), $affectedGameID);
```

This approach would place the burden on the callers of these functions to always case-correct the user input.

To avoid this issue, I'd like to propose a different pattern. Rather than using magic strings as cache keys, we could create a utility/helper function to construct cache keys. This function would take care of case correction and ensure a predictable, consistent key format, like so:

```php
// Before
return Cache::remember("user:$user:gameUnlocks:$gameID:$flags", ...);

// After - note: i'm not set on this function api, this is just for demo purposes
$cacheKey = constructCacheKey(
  "user",
  $user,
  "gameUnlocks",
  [$gameID, $flags],
)
```

Now before calls to `Cache::remember()` or `Cache::forget()` are made, they can utilize this util to construct predictable keys which are normalized/case-corrected. It is possible I've overlooked something, I'm not so well-versed on the `Cache` API, however I would appreciate your thoughts/feedback on this idea.

Jamiras:
> Currently, several code paths lead to getUserAchievementUnlocksForGame(), including getGameMetadata(). Assuming I am interpreting correctly, I believe this proposed change could inadvertently lead to the creation of superfluous cache entries.

The original bug is the result of superfluous cache entries.

Username is used as part of the following keys: `user:$user:card-data`, `user:$user:recentGames`, `user:$user:canTicket` and  `user:$user:rank`. The username should probably be normalized for all of these keys. Having a generating function seems like a reasonable solution.

All of the other keys use numeric ids (`game:$id:card-data`, `game:$id:top-achievers`, `achievement:$id:card-data`, and `rankedUserCount:$type`).

wescopeland:
Following feedback & discussion [here](https://discord.com/channels/476211979464343552/1026595325038833725/1129772512943812668), I have implemented a new support class: `CacheKey`.

All mentioned keys where username is used should now be utilizing this service to generate the key.

app/Support/Cache/CacheKey.php
```diff
@@ -0,0 +1,82 @@
...
+    public static function buildUserCardDataCacheKey(string $userName): string
+    {
+        return self::buildNormalizedUserCacheKey($userName, "card-data");
+    }
+
+    /**
+     * @param string $userName the name of the user for which the cache key is being constructed
+     * @param int $gameID the target game ID for the unlocks data
+     * @param int $flags The type of achievement flag which should correspond to values in the `AchievementType` enum.
+     *                   3 for 'OfficialCore' (default), 5 for 'Unofficial'.
```
wescopeland, line 23: deptrac got really unhappy when I tried to pull in `AchievementType` into this class. This also occurred further down for `RankType`.

Documentation like this felt like the best compromise to still preserve developer experience.

app/Support/Cache/CacheKey.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Support\Cache;
+
+class CacheKey
+{
+    public static function buildUserCanTicketCacheKey(string $userName): string
```
luchaos, line 9: Most of the code uses `$username` - is `$userName` more correct? I usually use the former

app/Support/Cache/CacheKey.php
```diff
@@ -0,0 +1,82 @@
...
+     * @param string $keyKind the kind of the cache key
+     * @param array  $params optional parameters for the cache key
+     *
+     * @return string the constructed cache key
+     */
+    private static function buildNormalizedUserCacheKey(string $userName, string $keyKind, array $params = []): string
+    {
+        $normalizedUserName = strtolower($userName);
+
+        $cacheKey = "user:$normalizedUserName:$keyKind";
```
luchaos, line 74: It seems we have quite the mix of cases in our code already - initially we went for kebap-case for cache keys (`card-data`), now we have some camelCase in there, too.
I think we should at least normalize cache keys by lowercasing it fully. Not perfect but one option less to mistype it. What do you think?

wescopeland: I agree it makes sense to keep one form of casing. At the same time, I didn't want to break any existing caching that may already be set up. What do you think would be best?

app/Support/Cache/CacheKey.php
luchaos: I see, deptrac only allows for the `Site` layer to be used in `Support`, apart from that `Support` is assumed to be shared but knowing nothing about the other layers 🤔 

Works like this, too, for now. That achievement type flag is a bit of a mess throughout the code, literally. It could just be replaced with a boolean here imo and add a suffix if it's set.

app/Support/Cache/CacheKey.php
wescopeland: No problem at all, changed to `$username` in https://github.com/RetroAchievements/RAWeb/pull/1659/commits/b8cdfcf5fee4355ad19bbd324d3308c9b032de44.

app/Support/Cache/CacheKey.php
Jamiras: As long as the changes are consistent between reads/writes/deletes, it doesn't matter what the old keys were - they'll be purged when they expire.

app/Support/Cache/CacheKey.php
Jamiras: `username` is more correct. It's more well-understood to represent an account. `userName` infers `user name`, which is the user's name, and typically refers to their real name.

app/Support/Cache/CacheKey.php
wescopeland: Cases have been corrected in https://github.com/RetroAchievements/RAWeb/pull/1659/commits/01bbb1c2d5a090e2536286aaa68e04990aee8883.

app/Support/Cache/CacheKey.php
wescopeland: I've replaced this with a boolean as recommended in https://github.com/RetroAchievements/RAWeb/pull/1659/commits/01bbb1c2d5a090e2536286aaa68e04990aee8883.

This still leaves the implicit use of `RankType` on line 36, which has three potential states (hardcore, softcore, trueratio). Not sure if we want to do anything about that one or not.

resources/views/community/components/cards/user.blade.php
```diff
@@ -28,9 +29,13 @@
 }
 
 if (empty($userData)) {
-    $userData = Cache::store('array')->rememberForever('user:' . $userName . ':card-data', function () use ($userName) {
-        return User::firstWhere('User', $userName);
-    });
+    $userData = Cache::store('array')->remember(
+        $userCardDataCacheKey,
```
Jamiras, line 33: Why pass this in instead of just calculating it here?

Jamiras (approved):

resources/views/community/components/cards/user.blade.php
wescopeland: My original implementation did precisely that, however deptrac throws an error if this Blade component attempts to directly use any class from `Support`. Slightly related: https://github.com/RetroAchievements/RAWeb/pull/1659#discussion_r1264519669

Issue #1583, opened 2023-05-12
Game List game tooltip sometimes won't display console name


laytonloztew:
Go to a game list [https://retroachievements.org/gameList.php](https://retroachievements.org/gameList.php) and hover over a game's name or icon. Sometimes the console name won't be printed underneath the game name. If it does display, hovering over other games won't display their console name. Refreshing the page resets this, so the first game hovered may display the console name but others will not.

![ra_tooltipbug1](https://github.com/RetroAchievements/RAWeb/assets/12983268/a556ced4-5761-4e04-b073-33f42b58b036)
![ra_tooltipbug2](https://github.com/RetroAchievements/RAWeb/assets/12983268/3e0765fe-1e61-43ea-8c25-4e6236376781)

I first observed this on my dev game list [https://retroachievements.org/gameList.php?d=Layton](https://retroachievements.org/gameList.php?d=Layton) Seems to be any page using gameList.php, other pages don't exhibit this behavior. 

Windows 11
Chrome 113.0.5672.93 (Official Build) (64-bit)

wescopeland:
This seems to have been introduced by https://github.com/RetroAchievements/RAWeb/pull/1156. I am investigating if it's possible to keep the console name stripped but still pass it to the tooltip on the page.

PR #1660, opened 2023-07-10
fix(gameList): add missing consoles to tooltips

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1583.

**Root Cause**
In https://github.com/RetroAchievements/RAWeb/pull/1156, console names were removed from game titles on the game list. This was done by `unset($gameLabelData['ConsoleName']);` and `$gameLabelData` was then passed to `gameAvatar()` immediately afterwards. This had the intended effect of removing console names from the list items, but had a side effect of removing the console name from their tooltips because the tooltip itself is also generated by `gameAvatar()`.

**Solution**
`gameAvatar()`'s `$title` variable can now be optionally overridden by the caller as a new argument. In this case, we pass it a custom title of the game name, excluding the console name.


luchaos (approved):
Seems to solve the issue.
However, it feels like the avatar helper functions received quite a lot of exceptions ever since we introduced them.
With moving to blade avatar components we can hopefully reduce that again.

wescopeland:
> However, it feels like the avatar helper functions received quite a lot of exceptions ever since we introduced them.
With moving to blade avatar components we can hopefully reduce that again.

I agree, they all feel quite busy and difficult to reason about. They are probably good candidates for a near-term Blade migration since they are so ubiquitous in the app.

Issue #1090, opened 2022-08-13
Related Games self entry


Tsearo:
Adding or removing a game to it's own related games adds duplicate messages
![image](https://user-images.githubusercontent.com/4047771/184509098-80853861-86de-4eab-a891-3655da8a6dff.png)

A simple fix to this would be to add a check in this foreach in `modifyGameAlternatives` to ignore when $id is $gameID
```
foreach ($ids as $id) {
    addArticleComment('Server', ArticleType::GameModification, $id, $message);
}
```

But a better overall fix would probably be to prevent adding an entry to it's own related games by modifying `$arrayFromParameter` or elsewhere.

PR #1661, opened 2023-07-10
fix: prevent games and hubs from being linked to themselves

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1090.

Currently, it is possible for games and hubs to be linked to themselves by entering their own IDs into the devbox form and pressing Submit.

I opted to leave `modifyGameAlternatives()` alone for the solution implementation and instead focus solely on filtering these out in the _create.php_ endpoint. This maintains a separation of concerns: data validation/sanitization remains in the controller layer.

I am not sure what, if any, games or hubs are currently exhibiting this issue in prod, however after this solution they are still be able to be manually unlinked/fixed.


Soulsderfuchs:
The only hub that used to be linked to itself was the Mature Hub for the purpose of showing the warning page upon viewing the hub but I belive this was fixed to just show the warning regarless, but I'm not sure.

wescopeland:
Thanks for the heads up. It appears the [Theme - Mature] hub is linked to itself on prod, however this has no effect and the mature content warning is not displayed. Probably a bug.

public/request/game-relation/create.php
```diff
@@ -10,9 +10,16 @@
...
 
-modifyGameAlternatives($user, (int) $input['game'], toAdd: $input['relations']);
+$gameId = (int) $input['game'];
+
+// Filter out instances where a game might be linked to itself.
+$relationsArray = explode(",", $input['relations']);
+$filteredArray = array_diff($relationsArray, [$input['game']]);
+$filteredRelationsCsv = implode(",", $filteredArray);
+
+modifyGameAlternatives($user, $gameId, toAdd: $filteredRelationsCsv);
```
Jamiras, line 23: I think it would make more sense to pass `$filteredArray` to `modifyGameAlternatives` to avoid deconstructing the CSV again, but you'd lose the input validation (and this field is user-entered), so it's probably fine like this.

Jamiras (approved):

PR #1663, opened 2023-07-10
fix(game): show mature content warning on mature hub page

wescopeland:
Resolves an issue raised at https://github.com/RetroAchievements/RAWeb/pull/1661#issuecomment-1629765567.

Currently, even though the mature content hub is linked to itself, the mature content warning is not being displayed. This is easily reproducible by visiting the hub using an incognito browser window (id 8323).

With this change, that particular hub will render the warning even when it is not linked to itself.


public/gameInfo.php
```diff
@@ -87,6 +87,8 @@
             $gate = true;
         }
     }
+} elseif (str_contains($gameTitle, '[Theme - Mature]')) {
```
Jamiras, line 90: you need to honor the `v` parameter, or a user who hasn't set their profile permission can never access the page.

public/gameInfo.php
```diff
@@ -87,6 +87,8 @@
...
+    $gate = !($userDetails && BitSet($userDetails['websitePrefs'], $matureContentPref));
```
Jamiras, line 91: Consider improving readability by distributing the not:
```
$gate = !$userDetails || !BitSet(...);
```

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1663/commits/c2190473a6afc0e5fc6ccacd1d7a6f3a524f9b9d.

public/gameInfo.php
wescopeland: Thanks, addressed in https://github.com/RetroAchievements/RAWeb/pull/1663/commits/c2190473a6afc0e5fc6ccacd1d7a6f3a524f9b9d.

Jamiras (approved):

PR #1664, opened 2023-07-10
Fix outdated seeders and factories

luchaos:
Adjust existing seeders and factories to work with the merged V3 schemas.

There are many more areas in the V2 code where column names weren't changed back to the current naming/casing. Those will be touched as more affected code of is rewritten and introduced.

> **Note**
> This is meant for seeding empty databases. It can also be run against existing data but will not seed data in some cases if data already exists in a table.
> If you want to keep data in your existing database do NOT run `composer mfs` or `php artisan migrate:fresh --seed` which will reset and seed each time. That's the intended way for local development however.

Run `php artisan db:seed` to run the seeders. More seeders to come in the future, maybe even sync capabilities via Web API using a personal API key.


database/seeders/GamesTableSeeder.php
```diff
@@ -43,7 +44,10 @@ public function run(): void
          * add achievements to games
          */
         Game::all()->each(function (Game $game) {
-            $game->achievements()->saveMany(Achievement::factory()->count(random_int(0, 10))->create());
+            $game->achievements()->saveMany(Achievement::factory()->count(random_int(0, 10))->create([
+                'GameID' => $game->ID,
+                'Flags' => AchievementType::OfficialCore,
+            ]));
```
Jamiras, line 50: These achievements should be assigned to valid users (preferably only one user for most games, and 2-4 on a handful of others). They're currently generating unique user names, so all the achievements are owned by unique non-existent users:

image

Jamiras (approved):
Verified the seeders run and populate the database, but the statistics section doesn't seem to be accurate:
image

```
ariaDB [retroachievements-web-blank]> select count(*) from GameData;
+----------+
| count(*) |
+----------+
|      246 |
+----------+
1 row in set (0.001 sec)

MariaDB [retroachievements-web-blank]> select count(*) from Achievements;
+----------+
| count(*) |
+----------+
|     1160 |
+----------+
1 row in set (0.002 sec)

MariaDB [retroachievements-web-blank]> select count(*) from UserAccounts;
+----------+
| count(*) |
+----------+
|       29 |
+----------+
1 row in set (0.001 sec)

MariaDB [retroachievements-web-blank]> select count(*) from Awarded;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.001 sec)
```

I think this (and the achievement author thing) can be addressed separately.

database/seeders/GamesTableSeeder.php
luchaos: Sounds good, thanks for looking into it this closely! 
The users seeds include one user per upcoming permission matrix role - we could reuse those for the various author relations, too. I'll have a look at that another time - the seeders need more love in general.

PR #1671, opened 2023-07-15
Resolves #1670 By adding ConsoleId to the ClaimKey check

### LINKED ISSUE #1670: New Sets/Revisions isn't using console names for filtering
Screenshot 2023-07-15 at 1 22 22 PM

lordpsycho released two sets with the same game name, but only one is appearing on the list. One game is for PS2, another is for PSP.

fatfingers23:
This resolves #1670 by adding the ConsoleId on the edge case if a game shares a name with one on another console then it does not get flagged. I know the others were using the name of user and game, but figured ConsoleId would be a bit more unquie, smaller in size and maybe stop any other possible edge cases.

Also added a Seeder to populate some claims. Set `Status => ClaimStatus::Complete` so it has some data to display on the front page.

### DIFF
--- a/database/seeders/AchievementSetClaimSeeder.php
+++ b/database/seeders/AchievementSetClaimSeeder.php
@@ -13,8 +13,8 @@ class AchievementSetClaimSeeder extends Seeder
 {
     public function run(): void
     {
-        //
-        if(AchievementSetClaim::count() > 0) {
+
+        if (AchievementSetClaim::count() > 0) {
             return;
         }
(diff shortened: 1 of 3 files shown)


wescopeland (approved):
I've tested and approve the changes in _finished-claims.blade.php_. I have no opinion on the seeder code and would rather luchaos or Jamiras weigh in.

luchaos (approved):
lgtm, thanks!

PR #1673, opened 2023-07-15
add tests for awardachievement API

Jamiras:


app/Helpers/database/player-achievement.php
```diff
@@ -40,74 +40,64 @@ function playerHasUnlock(?string $user, int $achievementId): array
     return $retVal;
 }
 
-function unlockAchievement(string $user, int $achIDToAward, bool $isHardcore): array
+function unlockAchievement(string $username, int $achIDToAward, bool $isHardcore): array
```
luchaos, line 43: `int $achievementId` for simpler parameter naming?

luchaos (approved):
Awesome 👏

app/Helpers/database/player-achievement.php
Jamiras: I plan to follow this up with a migration to a command. I'll make the change there.

Jamiras: Looks like there's already a partially implemented `UnlockPlayerAchievementAction` using the new schemas. I'll make the change here.

PR #1674, opened 2023-07-16
refactor(user-card): migrate logic to class and add test coverage

wescopeland:
This PR migrates the user card from an anonymous Blade component to a standard component.

**Why?**
_user.blade.php_ is fairly logic-dense and cumbersome. As an anonymous Blade component, it also isn't able to directly use the `CacheKey` helper service. This refactor better separates the logic and view concerns.

**Implementation Notes:**
* `userCardDataCacheKey` no longer is passed down as a prop. It is instead computed within the `UserCard` class.
* The user is now stored as an array (via `toArray()`) in the cache. This remediates a breakage on the cards test page.
* Several visibility tests have been added to validate the conditional renderings of the card UI.
* _card.css_ has been deleted as it is all dead code.


app/Site/Models/User.php
```diff
@@ -151,6 +151,10 @@ class User extends Authenticatable implements CommunityMember, Developer, HasCom
         "ContribYield",
         "TrueRAPoints",
         "websitePrefs",
+        "Motto",
+        "Untracked",
+        "LastLogin",
+        "avatarUrl",
```
wescopeland, line 157: I am not sure if it is good practice to expose these. This way the only way to have these natively available when calling `toArray()`. Another alternative would be to delete these lines and instead do something like:

```php
return array_merge(
    $foundUser->toArray(),
    [
        'avatarUrl' => $foundUser->AvatarUrl,
        // ...
    ]
);
```

app/View/Components/Community/UserCard.php
```diff
@@ -0,0 +1,163 @@
...
+                return $foundUser ? $foundUser->toArray() : null;
+            }
+        );
+    }
+
+    private function buildAllCardViewValues(string $username, array $rawUserData): array
+    {
+        $cardBioData = $this->buildCardBioData($rawUserData);
+        $cardRankData = $this->buildCardRankData($username, $rawUserData['RAPoints'], $rawUserData['RASoftcorePoints'], $rawUserData['Untracked'] ? true : false);
+        $cardRoleData = $this->buildCardRoleData($username, (int) $rawUserData['Permissions']);
```
wescopeland, line 73: phpunit in CI is failing without this `(int)` cast. However, locally everything passes. I'm not entirely clear why this would be the case. Maybe there is a subtle difference in how something is configured in our GitHub Actions job?

resources/views/components/community/user-card.blade.php
wescopeland, line 1: With the exception of lines 7 and 8, this file is unchanged from the previous implementation.

app/Site/Models/User.php
Jamiras: I'm under the impression that it's fine to expose anything that's not sensitive (email address, encrypted password, api token, etc).

app/View/Components/Community/UserCard.php
```diff
@@ -0,0 +1,163 @@
...
+
+    private function getUsername(string|array $user): ?string
+    {
+        return is_string($user) ? $user : ($user['User'] ?? null);
+    }
+
+    private function getUserData(string $username): ?array
+    {
+        return Cache::store('array')->remember(
+            CacheKey::buildUserCardDataCacheKey($username . "1"),
```
Jamiras, line 59: This seems prone to error. If appending "1" is meant to differentiate the behavior from some other cache entry, what happens if there is a user whose name is "$username1" and they read/write this key?

app/View/Components/Community/UserCard.php
Jamiras: I'm not sure what might differ between local and CI, but instead of casting here, you probably want to add `'Permissions' => 'integer'` in the model's `casts` array.

app/View/Components/Community/UserCard.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        $username = $rawUserData['User'] ?? "";
+        $motto = $rawUserData['Motto'] ?? null;
+        $avatarUrl = $rawUserData['avatarUrl'] ?? null;
+        $hardcorePoints = $rawUserData['RAPoints'] ?? 0;
+        $softcorePoints = $rawUserData['RASoftcorePoints'] ?? 0;
+        $retroPoints = $rawUserData['TrueRAPoints'] ?? 0;
+        $isUntracked = $rawUserData['Untracked'] ? true : false;
+        $permissions = $rawUserData['Permissions'] ?? Permissions::Unregistered;
+        $memberSince = $rawUserData['Created'] ?? Carbon::now();
```
Jamiras, line 88: I'm assuming Created isn't visible, as all my users have been members since today.
image

app/View/Components/Community/UserCard.php
```diff
@@ -0,0 +1,163 @@
...
+            'rankType',
+            'rankLabel',
+            'rankPctLabel',
+            'rankMinPoints',
+        );
+    }
+
+    private function buildCardRoleData(string $username, int $permissions): array
+    {
+        $canShowUserRole = $permissions >= Permissions::JuniorDeveloper;
```
Jamiras, line 150: Not sure if we care, but Banned and Spam accounts are no longer flagged as such on the card.

app/View/Components/Community/UserCard.php
wescopeland: Apologies, I was debugging something and forgot to back this out. "1" shouldn't be appended anymore as of https://github.com/RetroAchievements/RAWeb/pull/1674/commits/87b59e67ccb06ac73265d9fd3adbc75ac87d2ed4.

app/View/Components/Community/UserCard.php
wescopeland: Wow, that worked. Good to know. Updated in https://github.com/RetroAchievements/RAWeb/pull/1674/commits/87b59e67ccb06ac73265d9fd3adbc75ac87d2ed4.

app/View/Components/Community/UserCard.php
wescopeland: Thanks, fixed in https://github.com/RetroAchievements/RAWeb/pull/1674/commits/87b59e67ccb06ac73265d9fd3adbc75ac87d2ed4.

app/View/Components/Community/UserCard.php
Jamiras: I know we briefly talked about it in the forum refactor, and the decision was to not highlight the ban status in the forum entry user panel. The user-card does currently show the banned status, so I'm just noting that this is a change in behavior.

I think it's probably fine to not include it here.

There's still a bit of an oddity that clicking the user name will go to a Not Found page, but I don't think there's much of a demand for going to a user profile from a thread post.

app/View/Components/Community/UserCard.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return Cache::store('array')->remember(
+            CacheKey::buildUserCardDataCacheKey($username),
+            Carbon::now()->addMonths(3),
+            function () use ($username): ?array {
+                $foundUser = UserModel::firstWhere('User', $username);
+
+                return $foundUser ? $foundUser->toArray() : null;
+            }
+        );
```
Jamiras, line 66: I'm a bit concerned that this code is caching a `UserModel->toArray`, and the code in `userAvatar` is caching the output of `getAccountDetails` under the same cache key.

Both are using in-request caches, so there's minimal chance of the conflict actually causing issues, but you can see at least one aspect of the problem on the cards test page. If you forcibly expire the cache entry before doing the lookup in `UserCard::getUserData`, the embedded card will gain an image that's not present when it leverages the data cached by `userAvatar`.

wescopeland: It is a valid concern. After studying `userAvatar()` for a moment, it appears `getAccountDetails()` is being called simply to do an existence check for the user.

This is probably overkill. I suspect something like `User::where('User', $username)->exists()` is more performant and the caching can be eliminated here entirely.

I've made this change in https://github.com/RetroAchievements/RAWeb/pull/1674/commits/da210d147e73667624ed79aea9c3cfd1d7f1f3c8.

app/Helpers/render/user.php
```diff
@@ -49,8 +37,6 @@ function () use ($username) {
         return "<span class='inline whitespace-nowrap'><span class='inline-block'>" . $iconLabel . $usernameLabel . "</span></span>";
     }
 
-    $username = $user['User'] ?? null;
```
Jamiras, line 52: Without this, `[user=X]` links are no longer case corrected. `[user=Jamiras]` goes to `user/Jamiras`, but `[user=jamiras]` goes to `user/jamiras`, which does do the case correction at that point, but the link is incorrectly cased.

wescopeland: You're right, I didn't recognize that was the intent of this line. Given that, an existence check unfortunately won't be sufficient.

In https://github.com/RetroAchievements/RAWeb/pull/1674/commits/5f637768bc60a06f00d234f902b6c804b12031f7, I've instead opted to utilization the same caching implementation that user cards are using (with a subsequent `toArray()` call within the caching function). Now, at the very least, there will not be caching conflicts.

Jamiras (approved):

Issue #299, opened 2019-05-24
Adding link to Video for Leaderboards


magusLOGS:
A option to adding/editing a link to a video/recorded-stream on the Leaderboards once a Score/Time has been submitted.

Example:
![Video Link on Leaderboards](https://i.imgur.com/ZGDSvKQ.png)

Tsearo:
I'd love to see this and can try to see how much work it would be to implement. It would likely require a new column in `leaderboardentry` for storing/updating the link and then would just need to find a good way to add an option for the user to set/update the link for their own entries. Could possibly have an expandable div that displays the embedded video below the entry when clicked.

PR #1675, opened 2023-07-17
Adding the ability for a user to link to a video from a leaderboard entry of the user

cIymax:
This prototype PR of mine has been made and sitting in my personal repository since March.

Given my hurdles in setting up a proper local test environment, I thought I'd just go ahead and make a draft PR on this so that maybe this can gain some traction.

It addresses #765.  See also #299.


wescopeland (changes requested):
There are userscripts in this PR. This PR also adds another `s_mysql_query()` call, which is deprecated. It would also be practical if new UI used Blade components instead of legacy `echo` statements.

PR #1676, opened 2023-07-18
fix(Shortcode): remediate display issue with stripped/clamped hub titles

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1130987662514323536.

Currently, hub titles are being stripped by the shortcode `stripAndClamp()` function.

**Root Cause**
`stripAndClamp()` has a bug where game titles containing square brackets are not properly displayed. These characters are interpreted as shortcodes by the function's regex patterns in the processing subsequent to game shortcode processing, resulting in their removal. This fix addresses the issue by adding escape sequences for the square brackets before injecting the game data in the replaced input text.


Jamiras (commented):
This seems to be due to the "opening tags without closing tags" regex being too greedy. It only affects hubs that start with `b`, `i`, `u`, or `s` (like "Series" - it doesn't affect "Genre").

I propose updating the regex to match whole words (require a non-alphabetic character after the tag match):
```
'~\[(b|i|u|s|img|code|url|link|spoiler|ach|game|ticket|user)[^a-z][^\]]*?\]~i' => '',
'~\[(b|i|u|s|img|code|url|link|spoiler|ach|game|ticket|user)[^a-z][^\]]*?$~i' => '...',
```

It also seems like this is something that should be added to the unit tests.

wescopeland:
I like your proposal. I think we can use `\b`, which represents a word boundary. The change appears to work. 

I agree with the testing recommendation, these expression patterns are beginning to make my head spin. I've added several more test cases.

Updates made in https://github.com/RetroAchievements/RAWeb/pull/1676/commits/b1e1f6fee000cd6abd382ac1dcb67877a10aa6c5.

tests/Feature/Community/ShortcodeClampTest.php
wescopeland, line 1: Open to recommendations on a better file+class name. Wanted to separate these test cases from the unit tests because they use the mock DB.

Jamiras: My suggestion would be `ShortcodeEmbedTest` or something similar, indicating the dependency on the database.

wescopeland: Renamed in https://github.com/RetroAchievements/RAWeb/pull/1676/commits/dcca6109dcfa12070a3519ab7d397a2f9dfe950e.

Jamiras (approved):

PR #1681, opened 2023-07-21
feat: add multiline game avatar component

wescopeland:
This PR introduces a new component: `<x-game.multiline-avatar />`, which is modeled closely after the game display in the new AOTW component.

The intent of this PR revolves more around the agreement of the component's API/props design before we expand the component's usage to places on the site where it makes sense. As a result, this PR only integrates the component into the claims tables on the home page.

**Technical Notes**
* Rather than accepting a single model object, the props are decomposed across several values and the responsibility is on the caller to map the input data correctly. This has the benefit of being agnostic to the retrieval shape of the game by the caller; as long as the data can be mapped, the component can be used.
* Console information is nullable. This may be useful in scenarios where console information is better to remain hidden (such as hubs or events?).
* By default, the badge and label link to the game page, but this can be overridden with an `href` prop.

**BEFORE**
Screenshot 2023-07-20 at 8 30 36 PM

**AFTER**
Screenshot 2023-07-20 at 8 28 46 PM


resources/views/community/components/claims/finished-claim-table-row.blade.php
```diff
@@ -12,8 +12,16 @@
 ?>
 
 <tr>
-    <td class="pr-0">{!! gameAvatar($claim, label: false, tooltip: false) !!}</td>
-    <td class="w-full">{!! gameAvatar($claim, icon: false, tooltip: false) !!}</td>
+    <td class="py-1.5">
+        <x-game.multiline-avatar
+            :gameId="$claim['ID']"
```
Jamiras, line 17: This needs to be the game ID, not the claim ID.

Master (castlevania=1462):
image

This branch:
image

resources/views/community/components/claims/new-claim-table-row.blade.php
```diff
@@ -6,8 +6,16 @@
 ?>
 
 <tr>
-    <td class="pr-0">{!! gameAvatar($claim, label: false, tooltip: false) !!}</td>
-    <td class="w-full">{!! gameAvatar($claim, icon: false, tooltip: false) !!}</td>
+    <td class="py-1.5">
+        <x-game.multiline-avatar
+            :gameId="$claim['ID']"
```
Jamiras, line 11: Same here

Jamiras (changes requested):

resources/views/community/components/claims/finished-claim-table-row.blade.php
wescopeland: Thanks, fixed in https://github.com/RetroAchievements/RAWeb/pull/1681/commits/0e2f4c7f3c9d81138983bb1f36943dd5b13c62df.

resources/views/community/components/claims/new-claim-table-row.blade.php
wescopeland: https://github.com/RetroAchievements/RAWeb/pull/1681/commits/0e2f4c7f3c9d81138983bb1f36943dd5b13c62df

Jamiras (approved):

PR #1683, opened 2023-07-21
feat(forum): add 'Preview' support

wescopeland:
This PR adds preview functionality to the site forum for creating new posts, editing existing posts, and posting new replies.

**Technical Implementation**
* A new endpoint has been added to construct forum post previews. It accepts text content, sends it to the shortcode parser, then returns the HTML content for what the post would look like (rendered within the forum post Blade component) in its response.
* Upon clicking the Preview button, a new JavaScript function, `loadPostPreview()`, is called which makes the new API request. This function also enables visibility of a loading indicator and guards against accidental duplicate preview submissions.
* When the API call resolves, the JavaScript function replaces the inner HTML of an empty "preview" div with the returned HTML content from the endpoint's response.

Currently, this is implemented only for the forum. It is not implemented for user private messages. I was afraid of blowing up the scope of this PR any further; private messages can come next.

---

https://github.com/RetroAchievements/RAWeb/assets/3984985/e91b7809-10bb-4dbe-a183-17e7758870ab


Jamiras:
There's something about this post (http://localhost/editpost.php?comment=145736) that's causing a problem in the preview.

Actual:
image

Preview:
image

The spoiler contents aren't inside the spoiler.

wescopeland:
@Jamiras Want to make sure we're seeing the same thing, when I preview editing this post I am seeing content inside the spoiler. If you do `npm run build` and then don't spin up your local Vite dev server, are you still able to reproduce this issue?


https://github.com/RetroAchievements/RAWeb/assets/3984985/746f0de4-53c9-403d-ac6a-61381b1e34d7

luchaos:
> @Jamiras Want to make sure we're seeing the same thing, when I preview editing this post I am seeing content inside the spoiler. If you do `npm run build` and then don't spin up your local Vite dev server, are you still able to reproduce this issue?

Seems to be working fine with a production build of the frontend assets:
https://preview.retroachievements.org/editpost.php?comment=145736

public/request/forum-topic-comment/preview.php
```diff
@@ -0,0 +1,31 @@
...
+$parsedPostContent = Shortcode::render($input['body']);
+
+$postPreviewHTML = '';
+ob_start();
+echo "<div class='my-2'>";
+echo Blade::render('<x-forum.post :parsedPostContent="$parsedPostContent" isPreview="true" />', [
+    'parsedPostContent' => $parsedPostContent,
+]);
+echo "</div>";
+$postPreviewHTML = ob_get_clean();
```
luchaos, line 26: Maybe the output buffer is the reason for the response being returned short.
Is it needed in this case? Can you please try concatenating it manually:
```php
$postPreviewHTML = "<div class='my-2'>";
$postPreviewHTML .= Blade::render('<x-forum.post :parsedPostContent="$parsedPostContent" isPreview="true" />', [
    'parsedPostContent' => $parsedPostContent,
]);
$postPreviewHTML .= "</div>";
```

public/request/forum-topic-comment/preview.php
```diff
@@ -0,0 +1,31 @@
...
+    return back()->withErrors(__('legacy.error.permissions'));
+}
+
+$input = Validator::validate(Arr::wrap(request()->post()), [
+    'body' => 'required|string|max:60000',
+]);
+
+$parsedPostContent = Shortcode::render($input['body']);
+
+$postPreviewHTML = '';
```
luchaos, line 19: Minor nit: While it's not defined in the PSR coding standard guidelines, acronyms are usually treated like this: `$postPreviewHtml`.
We have a few areas in this codebase where this should be cleaned up.

Jamiras:
> @Jamiras Want to make sure we're seeing the same thing, when I preview editing this post I am seeing content inside the spoiler. If you do `npm run build` and then don't spin up your local Vite dev server, are you still able to reproduce this issue?

Yes, I'm still seeing it after running `.hooks/post-merge`, which updates all the dependencies then runs `npm run build`.

luchaos has also pointed out that it is occurring on the preview site:
> seems there's still an issue with how the preview renders it 🤔
> https://preview.retroachievements.org/viewtopic.php?t=19615&c=145736#145736
> https://preview.retroachievements.org/editpost.php?comment=145736
> 
> The post seems to be cut off? All the shortcode/post body is sent from what i can tell but the response is not complete.
> Not quite sure yet why the rendered post stops short in the middle of it, resulting in spoiler tags not closing etc

I'm seeing the same thing in both places. The post does appear to be truncated:
image

Jamiras:
It seems to be getting split up into multiple variables when sent to the server:
image

Corresponding input:
```
[url=https://retroachievements.org/historyexamine.php?d=1678492800&u=Searo]Link[/url]
[ach=174916]
```

Seems to be the ampersand. Can reproduce with this simple case:
image

wescopeland:
I've made the suggested changes; I'm able to repro the issue now with the input "M&M's are great!". I'll work now to resolve.

wescopeland:
```ts
        body: `body=${encodeURIComponent(postContent)}`,
```

This may be the solution, with this change I'm able to see a preview of "M&M's are great!". The ampersand was a good idea, I bet it was being treated as a delimiter before.

Jamiras (approved):

PR #1685, opened 2023-07-22
chore(cs): replace cs-fixer with pint

luchaos:
Use Laravel Pint which is based on CS-Fixer. Rules stay the same.

Replaces required `php-cs-fixer` GitHub action with `pint` (renamed).

Ran it on existing codebase and included some minor suggested changes.

Run as before: `composer lint` and `composer fix`.

`.php-cs-fixer.cache` has been removed from `.gitignore` as it will not be generated anymore, can be deleted.


wescopeland (approved):
Verified things appear to be working great here. `pint` seems noticeably faster.

Future changes we may want to consider (though not in this PR):
- Setting `braces` to `true`.
- Setting `declare_strict_types` to `true`.

luchaos:
> Verified things appear to be working great here. `pint` seems noticeably faster.
> 
> Future changes we may want to consider (though not in this PR):
> 
> * Setting `braces` to `true`.
> * Setting `declare_strict_types` to `true`.

Absolutely 👍 both of those rules were mentioned in the previous php cs fixer configuration but could not be applied yet due to some required refactoring in the view layer.
Thanks for mentioning it here for reference!

PR #1688, opened 2023-07-23
make point updates atomic

Jamiras:
fixes an issue where simultaneous unlocks would not update points correctly


app/Helpers/database/achievement-creator.php
```diff
@@ -159,10 +159,9 @@ function attributeDevelopmentAuthor(string $author, int $count, int $points): vo
     $oldContribCount = $user->ContribCount;
     $oldContribYield = $user->ContribYield;
 
-    $user->ContribCount += $count;
-    $user->ContribYield += $points;
-    $user->timestamps = false;
-    $user->save();
+    // use raw statement to perform atomic update
+    legacyDbStatement("UPDATE UserAccounts SET ContribCount = ContribCount + $count," .
```
luchaos, line 163: Now that we have reduced the DB connection to just one and those functions using the exact same signature we could deprecate/move away from the legacy* helpers and use the features directly.
```php
// used to have the connection
// DB::connection('mysql_legacy')->statement($query, $bindings);
// now it's just the same
DB::statement($query, $bindings);
```
I'll prepare a separate PR for that.

luchaos (approved):

luchaos:
Repost from Discord:

That's where the queued jobs will help. the unlock will trigger a queued job which takes care of the aggregation, multiple unlocks will push multiple jobs to the queue. those will then be worked on sequentially.
Whichever jobs runs last will certainly have all the data it needs to calculate the correct sum which is more reliable than increments.
Eventually I'd remove the calculations from those areas and work with asynchronous jobs instead

Jamiras:
> Eventually I'd remove the calculations from those areas and work with asynchronous jobs instead

That may be difficult in this particular scenario as the updated score is returned by the API. It's probably acceptable to return the slightly incorrect value on a multi-unlock as it will be correct on the site and updated on the client on a future unlock or login.

PR #1689, opened 2023-07-23
fix(main.blade.php): use alternative solution for minimizing layout shift

wescopeland:
This PR removes max-width in favor of swapping the position of the `<article>` and `<aside>` tags in the template. This has the benefit of remediating layout shift on the home page while also getting fluid layouts to work (`/demo`).

**Why this works**
The layout shift on the home page is caused by the Active Players widget's `<script>` tag, which the browser loads synchronously. During loading, HTML parsing is blocked and the browser renders the content available in that frame, which just happens to be missing the right-hand column. This is what is causing the layout shift on the home page.

This solution remediates the issue by forcing the browser to parse the right-hand column first.


luchaos (approved):
Fantastic 👏

PR #1690, opened 2023-07-23
feat(game): migrate alt games rows to blade

wescopeland:
This PR modifies the `RenderGameAlts()` function to return rows containing a new blade component: `<x-game.similar-game-table-row />`, which itself utilizes the new `<x-game.multiline-avatar />`. component.

**Notes**

* **I am not sure adding the "RP" label is desirable.** Without it, the number feels like it's floating in the abyss. The label doesn't feel like the perfect solution, but it feels better than having nothing at all. I tried other label variants such as "RetroPoints", but had a hard time getting something to fit in the constrained width.

* `<x-game.multiline-avatar />` has been modified to have the ability to not render a console icon (by not passing a `consoleId` prop). This came up in #1500. I believe for these table rows not having the icon is desirable to reduce clutter.

* Another new component, `<x-retropoints-container />`, has been added. This component accepts a slot, applies the `TrueRatio` classname, and adds a tooltip. Ideally, this could be reused in many places, but I wanted to try fishing it out here first. The aim here is to ensure consistency in how RetroPoints displays are rendered, as well as provide a helpful tip for answering one of the most common FAQs on the site ("what are white points?").

* These changes are also visible on hubs.

---

**Before & After**
Screenshot 2023-07-23 at 4 07 56 PM

Screenshot 2023-07-23 at 4 05 05 PM


resources/views/components/retropoints-container.blade.php
luchaos, line 1: -> `views/platform/components`?

resources/views/components/game/similar-game-table-row.blade.php
luchaos, line 1: -> `views/platform/components`?

the `views/components` directory is meant for "base" components that are shared among domains (mostly applied atm). if possible please add/move new components to the corresponding domain's directory (community/platform).

app/Helpers/render/game.php
```diff
@@ -270,29 +270,29 @@ function RenderGameAlts(array $gameAlts, ?string $headerText = null): void
...
+            <x-game.similar-game-table-row
+                :gameId="$gameId"
+                :gameTitle="$gameTitle"
+                :gameImageIcon="$gameImageIcon"
+                :consoleName="$consoleName"
+                :totalPoints="$totalPoints"
+                :totalRetroPoints="$totalRetroPoints"
+                :isFullyFeaturedGame="$isFullyFeaturedGame"
+            />
+        ', $rowData);
```
luchaos, line 293: Allows to pass the `$rowData` array directly as the local variable isn't used anywhere else

luchaos (changes requested):

resources/views/components/game/multiline-avatar.blade.php
luchaos, line 1: -> `views/platform/components/game`

luchaos:
> The aim here is to ensure consistency in how RetroPoints displays are rendered 
I'd prefer if the the parentheses were removed

luchaos:
> The aim here is to ensure consistency in how RetroPoints displays are rendered

I'd prefer if the the parentheses were removed and all three variant of points represented visually with icons.

For example:
Hardcore points https://blade-ui-kit.com/blade-icons/fas-circle
Softcore points https://blade-ui-kit.com/blade-icons/far-circle
Weighted points (TruePoints) https://blade-ui-kit.com/blade-icons/phosphor-circles-three-fill

This would allow the augment labels with icons and tooltips to save space and have a color independent representation.

app/Helpers/render/game.php
wescopeland: Sorry, I'm not sure I'm following.

wescopeland:
>I'd prefer if the the parentheses were removed and all three variant of points represented visually with icons. RP is already used as "Rich Presence" and might be misleading?

@luchaos I agree. Your suggestion aligns with my initial thoughts before opening this PR. However, I was concerned that incorporating such a change could dramatically broaden the scope of this particular component migration.

I have an alternative proposal. Let's first remove the ambiguous "RP" label so that we maintain feature parity with the existing legacy template. Immediately after this, I can open another PR dedicated to transitioning towards iconography. Ideally, this would allow us to visually represent the count of achievements, the points sum, and the RetroPoints sum. Without icons, we lack the necessary space to achieve this effectively, so I am in strong agreement that it is desirable.

I'm open to tackling it in this PR too, but addressing this in a subsequent PR would allow this migration to get through and allow for a more nuanced discussion on the topic of iconography, which is likely to influence the platform in ways that extend beyond this single component.

app/Helpers/render/game.php
luchaos: Could be passed directly like this - not very important though :)
```php
echo Blade::render('
    <x-game.similar-game-table-row
        :gameId="$gameId"
        :gameTitle="$gameTitle"
        :gameImageIcon="$gameImageIcon"
        :consoleName="$consoleName"
        :totalPoints="$totalPoints"
        :totalRetroPoints="$totalRetroPoints"
        :isFullyFeaturedGame="$isFullyFeaturedGame"
    />
', [
    'gameId' => $nextGame['gameIDAlt'],
    'gameTitle' => $nextGame['Title'],
    'gameImageIcon' => $nextGame['ImageIcon'],
    'consoleName' => $consoleName,
    'totalPoints' => $points,
    'totalRetroPoints' => $totalTP,
    'isFullyFeaturedGame' => $isFullyFeaturedGame,
]);
```

wescopeland: No worries at all, addressed in https://github.com/RetroAchievements/RAWeb/pull/1690/commits/3909b0522d92c54474d3cf290cdf48e53e3717ad.

resources/views/components/game/multiline-avatar.blade.php
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1690/commits/148fe7f05b5eb8f0555446cc8bd0644dc1f1f571.

resources/views/components/game/similar-game-table-row.blade.php
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1690/commits/148fe7f05b5eb8f0555446cc8bd0644dc1f1f571.

resources/views/components/retropoints-container.blade.php
wescopeland: Renamed to `<x-points-weighted-container />` in https://github.com/RetroAchievements/RAWeb/pull/1690/commits/063c71cdf43abcd268eab9fbcd1440a9e790c1e4.

I kept the container suffix as currently this component is not prescriptive on how the RetroPoints themselves are rendered, though it is very likely that will change in the subsequent iconography PR.

PR #1692, opened 2023-07-24
return HTTP 401 and code field on dorequest authentication failures

Jamiras:
also adds `Code` and `Status` to error responses and Status is reflected in HTTP Response Code per discussion https://discord.com/channels/476211979464343552/757767535293890682/1132370669938671758

### DIFF
--- a/app/Helpers/database/user-auth.php
+++ b/app/Helpers/database/user-auth.php
@@ -226,9 +226,6 @@ function authenticateFromAppToken(
     if (empty($userOut)) {
         return false;
     }
-    if (!isValidUsername($userOut)) {
-        return false;
-    }
     if (empty($token)) {
         return false;
     }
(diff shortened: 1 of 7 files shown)


app/Helpers/database/user-auth.php
```diff
@@ -162,17 +134,20 @@ function changePassword(string $user, string $pass): bool
 
 function hashPassword(string $pass): string
 {
-    return password_hash($pass, PASSWORD_ARGON2ID, [
+    $algorithm = defined('PASSWORD_ARGON2ID') ? PASSWORD_ARGON2ID : PASSWORD_DEFAULT;
```
luchaos, line 137: very nice - this solves the potentially missing argon support in php in some local setups.

I'll probably refactor this to use Laravel's capabilities at some point, should be fairly easy to make this compatible.

luchaos (approved):

luchaos (approved):

PR #1696, opened 2023-07-25
fix(tooltip): capture mouse coordinates on mouseover events

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1133391926402490388.

**Root Cause**
When landing on a tooltipped element using only the mouse wheel and no mouse movement, an `onmouseover` event is fired to pop the tooltip. However, the event is not capturing the mouse coordinates, resulting in the tooltip to be incorrectly positioned on the top left corner of the screen.

To resolve the issue, the mouse coordinates are now captured and utilized correctly.

### DIFF
--- a/app/Helpers/database/user-activity.php
+++ b/app/Helpers/database/user-activity.php
@@ -77,6 +77,7 @@ function postActivity(string|User $userIn, int $type, ?int $data = null, ?int $d
                 return false;
             }
             $activity->data = (string) $data;
+            $activity->data2 = (string) $data2;
             break;
 
         case ActivityType::Login:
--- a/app/Helpers/render/game.php
+++ b/app/Helpers/render/game.php
@@ -1,6 +1,7 @@
 <?php
 
 use App\Site\Enums\Permissions;
+use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Facades\Cache;
 
 function gameAvatar(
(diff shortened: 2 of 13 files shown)


Jamiras:
These seems to have fixed the issue on the Game Awards list. Instead of the popup appearing in the upper left corner as reported, mine just sometimes didn't show up at all, and with this change, it does always show up.

I'd like to report a similar issue on the active claims widget and recent forum posts. If you hover over a user to get a tooltip, then mousewheel so the cursor is over a different user, the tooltip appears in the same place as the first user. This seems to be related to fetching the user card info. It only happens the first time after refreshing the page.

Here, I hovered over GameZone, then mousewheeled to televandalist. You can see televandalist highlighted, but the tooltip appears next to GameZone.
image

wescopeland:
@Jamiras To confirm, is this also happening locally with this branch (want to know if resolved by this conditional https://github.com/RetroAchievements/RAWeb/pull/1696/files#diff-240d4470ee3ede50a8ab91e0f5db2692756d9e8c9311ae74060c94cca2fa695bR38)?

wescopeland:
Okay, I've been banging around on that page and I've just pushed a commit to fix the reported issue & another race condition I found with these dynamic tooltips where they sometimes would freeze onto the page if the user moused away too quickly.

Let me know if things look any better on your end now.

Jamiras (approved):
Confirmed functionality on user awards (original issue) and user tooltips on claims and recent posts (secondary issue).

PR #1699, opened 2023-07-27
ignore deleted forum topics when reporting unverified forum posts

Jamiras:
When a forum topic is deleted, the topic is marked as deleted (soft delete), but the posts remain. This prevents the "deleted" posts from appearing in the list of unverified forum posts and search results.

https://discord.com/channels/476211979464343552/1002689485005406249/1134137157636870164

Side note: should the forum topic comments for the deleted topic also be marked as deleted?


luchaos (approved):
nice catch!

I don't think marking related comments deleted is necessary.
if we fetch comments for public display we should check for the topic's status just as you did here imo.

PR #1700, opened 2023-07-29
feat(card)!: modernize game and hub tooltips

wescopeland:
This PR reimplements game and hub tooltips throughout the entire site. Presently, these tooltips feature largely redundant data (such as simply game and console name). Now, the tooltips show dynamic data related the game's achievement set as well as data pertinent to a currently authenticated or otherwise context-adjacent user.

### Use Case 1: No Active Core Set
If a game has no active core set, this is valuable contextual information for a user. This use case will commonly occur on game lists and game tables, especially lists of similar games on game pages. They are presented with the game title, console name, and copy stating there are no achievements for the game.

Screenshot 2023-07-28 at 9 16 45 PM

### Use Case 2: No Active Core Set + Has Active Claim
If a game has no active core set but has a claim, this is valuable contextual information for a user. This use case will commonly occur on the claims list (including the home page). Users will be presented with similar information as Use Case 1, however now they will see basic metadata about the claimants.

Collaboration claims also have a dedicated experience, which is driven by using a new "avatar stack" common component.

Screenshot 2023-07-28 at 9 19 47 PM

### Use Case 3: Active Core Set
This is by far the most common use case that everyday users will see. Once a game has an active core set, we can start showing basic metadata about the core set such as achievements count, points sum, RetroPoints sum, retro ratio, and last updated date.

Screenshot 2023-07-28 at 9 23 15 PM

### Use Case 4: Active Core Set + Site Award
This is likely the second most common use case that everyday users will see. Once a user has earned a completion or mastery badge for a game, it is noted on the card. This use case is also visible when a user visits another player's profile and views their badges: the username that the card receives to check for awards status is dynamic and context-driven.

Screenshot 2023-07-28 at 9 27 28 PM

### Use Case 5: Active Core Set + Revision Claim
If a game has an active core set but also has a revision claim, this is valuable contextual information for a user. Users will be presented with similar information as Use Case 2, however now they will see basic metadata about the claimants.

Collaboration claims also have a dedicated experience nearly identical to Use Case 2. Award displays from Use Case 4 are still respected. The revision metadata is "bolted on".

Screenshot 2023-07-28 at 9 30 31 PM

### Use Case 6: Retired Sets
If a set has been ~Z~'d out, it will be shown as being retired in the tooltip. This will most commonly occur on longstanding users' profile pages, where site awards are present for games that now have deprecated sets.

Screenshot 2023-07-28 at 9 38 24 PM

### Use Case 7: Hubs
Hubs now have a slightly improved treatment, showing the count of links (alt game IDs) and when the hub was last updated.

Screenshot 2023-07-28 at 9 39 15 PM

## Technical Implementation

* Game tooltip cards no longer render statically via `useCard()`. Each avatar component now makes a call to the _card.php_ endpoint, similar to user cards. The game ID is used to dynamically fetch the card template. For documents containing many game avatars, this results in a non-trivial load-time performance improvement.

**NES Console Game List (862 games):**
Before - 104kB
After - 84kB (21% reduction)

**Large User Profile (701 badges):**
Before - 162kB
After: 106kB (42% reduction)

* The `context` input for game cards is now dedicated to the username. Previously, it was statically-rendered mastery metadata. This is used for fetching metadata about _other_ users' site awards, such as on user profiles.

* The cards test page has been updated with lots of examples from the use cases listed above.

* To further improve performance, a polyfill has been added for `requestIdleCallback()` that should mitigate UI thread locking on every browser except Safari.

* A bug has been fixed in `updateTooltipPosition()` that could cause some game cards on user profiles to render in an infinite loop (note: this does not happen in prod today, it is surfaced by the new game cards themselves).

* The "Last Updated" metadata is still contentious and will differ from what is shown on game lists. @Jamiras: I implemented the suggestion from https://discord.com/channels/476211979464343552/1026595325038833725/1134272667780919427, however this caused some bizarre readings on older sets that have never been subject to the claim system (such as game ID 1), so I rolled back to the implementation that is checking for achievement last updates. I know this solution is also fraught with trade-offs.

* I have checked everywhere I can find on the site that I know that renders game avatars to ensure they're working properly. I've also attached the tooltips to the AOTW component.

* I've written a fairly thorough set of integration tests covering all of the use cases listed above.


wescopeland:
Tests are passing locally but failing in CI. I may need a hand in figuring this out.

The four failing tests all use claims via:

```php
insertClaim(
    $user->User,
    $game->ID,
    ClaimType::Primary,
    ClaimSetType::NewSet,
    ClaimSpecial::None,
    Permissions::Developer,
);
```

However, we have other tests that do this too and they pass in CI.

Alexdatadestroyer:
I have some feedback regarding this feature:

**a)** How Events pages (Console = 101) will be treated on Card Tooltip when an Event is highlighted? It's not an ordinary set and neither a Hub case. I don't think that showing it like an ordinary set (achievements, points, retro ratio, last updated) would be good: Maybe, only showing if the Event was mastered or not and when was last updated then?  

**b)** Is it possible to replace the blue dot and yellow dot for a medal and a crown on their respective cases? Softcore completion is already represented by a Medal while Hardcore completion is represented as a Crown on User Profile pages. Their icon is tiny enough to fit on Card Tooltip, so why not?

(PS: I know that Blue indicates Softcore while Yellow indicated Hardcore, indeed. However, on UP is represented differently for some reason: Hardcore progress is Blue while Softcore progress is Grey.)

**c)** For Use Case 6 (Retired Sets), if the idea is using dark-white color style (similar to when an achievement is still locked) to let the user know that the Set is now deprecated, then, why not suppressing the ~Z~ tag after Set Name knowing that an Available Set will never be shown like this?

A bolded effect on "This achievement set has been retired" message will be good to give it an extra attention to the user.

Looking forward! 🚀

Alexdatadestroyer:
[Easter Egg Proposal]

Now, I was thinking about something fun which would be great if this comes to reality using this feature...

Well, you know... Some games might have easter eggs. We are on a (retro) gaming website. So why not create an easter egg thing on RAWeb? This can be a fun part of helping this site overall!

As far as I know, the only easter egg that RAWeb has it's the random reactions of RA Mascot on such error pages. Some specific pages have static reaction though (Like HTTP 403 Forbidden case). The sentence "Git gud" can be seem on such situations.

I am saying that because I figured a chance of another easter egg creation:

- There is a [glitched set page](https://retroachievements.org/game/11782) (Some Errors Are More Error Than Others) which was created due to a Database Error in past without ConsoleID at the time. Now, it's filed on Event console category. Knowing that this will remain forever as invalid set page, then, it's comes the idea of addition of an easter egg feature on it: It will show to any user that finds the page using Site Search feature:

Instead of Card Tooltip showing Event details like it was expected to, it could show something like:

> Some Errors Are More Error Than Others
>  Events
> 
>
> **Whoa dude! You just found a bugged set page created by a database error years ago! Congratulations dude, you are a legend!**

Some time ago, a similar sentence like posted above used to be a meme spread by some users on RA community to ironically congratulate a user that have mastered such a meme/easy set or just to make fun of a friend.

wescopeland:
Hello @Alexdatadestroyer,

Thanks for leaving your thoughts. I will respond to each line item individually below.

>How Events pages (ConsoleID = 101) will be treated on Card Tooltip when an Event is highlighted? It's not an ordinary set and neither a Hub case. I don't think that showing it like an ordinary set (achievements, points, retro ratio, last updated) would be good: Maybe, only showing if the Event was mastered or not and when was last updated then?

This is how event awards currently appear:

Screenshot 2023-07-29 at 7 34 54 AM

Unless a web maintainer feels otherwise, I don't think we should change any of this. I don't want the cards to be prescriptive to event organizers on achievement counts or whether an event can be mastered or completed.

>Is it possible to replace the blue dot and yellow dot for a medal and a crown on their respective cases?

I am earnestly against this recommendation for two reasons. One, emoji are difficult to get right when it comes to accessibility. Two, emoji are not vendor-standardized and render differently on every browser and OS.

This is probably the appropriate time to say I'm in favor of removing all emoji from the app and replacing them with standardized icons that better represent our brand.

>For Use Case 6 (Retired Sets), if the idea is using dark-white color style (similar to when an achievement is still locked) to let the user know that the Set is now deprecated, then, why not suppressing the Z tag after Set Name knowing that an Available Set will never be shown like this?

The game badge images have not been altered in these screenshots, and the component applies no filters to them.

I agree with you that the Z tag is confusing and unhelpful. However, changing it or suppressing it requires adding conditional logic to the `renderGameTitle()` function, which is called on nearly every page of the site. This PR is already large; I don't want to increase its scope or size.

I would like to revisit this in a future PR, as the Z tag is something that has room for improvement.

>A bolded effect on "This achievement set has been retired" message would be nice to give it an extra attention to the user.

Under the current design, labels are bolded and descriptions are using the base font weight. Subjective: the Z tag is uncommon enough that it probably isn't worth adding an exception here.

>[Easter Egg Proposal]

A few of the site's event awards are fairly prominent on developer profiles, and some of them (such as DevQuest awards) take months, if not years, of hard work and steady contributions to earn. My personal opinion is adjusting their design in this way would be contentious at best. I would only be willing to make this change if the community was asking for it.

Jamiras:
> > How Events pages (ConsoleID = 101) will be treated on Card Tooltip when an Event is highlighted? It's not an ordinary set and neither a Hub case. I don't think that showing it like an ordinary set (achievements, points, retro ratio, last updated) would be good: Maybe, only showing if the Event was mastered or not and when was last updated then?
> 
> This is how event awards currently appear:
> 
> Screenshot 2023-07-29 at 7 34 54 AM
> 
> Unless a web maintainer feels otherwise, I don't think we should change any of this. I don't want the cards to be prescriptive to event organizers on achievement counts or whether an event can be mastered or completed.

I don't feel that the achievement count matters, most events will have exactly 6 achievements as that's how many are required to award a badge. You could also change the wording from "Completed"/"Mastered" to "Awarded" (for both hardcore and softcore).

Do you have any plans to spruce up the psuedo-awards?
image
image

wescopeland:
>I don't feel that the achievement count matters, most events will have exactly 6 achievements as that's how many are required to award a badge. You could also change the wording from "Completed"/"Mastered" to "Awarded" (for both hardcore and softcore).

Fair enough, I agree that it does seem reasonable to align the language with the non-event/non-game site awards. I've made the changes in https://github.com/RetroAchievements/RAWeb/pull/1700/commits/504938aa31b6b07b5658df48ea9a682d8c196382 which include adding an event card to the cards test page and adding another test case to handle this code path.

Screenshot 2023-07-29 at 10 56 43 AM

>Do you have any plans to spruce up the psuedo-awards? (also note the 'Awarded' phrasing)

Yes, most certainly! I would like to touch up every remaining tooltip kind on the site, which to my knowledge leaves achievements, tickets, and these straggler site award native tooltips as the ones outstanding. My intention is to tackle these tooltip kinds in subsequent PRs.

app/View/Components/Platform/Cards/Game.php
```diff
@@ -0,0 +1,304 @@
...
+                    'ConsoleID' => $foundGameConsoleId,
+                    'ConsoleName' => $foundGame->system->Name,
+                    'Achievements' => $foundGameAchievements,
+                    'Claims' => $foundClaims,
+                    'AltGames' => $foundAltGames,
+                ]
+            );
+        })();
+
+        Cache::store('array')->put($cacheKey, $loadGameCardData, Carbon::now()->addDays(7));
```
luchaos, line 106: Adding a TTL using the ephemeral `array` cache won't have an effect.

app/View/Components/Platform/Cards/Game.php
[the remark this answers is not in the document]
wescopeland: Got it, thanks for pointing this out. Should now be resolved in https://github.com/RetroAchievements/RAWeb/pull/1700/commits/a2da137cb9554818c12868ad94e986db44a26a49.

app/View/Components/Platform/Cards/Game.php
wescopeland: Understood, I've removed it in https://github.com/RetroAchievements/RAWeb/pull/1700/commits/a2da137cb9554818c12868ad94e986db44a26a49.

Should it also be removed from _UserCard.php_ here?: https://github.com/RetroAchievements/RAWeb/blob/515265d2cdf80d325f0d2a98521f775e9805143c/app/View/Components/Community/UserCard.php#L60

luchaos: that'd work as it won't have an effect there either 👍

wescopeland: No problem at all, addressed in https://github.com/RetroAchievements/RAWeb/pull/1700/commits/59f3825106064a9ce969faace1c0b20ed7449d6c.

luchaos:
Awesome stuff!

> This is probably the appropriate time to say I'm in favor of removing all emoji from the app and replacing them with standardized icons that better represent our brand.

I really like those dots; yet, maybe we implement different icons for completion? I think the current dot might resemble future points icons too closely.

> Use Case 4: Active Core Set + Site Award

When hovering the game "Super Mario 64" in "Completion Progress" first and the badge afterwards it will not contain the mastery information, when doing it the other way around it will - is it missing a context specifier for the cache key or is this by design?

https://preview.retroachievements.org/user/luchaos

app/Platform/Models/Game.php
```diff
@@ -85,6 +85,7 @@ class Game extends BaseModel implements HasComments, HasMedia
         'IsFinal',
         'RichPresencePatch',
         'GuideURL',
+        'Updated',
```
luchaos, line 88: Does it still have to be exposed or is it still used?

tests/Feature/Community/UserCardTest.php
luchaos, line 1: What do you think of moving the component tests into a dedicated `Components` sub-directory? I think we'll have a lot more of those in the future :)

app/Platform/Components/GameCard.php
```diff
@@ -0,0 +1,304 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Platform\Components;
+
+use App\Community\Enums\AwardType;
+use App\Community\Enums\ClaimStatus;
+use App\Community\Models\AchievementSetClaim;
+use App\Platform\Models\Game as GameModel;
```
luchaos, line 10: Is the alias needed? Usually resources in model form do not get a type suffix, any other class for the same resource would (e.g. `GameResource`, `GameFactory`, ... )

app/Support/Cache/CacheKey.php
```diff
@@ -49,6 +54,34 @@ public static function buildUserRecentGamesCacheKey(string $username): string
...
+     *
+     * This function creates a cache key for a specific entity with a given kind and optional parameters.
+     * Cache keys follow the format: "entityKind:12345:keyKind:{param1}:{param2}".
+     *
+     * @param string     $entityKind the kind of entity to cache data for (eg: "user", "game", "achievement")
+     * @param string|int $identifier the unique id associated with this entity (eg: 12345, "Scott")
+     * @param string     $keyKind the specific kind of data being cached (eg: "card-data")
+     * @param array      $params optional parameters for the cache key
+     *
+     * @return string the constructed cache key
```
luchaos, line 68: I tend to omit `@param` and `@return` in docblock if the description doesn't add context for highly complex structures. Usually it's easy to determine what parameters are for and/or what methods will return; especially since PHP allows us to specify types. It'S easier to maintain, too. We removed most of them when we introduced types.

However, if you think it's still worth keeping those I'd prefer the non-aligned variants as that's how the other docblocks formatted.

app/Support/Shortcode/Shortcode.php
```diff
@@ -232,7 +232,7 @@ private function embedAchievement(int $id): string
 
     private function embedGame(int $id): string
     {
-        $data = Cache::store('array')->rememberForever('game:' . $id . ':card-data', fn () => getGameData($id));
+        $data = Cache::store('array')->rememberForever('game:' . $id . ':game-data', fn () => getGameData($id));
```
luchaos, line 235: This makes the key slightly redundant - what is the reasoning for this change?
`game:1:game-data` vs `game:1:card-data` - would it make sense to specify its context here like `game:1:avatar-data`?

wescopeland: Without this change, there is a key collision (visible most especially on the cards demo page, which crashes altogether).

The two different keys are fetching different data, so this change seemed reasonable.

wescopeland:
>I really like those dots; yet, maybe we implement different icons for completion? I think the current dot might resemble future points icons too closely.

Definitely understand where you're coming from here, I want to argue against it though, for a few reasons.

1. What the Points branding will look like yet doesn't appear to be a settled discussion (though it will probably involve some spherical or rectangular shape, indeed).
2. I don't know that we're in a position to settle on Mastery branding, and not sure this should happen before settling on Points branding.
3. If we don't settle on Mastery branding but still want to pivot from the dot, to be consistent that means bringing over the emoji, which I'm strongly against.

I don't think we should settle on Mastery branding without polling the community and iterating once or twice. I definitely consider the dots as placeholders for whatever that branding winds up becoming; I'm just not comfortable doing something more involved here without first doing some kind of study or survey from the community.

However, perhaps we can meet in the middle. How would you feel about a common component being made for these, similar to what we did for the RetroPoints container? That way once we _do_ decide on a branding, we already have a common component ready to go for it. For the time being, that component can just use the dots.

app/Support/Cache/CacheKey.php
wescopeland: That makes sense. Some of these are definitely redundant, I didn't know if it was a good practice to only keep around the ones that add useful context while omitting the ones that are redundant. I've trimmed out a lot of these in https://github.com/RetroAchievements/RAWeb/pull/1700/commits/503bff6530c5ea54824c288909886aa2d3df93ac so hopefully the only params that could be ambiguous are still documented.

wescopeland:
>When hovering the game "Super Mario 64" in "Completion Progress" first and the badge afterwards it will not contain the mastery information, when doing it the other way around after a reload it will - is it missing a context specifier for the cache key or is this by design?

Great catch, this is a bug caused by a missing context specifier. Resolved in latest.

app/Platform/Models/Game.php
wescopeland: Indeed, still necessary.

Alexdatadestroyer:
Thanks for the answer @wescopeland! I read everything and I will give another answer regarding some aforementioned things below:

### 1. Answer given regarding Blue/Yellow dots. Medal/Crown emoji icons. Standardized icons idea of better branding representation...:

I didn't know that Medal and Crown images were emoji. Perhaps, we should create them as ordinary image to be referenced by HTML call instead of just being put as emoji? Or maybe... better to create our exclusives ones? Emojis are indeed variant, so, this might be an idea to solve this.

As a mockup, I meant something like this (Use Case 4 example):



I took the Crown and Medal icons through Source Code Inspector on Microsoft Edge Browser and edited on MS Paint to illustrate was I meant to say on my post before.

Also, 

> (PS: I know that Blue indicates Softcore while Yellow indicated Hardcore, indeed. However, on UP is represented differently for some reason: Hardcore progress is Blue while Softcore progress is Grey.)

This part of my comment wasn't answered before.

I agree with standardization of yellow (hardcore) and blue (softcore) colors - which also represents RA Mascot and RA Logo -, but why the personal player graph behaves differently? The progress graph used to behave the same way like happen on Set Pages.

- If you look on Set Pages, yellow progress represents hardcore unlocks and blue progress represents softcore unlocks. However, on User Profile, hardcore progress is blue and softcore progress is grey as shown below:



This little issue came to light when V2 was released in 2022 at beginning (19 Sep) before Playstation 2 release or on later versions. Can't map which version was exactly...

(**Extra Note:** If it's needed to open a separated issue to talk about this, then, tell me and I will make one later!)

### 2. Answer given regarding Use Case 6: ~Z~ tag. Pull Request scope is already too large. Labels are bolded and descriptions use base font weight...:

Understandable. I got some ideas for future discussion or for another Pull Request: What if we replace the "Z" styled block for "Retired" after Set Name? The ~Z~ tag would still apply on Set Page, but the Card Tooltip would show "Retired" instead. On RetroAchievements, we have 2 standardized colors: Blue and Yellow, so, adding a styled "X" mark on Set Badge would be another idea to indicate that the set is unavailable to be played anymore.

### 3. Answer given regarding Easter Egg Proposal...:

I think you misunderstood my point. I was talking about ONLY showing that message to the user without any Site Award to them. The idea of the Easter Egg is just to show something exclusive to that [bugged set page](https://retroachievements.org/game/11782) and nothing more. Since we are on a (Retro) Gaming Website and since this is something that is not so common to be found on RA Website that's why I proposed that as an Easter Egg feature.

- Let me see if I can give such an illustration through MS Paint editing:

The Card itself:



The Card Tooltip shown when Set Name is hovered:



This is just an example of the Easter Egg proposed above. I forgot to mention the inclusion of RA Mascot to it. Since the message is to be something ironic/funny, I added the "Amazed" reaction to it - which would fit well for this EE feature. "Gitgud" reaction would be fine too, I think.

This might be a subject of Community Votation as well. So, I agree with votation there on Discord to see if the community is fine with this feature or if this is useless. If approved, I don't think that might to be hard to implement that. If so, feel free to scrap that out.

For other RA Events, Card Tooltip behavior would remain normal like it was posted on first post!

Sorry for such a delayed answer since the revisions are came to light...

wescopeland:
Hi @Alexdatadestroyer,

I appreciate you taking the time to share your thoughts. I want to provide further clarity of the scope of this PR.

The purpose of this PR is specifically to implement the new game tooltip designs, which I outlined in detail in the initial description. Feedback on these designs was solicited from numerous stakeholders before and during implementation. While I always welcome and desire constructive feedback, I'd like to keep the discussion focused on the changes in this PR. This is not the appropriate venue to raise other issues or propose new features.

I've written in detail my opinion on completion icons here https://github.com/RetroAchievements/RAWeb/pull/1700#issuecomment-1658162225. I suggest creating separate issues or discussions for those ideas so they can get proper consideration. Pull requests work best when kept focused on the code at hand.

I don't want you to feel discouraged - on the contrary, we are fortunate to have someone who is energized and willing to help test issues on the site. But for this PR specifically, I'd like to keep the conversation centered on the code changes and implementation details contained therein.

Please let me know if I have not clarified this well enough. I appreciate you taking the time to provide feedback in RAWeb and the development process.

luchaos:
> > I really like those dots; yet, maybe we implement different icons for completion? I think the current dot might resemble future points icons too closely.
> 
> I appreciate this suggestion; I want to propose a different approach due to several considerations.
> 
> 1. The visual identity for Points branding remains an open discussion. Although it seems likely it'll incorporate a spherical or rectangular shape, it does seem to remain unsettled.
> 2. I believe we should not determine completion branding before reaching a consensus on Points branding. Establishing one before the other might lead to inconsistencies in our design language.
> 3. A pivot from the dot, in the absence of a decided completion branding, would likely mean defaulting to the emoji(s). This is an option I'm not in favor of.
> 
> I propose that, similar to the vote that recently occurred on RetroPoints branding (which we collected a lot of great feedback from), we perform a similar study/review on completion branding. For something this important, I am guarded and would like to gather community feedback and go through an iteration or two. In my view, the dots are temporary placeholders until we find a branding that suits our needs. That said, I'm hesitant to implement something more definitive without first conducting the surveys mentioned earlier.
> 
> Perhaps we can meet closer in the middle on this though. What if we create a common component, similar to our approach with the RetroPoints container? This way, once we agree on a branding or visual identity for completion indicators, we have a ready-to-go component for the new design. If it helps, maybe we can also change the dot to a square? Meanwhile, this common component can retain the dots as a placeholder.

Very good reasoning - thanks for all the context! Let's go with the dots 👍

luchaos (approved):

PR #1705, opened 2023-07-30
fix(layout): prevent layout shift with sidebars

luchaos:
Apply `lg:grid grid-cols-[1fr_340px] gap-4` in top brand, nav, and main content areas. 

Prevents layout shift and works with wide layouts (see /demo).

### DIFF
--- a/app/Helpers/render/layout.php
+++ b/app/Helpers/render/layout.php
@@ -107,7 +107,7 @@ function RenderToolbar(): void
 
         ],
     ];
-    echo "<ul class='flex-1'>";
+    echo "<ul>";
     echo "<li><a href='#'>Games</a>";
     echo "<div>";
     foreach ($menuSystemsList as $column) {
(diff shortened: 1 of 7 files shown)


wescopeland:
Not sure if intentional, but noticed a subtle change. The sidebar now extends the full height of the column. I don't think this is the end of the world, just want to call it out in case it was unintended.

Screenshot 2023-07-30 at 5 59 08 PM

luchaos:
> Not sure if intentional, but noticed a subtle change. The sidebar now extends the full height of the left container. I don't think this is the end of the world, just want to call it out in case it was unintended.

whops, thanks! i shouldn't have removed `align-self: start;` - i'll add that back

wescopeland (approved):
Great! I've confirmed no layout shift is present on both home and /demo:

Screenshot 2023-07-30 at 6 10 56 PM

Always love anything that gets us closer to enabling XL.

PR #1706, opened 2023-07-30
refactor: rename AchievementType to AchievementFlags

wescopeland:
Wanted to go ahead and get this unit of work out of the way so for the "type" PR we can have a better signal to noise ratio.

I'll continue on by branching off this PR.


app/Platform/Enums/AchievementFlags.php
luchaos, line 1: I'd go with `AchievementFlag`, what do you think?

wescopeland: No problem at all, that would align with me renaming the variables from `$flags` to `$flag`. I will make this adjustment tomorrow.

wescopeland: Actually was pretty trivial to do with Replace Symbol 😅 Should be resolved as of latest.

luchaos (approved):

PR #1714, opened 2023-08-04
feat: add back-end support for progression tier achievements and beaten game tracking

wescopeland:
This PR introduces database, back-end, and dev tooling support for beaten game tracking. 

>🔴🔴🔴 
**Note: This PR is not intended to consider any player-facing visible UX elements at this time. These will come once back-end support is ready. The scope of this PR is intentionally limited to what is mentioned in the first sentence of this description.**

The following discussions and conversations were referenced during the development process:
https://github.com/RetroAchievements/RAWeb/discussions/744
https://discord.com/channels/476211979464343552/1026595325038833725/1135226121546383472
https://discord.com/channels/476211979464343552/1026595325038833725/1135226612577742949

## Abstract
We aim to allow players to track the games they have beaten on the platform. This PR integrates this tracking support into the site's database and back-end. This is implemented through the introduction of "types", which can be assigned to achievements.

There are two types: "Progression" and "Win Condition". A game is considered beaten if:
* A game has both Progression and Win Condition achievements, and the player unlocks ALL Progression achievements and ANY Win Condition achievements.
* A game has Progression but no Win Condition achievements, and the player unlocks ALL Progression achievements.
* A game has Win Condition achievements but no Progression achievements, and the player unlocks ANY Win Condition achievements.

These scenarios are documented in [BeatenGameTest.php](https://github.com/RetroAchievements/RAWeb/pull/1714/files#diff-36d6fb891410dbb34298ea1241b4314b4889ddf84b975805044c400cdbdb50d7).

Upon unlocking an achievement, the system checks whether the player has beaten the game. If they have, a row is added to the `SiteAwards` table.

In the unique circumstance where the definition of beating a game changes (for example, due to the addition of a missed Progression achievement) after the player has already been awarded the beaten game status, the award will eventually be revoked. This stands apart from our current award offerings as it involves "taking" something from players. However, this action essentially acknowledges a potential oversight on our platform's part, suggesting that the user didn't actually beat the game under the correct conditions. In this light, the record is being corrected rather than penalizing the player, which makes this behavior reasonable and desirable.

When a user performs a score recalculation, their beaten game records will update. This will also occur automatically over time through the site's natural user scanning system for performing recalculations.

## High-level Summary of Changes
* A new column, `type`, has been added to the `Achievements` table.
* A new enum, `AchievementType`, has been introduced to the codebase, corresponding to the new column. This enum contains two values: "Progression" and "Win Condition".
* Through the achievement info and achievement inspector pages, developers can now associate achievements with types. An achievement can only have one type.
* A new function, `testBeatenGame()`, has been added and is called in `unlockAchievement()`. This function runs checks against the achievement types for the game and adds (or removes) beaten records to the `SiteAwards` table.
* The _recalculate-score.php_ endpoint now also recalculates the player's beaten game records.
* The _UpdatePlayerPoints.php_ command now also recalculates the player's beaten game records.

## Developer UX

**Achievement Info Devbox: New Optional Field**
Screenshot 2023-08-04 at 8 53 58 AM

**Achievement Inspector: New Toolbox, Improved Readability, Multi-adjust Types**
Screenshot 2023-08-04 at 8 54 54 AM

## Considerations
* Should anything be attached to the `$retVal` of _dorequest.php_ so emulator clients can be notified when a game is beaten?
* Without jobs and queues, it is not feasible to add beaten recalculations to the games themselves (similar to RetroPoints recalcs).
* I am currently setting `AwardDate` to when the user unlocked the achievement that granted them "beaten" status for the game. I assume this is desirable, as it will allow us to later build things like timelines which display when users actually beat the games for retroactive type assignments.


Alexdatadestroyer:
Nice feature!

However, in my point of view, the question that comes to my mind now is: **EVERY OfficialCore Achievement would have to be assigned to a Type (Progression / Win Condition) or it's nullable? (None)**

This kind of feature is confusing at first. Even reading most of conversations at the time to now (2018~2023), it's not possible for me to fully understand this. Those conversations are fragmented a lot! This might expect some RA Developers to go through a "heavy brainstorming" themselves before assigning a type to an Achievement in order to avoid future issues.

Besides, I hope this to be well documented on RADocs as if it's get released - **especially if this feature turns to be mandatory indeed.**  Don't take me as a naysayer toward this feature... I am saying that because there are some RA features that aren't well documented on RADocs yet :/

I would better go to "None" instead of addressing types for now to avoid any future issues with players or with Developer Compliance or Quality Assurance Teams. However, if it's turns to be mandatory, then, I would have some struggle before releasing a set on future as it would put more strings on it (Achievements).

wescopeland:
The type value is indeed nullable.

I have no insight into or opinion on whether this will be mandatory if the PR is merged. That would be something for DevC and/or QA to decide. I assume if it were hypothetically made mandatory, there would be plenty of good examples to reference and follow, and helpful community members to offer guidance and support along the way.

luchaos (approved):

PR #1729, opened 2023-08-09
perf: eliminate lodash

wescopeland:
There is only one lodash call in the entire app: `_.filter()` in the Active Players widget.

The lodash dependency is roughly 27KB unminified, and the entire app pays this tax due to this one call. On mobile 3G connections, this adds roughly 500ms of load time to an end user's first page view.

This PR migrates the `_.filter()` call to native JS `Array.prototype.filter()` and removes the lodash dependency.


luchaos (approved):
❤️

PR #1730, opened 2023-08-09
perf: eliminate jquery ui modal

wescopeland:
This PR swaps the site's modal implementation from jQuery UI Modal to Alpine.js.

Currently, only one page uses jQuery UI Modal, however it is depended upon by every page on the site. While the actual sizes of the jQuery UI Modal dependencies are very small, they are render-blocking.

A new component has been added: `<x-modal-trigger />`, which pops an accessible modal that works well across all breakpoints. It accepts content via a slot.

>ℹ️ NOTE: This PR simply ports the modal dialog content for these three modals to new Blade components. Some of this content is incorrect and needs to be changed later down the road.

**Desktop**

https://github.com/RetroAchievements/RAWeb/assets/3984985/3ac791b3-dbee-435c-bb93-98ff01f25042

**Mobile**

https://github.com/RetroAchievements/RAWeb/assets/3984985/7a57ad63-76ef-4c99-848e-2cf547314080


public/views/.gitkeep
wescopeland, line 1: Not sure if keeping this _views_ folder hanging around is desirable. If not, I can delete this _.gitkeep_ file.

package.json
```diff
@@ -14,6 +14,7 @@
         "apidoc:connect": "apidoc -i src/Connect/ -c docs/src/connect-api -t docs/src/connect-api/template -o docs/dist/api/connect"
     },
     "dependencies": {
+        "@alpinejs/focus": "^3.12.3",
```
wescopeland, line 17: https://bundlephobia.com/package/@alpinejs/focus@3.12.3

public/views/.gitkeep
luchaos: can be removed 👍

luchaos (approved):

public/views/.gitkeep
wescopeland: Removed in latest

PR #1733, opened 2023-08-10
perf(tooltip): use explicit event listeners

wescopeland:
This PR resolves a UI deadlock on the homepage for any browser that does not support `requestIdleCallback()`. Currently on these browsers, such as iOS Safari, the home page freezes for a few seconds before the user is able to do anything.

**Root Cause**
The Active Players widget has 2 tooltipped elements for every row on the table. Tooltipped elements are manually attaching event listeners right now when they're added to the DOM. Because this is happening thousands of times on the home page, it's causing the page to freeze until the work is done.

**Resolution**
The tooltip code has been refactored to a formal Alpine.js component. Tooltip consumers are now responsible for manually attaching their required event listeners.

There are some other cleanup tasks I took care of while finishing this work:
* All Alpine.js components in _resources/js_ are now suffixed with the word "component".
* All the tooltip code has been moved from _resources/js/tooltip_ to _resources/alpine/tooltipComponent_.


public/codenotes.php
```diff
@@ -178,9 +178,7 @@ function saveCodeNote(rowIndex, isDeleting = false) {
             rowEl.querySelector('td[data-current-author]').dataset.currentAuthor = currentUsername;
 
             const authorAvatarSpan = authorAvatarCellEl.querySelector('span.inline.whitespace-nowrap');
-            const newXInitValue = `attachTooltipToElement($el, { dynamicType: 'user', dynamicId: '${currentUsername}', dynamicContext: '' })`;
-            authorAvatarSpan.setAttribute('x-init', newXInitValue);
-            attachTooltipToElement(authorAvatarSpan, { dynamicType: 'user', dynamicId: currentUsername, dynamicContext: '' });
+            authorAvatarSpan.removeAttribute('@mouseover');
```
wescopeland, line 181: It is too much of a hassle to keep juggling the user prop on the tooltip as we're dynamically swapping it. Since the prop is just being changed to ourselves when we edit a code note, remove showing the tooltip altogether.

resources/js/alpine/tooltipComponent/tooltipComponent.ts
wescopeland, line 1: The implementation here is very nearly identical to the previous implementation, we're just returning functions now that can be used by the component's event listeners instead of attaching them in a `.forEach()` loop.

resources/js/tooltip/attachTooltipToElement.ts
```diff
@@ -1,141 +0,0 @@
...
-
-  const tooltipListeners = [
-    ['mouseover', handleMouseOver],
-    ['mouseleave', handleMouseLeave],
-    ['mousemove', handleMouseMove],
-  ];
-
-  tooltipListeners.forEach(([event, listenerFn]) => {
-    anchorEl.addEventListener(event as keyof HTMLElementEventMap, listenerFn as EventListener);
-  });
```
wescopeland, line 78: FWIW, this was the offending code.

luchaos (approved):

PR #1736, opened 2023-08-11
fix: use infobox icons to designate when title attrs are available

wescopeland:
Replaces the `*` on Guide URL and Type fields with an infobox icon. Also makes a few a11y improvements to those labels.

Screenshot 2023-08-10 at 8 06 55 PM
Screenshot 2023-08-10 at 8 07 17 PM

### DIFF
--- a/public/achievementInfo.php
+++ b/public/achievementInfo.php
@@ -297,7 +297,16 @@ RenderContentStart($pageTitle);
                 echo "</select>";
                 echo "</td></tr>";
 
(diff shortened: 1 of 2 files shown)


luchaos (approved):

luchaos:
Just remembered where i saw a similar thing before: https://retroachievements.org/reportissue.php?i=14071
The modal triggers use `?` for the button labels - we might want to find a standardized way to present both info titles and info modals at some point

wescopeland:
Agreed. FWIW, these are slightly different on stage to provide more than just "?":

Screenshot 2023-08-11 at 8 25 59 AM

luchaos:
> Agreed. FWIW, these are slightly different on stage to provide more than just "?":

oh, wonderful!

PR #1746, opened 2023-08-12
fix(userProfile): localize score and achievement count to match elsew…

drisc:
…here

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -285,7 +285,7 @@ RenderContentStart($userPage);
             $totalTruePoints = $userMassData['TotalTruePoints'];
 
             $retRatio = sprintf("%01.2f", $totalTruePoints / $totalHardcorePoints);
-            echo "Hardcore Points: " . localized_number($totalHardcorePoints) . "<span class='TrueRatio'> (" . localized_number($totalTruePoints) . ")</span></span><br>";
+            echo "Hardcore Points: " . localized_number($totalHardcorePoints) . "<span class='TrueRatio'> (" . localized_number($totalTruePoints) . ")</span><br>";
             echo "Hardcore Achievements: " . localized_number($totalHardcoreAchievements) . "<br>";
 
             echo "Site Rank: ";
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -285,8 +285,8 @@ RenderContentStart($userPage);
             $totalTruePoints = $userMassData['TotalTruePoints'];
 
             $retRatio = sprintf("%01.2f", $totalTruePoints / $totalHardcorePoints);
-            echo "Hardcore Points: $totalHardcorePoints <span class='TrueRatio'> ($totalTruePoints)</span></span><br>";
-            echo "Hardcore Achievements: $totalHardcoreAchievements<br>";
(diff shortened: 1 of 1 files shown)


public/userInfo.php
```diff
@@ -298,7 +298,7 @@ function resize() {
                 $userRank = $userMassData['Rank'];
                 $rankPct = sprintf("%1.2f", ($userRank / $countRankedUsers) * 100.0);
                 $rankOffset = (int) (($userRank - 1) / 25) * 25;
-                echo "<a href='/globalRanking.php?s=5&t=2&o=$rankOffset'>$userRank</a> / $countRankedUsers ranked users (Top $rankPct%)";
+                echo "<a href='/globalRanking.php?s=5&t=2&o=$rankOffset'>" . localized_number($userRank) . "</a> / " . localized_number($countRankedUsers) . " ranked users (Top $rankPct%)";
```
wescopeland, line 301: This doesn't match the user cards logic for rendering site rank. See `buildCardRankData()` in _UserCard.php_.

In the user card, the rank is prefixed with a "#", and there is conditional logic for whether the Top % copy should display.

public/userInfo.php
```diff
@@ -321,7 +321,7 @@ function resize() {
                 $userRankSoftcore = getUserRank($userPage, RankType::Softcore);
                 $rankPct = sprintf("%1.2f", ($userRankSoftcore / $countRankedUsers) * 100.0);
                 $rankOffset = (int) (($userRankSoftcore - 1) / 25) * 25;
-                echo "<a href='/globalRanking.php?s=2&t=2&o=$rankOffset'>$userRankSoftcore</a> / $countRankedUsers ranked users (Top $rankPct%)";
+                echo "<a href='/globalRanking.php?s=2&t=2&o=$rankOffset'>" . localized_number($userRankSoftcore) . "</a> / " . localized_number($countRankedUsers) . " ranked users (Top $rankPct%)";
```
wescopeland, line 324: Similar feedback as above for user rank display.

public/userInfo.php
drisc: Added "#" to the link, implemented the Top % hiding feature from UserCard.php

public/userInfo.php
```diff
@@ -285,8 +285,8 @@ function resize() {
             $totalTruePoints = $userMassData['TotalTruePoints'];
 
             $retRatio = sprintf("%01.2f", $totalTruePoints / $totalHardcorePoints);
-            echo "Hardcore Points: $totalHardcorePoints <span class='TrueRatio'> ($totalTruePoints)</span></span><br>";
-            echo "Hardcore Achievements: $totalHardcoreAchievements<br>";
+            echo "Hardcore Points: " . localized_number($totalHardcorePoints) . "<span class='TrueRatio'> (" . localized_number($totalTruePoints) . ")</span></span><br>";
```
wescopeland, line 288: This is unrelated to the PR, but I am just noticing that there's a bug on this line of code we should probably clean up while we're in here.

It appears there's an extra closing `</span>` tag here, possibly for something that used to exist but was removed.

drisc: Extra `span` removed

wescopeland (approved):

wescopeland:
Verified locally with several accounts:
Screenshot 2023-08-12 at 2 13 53 PM

Screenshot 2023-08-12 at 2 13 45 PM
Screenshot 2023-08-12 at 2 13 32 PM

luchaos (approved):

PR #1747, opened 2023-08-12
fix(achievement): dont save type: none as an empty string [V3.2.1]

wescopeland:


Tsearo (commented):
Fixes when updating the type back to None or when creating a new achievement so looks good to me. 
I may suggest having someone with root access running the following query to make sure there isn't any data discrepancy in the table from before this PR goes in as any achievements created between or set to None will be a blank string as opposed to NULL

`UPDATE Achievements
SET type = NULL 
WHERE type = ''`

luchaos (approved):

luchaos:
> someone with root access running the following query

will do

PR #1748, opened 2023-08-12
fix(game): remediate issue where more than 100% progress is shown [V3.2.1]

wescopeland:
Mitigates https://discord.com/channels/476211979464343552/1026595325038833725/1139734338041299065, which was introduced by the optimizations in #1712.

Now if unlock rate % is greater than 100% because Untracked players are being included, the UI will just fall back to showing 100%.


Tsearo (commented):
Looks good now with the conditionals back in

luchaos (approved):

PR #1749, opened 2023-08-12
fix(site-awards): filter out beaten games from awards displays [V3.2.1]

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1139786829290291241.

* Ensures beaten games are being fetched properly by the all-purpose `getUsersSiteAwards()` function.
* Ensures they're filtered out in the `SeparateAwards()` function so they don't increase the count of Site Awards on user profiles.


Tsearo (approved):
Fixes the count issue

PR #1757, opened 2023-08-13
fix(game): resolve screenshot height issue [V3.2]

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1140208098909372457.

The carousel styles are only visible now at the XS breakpoint. SM+ sees exactly the content that was present in release 3.1.

**Before**
Screenshot 2023-08-13 at 9 19 27 AM

**After**
Screenshot 2023-08-13 at 9 18 12 AM

### DIFF
--- a/resources/views/platform/components/game/screenshots.blade.php
+++ b/resources/views/platform/components/game/screenshots.blade.php
@@ -16,23 +16,35 @@ function carousel() {
 }
 </script>
 
-<div class="relative" x-data="carousel()">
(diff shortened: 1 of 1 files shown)


Tsearo (commented):
These still look a bit blown up and blurry when checking on mobile sized screens though I'm not sure if that is as much of a concern or is just due to using the device testing tool? For normal sized views on desktop it looks to be working fine

wescopeland:
Mobile is stretching to fit the width of the viewport. That is the pre-3.2 behavior, not sure if there's any good way around that while still maintaining a carousel component on mobile.

Tsearo (approved):

RALordAndrew:
DS screenshots are particularly susceptible to blurriness. Other consoles' screenshots get scaled down to 320 px wide on upload, but 256x384 px DS screenshots are scaled down to 160x240 due to the maximum height restriction. https://github.com/RetroAchievements/RAWeb/blob/master/app/Helpers/util/media.php#L133

PR #1759, opened 2023-08-13
chore: remove deleted models feature

luchaos:
After merging v1 and v2 schemas in v3 keeping track of deleted models is not necessary anymore.

This table/model was meant to keep track of deletions across multiple tables during v1 -> v2 syncs. Due to the in-place schema merge no additional tables were added for the resources that were recorded in this table.

An activity/audit log will keep track of deletions and other model changes in the future.


wescopeland (approved):
Verified the migration runs and all affected code paths are still working correctly.

PR #1760, opened 2023-08-13
chore: php8.2 upgrade

luchaos:
We're already running PHP 8.2 and Ubuntu 22.04 on our servers. This update reflect those changes in the application. To be able to upgrade to Laravel 10 and use real Enums we need 8.1+. 8.0 will be EOL (end of life) in two months.

Note: a composer update to pull in PHP 8.2 dependencies will follow in a separate PR because it will require some changes.

Please update your local environment:
- XAMPP/LAMPP/WAMPP: re-install/update
- Docker: `sail up`
- Valet: `valet use 8.2` or `valet isolate php@8.2`
- VM/WSL/Ubuntu: dist-upgrade to 22.04 (optional, i think. please verify)


wescopeland (approved):
Verified that running `composer setup` and `sail up` work great. I had no issues upgrading and visiting numerous pages on the site.

luchaos:
> Verified that running `composer setup` and `sail up` work great. I had no issues upgrading and visiting numerous pages on the site.

Fantastic, thanks for confirmation!
I have it running with a hybrid Valet and Docker/Sail setup on macOS so that's definitely working, too.

Let's wait for some feedback from Windows/Linux users before merging. Maybe those setups require more steps that should be documented.

luchaos:
Thinking about it - this update does not _require_ anyone to upgrade to php 8.2 yet.
Upcoming dependency updates will. Until then we can figure out any setup issues on other systems.

PR #1761, opened 2023-08-13
fix(user-game-lists): unique set requests

luchaos:
Manually remove duplicates from SetRequest table:
```sql
SELECT sr.User, sr.GameID, COUNT(*) count FROM SetRequest sr GROUP BY sr.User, sr.GameID HAVING count > 1;
```

Populate `user_id` column:
```sql
UPDATE SetRequest sr SET user_id = (SELECT ID FROM UserAccounts ua WHERE ua.`User` = sr.`User`);
```

Populate `type` column while at it:
```sql
UPDATE SetRequest sr SET `type` = 'achievement_set_request';
```


wescopeland:
Do you think it would be also worth disabling the Request Set button on click to mitigate superfluous calls to the API layer?

luchaos:
> Do you think it would be also worth disabling the Request Set button on click to mitigate superfluous calls to the API layer?

I think it was discussed at some point to make that the default behaviour for _all_ submit buttons at some point. So a global handler would make sure to disable all interactive buttons during a request.
We might want to explore that again when the form components are all in place so it's easier to wire it up? This is not the only button which is susceptible to this problem i assume

wescopeland (approved):
Verified as functional. I am no longer able to make duplicate set requests.

PR #1762, opened 2023-08-13
feat(GameCard): use 96x96 game and hub icons

wescopeland:
This PR removes all scaling from game and hub icons in the game card component.

**Summary of Changes**
* `<x-card.container />` accepts a new prop: `imgKind`, which defaults to "user". If this prop is set to "game", the card image uses 96x96 resolution.
* The "Last Updated" row in game cards has been removed because it's not entirely accurate. The row still remains in hub cards.
* The font size has been fixed to match user cards.
* RetroPoints has been combined with the Retro Ratio row (which also has the added benefit of establishing a stronger visual connection between the two tightly-coupled elements).
* The Active Players widget now is using the new cards for games, it was omitted by accident.
* A few more cases have been added to the cards test page, and a CSS adjustment has been made so the card components respect their internal heights rather than the height of the row they're on.

**Before**
image

**After**
Screenshot_2023-08-13_at_3 45 53_PM


wescopeland:
I am interested on getting feedback for if we should put a label after the ratio label.

In the screenshot in the OP, the RetroPoints row currently reads:
**RetroPoints:** 585 - 1.46x

Maybe
"1.46x Rarity" or "1.46x Ratio"

resources/views/platform/components/cards/game.blade.php
```diff
@@ -39,32 +39,30 @@
...
-                        {{ $lastUpdated->format('j F Y') }}
-                    </x-card.info-row>
-                @endif
+                        <!-- RetroPoints & Retro Ratio -->
+                        <x-card.info-row label="RetroPoints">
+                            <span>
+                                {{ $retroPointsSum > 0 ? localized_number($retroPointsSum) : 'None yet' }}
+                                @if ($retroRatio != 0)
+                                    &ndash;
+                                    <em>{{ $retroRatio }}x</em>
```
luchaos, line 60: The `&times;` symbol could be a good fit here?

luchaos (approved):

luchaos:
> I am interested on getting feedback for if we should put a label after the ratio indicator.
> 
> 
> 
> In the screenshot in the OP, the RetroPoints row currently reads:
> 
> **RetroPoints:** 585 - 1.46x
> 
> 
> 
> Maybe
> 
> "1.46x Rarity" or "1.46x Ratio"

I think I'd go with a ratio rather than rarity label and put it in paratheses as it's directly related to the value in this case 🤔

wescopeland:
I tried without the "Rarity" label and agree it probably lacks context for newcomers. I've added `&times;`, the label now reads like:

RetroPoints: 585 (×1.46 Rarity)

Screenshot 2023-08-16 at 5 11 16 PM

PR #1765, opened 2023-08-15
fix(home): remove console icons from claims rows

wescopeland:
This PR simply removes console icons from the "New Sets/Revisions" and "Sets in Progress" components on the home page for both desktop and mobile.

With the amount of emoji in the Active Players widget which these tables surround, the icons add to the overall visual clutter and noise on the screen. Removing them helps mitigate the perceived clutter.

**Before**
Screenshot 2023-08-15 at 4 14 02 PM

**After**
Screenshot 2023-08-15 at 4 12 16 PM


luchaos (approved):
Going through the main menu's "Games" section to look for inconsistencies I realized that only "Most Requested Sets" exposes the system name, all the others do not even include it (might be a worth an improvement).
"Most Requested Sets" does not include the icon either in its current state so it looks consistent to me so far 👍

PR #1766, opened 2023-08-15
feat(achievementInspector): hide code rows by default

wescopeland:
Based on feedback from #devs-chat, hides code rows by default on the achievement inspector page. They can still be enabled by clicking the new "Toggle Code Rows" button introduced in V3.2.

Testable on a URL such as /achievementinspector.php?g=20491.


Tsearo (approved):
Works great and feels much better being hidden by default 👍

PR #1767, opened 2023-08-15
Add achievement type filtering to Ticket Manager

Tsearo:
Adds functionality to the Ticket Manager to support preparing for #1714 by allowing devs to filter only progression type achievements (Progression or Win Condition).

image

For individual tickets it also adds a Type row that displays if a type is set

Normal:
image

Progression:
image


wescopeland:
A great addition! I've verified locally and confirmed this is working. There is only one thing I'm getting hung up on:

Screenshot 2023-08-15 at 9 00 19 PM

I'm thinking "Type:" as a label for this row is probably ambiguous. Maybe it should be more explicit similar to the "Report Type:" row?

wescopeland (approved):

luchaos:
> I'm thinking "Type:" as a label for this row is probably ambiguous. Maybe it should be more explicit similar to the "Report Type:" row?

I agree, it should be something like "Achievement Type" or "Concept Type"? (I like to see achievements as concepts rather than implementation details, but that might be a bit much at this point)

luchaos:
Should we change the label in this PR or merge it as-is and think about the label in a separate PR?

wescopeland:
The label appears to have been adjusted in https://github.com/RetroAchievements/RAWeb/pull/1767/commits/a032ebf7b9f75b573d540c74c0f8ab2793f44f41. We are probably good to merge.

PR #1770, opened 2023-08-17
fix(seeders): adjust seeders and allow login with seeded users

luchaos:
Each role and user status gets a seeded user, password is the same as username for login:
```
Root
Administrator
Moderator
Developer
Verified
Unverified
Unranked
Banned
Spammer
```


database/seeders/NewsTableSeeder.php
```diff
@@ -15,6 +15,6 @@ public function run(): void
             return;
         }
 
-        News::factory()->count(5)->create();
+        News::factory()->count(15)->create();
```
luchaos, line 18: 15 news entries to make sure the carousel works (requires 10)

database/factories/NewsFactory.php
```diff
@@ -24,7 +25,7 @@ public function definition(): array
     {
         return [
             'Title' => ucwords(fake()->words(2, true)),
-            'user_id' => $this->seedUserByUsername(fake()->userName)->id,
+            'user_id' => User::inRandomOrder()->first()->ID,
```
luchaos, line 28: use existing seeded users as news authors

app/Helpers/util/string.php
```diff
@@ -29,6 +29,10 @@ function attributeEscape(?string $input): string
 
 function isValidUsername(?string $username): bool
 {
+    if (Str::startsWith($username, '__')) {
+        return true;
+    }
```
luchaos, line 34: make sure that seeded users can be used and accessed

Jamiras: The `CtypeAlnum` validator used by the API (and the registration page) doesn't allow underscores:
```
$ curl "localhost/API/API_GetUserSummary.php?y=engineer-secret&u=__administrator"

{"message":"The u must only contain unaccented letters and numbers.",
 "errors":{"u":["The u must only contain unaccented letters and numbers."]}}
```
It will also choke on the dashes in things like `release-manager`.

For a purely seeded database, why can't we just use PascalCase? `Administrator`, `ReleaseManager`, etc.

database/factories/NewsFactory.php
luchaos: Good point, I'll add those

app/Helpers/util/string.php
luchaos: I think I initially added those prefixes to not run into conflicts with an existing imported dump but that's not how it should be used in the first place.
I'll remove those and make sure that the "safe" variants are used for usernames to make them work with our validation rules - thanks for finding that!

luchaos: Went with your suggestion to use PascalCase: https://github.com/RetroAchievements/RAWeb/pull/1770/commits/90627af82e73f4be7f15940613d84cd4b2a580c4

database/factories/NewsFactory.php
luchaos: Adressed in https://github.com/RetroAchievements/RAWeb/pull/1770/commits/90627af82e73f4be7f15940613d84cd4b2a580c4

Jamiras (approved):

PR #1774, opened 2023-08-17
chore: normalize buttons

luchaos:
One of the first components PRs extracted from the prototype/preview branch.

Normalize button usage:
- No `<input type="submit">`
- No `<button type="submit">` (`submit` is implicit)
- No implicit styling of `<button>` (relevant for upcoming PRs)
- Buttons with JS handlers use `type="button"`

That's the third or fourth time I'm normalizing those buttons now 🤔 
I hope this time it sticks 😅

### DIFF
--- a/app/Helpers/render/layout.php
+++ b/app/Helpers/render/layout.php
@@ -261,7 +261,7 @@ function RenderToolbar(): void
     echo "<form class='flex searchbox-top' action='/searchresults.php'>";
     // echo "Search:&nbsp;";
     echo "<input name='s' type='text' class='flex-1 searchboxinput' value='$searchQuery' placeholder='Search the site...'>";
-    echo "<input type='submit' value='🔎︎' title='Search the site'>";
+    echo "<button class='btn' title='Search the site'>🔎</button>";
     echo "</form>";
 }
 
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -139,12 +139,12 @@ if ($v != 1) {
                 <div class="flex flex-col sm:flex-row gap-4 sm:gap-2">
                     <form id='escapeform' action='/gameList.php'>
                         <input type='hidden' name='c' value='<?= $consoleID ?>'/>
-                        <input type='submit' class='leading-normal' value='No. Get me out of here.'/>
+                        <button class='btn leading-normal'>No. Get me out of here.</button>
                     </form>
 
                     <form id='consentform' action='/game/<?= $gameID ?>'>
                         <input type='hidden' name='v' value='1'/>
-                        <input type='submit' class='leading-normal' value='Yes. I&apos;m an adult.'/>
+                        <button class='btn leading-normal'>Yes. I'm an adult.</button>
                     </form>
 
                     <?php if ($userWebsitePrefs): ?>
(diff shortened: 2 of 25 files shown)


resources/views/components/modal-trigger.blade.php
```diff
@@ -39,13 +39,13 @@ class="bg-box-bg relative w-full h-screen max-h-screen overflow-y-auto sm:h-auto
             >
                 <div class="flex items-center justify-between pb-2">
                     <p class="text-lg font-semibold mb-4">{{ $modalTitleLabel }}</p>
-                    <button
```
luchaos, line 42: It looks like VS Code ignores the editorconfig for trimming whitespace?
https://github.com/RetroAchievements/RAWeb/blob/master/.editorconfig#L9
`trim_trailing_whitespace = true`

resources/css/button.css
```diff
@@ -1,4 +1,4 @@
-.btn, button, [type="button"], [type="reset"], [type="submit"] {
```
luchaos, line 1: While popular in CSS frameworks like bootstrap and Daisy UI (where this is from) it's too greedy for our use

app/Helpers/render/layout.php
```diff
@@ -261,7 +261,7 @@ function RenderToolbar(): void
     echo "<form class='flex searchbox-top' action='/searchresults.php'>";
     // echo "Search:&nbsp;";
     echo "<input name='s' type='text' class='flex-1 searchboxinput' value='$searchQuery' placeholder='Search the site...'>";
-    echo "<input type='submit' value='🔎︎' title='Search the site'>";
+    echo "<button class='btn' title='Search the site'>🔎</button>";
```
wescopeland, line 264: Adding a border to this button probably isn't desirable.

Prod:
Screenshot 2023-08-17 at 5 40 53 PM


PR:
Screenshot 2023-08-17 at 5 40 10 PM

public/gameInfo.php
wescopeland, line 1: The 3rd button on the page was missed:

Screenshot 2023-08-17 at 5 44 41 PM

resources/views/components/modal-trigger.blade.php
wescopeland: Confirmed, VSCode does indeed ignore _.editorconfig_ unless a third-party extension is installed.

resources/css/button.css
wescopeland: We have a few of these global style overrides lurking, hopefully we can systematically remove them over time.

public/gameInfo.php
luchaos: Good catch! Fixed in https://github.com/RetroAchievements/RAWeb/pull/1774/commits/8bbcf4791a54bd8d54be2004ed3eb0ea4a955432

app/Helpers/render/layout.php
luchaos: Good catch! Fixed in https://github.com/RetroAchievements/RAWeb/pull/1774/commits/8bbcf4791a54bd8d54be2004ed3eb0ea4a955432

wescopeland (approved):

PR #1776, opened 2023-08-17
feat: navbar blade components

luchaos:
- Port navbar to blade components
- (Re)introduce dropdown components
- Adjustments for small screens (can certainly be improved)
- Adjust policies to move blade components and controllers to model abilities (`@can()` and `authorize()`)
- Add point variants to navbar

Most of these changes are in preparation for the introduction of the permission matrix and getting rid of the user top card on sub-pages, hence moving point variants to navbar, too.

Increased font size, 4 columns in "Games", ported system icons:
Screenshot 2023-08-18 at 01 26 31

Search input expands on focus, no input on small screens (links to search page instead):
Screenshot 2023-08-18 at 01 26 55

Manage dropdown:
Screenshot 2023-08-18 at 01 27 12

"My pages" dropdown moved to profile menu with avatar:
Screenshot 2023-08-18 at 01 27 21

Small screen variants:
Screenshot 2023-08-18 at 01 35 23
Screenshot 2023-08-18 at 01 35 32
Screenshot 2023-08-18 at 01 35 43


wescopeland:
Mobile search does not appear to be working.


https://github.com/RetroAchievements/RAWeb/assets/3984985/fb15db9c-e255-49b7-99dd-27d4058bc56f

It probably makes sense for search to be on its own row on mobile similar to prod today. I believe having the huge tap target is desirable:
Screenshot 2023-08-17 at 7 38 57 PM

resources/views/community/components/news/carousel-item.blade.php
```diff
@@ -5,8 +5,8 @@ class="relative box-content flex w-full flex-none snap-start min-h-[270px]"
     {{ $index === 0 ? 'id=first-news-item' : '' }}
     {{ $index === $totalCount - 1 ? 'id=last-news-item' : '' }}
 >
-    <div 
-        class="z-10 py-6 px-8 absolute top-0 left-0 w-full h-full"
```
luchaos, line 9: Not sure if the z-index was necessary because of the menu dropdowns before but leaving those im place put the news items in front of the dropdowns.

wescopeland: Without this z-index value, the text appears behind the images.
This can be directly observed by moving the mouse over the author's name. The tooltip no longer appears because the text is not being hovered over.

luchaos:
> Mobile search does not appear to be working.

That's on purpose:
> no input on small screens (links to search page instead)

I can take a look at it again but I imagined the search to be improved in a separate PR where it becomes a full screen overlay and offers more options.

resources/views/community/components/news/carousel-item.blade.php
luchaos: I see! I'll revisit that then

wescopeland:
>I imagined the search to be improved in a separate PR where it becomes a full screen overlay and offers more options.

Even though its UX is gross, from a purely functional point of view, search in its present state is quite useful on mobile. I will of course defer to your judgment but I do hold the opinion that not having the ability to use search without being on a dedicated page does seem like a net loss.

wescopeland:
I am not sure why, but the carousel scroll buttons now have enlarged, blurry, and off-center icons. Perhaps this is only happening on my local?

**This PR**
Screenshot 2023-08-17 at 7 47 34 PM

**Prod**
Screenshot 2023-08-17 at 7 47 43 PM

resources/views/components/menu/account.blade.php
```diff
@@ -1,49 +1,74 @@
...
-            </x-form>--}}
-        </x-nav-dropdown>
-    </ul>
-    {{--<ul class="navbar-nav">
-        <li class="nav-item hidden lg:inline-block"><span class="nav-link">{{ request()->user()->points_total }}</span></li>
-    </ul>--}}
+    <div class="nav-link flex-col justify-center text-xs" style="line-height: 1em">
+        <div class="text-color">{{ number_format($user->points_total) }}</div>
+        <div class='trueratio'>{{ number_format($user->points_weighted_total) }}</div>
+        <div class='softcore'>{{ number_format($user->RASoftcorePoints) }}</div>
```
wescopeland, line 33: nit: It may be worth adding a title to these divs (with the exception of the weighted points row), otherwise we're only expressing meaning via color.

resources/views/components/menu/account.blade.php
```diff
@@ -1,49 +1,74 @@
...
-            {{--<x-dropdown-item :link="route('settings')">{{ __res('setting') }}</x-dropdown-item>--}}
-            {{--<x-form :action="route('logout')">
-                <button class="dropdown-item">{{ __('Sign Out') }}</button>
-            </x-form>--}}
-        </x-nav-dropdown>
-    </ul>
-    {{--<ul class="navbar-nav">
-        <li class="nav-item hidden lg:inline-block"><span class="nav-link">{{ request()->user()->points_total }}</span></li>
-    </ul>--}}
+    <div class="nav-link flex-col justify-center text-xs" style="line-height: 1em">
```
wescopeland, line 30: How would you feel about adding `items-end` to this, so the numbers are right-aligned?

wescopeland, line 30: `line-height: 1em` is just tight enough that the comma separators are bleeding into the next line.
How do you feel about `text-2xs` and a line height of 1.1em? With those adjustments I seem to have fixed the line bleed.

luchaos:
> > I imagined the search to be improved in a separate PR where it becomes a full screen overlay and offers more options.
> 
> Even though its UX is gross, from a purely functional point of view, search in its present state is quite useful on mobile. I will of course defer to your judgment but I do hold the opinion that not having the ability to use search without being on a dedicated page does seem like a net loss.

That's fair! Revisited in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/fde4f681d39001bfa501256dbf8f3e6d6b2a48b1, showing the input on small screens again. Still not the best solution but at least it's possible?

luchaos:
> I am not sure why, but the carousel scroll buttons now have enlarged, blurry, and off-center icons. Perhaps this is only happening on my local?

Missed the `btn` class on those; addressed in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/aa5e2af4b190b0bddd3d3e343e0224c6cd57ba62

wescopeland:
> That's fair! Revisited in [fde4f68](https://github.com/RetroAchievements/RAWeb/commit/fde4f681d39001bfa501256dbf8f3e6d6b2a48b1), showing the input on small screens again. Still not the best solution but at least it's possible?

This is way better. I use search on my tablet all the time. The only thing I'm noticing now is things get a bit wonky on iPhone SE. Other iPhone sizes all work great though.

Screenshot 2023-08-17 at 8 10 33 PM

resources/views/components/menu/account.blade.php
luchaos: My words :) I'll add that right away

resources/views/components/menu/account.blade.php
luchaos: Fantastic! Addressed in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/7b4ecc810aa992393c89f6e0a85b89e2a11b7e80

luchaos: Very nice, addressed in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/7b4ecc810aa992393c89f6e0a85b89e2a11b7e80

resources/views/components/menu/account.blade.php
```diff
@@ -1,49 +1,74 @@
...
+        <div class="text-color">{{ number_format($user->points_total) }}</div>
+        <div class='trueratio'>{{ number_format($user->points_weighted_total) }}</div>
```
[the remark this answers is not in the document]
luchaos: Added in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/7b4ecc810aa992393c89f6e0a85b89e2a11b7e80

luchaos:
> This is way better. I use search on my tablet all the time. The only thing I'm noticing now is things get a bit wonky on iPhone SE. Other iPhone sizes all work great though.

Targeting iPhone SE's 375px viewport with 1,000,000 points in: https://github.com/RetroAchievements/RAWeb/pull/1776/commits/6bb36056619162897109c982d7bfddd31127c935

resources/views/components/menu/account.blade.php
luchaos: Assuming we'll introduce similar containers like `x-points-weighted-container` at some point, what do you think would be good explanatory titles?

wescopeland: Good question. Maybe to be safe for now, we keep it as generic and factual as possible.

`title="Points earned in hardcore mode"`
`title="Points earned in softcore mode"`

resources/views/community/components/news/carousel-item.blade.php
luchaos: Changing the render order of items seems to do the trick for me without z-index 🤔 
https://github.com/RetroAchievements/RAWeb/pull/1776/commits/32c6f8686940e846825bfcac3593c6a332b39913

wescopeland: Confirmed, news carousel looks good now

wescopeland:
> Targeting iPhone SE's 375px viewport with 1,000,000 points in: [6bb3605](https://github.com/RetroAchievements/RAWeb/commit/6bb36056619162897109c982d7bfddd31127c935)

Issue is resolved now, I can no longer cause the overflow issue to happen.

wescopeland:
It appears the nav isn't sticking to the top of the screen. Almost like there's 1 or 2 pixels of margin. Can be seen by watching right above the nav while scrolling the page, you can see remnants of text flying by:


https://github.com/RetroAchievements/RAWeb/assets/3984985/16ff977a-adbe-4cb1-a7b8-640c6db49af8

resources/views/components/menu/account.blade.php
luchaos: sgtm! added in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/08c46f7ac135db041966e9c8756398b1e0e4e914

luchaos:
> It appears the nav isn't sticking to the top of the screen. Almost like there's 1 or 2 pixels of margin. Can be seen by watching right above the nav while scrolling the page, you can see remnants of text flying by:
> 
>  Sticky.Nav.mp4

Ah, yes, that fixed height was in preparation for when the whole top area is in its final state. Too soon. Removed in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/49c74b16dcaad35ff220365c4a3051ffea426dd9

wescopeland:
Still appears to be happening on latest 😞
I took a more zoomed in recording to help see. This is after pulling latest and restarting Vite.


https://github.com/RetroAchievements/RAWeb/assets/3984985/68d9b4fc-becd-413e-ac76-9af44690c867

public/userList.php
```diff
@@ -94,7 +94,7 @@
         }
         echo "<th><a href='/userList.php?s=$sort1&p=$perms" . ($showUntracked ? "&u=1" : '') . "'>User</a></th>";
         echo "<th><a href='/userList.php?s=$sort2&p=$perms" . ($showUntracked ? "&u=1" : '') . "'>Points</a></th>";
-        echo "<th><a href='/userList.php?s=$sort3&p=$perms" . ($showUntracked ? "&u=1" : '') . "'>Num Achievements Earned</a></th>";
```
luchaos, line 97: That edit wasn't supposed to be in this PR. Still, a change I wanted to see for quite some time 😅

luchaos:
> Still appears to be happening on latest 😞 I took a more zoomed in recording to help see. This is after pulling latest and restarting Vite.
> 
>  Area.mp4

Very weird, I can't seem to reproduce it on any type of page 🤔

wescopeland:
So strange - when switching to a different monitor I don't see the issue anymore 😵‍💫
It is probably just something specific to my setup then. If we see it after this is merged I'll see if I'm able to figure out what the culprit is.

Tsearo:
I wasn't able to reproduce the nav bar error at first but then tried a few different device sizes and could reproduce with Nest Hub Max (1280x800) at 95% zoom (100% zoom did not have the effect)

luchaos:
Just opened the page on my other machine and saw the the same gap:
Screenshot 2023-08-18 at 10 34 56

However, that's an old state of the master branch without any of the changes of this PR.
I think we should look into this separately.

wescopeland (approved):

resources/views/components/menu/account.blade.php
```diff
@@ -1,49 +1,76 @@
-{{--@guest
+<?php
```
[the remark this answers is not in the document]
luchaos: I agree and there certainly is! The prototype went quite a few steps farther than this and it was planned to remove the card eventually. 
We can have this right away, yet, while working on this I realized that it would blow up this PR quite a bit though because it requires some additional changes.
I think it would be best for me to follow up with another PR to address/discuss those changes separately?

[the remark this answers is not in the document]
Jamiras: What about just removing the icon and points from the top widget so it's just username, mail, and tickets?

resources/views/components/menu/search.blade.php
```diff
@@ -15,6 +15,6 @@
 }
 ?>
 <form class="flex searchbox-top" action="/searchresults.php">
-    <input name="s" type="text" class="flex-1 searchboxinput hidden lg:block" value="{{ $searchQuery }}" placeholder="{{ __('Search') }}">
+    <input name="s" type="text" class="flex-1 searchboxinput" value="{!! $searchQuery !!}" placeholder="{{ __('Search') }}">
```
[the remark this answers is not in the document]
luchaos: Good call. Tbh, I wasn't sold on that particular retro icons set myself.
Changed to fontawesome icons in: https://github.com/RetroAchievements/RAWeb/pull/1776/commits/ef11590615eacae0abdc7648e84db8dcf8073ae2
Up for discussion if we want to change those icons in the future. The important part here is that they exist for small screens.

Screenshot 2023-08-27 at 13 05 10

resources/views/components/menu/account.blade.php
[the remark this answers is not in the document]
luchaos: That works for now, too - done in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/1e2bde229b1f399ed5f7186e353ad77aef9fac4d

Screenshot 2023-08-27 at 13 04 26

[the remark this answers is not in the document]
luchaos: Heads up: this is what it looks like with the remaining elements moved to the navbar:
Screenshot 2023-08-27 at 13 08 55
Screenshot 2023-08-27 at 13 09 01

Allows us to drop the top card, make the brand logo more prominent, or get rid of the whole top brand area altogether (be it just on pages except on home, or on home, too)

resources/css/dropdown.css
```diff
@@ -0,0 +1,67 @@
...
+.dropdown-header {
+  padding: 0 12px;
+  margin-top: 8px;
+  text-align: center;
+  color: var(--text-color);
+}
+
+.dropdown-divider {
+  border-bottom: 1px solid var(--embed-color);
+  margin: 2px 10px;
```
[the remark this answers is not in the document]
luchaos: I thought it's fine to make them less present as it's mostly about the spacing - yet, I can see how it's a bit, too dim.
Introduced divider color to have better control over contrast: https://github.com/RetroAchievements/RAWeb/pull/1776/commits/6a39febd05eca98756fae1a897b0929f3b410dc1
Screenshot 2023-08-27 at 16 00 26

resources/views/components/menu/management.blade.php
```diff
@@ -1,83 +1,87 @@
...
+            <x-dropdown-header>{{ __('System') }}</x-dropdown-header>
             @can('viewHorizon')
                 <a class="dropdown-item" href="{{ url('horizon') }}" target="_blank">
                     Horizon Queue
                     <x-fas-external-link-alt/>
                 </a>
             @endcan
             @can('viewLogs')
-                <a class="dropdown-item" href="{{ url('log-viewer') }}" target="_blank">
+                <a class="dropdown-item" href="{{ route('log-viewer.index') }}" target="_blank">
```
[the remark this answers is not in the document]
luchaos: It's handled in the [AuthServiceProvider](https://github.com/RetroAchievements/RAWeb/blob/master/app/Site/AuthServiceProvider.php#L120):
```php
Gate::define('viewLogs', fn (Authenticatable $user) => $user->can('root'));
```

Because there's no explicit `root` gate defined it will default to the [`Gate::before()` closure](https://github.com/RetroAchievements/RAWeb/blob/master/app/Site/AuthServiceProvider.php#L42-L57)
```php
if ($user->Permissions >= Permissions::Root) {
  return true;
}
```

The root permission is not assigned to anyone but will be introduced again to make the permission matrix portable. This was not meant to be in this PR yet - reverted in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/ce38b521388f9ecf270e85e31c2c736b5f1c944d

resources/views/components/menu/account.blade.php
```diff
@@ -1,49 +1,78 @@
...
+        @if($user->RASoftcorePoints)
+            <div class='softcore cursor-help' title="Points earned in softcore mode">{{ localized_number($user->RASoftcorePoints) }}</div>
+        @endif
+    </div>
+    <x-nav-dropdown trigger-class="py-0" dropdown-class="dropdown-menu-right">
+        <x-slot name="trigger">
+            <x-user.avatar :user="$user" display="icon" iconSize="sm" :link="false" :tooltip="false"/>
+        </x-slot>
+        <x-dropdown-header>{{ $user->username }}</x-dropdown-header>
+        <div class="dropdown-divider"></div>
```
Jamiras, line 48: Minor: user menu header has a separator between [username] and Profile.
image

Other menus do not:
image

image

resources/views/components/menu/account.blade.php
```diff
@@ -1,49 +1,78 @@
...
+        <x-dropdown-item :link="route('user.show', $user)">{{ __res('profile', 1) }}</x-dropdown-item>
+        @if($user->ContribCount > 0 || $user->Permissions >= Permissions::JuniorDeveloper)
+            <div class="dropdown-divider"></div>
```
Jamiras, line 51: Extra divider if I set my permissions to Registered:
image

Why is this dependent on ContribCount?

I think you mistranslated this logic: https://github.com/RetroAchievements/RAWeb/blob/6baf30005587144bacebfadcb246c691360e7f3b/app/Helpers/render/layout.php#L193-L200

resources/views/components/menu/management.blade.php
```diff
@@ -1,83 +1,87 @@
...
             @endcan
-            <a class="dropdown-item {{ request()->routeIs('integration.release*') ? 'active' : '' }}"
-               href="{{ route('integration.release.index') }}">
-                Integration
-            </a>
-            <div class="dropdown-divider"></div>
-        @endcan
+            @can('manage', App\Community\Models\Forum::class)
+                {{--<x-dropdown-item :link="route('forum-topic.clearing')">Forum Clearing</x-dropdown-item>--}}
+                <x-dropdown-item :link="url('viewforum.php?f=0')">Forum Clearing</x-dropdown-item>
```
Jamiras, line 38: While "Invalid Forum Posts" implies there's something wrong with them, this makes no sense to me. Maybe "Unvalidated Forum Posts" or "Unauthorised Forum Posts"?

luchaos: Ah, yes, that's a leftover term from the prototype. What do you think about "Forum Verification" to make "Forum" the first word?

resources/views/components/menu/account.blade.php
luchaos: That's fair, removed in https://github.com/RetroAchievements/RAWeb/pull/1776/commits/d119b22aa484f3576d9516d20afcdfd489b8309b

resources/views/components/menu/management.blade.php
Jamiras: I'm fine with "Forum Verification".

Jamiras: Maybe both?
image

luchaos: Looks very good to me, let's go with that

Jamiras (approved):

resources/views/components/dropdown-item.blade.php
```diff
@@ -1,12 +1,12 @@
...
-    <x-link class="dropdown-item"
+    <x-link class="dropdown-item {{ $class ?? null }}"
             :active="$active ?? null"
             :external="$external ?? null"
             :link="$link">
         {{ $slot }}
     </x-link>
 @else
-    <span class="dropdown-item {{ ($active ?? false) ? 'active' : '' }}">
+    <span class="dropdown-item {{ $class ?? null }} {{ ($active ?? false) ? 'active' : '' }}">
```
wescopeland, line 9:
```suggestion
    <span class="dropdown-item [&>div]:flex [&>div]:items-center [&>div]:gap-x-2 {{ $class ?? null }} {{ ($active ?? false) ? 'active' : '' }}">
```

With this change, the classnames can be removed from the child div in _main.blade.php_, decreasing the overall document size.

wescopeland:
Lighthouse is throwing several accessibility warnings on mobile:

Screenshot 2023-08-28 at 4 26 47 PM

It is also giving an SEO warning for tap target size. I believe it throws this warning if the tap targets are less than 48x48px:

Screenshot 2023-08-28 at 4 27 59 PM

resources/views/components/dropdown-item.blade.php
luchaos: I chose to add the classes from the outside as i find it easier to read? https://github.com/RetroAchievements/RAWeb/pull/1776/commits/95e8a2f27f6091d502564968e0df8473c3a244c2

Arguably those styles could be applied to the .dropdown-item styles directly.

luchaos:
> Lighthouse is throwing several accessibility warnings on mobile:
> 
> It is also giving an SEO warning for tap target size. I believe it throws this warning if the tap targets are less than 48x48px:

Thank you for the thorough feedback! I chose to add `title` attributes instead of `aria-label` to stay consistent - makes lighthouse happy just as much.

For the tap areas (even though I do not agree with Google's assessment of the 48px rule ever since it was conceived) i removed the explicit sizing of icons and it seems to hit the mark.

resources/views/components/dropdown-item.blade.php
wescopeland: I think applying the styles to .dropdown-item directly would be a good approach. As it stands, these are adding unnecessary bloat to the document

Screenshot 2023-09-01 at 4 56 54 PM

luchaos: Agreed, good call, let's have that right away: https://github.com/RetroAchievements/RAWeb/pull/1776/commits/c87067457783405ab54993140fbb4a9d6d708cbe

wescopeland (approved):

PR #1780, opened 2023-08-20
feat(home): add user progression statistics [progression]

wescopeland:

This PR implements the new progression tracking features to the home page that have been discussed in:
https://discord.com/channels/476211979464343552/1139723030956871710/1141544886143828029
https://discord.com/channels/476211979464343552/1142257722788679780/1142257725275901963

When testing, if you see broken CSS, try restarting your Vite dev server.

---

**UI/UX**
Screenshot 2023-08-19 at 8 07 39 PM
Screenshot 2023-08-19 at 8 07 49 PM

---

**Implementation Details**
* The new homepage data is driven by new columns added to the `StaticData` table:
  * `num_hardcore_mastery_awards`
  * `num_hardcore_game_beaten_awards`
  * `last_game_hardcore_mastered_game_id`
  * `last_game_hardcore_mastered_user_id`
  * `last_game_hardcore_mastered_at`
  * `last_game_hardcore_beaten_game_id`
  * `last_game_hardcore_beaten_user_id`
  * `last_game_hardcore_beaten_at`

* A new migration has been built [here](https://github.com/RetroAchievements/RAWeb/pull/1780/files#diff-7da030617483cf369b7f7ecfb12424930fd6e759fb9fae645681eb5bde6e88bc), and accompanying updates have been included in _StaticDataFactory.php_ [here](https://github.com/RetroAchievements/RAWeb/pull/1780/files#diff-672dcfe1bc5dd16fd9f10e8496ca7427cbeed807e4ee11ba7568d65202fccf76). You will find SQL commands at the end of this PR description which should populate the table with accurate data.

* The logic for the stats component has been moved out of the template file and into a class.
* Currently, the counts for games mastered and games beaten links to recentMastery.php. There's a bit of a chicken and egg problem here - this page isn't able to filter by award type currently. We'll need to update these links in another PR.
* On a hardcore game being beaten or mastered, a call is made to new static data helper functions to update the data accordingly. If the player is untracked, the data is not updated.

---

Laravel command for updating the data:
`ra:platform:update-awards-static-data`

This can be executed with `php artisan ra:platform:update-awards-static-data`, or `sail artisan ra:platform:update-awards-static-data`.


Tsearo (commented):
Everything looks great visually 👍 The hyperlinks for mastery/beat may need to be tweaked if/when that page is updated but that can be taken care of then. Achievement Unlocks linking to recent masteries seems a bit odd but it would also be odd to be the only non-hyperlinked stat and I don't know a better place to link to.

If you beat and master a game the mastery and beat counts increment appropriately, however if you reset an achievement the mastery and beat counts do not decrease. Then re-earning the reset achievement (with progression type) both counts will then again increment. 

One thing to note is that the existing behavior for achievement unlocks in the Statistics does this same thing (does not decrement on resets) though so it may be known and intended for the unlocks. I'll leave it to someone else to decide if this is something acceptable or if we really should be properly decrementing these counts when things are reset or changed.

Not necessarily related to this specific page but through more testing I also noticed if I have a beat award for a game with a single progression achievement if I change that single achievement type to none (no longer any progression) and recalculate my score I still retain the beat award in the database which I would expect is no longer valid and gets removed. This will likely be more apparent when we have the user profile update.

wescopeland:
> The hyperlinks for mastery/beat may need to be tweaked if/when that page is updated but that can be taken care of then. Achievement Unlocks linking to recent masteries seems a bit odd but it would also be odd to be the only non-hyperlinked stat and I don't know a better place to link to.

Now that https://github.com/RetroAchievements/RAWeb/pull/1782 is opened, I've updated the hyperlinks to have the correct (new) query parameters.

> If you beat and master a game the mastery and beat counts increment appropriately, however if you reset an achievement the mastery and beat counts do not decrease. Then re-earning the reset achievement (with progression type) both counts will then again increment.

One thing to note is that the existing behavior for achievement unlocks in the Statistics does this same thing (does not decrement on resets) though so it may be known and intended for the unlocks.

I looked into building some decrement helper functions for this new static data, however I've started to believe the correct way to go is to instead write a command (similar to _UpdatePlayerMasteries.php_ and _UpdatePlayerPoints.php_) that periodically runs which "fixes" the count values to be correct. 

The trade-off with this approach is numbers will be more and more fuzzed the longer the system goes without doing a periodic update, but I think this approach should tremendously reduce the complexity required to solve this problem. We may only need to run these updates once a week, and the queries are pretty fast. If we handle this without a command, we'll need to account for all kinds of circumstances such as when a player deletes their account, when a player is marked untracked, etc. There's a lot of complexity there and, after chipping away at it for a bit, I'm not convinced it's worth going down that road.

> Not necessarily related to this specific page but through more testing I also noticed if I have a beat award for a game with a single progression achievement if I change that single achievement type to none (no longer any progression) and recalculate my score I still retain the beat award in the database which I would expect is no longer valid and gets removed.

Great catch, resolved in https://github.com/RetroAchievements/RAWeb/pull/1781.

database/migrations/2023_08_19_000000_update_staticdata_table.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
```
Jamiras, line 10: Shouldn't the "Required Prod SQL Commands" be part of the migration step?

resources/views/community/components/global-statistics/recent-game-progress.blade.php
```diff
@@ -0,0 +1,68 @@
...
+
+                <p class="absolute pl-4 top-[-2px] left-[58px] max-w-fit line-clamp-2 mb-px">
+                    {!! $renderedGameTitle !!}
+                </p>
+            </a>
+
+            <div class="-mt-1 w-full">
+                <!-- Provide invisible space to slide the console underneath -->
+                <p class="invisible max-w-fit line-clamp-2 mb-px">
+                    {!! $renderedGameTitle !!}
```
Jamiras, line 55: Should the console line have an icon?
image

app/Helpers/database/player-game.php
```diff
@@ -114,6 +114,10 @@ function testBeatenGame(int $gameId, string $user, bool $postBeaten): array
                 awardDate: Carbon::parse(calculateBeatenGameTimestamp($userAchievements)),
                 displayOrder: 0
             );
+
+            if ($isBeatenHardcore) {
+                static_addnewhardcoregamebeaten($gameId, $user);
+            }
```
Jamiras, line 120: I assume it's because not all games have their Beaten conditions defined yet, but it's weird that Beaten is so much lower than Mastered:
image

app/Helpers/database/static.php
```diff
@@ -39,6 +40,44 @@ function static_addnewregistereduser(string $user): void
...
 }
 
+function static_addnewhardcoremastery(int $gameId, string $username): void
+{
+    $foundUser = User::firstWhere('User', $username);
+    if ($foundUser->Untracked) {
+        return;
+    }
+
+    $query = <<<SQL
```
Jamiras, line 50: Using this syntax is inconsistent with the rest of the file (and largely the rest of the codebase). The only other "<<<SQL" is in another recently added function related to the Beaten feature. Does it provide some benefit over a multi-line string? It's not like it's avoiding escaping any characters because the characters that would need to be escaped are eliminated through the bound variables.

Personally, I don't like it because I search for "<<<" when trying to resolve conflicts.

database/migrations/2023_08_19_000000_update_staticdata_table.php
wescopeland: I wasn't sure whether or not writing data in that way via a migration was considered a best practice. Would you advise I move that stuff into the migration file instead?

resources/views/community/components/global-statistics/recent-game-progress.blade.php
wescopeland: I lean towards no, but happy to hear if you have a different opinion. My reasoning is for games that don't have 2 lines of content in their name, we have just enough space for the content to be flush against the game icon. If we add a console icon, the line height of the console name must increase. To maintain the "flush" effect, we'd need to compensate by increasing the game image size, which I don't think we should do.

database/migrations/2023_08_19_000000_update_staticdata_table.php
Jamiras: That's probably a question for @luchaos. I'm worried that a manual step will be forgotten.

app/Helpers/database/player-game.php
wescopeland: Your assumption is correct. A handful of devs are diligently working through defining those conditions right now, but there is still a lot of work to be done in that regard.

resources/views/community/components/global-statistics/recent-game-progress.blade.php
Jamiras: I was just aiming for consistency, but I didn't notice the sets widget doesn't have icons either. This is fine.

app/Helpers/database/static.php
wescopeland: The only benefit is improved syntax highlighting for the SQL within the string in VSCode. I'm probably in the minority when it comes to using VSCode, and there's likely more value in removing it for the reasons you've described. Refactored in https://github.com/RetroAchievements/RAWeb/pull/1780/commits/831e813c110827fc54f9e5028e1769ada8564e5f.

Jamiras:
From discord:
> 1,000 awards are granted per day give or take

That's one every 9 seconds (on average). Is it really worth showing the most recently awarded if it's going to be replaced before the player can even see it?

wescopeland:
@Jamiras Yes, I think so. My reasoning: most of the awards being granted right now are being backdated to when the player unlocked the achievement which would have been considered as beating the game. Some beaten game awards currently go back as far as 2014.

The cumulative awards granted over time can be viewed at: https://metabase.retroachievements.org/question/219

Truly new awards are happening at a much slower rate, viewable at: https://metabase.retroachievements.org/question/218

I have added some conditional logic in https://github.com/RetroAchievements/RAWeb/pull/1780/commits/7e4f79d43ccb82f6cfeabe7a92be53fd4aba2382 to ensure backdated awards don't wind up on the home page.

The backdating function can be found here: https://github.com/RetroAchievements/RAWeb/blob/master/app/Helpers/database/player-game.php#L153

database/migrations/2023_08_19_000000_update_staticdata_table.php
luchaos: I tend to go for commands that can be re-run instead, no data migrations in schema migrations.

wescopeland: Had a lot of fun learning commands, what a useful feature.

In https://github.com/RetroAchievements/RAWeb/pull/1780/commits/64713552b4861a88617f1a4a22f9e92d6c7f6350, I've edited the PR description to remove all the SQL queries, and migrated them to a new command: `ra:platform:update-awards-static-data`.

Jamiras:
**REMINDER**: run `php artisan ra:platform:update-awards-static-data` after deployment

Jamiras (approved):

luchaos:
> **REMINDER**: run `php artisan ra:platform:update-awards-static-data` after deployment

Added an [upgrade doc for 3.x -> 3.3](https://github.com/RetroAchievements/RAWeb/blob/master/docs/upgrade/upgrade-v3-to-v3.3.md) which I forgot to add before release.
Should we put this command into a 3.3 -> 3.4 doc, too?

wescopeland:
Updated to support feature flags. This PR can now be merged at will.

When the "beat" feature is not enabled, the new row of stat boxes is hidden and the most recent game beaten component is hidden:

Screenshot 2023-08-28 at 5 29 51 PM

PR #1782, opened 2023-08-20
feat(recentMastery): add beaten games and more filtering options [progression]

wescopeland:
This PR adds beaten game rows and additional filtering options on the _recentMastery.php_ page. It also renames the page in the navbar from "Recent Masteries" to "Recent Progression Awards".

At some point we may want to rename the file from _recentMastery.php_ to something else.

With the addition of the new filtering options, we can improve the URLs for the new stat boxes added in https://github.com/RetroAchievements/RAWeb/pull/1780 to filter by masteries and beaten games in hardcore mode.

The table now also uses multiline game avatars.

**Before**
Screenshot 2023-08-20 at 1 13 15 PM

**After**
Screenshot 2023-08-20 at 1 12 54 PM


public/recentMastery.php
```diff
@@ -1,67 +1,66 @@
...
+use App\Community\Enums\AwardType;
+use App\Platform\Enums\UnlockMode;
+use Illuminate\Support\Facades\Blade;
+
 authenticateFromCookie($user, $permissions, $userDetails);
 
 $maxCount = 25;
 // First day with Game Awards
-$minDate = '2014-09-29';
+$minDate = '2014-08-22';
```
wescopeland, line 11: I changed this based on what I see in the DB. But it may be worth just changing this to 2013-03-02.

resources/views/community/components/recent-awards/meta-panel.blade.php
```diff
@@ -0,0 +1,82 @@
+@props([
+    'minAllowedDate' => '2014-08-22',
+    'selectedAwardType' => null,
+    'selectedDate' => null,
+    'selectedUsers' => 'all',
+    'unlockModeSoftcore' => 0,
+    'unlockModeHardcore' => 1,
```
wescopeland, line 7: The `UnlockMode` abstract class can't be imported into this file.

app/Helpers/database/site-award.php
```diff
@@ -192,21 +192,39 @@ function SetCertifiedLegend(string $usernameIn, bool $enable): void
...
         $friendSubquery = GetFriendsSubquery($friendsOf);
         $friendCondAward = "AND saw.User IN ($friendSubquery)";
     }
 
+    $onlyAwardTypeClause = "
+        WHERE saw.AwardType IN (" . AwardType::Mastery . ", " . AwardType::GameBeaten . ")
+    ";
+    if ($onlyAwardType) {
+        $onlyAwardTypeClause = "WHERE saw.AwardType = $onlyAwardType";
+    }
```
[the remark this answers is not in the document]
wescopeland: Good idea. I think I've landed on a desirable result with some adjustments to the SQL query in https://github.com/RetroAchievements/RAWeb/pull/1782/commits/331fd0699ee872d4e5f780c1ca0d9557402e38ca.

Now, if the GameBeaten and Mastery`AwardType`s are earned by a user within a very short timespan and both are returned by the SQL query, only the Mastery award type will be in the final result (ie: a completion or mastery). This does mean if a filter is applied for beaten games, that row will still be visible in that view, but this is probably okay.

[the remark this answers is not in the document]
Jamiras: I see Road Runner is collapsed, but further down on the same page, Crash Bandicoot is not.

image 

This is indeed accurate, as the player earned one final achievement two minutes after the Beaten condition. I think this is correct, but it still looks odd.

resources/views/community/components/recent-awards/meta-type-filter.blade.php
```diff
@@ -0,0 +1,15 @@
...
+
+<div class="grid gap-y-1 sm:pl-[30px] sm:pr-8">
+    <label class="text-xs font-bold" for="filter-by-award-kind">Award kind</label>
+
+    <select id="filter-by-award-kind" class="w-full" onchange="handleAwardTypeChanged(event)">
+        <option value="all" @if (!$selectedAwardType) selected @endif>All</option>
+        <option value="beaten-softcore" @if ($selectedAwardType === 'beaten-softcore') selected @endif>Beaten (softcore)</option>
+        <option value="beaten-hardcore" @if ($selectedAwardType === 'beaten-hardcore') selected @endif>Beaten</option>
+        <option value="completed" @if ($selectedAwardType === 'completed') selected @endif>Completed</option>
+        <option value="mastered" @if ($selectedAwardType === 'mastered') selected @endif>Mastered</option>
```
Jamiras, line 13: The "Completed" filter also shows Masteries:
image

Similarly, the "Beaten (softcore)" filter also shows "Beaten" records:
image

"Mastered" and "Beaten" appear to filter correctly.

wescopeland: Thanks, resolved in https://github.com/RetroAchievements/RAWeb/pull/1782/commits/17df9774823f44857764a86fdac723b87a8957c8.

app/Helpers/database/site-award.php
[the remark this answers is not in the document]
wescopeland: I've made a slight adjustment to the query in https://github.com/RetroAchievements/RAWeb/pull/1782/commits/17df9774823f44857764a86fdac723b87a8957c8.

Maybe 1 minute was too permissive for collapsing the rows. I've expanded it to 5 minutes.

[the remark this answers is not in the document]
Jamiras: I'm not sure there's a correct value for the gap. It's probably better the way it was. If there were even one other event between the two, I probably wouldn't have even noticed.

In this case, the events _are separate_. The user beat the set, then a couple minutes later mastered it. Normally, the events would be farther apart or have other events between them.

resources/views/community/components/recent-awards/meta-type-filter.blade.php
Jamiras: Now it seems like Mastered and Beaten are missing from All.
image

image

image

app/Helpers/database/site-award.php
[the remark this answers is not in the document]
wescopeland: Reverted to a 1 minute gap.

resources/views/community/components/recent-awards/meta-type-filter.blade.php
wescopeland: Indeed. Resolved in https://github.com/RetroAchievements/RAWeb/pull/1782/commits/5fe3f5cb298a24a2b938d154ec7f2e25bf9c2feb.

Jamiras (approved):

Jamiras:
> @Jamiras Taking a 2nd look at this, do you have any reservations against me changing the page title from "Recent Progression Awards" to "Recent Game Awards"?

Nope. Sounds good.

resources/views/community/components/recent-awards/meta-type-filter.blade.php
```diff
@@ -0,0 +1,15 @@
+@props([
+    'selectedAwardType' => null,
+])
+
+<div class="grid gap-y-1 sm:pl-[30px] sm:pr-8">
+    <label class="text-xs font-bold" for="filter-by-award-kind">Award kind</label>
+
+    <select id="filter-by-award-kind" class="w-full" onchange="handleAwardTypeChanged(event)">
```
Tsearo, line 8: Changing the Award filter should likely reset the offset since it will not be uncommon if you move to a later page and then filter down to just a specific award type that there will be nothing appearing in the list without moving going back manually

wescopeland: I think this may be confusing for the user experience though. Since we can't set multiple filters simultaneously, I find myself changing the date first and then setting the award filter I want (largely because date is the first filter in the filter list).

Tsearo: Sorry to clarify a bit I'm just suggesting to reset the offset (next/prev 25). Ex 
I picked some date and see various awards (Mastered and Completed). I click next 25 and see more awards of each type for that day. Then I change the filter to Completed awards which I saw some on the first and second group of 25 users for that date and now I see none because I am still on the second page (1 offset over) so I'm confused because I know I saw some Completed awards but now none are showing.

wescopeland: Ah, sorry, I misunderstood. You're right, I will make this change.

wescopeland: Should be all set in latest.

resources/views/community/components/recent-awards/meta-type-filter.blade.php
```diff
@@ -0,0 +1,15 @@
...
+        <option value="all" @if (!$selectedAwardType) selected @endif>All</option>
+        <option value="beaten-softcore" @if ($selectedAwardType === 'beaten-softcore') selected @endif>Beaten (softcore)</option>
+        <option value="beaten-hardcore" @if ($selectedAwardType === 'beaten-hardcore') selected @endif>Beaten</option>
```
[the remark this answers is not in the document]
wescopeland: IMO it probably shouldn't matter anymore. We've met our baselines at this point to ship the feature (all top 200 games, 70%+ of all sets on the site).

resources/views/community/components/recent-awards/meta-type-filter.blade.php
Jamiras: minor: the page offset doesn't reset when switching between "all" and "followed"

wescopeland: I have no way to test a fix for this and have any confidence in my change.

It will have to wait until new DB dumps are produced.

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1782/commits/ff7198c90d19b0e863395e20382f3f9c76fce4b3.

Jamiras (approved):

PR #1783, opened 2023-08-21
fix(UploadNewAchievement): dont strip `type` value on RAIntegration edits

wescopeland:
This PR resolves an issue where achievement edits from RAIntegration are removing an achievement's `type` value.

`null` is a valid `type`, so a different fallback ("not-given") was picked.
These exceptions can be removed if and when RAIntegration natively supports the `type` field.

### DIFF
--- a/app/Helpers/database/achievement.php
+++ b/app/Helpers/database/achievement.php
@@ -194,6 +194,7 @@ function UploadNewAchievement(
         return false;
     }
 
+    // TODO: remove "not-given" when RAIntegration supports `type`
     if ($type !== null && (!AchievementType::isValid($type) && $type !== 'not-given')) {
         $errorOut = "Invalid achievement type";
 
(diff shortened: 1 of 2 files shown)


Tsearo (approved):
Verified locally that it no longer changes the type on achievement edits through the emulator.

app/Helpers/database/achievement.php
```diff
@@ -194,7 +194,8 @@ function UploadNewAchievement(
         return false;
     }
 
-    if ($type !== null && !AchievementType::isValid($type)) {
+    // TODO: remove "not-given" when RAIntegration supports `type`
```
Jamiras, line 197: I'm not sure when that will occur. Even if a new version of RAIntegration were released tomorrow, we've been allowing versions all the way back to 1.0.4 because it's the last version that supported Windows XP, and there haven't been any critical anti-cheating changes released.

wescopeland: I see. It sounds like then it may be best to always assume the potential presence of this fallback value and remove these TODO comments.

wescopeland: Comments updated in https://github.com/RetroAchievements/RAWeb/pull/1783/commits/89db13bbf40d67f0122bb455f82a5159a0f9058a.

Jamiras (approved):

PR #1789, opened 2023-08-23
prevent inserting truncated UTF-8 into motto field

Jamiras:
Reccurrance of:
```
Error encoding model [App\\Site\\Models\\User] with ID  to JSON: Malformed UTF-8 characters, possibly incorrectly encoded 
at /home/forge/retroachievements.org/releases/2023-08-14T082503-3.2.2-83eca4ee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/JsonEncodingException.php:18)
```

Which was originally reported [here](https://discord.com/channels/310192285306454017/1135568221995536554/1135568221995536554).

This sanitizes the motto going into the database so the error shouldn't occur again in the future.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -64,7 +64,9 @@ class Handler extends ExceptionHandler
             if (str_ends_with($context['url'], 'dorequest.php')) {
                 unset($params['u']);
                 unset($params['t']);
-                if ($params['r'] == 'login') {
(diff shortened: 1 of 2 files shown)


wescopeland (approved):
Verified locally the the `mb_strcut()` call is working correctly. I did not verify the other change, but I understand what it's doing and agree it's desirable.

PR #1790, opened 2023-08-23
fix: disable submit button when exceeding counter maxlength

wescopeland:
Extracts the byte count util from the character count util and reuses it in some Alpine.js directives to disable the submit button when appropriate.


Jamiras:
Verified the behavior. I can no longer submit a motto that is 50 characters long, but more than 50 bytes long.

However, there's no visual indication that the button is disabled. The only feedback is the "Changed" banner _not_ appearing when clicking on the button.

Can we turn the button text red or remove the hover highlight when it's disabled?

wescopeland:
Good idea. I've added some styles to `.btn` to better give a visual cue if it's disabled. Since red is used semantically to designate an error or danger (including on buttons that are not disabled, such as resetting progress), I went with the muted text color instead.

Screenshot 2023-08-24 at 11 03 25 PM

Jamiras (approved):

wescopeland:
Good catch, added more handling for new lines to address this in https://github.com/RetroAchievements/RAWeb/pull/1790/commits/7b4055451f49a2f7bd2d057eefa6051b0104d7d3.

Jamiras:
Pasting the text immediately reports 2004/2000. The button remains non-functional until enough characters are deleted to display 2000/2000, but the counter still displays in red at that time (while the submission is enabled and succeeds). At 1999/2000 the counter changes to blue.

Per my previous comment, I felt that having the counter red at 2000/2000 was more important than the newlines, but thanks for fixing that anyway.

wescopeland:
As of latest, the counter will no longer receive the danger color if the character count is less than or equal to the maxlength limit.

Jamiras (approved):

PR #1793, opened 2023-08-24
fix: change '- points' label to 'No achievements'

wescopeland:
Changes "- points" on similar games rows on game pages to instead show "no achievements".

Screenshot 2023-08-24 at 5 27 40 PM

### DIFF
--- a/resources/views/platform/components/game/similar-game-table-row.blade.php
+++ b/resources/views/platform/components/game/similar-game-table-row.blade.php
@@ -23,17 +23,18 @@
             <div class="flex flex-col items-end">
                 <p class="text-2xs whitespace-nowrap {{ $totalPoints === 0 ? 'text-text-muted' : '' }}">
                     <span @if ($totalPoints > 0) class="font-semibold" @endif>
-                        {{ $totalPoints > 0 ? localized_number($totalPoints) : '-' }}
+                        @if ($totalPoints > 0)
+                            {{ localized_number($totalPoints) }} points
+                        @else
+                            No achievements
+                        @endif
                     </span>
-                    points
                 </p>
 
                 <x-points-weighted-container>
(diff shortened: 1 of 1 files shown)


Jamiras:
I like the text better, but I don't like the fact that it makes that column so much wider.

Before:
image

After:
image

Would "N/A", "none", or "no points" work?

wescopeland:
That is kind of gross, isn't it? How about "n/a" (all lowercased)? Do you have a preference on "N/A" vs "n/a"?

Jamiras:
Google isn't very clear on which is preferred. Wikipedia implies lowercase is used less, but the Cambridge dictionary prioritizes the lowercase version.

When in doubt, try it out!

image

image

I prefer the capitalized version, but not strongly enough to argue about it.

wescopeland:
No strong preference from me so I'll defer to the one you prefer.

Updated in https://github.com/RetroAchievements/RAWeb/pull/1793/commits/fe52832bf5f4c5548d55331596dde52b9d83a312. I did also add a `title` attribute just in case anyone gets confused by what "N/A" means.

luchaos:
Personally, I prefer to not show anything if there's nothing to show

wescopeland:
I think if the entire list contains games with an empty state, it is non-obvious that any of them are empty without performing some kind of interaction. That would be my only reservation.

wescopeland:
No problem. I've removed the "N/A" label and set a min width on the 2nd column if the game is fully featured but has no points. If it's a hub, the 2nd column is still removed.

Mindhral:
I know this is a bit old, but is there any reason for excluding hubs lines from having the same number of columns as the other lines ?
https://github.com/RetroAchievements/RAWeb/blob/b0e5d0ef7b3d2d240e41b372b7cbce04a5feb1bd/resources/views/platform/components/game/similar-game-table-row.blade.php#L39-L41

It kind of breaks the alternating colors of rows:
image

PR #1795, opened 2023-08-25
fix: use btn class on missed buttons

wescopeland:
Adds `.btn` class names to a few missed buttons from https://github.com/RetroAchievements/RAWeb/pull/1774. Currently, these button elements appear in the UI as plain unstyled text.


luchaos (approved):
I found two other files in https://github.com/RetroAchievements/RAWeb/pull/1796 that have unstyled buttons - should those remain styled as "links"?
(leaderboardList and ticketmanager)

wescopeland:
Good finds - I didn't catch those. Your PR was more thorough than mine 😅 I've cherry-picked those changes over to here.

PR #1796, opened 2023-08-25
fix(buttons): add missing btn classes

luchaos:
Some buttons lost their styling after the last button changes.


luchaos:
Duplicate of https://github.com/RetroAchievements/RAWeb/pull/1796 - should've reloaded the PRs list first...

PR #1797, opened 2023-08-25
fix(testBeatenGame): remove activity tracking

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1144590443456237681.


luchaos (approved):

app/Community/Enums/ActivityType.php
```diff
@@ -27,8 +27,6 @@ abstract class ActivityType
 
     public const ClosedTicket = 10;
 
-    public const BeatGame = 11;
```
Jamiras, line 30: There are 70k entries in the Activity table with ActivityType=11. Do those need to be deleted, or should this enum value be kept?

app/Helpers/database/player-game.php
```diff
@@ -115,10 +115,6 @@ function testBeatenGame(int $gameId, string $user, bool $postBeaten): array
                 displayOrder: 0
             );
         }
-
-        if (!RecentlyPostedProgressionActivity($user, $gameId, $awardMode, ActivityType::BeatGame)) {
-            postActivity($user, ActivityType::BeatGame, $gameId, $awardMode);
-        }
```
Jamiras, line 121: It does seem weird that `postActivity` would update the `LastLogin` field, but it seems like it's being used as a "Last Activity" timestamp, so it's logically correct.
image

My understanding of the issue is that this is being called from some sort of background task to fill in missing information. In that case, either the activity needs to be back-dated, or just ignored.

In the normal use case, this seems like it should be expected. Maybe the `$postBeaten` check on line 105 should only apply to line 119, as the `HasSiteAward` should be sufficient to prevent doubling up on the SiteAward. Then the normal case would always pass true, and the recalculate case would pass false.

For now, removing this code seems reasonable. It can be properly addressed later.

app/Community/Enums/ActivityType.php
wescopeland: I am not sure what the correct strategy here is. If we're planning on dropping the whole table eventually anyway, it may not be a big deal to keep these around.

Jamiras: @luchaos - thoughts? Keep the enum or delete the records?

luchaos: let's keep the entries and enum for now - this way we can technically reconstruct the last login dates by looking up BeatGame activities.

wescopeland: Value restored in https://github.com/RetroAchievements/RAWeb/pull/1797/commits/dbc409eb429380cb63da081b058bf3f72762959d.

PR #1800, opened 2023-08-25
fix(buttons): add missing btn classes and button types to buttons

luchaos:


app/Helpers/render/comment.php
```diff
@@ -187,10 +187,10 @@ class="comment-textarea"
                             id="comment_textarea_$commentId"
                             oninput="autoExpandTextInput(this)"
                         ></textarea>
-                        <button class="comment-submit-button">
+                        <button class="btn comment-submit-button">
                             <img src="$submitImageUrl" alt="Submit">
                         </button>
-                        <span class="comment-loading-indicator">
+                        <span class="btn comment-loading-indicator">
```
Jamiras, line 193: `btn` style on a span?

app/Helpers/render/layout.php
```diff
@@ -210,7 +210,7 @@ function RenderToolbar(): void
         echo "<li>";
         echo "<form action='/request/auth/logout.php' method='post'>";
         echo csrf_field();
-        echo "<button class='p-0 bg-transparent text-gray-200 border-0 w-full'>Logout</button>";
+        echo "<button class='btn p-0 bg-transparent text-gray-200 border-0 w-full'>Logout</button>";
```
Jamiras, line 213: This is in the menu. Do we really want it to look like a button? (I don't actually see any difference)

wescopeland:
Per https://discord.com/channels/476211979464343552/1026595325038833725/1144748736120103054, we may need to adjust the button on manageclaims.php as well.

luchaos:
> Per https://discord.com/channels/476211979464343552/1026595325038833725/1144748736120103054, we may need to adjust the button on manageclaims.php as well.

thanks! addressed in https://github.com/RetroAchievements/RAWeb/pull/1800/commits/f9f8dd5a77db6e5896b52c1d8f5bc4de85792ea4

app/Helpers/render/comment.php
luchaos: Huh, I thought it needs that for the loading icon to keep the same padding as the submit button but it turned out that loading indicator isn't even in use anymore. Removed it in https://github.com/RetroAchievements/RAWeb/pull/1800/commits/6ffe06ffd98eae79140f4b86646f465f597ed66e

app/Helpers/render/layout.php
luchaos: I thought it would require that to make the defaults work. Turned out it'S the other way around because of the global styling from before; it does not need any of those classes anymore because it's fully styled via explicit menu button styles.
Removed in https://github.com/RetroAchievements/RAWeb/pull/1800/commits/debc7dd377d33adc83f24a9c0782c788b265a3a7

Jamiras (approved):

PR #1809, opened 2023-08-27
add In Review claim status

Jamiras:
Adds a new `In Review` claim status that can be applied to Jr. Dev claims that are ready to be reviewed. The button to toggle the `In Review` status is only available to Moderators. While a claim is in Review, it locks the claim so it cannot be dropped by the Jr. Dev.

New moderator button:
image

Changes to "Complete Claim Review" when a claim is In Review.

While In Review, the Jr Dev will see this:
image

The Claims list can also be filtered by In Review status, but is still ordered by claim date. This gives Jr Devs (and other users) a view of how many things are in the Review queue, but not their order within the queue.

image


resources/views/platform/components/game/devbox-claim-management.blade.php
```diff
@@ -158,18 +175,29 @@ function completeClaim() {
...
-        onsubmit="return dropClaim()"
-    >
-        {!! csrf_field() !!}
-        <input type="hidden" name="game" value="{{ $gameId }}">
-        <input type="hidden" name="claim_type" value="{{ $claimType }}">
-        <input type="hidden" name="set_type" value="{{ $claimSetType }}">
-        <button class="btn">Drop {{ ClaimType::toString($claimType) }} Claim</button>
-    </form>
+    @if ($primaryClaimStatus === ClaimStatus::InReview && $primaryClaimUser === $user)
+        <div class="ml-2">Cannot modify Claim in Review status</div>
```
wescopeland, line 179: I wonder if there is alternative language we should use for this. Technically, the Junior Dev can still extend the claim (and will likely have to if it remains in the queue for some time), which is a subtle modification of the claim.

resources/views/platform/components/game/devbox-claim-management.blade.php
```diff
@@ -96,6 +102,20 @@ function dropClaim() {
...
+    const gameTitle = "{!! html_entity_decode($gameTitle) !!}";
+
+    const message = 'Are you sure you want to change the claim status for ' + gameTitle + ' to In Review?';
+    return confirm(message);
+}
+
+function activateClaim() {
+    const gameTitle = "{!! html_entity_decode($gameTitle) !!}";
+
+    const message = 'Are you sure you want to change the claim status for ' + gameTitle + ' to Active?';
```
wescopeland, line 115: nit: This message is indeed technically correct, but my brain twisted into a knot when reading it. I think it's because the button says "Complete", the status is currently "In Review", and we're initiating a transition to "Active", from which the claim will be completed imminently.

I'm wondering if the user-facing message here should just be something like:
* Are you sure you want to set this claim to no longer In Review?

resources/views/platform/components/game/devbox-claim-management.blade.php
wescopeland, line 1: Should Jr Devs still be able to complete claims if they're in review?

Screenshot 2023-08-27 at 6 35 37 PM

resources/views/platform/components/game/devbox-claim-management.blade.php
Jamiras: The Extend Claim button is still available when applicable.
image

The functionality only prevents dropping the claim. Maybe it should say that? "Cannot Drop Claim while being reviewed".

resources/views/platform/components/game/devbox-claim-management.blade.php
Jamiras: I kind of struggled with the wording here too. The user is completing the review, not the claim. After the review is completed, the lock is removed and the Jr. Dev is allowed to complete the claim.

resources/views/platform/components/game/devbox-claim-management.blade.php
Jamiras: No. I'll fix that.

wescopeland: When I was a Jr, though I couldn't promote the achievements, I was still responsible for completing my claim. Not sure if that process should change, but that was what I had to do at the time.

resources/views/platform/components/game/devbox-claim-management.blade.php
wescopeland: Yes, "Cannot Drop Claim while being reviewed" or "Cannot Drop Claim while in Review status" both make more sense.

resources/views/platform/components/game/devbox-claim-management.blade.php
Jamiras: I was suggesting that promoted achievements would imply a completed review. I've explicitly hidden the button when In Review.

resources/views/platform/components/game/devbox-claim-management.blade.php
Jamiras: Went with "Cannot Drop Claim while In Review"

resources/views/platform/components/game/devbox-claim-management.blade.php
wescopeland: It's fine to stay as-is. I do think this could be a point of confusion, but the user will probably only get confused the first time they use the control, if at all.

wescopeland (approved):

PR #1811, opened 2023-08-27
feat: add support for feature flags

wescopeland:
This PR adds support for feature flags, as well as adds a new Blade directive: `@hasfeature()`.

Assuming this is approved+merged, I'll begin adding feature flag-driven conditional rendering to the progression PRs so they can be merged but not yet enabled.

### DIFF
--- a/app/Site/AppServiceProvider.php
+++ b/app/Site/AppServiceProvider.php
@@ -18,9 +18,9 @@ use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Database\Eloquent\Factories\Factory;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\ServiceProvider;
 use Illuminate\Support\Str;
-use Illuminate\Support\Facades\Blade;
 
 class AppServiceProvider extends ServiceProvider
 {
--- a/app/Site/AppServiceProvider.php
+++ b/app/Site/AppServiceProvider.php
@@ -20,6 +20,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Support\ServiceProvider;
 use Illuminate\Support\Str;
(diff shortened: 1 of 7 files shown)


resources/views/demo.blade.php
```diff
@@ -51,6 +51,19 @@
...
+            <h3>Feature Flags</h3>
+            <div class="flex justify-between">
+                <p>Beaten Games Player-facing UX</p>
+                @hasfeature("beat")
+                    Enabled
+                @else
+                    Disabled
+                @endhasfeature
+            </div>
+        </x-section>
```
wescopeland, line 65: I didn't know if this was really the best place for this or not. Maybe the admin panel is better?

app/Site/AppServiceProvider.php
```diff
@@ -55,6 +56,10 @@ public function boot(): void
             $schedule->command(DeleteOverdueUserAccounts::class)->daily();
         });
 
+        Blade::if('hasfeature', function ($feature) {
```
luchaos, line 59: Now I'm not sure, does it allow for camel case? like `hasFeature`?

resources/views/demo.blade.php
luchaos: It's nice to have it in the demo page for reference :)
For inspection by moderators the admin panel would be better, yeah.

luchaos (approved):
Nice!

app/Site/AppServiceProvider.php
wescopeland: It does allow for camel case, but I wasn't sure if that was the way to go because of existing built-ins like `@foreach`. Seems the convention for these directives is to lowercase everything.

resources/views/demo.blade.php
wescopeland: In https://github.com/RetroAchievements/RAWeb/pull/1811/commits/f01048c7e7fee83cdee66b77188a020485e082cc, I've broken out a `<x-feature-flags />` component that is being rendered on the demo page and admin.php.

app/Site/AppServiceProvider.php
luchaos: ah, yes, that makes sense.

luchaos (approved):

PR #1814, opened 2023-09-01
Update queries to use aggregated data

Jamiras:
New queries are enabled by setting the `FEATURE_AGGREGATE_QUERIES` .env flag to true.

Data populated for testing was done by hand by modifying the code to run the old queries and update the new tables with the resulting data. No new rows were added, but all of the existing rows had uninitialized fields. I did not keep the related code - I apologize to anyone else who wants to test this before luchaos finishes his PR.


app/Helpers/database/game.php
```diff
@@ -85,88 +84,103 @@ function getGameMetadata(
...
-            $bindings = [
-                'gameId' => $gameID,
-                'achievementFlag' => $flag,
-            ];
-
-            $gameIdStatement = 'ach.GameID = :gameId';
-            if ($parentGameId !== null) {
+            if ($parentGameId === null) {
+                $query = "SELECT players_total AS NumDistinctPlayers FROM GameData WHERE ID=$gameID";
+                $gameMetrics = legacyDbFetch($query);
```
Jamiras, line 92: **game/1446**
With feature: 0ms
Without feature: 743ms (results are cached for one hour at a time)

NOTE: due to the data available in the aggregate columns, only `NumDistinctPlayers` is available. Previously, this function returned both `NumDistinctPlayersCasual` and `NumDistinctPlayersHardcore`. The game page only used `NumDistinctPlayersCasual`, which reflects both hardcore and non-hardcore players. There are two APIs that also return the data. I'm not sure how that's being used, but I've made the decision to deprecate the fields and return `NumDistinctPlayers` instead. The deprecated fields return the same value as `NumDistinctPlayers`.

app/Helpers/database/game.php
```diff
@@ -85,88 +84,103 @@ function getGameMetadata(
...
-                SELECT ach.ID AS AchievementID,
-                    (COUNT(aw.AchievementID) - SUM(IFNULL(aw.HardcoreMode, 0))) AS NumAwarded,
-                    (SUM(IFNULL(aw.HardcoreMode, 0))) AS NumAwardedHardcore
-                FROM Achievements AS ach
-                LEFT JOIN Awarded AS aw ON aw.AchievementID = ach.ID
-                WHERE ach.GameID = :metricsGameId AND ach.Flags = :metricsAchievementFlag
-                GROUP BY ach.ID
-            ) AS tracked_aw ON tracked_aw.AchievementID = ach.ID";
+        if (env('FEATURE_AGGREGATE_QUERIES')) {
+            $metricsColumns = 'ach.unlocks_total AS NumAwarded, ach.unlocks_hardcore_total AS NumAwardedHardcore,';
```
Jamiras, line 151: **game/1446**
With feature: 0ms
Without feature: 90ms

app/Helpers/database/player-achievement.php
```diff
@@ -183,8 +183,13 @@ function unlockAchievement(string $username, int $achievementId, bool $isHardcor
 
 function getAchievementUnlockCount(int $achID): int
 {
-    $query = "SELECT COUNT(*) AS NumEarned FROM Awarded
-              WHERE AchievementID=$achID AND HardcoreMode=0";
+    if (env('FEATURE_AGGREGATE_QUERIES')) {
+        $query = "SELECT COUNT(*) AS NumEarned FROM player_achievements
+                  WHERE achievement_id=$achID";
```
Jamiras, line 188: This function is only called when promoting/demoting achievements. I couldn't get an accurate reading on the timing, but running the query from within mysql:

**achievement/3159**
With feature: 4ms
Without feature: 6ms

app/Helpers/database/player-achievement.php
```diff
@@ -485,22 +492,52 @@ function getAchievementDistribution(
...
+        if ($flag === AchievementFlag::OfficialCore) {
+            $countColumn = $hardcore ? 'achievements_unlocked_hardcore' : 'achievements_unlocked';
+            $query = "SELECT pg.$countColumn AS AwardedCount, COUNT(*) AS NumUniquePlayers
+                      FROM player_games AS pg
+                      $joinStatementNew
+                      WHERE pg.game_id = :gameId AND pg.$countColumn > 0
+                      $requestedByStatement
+                      GROUP BY AwardedCount
+                      ORDER BY AwardedCount DESC";
+            unset($bindings['achievementFlag']);
```
Jamiras, line 505: **game/1446**
With feature: 121ms
Without feature: 115ms

app/Helpers/database/player-achievement.php
```diff
@@ -485,22 +492,52 @@ function getAchievementDistribution(
...
+                        LEFT JOIN Achievements AS ach ON ach.ID = pa.achievement_id
+                        $joinStatementNew
+                        WHERE ach.GameID = :gameId
+                        $hardcoreStatement
+                        AND ach.Flags = :achievementFlag
+                        $requestedByStatement
+                        GROUP BY pa.user_id
+                        ORDER BY AwardedCount DESC
+                    ) AS InnerTable
+                    GROUP BY InnerTable.AwardedCount";
```
Jamiras, line 521: **game/1446**
With feature: 310ms
Without feature: 91ms

The new tables don't provide any aggregate data for unofficial achievements. Apparently, calculating that from the `player_achievements` table is slower than doing so from the `Awarded` table.

app/Helpers/database/player-game.php
```diff
@@ -674,27 +674,43 @@ function getUsersCompletedGamesAndMax(string $user): array
...
+                pg.achievements_unlocked AS NumAwarded, pg.achievements_unlocked_hardcore AS NumAwardedHC, " .
+                floatDivisionStatement('pg.achievements_unlocked', 'gd.achievements_published') . " AS PctWon, " .
+                floatDivisionStatement('pg.achievements_unlocked_hardcore', 'gd.achievements_published') . " AS PctWonHC
+            FROM player_games AS pg
+            LEFT JOIN GameData AS gd ON gd.ID = pg.game_id
+            LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
+            LEFT JOIN UserAccounts ua ON ua.ID = pg.user_id
+            WHERE ua.User = '$user'
+            AND gd.achievements_published > $minAchievementsForCompletion
+            ORDER BY PctWon DESC, PctWonHC DESC, MaxPossible DESC, gd.Title";
```
Jamiras, line 690: **user/MaxMilyin**
With feature: 11ms
Without feature: 545ms

app/Helpers/database/user-activity.php
```diff
@@ -393,31 +393,23 @@ function getRecentlyPlayedGames(string $user, int $offset, int $count, ?array &$
...
-    // 08:05 01/10/2014 removed outer activitytype=3, added rating
-    // {$RatingType::Game}
-
+    if (env('FEATURE_AGGREGATE_QUERIES')) {
+        $query = "SELECT pg.last_played_at AS LastPlayed, pg.game_id AS GameID
+                  FROM player_games pg
+                  LEFT JOIN UserAccounts ua ON ua.ID = pg.user_id
+                  WHERE ua.User = :username
+                  ORDER BY pg.last_played_at desc
+                  LIMIT $offset, $count";
```
Jamiras, line 402: **user/MaxMilyin**
With feature: 10ms
Without feature: 3070ms

app/Helpers/database/game.php
luchaos: If I understand correctly I will add `players_total_hardcore` to the games table. This was an oversight on my part.

app/Helpers/database/game.php
```diff
@@ -85,88 +84,103 @@ function getGameMetadata(
     $metricsJoin = '';
     $metricsBindings = [];
     if ($metrics) {
-        $cacheKey = "game:$gameID:playercount";
-        $gameMetrics = Cache::get($cacheKey);
-        if ($gameMetrics === null) {
+        if (env('FEATURE_AGGREGATE_QUERIES')) {
```
luchaos, line 87: Please move those to `config/feature.php` and access it via `config('feature.aggregate_query')`.
Configs are cached and thus code loses direct access to env variables at runtime.

app/Helpers/database/game.php
luchaos: Makes sense. It might be interesting to know for reporting? To see how sets fare when it comes to opting for softcore, might be an indicator of a set's difficulty in the future?

app/Helpers/database/player-achievement.php
luchaos: Might be the missing index on `unlocked_hardcore_at`, i'll add that.

app/Helpers/database/user-activity.php
luchaos: Wonderful

app/Helpers/database/player-game.php
luchaos: Wonderful

public/API/API_GetGameExtended.php
```diff
@@ -9,8 +9,9 @@
  *  int        ConsoleID                  unique identifier of the console associated to the game
  *  string     ConsoleName                name of the console associated to the game
  *  int?       ParentGameID               unique identifier of the parent game if this is a subset
- *  int        NumDistinctPlayersCasual   number of unique players who have earned achievements for the game
- *  int        NumDistinctPlayersHardcore number of unique players who have earned achievements for the game in hardcore
+ *  int        NumDistinctPlayers         number of unique players who have earned achievements for the game
+ *  int        NumDistinctPlayersCasual   [deprecated] equal to NumDistinctPlayers
+ *  int        NumDistinctPlayersHardcore [deprecated] equal to NumDistinctPlayers
```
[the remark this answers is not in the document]
luchaos: Do we want to keep the distinction intact?
I'd like to re-introduce `players_total_hardcore` for reporting reasons mentioned above.

app/Helpers/database/game.php
luchaos: I'm absolutely fine with deprecating the distinction btw.

luchaos (approved):
Thank you very much for taking care of digging through those queries 🙏 That helps a lot!

I guess the best way to move forward to have those changes all in place behind the feature flag? This way i can rebase the following double-write PR on it and see where indexes are still missing (as was identified for at least one case)

luchaos:
In case you want to include and validate the missing index right away:
```php
// database/migrations/platform/2023_09_01_000000_update_player_achievements_table.php
<?php

declare(strict_types=1);

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class() extends Migration {
    public function up(): void
    {
        Schema::table('player_achievements', function (Blueprint $table) {
            $table->index('unlocked_hardcore_at');
        });
    }

    public function down(): void
    {
        Schema::table('player_achievements', function (Blueprint $table) {
            $table->dropIndex('unlocked_hardcore_at');
        });
    }
};
```

Jamiras:
Additional timings for the three disabled APIs:

**API_GetGameInfoAndUserProgress?g=1446&u=Jamiras**
With feature: 210ms
Without feature: 1813ms

**API_GetUserRecentlyPlayedGames?u=Jamiras**
With feature: 373ms
Without feature: 2339ms

**API_GetUserSummary?u=Maxmilyin**
With feature: 1962ms
Without feature: 3423ms

app/Helpers/render/leaderboard.php
```diff
@@ -538,21 +538,42 @@ function getGlobalRankingData(
...
+                FROM player_achievements AS aw
+                LEFT JOIN Achievements AS ach ON ach.ID = aw.achievement_id
+                LEFT JOIN UserAccounts AS ua ON ua.ID = aw.user_id
+                WHERE TRUE $whereDateAchievement $typeCond
+                $friendCondAchievement
+                $singleUserAchievementCond
+                $untrackedCond
+                GROUP BY ua.User
+                $orderCond
+                LIMIT $offset, $count";
```
Jamiras, line 559: **Daily friends leaderboard widget for user following the 10 top ranked users, and about a dozen other users in the top 50.**
With feature: 136ms
Without feature: 313ms

app/Helpers/database/player-achievement.php
luchaos: Do you see a need for improvement here, like an index that might improve group by performance? (Not sure if that'd work actually)
Or are we happy with the results so far?

Jamiras: I'm not sure that an additional index would help. The query boils down to:
```
SELECT pg.achievements_unlocked AS AwardedCount, COUNT(*) AS NumUniquePlayers
FROM player_games AS pg
WHERE pg.game_id = 1446 AND pg.achievements_unlocked > 0
GROUP BY AwardedCount
ORDER BY AwardedCount DESC
```

And there's already an index on game_id.

It's just another aggregation issue. The time is most likely spent tallying the 30k records into 150 buckets. The query is repeated again for softcore, and the results of both queries are used to generate the achievement distribution graph. I think the best solution is to move that to a secondary page for the game.

luchaos: Sounds good - let's have that as soon as we work on those subpages.

PR #1820, opened 2023-09-02
feat(docker): use mysql image with pv installed

luchaos:
Makes it nicer to import large dumps from within the container by showing progress and time estimate.

Run `sail up (-d)` to build, existing volume is kept intact so no data is lost.

Use pv:
```sh
pv ra-web.sql | mysql -u retroachievements -pretroachievements retroachievements-web
2.32GiB 2:12:03 [ 307KiB/s] [======================================================>] 100%
```


docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '3.7'
```
luchaos, line 1: See https://github.com/laravel/sail/pull/601

wescopeland (approved):

PR #1821, opened 2023-09-03
feat(layout): enable wide layout

luchaos:
Wide layout.

Eliminates fullcontainer, leftcontainer, rightcontainer.


wescopeland:
Going to attack this 1st thing in the morning. I'll do a sweep of the site to clean up any visual discrepancies.

luchaos:
The constraint used to be at `lg` with `1024px` as defined in `tailwind.config.js`.

Removing the constraint completely defaults to the `2xl` width which is currently defined with `1536px`.

Arguably we could go for the `xl` breakpoint's `1280px` if we think we don't need that extra width:
```css
.container {
  max-width: theme('screens.xl');
}
```

wescopeland:
I think we should cap at XL and not worry about 2XL.
2XL is probably too many breakpoints to manage without a UI team 😅

resources/css/legacy.css
```diff
@@ -1,63 +1,3 @@
-/* TODO get rid of mainpage, fullcontainer, leftcontainer, rightcontainer */
```
luchaos, line 1: Finally, this TODO finds peace.

resources/css/layout.css
```diff
@@ -37,16 +37,49 @@ body {
...
   overflow-x: hidden;
+  padding: 0;
+  position: relative;
+  min-height: 450px;
+  display: grid;
+  column-gap: 1rem;
+}
+
+main.with-sidebar {
+  grid-template-columns: 1.329fr 0.671fr;
```
wescopeland, line 54: Before I continue with the review, we should make a decision regarding this line. This is my proposed change, but I could be convinced it's not the right approach. I'm interested in your thoughts:

```suggestion
  grid-template-columns: 1.329fr 0.489fr;
```

Currently, 0.671fr adds about 100px of width to the sidebar.
0.489fr reduces this back down to 340px.

Arguments for 0.489fr:
* 340px has been fairly constrained for the sidebar in the past - maybe this is a good thing.
* We'll only need to optimize for MD and less on the sidebar.
* On XL, there won't be an "open invitation" to put primary content in the sidebar because of the extra new space.
* The sidebar on desktop will still flush neatly against the navbar and user card (though, the user card is going away before long).

Arguments for 0.671fr:
* 340px has been fairly constrained for the sidebar in the past - maybe this is _not_ a good thing.
* The main content area gets less space at XL, so it's probably less we need to fix.

I lean towards sticking with 0.489fr, but my confidence in this choice is not high and this will have design ramifications across the entire site for some time (hence why we should make a decision before moving on in this review).

IMO the 2nd column (sidebar) is much too at 0.671fr. This is harder to tell on dark mode, but easier to see on light mode:

Screenshot 2023-09-03 at 9 56 03 AM

Setting this to 0.489fr gets us back to a crisp 340px width.

Screenshot 2023-09-03 at 9 57 56 AM

Let's land on a conclusion to this important decision first and then continue with the review.

wescopeland:
I noticed that something is wrong with the subheadings in the game page main content area. "Achievements" should not have an underline/blue divider.

Screenshot 2023-09-03 at 10 10 01 AM

resources/css/layout.css
luchaos: Thanks for bringing this up! I had very similar thoughts when browsing through the pages.

I'd agree and say let's stick with the 340px sidebar for now and concentrate on the main content area instead? You have some arguments for sticking with that for now.

https://github.com/RetroAchievements/RAWeb/pull/1821/commits/a0b455acfe6d88f6b069333ea3442002efc0a526

luchaos:
> I noticed that something is wrong with the subheadings in the game page main content area. "Achievements" should not have an underline/blue divider.

Fixed. Tbh I don't like those underlines being attached to headlines much. With the introduction of sections and section headers (another PR) this will be moved away from the heading elements.

resources/css/layout.css
luchaos: We could give it (and the too card) 20-50px just let it breathe a bit.

I didn't find any major issue with giving everything more space in general. A lot has been prepared for this in the past so it doesn't completely fall apart when making it wider.

wescopeland: Agreed, I just tested a few different numbers and 380px seems "just right".

luchaos: Perfect, let's go with that 👍

luchaos: Changed in https://github.com/RetroAchievements/RAWeb/pull/1821/commits/c6551d505fa25e5575f8768fec2b9072ec61ab7d

public/gameList.php
wescopeland, line 1: I right-aligned a bunch of numbers in the table, added the RetroPoints wrapper, and removed the "Last Updated" column (which was misleading and confusing).

Screenshot 2023-09-03 at 8 59 50 PM

http://localhost:64000/gameList.php?c=4

public/globalRanking.php
wescopeland, line 1: Right-aligned columns, localized numbers, and RetroPoints wrappers.

Screenshot 2023-09-03 at 9 01 46 PM

http://localhost:64000/globalRanking.php

public/history.php
wescopeland, line 1: Updated some margins and fixed the table on the page.

Screenshot 2023-09-03 at 9 02 48 PM

public/setRequestList.php
wescopeland, line 1: Moved the "Claimed By" column a little further to the right on XL so it wasn't sitting near the center of the page.

Screenshot 2023-09-03 at 9 03 27 PM

http://localhost:64000/setRequestList.php

public/ticketmanager.php
wescopeland, line 1: This page looks way better at XL.

Screenshot 2023-09-03 at 9 04 13 PM

http://localhost:64000/ticketmanager.php

resources/css/button.css
```diff
@@ -38,7 +38,8 @@
   background: none;
 }
 .btn-link:hover {
-  border: none;
+  background: none !important;
+  border: none !important;
```
wescopeland, line 42: `!important` is kind of lame, but this solves the bug with link buttons causing a layout shift on hover.

resources/css/layout.css
```diff
@@ -51,7 +51,7 @@ main {
 }
 
 main.with-sidebar {
-  grid-template-columns: 1fr 380px;
+  @apply grid-cols-[1fr,340px] xl:grid-cols-[1fr,380px];
```
wescopeland, line 54: Keep LG at the same sidebar width.

resources/css/utilities.css
```diff
@@ -1,3 +1,9 @@
+@layer components {
+  .container {
+    @apply sm:max-w-[1024px] xl:max-w-[1280px];
+  }
+}
```
wescopeland, line 5: This fixes a bug where SM and MD no longer stretched the width of the page.

resources/views/community/components/news/carousel-2.blade.php
wescopeland, line 1: The news carousel only gets a few more pixels of width at XL. It's now in an embed to help mask how much space we're missing in the component. I don't think we should add more width, otherwise news posts may get too big and become unreadable on other device sizes.

Screenshot 2023-09-03 at 9 06 02 PM

resources/views/layouts/components/body-end.blade.php
wescopeland, line 1: Restyles the debug element:

Screenshot 2023-09-03 at 9 07 39 PM

resources/views/layouts/components/brand-top.blade.php
wescopeland, line 1: With our current logo, this is probably as good as it's going to get on XL. If the logo remains centered, it does seem weird things to the visual continuity of the page.

Screenshot 2023-09-03 at 9 08 21 PM

resources/views/platform/components/game/screenshots.blade.php
```diff
@@ -39,7 +39,12 @@ function carousel() {
 </div>
 
 <!-- SM+ -->
-<div class="hidden sm:flex mb-3 mx-0 justify-around w-full gap-y-1 gap-x-5">
+<div class="
+    hidden sm:flex justify-around items-center mb-3 px-6 pt-2 pb-2.5 w-full
+    bg-embed border border-embed-highlight gap-y-1 gap-x-5
+    xl:px-4 xl:pt-4 xl:pb-5 xl:rounded-lg xl:mx-0 xl:w-full xl:min-h-[180px]
+"
+>
```
wescopeland, line 47: Screenshot 2023-09-03 at 9 14 29 PM

This lets us still visually have a "straight line" down, like:

Screenshot 2023-09-03 at 9 11 14 PM

Otherwise, the eye feels a lot of pressure and undue focus towards the strange whitespace around the screenshots.

wescopeland (commented):
I went ahead and figured it would be best just to commit directly to the branch.

I've verified every page across the entire app and made changes on the XL breakpoint where appropriate. Some pages were kinda weird and took a bit of tweaking, but we are all set now.

This is a big QoL improvement!

app/Helpers/render/site-award.php
```diff
@@ -140,7 +140,8 @@ function RenderAwardGroup(array $awards, string $title, string $awardsOwnerUsern
 
     echo "<div id='" . strtolower(str_replace(' ', '', $title)) . "'>";
     echo "<h3 class='flex justify-between gap-2'><span class='grow'>$title</span>$counters</h3>";
-    echo "<div class='component flex flex-wrap justify-start gap-2'>";
+    echo "<div class='flex justify-center'>";
+    echo "<div class='component flex flex-wrap gap-2 justify-start bg-embed w-full xl:rounded xl:py-2 xl:px-5'>";
```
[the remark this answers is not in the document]
luchaos: We could easily fit 6 per row 🤔 This would reduce the overall height occupied of awards in their current state. Just an idea.

Screenshot 2023-09-04 at 19 38 29

app/Helpers/render/user.php
```diff
@@ -145,11 +145,20 @@ function RenderCompletedGamesList(
 
         echo "<tr class='$isCompletedClassName'>";
 
-        echo "<td>";
-        echo gameAvatar($userCompletedGamesList[$i], label: false, context: $username);
-        echo "</td>";
-        echo "<td class='smaller'>";
-        echo gameAvatar($userCompletedGamesList[$i], icon: false, context: $username);
+        echo "<td class='py-2'>";
+        echo Blade::render('
```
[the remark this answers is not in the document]
luchaos: Nice!

app/Helpers/render/site-award.php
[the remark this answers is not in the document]
wescopeland: Considered this, but I think users would be pretty upset with the change. Based on everything learned from building drag & drop site awards, users are primarily ordering their awards for desktop with 5 badges per row (even though mobile is variable on # shown per row).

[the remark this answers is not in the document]
luchaos: Alright. Let's leave it at that for now.

On a side note: the question is which users - we'll get a lot of strong feedback from some on making the whole site wider, too. 
I'd like to modernize this site eventually ;)

resources/views/layouts/components/brand-top.blade.php
luchaos: lgtm! I have a general suggestion for the logo/brand area - that's a different PR though

resources/views/platform/components/game/screenshots.blade.php
luchaos: I like the use of "embedded" areas overall!

app/Helpers/render/site-award.php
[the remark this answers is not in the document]
Jamiras: > the question is which users

I'm one of them, and that's my profile that you've used to illustrate 6 per row.

Note how the Lolo achievements at the end of the list are in the middle of a row (they start a row on my live profile page). And less obviously, the Space Quest icon starts a block of adventure icons leading up to the Lolo achievements:

image

Then I've got five mickey mouse icons, then other cartoon-related stuff.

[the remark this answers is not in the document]
Jamiras: I wouldn't say I'd be upset by the change, but my badges are currently ordered under the assumption of 5 per row.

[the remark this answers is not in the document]
luchaos: fair enough :) it does get very dense with 6 per row, too, so i don't like it too much either after looking at it for longer.
that came off a bit strong i guess, sorry for that.

resources/css/button.css
luchaos: I can't seem to replicate the behaviour where those buttons cause layout shifts without `!important`. Is this happening on a specific page?

wescopeland: Yep, check out friends.php

Screenshot 2023-09-04 at 2 24 51 PM

These three buttons shift the layout around them when hovered

resources/css/utilities.css
luchaos: Interesting! I missed that it was overwriting that all that time oO
What do you think of using the tailwind theme helper here?
To have those breakpoints defined only in one place:
```css
@apply sm:max-w-[theme('screens.lg')] xl:max-w-[theme('screens.xl')];
```

wescopeland: That is better- I should've reached for this. I keep forgetting `theme()` is a thing.

wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1821/commits/829e3e1b556b1fbf8e8e6d55e2932d703e30a65c.

resources/css/button.css
luchaos: oh, i understand. it's because of the higher specifity of `:enabled`.
adding the border just for , even for `.btn-link` 🤔

Can we remove the `:enabled` selector as it's only applied to `button` elements, override hover for disabled instead? This way it doesn't interfere with the more generic `.btn-link` styles.

Something like this:
```css
.btn:hover {
  background-color: var(--embed-highlight-color);
  color: var(--link-hover-color);
  border: 1px solid var(--menu-link-color);
}
.btn:disabled, .btn:disabled:hover {
  color: var(--text-color-muted);
  border: 1px solid var(--embed-color);
  cursor: not-allowed;
}

.btn-link {
  border: none;
  background: none;
}
.btn-link:hover {
  border: none;
  background: none;
}
```

wescopeland: Same issue with this change on my local (even after restarting Vite). Going to play with it for just a minute...

wescopeland: Think I got it:

```css
.btn-link {
  border: 1px solid transparent;
  background: none;
}
.btn-link:hover {
  border: none;
}
```

luchaos: with that it starts jumping for me 😅

wescopeland: Latest changes in https://github.com/RetroAchievements/RAWeb/pull/1821/commits/0e42a8bae7d3e70dbad56ef3eb183e45c8c80151

luchaos: Looking good, thanks!

wescopeland (approved):

PR #1822, opened 2023-09-03
feat(inspector): allow Jr Devs to mark progression in bulk

drisc:
Manually marking progression sucks so now Jrs can do it in bulk on sets where they are the sole developer

### DIFF
--- a/public/achievementinspector.php
+++ b/public/achievementinspector.php
@@ -175,9 +175,6 @@ function toggleAllCodeRows() {
                     to see the new order on this page.
                 </p>
             HTML;
-        }
-
-        if ($partialModifyOK || $fullModifyOK) {
             echo <<<HTML
                 <p align="justify">
                     You can mark multiple achievements as '$progressionLabel' or '$winConditionLabel'. To do this, check the desired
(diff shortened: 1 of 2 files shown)


public/achievementinspector.php
wescopeland, line 1: While signed in as a Jr Dev and viewing the inspector for Core achievements, the behavior of the page should not change. The extra buttons, additional instructions description copy, and row checkboxes should not be available while inspecting Core achievements as a Jr Dev.

**Master**
Screenshot 2023-09-03 at 9 38 08 AM

**This PR**
Screenshot 2023-09-03 at 9 37 48 AM

wescopeland, line 1: While modifying Unofficial, the guard against changing types for non-self authored achievements works. However, it is unclear which of the achievements in this view I'm the author of. This is not uncommon if one Jr starts working on a set, gives up on the program, and another Jr takes up the set.

I believe the best UX in this case will be to hide the checkboxes on rows where I am not the author of that achievement.

Screenshot 2023-09-03 at 9 41 50 AM

In the screenshot above, my Jr Dev user did not author one of those achievements. I should only see one checkbox rather than two.

drisc: Fixed this in the latest commit

The checkbox no longer appears on achievements in unofficial that the user has not authored. The unlock-flag request also blocks manually crafted requests if the user is not the author.

image

wescopeland, line 1: As an admin, I can no longer see checkboxes for achievements I didn't author.

Screenshot 2023-09-03 at 2 40 10 PM

drisc: Added a check for permissions level, set to Developer and higher.

public/achievementinspector.php
```diff
@@ -177,7 +177,7 @@ function toggleAllCodeRows() {
             HTML;
         }
 
-        if ($fullModifyOK) {
+        if ($partialModifyOK || $fullModifyOK) {
```
wescopeland, line 180: I can't help but notice there is an identical conditional for another HTML block just above this.

Let's consolidate those conditionals into one:

```php
if ($partialModifyOK || $fullModifyOK) {
    echo <<<HTML
        <p align="justify">...</p>

        <p align="justify">...</p>
    HTML;
}
```

public/achievementinspector.php
```diff
@@ -325,7 +331,7 @@ function toggleAllCodeRows() {
                 echo "<a class='btn w-full flex justify-center py-2' href='/achievementinspector.php?g=$gameID&f=5'>Unofficial Achievement Inspector</a>";
             }
 
-            if ($fullModifyOK) {
+            if ($fullModifyOK || $partialModifyOK) {
```
wescopeland, line 334: This is already within a `$fullModifyOK || $partialModifyOK` conditional. Can this conditional be removed?

public/achievementinspector.php
drisc: Instructions combined and view conditional consolidated

public/achievementinspector.php
drisc: Yes, these have now been removed, functionality remains the same

wescopeland (approved):

PR #1823, opened 2023-09-03
attempt to address eslint warnings

Jamiras:
To eliminate extraneous warnings appearing from "Unchanged files with check annotations (beta)" feature that are appearing in every review.

### DIFF
--- a/public/js/all.js
+++ b/public/js/all.js
@@ -1,7 +1,3 @@
-function asset(uri) {
-  return window.assetUrl + '/' + uri.replace(/^\/|\/$/g, '');
-}
-
 function mediaAsset(uri) {
   return window.mediaAssetUrl + '/' + uri.replace(/^\/|\/$/g, '');
 }
@@ -44,40 +40,6 @@ $(document).ajaxSuccess(function (event, xhr) {
   }
 });
 
-var shortMonths = [
-  'Jan',
-  'Feb',
-  'Mar',
-  'Apr',
-  'May',
-  'Jun',
-  'Jul',
-  'Aug',
-  'Sep',
-  'Oct',
-  'Nov',
-  'Dec'];
-
-function htmlEntities(str) {
-  return String(str)
-    .replace(/&/g, '&amp;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;');
-}
-
-/**
- * Pads a number with 0s
- */
-function strPad(input, padLength, padString) {
-  input += '';
-  if (input.length >= width) {
-    return input;
-  }
-  padString = padString || '0';
-  return new Array(padLength - input.length + 1).join(padString) + input;
-}
-
 function getParameterByName(name) {
   name = name.replace(/[\[]/, '\\\[').replace(/[\]]/, '\\\]');
   var regexS = '[\\?&]' + name + '=([^&#]*)';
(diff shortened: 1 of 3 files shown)


public/js/all.js
```diff
@@ -1,7 +1,3 @@
-function asset(uri) {
```
Jamiras, line 1: This is the only one I'm concerned with. I think it's only used in loadDynamicTooltip.ts, but after removing it and rebuilding everything, the dynamic tooltip still loads, so maybe it's using the PHP version of the function?

Jamiras:
hrmph. now it's reporting a different set of errors in a bunch of .ts files.

wescopeland:
Will help out here in just a sec

wescopeland:
In _.eslintrc.yaml_, add this to the bottom of the file:

```yml
overrides:
  - files:
      - '*.test.ts'
    rules:
      '@typescript-eslint/no-explicit-any': 'off'
```

In tests it is safe to not enforce this rule.

wescopeland:
In _lazyLoadModuleOnIdFound.ts_:

```ts
(window as unknown as Record<string, unknown>)[moduleNameToAttachToWindow] = lazyLoadedModule;
```

wescopeland:
In _global.d.ts_, two lines should be added (without the comments to point them out):

```ts
declare global {
  var Alpine: Alpine;
  var assetUrl: string; // <--------------------- added this line
  var autoExpandTextInput: typeof AutoExpandTextInput;
  var cfg: Record<string, unknown> | undefined;
  var copyToClipboard: (text: string) => void;
  var getStringByteCount: typeof GetStringByteCount;
  var handleLeaderboardTabClick: typeof HandleLeaderboardTabClick;
  var hideEarnedCheckboxComponent: typeof HideEarnedCheckboxComponent;
  var initializeTextareaCounter: typeof InitializeTextareaCounter;
  var injectShortcode: typeof InjectShortcode;
  var loadPostPreview: typeof LoadPostPreview;
  var newsCarouselComponent: typeof NewsCarouselComponent;
  var showStatusFailure: (message: string) => void; // <--------------------- added this line
  var showStatusSuccess: (message: string) => void;
  var toggleUserCompletedSetsVisibility: typeof ToggleUserCompletedSetsVisibility;
  var tooltipComponent: typeof TooltipComponent;
}
```

public/js/all.js
wescopeland: This should be okay. It's using `asset()` from _helpers.ts_:

```
// loadDynamicTooltip.ts
import { asset, fetcher } from '@/utils';

// resources/js/utils/index.ts
export * from './helpers';

// helpers.ts
export const asset = (uri: string) => { ... };
```

resources/js/utils/helpers.ts
```diff
@@ -1,6 +1,6 @@
 export const config = (key: string) => window?.cfg?.[key];
 
 export const asset = (uri: string) => {
-  const assetsUrl = config('assetsUrl') ?? (window as any).assetUrl;
+  const assetsUrl = config('assetsUrl') ?? (window as unknown as Record<string, unknown>).assetUrl;
```
wescopeland, line 4: Sorry for the confusion.

By defining `assetUrl` in _global.d.ts_ as belonging to the global (window) module, we've communicated to TypeScript that `assetUrl` is type-safe and valid.

Given that, you should be able to do:

```ts
const assetsUrl = config('assetsUrl') ?? window.assetUrl;
```

TypeScript will recognize that `window.assetUrl` is expected to be a `string` value rather than being confused and thinking we've made a typo.

wescopeland (approved):

PR #1826, opened 2023-09-04
Remove pixelart icons, use fontawesome and google-material-design

luchaos:
The pixelart icons are not as legible as desired and do not scale well. 
I suggest we use the fontawesome solid set for now.


wescopeland (approved):
Verified locally, the new icons are looking great.

The info icon isn't resonating with me (due to the circle being filled), but it is better than what we currently have now. I have an SVG asset that may be a better drop-in replacement, but it is not enough to hold this PR up over.

luchaos:
> Verified locally, the new icons are looking great.
> 
> The info icon isn't resonating with me (due to the circle being filled), but it is better than what we currently have now. I have an SVG asset that may be a better drop-in replacement, but it is not enough to hold this PR up over.

As discussed: added google material design icons and replaced the solid fontawesome info icon in https://github.com/RetroAchievements/RAWeb/pull/1826/commits/1255fa8d83eef6e12d750b1ac6d3f271597c2504

wescopeland (approved):
Reverified. Very happy with this choice, it is perfect 👍

Screenshot 2023-09-04 at 3 37 34 PM

PR #1831, opened 2023-09-05
feat: add auth, notifications, and messages to navbar

luchaos:
This PR is inspired by the prototype, the final one addressing the navbar whilst adding new elements and features that have been replaced in the top area:

**Remove top-card, add brand top with square logo, text and claim**

_Dark_
Screenshot 2023-09-05 at 17 27 46
_Light_
Screenshot 2023-09-05 at 17 28 36

**Add bell icon to nav with badge for claim expiry & tickets notifications, add envelope icon to nav with badge for messages**

Screenshot 2023-09-05 at 15 59 21

_Claim expiry & ticket notifications moved to the dropdown_
Screenshot 2023-09-05 at 17 58 47

_Small screens (375px+)_
Screenshot 2023-09-05 at 16 52 35

**Add "Sign in" & "Sign up" to navbar**

Screenshot 2023-09-05 at 18 01 09

**(Re)enable fortify as auth handler**:
- dedicated login route (/login) with prompt layout, replacing login request script
- login prompt inspired by GitHub's login & registration
- rate limiting on login attempts
- use existing form blade components
- "remember me" checkbox
- login event triggers writing activity (updates last login)

_Dark_
Screenshot 2023-09-05 at 18 34 03

_Light_
Screenshot 2023-09-05 at 18 33 52

**Registration page requires accepted terms**

Screenshot 2023-09-05 at 15 53 59

Registration will be replaced with fortify eventually.

**Planned Follow-ups**
- Laravel Fortify: registration, email verification, password management, 2FA, ...
- Laravel Passport: prompt layout will be used for OpenID auth screens


resources/views/components/menu/notifications.blade.php
```diff
@@ -1,19 +1,9 @@
...
-            </a>
-        </li>
-    </ul>
+    {{--<li class="nav-item">
+        <a class="nav-link" href="{{ route('activity') }}" title="{{ __('Activity') }}" data-toggle="tooltip">
+            <x-fas-bolt />
+        </a>
+    </li>--}}
+    <livewire:notification-icon class="{{$class ?? ''}}" />
+    <livewire:message-icon class="{{$class ?? ''}}" />
```
Jamiras, line 8: I'm not sure if there's an easy solution for this, but I find myself trying to move diagonally from the icon to the menu item, which mouses over the Manage button and changes the menu before I can click on the notification submenu item.

image

Jamiras: There is much less right-padding when there are no notifications
image

resources/views/components/menu/brand.blade.php
```diff
@@ -1,9 +1,3 @@
-<a class="navbar-brand {{ request()->routeIs('home') ? 'lg:hidden' : '' }}"
-   href="{{ route('home') }}">
-    @if(request()->cookie('logo') === 'retro')
-        <img class="hidden md:inline-block" src="{{ asset('assets/images/ra-logo-sm.webp') }}" alt="{{ config('app.name') }}">
-        <img class="inline-block md:hidden" src="{{ asset('assets/images/ra-icon.webp') }}" alt="{{ config('app.name') }}">
-    @else
-        <img class="h-6" src="{{ asset('assets/images/ra-icon.webp') }}" alt="{{ config('app.name') }}">
-    @endif
+<a class="ml-3 {{ request()->routeIs('home') ? 'lg:hidden' : '' }}" href="{{ route('home') }}">
+    <img class="h-6" src="{{ asset('assets/images/ra-icon.webp') }}" alt="{{ config('app.name') }}">
```
Jamiras, line 2: I'm not sure how well this comes across in an image, but if you scroll down on the main page so the navbar is docked to the top, then go to some other page, the menu elements shift down a pixel (but not the red dots)

image

I assume the icon is causing the height of the navbar to change a little.

resources/views/components/menu/account.blade.php
```diff
@@ -75,8 +72,8 @@
         <x-dropdown-item :link="url('controlpanel.php')">Settings</x-dropdown-item>
         <div class="dropdown-divider"></div>
         {{--<x-form :action="route('logout')">--}}
-        <x-form :action="url('request/auth/logout.php')">
-            <button class="dropdown-item">{{ __('Sign Out') }}</button>
+        <x-form :action="route('logout')">
+            <button class="dropdown-item">{{ __('Sign out') }}</button>
```
Jamiras, line 76: message/icon.blade.php doesn't appear in this diff, so I'm putting the comment here indirectly.

Now that we have a Messages icon at the top level, do we still need a Messages menu item in the Account menu? Alternatively, should we remove the Messages button from the Nav Bar, keep the Messages link in the Account menu, and list unread messages under Notifications (with the tickets/claim notifications)?

If we keep the navbar item for notifications, would it be reasonable to report the number of unread messages in the tooltip for the navbar button? "Messages (2 unread)"?

resources/views/auth/login.blade.php
```diff
@@ -0,0 +1,22 @@
...
+        <x-button type="submit" class="w-full text-center py-2 mb-4">{{ __('Sign in') }}</x-button>
+        <div class="flex flex-col justify-between">
+            <a class="btn btn-link" href="{{ url('resetPassword.php') }}">
+                {{ __('Forgot your password?') }}
+            </a>
+            <a class="btn btn-link" href="{{ url('createaccount.php') }}">
+                {{ __('Sign up') }}
+            </a>
+        </div>
+    </x-form>
```
Jamiras, line 21: Failed login reports the error twice.
image

Jamiras: Successful login always returns to the home page. Before this PR, the login would remain on the current page. I'm not sure if it's a deal-breaker, but I often notice that I'm not logged in when trying to access the Dev dropdown on a game page and seeing it's not there. Being able to login without losing my context was a nice feature. I understand that having an intermediate page complicates things.

resources/views/components/menu/management.blade.php
```diff
@@ -5,81 +5,66 @@
...
-        {{--@can('manage', App\Platform\Models\IntegrationRelease::class)
-            <x-dropdown-header>Releases</x-dropdown-header>
-            @can('manage', App\Platform\Models\Emulator::class)
-                <x-dropdown-item :link="route('emulator.index')" :active="request()->routeIs('emulator*')">Emulators</x-dropdown-item>
-            @endcan
-            <x-dropdown-item :link="route('integration.release.index')" :active="request()->routeIs('integration.release*')">Integration</x-dropdown-item>
-        @endcan--}}
-        @if($user->Permissions >= Permissions::Moderator)
-            <x-dropdown-item :link="url('admin.php')">Admin Tools</x-dropdown-item>
+    <div class="md:flex">
```
Jamiras, line 16: On sizes less than LG, the nav bar doesn't float (it scrolls with the page). Is that intentional?

resources/views/components/menu/account.blade.php
```diff
@@ -7,24 +7,21 @@
...
-            </li>
-        @endif
-        {{--<li class="nav-item">
-            <a class="nav-link" href="{{ route('login') }}">
-                <x-fas-power-off />
-                <span class="sr-only">{{ __('Sign In') }}</span>
-                <span class="hidden lg:inline-block">{{ __('Sign In') }}</span>
+    <div class="nav-item">
+        <a class="nav-link" href="{{ route('login') }}">
+            <x-fas-power-off class="mr-1" />
```
Jamiras, line 12: It feels odd for Sign In to have an icon, but not Sign Up.
image

Perhaps `fas-user-plus` for sign up, and `fas-user-pen` for sign in? 
image

resources/views/auth/login.blade.php
wescopeland: It would be ideal if there was some kind of callbackUrl or redirectUrl that after login is successful, sends the user back to that URL. Could be passed to the page as a query param.

luchaos: Laravel supports that put of the box as "intended_url" session parameter and should redirect back automatically.
I'll look into why that's not happening.

luchaos: Seems that redirect after login has to be done by hand with fortify 🤔 
Fixed in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/704d4bc4e40ba259b41c9bad114c1796287b6be9

Removed global messages from prompt layout in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/e27de4cbf0818b3ff02e6961616f5a33f26e6c22

resources/views/components/menu/brand.blade.php
luchaos: What an interesting find! Turned out that rounding errors in the top area texts' line heights cause the height of the top area to shift ever so slighty, affecting the whole page.
Solved by setting a fixed height for the top brand area: https://github.com/RetroAchievements/RAWeb/pull/1831/commits/954dfefbcebd17b9d4149ccf4a77973098d6da15

resources/views/components/menu/notifications.blade.php
luchaos: I found myself running into this frequently, too.
By removing the fixed width of 250px of dropdown items it would adjust its width automatically and there's maybe less of an urge to go sideways?
Screenshot 2023-09-05 at 20 21 56
It would allow the games dropdown to scale more naturally, too, actually. It's very wide already.
Screenshot 2023-09-05 at 20 23 50

resources/views/components/menu/account.blade.php
luchaos: I think adding "x unread messages" as a notification item is a neat idea.
It would cut down the amount of items listed there.

resources/views/components/menu/notifications.blade.php
luchaos: Maybe right aligned just for the notifications dropdown? It doesn't work too well for the management and profile dropdowns, but feels better for notifications:
Screenshot 2023-09-05 at 20 49 57

resources/views/components/menu/account.blade.php
luchaos: Screenshot 2023-09-05 at 20 49 57

See https://github.com/RetroAchievements/RAWeb/pull/1831/commits/358fe16e415f39a6d6e9ca2434cf3dde4f2e5e96

luchaos:
I forgot to mention, please copy `storage/app/settings.dist.json` to `storage/app/settings.json` (see https://github.com/RetroAchievements/RAWeb/pull/1831/commits/781b0dee4a716cad10588487b7670abb0114feab).
This way it will add the system management links and you can add your own:
Screenshot 2023-09-05 at 21 00 20

resources/views/components/menu/management.blade.php
luchaos: Yes, kinda. It's still the same behaviour as before and the easiest way to make sure dropdowns with long lists of items can be scrolled down fully without being cut off at the bottom on small screens. 
There are certainly more sophisticated solutions for that (make the dropdown fixed width, make itself scrollable, make menus scrolalble overlays to make proper use of small screens, etc) but that might be a good candidate for an improvement PR.

resources/views/components/menu/account.blade.php
luchaos: Was thinking about that, too, but got close to removing the icon from "Sign in" rather
(I used the very same on completionist.me 😅 which is fine)
That looks very bare on the other hand 
Screenshot 2023-09-05 at 21 13 26

resources/views/components/menu/notifications.blade.php
Jamiras: I think the right alignment is enough for me to move the cursor straight down instead of down and left, as that's where the words are instead of a gap my brain says isn't part of the link.

wescopeland:
nit: In lang/en_US/app.php:

```php
<?php

return [
    'description' => 'Adding achievements to your favourite retro games since 2012',
];
```

This is the text appearing in the heading, however "favourite" is not the correct en_US spelling. That is the spelling for en_GB. The correct spelling for en_US is "favorite".

wescopeland:
On mobile, there is a strange gap at the top of the page:

Screenshot 2023-09-05 at 4 21 10 PM

wescopeland:
Lighthouse is throwing an a11y warning on the notification button:

Screenshot 2023-09-05 at 4 22 20 PM

Jamiras:
> This is the text appearing in the heading, however "favourite" is not the correct en_US spelling. That is the spelling for en_GB. The correct spelling for en_US is "favorite".

That's the site slogan. Scott is British. It's been that way for quite some time.

image

But yes, it does feel wrong to have that in the en_US file.

wescopeland:
Yes, I am not disputing those facts. I am only suggesting that because we are going to surface this copy more heavily and because we are leaning into i18n and l10n more strongly, we should aim for correctness in our locale files.

luchaos:
that's just because en_us is the default, i had en_GB as default for some time but that seemed counterproductive. it wouldn't change for german or portugese. the slogan is technically correct like this i feel, but if scott is fine with it we can change it and add this spelling to en_GB

luchaos:
or drop "your favourite" altogether.

wescopeland:
nit: I really like how this page looks! One thing I am wondering, since both fields are required would it be appropriate to drop the "*" designator on the labels? If everything is required, it may be redundant to specify it as so. No strong opinion on this one, but wanted to throw it out there.

Screenshot 2023-09-05 at 4 35 49 PM

resources/views/components/form-field.blade.php
```diff
@@ -0,0 +1,39 @@
...
+        @if(!empty($help))
+            <p class="help-block text-secondary mb-0">
+                {!! $help  !!}
+            </p>
+        @endif
+        @if($attribute ?? null)
+            @error($attribute)
+            <p class="help-block text-danger mb-0">
+                <x-fas-exclamation-triangle /> {{ $errors->first($attribute) }}
+            </p>
```
wescopeland, line 35: When these are visible, is there a good way for the input field to have an `aria-describedby` attribute that points to their IDs?

wescopeland (commented):
Only one thing other than what I've already noted jumped out at me. Overall I think this is a great change!

luchaos:
> On mobile, there is a strange gap at the top of the home page:
> 
> Screenshot 2023-09-05 at 4 21 10 PM

Whoops, good catch; fixed in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/a9253fcff2613631fcc2693691de4a5e923f5115

luchaos:
> Lighthouse is throwing an a11y warning on the notification button:

Fixed in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/bee5bddaf0aa82fbc1c6d446aa7821f179925991

I found that the open graph prefix was incorrectly set in the doctype which caused a lower lighthouse score due to falling back to quirks mode. Moved it to head, where the open graph meta tags are actually used, and the score increased for best practices, too: https://github.com/RetroAchievements/RAWeb/pull/1831/commits/788a75ab846c351c6726a7a571ee8bb464ee6272

resources/views/components/form-field.blade.php
luchaos: Phew, that's a neat detail. Addressed in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/12f89de8f7c536e07e7de6f667e008c11be29367

luchaos:
Slogan update and alignment in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/5720b123709036e3430ba7beace6dc36f1fa6b35
Screenshot 2023-09-06 at 23 17 03

resources/views/components/menu/account.blade.php
luchaos: Updated in https://github.com/RetroAchievements/RAWeb/pull/1831/commits/0d90580495763a63bf61ef6ac7b376d928eec545
Screenshot 2023-09-06 at 23 07 55
Screenshot 2023-09-06 at 23 08 06

Jamiras (approved):

wescopeland (approved):

PR #1832, opened 2023-09-05
use multiline avatar for active claims and expiring claims pages

Jamiras:


wescopeland (approved):
Verified both pages locally, they are both looking great.

PR #1834, opened 2023-09-06
fix(home): use smalldate on recent awards

wescopeland:
Screenshot 2023-09-06 at 6 41 35 PM

Side note: It seems `.smalldate` has a forced min-width. We may want to get rid of that at some point, it is combative with flex layouts :-(


luchaos (approved):
awesome!
yeah, agreed. the smalldate styles are a bit of a leftover - would be a good improvement to clean those up and get the class name in line while at it.

PR #1837, opened 2023-09-07
ci: use paratest on pre-push

wescopeland:
Switches from `phpunit` to `paratest` on pre-push to:
* Match our GH Actions CI pipeline.
* Reduce test execution time by roughly 75%.

I am not sure if there is a major drawback to doing this, so I am opening this a bit naively. However, our test execution time is going to be creeping up more and more, so taking some step to mitigate seems appropriate at this point.

### DIFF
--- a/.hooks/pre-push
+++ b/.hooks/pre-push
@@ -17,7 +17,7 @@ npx eslint --quiet
 printf "\n⏳ npm run test\n"
 npm run test
 
-printf "\n⏳ composer test\n"
-vendor/bin/phpunit
+printf "\n⏳ composer paratest\n"
+vendor/bin/paratest
 
 printf "\n✅ pre-push OK\n\n"


luchaos:
I like the idea and consistency but there is a reason indeed, well was at least.
Running paratest caused issues on some Windows setups in the past - maybe this changed? Would be nice to have it.

drisc:
Was that on Windows setups running WSL or on Windows itself, I can give WSL a test later today.

luchaos:
> Was that on Windows setups running WSL or on Windows itself, I can give WSL a test later today.

Windows itself iirc. WSL most likely works because of the linux context. That's what i hooe at least.
Would be great to find out for both 🙏

wescopeland:
I figured Windows might be the culprit. We can use `uname` in this script to detect if a Windows OS is running (both in WSL and native) and gracefully fall back to phpunit, else run paratest.

.hooks/pre-push
```diff
@@ -17,7 +17,7 @@ npx eslint --quiet
 printf "\n⏳ npm run test\n"
 npm run test
 
-printf "\n⏳ composer test\n"
-vendor/bin/phpunit
+printf "\n⏳ composer paratest\n"
+vendor/bin/paratest
```
Jamiras, line 21: is there a difference between this and `composer paratest`, which runs `php artisan test --parallel`?

luchaos: by running it through composer it might be subjetct to composer's execution time limit. realistically we could switch all of those commands to go through composer scripts as the echo statements imply.

PR #1844, opened 2023-09-08
feat(navbar): put softcore points on top for softcore players

wescopeland:
Screenshot 2023-09-07 at 8 03 21 PM

If the user has more softcore points than hardcore points, use softcore points as the first points indicator in the navbar.

### DIFF
--- a/resources/views/components/menu/account.blade.php
+++ b/resources/views/components/menu/account.blade.php
@@ -27,15 +27,21 @@ $user = request()->user();
 @endguest
 @auth
     <div class="nav-link flex-col justify-center items-end text-2xs" style="line-height: 1.1em">
+        @if($user->RASoftcorePoints && $user->RASoftcorePoints > $user->points_total)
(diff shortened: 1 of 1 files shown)


resources/views/components/menu/account.blade.php
```diff
@@ -27,15 +27,21 @@
...
         @endif
+
         @if($user->points_weighted_total)
             <x-points-weighted-container>
                 <span class='trueratio'>{{ localized_number($user->points_weighted_total) }}</span>
             </x-points-weighted-container>
         @endif
-        @if($user->RASoftcorePoints)
+
+        @if($user->RASoftcorePoints && $user->RASoftcorePoints < $user->points_total)
```
Jamiras, line 44: This should be `<=`. Softcore points disappear entirely if they're equal to hardcore points.

wescopeland: Great catch, resolved in https://github.com/RetroAchievements/RAWeb/pull/1844/commits/d9a4758fca15826acb323b0d356bd837f959f757.

Jamiras (approved):

PR #1847, opened 2023-09-08
fix(dropdown): text font-size

w400pedro:
A very simple change that lowers the font size of **dropdown elements** by 1 pixel to make it cleaner. (without including the developer's header.)

13px -> 12px

The navbar elements have the same font size as the dropdown menu elements, which makes it unnecessarily big

13px (before):
image

12px (after): 
image

### DIFF
--- a/resources/css/dropdown.css
+++ b/resources/css/dropdown.css
@@ -40,6 +40,7 @@
   padding: 4px 12px;
   border: none;
   border-radius: 0;
+  font-size: 12px;
 }
 
 .dropdown-item:last-child {


luchaos:
The font size change was intentional to move towards an overall larger font size. This site is still stuck with very small text which is an accessibility concern.
Personally, I'm fine with small font sizes - not sure if decreasing it in the menus was discussed beforehand?

resources/css/dropdown.css
```diff
@@ -40,6 +40,7 @@
   padding: 4px 12px;
   border: none;
   border-radius: 0;
+  font-size: 12px;
```
luchaos, line 43: Note that this is a global component and will not be only used in the menu. Was it intentional to change it globally or just for the manu?

w400pedro: It wasn't intentional to modify the dropdowns globally, but only those in the navbar. I only found this class in the navbar ones, so I made this modification directly through it. By the way, if the goal of keeping the font at this larger size is accessibility, then I think it's fair to maintain it, at first, I thought it was a silly mistake.

PR #1850, opened 2023-09-08
prevent invalid argument error when user has no awards

Jamiras:


public/globalRanking.php
```diff
@@ -285,7 +285,7 @@
                 echo "<td class='text-right'>" . localized_number($dataPoint['Points']) . "</td>";
             }
 
-            echo "<td class='text-right'>" . localized_number($dataPoint['TotalAwards']) . "</td></tr>";
+            echo "<td class='text-right'>" . localized_number($dataPoint['TotalAwards'] ?? 0) . "</td></tr>";
```
luchaos, line 288: We don't show a number when it's 0 in some areas. 
We could adjust localized_number to accept null to then return an empty string for convenience.

Jamiras: In this case, I think saying the user has "0 "masteries is more appropriate than leaving the column blank.

luchaos (approved):

Issue #1855, opened 2023-09-08
Badges descentralized


w400pedro:
**Describe the bug**
Apparentelly the badge in profile is not centralized with the background 

**To Reproduce**
Take a look at some profile with 5 or more badges

**Expected behavior**
The badges be re-centralized in the background

**Screenshots**

Desktop: 
![image](https://github.com/RetroAchievements/RAWeb/assets/47864347/97a51db7-ab58-4ce3-b5f4-c7caf168c7a7)

Mobile: 
![image](https://github.com/RetroAchievements/RAWeb/assets/47864347/126590be-675d-4dbf-a647-bb96b54cda7f)

**Adittional Context**
Note that in the Desktop version, the badges are a few more pixels to the left than to the right, while in the mobile version, it is fully positioned on the left

PR #1856, opened 2023-09-08
fix: always justify-center badges

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1855.


Jamiras:
Is this really the look you're after?
image

wescopeland:
No. That's pretty bad. Let's leave this as-is.

PR #1857, opened 2023-09-08
feat(game): allow sorting by achievement type

wescopeland:

This PR adds a new sorting option to game pages: "Type", which orders achievements by Progression, Win Condition, and then `null`.

https://github.com/RetroAchievements/RAWeb/assets/3984985/85974932-6d72-46ef-91eb-f7e9c9665a03


Tsearo (approved):
Works great 👍 

Just a thought but would it make sense to hide the Type sort when there are no existing typed achievements? All of the other sorts always apply but this one is semi-conditional at least until all games are marked which may take some time after the feature goes live (and also for subsets which won't get marked). Likely not worth the effort but thought I'd at least bring it up

wescopeland:
Yes, very good idea. We should make that change-- I'll update the code in this PR in just a bit.

wescopeland:
Conditional rendering logic added in https://github.com/RetroAchievements/RAWeb/pull/1857/commits/eebadf47bd6ea370f64de3b74d3f1274442a66a7.

app/Helpers/database/game.php
```diff
@@ -68,6 +68,10 @@ function getGameMetadata(
         14 => "ORDER BY ach.Points DESC, ach.ID DESC ",
         5 => "ORDER BY ach.Title, ach.ID ASC ",
         15 => "ORDER BY ach.Title DESC, ach.ID DESC ",
+        6 => "ORDER BY 
+            CASE WHEN ach.type IS NULL THEN 2 WHEN ach.type = 'progression' THEN 0 WHEN ach.type = 'win_condition' THEN 1 ELSE 3 END, 
+            ach.ID ASC ",
```
Jamiras, line 73: It would probably be better to use `ach.DisplayOrder` as the secondary sort, with `ach.ID` as the tertiary sort (i.e. use the "default=>" sort as the secondary sort).

app/Helpers/database/game.php
```diff
@@ -68,6 +68,10 @@ function getGameMetadata(
...
+        16 => "ORDER BY ach.type, ach.ID DESC ",
```
Jamiras, line 74: `16` should result in the same list as `6`, just reversed. I would expect the Case logic to be present here too. As it stands, because `progression` is alphabetically before `win_condition`, the win condition is not in the right place.

app/Helpers/database/game.php
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1857/commits/dc335a38c85df9a807b4276afa4966fac32151cc.

app/Helpers/database/game.php
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1857/commits/dc335a38c85df9a807b4276afa4966fac32151cc.

Jamiras (approved):

PR #1858, opened 2023-09-08
fix(current-progress): improve secondary actions menu ux

wescopeland:
This PR slightly adjusts the game progress component's secondary actions menu. This has the desirable effect of ensuring the mouse doesn't immediately land on the "Reset All Progress" button, which could otherwise cause some anxiety for the user.

Screenshot 2023-09-09 at 7 11 11 AM


Krylan:
I also felt that this should be changed, but had another idea for that:
In my opinion, dropdown should not cover the button, which opens it.
So instead of adding a label, I would just move the actions menu below the button.
image

This way action isn't located in the same place where user clicked (otherwise user could accidentaly double click and activate the option they don't intended). Also, button is still available to close the menu (I know that clicking anywhere closes the menu, but I guess that it's good practice to make the button available, as it would be intuitive).

wescopeland:
I agree with your opinion.

I have modified the border radius of the button, the position of the menu, and its open/close animation.


https://github.com/RetroAchievements/RAWeb/assets/3984985/8a9c9866-6bed-46ac-96fb-c4ecdbec5c80

Jamiras (approved):

PR #1862, opened 2023-09-10
feat: add Beaten Games global leaderboard [progression]

wescopeland:

**This is the final large UI piece for achievement types.**

This PR adds the global hardcore beaten games leaderboard. The leaderboard has various filtering options to allow a user to get precisely the view they would like. The user is always able to see their standing against their current filter criteria.

The new page can be accessed at `/ranking/beaten-games`.

**CAVEAT: Softcore beaten games do not have a leaderboard.**

### Entrypoints

There are two entrypoints to the new page: the app navbar and the home page. "Global Ranking" has been renamed to "Global Points Ranking" on both the home page and the navbar. The new link is titled "Global Beaten Games Ranking".

**App navbar**
Screenshot 2023-09-09 at 7 42 48 PM

**Home page**
Screenshot 2023-09-09 at 7 43 10 PM

On the home page, "Download Emulator" has been moved to the top so the two leaderboard links can still live next to each other. The icon for "Global Points Ranking" now also uses the same icon as the "Achievements" heading in the navbar, which seems logical as points are directly tied to the achievements that grant them.

---

### Empty State

If there are no rows to display, an empty state is shown.

**Mobile**
Screenshot 2023-09-10 at 8 58 52 AM

**Desktop**
Screenshot 2023-09-10 at 8 58 57 AM

---

### Filter Criteria

The user has two filter categories: selected system and selected game kind(s). By default, all systems and all game kinds are selected. However, the user can drill down into the specific leaderboard filter criteria they care most about with ease.

**Mobile**
Screenshot 2023-09-10 at 8 59 18 AM

**Desktop**
Screenshot 2023-09-10 at 8 59 15 AM

After discussions with @luchaos, these filters are implemented like so:
* `?filter[system]=1` (only show Mega Drive standings)
* `?filter[retail]=false` (exclude all games that don't have two "~" characters in the title)
* `?filter[hacks]=false` (exclude all hacks from standings)
* `?filter[homebrew]=false` (exclude all homebrew from standings)
* `?filter[unlicensed]=false` (exclude all unlicensed games from standings)
* `?filter[prototypes]=false` (exclude all prototype games from standings)
* `?filter[demos]=false` (exclude all demos from standings)

**CAVEAT: `~Subset~`, `~Test Kit~`, `~Multi~`, and events are _always_ excluded from standings.**

**CAVEAT: The homebrew filter excludes all titles from the Arduboy, WASM-4, and Uzebox consoles, as these consoles _exclusively_ have homebrew games.**

---

### Leaderboard UI

The leaderboard shows 25 rows of rankings per page. If the user is authenticated, they can always see what their rank is (as long as they have one for their current filter criteria). 

On desktop, each row contains the rank, associated username, their last game beaten, when they beat that game, and how many games they've beaten.

On mobile, each row contains the rank, associated username, and how many games they've beaten.

**Desktop**
Screenshot 2023-09-09 at 7 56 50 PM

Screenshot 2023-09-09 at 7 56 55 PM

**Mobile**
Screenshot 2023-09-09 at 7 57 10 PM

Screenshot 2023-09-09 at 7 57 02 PM

---

### Pagination Controls

25 rows are returned per page. There is a "Previous" and "Next" button clickable by the user. Additionally, if the user has a rank based on their current filter criteria, another button appears letting them jump directly to the page they're on. If the user is on Page 1, the Previous page button is visible but disabled. If the user is on the last page, the Next page button is visible but disabled.

**Desktop**
Screenshot 2023-09-09 at 7 58 48 PM

**Mobile**
Screenshot 2023-09-09 at 7 58 55 PM

The pagination query param is implemented like so:
* `page[number]=2`

---

### All Noted Caveats
* Softcore beaten games do not have a leaderboard.
* `~Subset~`, `~Test Kit~`, `~Multi~`, and events are _always_ excluded from standings.
* The homebrew filter excludes all titles from the Arduboy, WASM-4, and Uzebox consoles, as these consoles _exclusively_ have homebrew games.

---

https://github.com/RetroAchievements/RAWeb/assets/3984985/c23e1895-31a2-410f-8ee4-816ff5e51aeb


wescopeland:
https://github.com/RetroAchievements/RAWeb/pull/1862/commits/aeba4452d6bf9ba8fbc36687e9bcf05b6039a835 was co-authored by @Tsearo. I messed up coauthorship credit in the commit body 😵‍💫 I'll try to attach it when we merge the PR.

resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
```diff
@@ -0,0 +1,38 @@
...
+    'allSystems' => null,
+    'gameKindFilterOptions' => [],
+    'selectedConsoleId' => null,
+    'selectedAllowHacks' => true,
+])
+
+<script>
+function handleConsoleChanged(event) {
+    const newQueryParamValue = event.target.value;
+    window.updateUrlParameter('filter[system]', newQueryParamValue);
```
Jamiras, line 11: Changing filters does not reset the current page. This can result in misleading "There aren't any rows matching your current filter criteria." errors, even though there would be rows on the first page.

To reproduce: go to page 2 or 3, then select a system with very few awards (like TV Games Computer)

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -0,0 +1,196 @@
...
+            ->distinct('filteredGames.user')
+            ->count('filteredGames.user');
+
+        $paginator = new LengthAwarePaginator($beatenGameAwardsRankedRows, $rankedRowsCount, $this->pageSize, $currentPage, [
+            'path' => $request->url(),
+            'query' => $request->query(),
+        ]);
+
+        // Grab all the systems so we can build the system filter options.
+        $allSystems = System::orderBy('Name')->get(['ID', 'Name']);
```
Jamiras, line 77: Events appears to still be a selectable item.
image

And the list seems to be using something other than `Name` for display. For example, `Name` for console 63 is "Watara Supervision", but it just shows "Supervision" in the dropdown, which seems out of place given the sort was on "Watara".
image

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -0,0 +1,196 @@
...
+        // We also need to know how many rows there are, otherwise the
+        // paginator can't determine what the max page number should be.
+        $rankedRowsCount = $this->buildFilteredGamesQuery($targetSystemId, $gameKindFilterOptions)
+            ->distinct('filteredGames.user')
+            ->count('filteredGames.user');
+
+        $paginator = new LengthAwarePaginator($beatenGameAwardsRankedRows, $rankedRowsCount, $this->pageSize, $currentPage, [
+            'path' => $request->url(),
+            'query' => $request->query(),
+        ]);
```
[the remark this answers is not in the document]
wescopeland: No problem. In latest, I've added the ability to jump to first if on a page greater than 2. I've also added arbitrary page selection. Agreed that jumping to last probably doesn't make sense in this context.

Screenshot 2023-09-10 at 9 03 12 PM

resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
wescopeland: Addressed in latest, these JS functions responsible for managing the filter query params will now also simultaneously reset the page number to 1.

https://github.com/RetroAchievements/RAWeb/pull/1862/commits/b113ee264709ee466b4c1087f32adcabd84b3d15#diff-a5605035ffaeff131236bb669ee4856fc0a03efb14891106d3d6ab8f482af09cR10-R20

app/Platform/Controllers/BeatenGamesLeaderboardController.php
wescopeland: Using `Name` in latest: https://github.com/RetroAchievements/RAWeb/pull/1862/commits/b113ee264709ee466b4c1087f32adcabd84b3d15#diff-b591e62d226f39ad0c2697521c831ba732612a8dc708272dd6e816a7a22ff7e1R16

Also, "Events" is filtered out now:
https://github.com/RetroAchievements/RAWeb/pull/1862/commits/b113ee264709ee466b4c1087f32adcabd84b3d15#diff-b591e62d226f39ad0c2697521c831ba732612a8dc708272dd6e816a7a22ff7e1R16

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -0,0 +1,196 @@
...
+        ?int $targetSystemId = null,
+        array $gameKindFilterOptions
+    ): mixed {
+        return $this->buildBaseLeaderboardQuery($targetSystemId, $gameKindFilterOptions)
+            ->offset($currentOffset)
+            ->limit($this->pageSize)
+            ->get();
+    }
+
+    private function buildUserRanking(
```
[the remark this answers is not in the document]
wescopeland: Yes, I think what we'll need to do is fetch the user's rank on its own (~2 seconds) and then pull their most recent beaten game straight off the site awards table.

wescopeland:
With the newly optimized queries, the "Jump to your ranking's page" button had to be axed, as we no longer have the user's row number, only their rank number.

Tsearo:
We can add the users exact index back in by using ROW_NUMBER OVER if you'd like to still keep the Jump to your ranking's page functionality. It shouldn't add any extra overhead to the query either
`ROW_NUMBER() OVER (ORDER BY COUNT(s.AuthUser) DESC) RankNum, COUNT(s.AuthUser) AS indexRow,`

resources/views/platform/components/beaten-games-leaderboard/leaderboard-card-row.blade.php
```diff
@@ -0,0 +1,20 @@
...
+])
+
+<?php
+$isHighlighted = $isHighlighted || ($paginatedRow->User === $myUsername);
+?>
+
+<div class="rounded bg-embed p-2 {{ $isHighlighted ? 'my-2' : '' }}" @if ($isHighlighted) style="outline: thin solid;" @endif>
+    <div class="flex items-center justify-between">
+        <div class="flex gap-x-3 items-center">
+            <p class="text-sm">#{{ localized_number($paginatedRow->rank_number) }}</p>
```
Tsearo, line 14: This can attempt to read a rank_number property on null if you go to a page without any users.

resources/views/platform/components/beaten-games-leaderboard/leaderboard-table.blade.php
```diff
@@ -0,0 +1,54 @@
...
+    <tbody>
+        @foreach ($paginator as $paginatedRow)
+            <x-beaten-games-leaderboard.leaderboard-table-row
+                :myUsername="$myUsername"
+                :paginatedRow="$paginatedRow"
+            />
+        @endforeach
+
+        @if ($myRankingData && !$isUserOnCurrentPage)
+            <tr class="do-not-highlight"><td colspan="4">&nbsp;</td></tr>
```
Tsearo, line 45: I am still seeing this issue and believe it needs to be a colspan of 5 instead of 4, it cuts off slightly early for me still in the latest
image

wescopeland: I think the reason I'm not seeing it is our browsers have a different opinion of which row to colorize based on the CSS class.

Screenshot 2023-09-12 at 8 30 39 PM

Resolved in https://github.com/RetroAchievements/RAWeb/pull/1862/commits/ebe4309effa535db7dc91e682a2130773bbc29ce.

resources/views/platform/components/beaten-games-leaderboard/leaderboard-card-row.blade.php
wescopeland: Just out of curiosity, how are you landing on a page with no users?

wescopeland:
>We can add the users exact index back in by using ROW_NUMBER OVER if you'd like to still keep the Jump to your ranking's page functionality. It shouldn't add any extra overhead to the query either
ROW_NUMBER() OVER (ORDER BY COUNT(s.AuthUser) DESC) indexRow

I can't seem to get Eloquent ORM to be happy with this.

```php
->select(
    'ua.User as User',
    DB::raw('FIRST_VALUE(gd.ID) OVER (PARTITION BY ua.User ORDER BY sa.AwardDate desc) as most_recent_game_id'),
    DB::raw('ROW_NUMBER() OVER (ORDER BY COUNT(s.AuthUser) DESC) indexRow'),
    'gd.Title',
    'gd.ImageIcon',
    'c.name as ConsoleName',
    'sa.AwardDate',
)
```

Always produces an error:

`SQLSTATE[42S22]: Column not found: 1054 Unknown column 's.AuthUser' in 'window order by'`

resources/views/platform/components/beaten-games-leaderboard/leaderboard-card-row.blade.php
wescopeland: I've added an existence check in latest.

Tsearo: Yep I'm no longer seeing the issue, I was just selecting a System such as Arcadia 2001 which had no awards in my local environment but now it displays the message that there are no rows matching 👍

resources/views/platform/components/beaten-games-leaderboard/leaderboard-table.blade.php
Tsearo: Looks like that fixed it, could have also been just landing on an odd or even row based on users I'm not too sure

wescopeland:
`ROW_NUMBER() OVER (ORDER BY COUNT(s.AuthUser) DESC)` is working, and as of latest, the Jump to Page button should be functional with the optimized queries.

resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
Jamiras: Now I get an `Attempt to read property "User" on null` error following the steps I listed above. I seem to get it switching to any system, even on the first page.

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -0,0 +1,225 @@
...
+
+        // We need to know how many rows there are, otherwise the
+        // paginator can't determine what the max page number should be.
+        $rankedRowsCount = $this->getLeaderboardRowCount($targetSystemId, $gameKindFilterOptions);
+
+        // Now get the current page's rows.
+        $currentPage = $validatedData['page']['number'] ?? 1;
+        $offset = (int) ($currentPage - 1) * $this->pageSize;
+
+        $beatenGameAwardsRankedRows = $this->getLeaderboardDataForCurrentPage($offset, $targetSystemId, $gameKindFilterOptions);
```
Jamiras, line 57: The sort seems to be alphabetical by user now. This is page 1:
image

wescopeland: Could this be a side effect of using MariaDB? I am definitely not seeing this.

Screenshot 2023-09-13 at 10 08 39 AM

@Tsearo Can you repro this issue?

Tsearo: Rank ordering worked fine for me during testing (MySQL) so it's possible it could be a MariaDB issue

Jamiras: The test passes. Here's the query that's being run:
```
select * from (select `User`, 
    RANK() OVER (ORDER BY COUNT(s.User) DESC) as rank_number,
    ROW_NUMBER() OVER (ORDER BY COUNT(s.User) DESC) as leaderboard_row_number,
    COUNT(s.User) as total_awards, `most_recent_game_id`, `Title` as `GameTitle`,
    `ImageIcon` as `GameIcon`, `ConsoleName`, 
    `AwardDate` as `last_beaten_date` from (
        select `ua`.`User` as `User`,
            FIRST_VALUE(gd.ID) OVER (PARTITION BY ua.User ORDER BY sa.AwardDate desc) as most_recent_game_id,
            `gd`.`Title`, `gd`.`ImageIcon`, `c`.`name` as `ConsoleName`, `sa`.`AwardDate`
            from `SiteAwards` as `sa` inner join `GameData` as `gd` on `sa`.`AwardData` = `gd`.`ID`
            inner join `Console` as `c` on `gd`.`ConsoleID` = `c`.`ID`
            inner join `UserAccounts` as `ua` on `sa`.`User` = `ua`.`User`
            where `sa`.`AwardType` = 8 and `sa`.`AwardDataExtra` = 1 and `ua`.`Untracked` = 0
            and `gd`.`Title` not like '~Subset~%' and `gd`.`Title` not like '~Test Kit~%'
            and `gd`.`Title` not like '~Multi~%' and `gd`.`ConsoleID` not in (100, 101)
    ) as s
    group by `User`
) as b limit 25 offset 0
```
There probably needs to be an outer-level ORDER BY.

resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/1862/commits/38a30ffa76b61f7f2e0da092e868e649ad0ea141.

app/Platform/Controllers/BeatenGamesLeaderboardController.php
wescopeland: Outer-level `ORDER BY` added in https://github.com/RetroAchievements/RAWeb/pull/1862/commits/38a30ffa76b61f7f2e0da092e868e649ad0ea141.

Jamiras:
Minor: My "Jump to Your Ranking's Page" goes to page 15, but my entry is on page 14, and my rank actually spreads from pages 13-15.

Jamiras (approved):

wescopeland:
>My "Jump to Your Ranking's Page" goes to page 15, but my entry is on page 14, and my rank actually spreads from pages 13-15.

I'm hoping this is caused by a 2nd missing outer `ORDER BY`. Added in https://github.com/RetroAchievements/RAWeb/pull/1862/commits/d16bbbd2f461d635fc65e1c8cc1c37fa5497bfbc.

Jamiras:
> I'm hoping this is caused by a 2nd missing outer `ORDER BY`. Ac).

Unfortunately, it seems to have had the opposite effect. Now I'm on page 15 and it jumps to page 14.

The sort does seem to be inconsistent for tied rankings. Most of the time my page 15 starts like this:
image

But sometimes it starts like this:
image

wescopeland:
Reverted the `ORDER BY` change. I am not seeing this locally; it doesn't seem to be a showstopper but I am going to hope it's a MariaDB discrepancy.

docker/mysql/mysql.cnf
```diff
@@ -13,3 +13,4 @@ collation-server = utf8mb4_unicode_ci
 
 character-set-client-handshake = FALSE
 init_connect = "SET NAMES utf8mb4 COLLATE utf8mb4_unicode_ci"
+sql_mode = IGNORE_SPACE,STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION
```
wescopeland, line 16: Before merging:
* We should verify whether or not this is still needed with MariaDB.
* We should ensure this PR works with MariaDB, as there were reported discrepancies between MySQL and MariaDB during testing.

luchaos: sql mode is set on the connection level in the database config - server sql mode should stay as-is.
I'll look into that further

wescopeland: Confirmed this line of config no longer seems necessary. Removed in https://github.com/RetroAchievements/RAWeb/pull/1862/commits/63c22cc0e23e87ad97eecacae8336b6963111cc6.

wescopeland:
Verified everything appears to be functionally valid with MariaDB.

However, the page does seem to be dreadfully slow.

Screenshot 2023-10-29 at 8 16 24 PM

wescopeland:
I'm not sure how comfortable I am shipping this feature with the perf numbers I'm seeing locally.

I've tried a few naive optimizations using `player_games` instead of `SiteAwards` to no avail. May need to think more deeply about this one for a bit.

wescopeland:
Incorporated `FOUND_ROWS()` in https://github.com/RetroAchievements/RAWeb/pull/1862/commits/5c9b06fb4dde1e237ecd1f33f131b7209cd4ab05 to remove one of the three slow queries.

Locally, system-specific leaderboards all are loading for me in 3 seconds or less, regardless of what filters are applied.

The "all systems" leaderboard is around 11s total (down from 16s before the most recent change).

Per discussions with @luchaos, I'd like to merge this as-is so subsequent optimization work that will immediately follow is more tightly focused in a follow-up PR.

PR #1866, opened 2023-09-11
add PPSSPP to download page

Jamiras:


wescopeland:
Does this need a `minimum_version` of 1.16?

Jamiras:
> Does this need a `minimum_version` of 1.16?

I'm pretty sure `minimum_version` is only needed/used by the emulators that integrate with the DLL. I'll add it anyways.

wescopeland (approved):

PR #1815, opened 2023-09-01
Add 'Want to Play Games' list

Jamiras:
Adds a new button to individual games pages that allows a user to bookmark a game to play in the future.
image

When the button is clicked, a server request is made to add the game to the list, and the button changes from a plus to a check mark. In both states, a tooltip on the button helps to clarify the functionality. 
image

NOTE: The white button is due to mouseover, and does not reflect the fact that the item is or isn't on the list.

To access the list, a new item was added to the user menu:
image

I put it near the top as I expect it to be a prominent feature used by many users. If there's a better place to put it, let me know.

Clicking the link opens the list, which is rendered using the same logic as the "All Games" list.
image

The feature is currently restricted to Registered users, but there's no limit to the number of items a user can add to their list. Should there be?

-----

**NOTE: Requires running the two UPDATE queries from #1761**

PR #1833, opened 2023-09-06
feat: add Completion Progress page [progression]

wescopeland:
> ✅ Safe to merge, supports the "beat" feature flag.

This PR adds the new Completion Progress page, which is intended to be the destination of the component built in https://github.com/RetroAchievements/RAWeb/pull/1816.

The page is accessible at `/user/{username}/progress`.

## UX Walkthrough

Screenshot 2023-09-06 at 6 17 13 PM

The Completion Progress page provides an overview of a user's entire play history on RetroAchievements. Users can see their gameplay history, milestones, awards metadata, and can filter down into specific data to see the games they're interested in at the moment.

The game list consists of rows that look like:

Screenshot 2023-09-06 at 6 18 39 PM

The date is when the user last earned an achievement for that game.

If the game has an award associated with it, the award indicator changes and a new line of copy is added to the row:

Screenshot 2023-09-06 at 6 19 45 PM

The embed at the top of the page shows counts of games by various award statuses. These counts are dynamic, changing based on which console you're filtered onto at the time. Clicking one of the counts applies its respective filter:

Screenshot 2023-09-06 at 6 20 18 PM

Many progression status filtering options are provided to choose from:

Screenshot 2023-09-06 at 6 21 51 PM

Additionally, milestones can be found in the sidebar. For now, these are just showing awards granted at a handful of landmark ordinal intervals. In the future, after we cut over to aggregated queries, it would make sense to show things such as: first achievement earned, rarest achievement earned, rarest mastery, rarest beaten game, etc.:

Screenshot 2023-09-06 at 6 23 15 PM

Like award totals, milestones also update dynamically based on your current filter criteria.

Only 100 results are shown per page. A pagination control is provided at the bottom of the screen:
Screenshot 2023-09-06 at 6 24 28 PM

The UI is mobile-first:

Screenshot 2023-09-06 at 6 31 17 PM

---

https://github.com/RetroAchievements/RAWeb/assets/3984985/82644715-484d-4009-b1f8-b9274b249342

---

### Implementation Details

* This page is fully built from top to bottom using Blade components. The only exception is a single call to `userAvatar()` and a few calls to `renderGameTitle()`.
* Nearly all of the data on the page is derived from a single call to `getUsersCompletedGamesAndMax()` and `getUsersSiteAwards()`. Where applicable, aggregated queries have been updated to support the imminent cutover.
* The page is currently protected by auth. If the user is not authenticated, the page returns a 401.
[...]

PR #1867, opened 2023-09-11
Want to Develop list

Jamiras:
Expands upon #1815 by adding a Want to Develop Games list, accessible to all full-Developers. It was decided that the feature should not be available to Junior Developers as they should be focused on making the best single set they can, which would lead to becoming a full-Developer and access to the feature.

The "Want to Develop" button is embedded in the Dev panel, next to the "Make Claim" button.
image

Once a claim is made, the developer can view other interested parties via a secondary link in the Dev panel:
image

That link goes to a page similar to the "View Set Requests" page, listing all the developers who have added the game to their "Want to Develop" list. This gives the claimant insight into other devs that they could bounce ideas off of, or inquire about collaborating.

The "View Developer Interest" link is also visible to moderators, even if the game isn't claimed.

While users who are interested in developing for a specific game may be viewed by a moderator, or a user who has claimed the game, the full list associated to an individual developer is only accessible by the developer.

-----

A link to the developer's list has been added to the developer section of the account dropdown.
image

In addition to adding games the developer has interest in making a set for, games with achievements can be added to the list in case the developer wants to revise a set, or work on tickets.

image

Games are never automatically removed from the Want to Develop list. The developer may choose to remove the game from their list once it's claimed, once the set is released, or whenever they decide it's just cluttering up the list.

-----

I had considered adding site notifications/emails for when a Want to Develop game gets claimed, but decided that should wait until we have a more robust collaboration system. When a user claims a game, they may not want other interested developers jumping on the claim, and there currently isn't any way to prevent that. Interested users can still see the claim on the front page, or in the discord #set-claim logs, or even on the Want to Develop list, and engage with the claimant in the same manner they would currently.


app/Helpers/database/set-claim.php
```diff
@@ -182,7 +182,16 @@ function getClaimData(int $gameID, bool $getFullData = true): array
...
+    if (is_array($gameID)) {
+        if (empty($gameID)) {
+            return [];
+        }
+        $gameIDs = implode(',', $gameID);
+    } else {
+        $gameIDs = $gameID;
+    }
+
+    $query .= " FROM SetClaim sc WHERE sc.GameID IN ($gameIDs)
```
Jamiras, line 194: Fetching 25 claims in one query took about as long as fetching 2 claims individually.

public/gameInfo.php
```diff
@@ -1031,6 +1047,17 @@ function submitSetRequest(user, gameID) {
 
                 // Display the claims links if not an event game
                 if (!$isEventGame) {
+                    if ($permissions >= Permissions::Developer) {
+                        $gameMetaBindings['developListType'] = UserGameListType::Develop;
+                        echo Blade::render('
+                            <x-game.add-to-list
+                                :gameId="$gameID"
+                                :type="$developListType"
```
wescopeland, line 1055: It's hard not to notice the new button having a different letter spacing, font size, and height from the other buttons in the devbox. Visually, the difference is striking enough to mistake the new button as being a "primary action" (the new button is nearly 2x the height of the others).

Recommend when the `type` prop is `UserGameListType::Develop`, the button styles change to match the other buttons in the devbox.

Screenshot 2023-09-11 at 6 57 20 PM

public/dev-interest.php
wescopeland, line 1: While as a team we haven't formally stated new pages should be using Laravel MVC, I have a feeling we will later regret adding new raw .php pages like this, even for more trivial features.

#1833 and #1862 both avoid this by:
* Add a new controller to handle the view: https://github.com/RetroAchievements/RAWeb/pull/1862/files#diff-09011a6920853a763a114b2c30c48af5e3352c02b1718f01f6a74b6818bf14f1

* Add a new root blade component for the page which uses `<x-app-layout>` as the root node and the `pageTitle` prop for the `<title>` tag: https://github.com/RetroAchievements/RAWeb/pull/1862/files#diff-fd452dc927a97fa89a704b32391db8d0dd9b0e9a53350a086e196e58b3451450

* Add a new line to _RouteServiceProvider.php_ to hook up the new route (and keep the PHP file extension out of the URL): https://github.com/RetroAchievements/RAWeb/pull/1862/files#diff-4454baa45c0491c9db3f881c7a954ba52eed2713a53009a7d5c972d2f39f2c52R66

Now the back-end logic is separated in a controller away from the UI. The UI doesn't need any `echo` statements and can be 100% Blade:

```php
<x-app-layout pageTitle="{{ $gameTitle['Title'] }} - Developer Interest">
  <h2>List of Interested Users</h2>

  ...
</x-app-layout>
```

I'm happy to help with this conversion in this PR if you would like.

public/gameInfo.php
Jamiras: I can fix the font size by removing the `text-sm` from the span block, but everything I try to set for the image sizes cause the image to disappear.

https://github.com/RetroAchievements/RAWeb/blob/4d1acc47288928caadae2c790cdb00dcb23d9b5c/resources/views/platform/components/game/add-to-list.blade.php#L30-L36

image

I don't think that the logic need be conditional. A slightly smaller button in the header would be fine.

public/dev-interest.php
Jamiras: I'll see what I can do. I'll message you when/if I need help.

app/Platform/Controllers/GameDevInterestController.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    public function __invoke(Request $request): View
+    {
+        if (!$request->user()) {
+            abort(401);
+        }
+
+        $permissions = $request->user()->getAttribute('Permissions');
+        if ($permissions < Permissions::Developer) {
+            abort(401);
```
wescopeland, line 24: nit: This should probably be a 403. The user is authenticated, but they are not authorized.

app/Platform/Controllers/GameDevInterestController.php
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        $gameId = (int) $request->route('game');
+        $gameData = getGameData($gameId);
+        if ($gameData === null) {
+            abort(404);
+        }
+
+        if ($permissions < Permissions::Moderator && !hasSetClaimed($request->user()->User, $gameId, true)) {
+            abort(401);
```
wescopeland, line 34: nit: This should probably be a 403. The user is authenticated, but they are not authorized.

public/gameInfo.php
wescopeland: It looks like something in the SVG for the icons is causing the issue. This is where I ended up:

```php
<button 
    id='{{ $type }}-list-button'
    class='btn'
    type='button'
    title='{{ $buttonTooltip }}'
    onClick="toggleListItem('{{ $type }}')"
>
    <div class="flex items-center gap-x-1">
        <div
            id='add-to-{{ $type }}-list'
            class="{{ $addVisibility }} {{ $type === 'play' ? 'w-[18px] h-[18px]' : '' }}"
        >
            @if ($type === 'play')
                <x-icon.add-to-list />
            @else
                <x-fas-plus class="-mt-0.5 w-[12px] h-[12px]" />
            @endif
        </div>

        <div
            id='remove-from-{{ $type }}-list'
            class="{{ $removeVisibility }} {{ $type === 'play' ? 'w-[18px] h-[18px]' : '' }}"
        >
            @if ($type === 'play')
                <x-icon.remove-from-list />
            @else
                <x-fas-check class="-mt-0.5 w-[12px] h-[12px]" />
            @endif
        </div>

        <span>
            {{ __('user-game-list.' . $type . '.name') }}
        </span>
    </div>
</button>
```

Screenshot 2023-09-12 at 7 57 46 PM

televandalist:
`I had considered adding site notifications/emails for when a Want to Develop game gets claimed, but decided that should wait until we have a more robust collaboration system. When a user claims a game, they may not want other interested developers jumping on the claim, and there currently isn't any way to prevent that. Interested users can still see the claim on the front page, or in the discord #set-claim logs, or even on the Want to Develop list, and engage with the claimant in the same manner they would currently.`

Maybe instead of having a single button for making a primary claim, have two? One for an "open" primary claim and another for a "closed" primary claim. If it's open, interested developers would get a notification and would be able to hop aboard for a collaboration claim. If it's closed, then it's basically locked until it's completed, dropped, expired, or set to open.

Just a thought, but I feel that it would work.

televandalist:
Something else that I think would be useful is being able to make notes/comments of some sort on games marked as "want to develop." It would be a good way for someone to express what they'd be interested in doing for the set, whether it's just badges, RP, a couple specific challenges, etc. 

Maybe a comment system for the "View Developer Interest" page?

public/gameInfo.php
Jamiras: I think the non-circle icons look find for the Want to Play button and removed the conditional logic. Thanks for your help.

wescopeland (approved):
Verified locally, everything is working great.

PR #1873, opened 2023-09-14
chore: disable web-based leaderboard editing

wescopeland:
Sets the start, cancel, submit, and value fields in leaderboardList.php to read-only. Now that RAIntegration supports editing leaderboards, and given that we have no validation of input on the web, it doesn't make sense for these fields to be editable in the web UI anymore.

Screenshot 2023-09-14 at 5 45 14 PM

### DIFF
--- a/public/leaderboardList.php
+++ b/public/leaderboardList.php
@@ -113,7 +113,7 @@ if ($permissions >= Permissions::JuniorDeveloper) {
 
 $listCount = 0;
 
-$bgColorClassNames = ["!bg-box-bg", "!bg-embed"];
+$bgColorClassNames = ["!bg-embed", "!bg-box-bg"];
 $currentBgColorIndex = 1;
 
 foreach ($lbData as $nextLB) {
--- a/public/leaderboardList.php
+++ b/public/leaderboardList.php
@@ -82,37 +82,6 @@ if (isset($gameData['ID'])) {
     echo "</div>";
 }
 
-if ($permissions >= Permissions::JuniorDeveloper) {
-    $numGames = getGamesList(0, $gamesList);
-
-    echo "<div class='devbox'>";
-    echo "<span onclick=\"$('#devboxcontent').toggle(); return false;\">Dev ▼</span>";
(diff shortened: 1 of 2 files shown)


Tsearo (commented):
Should we remove the `Create Leaderboard` button as well since there isn't much reason to create them through the site if we can't edit them.

The Duplicate leaderboard option may still have some niche use but could also be removed since you can clone leaderboards in the toolkit

wescopeland:
Thanks, I didn't realize those options were even there. Removed in https://github.com/RetroAchievements/RAWeb/pull/1873/commits/f1c954ef53bb4526bf942a191989a85fa899a91a.

public/leaderboardList.php
```diff
@@ -245,28 +214,32 @@ function UpdateLeaderboard(lbID) {
         echo "<tr>";
         echo "<td style='width:10%;'>Start:</td>";
         echo "<td>";
-        echo "<input type='text' id='LB_" . $lbID . "_Mem1' value='$memStart' style='width: 100%;' " . ($editAllowed ? "" : "readonly") . "/>";
+        echo "<p>$memStart</p>";
```
Jamiras, line 217: Could probably just remove the definition dumps here, as they're provided more explicitly in the leaderboard page itself:

image

Or at least hide them by default like the Code fields in the Manage Achievements page.

wescopeland: Removed the definitions in latest. Also made a few extra QoL tweaks:

Screenshot 2023-09-17 at 10 19 34 AM

Alternating colors now works based on the leaderboard (now the explicit `<tr>`), and the page does a hard reload after a successful save.

Jamiras: Probably don't need the code notes side panel on this page any longer.

wescopeland: Yep, you're probably right. Removed in https://github.com/RetroAchievements/RAWeb/pull/1873/commits/c9ff1bfbb4cb45a0b175b4d0ca20e79d4e123da8.

Jamiras (approved):
I think the column widths could use some tweaking, but it's not a dealbreaker.

PR #1874, opened 2023-09-16
feat: async double write and metrics aggregation

luchaos:
Adding asynchronous double-write and metrics aggregation for games, player-games, player-sessions, player-achievements while preserving legacy features.

This change adds complexity and some boilerplate overhead but is very much worth maintaining so we can scale the platform better. 

`player_games` and `player_achievement_sets` tables have been updated:
```sh
php artisan migrate
```

> **Note**
> Setting `QUEUE_DRIVER` to `sync`, which is the default locally for new setups, will run the jobs synchronously as the driver name implies.
> Handling jobs synchronously has a major impact on execution time. 
> Handling jobs asynchronously, as it will happen on production, is done by running `./artisan horizon` when `QUEUE_DRIVER` is set to `redis`.
> Tests use the `sync` queue driver.

This implementation makes us of Laravel's Events, Listeners, Jobs, Commands, which eventually call re-usable Actions. 
Events dispatch queued listeners which in turn will dispatch Jobs with specific rate limiting behaviour.
Only in rare cases are Jobs dispatched directly.

To see how events cascade into listeners take a look at `app/Platform/EventServiceProvider`.

`dorequest` is heavily affected by this, thus tests have been updated to take new behaviour into account.

How to test:
- Validate expected data manipulation in database tables: `player_sessions`, `player_games`, `Awarded`, 'player_achievements', 'UserAccounts', 'GameData', `user_activities`, `Activity`
- Requests towards Connect API (`dorequest.php`), ideally end-to-end via emulators
- Reset progress via settings (`controlpanel.php`)
- Run provided commands, run any of the below without arguments and the `-h` option to see usage infos

```sh
./artisan ra:platform:player:unlock-achievement luchaos 1
./artisan ra:platform:player:unlock-achievement luchaos 1,2,3,4,5,6,8,9,15,16,17,18,19,20,21,22,23,24,25,32,38,60,129,130 --hardcore
./artisan ra:platform:player:update-game-metrics luchaos 1,228
./artisan ra:platform:game:update-metrics 1,228
./artisan ra:platform:player:update-game-metrics luchaos --outdated
```

> **Note**
> There are zombie entries in the `player_achievements` table due to progress resets that have not been taken into account yet and only removed unlocks from `Awarded` until now.
> To clean those efficiently I'm looking for help with that so i can focus on tuning the queues and bug-fixing in the meantime. 

Jobs implement various framework-native rate limiting contracts to control scaling.
Adjusting rate-limiting and locking jobs to be unique will be worth looking into. After we gained more insights on how the system behaves running the queues in parallel.
As it's running in parallel, double-writing, this will give us enough time to validate all data and behaviour before doing the switch.

> **Note**
> Jobs only take primitives as input arguments to make sure we do not run into deserialization issues with changes in models in the future.


wescopeland:
Is manual score recalculation no longer possible?

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -104,11 +154,156 @@ public function execute(PlayerGame $playerGame): void
...
+    {
+        $playerAchievements = $beatProgress['playerAchievements'];
+        $isBeatenSoftcore = $beatProgress['isBeatenSoftcore'];
+        $isBeatenHardcore = $beatProgress['isBeatenHardcore'];
+
+        $badge = PlayerBadge::where('User', $user->username)
+            ->where('AwardType', AwardType::GameBeaten)
+            ->where('AwardData', $game->id);
+        $softcoreBadge = $badge->where('AwardDataExtra', UnlockMode::Softcore);
+        $hardcoreBadge = $badge->where('AwardDataExtra', UnlockMode::Softcore);
```
wescopeland, line 228:
```suggestion
        $hardcoreBadge = $badge->where('AwardDataExtra', UnlockMode::Hardcore);
```

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -104,11 +154,156 @@ public function execute(PlayerGame $playerGame): void
...
+            $query = $user->playerBadges()
+                ->where('AwardType', AwardType::Mastery)
+                ->where('AwardData', $game->id);
+
+            if ($query->exists()) {
+                // TODO
+                $user->playerBadges()
+                    ->where('AwardType', AwardType::Mastery)
+                    ->where('AwardData', $game->id)
+                    ->delete();
```
wescopeland, line 213:
```suggestion
            $query = $user->playerBadges()
                ->where('AwardType', AwardType::Mastery)
                ->where('AwardData', $game->id)
                ->delete();
```

Pretty sure we can get rid of the double query and just call `delete()`, and if the record doesn't exist everything will be fine.

app/Platform/Actions/UpdateGameMetricsAction.php
```diff
@@ -4,47 +4,106 @@
...
+        // Note that those might be multiple thousand entries depending on a game's players count
 
-        // $game->save();
+        $updateReason = null;
+        if ($pointsWeightedChange) {
+            $updateReason = 'weighted_points_outdated';
+        }
+        if ($achievementSetVersionChanged) {
+            $updateReason = 'version_mismatch';
+        }
```
wescopeland, line 88: Is it possible that `$pointsWeightedChange` and `$achievementSetVersionChanged` could both be truthy simultaneously?

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -104,11 +154,156 @@ public function execute(PlayerGame $playerGame): void
...
+                    ? $achievement->Date
+                    : $earliestWinConditionDate;
+            }
+        }
 
-        $playerGame->update($metrics);
+        // Return the latest date between the progression and win condition achievements.
+        return $progressionAchievementsUnlocked > 0
+            ? ($latestProgressionDate ? max($latestProgressionDate, $earliestWinConditionDate) : $earliestWinConditionDate)
+            : $earliestWinConditionDate;
```
wescopeland, line 307: In softcore mode, I unlocked a few achievements for a new user: TestMan123. I executed the following commands in order:

```
sail artisan ra:platform:player:unlock-achievement TestMan123 9,16,18,15,17,19
sail artisan ra:platform:player:unlock-achievement TestMan123 79434,79435,79436
sail artisan ra:platform:player:update-game-metrics TestMan123 1,14402
```

The `update-game-metrics` command crashed:

```
   TypeError

  App\Platform\Actions\UpdatePlayerGameMetricsAction::calculateBeatenGameTimestamp(): Return value must be of type string, null returned

  at app/Platform/Actions/UpdatePlayerGameMetricsAction.php:307
    303▕
    304▕         // Return the latest date between the progression and win condition achievements.
    305▕         return $progressionAchievementsUnlocked > 0
    306▕             ? ($latestProgressionDate ? max($latestProgressionDate, $earliestWinConditionDate) : $earliestWinConditionDate)
  ➜ 307▕             : $earliestWinConditionDate;
    308▕     }
    309▕ }
    310▕
```

Dragster (game 14402) does not have any typed achievements (no progression or win condition) in my local database.

wescopeland: After running:

```
sail artisan ra:platform:player:update-game-metrics TestMan123 1
sail artisan ra:platform:game:update-metrics 1`
```

I see TestMan123 as having unlocked every achievement for game 1.

Screenshot 2023-09-16 at 11 32 29 AM

wescopeland:
>`ra:platform:player:update-metrics luchaos --outdated`

`--outdated` doesn't appear to be a valid flag, however the command does execute without it.

app/Platform/Actions/UpdateGameMetricsAction.php
luchaos: Absolutely, `version_mismatch` would take precedence in that case, which is fine.
Then again, i'm not too happy with that area yet - it was the latest addition to the process and there is no entry path yet where all of the outdated player games get added to the queue.

luchaos:
> > `ra:platform:player:update-metrics luchaos --outdated`
> 
> `--outdated` doesn't appear to be a valid flag, however the command does execute without it.

Sloppy edit on my part, that should've been `./artisan ra:platform:player:update-game-metrics luchaos --outdated`. Updated in the description.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
luchaos: That part isn't quite finished yet, I wasn't sure if the else branch actually follows the right condition yet after copying it over.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
Jamiras: I can generate the same error with this command:
```
php artisan ra:platform:player:update-game-metrics MikeyG 11008
```
The game has progression achievements flagged, but the user has not unlocked any of the progression achievements.

Jamiras: Also when resetting all achievements for the game:
image

Jamiras (commented):
> Requests towards Connect API (dorequest.php), ideally end-to-end via emulators

Other than the unlock error, I was able to log in, submit rich presence, upload code notes, achievements, and leaderboards, and submit leaderboard entries.

app/Platform/Actions/UnlockPlayerAchievementAction.php
```diff
@@ -4,126 +4,85 @@
...
+        ?User $unlockedBy = null
+    ): void {
+        $timestamp ??= Carbon::now();
+
+        if (!$achievement->game) {
+            throw new Exception('Achievement does not belong to any game');
+        }
+
+        // make sure to resume the player session which will attach the game to the player, too
+        $achievement->loadMissing('game');
```
Jamiras, line 30:
```
$ php artisan ra:platform:player:unlock-achievement Jamiras 174647,174648,174649,174650 --hardcore
Unlocking 4 [hardcore] Achievements for user [Jamiras]
 0/4 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
   Illuminate\Database\LazyLoadingViolationException 

  Attempted to lazy load [game] on model [App\Platform\Models\Achievement] but lazy loading is disabled.
```

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
Jamiras: I assume it's related to this error, but unlocked achievements are not awarding contribution points or mastery badges.

public/gameInfo.php
```diff
@@ -307,7 +307,7 @@ class='btn break-words whitespace-normal leading-normal'
             && !$isBeatenHardcore
             && !$isBeatenSoftcore
         ) {
-            $beatenGameRetVal = testBeatenGame($gameID, $user, true);
+            $beatenGameRetVal = testBeatenGame($gameID, $user);
```
wescopeland, line 310: One thing I am curious about (sorry if it is answered elsewhere in the PR).

Other than visiting a game page directly, how will the self-healing of beaten games work?

Example:
Game X has 100 achievements. 10 of them are Progression.
50 players have earned _only_ those 10 Progression achievements. Today, they have the beaten game award.
Tomorrow, a developer sets an 11th achievement as Progression.

Currently in V4, as `UpdatePlayerPoints` slowly crawls through the userbase, it will eventually correct everyone's beaten game awards over time, even if they don't visit the game page again or do a manual rescore.

Is there a similar automated mechanism in this PR to capture these cases?

database/migrations/platform/2023_09_15_000000_update_player_games_player_achievement_sets.php
```diff
@@ -0,0 +1,82 @@
...
+    public function down(): void
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+            $table->dropColumn('achievements_beat');
+            $table->dropColumn('beaten_percentage');
+            $table->dropColumn('beaten_percentage_hardcore');
+            $table->dropColumn('beaten_dates');
+            $table->dropColumn('beaten_dates_hardcore');
+            $table->dropColumn('beaten_at');
+            $table->dropColumn('beaten_hardcore_at');
```
wescopeland, line 65:
```suggestion
            $table->dropColumn('achievement_set_version_hash');
            $table->dropColumn('achievements_beat');
            $table->dropColumn('beaten_percentage');
            $table->dropColumn('beaten_percentage_hardcore');
            $table->dropColumn('beaten_dates');
            $table->dropColumn('beaten_dates_hardcore');
            $table->dropColumn('beaten_at');
            $table->dropColumn('beaten_hardcore_at');
            $table->dropColumn('update_status');
```

Added achievement_set_version_hash and update_status

app/Platform/Actions/UnlockPlayerAchievementAction.php
wescopeland: This is being caused by this conditional in _UnlockPlayerAchievement.php_:

```php
if (!$achievement->game) {
    throw new Exception('Achievement does not belong to any game');
}
```

The game relationship is being accessed, but the query to load the game hasn't been executed yet. Beginning from Laravel 8, it appears that lazy loading is disabled by default to prevent N+1 query problems.

There is a `loadMissing()` call which happens right after the conditional which, if placed before the conditional, resolves the problem.

From an architectural standpoint, I am unsure if lazy or eager loading is best within the context of this action. It seems like eager loading could be more beneficial here because we know in advance we're going to access this related model, and eager loading reduces the count of queries made to the database. Given this particular action, it seems like the game relationship is crucial for the action to proceed.

I think we could eager load the game by adding this conditional to the very top of the action:

```php
if (!$achievement->relationLoaded('game')) {
  $achievement->load('game');
}
```

app/Platform/Actions/UpdateGameAchievementsMetrics.php
```diff
@@ -0,0 +1,72 @@
...
+            if ($unlocksHardcoreCalc == 0) { // force all unachieved to be 1
+                $unlocksHardcoreCalc = 1;
+            }
+
+            $weight = 0.4;
+            $pointsWeighted = (int) ($achievementPoints * (1 - $weight)) + ($achievementPoints * (($playersHardcoreCalc / $unlocksHardcoreCalc) * $weight));
+            $pointsWeightedTotal += $pointsWeighted;
+
+            $achievement = Achievement::find($achievementId);
+            $achievement->unlocks_total = 0;
```
wescopeland, line 59: Not sure if still work in progress. This seems to force the softcore unlock count to always be 0.

Jamiras: I reported that earlier: https://github.com/RetroAchievements/RAWeb/pull/1874#discussion_r1328012727

wescopeland: Thanks, I could've sworn you had but wasn't able to find the comment. I'll hide this one in favor of yours.

app/Platform/Actions/UpdateGameAchievementsMetrics.php
```diff
@@ -0,0 +1,72 @@
...
+
+        // TODO refactor to player_achievements
+
+        $achievementsData = legacyDbFetchAll("SELECT ach.ID, ach.Points, SUM(CASE WHEN NOT ua.Untracked THEN aw.HardcoreMode ELSE 0 END) AS UnlocksHardcore
+              FROM Achievements AS ach
+              LEFT JOIN Awarded AS aw ON aw.AchievementID = ach.ID
+              LEFT JOIN UserAccounts AS ua ON ua.User = aw.User
+              WHERE ach.GameID = $game->id
+              AND ach.Flags = " . AchievementFlag::OfficialCore . "
+              GROUP BY ach.ID");
```
wescopeland, line 29: This includes the softcore unlocks and also filters out Untracked users:

```php
$achievementsData = PlayerAchievement::select([
    'Achievements.ID',
    'Achievements.Points',
    DB::raw('COUNT(player_achievements.id) AS Unlocks'),
    DB::raw('SUM(CASE WHEN player_achievements.unlocked_hardcore_at IS NOT NULL THEN 1 ELSE 0 END) AS UnlocksHardcore'),
])
    ->leftJoin('Achievements', 'player_achievements.achievement_id', '=', 'Achievements.ID')
    ->leftJoin('UserAccounts', 'player_achievements.user_id', '=', 'UserAccounts.ID')
    ->where('Achievements.GameID', $gameId)
    ->where('Achievements.Flags', AchievementFlag::OfficialCore)
    ->where('UserAccounts.Untracked', 0)
    ->groupBy('Achievements.ID')
    ->get();
```

In the loop, the only thing that needs to be cast is `$unlocksHardcore`.

```php
$achievementId = $achievementData['ID'];
$achievementPoints = $achievementData['Points'];
$unlocks = $achievementData['Unlocks'];
$unlocksHardcore = intval($achievementData['UnlocksHardcore']);
```

app/Platform/Models/PlayerAchievement.php
```diff
@@ -16,7 +16,7 @@ class PlayerAchievement extends BasePivot
 
     protected $table = 'player_achievements';
 
-    public const CREATED_AT = 'unlocked_at';
+    public const CREATED_AT = null;
```
wescopeland, line 19: I believe this causes `SyncPlayerGames` to throw an error, largely because of this in _config/sync.php_:

```php
'player_games' => [
    // ...
    'reference_key' => PlayerAchievement::CREATED_AT,
    // ...
],
```

Perhaps a good `reference_key` value would instead be the `unlocked_at` column?

luchaos: Good catch! Addressed in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/7f20d681045ef05c76819bc54a86f4731aca9498

app/Platform/Actions/UpdateGameAchievementsMetrics.php
luchaos: Thanks! I went a slightly different route to make it easier to reason about: https://github.com/RetroAchievements/RAWeb/pull/1874/commits/2ee360713a517a5e84962f897b8305bc67ba5e9c

Note: Casing using `(int)` is more efficient than `intval()`.

app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
```diff
@@ -0,0 +1,72 @@
...
+            if ($unlocksHardcoreCalc == 0) { // force all unachieved to be 1
+                $unlocksHardcoreCalc = 1;
+            }
+
+            $weight = 0.4;
+            $pointsWeighted = (int) ($achievementPoints * (1 - $weight)) + ($achievementPoints * (($playersHardcoreCalc / $unlocksHardcoreCalc) * $weight));
+            $pointsWeightedTotal += $pointsWeighted;
+
+            $achievement = Achievement::find($achievementId);
+            $achievement->unlocks_total = 0;
```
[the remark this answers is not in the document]
luchaos: Should be addressed in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/2ee360713a517a5e84962f897b8305bc67ba5e9c

database/migrations/platform/2023_09_15_000000_update_player_games_player_achievement_sets.php
```diff
@@ -0,0 +1,81 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+        });
```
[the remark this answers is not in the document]
luchaos: A leftover, removed in latest

app/Platform/Events/AchievementPublished.php
```diff
@@ -17,7 +17,7 @@ class AchievementPublished
     use SerializesModels;
 
     public function __construct(
-        public Achievement $achievement
+        public Achievement|int|array $achievement
```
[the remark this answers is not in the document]
luchaos: For multiple achievement IDs, will be removed.

app/Platform/Events/AchievementCreated.php
```diff
@@ -17,7 +17,7 @@ class AchievementCreated
     use SerializesModels;
 
     public function __construct(
-        public Achievement $achievement
+        public Achievement|int $achievement
```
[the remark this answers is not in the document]
luchaos: Absolutely, those types will all be reduced to just model classes.
I had to allow for more during development because in some areas where those events are dispatched the respective models were not present yet.
Will be cleaned up in one of the upcoming commits.

app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
```diff
@@ -0,0 +1,72 @@
...
+            $achievement->unlocks_hardcore_total = $unlocksHardcore;
+            $achievement->unlock_percentage = 0;
+            $achievement->unlock_hardcore_percentage = $unlocksHardcore ? $unlocksHardcore / $playersHardcore : 0;
+            $achievement->TrueRatio = $pointsWeighted;
+            $achievement->save();
```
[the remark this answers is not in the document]
luchaos: Updating `updated_at` is fine in these cases as that's when the row itself was updated after all. If we want to be more specific with which set of data was updated when then those should be explicitly new timestamps. E.g. `achievement_set_updated_at`, or `metrics_updated_at`. I don't see a need for those though as most of those updates should be derived from elsewhere, be it an audit log or from a future achievement set model.

public/admin.php
```diff
@@ -70,14 +71,14 @@
...
-                $hardcorePoints = $userPoints['RAPoints'];
-                $softcorePoints = $userPoints['RASoftcorePoints'];
+                dispatch(
+                    new UnlockPlayerAchievementJob(
+                        $user->id,
+                        $nextID,
+                        (bool) $awardAchHardcore,
+                        unlockedByUserId: request()->user()->id
+                    )
+                );
```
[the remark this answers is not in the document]
luchaos: `unlockAchievement()` is marked as deprecated and does not take the new code path for unlocks into account.
Similar to that `UnlockPlayerAchievementJob()` will not touch the then legacy `Awarded` table.

app/Community/Enums/UserActivityType.php
```diff
@@ -16,7 +16,8 @@ abstract class UserActivityType
 
     public const EditAchievement = 'achievement.update';
 
-    public const CompleteGame = 'achievement-set.complete';
+    public const CompleteAchievementSet = 'achievement-set.complete';
+    public const BeatAchievementSet = 'achievement-set.beat';
```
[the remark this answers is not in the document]
luchaos: Good catch, added in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/d4808412aec8224195ad1cecab9495697644a5fa

database/migrations/platform/2023_09_15_000000_update_player_games_player_achievement_sets.php
luchaos: Thanks! Seems i missed a few more columns: https://github.com/RetroAchievements/RAWeb/pull/1874/commits/f340b836f4467a6b838900025e4851647bd7ebef

app/Platform/Actions/UnlockPlayerAchievementAction.php
luchaos: > The game relationship is being accessed, but the query to load the game hasn't been executed yet. Beginning from Laravel 8, it appears that lazy loading is disabled by default to prevent N+1 query problems.

The forced eager loading behaviour is explicitly enabled here: https://github.com/RetroAchievements/RAWeb/blob/master/app/Site/AppServiceProvider.php#L48

`loadMissing()` does pretty much the same as checking for `relationLoaded` manually. 
I moved the `loadMissing()` call above the check.

app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -10,64 +10,74 @@
...
             // if the session hasn't been updated in the last 10 minutes, start a new session
-            if ($now->diffInMinutes($playerSession->updated_at) < 10) {
-                $playerSession->duration = $now->diffInMinutes($playerSession->created_at);
+            if ($timestamp->diffInMinutes($playerSession->updated_at) < 10) {
+                $playerSession->duration = max(1, $timestamp->diffInMinutes($playerSession->created_at));
 
                 if ($presence) {
                     $playerSession->rich_presence = $presence;
-                    $playerSession->rich_presence_updated_at = $now;
+                    $playerSession->rich_presence_updated_at = $timestamp;
```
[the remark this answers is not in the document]
luchaos: Added the date comparison: https://github.com/RetroAchievements/RAWeb/pull/1874/commits/f1cf0039f8c434dad20051cc4fe751c9dc7f336e

Passing the timestamp in this and other actions allows the take a delay into account that occurs when asynchronously processing the job a few seconds later.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
luchaos: `calculateBeatenGameTimestamp()` wasn't fully refactored yet.
Please check for those errors again in latest.

app/Platform/Actions/UpdateDeveloperContributionYieldAction.php
```diff
@@ -0,0 +1,171 @@
...
+                        [
+                            'User' => $username,
+                            'AwardType' => AwardType::AchievementUnlocksYield,
+                            'AwardData' => $countLevel,
+                            'AwardDate' => $unlock->Date,
+                        ],
+                    ],
+                    ['User', 'AwardType', 'AwardData'],
+                    ['AwardDate']
+                );
```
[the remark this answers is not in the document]
luchaos: `UpdateDeveloperContributionYieldAction` is yet to be refactored due to it being executed with each unlock.

The tests assume that the contrib count and yield are updated with each unlock, too.

Would it be enough to calculate contribution at a much later time? E.g. daily for all the affected authors.

Should developers ever lose a badge or should they be permanent?

public/gameInfo.php
luchaos: Technically this is addressed ad-hoc by listening on AchievementPublished events (achievement set version changes) which will mark all affected player_games to be updated.

During player_games metrics updates the badges will be re-evaluated for eligibility, too, which will act accordingly and remove the badge.

How the update of player games based on achievement set version changes will work in detail is still open - most likely it will be a scheduled task that takes care of scheduling all the player games marked for update (update_status) or triggered by a user logging in.

The previous "loops" that were crawling games and players do not scale well, thus I removed them completely. The StaticData table's references to those will be dropped, too, eventually.

app/Platform/Events/AchievementPublished.php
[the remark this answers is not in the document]
luchaos: Cleand up in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/6263982769e2405a1f263d4a187e4c3498542809

app/Platform/Listeners/DispatchUpdateGameMetricsJob.php
```diff
@@ -0,0 +1,55 @@
...
+                break;
+        }
+
+        if (!$game instanceof Game) {
+            if (is_int($game)) {
+                $game = Game::find($game);
+            } elseif ($achievement instanceof Achievement) {
+                $achievement->loadMissing('game');
+                $game = $achievement->game;
+            }
```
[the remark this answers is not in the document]
luchaos: Cleaned up in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/6263982769e2405a1f263d4a187e4c3498542809

app/Platform/Events/AchievementCreated.php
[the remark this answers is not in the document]
luchaos: Cleaned up in https://github.com/RetroAchievements/RAWeb/commit/6263982769e2405a1f263d4a187e4c3498542809

app/Platform/Actions/UpdateGameMetricsAction.php
```diff
@@ -4,47 +4,106 @@
...
+                            ->orWhereNull('achievement_set_version_hash');
+                    }
+                })
+                ->update([
+                    'update_status' => $updateReason,
+                    'achievements_total' => $game->achievements_published,
+                    'points_total' => $game->points_total,
+                    'points_weighted_total' => $game->TotalTruePoints,
+                ]);
+        }
```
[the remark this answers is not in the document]
luchaos: See https://github.com/RetroAchievements/RAWeb/pull/1874#discussion_r1335158595
> How the update of player games based on achievement set version changes will work in detail is still open - most likely it will be a scheduled task that takes care of scheduling all the player games marked for update (update_status) or triggered by a user logging in.

I'm still trying to find a good and scalable way to cascade into player games metrics update which also will make sure that badge eligibility is taken into account.

> // ad-hoc updates for player games, so they can be updated the next time a player updates their profile
> // Note that those might be multiple thousand entries depending on a game's players count

app/Platform/EventServiceProvider.php
```diff
@@ -84,6 +84,7 @@ class EventServiceProvider extends ServiceProvider
             ResumePlayerSession::class,
         ],
         PlayerRankedStatusChanged::class => [
+            // TODO Update all affected games
```
wescopeland, line 87: Good call, I forgot about this situation.

public/gameInfo.php
wescopeland: At the very least, we should be able to easily detect if the award is outdated when visiting a game page. Take a look at this screenshot:

Screenshot 2023-09-24 at 11 46 09 AM

In the screenshot above, the user got the beaten game award, then a developer marked a new achievement as Progression. When this occurs, is it reasonable to fire off an event to update the player's game metrics before the page renders? Otherwise, we end up with situations where clearly I haven't met the beaten criteria, but the page still says I've beaten the game.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
wescopeland: These issues now appear to be resolved for me.

app/Platform/Actions/UnlockPlayerAchievementAction.php
wescopeland: I can confirm this is now working for me.

app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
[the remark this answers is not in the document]
luchaos: Note that Laravel won't update the timestamp if it detects no changes in the model.
This would have to be enforced with the `touch` option.

app/Platform/Actions/UpdateDeveloperContributionYieldAction.php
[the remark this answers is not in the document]
Jamiras: I assume they would be permanent and unchanging. The developer was awarded the 10000 point badge on 12 Dec 2020. If someone reset an achievement older than 12 Dec 2020, it doesn't change then they were awarded the badge, or the fact that they were awarded it in the first place.

app/Platform/Events/AchievementSetCompleted.php
```diff
@@ -19,8 +19,8 @@ class AchievementSetCompleted
 
     public function __construct(
         public User $user,
-        public Game $game,
-        public bool $hardcore
+        public AchievementSet $achievementSet,
+        public ?bool $hardcore = false,
```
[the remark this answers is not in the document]
luchaos: Good catch, I'll change that to non-nullable.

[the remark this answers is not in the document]
luchaos: Changed in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/f0d1b7a054b169a46b266e378cde66e5407bd78b

app/Platform/Actions/RevalidateAchievementSetBadgeEligibility.php
```diff
@@ -0,0 +1,308 @@
...
+    //             PlayerGameCompleted::dispatch($user, $game->id);
+    //
+    //             if ($awardBadge === UnlockMode::Hardcore) {
+    //                 static_addnewhardcoremastery($game->id, $user->username);
+    //             }
+    //         }
+    //
+    //         expireGameTopAchievers($game->id);
+    //     }
+    // }
```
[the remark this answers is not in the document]
luchaos: Deleted commented code with re-introduction of the 6 achievements count minimum for badge eligibility in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/d4e4c0d2536ce8f3a85e76f053a778ab0f670e80

public/gameInfo.php
luchaos: Ideally this should've been caught fairly soon after the new achievement was published or its type changed so it will cascade down into all the players' games metrics to then remove any ineligibly awards.
I'd like to not cause any writes-on-visit side-effects anymore - this stuff has to happen in the back in a timely manner.

What i noticed though is that the type does have an event but no listener yet.
Instead of a listener I think it'd would make sense to make the achievement type part of the achievement version hash 🤔

wescopeland: I agree that side effects on page visit are undesirable.

No strong opinion on the best way to handle this- it does make sense that on a type change some kind of event will cause a cascading update so player's don't really notice this (in some kind of realistic timeframe). We can also update the UI for the progress component to not display it to the player (we have all the info we need on the page to pull off this kind of optimistic visual update).

If we have a strategy for handling the data updates in the back-end, I can take care of the aforementioned UI piece. How do you think we should move forward?

luchaos: I like that - if we know on the game level that there's no beaten award to be had then there's no need to even check for a player's progress.

wescopeland: If you can take care of the back-end, I will open a separate PR this weekend to adjust the UI in that component so the front-end is taken care of.

luchaos: Perfect, thanks! I'm already working on adding the achievement type to the achievement set version hash.

luchaos: Addressed in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/239907663b08f50598a8e8d622455560d052ea15

wescopeland: UI piece should be addressed in https://github.com/RetroAchievements/RAWeb/pull/1900.

luchaos: Fantastic! With the aggregated metrics in place this may become even simpler by just reading from the player's game data directly.

app/Platform/Actions/UpdatePlayerMetrics.php
```diff
@@ -0,0 +1,49 @@
...
+        $user->achievements_unlocked_hardcore = $playerGames->sum('achievements_unlocked_hardcore');
+        $user->completion_percentage_average = $playerGames->average('completion_percentage');
+        $user->completion_percentage_average_hardcore = $playerGames->average('completion_percentage_hardcore');
+
+        // TODO refactor to use aggregated player_games metrics
+        $user->RAPoints = $user->achievements()->published()->wherePivotNotNull('unlocked_hardcore_at')->sum('Points');
+        $user->RASoftcorePoints = $user->achievements()->published()->wherePivotNull('unlocked_hardcore_at')->sum('Points');
+        $user->TrueRAPoints = $user->achievements()->published()->wherePivotNotNull('unlocked_hardcore_at')->sum('TrueRatio');
+
+        // TODO refactor to use the above implementation only
```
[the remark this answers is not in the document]
luchaos: Removed the whole block in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/a9552814b9248b7cb6b2e8319c9701f3dce3ac25

I wasn't sure if we'd risk running into temporary points/rank jumps when basing the calculation on player_achievements because of known zombie entries.
As cleaning up those zombie entries before this code even hits production this shouldn't be an issue eventually.

public/gameInfo.php
wescopeland: I've merged https://github.com/RetroAchievements/RAWeb/pull/1900, which removes the potential `testBeatenGame()` call on page visit when aggregate queries are enabled and updates the progress status label on an optimistic basis. I think it's safe to mark this feedback as resolved.

app/Platform/Actions/UpdateDeveloperContributionYieldAction.php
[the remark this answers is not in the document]
luchaos: Disabled developer contribution calculations completely in https://github.com/RetroAchievements/RAWeb/pull/1874/commits/cb88845929454499af148be5fd4efae24fe94374

Let's revisit this feature when the rest is done?

wescopeland (approved):

Jamiras (approved):

PR #1875, opened 2023-09-16
Highlight top searchbox to make label more visible and with transition

Krylan:
Changing styles to make searchbox more visible, so input will be easier to notice and click when starting search, and also adding a transition on input's width to make it cool when focusing on input.

**Before**
image
**After**
image


wescopeland:
Recommend setting the focus ring to `:focus-visible`. I see it on every click in the Chromium browser I'm using.

Screenshot 2023-09-16 at 8 34 51 AM

Screenshot 2023-09-16 at 8 34 56 AM

resources/css/search.css
```diff
@@ -9,8 +9,13 @@
 }
 
 .searchbox-top {
-  background-color: var(--embed-color);
-  border-radius: 4px;
+  height: 22px;
```
wescopeland, line 12: 22px feels a bit cramped. We can probably up this to 30px without impacting the navbar height.

wescopeland:
Recommend if the placeholder is not visible (`:not(:placeholder-shown)`), to keep the searchbox expanded. Otherwise, it contracts when landing on the search page:

Screenshot 2023-09-16 at 8 52 00 AM

Krylan:
> Recommend setting the focus ring to `:focus-visible`. I see it on every click in the Chromium browser I'm using.
> 
> Screenshot 2023-09-16 at 8 34 51 AM
> 
> Screenshot 2023-09-16 at 8 34 56 AM

@wescopeland I'm afraid it has no effect on text inputs, as browsers trigger `:focus-visible` regardless of the way, how these were focused. I checked that and in Chrome it still shows outline. Discussions on the Internet seems to confirm this behaviour as expected and along with recommendations:
https://github.com/WICG/focus-visible/issues/131
https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo

So solution there would be leaving it this way, removing outline from this element (wrong for accessibility) or checking focus by JS to detect how it was triggered (a bit overwhelming for such an issue).

I agree with the other remarks and will include changes in next commit.

wescopeland:
Yep, looks like :focus-visible is not going to be helpful here and we'd need to pull this off using JS. Let's not worry about it for now then.

wescopeland (approved):

PR #1877, opened 2023-09-17
fix(userInfo): dont render Progression Status if only event games

wescopeland:
Fixes an edge case on user profiles where the Progression Status can component will render if the user has only games for the "Events" console.

This notably happens on WilHiteWarrior's user profile.

**Before**
Screenshot 2023-09-17 at 12 23 07 PM

**After**
Screenshot 2023-09-17 at 12 21 12 PM

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -480,9 +480,13 @@ RenderContentStart($userPage);
         echo "</div>"; // devbox
     }
 
-    // The component isn't as useful if we don't have data for the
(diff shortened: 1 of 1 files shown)


Jamiras:
Oddly enough, I see 1 PS2 entry for WilHiteWarrior with `AGGREGATE_QUERIES` off, and 390 games with `AGGREGATE_QEURIES` on.
image

Jamiras (approved):

PR #1881, opened 2023-09-20
refactor(reportissue): migrate from Knockout.js to Alpine

wescopeland:
This PR migrates _reportissue.php_'s client-side code from Knockout.js to Alpine. Once this and https://github.com/RetroAchievements/RAWeb/pull/1876 are merged, there will be nothing left in the project depending upon Knockout.js and it can be removed as a dependency.

This is a 1:1 conversion with no feature/function changes.

**How to test this PR**
Visit localhost/reportissue.php?i=116522.

* In the Description field, try typing the string "manual unlock". You should see some helper text appear.
* In the Description field, try typing the string "didn't work". Helper text should appear and the submit button should be disabled.
* In the Emulator field, select "Bizhawk". The "Core" field should vanish.


public/reportissue.php
```diff
@@ -181,43 +222,40 @@ function displayCore() {
...
+
+                    <p x-show="descriptionIsNetworkProblem">
+                        Please do not use this tool for network issues.
+                        See <a href='https://docs.retroachievements.org/FAQ/#how-can-i-get-credit-for-an-achievement-i-earned-but-wasnt-awarded'>here</a>
+                        for instructions on how to request a manual unlock.
+                    </p>
+                    <p x-show="descriptionIsUnhelpful">
+                        Please be more specific with your issue&mdash;such as by adding specific reproduction steps or what you 
+                        did before encountering it&mdash;instead of simply stating that it doesn't work. The more specific, the better.
+                    </p>
```
Jamiras, line 247: This text is briefly visible when loading (or reloading) the page.
image

public/reportissue.php
```diff
@@ -181,43 +222,40 @@ function displayCore() {
...
-                    <p data-bind="visible: descriptionIsNetworkProblem">Please do not use this tool for network issues. See <a href='https://docs.retroachievements.org/FAQ/#how-can-i-get-credit-for-an-achievement-i-earned-but-wasnt-awarded'>here</a> for instructions on how to request a manual unlock.</p>
-                    <p data-bind="visible: descriptionIsUnhelpful">Please be more specific with your issue&mdash;such as by adding specific reproduction steps or what you did before encountering it&mdash;instead of simply stating that it doesn't work. The more specific, the better.</p>
+                    <textarea
+                        class="w-full forum"
+                        name="description"
+                        id="description"
+                        style="height:160px"
+                        rows="5"
+                        cols="61"
+                        placeholder="Describe your issue here..."
```
Jamiras, line 232: The cursor acts like the watermark is there, but it's not visible
image
image

Jamiras: And typing doesn't clear it out either. I can click in the middle of the placeholder text and add some letters.
image

The submitted issue captures the behavior as whitespace:
image

Jamiras: And it seems to interfere with the regex matching:
image

image

public/reportissue.php
wescopeland: Thanks, I didn't see this with a local dev build. Resolved in latest with the addition of an `x-cloak` directive.

public/reportissue.php
wescopeland: Resolved in latest. This turned out to be a rendering issue with Blade templates.

Note: The "no work" matching does cut off at 25 characters.

Jamiras (approved):

PR #1884, opened 2023-09-25
refactor(game): migrate recent game players to blade and improve mobile ux

wescopeland:
This PR migrates the game page's "Recent Players" section to Blade and improves the mobile UX for the component.

**Mobile Before**
Screenshot 2023-09-25 at 5 09 42 PM

**Mobile After**
Screenshot 2023-09-25 at 5 09 14 PM

**Desktop Before**
Screenshot 2023-09-25 at 5 09 50 PM

**Desktop After**
Screenshot 2023-09-25 at 5 08 53 PM

---

UX Notes:
* "Recent Players" is now an H2 tag similar to the "Achievements" heading higher up on the page. I wasn't sure if this was perhaps too loud. If we think it's too loud, let's reduce the size but give the tag a `role` attribute with a "heading" value and an `aria-level` of "2".

### DIFF
--- a/resources/views/platform/components/game/recent-game-players.blade.php
+++ b/resources/views/platform/components/game/recent-game-players.blade.php
@@ -1,0 +1,73 @@
+@props([
+    'recentPlayerData' => [],
+    'gameTitle' => '',
+])
+
+<?php
+use Illuminate\Support\Carbon;
+
+$processedRecentPlayers = [];
+
+foreach ($recentPlayerData as $recentPlayer) {
+    $date = Carbon::parse($recentPlayer['Date'])->format('d M Y, g:ia');
+    $isBroken = mb_strpos($recentPlayer['Activity'], 'Unknown macro') !== false;
+
+    $processedRecentPlayers[] = [
+        'User' => $recentPlayer['User'],
+        'Date' => $date,
+        'IsBroken' => $isBroken,
+        'Activity' => $recentPlayer['Activity'],
+    ];
+}
+?>
+
+<h2 class="text-h4">Recent Players</h2>
+
+<div class="sm:hidden flex flex-col gap-y-2">
+    @foreach ($processedRecentPlayers as $recentPlayer)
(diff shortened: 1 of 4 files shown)


resources/views/platform/components/game/recent-game-players.blade.php
```diff
@@ -0,0 +1,73 @@
...
+        <div class="flex flex-col gap-y-0.5 px-2 py-1.5 odd:bg-embed">
+            <div class="w-full flex items-center justify-between">
+                {!! userAvatar($recentPlayer['User'], iconClass: 'rounded-sm', iconSize: 20) !!}
+                <p class="smalldate">{{ $recentPlayer['Date'] }}</p>
+            </div>
+
+            @if ($recentPlayer['IsBroken'])
+                <div class="cursor-help text-xs" title="{{ $recentPlayer['Activity'] }}">
+                    <span>⚠️</span>
+                    <span>Playing {{ $gameTitle }}</span>
```
Jamiras, line 37: The font stride seems to be slightly different when the warning symbol is present.
image

wescopeland: Good catch. Resolved in https://github.com/RetroAchievements/RAWeb/pull/1884/commits/86bfd858e915d8badc6cff1d680d04b21593acd9.

Jamiras: You changed the second table below. Does the same change need to be applied here? Why are there two tables?

wescopeland: The mobile view doesn't use a table layout. It uses cards with a flex layout. SM and above uses a `<table>` tag. Due to the restrictions of HTML and CSS, it isn't possible to have a single component operate as both a flex and table layout based on the breakpoint.

The mobile view uses `text-xs` for all string content. The table does not, and this was the source of the bug you observed.

Jamiras (approved):

PR #1896, opened 2023-09-30
fix(game): grant Jr Devs access to Developer Interest page

wescopeland:
This PR adjusts the visibility of the "View Developer Interest" link on the game page devbox such that it is never visible to Junior Developers.

Currently, the URL is guarded from Jr Devs, so it's causing some confusion that the link itself is appearing for them:
https://github.com/RetroAchievements/RAWeb/blob/a3be5f7b4260fe4128f8f721144b6ccfbe161aca/app/Platform/Controllers/GameDevInterestController.php#L23-L25


Jamiras:
I don't see why a Jr. Dev with a claim shouldn't be allowed to see other people who might be interested in dev'ing the game. The link is made available to Dev claimants per the following:

> This gives the claimant insight into other devs that they could bounce ideas off of, or inquire about collaborating.

Are there concerns that a Jr. Dev would harass other developers interested in the title?

It would probably work to just remove the Developer check you've highlighted, and let the check on line 33 deal with it:
https://github.com/RetroAchievements/RAWeb/blob/a3be5f7b4260fe4128f8f721144b6ccfbe161aca/app/Platform/Controllers/GameDevInterestController.php#L33-L35

wescopeland:
No concern on my end - I assumed Jr Devs being able to even see the link was a bug given they couldn't access the page the link took them to.

I've reverted the _gameInfo.php_ change and removed the conditional I highlighted in the OP.

suspect15:
No concern about Jr Devs seeing additional interest from the CR team.  Per a recent vote, we are not permitting Jr Devs to collab  until after the mentor program is stood up due to significant issues with recent collabs, but there's no reason they shouldn't be able to see developer interest.

app/Platform/Controllers/GameDevInterestController.php
```diff
@@ -20,9 +20,6 @@ public function __invoke(Request $request): View
         }
 
         $permissions = $request->user()->getAttribute('Permissions');
-        if ($permissions < Permissions::Developer) {
```
Jamiras, line 23: This should still restrict access to JuniorDevelopers or non-developers could access the page directly via a URL.

wescopeland: Good catch, resolved in https://github.com/RetroAchievements/RAWeb/pull/1896/commits/bd7b2d4bc260bca71729ba8d7c4dd2967669b613.

Jamiras (approved):

PR #1897, opened 2023-09-30
fix(renderAchievementTitle): remediate blade rendering with special chars

wescopeland:
This PR resolves this issue:

image

**Root Cause**
When using `renderAchievementTitle()` in a Blade component, the title must be rendered using the `{!! !!}` expression. This syntax tells Blade to output the data as raw HTML without escaping it. When special characters are included in an expression like this, those special characters are interpreted as raw HTML.

We resolve this by using `htmlspecialchars()` to escape just the `$title` string but not the missable tag.

**How to test this PR:**
Change an achievement title to:
* Sam the >!</# Flame

### DIFF
--- a/app/Helpers/render/achievement.php
+++ b/app/Helpers/render/achievement.php
@@ -70,7 +70,7 @@ function renderAchievementTitle(?string $title, bool $tags = true): string
         return '';
     }
     if (!Str::contains($title, '[m]')) {
-        return $title;
+        return htmlspecialchars($title, ENT_QUOTES, 'UTF-8');
     }
 
     $missableTag = '';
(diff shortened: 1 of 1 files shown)


app/Helpers/render/achievement.php
```diff
@@ -70,7 +70,7 @@ function renderAchievementTitle(?string $title, bool $tags = true): string
         return '';
     }
     if (!Str::contains($title, '[m]')) {
-        return $title;
+        return htmlspecialchars($title, ENT_QUOTES, 'UTF-8');
```
Jamiras, line 73: This breaks the display on the achievement details page:
image

image

PR #1898, opened 2023-09-30
chore: migrations cleanup to match production schema more closely

luchaos:
In preparation for better seeders and to make sure the migrations are in a clean state moving forward in general.

I ran into at least two issues that i thought were bugs when it was the migrations being out of sync.


Jamiras (approved):
I didn't validate the changes against the production schema, but I did reset my database with them and the seeded site seems to be at least smoketest-functional.

PR #1900, opened 2023-09-30
fix(game): show beaten progress on an optimistic basis

wescopeland:
This PR adjusts how the progress component on game pages renders Beaten and Beaten (softcore) statuses.

Rather than explicitly checking for the award, the status is rendered now on an optimistic basis dependent upon the user's current unlocks for the game. This better facilitates self-healing of beaten game awards when async metrics aggregation is used.


public/gameInfo.php
```diff
@@ -234,20 +244,29 @@
...
+
+        $isBeatenHardcore = (
+            $totalEarnedProgressionHardcore === $totalProgressionAchievements
+            && $totalEarnedWinConditionHardcore >= $neededWinConditions
+        );
+
+        // TODO: Remove this side effect when switching to aggregate queries.
+        // Without aggregate queries, the side effect is part of the beaten games
+        // self-healing mechanism.
+        if (!config('feature.aggregate_queries')) {
```
Jamiras, line 266: You could move this up to line 249. There's no reason to calculate `$isBeatenSoftcore` or `$isBeatenHardcore` unless it goes into this if.

Jamiras (approved):
> Rather than explicitly checking for the award, the status is rendered now on an optimistic basis dependent upon the user's current unlocks for the game. 

It seems like it was doing that before.

The only changes I see here are calling `testBeatenGame` is dependent on the aggregate queries feature being disabled, and it will still check for a hardcore beaten badge if the softcore beaten badge was awarded. Otherwise, the functionality appears the same.

public/gameInfo.php
wescopeland: I'm not sure I follow - the optimistic calculation without the `testBeatenGame()` call is the purpose of the PR.

Previously, these values were being set from either the SiteAwards table or based on the return value of `testBeatenGame()`.

Based on [this conversation](https://github.com/RetroAchievements/RAWeb/pull/1874#discussion_r1335158595) in #1874, there is a desire to remove the `testBeatenGame()` side effect from the game page entirely.

However, due to the upcoming async nature of the award updates in the back-end, we can't guarantee the user won't erroneously see a status label for a beaten award they should or shouldn't have. By deriving the award status optimistically through the achievement data already loaded for the page, we can mitigate such scenarios.

Jamiras: > I'm not sure I follow - the optimistic calculation without the testBeatenGame() call is the purpose of the PR.

Logic before:
```
        if (
            $totalEarnedProgression === $neededProgressions
            && $totalEarnedWinCondition >= $neededWinConditions
            && !$isBeatenHardcore
            && !$isBeatenSoftcore
        ) {
            $beatenGameRetVal = testBeatenGame($gameID, $user, true);
```

It doesn't look like the logic that gates the call to `testBeatenGame` has changed. The first two lines become `$isBeatenSoftcore`, and the last line is the logical comparison that the award hasn't been previously given.

New logic:
```
        $isBeatenSoftcore = (
            $totalEarnedProgression === $totalProgressionAchievements
            && $totalEarnedWinCondition >= $neededWinConditions
        );
        
        ...
        
        if ($isBeatenSoftcore !== $hasBeatenSoftcoreAward || $isBeatenHardcore !== $hasBeatenHardcoreAward) {
            $beatenGameRetVal = testBeatenGame($gameID, $user, true);
        }
````

Jamiras: > However, if you then changed a 9th achievement to be progression, you would continue to see the label even though you hadn't earned that 9th progression achievement. After the changes in this PR, that is no longer the case.

I don't think that would have worked before either as this would not have been true: `$totalEarnedProgression === $neededProgressions`

wescopeland: > I don't think that would have worked before either as this would not have been true:
`$totalEarnedProgression === $neededProgressions`

You're right, it didn't work before either. It does now with this change.

**Before**
In this screenshot, an achievement ("Pro Collector", last on the list in the image) has been given a Progression type _after_ I've earned the award. I still visually see the "Beaten" status text, even though I shouldn't.

Screenshot 2023-09-30 at 7 34 25 PM

**After**
Changing nothing else besides switching from `master` to this branch, the label now says "Unfinished", even though I technically still have the award in the database. Because of `$isBeatenHardcore` and `$isBeatenSoftcore` being optimistically adjusted, the erroneous label no longer displays. Eventually, the awards data in the database will catch up.

Screenshot 2023-09-30 at 7 36 20 PM

PR #1906, opened 2023-10-08
add tests for dorequest?r=uploadachievement

Jamiras:


app/Platform/EventServiceProvider.php
```diff
@@ -33,20 +34,21 @@ class EventServiceProvider extends ServiceProvider
 {
     protected $listen = [
         AchievementCreated::class => [
+            DispatchUpdateGameMetricsJob::class, // dispatches GameMetricsUpdated
         ],
         AchievementPublished::class => [
             DispatchUpdateGameMetricsJob::class, // dispatches GameMetricsUpdated
-            // TODO DispatchUpdateDeveloperContributionYieldJob::class,
```
Jamiras, line 39: Is there some reason this was TODO'd?

wescopeland: Probably from line 1 in this comment: https://discord.com/channels/476211979464343552/1026595325038833725/1159436046837223515

app/Helpers/database/achievement.php
```diff
@@ -226,67 +221,84 @@ function UploadNewAchievement(
...
+        addArticleComment(
+            "Server",
+            ArticleType::Achievement,
+            $idInOut,
+            "$author uploaded this achievement.",
+            $author
+        );
+
+        // uploaded new achievement
+        AchievementCreated::dispatch($achievement);
```
luchaos, line 249: Would it make sense to dispatch AchievementPublished if the flag matches? This way it would already cascade into the games metrics update.
Not sure if we actually need AchievementCreated for that.

app/Platform/EventServiceProvider.php
luchaos: https://github.com/RetroAchievements/RAWeb/pull/1874#discussion_r1347230085

tests/Feature/Connect/UploadAchievementTest.php
```diff
@@ -0,0 +1,895 @@
...
+        $this->assertEquals($game->points_total, 10);
+
+        // ====================================================
+        // unlock achievement; contrib yield changes
+        $this->addHardcoreUnlock($author, $achievement2);
+        $this->addHardcoreUnlock($this->user, $achievement2);
+
+        $author->refresh();
+        $this->assertEquals($author->ContribCount, 1);
+        $this->assertEquals($author->ContribYield, 10);
```
luchaos, line 141: I'd skip contrib recalculation for now until we have a scalable solution for those in place.
https://github.com/RetroAchievements/RAWeb/pull/1874#discussion_r1347230085

app/Helpers/database/achievement.php
luchaos: nevermind - makes sense for `achievements_unpublished` as you mentioned

Jamiras: Achievements created in the DLL are always uploaded to Unofficial, then promoted from there. There's no way (currently) to upload directly to Core. As such, `AchievementPublished` seems incorrect.

app/Platform/EventServiceProvider.php
Jamiras: Ok. I'll put these back and comment out the related assertions.

app/Helpers/database/achievement.php
luchaos: perfect, thanks for the clarification!

luchaos (approved):

PR #1908, opened 2023-10-08
feat: update player games in batches after game metrics updates

luchaos:
Allow game metrics updates to cascade into player games updates in batches.

Player games batches are processed on a separate queue and will be worked on with a lower priority than the rest.

Batches of jobs will not be created when using the `sync` queue driver and/or for tests.

Run `php artisan migrate` to create the batches table.

```
ID      Title                               Players
228     Super Mario World                   49556
10003   Super Mario 64                      31605
337     Donkey Kong Country                 31209
1446    Super Mario Bros.                   28812
1       Sonic the Hedgehog                  26651
1995    Super Mario Bros. 3                 24234
355     Legend of Zelda, The                22750
232     Super Mario Kart                    20965
253     Super Mario All-Stars               20600
11240   Castlevania: Symphony of the Night  18802
236     Super Metroid                       18681
319     Chrono Trigger                      18309
637     Mega Man X                          17943
1462    Castlevania                         17085
379     Aladdin                             16976
```


config/queue.php
```diff
@@ -69,6 +69,22 @@
...
+    | batching information. These options can be updated to any database
+    | connection and table which has been defined by your application.
+    |
+    */
+
+    'batching' => [
+        'database' => env('DB_CONNECTION', 'mysql'),
+        'table' => 'queue_job_batches',
+    ],
+
```
luchaos, line 87: taken from the Laravel 10 skeleton - it's been configurable since Laravel 8.

config/horizon.php
```diff
@@ -194,9 +194,10 @@
                 'default',
                 'player-achievements',
                 'player-game-metrics',
-                'game-metrics',
                 'player-metrics',
+                'game-metrics',
                 'developer-metrics',
+                'player-game-metrics-batch',
```
luchaos, line 200: Lowest priority for `player-game-metrics-batch`

app/Platform/Actions/UpdateGameMetrics.php
```diff
@@ -81,23 +84,34 @@ public function execute(Game $game): void
...
+                ->chunk(1000, function (Collection $chunk) {
+                    // map and dispatch this chunk as a batch of jobs
+                    Bus::batch(
+                        $chunk->map(
+                            fn (PlayerGame $playerGame) => new UpdatePlayerGameMetricsJob($playerGame->user_id, $playerGame->game_id)
+                        )
+                    )
+                        ->onQueue('player-game-metrics-batch')
+                        ->dispatch();
+                });
```
Jamiras, line 109: I'm a bit concerned that `UpdateGameMetrics` updates both the achievement set metrics (`achievements_published`, `achievements_unpublished`, `points_total`) and set player metrics (`players_total`, `players_hardcore`).

Because of this dual-purposing, it gets fired when the achievement set changes (fine) or when an achievement is unlocked (potentially very frequently for these common games):
https://github.com/RetroAchievements/RAWeb/blob/8d9eff006f1fafea8cd7f1977df54d1b808d0d8e/app/Platform/EventServiceProvider.php#L58-L61

If I understand the logic here, every time any user unlocks an achievement in Super Mario World, the entire Super Mario World userbase will be queued up to have their game metrics re-evaluated. Even if the filter on the hash changing prevents that, we shouldn't have to generate that hash and query the player_games table to see if it's been changed just when some random user unlocks an achievement that won't have any impact on anything other than possibly the total number of players in the set.

I think we should separate the player count stuff out of this action and put it in a separate one that doesn't generate the achievement_set_hash or attempt to batch up changes. Maybe `UpdateGamePlayerMetrics`?

Jamiras (commented):
I have not tested these changes, but they seem reasonable. Please see my note about splitting the `UpdateGameMetricsAction`.

app/Platform/Actions/UpdateGameMetrics.php
luchaos: valid concerns, i had similar thoughts. it's very hard to tell how it scales without running it on the server. not all of the unlocks will cascade into a game metrics update - it can only ever be queued once.

do i understand correctly, `UpdateGamePlayerMetrics` would only include the affected player games batch updates after the achievement set version changed?
i think returning early after `GameMetricsUpdated::dispatch()` in `UpdateGameMetrics` when there was no achievement set version change would result in the same behaviour for now?

keeping it everything together makes it easier to reason about but there's definitely room for improvement. especially since the batching is very greedy in itself and could be much smarter by checking what actually changed. 
personally, i'd like to see it in action before doing too much optimization or splitting that might not be necessary.

Jamiras: I'm just thinking that these really popular games are almost always going to have _someone_ playing them. As such, they're constantly going to be calculating the hash and scanning the 40k `player_games` rows looking for hash mismatches. Since we can pretty easily associate the hash to specific changes to the achievements (which happen very infrequently), we should avoid that overhead as much as possible.

> UpdateGamePlayerMetrics would only include the affected player games batch updates after the achievement set version changed?

Yes, leave this action for changes to the achievement set. And add a new action for changes to the players of the set. Changing the set would also update the players as the player count is dependent on the "exposed" achievements.  But the much more frequent unlocking of achievements (and the inverse - resetting of achievements) would be streamlined to just update the player counts.

It could be even further optimized to only be invoked on unlocking the first achievement or resetting the last achievement - though that doesn't play into the automatic mapping of the unlock event to the metrics update through the listener.

> i think returning early after GameMetricsUpdated::dispatch() in UpdateGameMetrics when there was no achievement set version change would result in the same behaviour for now?

That would avoid scanning the 40k rows, but it would still have to generate the hash every time an achievement was unlocked in the set just to see if it changed.

> not all of the unlocks will cascade into a game metrics update - it can only ever be queued once.

Yes, but if a large number of people are playing a popular game at the same time and unlocking a reasonable number of achievements, the game metrics update will be reinserted back into the queue almost as quickly as it cycles out.

As you've said, it would be best to see exactly how things perform in production, but I don't think we can simulate the load the site actually sees.

app/Platform/Jobs/UpdatePlayerMetricsJob.php
```diff
@@ -25,6 +27,17 @@ public function __construct(
...
     public function handle(): void
     {
+        if ($this->batch()?->cancelled()) {
+            return;
+        }
+
+        $user = User::find($this->userId);
+
+        if (!$user) {
+            // might've been deleted
```
wescopeland, line 37: nit: It may be worth logging this.

app/Console/Kernel.php
```diff
@@ -15,6 +15,7 @@ protected function schedule(Schedule $schedule)
         // $schedule->command('websockets:clean')->daily();
 
         $schedule->command('horizon:snapshot')->everyFiveMinutes();
+        $schedule->command('queue:prune-batches --hours=48 --unfinished=72 --cancelled=72')->daily();
```
wescopeland, line 18: nit: It may be worth adding `onFailure()` logic here for diagnostics purposes just in case something goes awry.

luchaos: I'm not sure we're that interested in failures of the cleanup command 🤔 
it's a default framework command that is expected to work as documented.
am i missing something?

app/Platform/Jobs/UpdatePlayerMetricsJob.php
luchaos: ideally we should not ever run into this - this check is to be removed, too.
the only reason i added it is because the player_games table is not yet properly cleaned up after account deletions. i have a draft PR up for that.

Jamiras (approved):

luchaos:
Went ahead and merged as i need this on stage before enabling it again to prevent unexpected behaviour.
We can revisit the open items in follow-up PRs, I hope that's good enough for now.

PR #1911, opened 2023-10-12
fix: achievements weighted points (TrueRatio) metrics calculation

luchaos:
Add tests for metrics while at it.


app/Platform/Actions/UpdateGameAchievementsMetrics.php
```diff
@@ -34,7 +34,10 @@ public function execute(Game $game): void
             // force all unachieved to be 1
             $unlocksHardcoreCalc = $unlocksHardcoreCount ?: 1;
             $weight = 0.4;
-            $pointsWeighted = (int) ($achievement->points * (1 - $weight)) + ($achievement->points * (($playersHardcoreCalc / $unlocksHardcoreCalc) * $weight));
+            $pointsWeighted = (int) (
+                $achievement->points * (1 - $weight)
+                + $achievement->points * (($playersHardcoreCalc / $unlocksHardcoreCalc) * $weight)
+            );
```
wescopeland, line 40: nit: I can't help but feel that `$weight` is a magic number, so perhaps it would make sense to encapsulate this work into a helper method, eg:

```php
private function calculateWeightedPoints($points, $playersHardcoreCalc, $unlocksHardcoreCalc): int {
    // 0.4 means 40% weight is given to hardcore unlocks.
    const WEIGHT_FACTOR = 0.4;

    return (int) (
        $points * (1 - WEIGHT_FACTOR)
        + $points * (($playersHardcoreCalc / $unlocksHardcoreCalc) * WEIGHT_FACTOR)
    );
}
```

wescopeland (approved):

Jamiras (approved):

app/Platform/Actions/UpdateGameAchievementsMetrics.php
luchaos: I was thinking about that but I'd like to keep it together for now for brevity. It doesn't have to be reused yet, maybe in the future?

wescopeland: No worries at all! I was borderline on this too so I went ahead and approved after writing the comment. Marking as resolved.

PR #1912, opened 2023-10-13
allow aggregate data use on global ranking page

Jamiras:


app/Platform/Actions/UpdatePlayerGameMetrics.php
```diff
@@ -37,7 +37,7 @@ public function execute(PlayerGame $playerGame, bool $silent = false): void
 
         $points = $achievementsUnlocked->sum('Points');
         $pointsHardcore = $achievementsUnlockedHardcore->sum('Points');
-        $pointsWeighted = $achievementsUnlocked->sum('TrueRatio');
+        $pointsWeighted = $achievementsUnlockedHardcore->sum('TrueRatio');
```
Jamiras, line 40: fixes `points_weighted` calculation for `player_games`

app/Platform/Actions/UpdatePlayerMetrics.php
```diff
@@ -11,16 +11,18 @@ class UpdatePlayerMetrics
 {
     public function execute(User $user): void
     {
-        $playerGames = $user->playerGames()->where('achievements_unlocked', '>', 0);
+        $playerGames = $user->playerGames()
+            ->join('GameData', 'GameData.ID', '=', 'player_games.game_id')
+            ->whereNotIn('GameData.ConsoleID', [100, 101]) // ignore events and hubs
```
Jamiras, line 16: exclude events and hub achievements from player achievement counts

app/Platform/Actions/UpdatePlayerMetrics.php
```diff
@@ -11,16 +11,18 @@ class UpdatePlayerMetrics
...
         $user->completion_percentage_average = $playerGames->average('completion_percentage');
         $user->completion_percentage_average_hardcore = $playerGames->average('completion_percentage_hardcore');
 
-        // TODO refactor to use aggregated player_games metrics
-        $user->RAPoints = $user->achievements()->published()->wherePivotNotNull('unlocked_hardcore_at')->sum('Points');
-        $user->RASoftcorePoints = $user->achievements()->published()->wherePivotNull('unlocked_hardcore_at')->sum('Points');
-        $user->TrueRAPoints = $user->achievements()->published()->wherePivotNotNull('unlocked_hardcore_at')->sum('TrueRatio');
+        $user->RAPoints = $playerGames->sum('points_hardcore');
+        $user->RASoftcorePoints = $playerGames->sum('points') - $user->RAPoints;
+        $user->TrueRAPoints = $playerGames->sum('points_weighted');
```
luchaos, line 25: Will this PR follow in a later release? otherwise we'll have to wrap this and the replaced code in the aggregated queries feature flag

app/Platform/Actions/UpdatePlayerGameMetrics.php
luchaos: Good catch!

app/Platform/Actions/UpdatePlayerMetrics.php
Jamiras: I hadn't really considered the fact that these values are currently being used. This cannot be called until player_games is fully populated. I'll wrap it in the feature flag and restore the old code as the default path.

app/Platform/Commands/UpdatePlayerGameMetrics.php
```diff
@@ -49,11 +51,19 @@ public function handle(): void
         $progressBar = $this->output->createProgressBar($playerGames->count());
         $progressBar->start();
 
+        // if running in sync mode, just call updatePlayerMetrics once manually after updating
+        // all of the player_games instead of letting it cascade for every player_game updated.
+        $isSync = (config('queue.default') === 'sync');
+
```
luchaos, line 57: Very nice, I forgot to add this to the command, too.

luchaos (approved):

PR #1918, opened 2023-10-15
use aggregate_queries feature for high scores on game page

Jamiras:
Takes ~250ms to build the high scores list for Super Mario World vs 2.5s with the feature disabled.

250ms is still a reasonably large amount of time, so I recommend we continue to cache the result (line 733 above).

### DIFF
--- a/app/Helpers/database/player-game.php
+++ b/app/Helpers/database/player-game.php
@@ -751,7 +751,7 @@ function getGameTopAchievers(int $gameID): array
         $query = "SELECT ua.User, pg.achievements_unlocked_hardcore AS NumAchievements,
                          pg.points_hardcore AS TotalScore, pg.last_unlock_hardcore_at AS LastAward
                     FROM player_games pg
-                    LEFT JOIN UserAccounts ua ON ua.ID = pg.user_id
+                    INNER JOIN UserAccounts ua ON ua.ID = pg.user_id
                     WHERE ua.Untracked = 0
                     AND pg.game_id = $gameID
                     AND pg.achievements_unlocked_hardcore > 0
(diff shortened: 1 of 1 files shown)


app/Helpers/database/player-game.php
```diff
@@ -747,18 +747,28 @@ function getGameTopAchievers(int $gameID): array
...
-                  AND gd.ID = $gameID
-                  AND aw.HardcoreMode = " . UnlockMode::Hardcore . "
-                GROUP BY aw.User
-                ORDER BY TotalScore DESC, NumAchievements DESC, LastAward";
+    if (config('feature.aggregate_queries')) {
+        $query = "SELECT ua.User, pg.achievements_unlocked_hardcore AS NumAchievements,
+                         pg.points_hardcore AS TotalScore, pg.last_unlock_hardcore_at AS LastAward
+                    FROM player_games pg
+                    LEFT JOIN UserAccounts ua ON ua.ID = pg.user_id
+                    WHERE NOT ua.Untracked
```
wescopeland, line 755: nit:
```suggestion
                    WHERE ua.Untracked = 0
```

When using `NOT` in front of Untracked, it makes querying that data non-sargable, blocking any indexes on the `Untracked` column from being used. The `NOT` operator acts as a transformation on the column, preventing an index from being used.

Jamiras: That cut another 90ms off the query time. Still about 160ms for the query.

wescopeland: Other optimization ideas:

* Add `LIMIT 10`.
* Use an `INNER JOIN` instead of a `LEFT JOIN` on `UserAccounts`.

Unfortunately, I'm not able to test either of these at the moment with the state of my local database.

wescopeland (approved):
Query is a huge improvement, even without additional micro-optimizations. Marking as approved.

app/Helpers/database/player-game.php
Jamiras: We can't have a limit as it has to return all rows where users have mastered. In this case, 562:
image

`INNER JOIN` did cut off another 20ms. That's a change we can probably apply to a lot of the slow queries without needing to use the aggregate tables.

luchaos (approved):

PR #1919, opened 2023-10-15
fix client error retrieving softcore unlocks with aggregate queries on

Jamiras:
```
'DateEarnedHardcore' => $unlock->unlocked_hardcore_at?->format('Y-m-d H:i:s'),
```
would return a null entry for `DateEarnedHardcore`, whereas the non-aggregate code does not populate the key. This caused the filtering code for the `unlocks` and `startsession` requests to not produce the desired results.

https://github.com/RetroAchievements/RAWeb/blob/c0fa53753c0c1a16ce9418c17af8c72f23af1233/public/dorequest.php#L392-L394

https://github.com/RetroAchievements/RAWeb/blob/c0fa53753c0c1a16ce9418c17af8c72f23af1233/public/dorequest.php#L325-L335

In fact, the `strtotime` call would change the value from `null` to `false` resulting in unparseable timestamps for the client.
```
{"Success":true,"HardcoreUnlocks":[{"ID":361917,"When":false},{"ID":361918,"When":false},{"ID":361919,"When":false},
{"ID":361921,"When":false},{"ID":361922,"When":false},{"ID":361924,"When":false},{"ID":361925,"When":false},
{"ID":361928,"When":false},{"ID":361931,"When":false}],"ServerNow":1697396485}
```

### DIFF
--- a/tests/Feature/Connect/BootstrapsConnect.php
+++ b/tests/Feature/Connect/BootstrapsConnect.php
@@ -15,6 +15,8 @@ trait BootstrapsConnect
     {
         parent::setUp();
 
+        // config(['feature.aggregate_queries' => true]);
+
         /** @var User $user */
         $user = User::factory()->create(['appToken' => Str::random(16)]);
         $this->user = $user;
--- a/app/Helpers/database/player-game.php
+++ b/app/Helpers/database/player-game.php
@@ -322,11 +322,18 @@ function getUserAchievementUnlocksForGame(string $username, int $gameID, int $fl
                 'unlocked_hardcore_at',
             ])
             ->mapWithKeys(function (PlayerAchievement $unlock, int $key) {
-                return [$unlock->achievement_id => [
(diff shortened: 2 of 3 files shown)


tests/Feature/Connect/BootstrapsConnect.php
```diff
@@ -15,6 +15,8 @@ protected function setUp(): void
     {
         parent::setUp();
 
+        // config(['feature.aggregate_queries' => true]);
```
luchaos, line 18: Is this still needed/a reference to a TODO?

luchaos (approved):

tests/Feature/Connect/BootstrapsConnect.php
Jamiras: It's a way to toggle the tests to run against the aggregate queries code, which was a path not being tested, letting this problem through. I don't know if there's an easy way to test both code paths, so I left this in as an easy toggle.

PR #1920, opened 2023-10-21
fix(connect): do not record ping requests without a game ID

luchaos:
Fixes `App\Platform\Events\PlayerSessionHeartbeat::__construct(): Argument #2 ($game) must be of type App\Platform\Models\Game, null given`.

I don't think saving the last login date for users when no valid date was provided is worth the overhead. There are a lot of writes to the last login date already.

### DIFF
--- a/public/dorequest.php
+++ b/public/dorequest.php
@@ -230,7 +230,7 @@ switch ($requestType) {
      */
 
     case "ping":
-        if ($user === null) {
+        if ($user === null || !$gameID) {
             $response['Success'] = false;
         } else {
             $activityMessage = request()->post('m');


public/dorequest.php
```diff
@@ -230,7 +230,7 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
      */
 
     case "ping":
-        if ($user === null) {
+        if ($user === null || !$gameID) {
```
Jamiras, line 233: Verified that the log seems to be associated with pings for game 0 (wonder what's doing that), but the error could still occur if an unknown ID was provided. The exception just indicates `Game::find` returned null.

luchaos: An ID for a game that doesn't exist is fine I'd say. 0 looks like invalid input - it's just a quick fix, better validation will come when we migrate dorequest to controllers?

Jamiras (approved):

PR #1924, opened 2023-10-23
feat: use pre-aggregated data for unlocks on users list

luchaos:
This should get rid of one of the majorly slow queries.

### DIFF
--- a/app/Helpers/database/user.php
+++ b/app/Helpers/database/user.php
@@ -284,9 +284,8 @@ function getUserListByPerms(int $sortBy, int $offset, int $count, ?array &$dataO
         default => "ua.User ASC ",
     };
 
-    // TODO slow query (70) when ordering by NumAwarded
     $query = "SELECT ua.ID, ua.User, ua.RAPoints, ua.TrueRAPoints, ua.LastLogin,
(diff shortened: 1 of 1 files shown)


wescopeland (approved):

app/Helpers/database/user.php
```diff
@@ -284,9 +284,8 @@ function getUserListByPerms(int $sortBy, int $offset, int $count, ?array &$dataO
         default => "ua.User ASC ",
     };
 
-    // TODO slow query (70) when ordering by NumAwarded
     $query = "SELECT ua.ID, ua.User, ua.RAPoints, ua.TrueRAPoints, ua.LastLogin,
-                (SELECT COUNT(*) AS NumAwarded FROM Awarded AS aw WHERE aw.User = ua.User) NumAwarded
+                ua.achievements_unlocked NumAwarded
```
Jamiras, line 288: This also fixes an issue where hardcore unlocks were counted twice because the subquery doesn't have a filter on `HardcoreMode` or a `group by AchievementID` clause.

Before:
image

After:
image

Profile:
image

12049+90=12139

Jamiras (approved):

app/Helpers/database/user.php
luchaos: wonderful! i wondered why there were no additional filters 😅

PR #1925, opened 2023-10-24
use aggregated data for usergameactivity

Jamiras:
Because this page no longer reads the Activity table, some historical data is lost. Sessions have to be generated only from unlock timestamps - we can no longer rely on the StartedPlaying Activity entries. Similarly, achievements that have been reset will no longer appear on the page as the historical unlock information is no longer available.

New sessions (created after v5 was deployed) should now be accurately captured instead of estimated. 

For legacy data, the estimations are no longer as accurate. Here's one user playing an RPG:
image

Without the Activity table data, the same user/game:
image

7 sessions were extrapolated from the Activity table that contain no unlocks. These sessions would previously contribute to the overall time played through the estimated session extension logic. Similarly, without the StartedPlaying Activity entries, the lead up to the first achievement unlocked in a session is not captured (in this case it's almost two hours into the session):
image

As a result of the lead-up time missing, the session extension calculation is smaller, and without the "empty" sessions, the number of sessions that gets extended is smaller, which results in a much lower estimate (16 hours instead of 28).

Since it's just an estimate, we can't know which is more accurate. The loss of the lead-up time clearly makes the estimate too low, but the calculated session extension value could be overcorrecting for shorter sessions. And we have no idea whatsoever how long the sessions where achievements weren't earned were.

We want to get rid of the Activity table. Unfortunately, doing so does decrease the accuracy of the information provided on this page for historical data, but the newer data should be more accurate as it gets populated. Additionally, as this page is no longer database intensive, it might be possible to open it up to more users in the future.


app/Helpers/database/user-activity.php
```diff
@@ -440,100 +443,165 @@ function GetMostPopularTitles(int $daysRange = 7, int $offset = 0, int $count =
...
-                'StartTime' => strtotime($row['lastupdate']),
-            ];
+    $playerSessions = PlayerSession::where('user_id', '=', $user->ID)
+        ->where('game_id', '=', $gameID)
+        ->get();
+    foreach ($playerSessions as $playerSession) {
+        $session = [
+            'StartTime' => $playerSession->created_at->unix(),
+            'EndTime' => $playerSession->updated_at->unix(),
+            'IsGenerated' => $playerSession->created_at < Carbon::create(2023, 10, 14, 13, 16, 42),
```
Jamiras, line 469: The timestamp is from the first `player_session` created in the prod database. 

This is leftover from when I tried to backfill `player_sessionss from the `Activity` data, which turned out to be unreasonably expensive (estimate to completely backfill the database was nearly a year due to querying the Activity table multiple times for every one of the roughly eight million player_games rows).

luchaos (approved):

PR #1930, opened 2023-10-26
feat: switch to mariadb and convert to utf8mb4

luchaos:
Add the following to your `.env` until the mariadb utf8 dump is ready:
```env
DB_CHARSET=latin1
DB_COLLATION=latin1_general_ci
```

Rebuilding docker containers is not advised yet as it will remove the mysql container from compose.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,6 +50,9 @@ DB_PORT=3306
 DB_DATABASE=retroachievements-web
 DB_USERNAME=retroachievements
 DB_PASSWORD="${DB_USERNAME}"
+# TODO remove after utf8mb4 conversion
+#DB_CHARSET=latin1
+#DB_COLLATION=latin1_general_ci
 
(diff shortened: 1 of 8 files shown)


Jamiras (approved):
Verified new database created by `php artisan migrate:fresh --seed` contains tables with `DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci`.

Though should it be `utf8mb4_unicode_ci` per the mysql.cnf?

Jamiras:
Also verified explicitly setting `DB_CHARSET` fixes non-latin character display in existing dump.

luchaos:
> Though should it be `utf8mb4_unicode_ci` per the mysql.cnf?

mariadb uses general by default so the collation is set explicitly in the config with the lates commit

Jamiras (approved):

PR #1932, opened 2023-10-27
populate as much of Recent Players list as possible from aggregate data

Jamiras:
Fetching rich presence data from UserAccounts takes 2-5 times longer than fetching it from player_sessions. This attempts to fill as much of the Recent Players list as possible from player_sessions. If 10 rows can be generated, then the UserAccounts table doesn't have to even be looked at.

For games with less than 10 players since the v5 release, the result will be slower because two queries will be executed, but they will be taking less traffic, so that's an acceptable tradeoff.

### DIFF
--- a/app/Helpers/database/player-game.php
+++ b/app/Helpers/database/player-game.php
@@ -5,6 +5,7 @@ use App\Platform\Enums\AchievementType;
 use App\Platform\Enums\UnlockMode;
 use App\Platform\Models\Achievement;
 use App\Platform\Models\PlayerAchievement;
+use App\Platform\Models\PlayerSession;
 use App\Site\Enums\Permissions;
 use App\Site\Models\User;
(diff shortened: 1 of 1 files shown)


luchaos (approved):
Very nice

app/Helpers/database/player-game.php
```diff
@@ -698,26 +699,55 @@ function getTotalUniquePlayers(
     return (int) (legacyDbFetch($query, $bindings)['players_count'] ?? 0);
 }
 
-function getGameRecentPlayers(int $gameID, int $maximum_results = 0): array
+function getGameRecentPlayers(int $gameID, int $maximum_results = 10): array
 {
     $retval = [];
+    $filter = '';
+
+    if (config('feature.aggregate_queries')) {
```
luchaos, line 707: The feature flag config check is not needed anymore

PR #1935, opened 2023-10-27
chore: adjust index names

luchaos:
Minor cosmetic change: before starting imports on the new server I'd like to get those index names cleaned up. With data in the tables it takes unnecessarily long to change them.

The dumps won't include table schema, only data, meaning that the migrations will define how things are named in the new databases. Another cleanup of migrations to better match current prod as closely as possible happened prior.

### DIFF
--- a/database/migrations/2012_10_03_133633_create_base_tables.php
+++ b/database/migrations/2012_10_03_133633_create_base_tables.php
@@ -128,7 +128,7 @@ return new class() extends Migration {
                 } else {
                     $table->primary(['GameID', 'Address']);
                 }
-                $table->index('GameID');
+                $table->index('GameID', 'memory_notes_game_id_index');
             });
         }
 
@@ -157,7 +157,7 @@ return new class() extends Migration {
                 $table->timestampTz('Submitted')->useCurrent();
                 $table->timestampTz('Edited')->nullable();
 
-                $table->index('ArticleID');
+                $table->index('ArticleID', 'comments_commentable_id_index');
             });
         }
 
@@ -194,7 +194,7 @@ return new class() extends Migration {
         if (!Schema::hasTable('Forum')) {
             Schema::create('Forum', function (Blueprint $table) {
                 $table->increments('ID');
-                $table->unsignedInteger('CategoryID')->index();
+                $table->unsignedInteger('CategoryID')->index('forums_forum_category_id_index');
                 $table->string('Title', 50);
                 $table->string('Description', 250);
                 $table->unsignedInteger('LatestCommentID')->nullable();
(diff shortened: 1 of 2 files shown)


database/migrations/2012_10_03_133633_create_base_tables.php
```diff
@@ -157,7 +157,7 @@ public function up(): void
                 $table->timestampTz('Submitted')->useCurrent();
                 $table->timestampTz('Edited')->nullable();
 
-                $table->index('ArticleID');
+                $table->index('ArticleID', 'comments_commentable_id_index');
```
Jamiras, line 160: There actually is a `commentable_id` column in the prod db. Won't this name be confusing?

database/migrations/2012_10_03_133633_create_base_tables.php
```diff
@@ -286,8 +286,8 @@ public function up(): void
                     $table->bigIncrements('id')->first();
                 }
 
-                $table->string('User', 32)->index();
-                $table->string('Friend', 32)->index();
+                $table->string('User', 32)->index('user_relations_user_id_index');
+                $table->string('Friend', 32)->index('user_relations_related_user_id_index');
```
Jamiras, line 290: Same comment. There are `user_id` and `related_user_id` columns in the prod db.

database/migrations/2012_10_03_133633_create_base_tables.php
```diff
@@ -782,7 +782,7 @@ public function up(): void
                 $table->integer('RASoftcorePoints')->nullable()->default(0)->after('RAPoints');
 
                 // https://github.com/RetroAchievements/RAWeb/blob/master/database/20220810_000000_Add_UserAccount_SoftcorePoints_Key.sql
-                $table->index(['RASoftcorePoints', 'Untracked']);
+                $table->index(['RASoftcorePoints', 'Untracked'], 'users_points_softcore_unranked_index');
```
Jamiras, line 785: `untracked`, not `unranked`.

database/migrations/2012_10_03_133633_create_base_tables.php
luchaos: truw, yet `ArticleID` will be dropped completely at some point as it's redundant with the polymorphic commentable relation. this index doesn't have to be renamed, really.

database/migrations/2012_10_03_133633_create_base_tables.php
luchaos: this was a mistake - those should be username instead of user_id if anything. both User and Friend columns will be removed in the future, so adding an index name for those isn't necessary either.
removed custom index names in https://github.com/RetroAchievements/RAWeb/pull/1935/commits/b78ec3a435a8a5474e20a149ae209372d2b14b89

database/migrations/2012_10_03_133633_create_base_tables.php
luchaos: copied that from the User model where a todo states to rename untracked to either unranked or remove in favor of unranked_at but i'm not sold on the term `unranked` myself. that's a leftover from the v2 prototype.
adjusted in https://github.com/RetroAchievements/RAWeb/pull/1935/commits/b78ec3a435a8a5474e20a149ae209372d2b14b89

Jamiras (approved):

PR #1936, opened 2023-10-27
feat: use aggregated data only

luchaos:
delete dead code while at it.

PR #1940, opened 2023-10-27
Eliminate aggregate queries feature

Jamiras:
Includes all the changes in #1936.


app/Platform/Actions/UpdateGamePlayerGames.php
```diff
@@ -18,7 +18,10 @@ public function execute(Game $game): void
     {
         // don't do this without a proper queue
         if (config('queue.default') === 'sync') {
-            return;
+            // do allow this to cascade in the unit tests
+            if (!app()->environment('testing')) {
+                return;
+            }
```
luchaos, line 24: it should cascade for tests and returns if it's not testing which makes sense but confused me because testing is always run with sync driver. the only other variant that allows to run sync is when set locally.
could this be rephrased positively to read for `local` specifically or would this be more confusing?
what we want to prevent here specifically is that when running with a huge dataset locally that the sync queue would run for a very long time which is not a problem with tests.
`if (app()->environment('local')) {`

luchaos (approved):
Thank you for porting this over and making it work 🙇

app/Platform/Actions/UpdateGamePlayerGames.php
luchaos: the more i think about it the more it makes sense how you wrote it - feel free to discard :)

luchaos: it could be inlined though:
```php
// don't do this without a proper queue and not testing
if (config('queue.default') === 'sync' && !app()->environment('testing')) {
    return;
}
```

app/Platform/Actions/UpdateGamePlayerGames.php
```diff
@@ -39,11 +42,13 @@ public function execute(Game $game): void
...
                     ->name('player-game-metrics ' . $game->id . ' ' . $page)
                     ->allowFailures()
-                    ->finally(function (Batch $batch) {
+                    ->finally(function (Batch $batch) use ($game) {
                         // mark batch as finished even if jobs failed
                         if (!$batch->finished()) {
                             resolve(BatchRepository::class)->markAsFinished($batch->id);
                         }
+                        // some game metrics depend on the player_games rows
+                        (new UpdateGameMetrics())->execute($game);
```
luchaos, line 51: That's a good addition - this was certainly missing when looking at some games' metrics

luchaos: thinking about it this definitely should be dispatching a job instead of running the game metris action directly

PR #1941, opened 2023-10-28
use aggregate data for achievement info page

Jamiras:


app/Helpers/database/player-achievement.php
```diff
@@ -94,9 +67,47 @@ function unlockAchievement(User $user, int $achievementId, bool $isHardcore): ar
...
+            $playerGame->save();
+        }
+    }
+
+    if ($playerGame) {
+        if ($isHardcore) {
+            $retVal['AchievementsRemaining'] = $achievement->game->achievements_published - $playerGame->achievements_unlocked_hardcore;
+        } else {
+            $retVal['AchievementsRemaining'] = $achievement->game->achievements_published - $playerGame->achievements_unlocked;
+        }
```
luchaos, line 110: I'm not sure that we should do any synchronous updates here. the unlock job does all of that, the unlockAchievement method was bound to be removed.
That stats are out of sync for a while it ok, the optimistic response is enough for clients.

This would allow for thousands of unlock requests to not overwhelm the database again as it did in the past.

luchaos: the idea is to getting rid of most of these global helper functions in the long run.
using this unlock method to write anything to the db defeats the purpose of moving most of the write operations to queued jobs.

Jamiras: The client is expecting an immediate response of the player's new score and how many achievements they have left to unlock. The optimistic response is fine if only one achievement is unlocked, but there are many times where multiple achievements get unlocked at the same time (defeat final boss, defeat final boss without continuing, have all items after defeating final boss, etc). In that case, to show the mastery notification on the client, the "AchievementsRemaining" must be 0 on one of the calls. If they all optimistically subtract 1 from the `player_games->achievements_unlocked` counter, the client will see "2 remaining" for all three requests and not show the notification.

Similarly, the client could receive "base points+achievement1 points", "base points+achievement2 points" and "base points+achievement3 points" and be out-of-sync with the server. I'm less concerned about this as few people actually look at their points in the client, but we need to do something to ensure the mastery is shown correctly. 

Maybe we could do something ephemeral with redis where it caches the number of pending unlocks?

luchaos: hrm, i was not aware that mastery popups are in play here, too. can we reduce the writing to player_games?
that'd leave points out of it and use player_games as the only reference to keep track. i don't think redis would make this much easier to maintain.

luchaos (approved):

PR #1943, opened 2023-10-28
use aggregate data for history

Jamiras:


public/API/API_GetAchievementsEarnedBetween.php
```diff
@@ -25,7 +25,13 @@
...
  */
 
-$user = request()->query('u');
+use App\Site\Models\User;
+
+$user = User::firstWhere('User', request()->query('u'));
+if (!$user) {
+    return response()->json([]);
+}
+
```
luchaos, line 34: We could introduce validation responses here similar to the public request scripts. Was planned to happen in future iterations of the API:
```php
    'u' => 'required|string|exists:UserAccounts,User|alpha_num|max:20',
]);
```
i'm not sure what the framework's json response would look like and if we'd want to keep it this way.

app/Platform/Models/Game.php
```diff
@@ -65,7 +65,6 @@ class Game extends BaseModel implements HasComments, HasMedia
     public const UPDATED_AT = 'Updated';
 
     protected $fillable = [
-        'system_id',
```
luchaos, line 68: This is not needed?

Jamiras: It doesn't exist. The field is "ConsoleID"

luchaos: good point! i think it was used for the seeders/test helpers but with factories that's apparently not required anymore. interesting

public/API/API_GetAchievementsEarnedBetween.php
Jamiras: It returns
```
{
      "message": "The u field is required.",
      "errors": {
          "u": [
              "The u field is required."
          ]
      }
  }
```

Instead of 
```
[]
```

So, instead of an empty array, the client would receive an error object.

Jamiras: Additionally, doesn't the `exists` validation have the query the database, thus adding additional overhead since we have to retrieve the record anyway?

luchaos: good point, exists is really fast though.
let's revisit the validation thing sepatately when refactoring the v2 web api into controllers.

apart from that, is empty array the correct response, should it be empty object?

luchaos: nevermind, in this case it's an array 👍

luchaos (approved):
👏

PR #1944, opened 2023-10-28
chore: remove player_games.update_status

luchaos:


database/migrations/platform/2023_09_15_000000_update_player_games_player_achievement_sets.php
```diff
@@ -11,7 +11,6 @@ public function up(): void
     {
         Schema::table('player_games', function (Blueprint $table) {
             $table->string('achievement_set_version_hash')->nullable()->after('game_hash_id');
-            $table->string('update_status')->nullable()->after('achievement_set_version_hash');
```
Jamiras, line 14: Does there need to be a step to remove this? The column already exists in the prod db.

luchaos: i would've removed those from the server manually but that's bot how migrations should be dealt with - it's dirty.
i'll add a proper migration for that 👍
at some point i'd like to merge all migrations to remove unnecessary migration steps but that will take some time and it's not very important. might speed up tests though.

luchaos: Added in https://github.com/RetroAchievements/RAWeb/pull/1944/commits/206b8f02a7e0d8e167f20bbd50c41ac346517668

Jamiras (approved):

PR #1945, opened 2023-10-28
fix: incorrect index and unique on tickets table

luchaos:


luchaos:
I have to merge this right away to proceed with new db server imports.

PR #1948, opened 2023-10-28
chore: move legacy last game and presence writes to ResumePlayerSession

luchaos:
Reduces synchronous writes in dorequest.


app/Platform/Actions/ResumePlayerSession.php
```diff
@@ -36,12 +36,24 @@ public function execute(
             ->orderByDesc('id')
             ->first();
 
+        if ($user->LastGameID !== $game->id) {
+            expireRecentlyPlayedGames($user->User);
+            // TODO deprecated, read from last player_sessions entry where needed
+            $user->LastGameID = $game->id;
+            $user->save();
```
Jamiras, line 43: There's an `$user->save()` on line 75 that doesn't look like it can be avoided. Perhaps remove this and line 56.

luchaos: removed this one but the other two are different paths with an early return in-between.
changed in https://github.com/RetroAchievements/RAWeb/pull/1948/commits/43bc4278db6444f7803c1c9db3bcb8127e78909e

luchaos: actually, i know why the highlighted save was kept in place - it'll probably stay longer than the other two as those will be replaced with player_sessions rich presence whereas the last game id will probably be changed independently.

Jamiras: You're right. I missed the return on line 61. This one probably should be restored because it won't happen if presence is null.

luchaos: done in https://github.com/RetroAchievements/RAWeb/pull/1948/commits/f4172e6a7b5631916bcff88aac65b6ee4748f911

Jamiras (approved):

PR #1954, opened 2023-10-29
fix: user created timestamp not cast to Carbon

luchaos:
Do not set timestamps to false on user object - this prevents datetime casting for Created and Updated attributes.

Change ticket `isAllowedToSubmitTickets()` check.

### DIFF
--- a/app/Helpers/database/ticket.php
+++ b/app/Helpers/database/ticket.php
@@ -23,27 +23,28 @@ function isAllowedToSubmitTickets(string $username): bool
 
     $cacheKey = CacheKey::buildUserCanTicketCacheKey($username);
 
-    $value = Cache::get($cacheKey);
-    if ($value !== null) {
-        return $value;
+    if (Cache::has($cacheKey)) {
+        return Cache::get($cacheKey);
     }
 
     $user = User::firstWhere('User', $username);
-    $value = $user->Created->diffInDays() > 1
+    $isAllowed = $user->Created->diffInDays() > 1
         && getRecentlyPlayedGames($username, 0, 1, $userInfo)
         && $userInfo[0]['GameID'];
 
-    if ($value) {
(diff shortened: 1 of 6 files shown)


app/Helpers/database/ticket.php
```diff
@@ -15,35 +15,36 @@
...
     }
 
-    $cacheKey = CacheKey::buildUserCanTicketCacheKey($user);
+    $cacheKey = CacheKey::buildUserCanTicketCacheKey($username);
 
-    $value = Cache::get($cacheKey);
-    if ($value !== null) {
-        return $value;
+    if (Cache::has($cacheKey)) {
+        return Cache::get($cacheKey);
```
Jamiras, line 27: Doesn't this cause a double lookup?

Jamiras: Without the join against Activity, we could probably just eliminate caching of this altogether. What about something like this?
```
function isAllowedToSubmitTickets(string $username): bool
{
    $user = User::firstWhere('User', $username);
    if (!$user || $user->Created->diffInDays() < 1) { 
        return false;
    }

    return PlayerGame::where('user_id', $user->id)
        ->where('time_taken', '>', 5)
        ->exists();
}
```

luchaos: very nice - changed in https://github.com/RetroAchievements/RAWeb/pull/1954/commits/8bbe110fd76f6fabf42f8596df0c39d3a09e1186

Jamiras: Should also remove the code for `buildUserCanTicketCacheKey`

luchaos: removed in https://github.com/RetroAchievements/RAWeb/pull/1954/commits/ea238e5d0169684a295cd71f79b748907b76a9c9

Jamiras (approved):

PR #1956, opened 2023-10-30
perf(game): add composite indexes for game_id and achievements unlocked

wescopeland:
This dramatically improves performance on the two queries to fetch the game's achievement distribution on game pages.

**Benchmark query:**
```sql
-- softcore mode, use NOW() to bypass local performance caching
SELECT
	NOW(),
	pg.achievements_unlocked AS AwardedCount,
	COUNT(*) AS NumUniquePlayers
FROM
	player_games AS pg
WHERE
	pg.game_id = 228
	AND pg.achievements_unlocked > 0
GROUP BY
	AwardedCount
ORDER BY
	AwardedCount DESC
```

**Before**
4.389s
4.697s
4.282s
4.723s
4.867s
Average: _4.591s_

**After**
38ms
23ms
8ms
24ms
19ms
Average: _22.4ms_

The addition of these indices results in a speedup factor of around 535x. This is especially desirable given the query **runs twice on every game page**.

### DIFF
--- a/database/migrations/platform/2023_10_29_000000_update_player_games_table.php
+++ b/database/migrations/platform/2023_10_29_000000_update_player_games_table.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+            $table->index(['game_id', 'achievements_unlocked'], 'game_id_achievements_unlocked_index');
+            $table->index(['game_id', 'achievements_unlocked_hardcore'], 'game_id_achievements_unlocked_hardcore_index');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+            $table->dropIndex('game_id_achievements_unlocked_index');
+            $table->dropIndex('game_id_achievements_unlocked_hardcore_index');
+        });
+    }
+};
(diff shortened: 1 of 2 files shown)


Jamiras:
I'm not sure if the NOW() is enough to bypass my local caching.

> Before: 69ms, 71ms, 67ms, 70ms, 68ms (average: 69ms)
> After: 8ms, 9ms, 8ms, 8ms, 8ms (average: 8ms)

Still a 8x improvement.

For 1446:

> Before: 149ms, 78ms, 116ms, 94ms, 230ms (average: 133ms)
> After: 5ms, 5ms, 8ms, 7ms, 6ms (average: 6ms)

That's a 22x improvement

Jamiras (approved):

database/migrations/platform/2023_10_29_000000_update_player_games_table.php
```diff
@@ -0,0 +1,25 @@
...
+            $table->index(['game_id', 'achievements_unlocked']);
+            $table->index(['game_id', 'achievements_unlocked_hardcore']);
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+            $table->dropIndex('player_games_game_id_achievements_unlocked_index');
+            $table->dropIndex('player_games_game_id_achievements_unlocked_hardcore_index');
```
luchaos, line 22:
```php
$table->dropIndex(['game_id', 'achievements_unlocked']);
$table->dropIndex(['game_id', 'achievements_unlocked_hardcore']);
```

luchaos:
> runs twice on every game page

Should probably be pre-aggregated on games/sets. Won't scale well in the future.

database/migrations/platform/2023_10_29_000000_update_player_games_table.php
wescopeland: Thanks, that's much easier indeed. Resolved in latest. I'll keep hunting for more optimizations.

luchaos (approved):

PR #1962, opened 2023-10-30
fix(user): only count games with at least 1 earned achievement as "Unfinished"

wescopeland:
This PR fixes a bug on user profiles and the user completion progress page where, now that aggregate queries are enabled, games with no achievements unlocked are appearing/counting in these areas as "Unfinished".

We should only count a game as Unfinished if the user has earned at least one achievement for the game. Simply booting a game should not qualify.

@luchaos , this issue is highly visible on your profile for these areas of the UI.

**Before**
Screenshot 2023-10-30 at 4 49 41 PM

Screenshot 2023-10-30 at 4 49 30 PM

Screenshot 2023-10-30 at 4 49 36 PM

**After**
Screenshot 2023-10-30 at 4 48 43 PM

Screenshot 2023-10-30 at 4 49 14 PM

### DIFF
--- a/app/Platform/Controllers/PlayerCompletionProgressController.php
+++ b/app/Platform/Controllers/PlayerCompletionProgressController.php
@@ -14,12 +14,10 @@ use Illuminate\Support\Facades\Auth;
 
 class PlayerCompletionProgressController extends Controller
 {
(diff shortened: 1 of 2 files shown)


luchaos (approved):
neat!

PR #1963, opened 2023-10-30
fix(user): dont hide profile completion progress widget row

wescopeland:
Resolves an issue on user profiles where the "see more..." link to take the user to the unfiltered Completion Progress page may never appear.

@luchaos, this is another issue that is exhibited on your profile 😅

**Before**
Screenshot 2023-10-30 at 5 08 51 PM

**After**
Screenshot 2023-10-30 at 5 08 40 PM

**Root Cause**
The conditional for determining whether or not to render the filter checkbox was driving the entire div that also contained the "see more" link. I've removed the conditional and its helper function entirely- IMO, they're overkill.

### DIFF
--- a/app/Helpers/render/user.php
+++ b/app/Helpers/render/user.php
@@ -71,25 +71,6 @@ function renderUserCard(string|array $user): string
     ]);
 }
 
-function getCompletedAndIncompletedSetsCounts(array $userCompletedGamesList): array
-{
-    $completedSetsCount = 0;
-    $incompletedSetsCount = 0;
-
-    foreach ($userCompletedGamesList as $game) {
-        $nextMaxPossible = $game['MaxPossible'];
-        $nextNumAwarded = $game['NumAwarded'];
-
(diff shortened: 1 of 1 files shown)


luchaos (approved):
very nice, less code is best code.
btw. we should think about a ubiquitous way to label those buttons, we have "more", "see more", thinking of "show more" sometimes.
same for "my", "your", or omitting posessive pronouns altogether.

wescopeland:
I agree. No strong opinions on what the label should be; it is bothering me that we have "more..." on the home page and "See more..." here. Maybe worth changing the label to "more..." here too?

luchaos:
> I agree. No strong opinions on what the label should be; it is bothering me that we have "more..." on the home page and "See more..." here. Maybe worth changing the label to "more..." here too?

yeah, i guess that'd be best for now to stay consistent. sorry, merged early.

wescopeland:
No worries at all, followup here https://github.com/RetroAchievements/RAWeb/pull/1964

PR #1966, opened 2023-10-30
perf(user): optimize getAchievementsEarnedBetween()

wescopeland:
`getAchievementsEarnedBetween()` is responsible for building the chart to show the user's points progress over the last ~2 weeks.

This is currently very slow for users with a lot of progress on the site, even though it is using denormalized data. After running `EXPLAIN` on the query, I identified the main driver behind this slowness is because the query has to scan potentially tens of thousands of rows on the `player_achievements` table due to a missing compound index.

**Benchmark Query**
```sql
SELECT
	COALESCE(pa.unlocked_hardcore_at, pa.unlocked_at) AS Date,
	SUM( IF(pa.unlocked_hardcore_at IS NOT NULL, ach.Points, 0)) AS HardcorePoints,
	SUM(ach.Points) AS SoftcorePoints
FROM
	player_achievements pa
	INNER JOIN Achievements AS ach ON ach.ID = pa.achievement_id
	INNER JOIN GameData AS gd ON gd.ID = ach.GameID
WHERE -- user = MaxMilyin
	pa.user_id = 16446 AND ach.Flags = 3 AND COALESCE(pa.unlocked_hardcore_at, pa.unlocked_at) BETWEEN '2023-10-16 21:52:14' AND '2023-10-30 21:52:14'
GROUP BY
	Date
ORDER BY
	Date ASC
LIMIT 0, 14;
```

**Before**
3.049s
2.748s
2.722s
2.773s
2.765s
Average: 2.81s

**After**
21ms
25ms
26ms
22ms
26ms
Average: 24ms

### DIFF
--- a/database/migrations/platform/2023_10_30_000000_update_player_achievements_table.php
+++ b/database/migrations/platform/2023_10_30_000000_update_player_achievements_table.php
@@ -12,7 +12,7 @@ return new class() extends Migration {
         Schema::table('player_achievements', function (Blueprint $table) {
             $table->index(
                 ['user_id', 'unlocked_at', 'unlocked_hardcore_at', 'achievement_id'],
-                'idx_user_date_achievement'
+                'player_achievements_user_date_achievement'
             );
         });
     }
@@ -20,7 +20,7 @@ return new class() extends Migration {
     public function down(): void
     {
         Schema::table('player_achievements', function (Blueprint $table) {
-            $table->dropIndex('idx_user_date_achievement');
+            $table->dropIndex('player_achievements_user_date_achievement');
         });
     }
 };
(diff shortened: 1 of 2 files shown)


database/migrations/platform/2023_10_30_000000_update_player_achievements_table.php
```diff
@@ -0,0 +1,26 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('player_achievements', function (Blueprint $table) {
+            $table->index(
+                ['user_id', 'unlocked_at', 'unlocked_hardcore_at', 'achievement_id'],
+                'player_achievements_user_date_achievement'
```
wescopeland, line 15: Without a custom index name, an error is thrown for the auto-generated default name being too long.

luchaos (approved):
this index might take a while to create?
should we plan for some downtime - or create the index ahead of time during off hours?
in the latter case we should add a check if the index already exists to skip it.

wescopeland:
@luchaos I've updated the migration per your feedback

luchaos (approved):

PR #1970, opened 2023-11-02
feat: use MariaDB uca1400_ai_ci and fix game title tilde sort

luchaos:
`utf8mb4_unicode_ci` is based on Unicode-4.0.0 which is fairly old. Use `uca1400_ai_ci` collation instead. (see https://jira.mariadb.org/browse/MDEV-27009)

The `utf8mb4_unicode_ci` mariadb container default can stay for now, Laravel will use `uca1400_ai_ci` for migrations. Changing the docker container default to `uca1400_ai_ci` resulted in `SQLSTATE[HY000] [2054] Server sent charset (0) unknown to the client`.
I'll look into that another time, not relevant for the server move.

Already migrated/imported local databases can stay as is - no need to change the collations just yet, but can be done if desired.

Fix tilde sort while at it.

Left is `utf8mb4_unicode_ci` collation without sort fix, right is with `uca1400_ai_ci` with the fix. Note the sorting of `Ł`:

Screenshot 2023-11-02 at 19 30 00

### DIFF
--- a/app/Helpers/database/game.php
+++ b/app/Helpers/database/game.php
@@ -347,7 +347,7 @@ function getGamesListByDev(
             if ($sortBy < 10) {
                 $orderBy .= ", IF(gd.Title LIKE '~%',1,0), gd.Title";
             } else {
-                $orderBy .= ", IF(gd.Title LIKE '~%',1,0), gd.Title";
+                $orderBy .= ", IF(gd.Title LIKE '~%',0,1), gd.Title DESC";
             }
         }
         if ($consoleID == 0) {
--- a/app/Helpers/database/game.php
+++ b/app/Helpers/database/game.php
@@ -321,8 +321,8 @@ function getGamesListByDev(
     }
 
     $orderBy = match ($sortBy) {
-        1 => "gd.Title",
-        11 => "gd.Title DESC",
+        1 => "IF(gd.Title LIKE '~%',1,0), gd.Title",
+        11 => "IF(gd.Title LIKE '~%',0,1), gd.Title DESC",
         2 => "NumAchievements DESC, MaxPointsAvailable DESC",
         12 => "NumAchievements, MaxPointsAvailable",
         3 => "MaxPointsAvailable DESC, NumAchievements DESC",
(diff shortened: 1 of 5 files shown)


app/Helpers/database/game.php
```diff
@@ -345,9 +345,9 @@ function getGamesListByDev(
     if (!empty($orderBy)) {
         if (!Str::contains($orderBy, "Title")) {
             if ($sortBy < 10) {
-                $orderBy .= ", Title";
+                $orderBy .= ", IF(gd.Title LIKE '~%',1,0), gd.Title";
             } else {
-                $orderBy .= ", Title DESC";
+                $orderBy .= ", IF(gd.Title LIKE '~%',0,1), gd.Title DESC";
             }
```
luchaos, line 351: Replacing this with just ascending sort for unrelated sorting feels more natural to me 🤔 
`$orderBy .= ", IF(gd.Title LIKE '~%',1,0), gd.Title";`

config/database.php
```diff
@@ -55,7 +55,7 @@
             'password' => env('DB_PASSWORD', ''),
             'unix_socket' => env('DB_SOCKET', ''),
             'charset' => env('DB_CHARSET', 'utf8mb4'),
-            'collation' => env('DB_COLLATION', 'utf8mb4_unicode_ci'),
+            'collation' => env('DB_COLLATION', 'uca1400_ai_ci'),
```
Jamiras, line 58: I tried recreating the test db, and got this error:
```
SQLSTATE[HY000]: General error: 1273 Unknown collation: 'uca1400_ai_ci' (SQL: SHOW FULL TABLES WHERE table_type = 'BASE TABLE')
```

fwiw, I'm on MariaDB 10.4:
```
Server version: 10.4.28-MariaDB Source distribution
```

luchaos: oh, that'd be a problem, yeah. docker container and server are using 10.11.
you can set the collation to unicode in .env in the meantime.

Jamiras (approved):
I can't test the collation change, but the tagged game sorting works.

PR #1976, opened 2023-11-04
feat: re-enable followed users ranking, move it to Followers page

wescopeland:
This PR:
* Makes some minor adjustments to the ranking query, swapping `LEFT JOIN` to `INNER JOIN` for a very minor performance improvement.
* Moves the followed users ranking component to _friends.php_.
* Adds a link to user profiles where the component used to be so users know where to find it now.

When following 800 users, the query always runs at less than 300ms for me. At more reasonable following counts, it's around 100ms. With it being off the user profile, it should have much less of an impact on the site.

Screenshot 2023-11-04 at 9 25 02 AM

Screenshot 2023-11-04 at 9 25 10 AM

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -649,8 +649,8 @@ RenderContentStart($userPage);
     echo "</div>";
 
     if ($user !== null && $user === $userPage) {
-        // FIXME: https://discord.com/channels/476211979464343552/1026595325038833725/1162746245996093450
-        // RenderPointsRankingComponent($user, true);
+        $friendCount = getFriendCount($user);
+        echo Blade::render('<x-user.followed-leaderboard-cta :friendCount="$friendCount" />', ['friendCount' => $friendCount]);
     }
     ?>
 </aside>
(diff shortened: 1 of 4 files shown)


public/userInfo.php
```diff
@@ -649,8 +649,8 @@ function resize() {
     echo "</div>";
 
     if ($user !== null && $user === $userPage) {
-        // FIXME: https://discord.com/channels/476211979464343552/1026595325038833725/1162746245996093450
-        // RenderPointsRankingComponent($user, true);
+        $friendCount = getFriendCount($user);
+        echo Blade::render('<x-user.followed-leaderboard-cta :friendCount="$friendCount" />', ['friendCount' => $friendCount]);
```
Jamiras, line 653: what is `cta` short for?

Jamiras (approved):

public/userInfo.php
wescopeland: Call to action

PR #1977, opened 2023-11-04
fix(setRequestList): remediate user request list sort order issue

wescopeland:
Remediates an issue where tilde characters in game titles are not being ignored in the sort order.

In the images below, look at the position of the unlicensed title, "Tiles of Fate".

**Before**
Screenshot 2023-11-04 at 9 56 40 AM

**After**
Screenshot 2023-11-04 at 9 56 33 AM

### DIFF
--- a/app/Helpers/database/set-request.php
+++ b/app/Helpers/database/set-request.php
@@ -36,7 +36,12 @@ function getUserRequestList(?string $user = null): array
         GROUP BY
             sr.GameID
         ORDER BY
-            GameTitle ASC";
+            CASE
+                WHEN gd.Title LIKE '~%' THEN
+                    SUBSTRING(gd.Title, LOCATE(' ', gd.Title) + 1)
+                ELSE
+                    gd.Title
+            END ASC";
 
     $dbResult = s_mysql_query($query);
 
--- a/app/Helpers/database/set-request.php
+++ b/app/Helpers/database/set-request.php
@@ -35,13 +35,7 @@ function getUserRequestList(?string $user = null): array
             sr.user = '$user' AND sr.type='" . UserGameListType::AchievementSetRequest . "'
         GROUP BY
             sr.GameID
-        ORDER BY
-            CASE
-                WHEN gd.Title LIKE '~%' THEN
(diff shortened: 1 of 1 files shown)


app/Helpers/database/set-request.php
```diff
@@ -36,7 +36,12 @@ function getUserRequestList(?string $user = null): array
         GROUP BY
             sr.GameID
         ORDER BY
-            GameTitle ASC";
+            CASE
+                WHEN gd.Title LIKE '~%' THEN
+                    SUBSTRING(gd.Title, LOCATE(' ', gd.Title) + 1)
+                ELSE
+                    gd.Title
+            END ASC";
```
luchaos, line 44: does this lead to the same sorting order as before the unicode change?
for game lists it was changed to
```php
ifStatement("gd.Title LIKE '~%'", 1, 0) . ", gd.Title"
```

wescopeland: It's slightly different. Is our intent to put all prefixed games (eg: Hack, Unlicensed, Homebrew) at the bottom of the list? If so, we can do:

```sql
        ORDER BY " . ifStatement("gd.Title LIKE '~%'", 1, 0) . ", gd.Title";
```

luchaos: at least that was the behaviour before if i'm no mistaken.
as soon as the tags are gone from the title and are added as tags they would sort the way you prepared it here.
we can go for that right away, too, but should probably change the other listings while at it?

oor, we remove all of that special behaviour and move those tags to the end of the title - like `[m]` achievements tags.

wescopeland: Let's keep it consistent with the game page per your suggestion. Truthfully, I can't remember what the previous behavior was. Updated in https://github.com/RetroAchievements/RAWeb/pull/1977/commits/fdd071bc93feca2ac72485c9a85aec0e76cbd574.

luchaos: alright, let's go with that for now and discuss our desired sorting strategy for future reference.
sorting special characters to the beginning of the list is technically correct - that's how most file systems do it, too.

luchaos (approved):

PR #1983, opened 2023-11-05
fix(game): only load google charts if necessary

wescopeland:
This PR resolves an uncaught JS error on game pages that do not have sets, which ultimately results in all kinds of front-end weirdness further downstream.

Because game pages that do not have sets do not render Google charts in the sidebar, the JS trying to act upon those charts is crashing. It's better to just not execute this JS at all.

This issue can be seen in prod on https://retroachievements.org/game/20786 if you open the browser devtools and inspect the Console tab. After this PR, that issue should go away.

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -343,246 +343,249 @@ sanitize_outputs(
 
 <?php RenderContentStart($pageTitle); ?>
 <?php if ($isFullyFeaturedGame): ?>
-    <script defer src="https://www.gstatic.com/charts/loader.js"></script>
-    <script>
-    document.addEventListener('DOMContentLoaded', function() {
-        if (typeof google !== 'undefined') {
-            google.load('visualization', '1.0', { 'packages': ['corechart'] });
-            google.setOnLoadCallback(drawCharts);
-        }
-    });
-
-    function drawCharts() {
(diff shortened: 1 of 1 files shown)


public/gameInfo.php
```diff
@@ -343,246 +343,249 @@ function generateGameMetaDescription(
...
 <?php if ($isFullyFeaturedGame): ?>
-    <script defer src="https://www.gstatic.com/charts/loader.js"></script>
-    <script>
-    document.addEventListener('DOMContentLoaded', function() {
-        if (typeof google !== 'undefined') {
-            google.load('visualization', '1.0', { 'packages': ['corechart'] });
-            google.setOnLoadCallback(drawCharts);
-        }
-    });
+    <?php if ($numDistinctPlayers): ?>
```
wescopeland, line 346: This (and the closing `endif`) are the relevant lines of code. Everything else is just whitespace adjustment.

Jamiras (approved):
Confirmed the presence of the uncaught error on the specified page, and it not being there after checking out this branch.

public/gameInfo.php
Jamiras: As a reviewer you can hide whitespace changes by adding a "&w=1" to the URL or clicking the gear between the tabs and the diff and checking the box.

PR #1987, opened 2023-11-05
prevent exception accessing linked hashes for invalid game id

Jamiras:
fixes
```
ErrorException(code: 0): Trying to access array offset on value of type null at
/home/forge/retroachievements.org/releases/2023-11-04T173719-5.1.1-8f93004d/public/linkedhashes.php:15
```


wescopeland (approved):
Repro'd locally with _linkedhashes.php?g=-1_ and verified this change resolves the issue.

PR #1989, opened 2023-11-05
feat(user): use new progression components for recently played

wescopeland:
This PR:

* Migrates the user profile "Last N games played" to Blade.
* Expands the functionality of the game list item components on the Completion Progress page to be optionally collapsible and accept a slot.
* Uses the aforementioned list item components in the "Last N games played" section.

**Key Functional Differences**
* Only the first item with achievements is pre-expanded by default.
* Beaten award displays **are not feature-flagged.** We are so close to shipping that I didn't think it was worth the extra effort.

**Before**
Screenshot 2023-11-05 at 12 56 22 PM

**After**
Screenshot 2023-11-05 at 1 11 06 PM

Screenshot 2023-11-05 at 12 55 52 PM

**Video**

https://github.com/RetroAchievements/RAWeb/assets/3984985/86403855-47a7-4c6c-ba44-13dcfbf18ffc


Jamiras:
> Only the first item with achievements is pre-expanded by default.

I find this behavior unexpected. If the first item doesn't have achievements, I wouldn't expect anything to be expanded. Once something becomes not-the-most-recent, it should be given less attention.

Jamiras:
I like the fade-in of the achievement when expanding, but I don't like the fact that they scroll up as they fade in. 
image

If anything, they should scroll down to mimic the control expanding.

wescopeland:
> I find this behavior unexpected. If the first item doesn't have achievements, I wouldn't expect anything to be expanded. Once something becomes not-the-most-recent, it should be given less attention.

No problem. After stepping away from it for a little while and coming back, I can see how this is unexpected. I've adjusted this in latest so that only the first item will be pre-expanded if it has achievements.

> I like the fade-in of the achievement when expanding, but I don't like the fact that they scroll up as they fade in. If anything, they should scroll down to mimic the control expanding.

That makes sense. I've adjusted this in latest. The change is here: https://github.com/RetroAchievements/RAWeb/pull/1989/commits/1703827acc0383880abd72094b41dda7ce73abb6#diff-995fc2eec60ca841b104cfadf3ac52b6c74fb5ec5e9a132a884deabc37856e5dR111 and was done by swapping `translate-y-1.5` to `-translate-y-1.5`.


https://github.com/RetroAchievements/RAWeb/assets/3984985/a40f4baa-6113-4f24-9020-22d0f3567e63

public/userInfo.php
```diff
@@ -502,113 +502,34 @@ function resize() {
...
+                :recentlyPlayedEntities="$recentlyPlayedEntities"
+                :recentAchievementEntities="$recentAchievementEntities"
+                :recentAwardedEntities="$recentAwardedEntities"
+                :targetUsername="$targetUsername"
+                :userAwards="$userAwards"
+            />
+        ', [
+            'recentlyPlayedCount' => $userMassData['RecentlyPlayedCount'],
+            'recentlyPlayedEntities' => $userMassData['RecentlyPlayed'],
+            'recentAchievementEntities' => $userMassData['RecentAchievements'],
```
Jamiras, line 519: I get an exception on this line accessing the profile page for the most recently registered user.

wescopeland: Good catch, fixed in https://github.com/RetroAchievements/RAWeb/pull/1989/commits/243254d177ab6b39570841da23932f30ebec9ab6 by setting some sane fallbacks.

Jamiras (approved):

wescopeland:
Safe to merge once #1999 is merged, as it is the only outstanding gating factor on shipping the beaten games feature.

PR #1992, opened 2023-11-06
feat(home): delay autoscrolling on the news' first slide

wescopeland:
A trivial UX improvement to the home page news carousel. Show the first slide for significantly longer (~20 seconds) before beginning autoscrolling to slide 2.

Whatever is on the first slide is likely the most important item for incoming users to see, so it should receive more prominence in the component.

### DIFF
--- a/resources/js/alpine/newsCarouselComponent/newsCarouselComponent.test.ts
+++ b/resources/js/alpine/newsCarouselComponent/newsCarouselComponent.test.ts
@@ -119,7 +119,7 @@ describe('Component: newsCarouselComponent', () => {
     render();
 
     // ACT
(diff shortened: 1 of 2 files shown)


Jamiras (approved):
I think the initial delay might be too much. According to my stopwatch, it was 21 seconds for the first panel, then 7 seconds for the other panels. 10 or maybe 15 seconds for the first panel might be better.

wescopeland:
Not a problem -- I've shortened the delay by about 7 more seconds in https://github.com/RetroAchievements/RAWeb/pull/1992/commits/bec313e74dd1875b3c07c837d43808dbdbbf6b0c.

PR #1995, opened 2023-11-06
show last game-related activity for claimant to moderators

Jamiras:
Adds an additional line for each claimant (only visible to Moderators) indicating the last time the claimant has done anything related to the game. The most simple case is just loading the game, as that's required to do anything else.
image

With the move to v5, we no longer have activity data prior to v5 (about 14 Oct 2023), so if no playing activity is seen, we look to see when the last achievement edit or code note was made. As time progresses, we'll have more v5 playing activity to look at.

image

image

For collaborations, a line is provided for each collaborator.
image

Moderators can use this information to ensure a claim is actively being worked on. This is particularly important if a claim gets extended as the claim is preventing other users who might be interested in the game from having a chance at it.


public/gameInfo.php
wescopeland, line 1: Functionally, everything seems fine.

This is a lot of new logic. Any objections to putting this in a new Blade component rendered by _gameInfo.php_ as opposed to _gameInfo.php_ itself?

wescopeland (approved):

PR #1996, opened 2023-11-06
Prevent exception viewing expiring claims without an active claim

Jamiras:
Fixes 
```
Undefined array key \"Expired\" at /home/forge/retroachievements.org/releases/2023-11-05T163253-5.1.2-749d077a/public/expiringclaims.php:32
```

Also restricts access of the page to match permissions for the menu item: 
https://github.com/RetroAchievements/RAWeb/blob/de0cf4264fffe77ab7df7c0ea3dfb0619e572e8c/resources/views/components/menu/management.blade.php#L36-L38


public/expiringclaims.php
```diff
@@ -5,9 +5,14 @@
 use App\Community\Enums\ClaimSorting;
 use App\Community\Enums\ClaimSpecial;
 use App\Community\Enums\ClaimType;
+use App\Community\Models\AchievementSetClaim;
 
 authenticateFromCookie($user, $permissions, $userDetails);
 
+if (!request()->user() || !request()->user()->can('manage', AchievementSetClaim::class)) {
```
luchaos, line 12: could be shortened to if i'm not mistaken:
```php
!request()->user()?->can('manage', AchievementSetClaim::class)
```

luchaos (approved):

public/expiringclaims.php
```diff
@@ -29,8 +34,10 @@
     $expired = $expiring = 0;
     if (isset($user)) {
         $expiringClaims = getExpiringClaim($user);
-        $expired = (int) $expiringClaims["Expired"];
-        $expiring = (int) $expiringClaims["Expiring"];
+        if (!empty($expiringClaims)) {
+            $expired = (int) $expiringClaims["Expired"];
+            $expiring = (int) $expiringClaims["Expiring"];
+        }
```
luchaos, line 40:
```php
$expired = (int) ($expiringClaims["Expired"] ?? 0);
$expiring = (int) ($expiringClaims["Expiring"] ?? 0);
```

PR #1998, opened 2023-11-08
Convert "My Sets" developer page to blade

Jamiras:
Not sure exactly why, but using a bunch of smaller queries makes the page load approximately 8x faster than before.

No longer shows "X of Y" if X equals Y. Just shows "Y" instead.

Removed TruePoints and Last Updated. Added Players and Progress.

Before:
image

After:
image

Also adds a bunch of sorting options and a breadcrumb banner:
image


app/Helpers/render/user.php
```diff
@@ -170,3 +170,26 @@ function RenderCompletedGamesList(
...
+    }
+
+    // Return next crumb (i.e `» text`), either as a link or not
+    $nextCrumb = fn ($text, $href = ''): string => " &raquo; " . ($href ? "<a href='$href'>$text</a>" : "<span class='font-bold'>$text</span>");
+
+    $html = "<a href='/userList.php'>All Users</a>"
+        . $nextCrumb($user->User, "/user/{$user->User}");
+
+    return $html;
+}
```
wescopeland, line 195: If we perceive a Blade component as, technically, a function which accepts some props as _arguments_ and then _returns_ HTML, it also seems like this should be a Blade component.

resources/views/community/components/developer/dev-games-page.blade.php
```diff
@@ -0,0 +1,185 @@
...
+    if ($sortOrder === $value) {
+        return "<a href='?sort=-$value$suffix'>$text &#x25B2;</a>";
+    }
+
+    if ($sortOrder == "-$value") {
+        return "<a href='?sort=$value$suffix'>$text &#x25BC;</a>";
+    }
+
+    return "<a href='?sort=$value$suffix'>$text</a>";
+};
```
wescopeland, line 29: Similar feedback as above, this is a function that accepts some props and returns HTML. Seems like it's actually doing the job of a component.

Blazekickn:
is retropoints or retro ratio too expensive to display? Its extremely useful as a quick rough estimate of how difficult a mastery is and it'd be unfortunate if it was cut

app/Platform/RouteServiceProvider.php
```diff
@@ -72,6 +73,7 @@ protected function mapWebRoutes(): void
             // Route::resource('leaderboard', LeaderboardController::class)->only('show');
 
             // Route::get('user/{user}/history', [PlayerHistoryController::class, 'show'])->name('user.history');
+            Route::get('user/{user}/dev-games', DeveloperGameListController::class)->name('user.dev-games');
```
wescopeland, line 76: Wondering if "dev" should be its own path parameter rather than part of the slug, eg:

`user/{user}/dev/sets`
`user/{user}/dev/games`

This question is coming up because I'm trying to figure out the best path for the Recently Obtained Achievements stuff, eg:

`user/{user}/dev/recent-unlocks`

Jamiras:
It's not expensive. There's just no room. I added two columns and removed two and a half.

There's a myriad of reasons why RetroPoints are inaccurate. It's doesn't reflect **difficulty** - it reflects **attrition** - how much effort is a player willing to put into the game. A hard game does require a lot of practice to earn the hard achievements, but a set with a high RetroRatio may just require time and not effort. It may require multiple playthroughs as part of the set, or it may just take hundreds of hours to complete all of the achievements. Even sets that don't take hundreds of hours may become boring before the player finishes them, thus inflating the RetroRatio. A bad game may push users away after the first few achievements, also inflating RetroRatio. Finally, revisions inflate RetroRatio as the new achievements only reflect the earn rate of people who play after the new achievements are added.

I think I can squeeze RetroRatio back in, but I'm going to leave out RetroPoints.

app/Helpers/render/user.php
Jamiras: The was just copied from `renderGameBreadcrumb`, but I could see it as a Blade component.

resources/views/community/components/developer/dev-games-page.blade.php
Jamiras: This is just a helper function for managing the sort parameters. I don't think it makes sense as a Blade component. I'm also considering ways to replace the sort implementation with a filter widget similar to what you've done on the beaten list.

Blazekickn:
Im aware retropoints are inaccurate, hence why i said rough estimate. Personally points are way more useful to me than ratio, since its easier to calculate the ratio via the points than it is the other way around. Id probably remove ratio or maybe players since that isn't super useful in most cases

Jamiras:
But players are important: https://discord.com/channels/310192285306454017/1171507488902881290/1171511417887805440
No one else has said anything about RetroPoints or RetroRatio.

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+                            :gameId="$game['ID']"
+                            :gameTitle="$game['Title']"
+                            :gameImageIcon="$game['ImageIcon']"
+                        />
+                    @endif
+                    </td>
+
+                    @if ($game['NumAuthoredAchievements'] == $game['achievements_published'])
+                        <td class='text-right'>{{ $game['NumAuthoredAchievements'] }}</td>
+                        <td class='text-right'>{{ $game['NumAuthoredPoints'] }}</td>
```
wescopeland, line 83:
```suggestion
                        <td class='text-right'>{{ localized_number($game['NumAuthoredPoints']) }}</td>
```

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+                    @else
+                        <td class='text-right'>{{ $game['NumAuthoredAchievements'] }} of {{ $game['achievements_published'] }}</td>
+                        <td class='text-right'>{{ $game['NumAuthoredPoints'] }} of {{ $game['points_total'] }}</td>
```
wescopeland, line 86:
```suggestion
                        <td class='text-right'>{{ localized_number($game['NumAuthoredPoints']) }} of {{ localized_number($game['points_total']) }}</td>
```

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+
+                    @if ($game['leaderboards_count'] == 0)
+                        <td></td>
+                    @elseif ($game['NumAuthoredLeaderboards'] == $game['leaderboards_count'])
+                        <td class='text-right'>{{ $game['NumAuthoredLeaderboards'] }}</td>
+                    @else
+                        <td class='text-right'>{{ $game['NumAuthoredLeaderboards'] }} of {{ $game['leaderboards_count'] }}</td>
+                    @endif
+
+                    <td class='text-right'>{{ $game['players_total'] }}</td>
```
wescopeland, line 99:
```suggestion
                    <td class='text-right'>{{ localized_number($game['players_total']) }}</td>
```

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+                </tr>
+            @endforeach
+            @if ($count > 1)
+                <tr>
+                    <td><b>Total:</b> {{ $count }} games</td>
+                    <td class='text-right'><b>{!! localized_number($achievementCount) !!}</b></td>
+                    <td class='text-right'><b>{!! localized_number($pointCount) !!}</b></td>
+                    <td></td>
+                    <td class='text-right'><b>{!! localized_number($leaderboardCount) !!}</b></td>
+                    <td></td>
```
wescopeland, line 137: Would definitely be interested in seeing sum of players here.

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+
+<x-app-layout
+    pageTitle="{{ $user->User }} - Developed Sets"
+    pageDescription="Games developed by {{ $user->User }}"
+>
+    <x-user.breadcrumbs :targetUsername="$user->User" currentPage="Developed Sets" />
+
+    <div class="mt-3 -mb-3 w-full flex gap-x-3">
+        {!! userAvatar($user->User, label: false, iconSize: 48, iconClass: 'rounded-sm') !!}
+        <h1 class="mt-[10px] w-full">{{ $user->User }}'s Developed Sets</h1>
```
wescopeland, line 18: Not sure if it matters too much, but there is a subtle difference here from how the Completion Progress page renders its header.

**This page**
Screenshot 2023-11-08 at 4 15 09 PM

**Completion Progress page**
Screenshot 2023-11-08 at 4 15 16 PM

On the Completion Progress page, if the username ends with an "s", an apostrophe is added without an additional "s" at the end.

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+            <tr>
+                <th style='width:34%'>Title</th>
+                <th style='width:12%' class='text-right'>Achievements</th>
+                <th style='width:10%' class='text-right'>Points</th>
+                <th style='width:10%' class='text-right'>RetroRatio</th>
+                <th style='width:10%' class='text-right'>Leaderboards</th>
+                <th style='width:8%' class='text-right'>Players</th>
+                <th style='width:8%' class='text-right'>Tickets</th>
+                @if ($userProgress !== null)
+                    <th style='width:8%' class='text-right'>Progress</th>
```
wescopeland, line 48: It took me a second to understand whose or what progress this is (mine?, the developer I'm looking at?, the set's playerbase?, etc).

This column is fun and I like it, but because it took me a second to fully understand I'm afraid it leans towards being ambiguous. I know there is an aversion to putting labels like "Your" or "Mine" on things, but it does feel appropriate here to reduce that ambiguity.

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+                <tr>
+                    <td><b>Total:</b> {{ $count }} games</td>
+                    <td class='text-right'><b>{!! localized_number($achievementCount) !!}</b></td>
+                    <td class='text-right'><b>{!! localized_number($pointCount) !!}</b></td>
+                    <td></td>
+                    <td class='text-right'><b>{!! localized_number($leaderboardCount) !!}</b></td>
+                    <td></td>
+                    <td class='text-right'><b>{!! localized_number($ticketCount) !!}</b></td>
+                    @if ($userProgress !== null)
+                        <td></td>
```
wescopeland, line 140: Average progress here would also be very interesting to see. I think this column as you've built it is really cool, this is another thing that could make it shine.

resources/views/community/components/developer/sets-meta-panel/sort-order.blade.php
wescopeland, line 1: When enabling a sort while in "Group by console" mode, the console names reappear.

Screenshot 2023-11-08 at 4 23 16 PM

wescopeland:
> Not sure exactly why, but using a bunch of smaller queries makes the page load approximately 8x faster than before.

I experienced this in #1999 as well: https://github.com/RetroAchievements/RAWeb/pull/1999/files#diff-09011a6920853a763a114b2c30c48af5e3352c02b1718f01f6a74b6818bf14f1R139

I think it's because the joins produce huge tables in memory, but I'm not 100% sure.

resources/views/community/components/developer/sets-page.blade.php
Jamiras: We need union, not sum. If one player played each of 5 sets, the total number of players is still 1, not 5.

app/Platform/Controllers/DeveloperGameListController.php
```diff
@@ -0,0 +1,191 @@
...
+            'progress' => function ($a, $b) {
+                return $a['CompletionPercentage'] <=> $b['CompletionPercentage'];
+            },
+        };
+        usort($games, $sortFunction);
+        if ($reverse) {
+            $games = array_reverse($games);
+        }
+
+        return view('community.components.developer.sets-page', [
```
wescopeland, line 182: The controller is in the Platform domain, seems like this should be:

`platform.components.developer.sets-page`

resources/views/community/components/developer/sets-page.blade.php
```diff
@@ -0,0 +1,153 @@
...
+    'consoles' => [],
+    'games' => [],
+    'sortOrder' => 'title',
+    'filterOptions' => [],
+    'userProgress' => null,
+])
+
+<x-app-layout
+    pageTitle="{{ $user->User }} - Developed Sets"
+    pageDescription="Games developed by {{ $user->User }}"
```
wescopeland, line 12: Apologies for leaving comments so sporadically, I'm basing the recent unlocks work on a lot of your work here.

```suggestion
    pageDescription="View achievement sets developed by {{ $user->User }} for various games on RetroAchievements"
```

This meta description is slightly better for SEO (this page will almost certainly get crawled).

resources/views/community/components/developer/sets-page.blade.php
Jamiras: Google says both ways are correct, but seems to favor using the `'s` for singular proper names.

Merriam-Webster says the second form implies that there's more than one Jamiras (as in a family name): "The Jones' house" is a house belonging to the Jones family. "Jones's dog" is a dog belonging to a guy named Jones.

The Associated Press requires that "any plural noun, including names that end with S, should not include an additional S after the apostrophe", though I don't understand how a name ending in S is automatically a plural noun. Maybe this one is misleading?

I also found this:
> Major style guides like APA Publication Manual, MLA, and Chicago recommend using apostrophes and extra letters S. This rule applies to both proper and common nouns.
> Other formal writing style guides only recommend adding an apostrophe after singular nouns ending in S, specifically proper nouns. This structure is also more common in informal writing or a hip style of texting.

wescopeland: Thanks for sharing; it's hard to argue against APA or MLA guidelines. I'll make a mental note to adjust this on the Completion Progress page later.

resources/views/community/components/developer/sets-page.blade.php
Jamiras: It's the logged in user's progress, and is meant to be propagated to pretty much every game list in the site as a replacement for the color script. As we haven't standardized how we want that to be displayed, I'm fine with removing it from this PR.

app/Platform/Controllers/DeveloperGameListController.php
Jamiras: I really don't understand the domains. Users are in Community, and there's already a developers component subdirectory there. And there's already a DeveloperController in the Platform domain, though it doesn't seem to be tied to anything.

resources/views/community/components/developer/sets-page.blade.php
wescopeland: I think it's neat and enhances the page. This minor ambiguity thing is the only thing about it that may need a slight change if it stays (my initial assumption was the bar represented the average progress of all players on the set).

Jamiras: Would adding a tooltip help?
image

wescopeland: Yep, I think so. Perhaps on the column title label as well?

app/Platform/Controllers/DeveloperGameListController.php
```diff
@@ -0,0 +1,191 @@
...
+
+class DeveloperGameListController extends Controller
+{
+    public function __invoke(Request $request): View
+    {
+        $username = $request->route('user');
+        $user = User::firstWhere('User', $username);
+        if ($user === null) {
+            abort(404);
+        }
```
wescopeland, line 26: When attempting to access a banned user's page, a division by zero error is thrown. May be worth adding something like a `getCanViewTargetUser()` function similar to the completion progress page.

Jamiras: I suspect you're using a very specific banned user, who has a set worth 0 points. Good find. I've corrected for that.

Banned users still have accessible profiles, so their list of dev'd sets should still be accessible as well.

wescopeland: Yes :)

Got it - I actually did not realize those dev profiles were still accessible (I assumed they weren't). Marking as resolved.

wescopeland (approved):
None of my other commentary is a showstopper. This is substantially better than what's out there on prod at the moment.

I'm still basing my current work off a mix of the implementation here & the Completion Progress page. If I find anything else pressing I'll leave a comment (and apologies again for leaving feedback so sporadically).

I think @luchaos should weigh in on the domains comment (https://github.com/RetroAchievements/RAWeb/pull/1998#discussion_r1387264568). I am also curious on this & the answer likely has implications for future work.

Blazekickn:
> But players are important: https://discord.com/channels/310192285306454017/1171507488902881290/1171511417887805440 Devs want to see how many people are playing their games without having to open every individual page.
> 
> No one else has said anything about RetroPoints or RetroRatio.

Fair enough, I'm viewing this from the perspective of a player who uses developer set lists the same way i use console set lists and hubs - to find new games to play. I'd still like Retro Points to be squeezed in somehow, but if not im sure a greasemonkey script could probably show it so whatever i can't stop you.

app/Platform/Controllers/DeveloperGameListController.php
luchaos: Community vs Platform domains are meant to distinguish between core platform features and community "addons". Developers are a weird thing in-between, i think we can move most of it to platform though?

Minor: `DeveloperGamesController` - `List` can be omitted imo

app/Platform/RouteServiceProvider.php
luchaos: Yup, me too. had something similar in mind:

`/user/{user}/developer` as the main developer profile
`/user/{user}/developer/X` as sub-pages for developer profile related resources and detail pages.
As mentioned, i'd try to keep lingo out of routes and in copy and not shorten "developer" to "dev" to make it sound somewhat official?

luchaos: I think the most fitting route name here would be `developer.games` - substituting `user` with `developer` as an alias in this context 🤔

wescopeland: That definitely sounds agreeable to me. Maybe `/user/{user}/developer` takes on individualdevstats.php ?

luchaos: > Maybe /user/{user}/developer takes on individualdevstats.php

Sounds good to me - at least the most applicable atm.
It could house the developer-specific awarded badges for contributions in the future.

luchaos: Throwing `creator` profile in the ring here again, too - just in case we plan on incorporating writers and artists on such pages and keep the "developer" nomenclature specifically for "trigger authors". May as well be just "developers". That's for another discussion - might affect how we want to name routes, though.

luchaos (approved):

PR #1999, opened 2023-11-08
perf(beaten-leaderboard): use aggregated data

wescopeland:
This PR results in beaten game leaderboards loading significantly faster than the current implementation in `main`. The unfiltered leaderboard goes from loading in 11s to loading in less than 200ms.

As a reminder, the route can be accessed at `/ranking/beaten-games`.

> **Note**
> I've added a command which lets you either build aggregates for a single user or queue up a full sync of the `player_stats` table.
> 
> **For a single user**, you can do `sail artisan ra:platform:player:update-stats Jamiras`.
> **For a full sync**, you can exclude the username/userid argument and simply do `sail artisan ra:platform:player:update-stats`.

## How to test this PR
* `sail artisan migrate`
* Test the command on a single user: `sail artisan ra:platform:player:update-stats Jamiras`. Visit `/ranking/beaten-games`.
* Prepare to do a full sync by running `sail artisan horizon` in another terminal tab.
* Queue a full sync by executing `sail artisan ra:platform:player:update-stats`. Visit `localhost:64000/horizon` to watch its progress.
* Visit `/ranking/beaten-games`.

## Theory of operation
Presently, beaten game leaderboards are constructed by performing complex queries where multiple tables are joined onto `SiteAwards`. It is much faster to instead build the leaderboards by using a new table containing rows with all the aggregate data computed in advance, similar to `player_games`, etc.

Let's pretend we want to put together a beaten game leaderboard for system_id 6 and only include retail games.

Our new table can contain the aggregate award counts by user_id for this system_id. Rows in the table might look like:

```
user_id = 1, system_id = 6, type = "games_beaten_hardcore_retail", value = 5
user_id = 2, system_id = 6, type = "games_beaten_hardcore_retail", value = 7
user_id = 3, system_id = 6, type = "games_beaten_hardcore_retail", value = 2
```

Assume `value` is the user's beaten game count. From these rows, we can build our leaderboard using a simple SQL query.

Now, for a more complex scenario. Let's pretend we want both retail games and hacks in our leaderboard view, again for system_id = 6.

```
user_id = 1, system_id = 6, type = "games_beaten_hardcore_retail", value = 5
user_id = 1, system_id = 6, type = "games_beaten_hardcore_hack", value = 5
user_id = 2, system_id = 6, type = "games_beaten_hardcore_retail", value = 7
user_id = 3, system_id = 6, type = "games_beaten_hardcore_retail", value = 2
user_id = 3, system_id = 6, type = "games_beaten_hardcore_hack", value = 1
```

In this case, we query the rows where system_id = 6 and types are "games_beaten_hardcore_retail" or "games_beaten_hardcore_hack". We sum the values of rows for the same user_id, and again build our leaderboard.

Finally, what if we don't care about the system_id at all, and only want the global leaderboard values?

```
user_id = 1, system_id = null, type = "games_beaten_hardcore_retail", value = 50
user_id = 2, system_id = null, type = "games_beaten_hardcore_retail", value = 70
user_id = 3, system_id = null, type = "games_beaten_hardcore_retail", value = 20
```

Our aggregate values do not store a system_id for total sums of all systems. This lets us bypass doing a lot of math on many possible rows in the table.

## How it works
* A new table, `player_stats`, is added (thanks @luchaos for guidance on this).

The table structure is:

```sql
CREATE TABLE `player_stats` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `user_id` bigint(20) unsigned NOT NULL,
  `system_id` int(10) unsigned DEFAULT NULL, -- if NULL, don't filter leaderboards by system_id
  `type` varchar(255) NOT NULL, -- PlayerStatType.php: games_beaten_hardcore_retail, games_beaten_hardcore_demo, etc.
  `last_game_id` bigint(20) unsigned DEFAULT NULL, -- the most recent game that affected the row
  `value` int(11) NOT NULL DEFAULT 0, -- how many games were beaten by the user for the system_id & type
  `created_at` timestamp NULL DEFAULT NULL,
  `updated_at` timestamp NULL DEFAULT NULL, -- datetime of award for most recent game that affected the row (ie: backdated)
)
```

* A new action, `UpdatePlayerStats`, has been added which is responsible for aggregating and storing stats for a given `User`.

* When `PlayerBadgeAwarded`, `PlayerBadgeLost`, or `PlayerRankedStatusChanged` events are dispatched, an `UpdatePlayerStatsJob` is queued.

* `BeatenGamesLeaderboardController` has been updated to use the new aggregate data from the `player_stats` table.

## UX Changes

* The checkbox filters have been dramatically simplified to a small series of radio buttons. These radio buttons mirror the mastery rankings that are presented monthly in RANews. "Retail" is the default option.

Screenshot 2023-11-11 at 10 03 48 AM

* There was a bug in the original implementation where the "Jump to your ranking's page" could take you to the wrong page. This has been fixed.


resources/views/platform/beaten-games-leaderboard-page.blade.php
```diff
@@ -18,6 +19,7 @@
     <x-beaten-games-leaderboard.meta-panel
         :allSystems="$allSystems"
         :gameKindFilterOptions="$gameKindFilterOptions"
+        :leaderboardKind="$leaderboardKind"
```
Jamiras, line 22: There seems to be a slight discrepancy between the aggregated and non-aggregated data.

master:
image

this branch:
image

wescopeland:
This PR is now ready for review.

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -90,136 +85,213 @@ public function __invoke(Request $request): View
...
+
+        // As a safeguard, set everything to false by default.
+        $gameKindFilterOptions = array_fill_keys($allFilterKeys, false);
+
+        // Show retail games only by default. It will be the default filter choice.
+        $selectedKind = $validatedData['filter']['kind'] ?? 'retail';
+
+        switch ($selectedKind) {
+            case 'retail':
+                $gameKindFilterOptions['retail'] = true;
```
Jamiras, line 191: Shouldn't retail also contain unlicensed? They were still sold at the store.

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -90,136 +85,213 @@ public function __invoke(Request $request): View
...
+        $systemIdCondition = $targetSystemId ? "AND r2.system_id = {$targetSystemId}" : "AND r2.system_id IS NULL";
+        $lastBeatenDateSubquery = DB::raw("(
+            SELECT r2.updated_at
+            FROM rankings AS r2
+            WHERE r2.user_id = rankings.user_id
+                {$systemIdCondition}
+                AND r2.type IN ({$typeBindings})
+            ORDER BY r2.updated_at DESC
+            LIMIT 1
+        ) AS last_beaten_date");
```
Jamiras, line 151: Why does this take two subqueries?
```
> select `user_id`, ( SELECT r1.game_id FROM rankings AS r1 WHERE r1.user_id = rankings.user_id AND r1.system_id IS NULL AND r1.type IN ('games_beaten_hardcore_retail') ORDER BY r1.updated_at DESC LIMIT 1 ) AS most_recent_game_id, ( SELECT r2.updated_at FROM rankings AS r2 WHERE r2.user_id = rankings.user_id AND r2.system_id IS NULL AND r2.type IN ('games_beaten_hardcore_retail') ORDER BY r2.updated_at DESC LIMIT 1 ) AS last_beaten_date, RANK() OVER (ORDER BY SUM(value) DESC) as rank_number, ROW_NUMBER() OVER (ORDER BY SUM(value) DESC) as leaderboard_row_number, SUM(value) as total_awards from `rankings` where `system_id` is null and `type` in ('games_beaten_hardcore_retail') group by `user_id` order by `total_awards` desc, `last_beaten_date` asc
...
39794 rows in set (0.454 sec)
```
vs
```
> SELECT sub.user_id, j.game_id, j.updated_at, sub.total, ROW_NUMBER() OVER(ORDER BY sub.total DESC, sub.updated_at ASC) FROM (SELECT r.user_id, r.system_id, SUM(r.value) AS total, MAX(r.updated_at) AS updated_at FROM rankings r WHERE r.system_id IS NULL AND r.type IN ('games_beaten_hardcore_retail') GROUP BY r.user_id, r.system_id ORDER BY total DESC) AS sub INNER JOIN rankings j ON sub.user_id=j.user_id AND sub.updated_at = j.updated_at AND j.system_id IS NULL ORDER BY sub.total DESC, sub.updated_at ASC
...
39839 rows in set (0.103 sec)
```

app/Platform/Controllers/BeatenGamesLeaderboardController.php
```diff
@@ -90,136 +85,213 @@ public function __invoke(Request $request): View
...
+        }
+        if ($gameKindFilterOptions['prototypes']) {
+            $includedTypes[] = RankingType::GamesBeatenHardcorePrototypes;
+        }
+        if ($gameKindFilterOptions['demos']) {
+            $includedTypes[] = RankingType::GamesBeatenHardcoreDemos;
+        }
 
-        $userRankingData = $result->isEmpty() ? null : $result->get(0);
+        return $includedTypes;
```
Jamiras, line 247: is there still some way to select multiple types?

app/Platform/Listeners/DispatchUpdatePlayerRanksJob.php
```diff
@@ -0,0 +1,33 @@
...
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class DispatchUpdatePlayerRanksJob implements ShouldQueue
+{
+    public function handle(object $event): void
+    {
+        $user = null;
+
+        switch ($event::class) {
+            case PlayerGameMetricsUpdated::class:
```
Jamiras, line 18: it looks like you're listening for `PlayerBadgeAwarded` and `PlayerBadgeLost`, not `PlayerGameMetricsUpdated`

app/Platform/Controllers/BeatenGamesLeaderboardController.php
wescopeland: Yes. The "All" radio button selects all types.

resources/views/platform/beaten-games-leaderboard-page.blade.php
wescopeland: I noticed this as well. I believe the latest developer dumps have some discrepancies between `player_games` and `SiteAwards`.

In https://github.com/RetroAchievements/RAWeb/pull/1999/commits/e4c2e5e5554d5a0669654338b46bf5dbe3676ac0#diff-d1cd771d569c8de22cd2a5fbb203f1679cdd74af5e55f288f169697ed8bbeff8, I changed `UpdatePlayerRanks` to use `SiteAwards` instead of `player_games` before building aggregates. It occurred to me that `SiteAwards` is likely the better source of truth for building these rankings. I believe it's possible I pushed this commit up right after you did a full sync operation.

app/Platform/Listeners/DispatchUpdatePlayerRanksJob.php
wescopeland: Want to confirm if this still needs to change now that `UpdatePlayerRanks` is using `SiteAwards` as its source of truth for aggregates:

https://github.com/RetroAchievements/RAWeb/pull/1999/files#diff-d1cd771d569c8de22cd2a5fbb203f1679cdd74af5e55f288f169697ed8bbeff8R28

app/Platform/Controllers/BeatenGamesLeaderboardController.php
wescopeland: You're right - using a single subquery in this way is significantly faster. I'm seeing load times go from roughly 500ms to 150ms for the overall leaderboard. The NES leaderboard now also loads in less than 100ms.

It took some tweaking to get the `rank_number` and `leaderboard_row_number` values right, but your example was enough to get me started. Thanks!

app/Platform/Controllers/BeatenGamesLeaderboardController.php
wescopeland: You're right. I hadn't considered that. Updated in latest.

database/migrations/platform/2023_11_06_000000_create_rankings_table.php
```diff
@@ -0,0 +1,43 @@
...
+    public function up(): void
+    {
+        /**
+         * A flexible structure to store different types of ranking data within the same table.
+         * The 'type' column is a string that specifies the category of the ranking, such as
+         * 'games_beaten_retail', and the 'value' column stores the numerical value associated
+         * with that ranking. This allows for a wide range of aggregate ranking types to be managed
+         * dynamically without requiring schema changes for each new ranking category.
+         */
+        Schema::create('rankings', function (Blueprint $table) {
```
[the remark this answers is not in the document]
luchaos: I suggested rankings but agree that the name is not as fitting as i thought.
player_stats? they can be used wothout system_id iirc (not sure if we should)

database/migrations/platform/2023_11_06_000000_create_rankings_table.php
```diff
@@ -0,0 +1,43 @@
...
+            $table->bigIncrements('id');
+            $table->unsignedBigInteger('user_id');
+            $table->unsignedInteger('system_id')->nullable(); // useful for system-specific leaderboards
+            $table->unsignedBigInteger('game_id')->nullable(); // useful for showing most recent game which affected the value
```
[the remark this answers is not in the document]
luchaos: do we even need this column for something in particular?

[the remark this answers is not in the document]
wescopeland: Yes, the leaderboard displays the most recent game that contributed to the user's ranking.

Screenshot 2023-11-11 at 11 36 21 AM

database/migrations/platform/2023_11_06_000000_create_rankings_table.php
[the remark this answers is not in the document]
wescopeland: Happy to go with either `player_system_stats` or `player_stats`.
`system_id` is indeed nullable, and in this leaderboard context is being used for "overall" values. However, it doesn't necessarily have to be used in that way in all contexts.

database/migrations/platform/2023_11_06_000000_create_rankings_table.php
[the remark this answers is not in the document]
luchaos: i see. last_game_id sounds fitting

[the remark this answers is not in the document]
wescopeland: In latest, renamed the column to `last_game_id` and moved to be after `type`.

database/migrations/platform/2023_11_06_000000_create_rankings_table.php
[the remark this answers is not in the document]
wescopeland: In latest, renamed `rankings` to `player_stats`.

I've also renamed all classes, etc that previously used the "ranks" or "rankings" nomenclature to instead use the "player stats" nomenclature.

The command has also changed as a result:
* `sail artisan ra:platform:player:update-stats`

app/Platform/Listeners/DispatchUpdatePlayerRanksJob.php
Jamiras: I'm referring to this code: https://github.com/RetroAchievements/RAWeb/blob/3775e153fab8d826fe1005afc595d87ae2ebdde1/app/Platform/EventServiceProvider.php#L74-L81

This listener will get called with a `PlayerBadgeAwarded` event or a `PlayerBadgeLost` event. I don't see anywhere where it would get called with a `PlayerGameMetricsUpdated` event.

wescopeland: Ah, I think I understand now. I've removed the `DispatchUpdatePlayerStatsJob` from these two in latest and moved it over to `PlayerGameMetricsUpdated`.

Jamiras: If the queries rely strictly on the `SiteAwards` table, wouldn't it be more succinct to have the listener watching for `PlayerBadgeAwarded` and `PlayerBadgeLost`?

`UpdatePlayerGameMetrics` disptches a `RevalidateAchievementsSetBadgeEligibility` action, which would raise the `PlayerBadgeAwarded` or `PlayerBadgeLost` when appropriate, and you wouldn't be regenerating the stats any time the player unlocked an achievement.

wescopeland: Sorry, I was getting confused about multiple things simultaneously, but I think I understand the feedback now.

I've updated `EventServiceProvider.php` so it's listening for `PlayerBadgeAwarded` and `PlayerBadgeLost` again (and not `PlayerGameMetricsUpdated`). I've also updated `DispatchUpdatePlayerStatsJob` to have cases for `PlayerBadgeAwarded` and `PlayerBadgeLost` events.

wescopeland: Latest commit https://github.com/RetroAchievements/RAWeb/pull/1999/commits/763b0405633a16800ff7003a0eb9887bd6cc9980 resolves an issue where `PlayerBadgeLost` would instantly fail due to a race condition from the badge itself no longer existing.

`PlayerBadgeLost` is now dispatched with a deconstructed badge instance:

```php
private function dispatchBadgeLostEvent(PlayerBadge $badge): void
{
    PlayerBadgeLost::dispatch(
        $badge->user,
        $badge->AwardType,
        $badge->AwardData,
        $badge->AwardDataExtra,
    );
}
```

app/Platform/Listeners/DispatchUpdatePlayerStatsJob.php
```diff
@@ -0,0 +1,36 @@
...
+class DispatchUpdatePlayerStatsJob implements ShouldQueue
+{
+    public function handle(object $event): void
+    {
+        $user = null;
+
+        switch ($event::class) {
+            case PlayerBadgeAwarded::class:
+            case PlayerRankedStatusChanged::class:
+                $user = $event->playerBadge->user;
```
Jamiras, line 21: This feels incorrect. `PlayerRankedStatusChanged` does not have a `playerBadge` member. I think you meant to group it with the `PlayerBadgeLost` handler.

wescopeland: Yep, silly typo. Should be all set now.

PR #2000, opened 2023-11-08
feat(active-players): support OR searches with a pipe operator

wescopeland:
Allows users to search by strings like "developing|fixing|inspecting".


Jamiras (approved):
Minor: empty strings are not ignored, so if you pause on `first|` before typing `second`, all the rows are returned.

wescopeland:
@Jamiras Good catch, resolved in latest. Though approved, I'll let you merge just in case you'd like to verify.

PR #2007, opened 2023-11-10
add admin tool to migrate achievements between games

Jamiras:
Eliminates needs for direct database access to move achievements to a subset, and ensures all appropriate aggregate data gets updated.

### DIFF
--- a/public/admin.php
+++ b/public/admin.php
@@ -106,7 +106,7 @@ if ($action === 'migrate-achievement') {
     }
 
     // determine which game(s) the achievements are coming from
-    $oldGames = Achievement::whereIn('ID', $achievementIds)->select(['GameID'])->distinct()->pluck('GameID')->toArray();
+    $oldGames = Achievement::whereIn('ID', $achievementIds)->select(['GameID'])->distinct()->pluck('GameID');
 
     // associate the achievements to the new game
     Achievement::whereIn('ID', $achievementIds)->update(['GameID' => $gameId]);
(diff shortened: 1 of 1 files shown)


public/admin.php
```diff
@@ -92,6 +98,43 @@
...
+    Achievement::whereIn('ID', $achievementIds)->update(['GameID' => $gameId]);
+
+    // add an audit comment to the new game
+    addArticleComment('Server', ArticleType::GameModification, $gameId,
+        "$user migrated " . Str::plural('achievement', count($achievementIds)) . ' ' .
+        implode(',', $achievementIds) . ' from ' .
+        Str::plural('game', count($oldGames)) . ' ' . implode(',', $oldGames) . '.');
+
+    // ensure player_game entries exist for the new game for all affected users
+    foreach (PlayerAchievement::whereIn('achievement_id', $achievementIds)->select(['user_id'])->distinct()->pluck('user_id')->toArray() as $userId) {
```
luchaos, line 121: `->toArray()` shouldn't be necessary as Collections implement the necessary iterator and ArrayAccess interfaces and can be used in `foreach` loops directly.

public/admin.php
```diff
@@ -92,6 +98,43 @@
...
 
+if ($action === 'migrate-achievement') {
+    $achievementIds = explode(',', requestInputSanitized('a'));
+    $gameId = requestInputSanitized('g');
+    if (!Game::where('ID', $gameId)->exists()) {
+        return back()->withErrors('Unknown game');
+    }
+
+    // determine which game(s) the achievements are coming from
+    $oldGames = Achievement::whereIn('ID', $achievementIds)->select(['GameID'])->distinct()->pluck('GameID')->toArray();
```
luchaos, line 109: Same as below ad `->toArray()`

luchaos (approved):

public/admin.php
Jamiras: This one is required, because $oldGames is also passed to `implode()`:
```
implode(): Argument #2 ($array) must be of type ?array, Illuminate\Support\Collection given
```

luchaos: can be done with collections, too, e.g. `$oldGames->implode(',')`

PR #2009, opened 2023-11-11
add admin tool to duplicate unlocks

Jamiras:
Implements most of #1806.

The simplest use case is to put one achievement ID in the first field, and one in the second ID. All unlocks from the first achievement will be applied to the second achievement (if the user hasn't already earned the second achievement) using the timestamps from the first achievement.

Both fields also support CSV values.

If the first field contains a CSV, the player must have earned all achievements in the CSV to unlock the achievement(s) in the second field. The latest unlock timestamp will be applied to the target achievement(s).

If the second field contains a CSV, the player will be awarded multiple achievements if they have all of the achievement(s) in the first field. The same timestamp will be applied to all of the target achievement(s).

#1806 mentions the ability to put provide a list of inputs and a list of outputs and have the simple use case applied to each pair in the list. This was not implemented, as it conflicts with the aforementioned CSV support. If two achievements were put into each field, the code wouldn't know whether it was a list mapping or if both of the first achievements were requires to unlock both of the target achievements.


public/admin.php
```diff
@@ -98,6 +99,44 @@
...
+        ->groupBy('user_id')
+        ->having('softcore_count', '=', $fromAchievementCount)
+        ->get();
+
+    // award the target achievements, copying the unlock times and hardcore state
+    $unlockerId = request()->user()->id;
+    foreach ($existing as $playerAchievement) {
+        $hardcore = ($playerAchievement->hardcore_count == $fromAchievementCount);
+        $timestamp = Carbon::parse($hardcore ? $playerAchievement->unlocked_hardcore_at : $playerAchievement->unlocked_softcore_at);
+        foreach ($toAchievementIds as $nextID) {
```
luchaos, line 124: minor: `$toAchievementId` instead of `$nextID`?

luchaos (approved):

public/admin.php
Jamiras: That's copied from https://github.com/RetroAchievements/RAWeb/blob/5c28b94e1c6b0d99b9ff2a41bc8f8be5743f6518/public/admin.php#L76-L84, but I can change it.

PR #2013, opened 2023-11-13
fix completion badges not getting revoked when resetting softcore achievements

Jamiras:
also adds a command to manually reset achievements from the command line (currently limited to local environment)


app/Platform/Actions/ResetPlayerProgress.php
```diff
@@ -47,13 +49,20 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
...
+                $playerBadge = $user->playerBadges()
                     ->where('AwardType', AwardType::Mastery)
                     ->where('AwardData', $achievement->game_id)
-                    ->where('AwardDataExtra', UnlockMode::Hardcore)
-                    ->delete();
+                    ->where('AwardDataExtra', $playerAchievement->unlocked_hardcore_at ? UnlockMode::Hardcore : UnlockMode::Softcore)
+                    ->first();
+                if ($playerBadge) {
+                    PlayerBadgeLost::dispatch($playerBadge);
+                    $playerBadge->delete();
```
luchaos, line 64: Event should follow the action to keep things in order

luchaos (approved):

app/Platform/Commands/ResetPlayerAchievement.php
```diff
@@ -0,0 +1,64 @@
...
+        parent::__construct();
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function handle(): void
+    {
+        if (!app()->environment('local')) {
+            $this->error('This command is only allowed in the local environment');
```
luchaos, line 32: Why is it restricted to local?

app/Platform/Actions/ResetPlayerProgress.php
Jamiras: This is following the logic in RevalidateAchievementSetBadgeEligibility: https://github.com/RetroAchievements/RAWeb/blob/c92999b87927ef4c0b2922720f6232f3e85317d6/app/Platform/Actions/RevalidateAchievementSetBadgeEligibility.php#L103-L104

Though there seems to be a problem with the lifetime of the badge even if it's deleted after raising the event: https://discord.com/channels/476211979464343552/1026595325038833725/1173661914556006431

app/Platform/Commands/ResetPlayerAchievement.php
Jamiras: It was a consideration, given the command is completely destructive and leaves no audit trail.

wescopeland (approved):
Verified locally. The command was also helpful for some of my own local testing, thank you.

Depending on who merges first, we'll have something that needs to be fixed with the `PlayerBadgeLost` dispatches added in this PR (as they are now being deconstructed in #1999).

PR #2014, opened 2023-11-14
feat: migrate inspector toolbox to blade, hide type additions to subsets

wescopeland:
This PR migrates the achievement inspector toolbox component (`/achievementinspector.php?g=1`) to Blade. It also hides the "Set Selected to Progression" and "Set Selected to Win Condition" buttons if the game is either a subset or a test kit.

In a subsequent PR, I will also adjust the logic for these cases in _achievementinfo.php_ and our API calls.

PR #2015, opened 2023-11-14
fix(achievementInfo): prevent adding types to subsets and test kits

wescopeland:
Related to #2014.

This PR prevents the user from adding types to an achievement if it is a subset or a test kit.

To test:
* You should be able to add types to `/achievement/9`.
* You should get an error if you try to add a type to `/achievement/241312`.
* You should always be able to remove a type, regardless of the game title.

### DIFF
--- a/public/achievementInfo.php
+++ b/public/achievementInfo.php
@@ -121,8 +121,21 @@ RenderOpenGraphMetadata(
 RenderContentStart($pageTitle);
 ?>
 <?php if ($permissions >= Permissions::Developer || ($permissions >= Permissions::JuniorDeveloper && $isAuthor)): ?>
+    <?php
+        $canHaveTypes = mb_strpos($gameTitle, "[Subset") === false && mb_strpos($gameTitle, "~Test Kit~") === false;
+    ?>
+
     <script>
     function updateAchievementDetails() {
+        const canHaveTypes = <?= $canHaveTypes ? 'true' : 'false' ?>;
+
+        const typeInputValue = $('#typeinput').val();
+
+        if (!canHaveTypes && typeInputValue) {
+            showStatusFailure('Error: Types are not supported on subsets or test kits.');
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


public/achievementInfo.php
```diff
@@ -121,8 +121,21 @@ function generateAchievementMetaDescription(
...
+    ?>
+
     <script>
     function updateAchievementDetails() {
+        const canHaveTypes = <?= $canHaveTypes ? 'true' : 'false' ?>;
+
+        const typeInputValue = $('#typeinput').val();
+
+        if (!canHaveTypes && typeInputValue) {
+            showStatusFailure('Error: Types are not supported on subsets or test kits.');
```
Jamiras, line 135: Recommend hiding the dropdown if the type is already None

public/achievementInfo.php
```diff
@@ -121,8 +121,21 @@ function generateAchievementMetaDescription(
 RenderContentStart($pageTitle);
 ?>
 <?php if ($permissions >= Permissions::Developer || ($permissions >= Permissions::JuniorDeveloper && $isAuthor)): ?>
+    <?php
+        $canHaveTypes = mb_strpos($gameTitle, "[Subset") === false && mb_strpos($gameTitle, "~Test Kit~") === false;
```
Jamiras, line 125: Suggest extending this to Events

Jamiras (commented):
Recommend hiding the dropd

public/achievementInfo.php
wescopeland: Updated to check for games that have a console ID of 101.

public/achievementInfo.php
wescopeland: Removed this new JS and set the dropdown to never appear if `$canHaveTypes` is falsy. Related SQL query we can execute after shipping: https://discord.com/channels/476211979464343552/1174126386349342901

Jamiras (approved):

PR #2024, opened 2023-11-16
fix(feed): dont show all players on player list refresh [V5.2]

wescopeland:
On its 5 minute refresh interval, the feed is loading all players rather than the filtered list for the developer's games. This PR ensures the list remains focused on only those game IDs.

### DIFF
--- a/resources/views/community/components/active-players/active-players.blade.php
+++ b/resources/views/community/components/active-players/active-players.blade.php
@@ -110,6 +110,10 @@ function activePlayersComponent() {
                 params.set('all', true);
             }
 
+            if (this.targetGameIds) {
+                params.set('targetGameIds', Object.values(this.targetGameIds));
+            }
+
             if (params.size > 0) {
                 requestUrl = `${requestUrl}?${params}`;
             }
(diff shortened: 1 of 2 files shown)


resources/views/community/components/active-players/active-players.blade.php
```diff
@@ -106,7 +106,7 @@ function activePlayersComponent() {
                 params.set('search', this.searchInput);
             }
 
-            if (getFullList || this.hasFetchedFullList) {
+            if (this.variant !== 'focused' && (getFullList || this.hasFetchedFullList)) {
```
Jamiras, line 109: This doesn't do what it needs to. `all` just says to return 5000 rows instead of 20. You need to pass in the gameIds that were used in the initial query.

wescopeland: Yep, you're right.

I think we should still keep this conditional, otherwise `?all=true` gets sent on the fetch URL. However, in latest, I've added some logic to send (and parse) the list of game IDs.

Jamiras (approved):

PR #2026, opened 2023-11-16
allow admins to manually unlock demoted achievements

Jamiras:
The `unlockAchievement` function has code that prevents unlocking Unofficial achievements. The `UnlockPlayerAchievementJob` has no similar qualms. Both were being called previously to ensure both new and old data structures were updated. The old data structure is no longer used, so there's no need to try to populate it.


wescopeland (approved):
Verified locally for both a hardcore and softcore unlock of a demoted achievement.

PR #2027, opened 2023-11-16
fix exception if game has leaderboards but no achievements

Jamiras:
for example: http://localhost:64000/user/s0uth/developer/sets


wescopeland (approved):
Verified issue is resolved locally.

PR #2030, opened 2023-11-17
fix(user): add validation to the 'g' query param

wescopeland:
Abort if non numeric or greater than 100.

https://discord.com/channels/476211979464343552/1026595325038833725/1174924736795512853

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -23,7 +23,7 @@ authenticateFromCookie($user, $permissions, $userDetails);
 
 $maxNumGamesToFetch = requestInputSanitized('g', 5, 'integer');
 
-if (!is_numeric($maxNumGamesToFetch) || $maxNumGamesToFetch > 100) {
+if ($maxNumGamesToFetch < 1 || $maxNumGamesToFetch > 100) {
     abort(400);
 }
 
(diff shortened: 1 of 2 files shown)


wescopeland:
Reduced to 100 in latest.

public/userInfo.php
```diff
@@ -23,6 +23,10 @@
 
 $maxNumGamesToFetch = requestInputSanitized('g', 5, 'integer');
 
+if (!is_numeric($maxNumGamesToFetch) || $maxNumGamesToFetch > 100) {
```
Jamiras, line 26: I don't think the `is_numeric` is necessary. The `integer` clause in `requestInputSanitized` will ensure the returned value is numeric.

If I do pass in a non-numeric value, I get "Last 0 Games played".

If I pass in a negative value, I get an exception.

Maybe there should be a minimum value as well? Not sure if 0, 1, or 5 would make more sense.

wescopeland: Makes sense.

In latest, I've set the minimum value to 1. Passing in 0 or a negative value does result in a 400. My personal opinion is we shouldn't try to "save" the user's response by coercing it into a more reasonable value, but this is not a strongly held opinion & I am happy to adjust further if you think we should go the other direction.

Jamiras (approved):

PR #2031, opened 2023-11-17
fix error viewing want-to-play list sorted by date

Jamiras:
fixes
```
SQLSTATE[HY093]: Invalid parameter number (SQL: SELECT  gd.ID, gd.Title, gd.ImageIcon, gd.TotalTruePoints,
...
#7 /home/forge/retroachievements.org/releases/2023-11-17T040722-5.2.1-054ded59/public/gameList.php(31): getGamesListByDev()
```

bindings left over from preliminary query were confounding primary query


wescopeland (approved):
Repro'd the issue locally and verified the fix.

PR #2034, opened 2023-11-17
feat(user): remove auth gating from completion progress

wescopeland:
Right now, the Completion Progress page requires authentication to view.

This PR removes the auth gate and adds a meta description to the page for crawlers.

### DIFF
--- a/resources/views/platform/completion-progress-page.blade.php
+++ b/resources/views/platform/completion-progress-page.blade.php
@@ -2,7 +2,7 @@
 use App\Site\Models\User;
 
 $targetUsername = $user->User;
-$isMe = $me->User === $targetUsername;
+$isMe = $me?->User === $targetUsername;
 
 $headingLabel = '';
 if ($isMe) {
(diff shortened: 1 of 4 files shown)


resources/views/platform/completion-progress-page.blade.php
```diff
@@ -13,12 +13,10 @@
     $headingLabel = $targetUsername . "'s Completion Progress";
 }
 
-// TODO: Once using denormalized data, come up with a good page description.
-// Doesn't matter right now because these pages don't generate any SEO juice.
-// $pageDescription = "";
+$pageDescription = "View {$targetUsername}'s game completion stats and milestones on RetroAchievements. Track their played, unfinished, and mastered games from various systems.";
```
Jamiras, line 16: Do we need to worry about the crawler following the filter links (Played, Unfinished, Beaten, Completed, Mastered)?

Jamiras (approved):

resources/views/platform/completion-progress-page.blade.php
wescopeland: Good call; I got nervous at first but it looks like Laravel handles this for us by setting a canonical URL for the page:

Screenshot 2023-11-17 at 6 07 41 PM

This is correctly set regardless of what filters are applied, so crawlers will not see filtered content as being duplicative.

PR #2036, opened 2023-11-18
chore(deps): use laravel 10

wescopeland:
This PR migrates RAWeb from Laravel 9 to Laravel 10. This was a [mostly seamless transition](https://laravel.com/docs/10.x/upgrade) with only one breaking change.

Summary of changes:
* Removed all explicit calls to `registerPolicies()`.
* Set `cache:prune-stale-tags` to run hourly.
* Added a few missing imports to some PHP files.
* Renamed `$routeMiddleware` to `$middlewareAliases`.
* Updated _validateEmail.php_ to account for an API change with the `Validator` class.

I also set the pre-push hook to use `composer paratest` instead of `composer test`.


luchaos:
i have an open merge for the upgrade which includes a few more things. i can state the PR in a few hours.
paratest has yet to be tested on windows.

wescopeland:
Sounds good, no worries at all. Will close this one.

PR #960, opened 2022-04-20
add tooltip to game rating

Jamiras:
Adds a tooltip to the game rating widget showing a distribution of the user-submitted ratings, and highlighting the user's rating.
image

When the user changes their rating, the user's rating portion of the tooltip is updated, but the distribution is not.
image

The rating will be hidden until there are at least five ratings submitted:
image
image
image

When a game has no ratings:
image

PR #2038, opened 2023-11-18
only fetch claimants for set requests on set requests page

Jamiras:
eliminates a ~40ms query from my profile page

### DIFF
--- a/public/userInfo.php
+++ b/public/userInfo.php
@@ -27,6 +27,11 @@ if ($maxNumGamesToFetch < 1 || $maxNumGamesToFetch > 100) {
     abort(400);
 }
 
+$userPageModel = User::firstWhere('User', $userPage);
+if (!$userPageModel) {
+    abort(404);
+}
+
 $userMassData = getUserPageInfo($userPage, numGames: $maxNumGamesToFetch);
 if (empty($userMassData)) {
     abort(404);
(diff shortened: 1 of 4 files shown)


public/userInfo.php
```diff
@@ -39,8 +44,7 @@
 $userPage = $userMassData['User'];
 $userMotto = $userMassData['Motto'];
 $userPageID = $userMassData['ID'];
-$setRequestList = getUserRequestList($userPage);
```
Jamiras, line 42: This was the expensive call that was eliminated. This page only needs the number of used requests, not information about the games they're associated to. In particular, determining if the games were claimed was expensive.

wescopeland (approved):
Verified locally.

Tangentially related: it seems kinda weird that on every page that doesn't have a set, we're making a client-side call to _set-requests.php_. We should probably consider refactoring this to live entirely on the server - the content being rendered from the response is causing a tiny layout shift because it isn't initially painted to the screen.

Jamiras:
> it seems kinda weird that on every page that doesn't have a set, we're making a client-side call to set-requests.php.

That logic has been there since it was introduced in #378. I assume it's a result of mimicking the set rating logic that's present when the game does have a set. That was eliminated in #960.

wescopeland:
Yep, not suggesting at all it was introduced by this PR or should even change in this PR. Just more of a note we should probably consider refactoring sometime soon-ish.

PR #2041, opened 2023-11-18
feat(user): prefetch achievement badges for recently played games

wescopeland:
This PR attaches an event to these items on the user profile:

Screenshot 2023-11-18 at 1 01 57 PM

When hovering over the rows, an event now fires to preload the badge images. This should hopefully reduce some "pop in" that can currently happen due to badges loading the moment a user expands a row.

### DIFF
--- a/app/Community/Components/UserRecentlyPlayed.php
+++ b/app/Community/Components/UserRecentlyPlayed.php
@@ -100,6 +100,16 @@ class UserRecentlyPlayed extends Component
         }
     }
 
+    private function deriveAchievementBadgeUrl(array $achievementData): string
+    {
+        $isAwarded = $achievementData['IsAwarded'];
+        $rawBadgeName = $achievementData['BadgeName'];
+
+        $processedBadgeName = $isAwarded ? "{$rawBadgeName}.png" : "{$rawBadgeName}_lock.png";
+
+        return media_asset("Badge/{$processedBadgeName}");
+    }
+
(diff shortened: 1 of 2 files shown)


Jamiras (commented):
This doesn't seem to provide much benefit if I just move onto the arrow and open the dropdown. It does help if I accidentally cross over one row while trying to get to other places and then come back and expand that row.

If the images are already cached locally, I see no difference with or without this code.

wescopeland:
The difference on my local is noticeable. With the cache disabled, there is a significant reduction of pop in.

**Before**

https://github.com/RetroAchievements/RAWeb/assets/3984985/09b1d242-1058-4b85-8ae8-616e77190901

**After**

https://github.com/RetroAchievements/RAWeb/assets/3984985/507963bb-9cd6-440b-be9b-a5250ffe4bc6

Jamiras:
I understand the intent of the feature. I'm just saying it doesn't seem to help for me.

Here's a video of the feature branch. The first user I've visited before, so all the achievement images are available. The second user has to download the achievement images, so the attempt to preload doesn't actually help.

https://github.com/RetroAchievements/RAWeb/assets/32680403/567f6f32-58d6-40fd-b7c9-2b08079409b3

master for comparison:

https://github.com/RetroAchievements/RAWeb/assets/32680403/44f36851-0054-4d4e-b16d-44f487fea65d

I did a `.hooks/post-merge` and `npm run build` after switching branches

PR #2042, opened 2023-11-19
fix: remediate subset achievement update error

wescopeland:
Resolves https://discord.com/channels/310192285306454017/310195377993416714/1175593754028158986.

**Root Cause**
When editing an achievement belonging to a subset, there isn't a type field anymore. Therefore, we need to add some safety before we navigate into the "type" array key of the `$input` variable.

### DIFF
--- a/public/request/achievement/update-base.php
+++ b/public/request/achievement/update-base.php
@@ -34,7 +34,7 @@ if (UploadNewAchievement(
     title: $input['title'],
     desc: $input['description'],
     points: $input['points'],
-    type: isset($input['type']) ? $input['type'] : null,
+    type: $input['type'] ?? null,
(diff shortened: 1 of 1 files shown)


public/request/achievement/update-base.php
```diff
@@ -34,7 +34,7 @@
     title: $input['title'],
     desc: $input['description'],
     points: $input['points'],
-    type: $input['type'],
+    type: isset($input['type']) ? $input['type'] : null,
```
Jamiras, line 37: Would be simpler to use `??`:
```
   type: $input['type'] ?? null,
```

wescopeland: Updated in latest

Jamiras (approved):

PR #2043, opened 2023-11-19
add indices to token authentication lookups

Jamiras:
Anecdotally should remove about 300ms from every Connect call, and about 100ms from every public API call.

This query (for example) appears a lot in the "show full processlist" for the prod server:
```
select * from `UserAccounts` where `appToken` = 'XXXX' and `UserAccounts`.`Deleted` is null limit 1;
```

Running it directly against my local server takes about 200ms and goes to 1ms with the index. The same query on the prod server takes about 300ms


database/migrations/platform/2023_11_19_000000_update_user_accounts_table.php
```diff
@@ -0,0 +1,47 @@
...
+    public function up(): void
+    {
+        Schema::table('UserAccounts', function (Blueprint $table) {
+            $sm = Schema::getConnection()->getDoctrineSchemaManager();
+            $indexesFound = $sm->listTableIndexes('UserAccounts');
+
+            if (!array_key_exists('users_apptoken_index', $indexesFound)) {
+                $table->index(
+                    ['appToken'],
+                    'users_apptoken_index'
```
wescopeland, line 19: nit: Do we need to even give these custom names? I was under the impression Laravel would automatically handle this for us, eg: https://github.com/RetroAchievements/RAWeb/blob/master/database/migrations/platform/2023_11_09_000000_update_site_awards_table.php

wescopeland (approved):

database/migrations/platform/2023_11_19_000000_update_user_accounts_table.php
luchaos: Indeed, providing custom names for the index is only necessary if the generated name exceeds 64 characters.
Can be shortened to:
```php
$table->index('appToken');
$table->index('APIKey');
```

luchaos (approved):

database/migrations/platform/2023_11_19_000000_update_user_accounts_table.php
Jamiras: Without the explicit alias, the name starts with `useraccounts_`, not `users_`:
```
...
  KEY `users_points_unranked_at_index` (`RAPoints`,`unranked_at`),
  KEY `users_points_softcore_unranked_at_index` (`RASoftcorePoints`,`unranked_at`),
  KEY `users_points_weighted_unranked_at_index` (`TrueRAPoints`,`unranked_at`),
  KEY `useraccounts_apptoken_index` (`appToken`),
  KEY `useraccounts_apikey_index` (`APIKey`)
```

luchaos: i completely forgot about that, nevermind! makes sense to define the name in this case 👍
i hope to get the table renaming in place soon to get rid of the inconsistencies - migration for that has been prepared.

PR #2045, opened 2023-11-19
feat: prettify query params like ?page[number]=1

wescopeland:
This PR results in human-readable query params being produced on some of our newer pages, via string replacement of "%5B" and "%5D" to their respective brackets: "[" and "]".

**How to test this PR:**
Visit either the developer sets page, the completion progress page, or the beaten games leaderboard. Apply any combination of filters and observe the produced URL in the browser navbar.

**Before**
```
/progress?filter%5Bsystem%5D=38&page%5Bnumber%5D=1&filter%5Bstatus%5D=unawarded
```

**After**
```
/progress?filter[system]=38&page[number]=1&filter[status]=unawarded
```

### DIFF
--- a/app/Support/Url/UrlBuilder.php
+++ b/app/Support/Url/UrlBuilder.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Support\Url;
+
+class UrlBuilder
+{
+    public static function prettyHttpBuildQuery(array $query): string
+    {
+        $queryString = http_build_query($query);
+        $queryString = str_replace(['%5B', '%5D'], ['[', ']'], $queryString);
+
+        return $queryString;
+    }
+}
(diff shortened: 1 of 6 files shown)


Jamiras:
Are you sure this is safe to do? `http_build_query` uses [RFC1738](https://www.ietf.org/rfc/rfc1738.txt) by default, which says:
>  Other characters are unsafe because gateways and other transport agents are known to sometimes modify such characters. These characters are "{", "}", "|", "\", "^", "~", "[", "]", and "`".

> All unsafe characters must always be encoded within a URL.

`http_build_query` also supports [RFC3986](https://www.ietf.org/rfc/rfc3986.txt), but that still encodes the brackets.
> A host identified by an Internet Protocol literal address, version 6 [RFC3513] or later, is distinguished by enclosing the IP literal within square brackets ("[" and "]").  This is the only place where square bracket characters are allowed in the URI syntax.

wescopeland:
Good point, let's sit on this for now.

PR #2046, opened 2023-11-19
convert set requests to blade

Jamiras:
eliminates initial async load of component as well


wescopeland (approved):
Verified functionality locally. The layout shift from this part of the page is now fixed.

PR #2051, opened 2023-11-21
feat(beaten-leaderboard): exclude homebrew systems from retail games

wescopeland:
This PR makes a slight adjustment to the logic in the `UpdatePlayerStats` action which is responsible for determining a game's kind (retail, homebrew, hack, etc.)

Three systems we support: Arduboy, Uzebox, and WASM-4 never had retail games in the conventional sense. These are homebrew systems. Therefore, it is reasonable that the fallback value for game kinds on these systems is not "retail", but is instead "homebrew".

* Arduboy, Uzebox, and WASM-4 games are now classed in player_stats as homebrew.
* When selecting one of these systems on the beaten leaderboard, the filter changes to "All" and the "Retail" radio button is disabled.

Screenshot 2023-11-20 at 7 27 19 PM

### DIFF
--- a/resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
+++ b/resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
@@ -5,6 +5,16 @@
     'selectedConsoleId' => null,
 ])
 
+<?php
+// Arduboy, WASM-4, and Uzebox are homebrew consoles.
+// They do not have games that would conventionally be considered "retail".
+$allowsRetail = (
+    $selectedConsoleId !== 71
+    && $selectedConsoleId !== 72
+    && $selectedConsoleId !== 80
+);
+?>
+
 <script>
 function handleGameKindsChanged(event) {
     window.updateUrlParameter(
@@ -14,9 +24,21 @@ function handleGameKindsChanged(event) {
 }
 
 function handleConsoleChanged(event) {
+    // The "Retail" filter is automatically disabled on homebrew systems.
+    // Default to the "All" option.
+    const homebrewSystemIds = [71, 72, 80];
+    if (homebrewSystemIds.includes(Number(event.target.value))) {
+        window.updateUrlParameter(
+            ['page[number]', 'filter[system]', 'filter[kind]'],
+            [1, event.target.value, 'all'],
+        );
+
+        return;
+    }
+
     window.updateUrlParameter(
-        ['page[number]', 'filter[system]'],
-        [1, event.target.value],
+        ['page[number]', 'filter[system]', 'filter[kind]'],
+        [1, event.target.value, 'retail'],
     );
 }
 </script>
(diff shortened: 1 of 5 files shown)


resources/views/platform/components/beaten-games-leaderboard/meta-panel/index.blade.php
```diff
@@ -5,6 +5,16 @@
...
 ])
 
+<?php
+// Arduboy, WASM-4, and Uzebox are homebrew consoles.
+// They do not have games that would conventionally be considered "retail".
+$allowsRetail = (
+    $selectedConsoleId !== 71
+    && $selectedConsoleId !== 72
+    && $selectedConsoleId !== 80
+);
```
Jamiras, line 15: suggest putting this logic in the `System` model (see also `isGameSystem`).

wescopeland: Good idea. Addressed in latest with two new static methods in `System`: `getHomebrewSystems()` and `isHomebrewSystem()`. This lets us rely on the `System` model for both the front-end and the back-end logic.

Jamiras (approved):
It's working as described, but I can't help but feel there's still something missing. Even though there aren't retail games for these systems, since the normal games aren't marked as ~homebrew~, there's no way to filter out the ~hack~s. Uzebox already has a couple ~hack~ sets, but there's nothing preventing hacks from being created for any of these consoles.

wescopeland:
> Even though there aren't retail games for these systems, since the normal games aren't marked as homebrew, there's no way to filter out the hacks.

Sorry, to make sure I totally understand, since "untagged" games are defaulting to a homebrew type, doesn't this mean hacks can be filtered out by selecting the "Homebrew" radio button?

Jamiras:
Uzebox all - SuperMeatBro has 17
image

Uzebox hacks - SuperMeatBro has 1
image

But there's no way to see SuperMeatBro with 16 non-hacks. Selecting Homebrew returns "no records found".

wescopeland:
I'm seeing something different locally:

`sail artisan ra:platform:player:update-stats SuperMeatBro`

Screenshot 2023-11-27 at 5 05 10 PM

Jamiras:
That was it. Manually updating his stats put him on the homebrew list. But running the command without a target user did not add anyone else to the list (yes, I waited for the queue to empty). Do we have a plan for updating users after this is deployed?

wescopeland:
> But running the command without a target user did not add anyone else to the list (yes, I waited for the queue to empty).

Here's something really weird I've learned which may be the cause of this. Running the command for a single user does not dispatch a job for it. Obviously, running it for all users dispatches many jobs. If Horizon is running before the code has changed for the action, it will not pick up the changes (???) and it executes jobs using the old action code. Every time I change code for an action, I have to restart Horizon to get accurate results.

> Do we have a plan for updating users after this is deployed?

I think the best plan is for this PR and https://github.com/RetroAchievements/RAWeb/pull/2050 to go out in the same release, and we kill two birds with one stone by doing a full sync on deploy. The leaderboard will have a few minutes of inaccurate results/downtime, then after a few minutes all the data integrity issues will be resolved.

Jamiras:
> If Horizon is running before the code has changed for the action, it will not pick up the changes (???) and it executes jobs using the old action code. Every time I change code for an action, I have to restart Horizon to get accurate results.

That was it.

PR #2056, opened 2023-11-24
fix(forum): resolve game title rendering issues in code blocks

wescopeland:
Resolves an issue where the Blade renderer's whitespace and `<pre>` tags are resulting in very spaced out game titles within forum code blocks:

**Before**
Screenshot 2023-11-24 at 9 29 57 AM

**After**
Screenshot 2023-11-24 at 9 29 47 AM

**How to test this PR**
Create a post with the following content:

```
[game=26470]

[code]
[game=26470]
[/code]
```


luchaos (approved):
I wished code blocks weren't abused like this 😞 
Somebody started using them for structuring posts when they actually shouldn't render anything within but only output un-rendered text - much like markdown.
Anyways, can't solve that easily now - you might want to add this case to https://retroachievements.org/test/shortcode.php for visual smoke testing?

AuburnRDM:
> I wished code blocks weren't abused like this 😞 Somebody started using them for structuring posts when they actually shouldn't render anything within but only output un-rendered text - much like markdown. Anyways, can't solve that easily now - you might want to add this case to https://retroachievements.org/test/shortcode.php for visual smoke testing?
> 
> See `formatting works in code?` - it really shouldn't imo 😅 that's not how code blocks usually work.
> 
> Maybe we can fix this in the future by replacing the `code` tag with something more appropriate.

Doing this would break a LOT of older event posts and require them to be reformatted unless [code][/code] would also still work or be automatically converted to the new tag.

luchaos:
> Doing this would break a LOT of older event posts and require them to be reformatted unless [code][/code] would also still work or be automatically converted to the new tag.

yes, replacing all usages of `code` tags with any new tag is implied. i'm aware of all those posts that rely on using the tags available for fanciness :)

televandalist:
I remember the concerns about misusing the code blocks from early on, but it's been going on for like two or three years. The tag doesn't behave how you'd expect it to, so we did what we could with it. Kinda like how we've using "game" entries for hubs for roughly four years now. It makes no sense, but we made the most of it because it was originally intended to be a temporary workaround until a tagging system was implemented. We still do it that way because it's still the only means to do so.

luchaos:
absolutely, i'm with you. i should've phrased it differently. all i wanted to say is that this is one of the areas where a better solution is due. i don't want to blame anyone making the best with the options that are given. that's perfectly normal.
i'm dedicated to get rid of all of those make-shift solutions and only wanted to remind myself/ourselves that the shortcode parser is one of those areas - too easy to go blind on such things

wescopeland:
Agree with all thoughts here. `code` tags are being used for fancy formatting right now than sharing actual code. I think swapping existing use to something like `mono` down the line and repurposing `code` to better support colorized code blocks is a good idea.

Jamiras (approved):

PR #2057, opened 2023-11-24
fix(game): adjust completion % rounding to match other pages

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1177243393735270532.

Currently, sometimes the progress component on game pages shows a different percentage than user pages. This is due to a difference in how the numbers are rounded.

**Before**
Screenshot 2023-11-24 at 10 21 18 AM

Screenshot 2023-11-24 at 10 21 24 AM

**After**
Screenshot 2023-11-24 at 10 22 12 AM

### DIFF
--- a/resources/views/platform/components/game/current-progress/progress-bar.blade.php
+++ b/resources/views/platform/components/game/current-progress/progress-bar.blade.php
@@ -16,11 +16,11 @@ if ($totalAchievementsCount > 0) {
     $hardcoreProgressWidth = ($numEarnedHardcoreAchievements / $totalAchievementsCount) * 100;
     $softcoreProgressWidth = ($numEarnedSoftcoreAchievements / $totalAchievementsCount) * 100;
 
-    $hardcoreCompletionPercentage = sprintf("%01.0f", floor($hardcoreProgressWidth));
-    $softcoreCompletionPercentage = sprintf("%01.0f", floor($softcoreProgressWidth + $hardcoreCompletionPercentage));
+    $hardcoreCompletionPercentage = sprintf("%01.0f", $hardcoreProgressWidth);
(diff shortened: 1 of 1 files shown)


resources/views/platform/components/game/current-progress/progress-bar.blade.php
```diff
@@ -16,11 +16,11 @@
...
     $softcoreProgressWidth = ($numEarnedSoftcoreAchievements / $totalAchievementsCount) * 100;
 
-    $hardcoreCompletionPercentage = sprintf("%01.0f", floor($hardcoreProgressWidth));
-    $softcoreCompletionPercentage = sprintf("%01.0f", floor($softcoreProgressWidth + $hardcoreCompletionPercentage));
+    $hardcoreCompletionPercentage = sprintf("%01.0f", $hardcoreProgressWidth);
+    $softcoreCompletionPercentage = sprintf("%01.0f", $softcoreProgressWidth + $hardcoreCompletionPercentage);
 }
 
-$completionPercentage = sprintf("%01.0f", floor($hardcoreProgressWidth + $softcoreProgressWidth));
+$completionPercentage = sprintf("%01.0f", round($hardcoreProgressWidth + $softcoreProgressWidth));
```
luchaos, line 23: I might be misunderstanding but isn't rounding to floor preferable to not ever display 100% when it's not?

wescopeland: Good call. Yes, I think you are right.

In latest, I've reverted the change to the game page progress component and updated the game list item progress bar component. Now the game list item progress should align to the game page instead of the other way around.

luchaos (approved):

PR #2061, opened 2023-11-25
chore: upgrade to Laravel v10 and Livewire v3

luchaos:
Update all vendor assets for packages.


resources/js/alpine/newsCarouselComponent/newsCarouselComponent.test.ts
```diff
@@ -10,6 +9,7 @@ import {
   it,
   vi,
 } from 'vitest';
+import Alpine from '../../../../vendor/livewire/livewire/dist/livewire';
```
luchaos, line 12: @wescopeland not sure how to deal with the failed test. Alpine doesn't seem to be happy how we're using it but i couldn't figure out how to overcome this.
Maybe you can help finding what the issue is? 🙏

resources/views/layouts/components/body-end.blade.php
```diff
@@ -1,5 +1,4 @@
-{{-- TODO add livewire--}}
-{{--<livewire:scripts />--}}
+<livewire:scripts />
```
luchaos, line 1: livewire assets are now included as designed - includes Alpine.js

resources/js/alpine/newsCarouselComponent/newsCarouselComponent.test.ts
wescopeland: I think the correct approach to solving this is to set up an alias in the vite config, eg:

```ts
resolve: {
  alias: {
    '@': resolve(__dirname, './resources/js'),
    alpine: resolve(__dirname, './vendor/livewire/livewire/dist/livewire.js')
  },
},
```

Unfortunately I'm not having much luck with this, even though it does appear the file is trying to export Alpine:

```js
// cjs (livewire.js)
window.Alpine = module_default;

// esm (livewire.esm.js)
var export_Alpine = import_alpinejs20.default;
export {
  export_Alpine as Alpine,
  Livewire2 as Livewire
};
```

I can get as far as the tests trying to run, but then some kind of DOM error is output from an Alpine.js internal and I can't get much further.

I think we have two options:
* Install alpine as a dev dependency for the sole purpose of testing.
* Delete this test suite.

I have no strong feelings or opinion about either approach.

luchaos: assuming we'd aim for replacing all of the custom alpine components and only use livewire directly in the future we might be good to remove the tests?

luchaos: maybe too soon - went with the dev dependency instead

luchaos: sounds good :) thank you for looking into this!

wescopeland: Sure thing!

resources/js/app.ts
```diff
@@ -1,8 +1,8 @@
-import Alpine from 'alpinejs';
-import focus from '@alpinejs/focus';
 // eslint-disable-next-line camelcase,import/no-unresolved
 // import { livewire_hot_reload } from 'virtual:livewire-hot-reload';
 
+// eslint-disable-next-line import/no-unresolved
+import { Livewire } from 'livewire/livewire.esm';
```
wescopeland, line 5: The TypeScript language server used by VSCode is throwing a type error here. I'm not sure why it's not showing up in the build itself, but I was able to resolve it locally with the following steps:

1. Adjust the alias to point at the ESM bundle:
```ts
        livewire: resolve(__dirname, './vendor/livewire/livewire/dist/livewire.esm'),
```

2. In _resources/js_ (the same directory as _app.ts_), create a new file called _livewire.d.ts_ with the following content:

```ts
declare module 'livewire' {
  export const Livewire: any;
}
```

3. In _app.ts_, adjust the import:
```ts
import { Livewire } from 'livewire';
```

luchaos: awesome, thanks! done in https://github.com/RetroAchievements/RAWeb/pull/2061/commits/e6af867b2d71338a286b4946329759bae914afc5

database/migrations/2014_10_12_000001_create_password_reset_tokens_table.php
```diff
@@ -9,11 +9,11 @@
 return new class() extends Migration {
     public function up(): void
     {
-        if (Schema::hasTable('password_resets')) {
```
Jamiras, line 12: Do we need a migration step for this? Or some sort of documentation for queries we need to run on our dumps?

database/migrations/2018_09_01_000001_create_permission_tables.php
```diff
@@ -9,79 +9,107 @@
...
+
+            $table->unique(['name', 'guard_name']);
         });
 
-        Schema::create($tableNames['roles'], function (Blueprint $table) {
-            $table->bigIncrements('id');
-            $table->string('name');
+        Schema::create($tableNames['roles'], function (Blueprint $table) use ($teams, $columnNames) {
+            $table->bigIncrements('id'); // role id
+            if ($teams || config('permission.testing')) { // permission.testing is a fix for sqlite testing
```
Jamiras, line 36: is `config('permission.testing')` preferred over `app()->environment('local', 'testing')`?

database/migrations/2018_09_01_000001_create_permission_tables.php
```diff
@@ -9,79 +9,107 @@
 return new class() extends Migration {
     public function up(): void
     {
+        $teams = config('permission.teams');
```
Jamiras, line 12: I'm a bid concerned about conditional migration code. Especially in files from migrations that should have occurred in 2018.  How does this logic impact the prod DB?

config/fortify.php
```diff
@@ -50,6 +50,19 @@
...
+    | Lowercase Usernames
+    |--------------------------------------------------------------------------
+    |
+    | This value defines whether usernames should be lowercased before saving
+    | them in the database, as some database system string fields are case
+    | sensitive. You may disable this for your application if necessary.
+    |
+    */
+
+    'lowercase_usernames' => true,
```
Jamiras, line 64: This seems like a significant change. What about all the existing users?

database/migrations/2018_09_01_000001_create_permission_tables.php
luchaos: that's from a vendor package update - only updated to keep things clean.
otherwise i absolutely agree, our application's own migrations should not be treated like this if possible.

database/migrations/2014_10_12_000001_create_password_reset_tokens_table.php
luchaos: `php artisan migrate` should be the only required command to run in this case

config/fortify.php
luchaos: good point, this will be changed. fortify is not yet used for user registration, will follow in another PR where public scripts will be replaced.

i'm still thinking about lowercasing all the usernames and moving the current usernames to the display name column. this makes the integration of fortify much easier and follows the v2 prototype's approach.
queries will have to be checked for this change, too, but as we're using a case insensitive collation we might be good without any additional changes.

database/migrations/2018_09_01_000001_create_permission_tables.php
luchaos: oh, i should check if this is actually coming from the package now that you mention it

database/migrations/2018_09_01_000001_create_permission_tables.php
luchaos: maybe we should actually incorporate the teams feature 🤔 i'll look into this.

database/migrations/2018_09_01_000001_create_permission_tables.php
luchaos: i ran `php artidan vendor:publish --force` to get updated assets for all packages - sometimes there are backported features in packages' migrations.

config/fortify.php
luchaos: nothing that has to be decided just yet imo. that's just the default

database/migrations/2018_09_01_000001_create_permission_tables.php
luchaos: Reset this migration to its previous state and added a new migration for the teams feature as per the package's documentation: https://spatie.be/docs/laravel-permission/v6/basic-usage/teams-permissions#content-enabling-teams-permissions-feature

PR #2062, opened 2023-11-27
Threaded messages

Jamiras:
Replaces the inbox/outbox with threaded messages, wherein a conversation between two users about a single subject can be discussed (similar to forum posts). Most importantly, reply chains are grouped together and it's easy to see both sides of the conversation.

A new message is created the same as before, and supports full shortcode syntax and preview:
image

This immediately appears in the user's list (there's no longer a distinction between inbox and outbox):
image

It also appears in the recipient's list (as unread):
image

The recipient can open the message and reply to it. (the sender can also open it and send additional information).
image

The sender is notified of the new message as part of an updated thread:
image

Both users can view the entire thread and update it at any time:
image

Either user can delete the thread (which removes it from their list, but not the other participant's list).
image

If the other participant replies again, it reappears in the first user's list.
image

If both users delete the thread, it's removed from the database.


resources/views/community/components/message/list-page.blade.php
```diff
@@ -0,0 +1,123 @@
...
+>
+    <x-user.breadcrumbs targetUsername="{{ $user->User }}" currentPage="Messages" />
+
+    <div class="mt-3 w-full flex gap-x-3">
+        {!! userAvatar($user->User, label: false, iconSize: 48, iconClass: 'rounded-sm') !!}
+        <h1 class="mt-[10px] w-full">{{ $user->User }}'s Messages</h1>
+    </div>
+
+    <div class='ml-2'>
+        You have {{ $unreadCount }} unread messages in {{ $totalMessages }} message threads.
```
wescopeland, line 44: Recommend adding pluralization here.

Screenshot 2023-11-27 at 4 15 58 PM

resources/views/community/components/message/list-page.blade.php
```diff
@@ -0,0 +1,123 @@
...
+            $mostRecentUpdate = Carbon::parse($message->last_message_at);
+            $humanDate = $mostRecentUpdate->format('F j Y, g:ia');
+
+            $num_unread = $message->num_unread;
+        ?>
+        <tr>
+            <td>
+                @if ($num_unread > 0)
+                    <b>
+                @endif
```
wescopeland, line 75: I think we can add these styles with less code by conditionally attaching a class to the parent `<td>`, eg:

```php
<td @if ($num_unread > 0) class="font-bold" @endif>
```

resources/views/community/components/message/list-page.blade.php
```diff
@@ -0,0 +1,123 @@
...
+                @else
+                    {{ $message->num_messages }} <b>({{ $num_unread }} unread)</b>
+                @endif
+            </td>
+
+            <td class='text-right'>
+                @if ($isShowAbsoluteDatesPreferenceSet)
+                    <span class='smalldate'>{{ $humanDate }}</span>
+                @elseif ($mostRecentUpdate < $monthAgo)
+                    <span class='smalldate'>{{ $humanDate }}</span>
```
wescopeland, line 110: nit: Can the `@elseif` be removed and this be condensed down to a single block?

resources/views/community/components/message/new-thread-page.blade.php
```diff
@@ -0,0 +1,56 @@
...
+                <td>
+                    <div class="w-full">
+                        <div style="float:right">
+                            <x-input.user-select-image for="recipient" :user="$toUser" size="48" />
+                        </div>
+                        <!-- TODO: why won't this field align properly? -->
+                        <div style='vertical-align:middle'>
+                            <x-input.user-select name="recipient" :user="$toUser" />
+                        </div>
+                    </div>
```
wescopeland, line 34: Floats can make it tricky. I recommend always reaching for flex before trying to float an element. This should resolve the issue outright or get you very close:

```html
<div class="w-full flex justify-between items-center">
  <div><x-input.user-select ... /></div>
  <div><x-input.user-select-image ... /></div>
</div>
```

Note that I reversed the order of the inputs so the flex layout can set them correctly.

In a flex layout, `justify-*` controls element alignment along the X-axis and `items-*` controls element alignment along the Y-axis.

`justify-between` forces the elements along the X-axis to be "pushed" as far away as possible from each other.
`items-center` forces the elements along the Y-axis to be horizontally centered.

This is a quick and somewhat silly resource for how to master flex: https://flexboxfroggy.com/

resources/views/community/components/message/view-thread-page.blade.php
```diff
@@ -0,0 +1,111 @@
...
+        @foreach ($messages as $message)
+            <div class="embedded mb-2">
+                <div class="flex justify-between items-center">
+                    <div>
+                    {!! userAvatar($participants[$message->author_id], iconSize: 24) !!}
+                    <?php $humanDate = $message->created_at->format('F j Y, g:ia'); ?>
+                    @if ($isShowAbsoluteDatesPreferenceSet)
+                        <span class='smalldate'>{{ $humanDate }}</span>
+                    @elseif ($message->created_at < $monthAgo)
+                        <span class='smalldate'>{{ $humanDate }}</span>
```
wescopeland, line 80: nit: Can this be condensed to a single conditional?

resources/views/components/input/shortcode-textarea.blade.php
```diff
@@ -0,0 +1,53 @@
+@props([
+    'name' => 'body',
+    'enabled' => true,
+    'watermark' => '',
```
wescopeland, line 4: nit: I recommend renaming this to `placeholder`.

`enabled` is fair so as to not create a double negative on the standard `disabled` attribute, but I perceive less value in renaming `placeholder`.

resources/views/components/input/shortcode-textarea.blade.php
```diff
@@ -0,0 +1,53 @@
...
+    'maxLength' => 60000,
+    'initialValue' => '',
```
wescopeland, line 6: nit: It may be more ergonomic to delete this prop and use `{{ $slot }}` instead for the initial value, just like the native `<textarea>`. I believe this would lower the learning curve of this component.

```php
<textarea ...>{{ $slot }}</textarea>
```

```php
<x-input.shortcode-textarea ...>Here's my initial content!</x-input.shortcode-textarea>
```

resources/views/components/input/shortcode-textarea.blade.php
```diff
@@ -0,0 +1,53 @@
...
+<?php
+
+$inputEnabled = $enabled ? '' : 'disabled';
+$buttonEnabled = $enabled ? ":disabled='!isValid'" : "disabled";
+
+$loadingIconSrc = asset('assets/images/icon/loading.gif');
+
+?>
+
+<div class='mt-4'>
```
wescopeland, line 18:
```suggestion
<div>
```

The component likely shouldn't have any self-applied layout styles like this on its container, otherwise reusability in different contexts becomes more difficult. Instead, this responsibility should belong to the parent:

```php
<div class="mt-4">
  <x-input.shortcode-textarea />
</div>
```

resources/views/community/components/message/list-page.blade.php
```diff
@@ -0,0 +1,123 @@
...
+
+<x-app-layout
+    pageTitle="Messages"
+    pageDescription="Messages involving {{ $user->User }}"
+>
+    <x-user.breadcrumbs targetUsername="{{ $user->User }}" currentPage="Messages" />
+
+    <div class="mt-3 w-full flex gap-x-3">
+        {!! userAvatar($user->User, label: false, iconSize: 48, iconClass: 'rounded-sm') !!}
+        <h1 class="mt-[10px] w-full">{{ $user->User }}'s Messages</h1>
```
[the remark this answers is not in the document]
Jamiras: Yes. It should only be visible to the current user. I felt this was consistent with the other pages under the User breadcrumb, although Messages could be its own top-level breadcrumb item instead of being nested under the User.

resources/views/community/components/message/list-page.blade.php
Jamiras: done

resources/views/community/components/message/list-page.blade.php
Jamiras: done

resources/views/community/components/message/list-page.blade.php
Jamiras: done

resources/views/community/components/message/new-thread-page.blade.php
Jamiras: Thanks.

resources/views/community/components/message/view-thread-page.blade.php
```diff
@@ -0,0 +1,111 @@
...
+    <div class="mt-3 w-full flex gap-x-3">
+        <h1 class="mt-[10px] w-full">{!! $thread->title !!}</h1>
+    </div>
+
+    <div class="w-full flex my-2">
+        <div class="mr-6">
+            <form action='/request/message/delete.php' method='post'>
+            {{ csrf_field() }}
+            <input type="hidden" name="thread_id" value="{{ $thread->id }}" />
+            <button class="btn btn-danger" onclick="return confirm('Are you sure you want to delete this message?')">Delete</button>
```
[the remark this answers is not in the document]
Jamiras: Changed to "Are you sure you want to delete this message thread?"

resources/views/community/components/message/view-thread-page.blade.php
Jamiras: Yep. I considered making it into a separate component too as it's logically duplicated from the list view, but logically the month ago and user preference lookup should be contained in the component and would be evaluated multiple times.

resources/views/components/input/shortcode-textarea.blade.php
Jamiras: The component was extracted from https://github.com/RetroAchievements/RAWeb/blob/d74d235417dcfd4ca26e9fbba744ce5746a3c6b4/public/viewtopic.php#L213-L254 and I just chose names for each customizable bit of data.

I've changed `watermark` to `placeholder` as suggested.

resources/views/components/input/shortcode-textarea.blade.php
Jamiras: done

resources/views/components/input/shortcode-textarea.blade.php
Jamiras: done

luchaos (commented):
I realize that my review got extremely long and I'm not sure I even addressed all things yet. If you want I can take care of those architectural changes. I wasn't expecting changes like this to be made until I refactored another feature to have some reference.

It's wonderful to finally see some scripts being removed from `/public`.

app/Community/Commands/SyncMessages.php
```diff
@@ -4,28 +4,133 @@
...
+use App\Community\Models\MessageThreadParticipant;
+use App\Site\Models\User;
 use Illuminate\Console\Command;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
 
 class SyncMessages extends Command
 {
-    use SyncTrait;
```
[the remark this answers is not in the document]
Jamiras: This differs from the other sync commands in that it should only need to be run once. Nothing will be writing to the old structure once the new code is deployed.

app/Site/Actions/ClearAccountDataAction.php
```diff
@@ -39,6 +39,8 @@ public function execute(User $user): void
         // TODO $user->subscriptions()->delete();
         DB::statement('DELETE FROM Subscription WHERE UserID = :userId', ['userId' => $user->ID]);
 
+        DB::statement('UPDATE message_thread_participants SET deleted_at=NOW() WHERE user_id = :userId', ['userId' => $user->ID]);
```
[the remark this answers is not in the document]
Jamiras: We're not actually deleting the thread at this point - just the user's participation in the thread. The other participant(s) should still be able to see the thread if they haven't also deleted it.

config/missing-page-redirector.php
```diff
@@ -90,7 +90,6 @@
          */
         '/controlpanel.php' => '/settings',
         '/friends.php' => '/friends',
-        '/createmessage.php' => '/message/create',
```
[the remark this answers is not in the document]
Jamiras: It does? The parameters have changed, so the page will probably have to still exist and just call whatever Action gets created.

resources/views/community/components/message/list-page.blade.php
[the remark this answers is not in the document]
Jamiras: Do you mean `community/message/index.blade.php`?

database/migrations/platform/2023_11_25_000002_update_messages_table.php
```diff
@@ -0,0 +1,116 @@
...
+            $table->renameColumn('Payload', 'body');
+            $table->renameColumn('TimeSent', 'created_at');
+            $table->renameColumn('ID', 'id');
+        });
+
+        // populate author_id so we can add a foreign key
+        DB::statement("UPDATE Messages m SET m.author_id = (SELECT u.ID FROM UserAccounts u WHERE u.User = m.UserFrom)");
+
+        // delete records associated to non-existant users
+        DB::statement("DELETE FROM Messages WHERE author_id=0");
```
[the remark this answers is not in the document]
Jamiras: The `author_id` column must be populated/sanitized before the foreign key can be added. The `thread_id` foreign key reference is already delayed to the sync command, so I could move this too.

app/Site/Actions/ClearAccountDataAction.php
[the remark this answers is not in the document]
luchaos: ah, i see. it'd become something like `$user->messageThreadParticipstiond()->delete()` requiring a corresponding relation in ActsAsCommunityMember

resources/views/community/components/message/list-page.blade.php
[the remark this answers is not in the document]
luchaos: ah, yes, all with the blade suffix 🙏 i tend to imply that and forget to mention. everything in `resources/views/` should to be `...blade.php`

app/Site/Actions/ClearAccountDataAction.php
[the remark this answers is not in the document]
Jamiras: done

app/Community/RouteServiceProvider.php
```diff
@@ -228,11 +229,10 @@ protected function mapWebRoutes(): void
...
 
-                //     /*
-                //      * inbox
-                //      */
-                //     Route::resource('message', MessageController::class)->except('index', 'edit', 'update');
-                //     Route::resource('messages', MessageController::class)->only('index')->names(['index' => 'message.index']);
+                /*
+                 * messages
+                 */
+                Route::resource('messages', MessageThreadsController::class)->except(['edit', 'update']);
```
Jamiras, line 235: I see the previous duality of plural for list/non-plural for items. Do you want me to switch back to that?

resources/views/community/components/message/list-page.blade.php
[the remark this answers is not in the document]
Jamiras: Not sure if it sways any opinions, but here's an illustration showing the mapping from the breadcrumb. "You" does appear in the subheading to clarify context.
image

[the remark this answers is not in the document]
wescopeland: My opinion is unswayed 😅 But this isn't a dealbreaker by any means 🙂

When I see "WCopeland's Messages", my inner voice is saying "of course it is, I'm the only one who can see the page!" Which leads me to the question... am I _really_ the only one who can see the page? Why else would the page remind me in the `<h1>` of who I am?

It is for this reason that on the completion progress page, if you're viewing your own page, the label changes:

Screenshot 2023-11-29 at 4 58 19 PM

Obviously, I'm heavily speaking from my biases in this conversation, but I don't perceive this as confusing. I think it's a better experience and appreciate the attention to detail when the app recognizes my context. This doesn't necessarily mean we have to have "Your X" or "My X" labels everywhere to @luchaos's point. But in this case, this is exclusively context-driven content which I believe is what makes it appropriate.

Having the `<h1>` just read "Messages" feels like a good compromise.

app/Community/Commands/SyncMessages.php
```diff
@@ -4,28 +4,133 @@
...
     public function handle(): void
     {
-        $this->sync('messages');
+        $count = Message::where('thread_id', 0)->count();
+
+        $progressBar = $this->output->createProgressBar($count);
+        $progressBar->start();
+
+        // have to do this in batches to prevent exhausting memory
+        // due to requesting payloads (message content)
```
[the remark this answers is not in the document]
luchaos: I see. In such cases `chunkById()` can help:

> If you are updating database records while chunking results, your chunk results could change in unexpected ways. If you plan to update the retrieved records while chunking, it is always best to use the `chunkById` method instead. This method will automatically paginate the results based on the record's primary key

from https://laravel.com/docs/10.x/queries#chunking-results

app/Community/Controllers/MessageThreadsController.php
```diff
@@ -0,0 +1,200 @@
...
+use App\Http\Controller;
+use App\Site\Models\User;
+use Illuminate\Contracts\View\View;
+use Illuminate\Http\RedirectResponse;
+use Illuminate\Http\Request;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Validator;
+
+class MessageThreadsController extends Controller
```
luchaos, line 23: `MessageThreadController`

app/Community/Controllers/MessageThreadsController.php
```diff
@@ -0,0 +1,200 @@
...
+        ]);
+    }
+
+    public function show(Request $request, int $threadId): View
+    {
+        $thread = MessageThread::firstWhere('id', $threadId);
+        if (!$thread) {
+            abort(404);
+        }
+
```
luchaos, line 79: To make use of route model binding (see suggested adjustments to RouteServiceProvider)
`public function show(Request $request, Message $messageThread): View`

This can be removed, route model binding takes care of fetching and 404 failovers:
```php
$thread = MessageThread::firstWhere('id', $threadId);
if (!$thread) {
    abort(404);
}
```

resources/views/community/components/message/list-page.blade.php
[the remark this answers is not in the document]
luchaos: > But in this case, this is exclusively context-driven content which I believe is what makes it appropriate.

That's fair. In those two examples in the screenshots i would've opted to be much more stern for brevity and less noise. I can see its merit though, too.
I don't have a strong opinion on that at this point.

config/missing-page-redirector.php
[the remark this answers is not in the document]
luchaos: E.g. those are linked in the .github issue template config (you might want to update those too while at it?) https://github.com/RetroAchievements/RAWeb/issues/new/choose

I'd assume that the script is referenced in forums, docs, news, too, though. It's not as easy to get rid of such pages. Having a missing page redirect even if the parameters don't match is what the missing page redirector has been set up to do for removed public scripts.

What this config entry needs is some adjustment to forward the query parameters:

`'/createmessage.php' => '/message-thread/create?to={t}&subject={s}',`

We'll run into similar situations as soon as all the resource listings are refactored. Not all parameters have to be forwarded or handled as fallback but I'd do it where it's easy enough to do.

app/Community/AuthServiceProvider.php
```diff
@@ -45,7 +45,7 @@ class AuthServiceProvider extends ServiceProvider
         ForumTopicComment::class => ForumTopicCommentPolicy::class,
         ForumTopic::class => ForumTopicPolicy::class,
         GameComment::class => GameCommentPolicy::class,
-        Message::class => MessagePolicy::class,
```
luchaos, line 48: I'd keep the MessagePolicy intact - we might still need it in the future.
E.g. for determining if a single message may be seen, edited, deleted.

app/Community/Policies/MessageThreadPolicy.php
```diff
@@ -0,0 +1,44 @@
...
+
+use App\Community\Models\MessageThread;
+use App\Site\Models\User;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class MessageThreadPolicy
+{
+    use HandlesAuthorization;
+
+    public function view(User $user, MessageThread $message): bool
```
luchaos, line 15: `MessageThread $messageThread`

app/Community/Policies/MessageThreadPolicy.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return true;
```
luchaos, line 17: This should check if a given user is allowed to view the thread (same for delete).
Otherwise it's possible to view threads that one is not a participant of?

resources/views/community/components/message/list-page.blade.php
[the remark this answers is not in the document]
wescopeland: The i18n angle is a fair point and not something I had previously considered.

app/Community/RouteServiceProvider.php
luchaos: just realized that those routes aren't guarded yet; should read like this:
```php
Route::group([
    'middleware' => ['auth'], // TODO add 'verified' middleware
], function () {
    Route::resource('messages', MessageThreadController::class)->only('index')->names(['index' => 'message.index']);
    Route::get('message/create', [MessageThreadController::class, 'create'])->name('message.create');
    Route::resource('message-thread', MessageThreadController::class)->except('index', 'create', 'edit', 'update');
});
```

app/Community/Commands/SyncMessages.php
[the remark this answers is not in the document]
Jamiras: done.

config/missing-page-redirector.php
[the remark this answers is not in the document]
Jamiras: done

app/Community/RouteServiceProvider.php
Jamiras: > I'd still go with separating controllers for messages and message threads.

I'm still struggling with this decision. The user doesn't care about the separation of the models. Why wouldn't everything be `message-thread`?

```
messages.index -> message-thread.index
messages.show -> message-thread.show
messages.create -> message-thread.create
messages.store -> message-thread.store
messages.destroy -> message-thread.destroy
```

And
```
Route::resource('message-threads', MessageThreadController::class)->only('index')->names(['index' => 'message-thread.index']);
Route::resource('message-thread', MessageThreadController::class)->except('index', 'edit', 'update');
```

Personally, I like the shorter URLS: `/messages` and `/message/id`, but I realize they're effectively for the threads.

app/Community/Policies/MessageThreadPolicy.php
Jamiras: The participant restriction is currently enforced in `MessageThreadController::show`. I can't actually get these methods to be called - I'm guessing it's related to the route stuff.

config/missing-page-redirector.php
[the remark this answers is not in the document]
luchaos: should be singular: `/message/create?to={t}&subject={s}&message={p}`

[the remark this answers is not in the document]
Jamiras: Yep. That'll get corrected once I sort out the routes: https://github.com/RetroAchievements/RAWeb/pull/2062#discussion_r1412457717

app/Community/RouteServiceProvider.php
Jamiras: As part of separating the parts of the controller, I settled on being able to create/store `messages`, and show/destory `message_threads`. `index` is associated to `message_threads`, but pathed through `messages`:
```
/messages                        message-thread.index

/message-thread/{message_thread} message-thread.show
/message-thread/{message_thread} message-thread.destroy

/message/create                  message.create
/message                         message.store
```
I'm still uncertain on the `message.store` route, as it creates a message and appends it to a thread. It's not really updating a `message` or a `message_thread`.

app/Community/Controllers/MessageThreadsController.php
Jamiras: That doesn't seem to work. Did I name my resource wrong?
```
Route::resource('message-thread', MessageThreadController::class)->only(['show', 'destroy']);
```

```
public function show(Request $request, MessageThread $thread): View
{
    print_r($thread->toArray());
    exit;
```
prints an empty array.

app/Community/Policies/MessageThreadPolicy.php
Jamiras: I'm clearly still missing something.
```
    public function create(Request $request): View
    {
        $this->authorize('create', Message::class);
```
And I put a divide by zero exception in MessagePolicy::create, and it doesn't fail.

app/Community/RouteServiceProvider.php
luchaos: This looks really good now imo! Thanks for adjusting it like this 🙏 

`message.store` is fine like this I think as it's affecting both resources and it would have to work like this either way. Having a store route that matches the create view looks good to me.

app/Community/Controllers/MessageThreadsController.php
luchaos: I think it's the parameter name that has to match so the autowiring of route model binding can do its job via reflection. I think it actually looks at parameter names.

Does this work? `MessageThread $messageThread`

app/Community/Policies/MessageThreadPolicy.php
Jamiras: Yep. That seems to have been my problem.

If admins circumvent the policies, we can't really put the participation logic into the Policy file, or admins could view any message thread. Do you still want me to put authorize calls in every method, but leave the participation check where it is?

app/Community/RouteServiceProvider.php
Jamiras: I'm still not sure it's appropriate in either place. If we leave it here, we'll have to move it later if we give users the ability to edit previous messages (not something that's usually allowed anyway). If we move it to `message_thread.store`, then we lose a similar ability to edit the message thread title.

luchaos (approved):
This is a good template to go by for the other routes that are yet to be refactored 👏

app/Community/RouteServiceProvider.php
luchaos: for editing `edit` and `update` routes are used. `create` and `store` are only for creation.

resources/views/community/components/message/list-page.blade.php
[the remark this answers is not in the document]
Jamiras: Changed to "Messages" and removed the icon/breadcrumb.

PR #2067, opened 2023-12-02
feat: allow redirects to login page

luchaos:
Redirects to login page will work as intended now.
After being redirected to an auth guarded page as a guest it will prompt to login and redirect back to the intended page after successfully logging in.
That's default framework behaviour.

### DIFF
--- a/app/Helpers/database/user-activity.php
+++ b/app/Helpers/database/user-activity.php
@@ -128,6 +128,12 @@ function expireRecentlyPlayedGames(string $user): void
 
 function getRecentlyPlayedGames(string $user, int $offset, int $count, ?array &$dataOut): int
 {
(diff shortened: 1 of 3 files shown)


Jamiras (approved):
Will this fix the `Argument #1 ($content) must be of type ?string, Illuminate\Routing\Redirector given` error for `"url":"https://retroachievements.org/login"`? It would be so nice to have a clean error log.

luchaos:
> Will this fix the `Argument #1 ($content) must be of type ?string, Illuminate\Routing\Redirector given` error for `"url":"https://retroachievements.org/login"`? It would be so nice to have a clean error log.

I'm not sure where this error is coming from, couldn't figure it out yet. I'll look into that.

PR #2072, opened 2023-12-03
Rewrite download.php to MVC

anegoda1995:


app/Site/RouteServiceProvider.php
```diff
@@ -91,7 +92,7 @@ protected function mapWebRoutes(): void
             Route::get('demo', [ContentController::class, 'demo'])->name('demo');
             Route::get('demo/error/{code}', [ContentController::class, 'errorDemo'])->name('demo.error');
             Route::get('terms', [ContentController::class, 'terms'])->name('terms');
-            // Route::get('downloads', [DownloadController::class, 'index'])->name('download.index');
+             Route::get('downloads', [DownloadController::class, 'index'])->name('download.index');
```
wescopeland, line 95:
```suggestion
             Route::get('download', [DownloadController::class, 'index'])->name('download.index');
```

Everything else is called "download" (singular), including the link itself. Not sure why the path would be any different.

app/Site/Controllers/DownloadController.php
```diff
@@ -5,47 +5,21 @@
...
 
 use App\Http\Controller;
-use App\Platform\Models\Emulator;
-use App\Platform\Models\IntegrationRelease;
 use Illuminate\Contracts\View\View;
-use Illuminate\Database\Eloquent\Collection;
 
 class DownloadController extends Controller
 {
     public function index(): View
```
wescopeland, line 12: This seems identical to `ContentController#download()`.

resources/views/download.blade.php
```diff
@@ -0,0 +1,71 @@
+<x-app-layout :page-title="__('Download a client')">
```
wescopeland, line 1: This page needs a description for SEO purposes.

resources/views/download.blade.php
wescopeland, line 1: I think this file needs to be further decomposed into more components. The biggest tell is the constant:

`mb_strtolower($emulator['keyName'] ?? '')`

I see at least three more components that should be broken out:

* The "longheader"
* The emulator description and supported systems
* The emulator download meta

wescopeland (changes requested):

luchaos:
All of the code in DownloadController assumes Emulators and Integration Release models to be in place.
I'd like to switch to MVS for the downloads page with those in place the earliest 🙏

app/Site/RouteServiceProvider.php
luchaos: index pages are laid out to be plural

wescopeland: It seems then that everything else should be changed to reflect this, eg:
_downloads.blade.php_ instead of _download.blade.php_.

luchaos:
Thanks for your PR! I like where this is going, yet, unfortunately this does not fit well into the currently planned schedule for refactoring.
I contacted you on Discord so we can discuss further steps 🙏

PR #2073, opened 2023-12-04
fix: remove teams permissions

luchaos:
not worth it - cause problems with seeding and make permissions unnecessarily complicated for our use.


Jamiras (approved):

Jamiras:
I noticed this doesn't have a down(), and after I went back to master, I got this error trying to load a page:
```
Column not found: 1054 Unknown column 'auth_roles.team_id' in 'where clause'
```
I was able to get things back in a stable state by doing another `artisan migrate:rollback --step=1`, and then re-migrating the addition of the teams stuff, but I didn't try loading a page while this migration was in place, so I'm not 100% confident the same error doesn't occur there.

Jamiras:
After manually dropping an index:
```
alter table auth_roles drop key auth_roles_name_guard_name_unique;
```
I was able to reapply this migration, and verified I do not see the `Unknown column 'auth_roles.team_id'` error.

PR #2077, opened 2023-12-04
feat(game): add missable achievement filter

wescopeland:
This PR adds a top-level missables count to game pages. When clicked, it filters the achievements list to show only missables.

**Why?**
While the new missable indicator is desirable for several reasons, a common piece of feedback received so far has been that it may now be more difficult to notice which achievements are actually missable.

The counter makes this abundantly clear in a nonobtrusive way, while also giving the user new functionality to specifically filter the list without needing to manually search for "[m]" or use a third-party extension.

Thanks @Jamiras for the idea on this new button placement.

https://github.com/RetroAchievements/RAWeb/assets/3984985/5b5d1490-b6ad-4809-a6ae-cd0fec92429f


resources/views/platform/components/game/achievements-list-meta.blade.php
```diff
@@ -0,0 +1,35 @@
+@props([
+    'isOfficial' => true,
+    'numAchievements' => 0,
+    'numMissableAchievements' => 2,
```
Jamiras, line 4: This doesn't appear to be getting passed through. Every page I visit has 2 missable achievements

resources/js/alpine/toggleAchievementRowsComponent/toggleAchievementRowsComponent.ts
```diff
@@ -0,0 +1,25 @@
...
+
+function toggleNonMissableRows(): void {
+  const allRows = document.querySelectorAll<HTMLLIElement>('#set-achievements-list > li');
+
+  allRows.forEach((row) => {
+    if (!row.classList.contains('missable-row')) {
+      row.classList.toggle('hidden');
+    }
+  });
+}
```
Jamiras, line 21: There's some weird interactions when toggling both states: i.e. if I hide unlocked achievements, then click the missables link, I get a list of all missables, not just locked ones - and the hide unlocked achievements button remains selected. Similarly, if I select missables first, then hide unlocked, I think it's showing me all achievements

resources/views/platform/components/game/achievements-list-meta.blade.php
```diff
@@ -0,0 +1,35 @@
...
+        {{ mb_strtolower(__res('point', $totalPossible)) }}.
+    </p>
+
+    @if ($numMissableAchievements > 0)
+        <button class="flex items-center gap-x-1 hover:underline hover:text-link-hover text-link" @click="toggleNonMissableRows">
+            <div class="w-5 h-5 p-0.5 rounded-full bg-embed border border-dashed border-stone-500 text-white">
+                <x-icon.missable />
+            </div>
+            This set has {{ $numMissableAchievements }} missable {{ mb_strtolower(__res('achievement', $numMissableAchievements)) }}.
+        </button>
```
Jamiras, line 33: I don't really link this button style. It looks like a link, which says to me that it'll open a new page.

wescopeland: When it takes a standard `.btn` classname, it becomes fairly loud on the page. Maybe we need a small meta surface above the achievements list which contains filtering / sorting options similar to other pages (ie: completion progress, most requested sets, recent game awards, etc).

resources/views/platform/components/game/achievements-list-meta.blade.php
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/2077/commits/2792b2efbe4231578b6970fb796742c4d9725a39.

resources/views/platform/components/game/achievements-list-meta.blade.php
wescopeland: In latest, this is now using a `.btn` class.

Screenshot 2023-12-05 at 6 00 12 PM

On clicking the button, the filter is applied and the text changes informing the user they can click again to reset the view.

I do think we should rework all the filtering and sorting using Livewire, but it's a fairly large undertaking likely beyond the scope of this PR.

resources/js/alpine/toggleAchievementRowsComponent/toggleAchievementRowsComponent.ts
wescopeland: In latest, it is no longer possible to enable both states.

If state A is enabled and the user attempts to enable state B, then state A will be forcefully disabled.

resources/views/platform/components/game/achievements-list-meta.blade.php
Jamiras: I agree that a better filtering system should be a separate PR, but I'm still not comfortable putting behavior on the summary detail. What about adding a second check mark?
image

Jamiras: Or maybe separate the button from the summary?
image

But at that point, the checkbox feels more consistent.

wescopeland: I think I like the checkbox most, as it keeps the sorting/filtering function in the same centralized location.

Jamiras (approved):

PR #2085, opened 2023-12-08
feat: Discord notifications for system user messages

luchaos:
Stage and production environment variables have been configured accordingly.


app/Community/Listeners/NotifyMessageThreadParticipants.php
```diff
@@ -61,6 +62,84 @@ public function handle(object $event): void
             if (BitSet($userTo->websitePrefs, UserPreference::EmailOn_PrivateMessage)) {
                 sendPrivateMessageEmail($userTo->User, $userTo->EmailAddress, $thread->title, $message->body, $userFrom->User);
             }
+
+            $this->forwardToDiscord($userFrom, $userTo, $thread, $message);
```
Jamiras, line 66: Shouldn't this be it's own listener?

Jamiras (approved):
Untested, but it looks like you've been testing it in the secondary discord server already, so I trust the changes.

app/Community/Listeners/NotifyMessageThreadParticipants.php
luchaos: I started with a separate listener but realized that it would include pretty much all the same code as the rest of this listener. I added it to this one to save some time to make get this feature back asap for the affected roles.
This line and the sending of mails will be refactored into a notification system which supports both mails and discord in the future.

Issue #618, opened 2021-06-15
allow retrieving game data from hash


Jamiras:
As a client developer, I'd like to be able to fetch the achievements/leaderboards/rich presence data for a game using the same server call as resolving the hash.

Almost every time a hash is resolved via the `gameid` API, the next thing to do is to call the `patch` API with the resulting game ID.

I'd like to propose passing the hash to the `patch` API and have it return the patch data (which includes the game ID) if the hash resolves, or a null `PatchData` if the hash does not resolve (possibly even a Success=false,Error='Unknown Hash'). The API currently returns empty arrays when passed a non-existant game ID. This would also be acceptable as long as the `PatchData.ID` field is not set, it's obvious the game doesn't exist.
```
{"Success":true,"PatchData":{"Achievements":[],"Leaderboards":[]}}
```

This would avoid a second server request when loading a game, making it load just a little bit faster.

luchaos:
Makes sense and should be easy to implement - in a backwards compatible fashion even. Ready for v1 anytime imo

Jamiras:
Added in #2857

PR #2090, opened 2023-12-09
feat(api): add API_GetGameFromHash endpoint

wescopeland:
This PR adds a new API endpoint: `API_GetGameFromHash.php`.

The endpoint accepts a single argument: `h`, which corresponds to a 32-character md5 string most commonly seen on the "Supported Game Files" page.

If a matching hash cannot be found, a 400 status code and message are returned.
If a matching hash is found, high-level minimal metadata regarding the game is returned.

```
GET http://localhost:64000/API/API_GetGameFromHash.php?z=Scott&y=x&h=2bc674be034e43c96b86487ac69d9293

{ "message": "Could not find a game matching the given hash." }
```

```
GET http://localhost:64000/API/API_GetGameFromHash.php?z=Scott&y=x&h=1bc674be034e43c96b86487ac69d9293

{
  "ID": 1,
  "Title": "Sonic the Hedgehog",
  "ConsoleID": 1,
  "ConsoleName": "Mega Drive"
}
```

### DIFF
--- a/public/API/API_GetGameFromHash.php
+++ b/public/API/API_GetGameFromHash.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ *  API_GetGameFromHash - looks up a game from a given hash string
+ *    h : game hash
+ *
+ *  int        ID                         id of the game
+ *  string     Title                      name of the game
+ *  int        ConsoleID                  unique identifier of the console associated to the game
+ *  string     ConsoleName                name of the console associated to the game
+ */
+
+use App\Platform\Models\GameHash;
+use App\Platform\Models\System;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'h' => ['required', 'size:32'],
+]);
+
+$givenHash = request()->query('h');
+
+$foundHash = GameHash::with('game')->where('MD5', $givenHash)->first();
+
+if (!$foundHash) {
+    return response()->json(['message' => 'Could not find a game matching the given hash.'], 400);
+}
+
+$foundSystem = System::find($foundHash->game->ConsoleID);
+
+return response()->json([
+    'ID' => $foundHash->game->ID,
+    'Title' => $foundHash->game->Title,
+    'ConsoleID' => $foundHash->game->ConsoleID,
+    'ConsoleName' => $foundSystem?->Name ?? null,
+]);
--- a/public/API/API_GetGameFromHash.php
+++ b/public/API/API_GetGameFromHash.php
@@ -24,7 +24,11 @@ $givenHash = request()->query('h');
 $foundHash = GameHash::with('game')->where('MD5', $givenHash)->first();
 
 if (!$foundHash) {
-    return response()->json(['message' => 'Could not find a game matching the given hash.'], 400);
+    return response()->json([
+        'status' => 404,
+        'code' => 'not_found',
+        'error' => "Unknown hash: $givenHash",
+    ], 404);
 }
 
 $foundSystem = System::find($foundHash->game->ConsoleID);
@@ -32,6 +36,7 @@ $foundSystem = System::find($foundHash->game->ConsoleID);
 return response()->json([
     'ID' => $foundHash->game->ID,
     'Title' => $foundHash->game->Title,
+    'Description' => $foundHash->Name,
     'ConsoleID' => $foundHash->game->ConsoleID,
     'ConsoleName' => $foundSystem?->Name ?? null,
 ]);
--- a/public/API/API_GetGameFromHash.php
+++ b/public/API/API_GetGameFromHash.php
@@ -6,6 +6,7 @@
  *
  *  int        ID                         id of the game
  *  string     Title                      name of the game
+ *  string     Description                description of the associated hash
  *  int        ConsoleID                  unique identifier of the console associated to the game
  *  string     ConsoleName                name of the console associated to the game
  */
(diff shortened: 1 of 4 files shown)


public/API/API_GetGameFromHash.php
```diff
@@ -0,0 +1,37 @@
...
+    return response()->json(['message' => 'Could not find a game matching the given hash.'], 400);
+}
+
+$foundSystem = System::find($foundHash->game->ConsoleID);
+
+return response()->json([
+    'ID' => $foundHash->game->ID,
+    'Title' => $foundHash->game->Title,
+    'ConsoleID' => $foundHash->game->ConsoleID,
+    'ConsoleName' => $foundSystem?->Name ?? null,
```
Jamiras, line 36: Would it be worth returning the hash description?
```
'Description' => $foundHash->name,
```

```
{
    "ID":1,
    "Title":"Sonic the Hedgehog",
    "Description":"Sonic The Hedgehog (USA, Europe).md",
    "ConsoleID":1,
    "ConsoleName":"Mega Drive"
}
```

public/API/API_GetGameFromHash.php
```diff
@@ -0,0 +1,37 @@
...
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'h' => ['required', 'size:32'],
+]);
+
+$givenHash = request()->query('h');
+
+$foundHash = GameHash::with('game')->where('MD5', $givenHash)->first();
+
+if (!$foundHash) {
+    return response()->json(['message' => 'Could not find a game matching the given hash.'], 400);
```
Jamiras, line 27: not found should return 404.

Jamiras, line 27: Recommend using `error` instead of `message`, so it's at least consistent with the API_GetTicketData endpoint:
https://github.com/RetroAchievements/RAWeb/blob/3d5d845ea70737f90d9a35eb867111a87018d387/public/API/API_GetTicketData.php#L154

API_GetAchievementUnlocks, API_GetUserPoints, and API_GetUserSummary just echo back the inputs on a 404 not found response.

Also consider a more expanded response as described [here](https://discord.com/channels/476211979464343552/757767535293890682/1132370669938671758).

```
HTTP/1.1 404 Not Found
{
  "status": 404,
  "code": "not_found",
  "error": "Unknown hash: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
}
```
> This kinda matches or at least come closest to a JSON:API style response (https://jsonapi.org/format/#error-objects); plans are to do a similar thing for the Web API.

@luchaos probably has a specific structure in mind.

wescopeland: Doesn't this imply the endpoint itself doesn't exist? In this case, the request is successful, but the server can't do anything with what the client has provided (4xx).

public/API/API_GetGameFromHash.php
wescopeland: Probably so, added in https://github.com/RetroAchievements/RAWeb/pull/2090/commits/8ec90f1122c9f8f7847bdfdebacc32a8be1e9d4a.

public/API/API_GetGameFromHash.php
wescopeland: Updated in https://github.com/RetroAchievements/RAWeb/pull/2090/commits/8ec90f1122c9f8f7847bdfdebacc32a8be1e9d4a.

wescopeland: Switched to `error` and the expanded response structure in https://github.com/RetroAchievements/RAWeb/pull/2090/commits/8ec90f1122c9f8f7847bdfdebacc32a8be1e9d4a.

public/API/API_GetGameFromHash.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+/**
+ *  API_GetGameFromHash - looks up a game from a given hash string
+ *    h : game hash
+ *
+ *  int        ID                         id of the game
+ *  string     Title                      name of the game
+ *  int        ConsoleID                  unique identifier of the console associated to the game
+ *  string     ConsoleName                name of the console associated to the game
```
Jamiras, line 10: Doc comments should include Description

public/API/API_GetGameFromHash.php
Jamiras: fwiw, hitting a non-existing endpoint returns the error in a nested structure.
```
{"message":"Not Found","errors":[{"status":404,"code":"not_found","title":"Not Found"}]}
```

https://github.com/RetroAchievements/RAWeb/blob/cf9953f7181df2c45a44ced1d39a10f30b3d5f2f/app/Exceptions/Handler.php#L101-L110

public/API/API_GetGameFromHash.php
wescopeland: Good catch, added in latest.

public/API/API_GetGameFromHash.php
wescopeland: Adjusted in latest to follow a similar structure:

```json
{
	"message": "Unknown hash: 2bc674be034e43c96b86487ac69d9293",
	"errors": [
		{
			"status": 404,
			"code": "not_found",
			"title": "Unknown hash: 2bc674be034e43c96b86487ac69d9293"
		}
	]
}
```

wescopeland:
Based on feedback from consumers, it may be more appropriate for this endpoint to accept an array of hashes, probably via a comma-separated list. Barring any objections, I'd like to convert this to something like `API_GetGamesFromHashes`, which can then return found/not-found values given a library of hashes.

luchaos:
This seems to address https://github.com/RetroAchievements/RAWeb/issues/618 if I'm not mistaken?

wescopeland:
It turns out this is already mostly supported by `API_GetGameList.php`, which can be specified by a consumer to return hashes associated with all sets for a console.

PR #2092, opened 2023-12-09
Add API endpoint for minimal user profile information

relm:
This PR adds a new endpoint to retrieve minimal user profile information - `API_GetUserProfile.php`

We use the existing `API_GetUserSummary.php` endpoint to check if a user exists as well as other details like their untracked/tracked status and then pull additional profile information based on that.  This results in some unnecessary DB queries as we only really need fields from the main user table. So I propose adding this as a lighter alternative.

One argument is accepted - `u` which represents the user's username

Example output:

```
{
	"User": "Root",
	"MemberSince": "2023-12-09 19:24:46",
	"RichPresenceMsg": null,
	"LastGameID": 0,
	"ContribCount": 0,
	"ContribYield": 0,
	"TotalPoints": 1000,
	"TotalSoftcorePoints": 0,
	"TotalTruePoints": 1010,
	"Permissions": 4,
	"Untracked": 0,
	"ID": 1000001,
	"UserWallActive": 1,
	"Motto": "I am root",
	"UserPic": "\/UserPic\/root.png"
}
```


wescopeland (changes requested):
Thanks for submitting a PR.

Please add tests for any new endpoint.

Also, let's not add new functionality to any _Helpers_ file unless absolutely necessary. We plan on phasing these out soon. As an alternative, perhaps it would just be best for this new functionality to live within the endpoint itself.

wescopeland:
https://github.com/RetroAchievements/RAWeb/pull/2090 may be a helpful reference.

relm:
Done! Moved the functionality to the endpoint itself and added tests.

public/API/API_GetUserProfile.php
```diff
@@ -0,0 +1,54 @@
...
+ *  int        ID                      unique identifier of the user
+ *  int        TotalPoints             number of hardcore points the user has
+ *  int        TotalSoftcorePoints     number of softcore points the user has
+ *  int        TotalTruePoints         number of "white" points the user has
+ *  int        Permissions             unique identifier of user's account type
+ *  datetime   MemberSince             when the user joined the site
+ *  int        Untracked               "1" if the user is untracked, otherwise "0"
+ *  string     UserPic                 site-relative path to the user's profile picture
+ *  string     Motto                   the user's motto
+ *  int        UserWallActive          "1" if the user allows comments to be posted to their wall, otherwise "0"
```
wescopeland, line 17: We should document that a "User" key of type `string` is also returned.

public/API/API_GetUserProfile.php
```diff
@@ -0,0 +1,54 @@
...
+    'LastGameID' => (int) $user->LastGameID,
+    'ContribCount' => (int) $user->ContribCount,
+    'ContribYield' => (int) $user->ContribYield,
+    'TotalPoints' => (int) $user->RAPoints,
+    'TotalSoftcorePoints' => (int) $user->RASoftcorePoints,
+    'TotalTruePoints' => (int) $user->TrueRAPoints,
+    'Permissions' => (int) $user->getAttribute('Permissions'),
+    'Untracked' => (int) $user->Untracked,
+    'ID' => (int) $user->ID,
+    'UserWallActive' => (int) $user->UserWallActive,
```
wescopeland, line 52: I've verified that if all of these `int` casts are removed, the endpoint still returns integer values for all of these fields.

public/API/API_GetUserProfile.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+use App\Site\Models\User;
```
wescopeland, line 3: nit: Let's move this down to live with the other imports in the file.

wescopeland (commented):
Looks great. A few minor nits.

relm:
Should be good now!

wescopeland (approved):

public/API/API_GetUserProfile.php
```diff
@@ -0,0 +1,55 @@
...
+/*
+ *  API_GetUserProfile
+ *    u : username
+ *
+ *  string     User                    name of user
+ *  int        ID                      unique identifier of the user
+ *  int        TotalPoints             number of hardcore points the user has
+ *  int        TotalSoftcorePoints     number of softcore points the user has
+ *  int        TotalTruePoints         number of "white" points the user has
+ *  int        Permissions             unique identifier of user's account type
```
Jamiras, line 12: Is there some benefit to exposing the user's permissions? We're going to be overhauling that functionality in the near future.

public/API/API_GetUserProfile.php
```diff
@@ -0,0 +1,55 @@
...
+ *  datetime   MemberSince             when the user joined the site
+ *  int        Untracked               "1" if the user is untracked, otherwise "0"
+ *  string     UserPic                 site-relative path to the user's profile picture
+ *  string     Motto                   the user's motto
+ *  int        UserWallActive          "1" if the user allows comments to be posted to their wall, otherwise "0"
```
Jamiras, line 17: Is `UserWallActive` really important to the caller? It's not like they can see the user wall or append to it.

public/API/API_GetUserProfile.php
```diff
@@ -0,0 +1,55 @@
...
+$user = User::firstWhere('User', request()->query('u'));
+
+if (!$user) {
+    return response()->json([], 404);
+}
+
+return response()->json([
+    'User' => $user->User,
+    'UserPic' => sprintf("/UserPic/%s.png", $user->User),
+    'MemberSince' => $user->Created?->__toString(),
```
Jamiras, line 42: I believe `->toDateTimeString()` is preferred. `__toString` is a magic php helper method for implicit conversions.

tests/Feature/Api/V1/UserProfileTest.php
```diff
@@ -0,0 +1,56 @@
...
+        /** @var User $user */
+        $user = User::factory()->create();
+
+        $this->get($this->apiUrl('GetUserProfile', ['u' => $user->User]))
+            ->assertSuccessful()
+            ->assertJson([
+                'User' => $user->User,
+                'UserPic' => sprintf("/UserPic/%s.png", $user->User),
+                'MemberSince' => $user->Created?->__toString(),
+                'RichPresenceMsg' => empty($user->RichPresenceMsg) || $user->RichPresenceMsg === 'Unknown' ? null : $user->RichPresenceMsg,
```
Jamiras, line 42: Duplicating the logic in the unit test doesn't really validate the expectations. You should have an explicit expectation of `$user->richPresenceMsg` or `null`.

public/API/API_GetUserProfile.php
luchaos: I was wondering the same, however it is included in the summary, too. Might as well stay consistent in this case?
It shouldn't be included in the next iteration of the API imo.

public/API/API_GetUserProfile.php
luchaos: Same as `UserWallActive` - shouldn't be included but it's present in summary, too.

Do we want to remove it from both?

wescopeland: `Permissions` seems reasonable to remove.

It may make sense for a future iteration of the API allows someone to read a user's wall posts. That's the only reason I can think of to continue exposing `UserWallActive`.

Jamiras: API_GetUserSummary just dumps the data returned by the query for populating the user page, which does display the user's "Account Type" (permissions) and the user wall.

relm: We currently use the Permissions field to check if a user is banned from the site and remove them from tracking. I don't know if it would be better to just a return a 404 for these types of users (as looking on the site itself, a not found page is returned when accessing a banned profile it seems). 

But I just kept the behavior the same as API_GetUserSummary in that regard - maybe it's better to tackle that in a future iteration of the API.

luchaos: I agree, let's keep it as is for now?
I'd consider this iteration of the API to be v1 and we should not change it too much to meet current clients' expectations.
Using dedicated resource classes which map to the current schema and payload would make it more robust - this would prevent unexpected payload changes because of query changes. We now have characterization tests for this in place to refactor such things confidently. I'll look into that in a separate PR.

PR #2095, opened 2023-12-10
fix: redirect to intended url without errors

luchaos:
The fix itself would've been to just rephrase `redirect(intended)` to `redirect()->to(intended)`. This PR removes the custom `intended_url` implementation and uses the framework's while at it.
Some related adjustments for Fortify are included to support the upcoming two-factor authentication.

How to replicate:
- [add sleep(5) locally to searchresults.php](https://github.com/RetroAchievements/RAWeb/compare/master...luchaos:redirect-intended-url?expand=1#diff-6fb5823eeafa72be7ea42fad67d6d6fd258f34733bb40fab3df2c45fdc6cbccdR27) to make sure it finishes loading after the login page
- go to front page, logged out
- ctrl-click search icon (new tab)
- ctrl-click "Sign in" (new tab)
- switch to login tab
- wait for the search tab to finish loading
- login
- error: `Symfony\Component\HttpFoundation\Response::setContent(): Argument #1 ($content) must be of type ?string, Illuminate\Routing\Redirector given`


Jamiras (approved):
I couldn't recreate the error with the provided steps, but I didn't experience any issues with the changes. So if this finally gets rid of that last weird error log message, I'm all for it!

luchaos:
> So if this finally gets rid of that last weird error log message, I'm all for it!

It should! All for clean logs :)

PR #2097, opened 2023-12-11
chore: update larastan package

luchaos:
Package moved to new GitHub org:
> This package is abandoned and no longer maintained. The author suggests using the larastan/larastan package instead.


config/logging.php
wescopeland, line 1: Seems like there are at least 3 instances of duplicate `replace_placeholders` keys in this file.

luchaos: whoops, nice catch, thanks! those are from a bad rebase

wescopeland (approved):
Verified after running `composer setup` and then `composer analyse`, the command continues to execute successfully.

PR #2099, opened 2023-12-11
ignore placeholders not replaced by missing-page-redirector

Jamiras:
Fixes '{s}' being left in the subject field when clicking links going to the old create message page.


luchaos (approved):
As mentioned, I'll look into a more generalized solution.

PR #2100, opened 2023-12-11
feat: improve legacy redirector's handling of unused query parameters

luchaos:


app/Support/Redirector/LegacyRedirector.php
```diff
@@ -27,11 +28,21 @@ public function getRedirectsFor(Request $request): array
...
+                // forward query string values
                 foreach ($queryParams as $key => $value) {
                     $redirectUrl = str_replace("{{$key}}", $value, $redirectUrl);
                 }
 
-                return [$request->getPathInfo() => $redirectUrl];
+                // remove remaining, unused query string markers
+                $parsedRedirectUrl = URL::fromString($redirectUrl);
+                foreach ($parsedRedirectUrl->getAllQueryParameters() as $queryParameter => $queryParameterValue) {
+                    if (str_starts_with($queryParameterValue, '{')) {
```
Jamiras, line 40: Though unlikely, this could falsely match a string that does, for some reason, start with a `{`. What about this?
```
                $queryParams = Query::parse($request->getQueryString());
                $redirectUrl = $redirects[$request->getPathInfo()];

                // replace query string markers
                $parsedRedirectUrl = URL::fromString($redirectUrl);
                foreach ($parsedRedirectUrl->getAllQueryParameters() as $queryParameter => $queryParameterValue) {
                    if (str_starts_with($queryParameterValue, '{') && str_ends_with($queryParameterValue, '}')) {
                        $key = substr($queryParameterValue, 1, -1);
                        if (array_key_exists($key, $queryParams)) {
                            // found a replacement, substitute it
                            $parsedRedirectUrl = $parsedRedirectUrl->withQueryParameter($queryParameter, $queryParams[$key]);
                        } else {
                            // did not find a replacement, discard it
                            $parsedRedirectUrl = $parsedRedirectUrl->withoutQueryParameter($queryParameter);
                        }
                    }
                }

                return [$request->getPathInfo() => (string) $parsedRedirectUrl];
```

Jamiras: Oh, but that wouldn't work for placeholders in the path...

Jamiras (approved):

app/Support/Redirector/LegacyRedirector.php
luchaos: I like it! ❤️ slightly adjusted and updated in latest: https://github.com/RetroAchievements/RAWeb/pull/2100/commits/3fc88194b01ec21b65dd26b556a0682e1fe7470f

luchaos: > Oh, but that wouldn't work for placeholders in the path...

good point - forgot about those 🤔 I'll revert then. The chances of query parameters including a valid `{` when redirecting are very slim

luchaos: done. rephrased comments to make sure both routes and query strings are affected

PR #2106, opened 2023-12-14
fix: migrate missable display pref detection to achievement list item

wescopeland:
This PR migrates the achievement list item's missable user pref detection to the type indicator component itself.

I initially was hesitant to do this because I didn't understand the performance implications of calling `Auth::user()` within a Blade component, however I've since learned that this isn't anything to worry about as there are no additional hits to the DB caused by doing this.

With this change, if the user has enabled the pref for not showing missable indicators, they will continue to not appear on game pages, and now they will no longer appear on achievement info pages due to the component sharing we're doing on that page.


resources/views/platform/components/game/achievements-list/type-indicator.blade.php
```diff
@@ -21,61 +23,71 @@
...
-                    group-hover:md:ml-1 group-hover:md:mr-2 group-hover:md:translate-x-0
-                    select-none font-semibold
-                ">
-                    Progression
-                </span>
-                <div class="w-[18px] h-[18px]" aria-label="Progression">
-                    <x-icon.progression />
-                </div>
-            @endif
+@if ($canRender)
```
wescopeland, line 36: The only thing changed within this block is white space.

Jamiras (approved):
Should the user setting also hide the count and checkbox?
image

wescopeland:
> Should the user setting also hide the count and checkbox?

Yes it should, thanks. Addressed in latest.

Jamiras (approved):

PR #2107, opened 2023-12-15
feat: add woz5 label

luchaos:
Adds woz5 to https://static.retroachievements.org/assets/images/labels/woz5.png

Note: the "real" URL should actually be 
https://static.retroachievements.org/assets/images/labels/atariage.png
instead of
https://retroachievements.org/Images/labels/atariage.png

Would be nice if we could clean that up some day - those labels are referenced a lot in forums afaik.


wescopeland (approved):

public/assets/images/labels/woz5.png
Jamiras, line 1: Why `woz5` instead of `wozaday`? The image matching code expects to find the labels from the database:
https://github.com/RetroAchievements/RAWeb/blob/010b8136f3af0f7c04656d45481ed3c98568e9d6/public/linkedhashes.php#L79-L84
```
> select Labels, count(*) from GameHashLibrary where Labels like '%woz%' group by 1 order by 2 desc;
+-----------+----------+
| Labels    | count(*) |
+-----------+----------+
| wozaday   |       19 |
| woz-a-day |        1 |
+-----------+----------+
2 rows in set (0.007 sec)
```

luchaos: I was curious about that `5` in the name. Should be `wozaday` then @televandalist ?

luchaos: preemptively updated in https://github.com/RetroAchievements/RAWeb/pull/2107/commits/37c3308a84393df1cfd62fc148676fb06d3c0444

PR #2110, opened 2023-12-16
feat(hashes): rewrite the Manage Hashes page, add new fields

wescopeland:
This PR rewrites the Manage Hashes page for games. It also adds two new fields to the form.

See: https://discord.com/channels/310192285306454017/1085433484048343050

**Before**
Screenshot 2023-12-16 at 5 31 24 PM

**After**
Screenshot 2023-12-16 at 5 30 41 PM

As a result of these changes, several desirable features are introduced:
* Hashes can now be linked to a RAPatches repo URL. This is powered by a new `internal_patch_url` column on `GameHashLibrary`.
* Hashes can now be linked to a source URL (eg: No Intro, Redump, RHDN, SMWCentral, itch.io).
* A new (currently unused) `internal_status` column has been added to `GameHashLibrary`, which will theoretically accept five possible values: NULL, Unverified, In Progress, Verified, or Problematic.

**Technical notes**
* Game sidebar links (Official Forum Topic, Supported Game Files, etc.) have been migrated to Blade in order to re-use them on this page in a way that makes sense.
* There is now a game breadcrumbs component. This PR does not proliferate it throughout the codebase.
* A future PR, branched from here, will utilize the new `internal_status` column.
* Server logs will show changes to the two new URL fields, including removals.
* `managehashes.php` has been deleted.

The new hash management page is accessible via `/game/:gameId/hashes/manage`.

---

https://github.com/RetroAchievements/RAWeb/assets/3984985/08e26768-6706-4987-8d4e-2244c57db095


database/migrations/2023_12_16_000000_update_gamehashlibrary_table.php
wescopeland, line 1: @luchaos FYI

Also: `source` is being used for the No Intro, Redump, RHDN, SMWCentral, itch.io, etc. URL. I believe that is indeed the intent of the field, but I want to be sure.

resources/views/platform/manage-hashes-page.blade.php
```diff
@@ -0,0 +1,91 @@
...
+                <x-slot name="title">Warning</x-slot>
+                <x-slot name="description">
+                    <div class="flex flex-col">
+                        <p>
+                            PLEASE be careful when using this tool. Mistakes can cause a lot of tickets to be created.
+                        </p>
+
+                        <p>
+                            If you're not <span class="underline font-semibold">100% sure</span> of what you're doing,
+                            <a href="{{ route('message.create') . '?to=RAdmin&subject=Attempt+to+Unlink+' . urlencode($game->Title) . '&message=%5Bgame=' . $game->ID . '%5D' }}">
```
Jamiras, line 49: I wonder if the subject line for this message should be changed. This page allows for so much more than just unlinking.

Jamiras (commented):
The new fields (patch/resource) don't appear outside of the manage hashes page. I would expect them to at least appear in the Supported Game Files page. I'm guessing that's coming as a separate future PR?

resources/views/platform/components/manage-hashes/pretty-hash.blade.php
```diff
@@ -0,0 +1,7 @@
+@props([
+    'hash' => '12345678901234567890123456789012',
+])
+
+<p class="font-mono text-neutral-200 light:text-neutral-700 cursor-help" title="{{ $hash }}">
+    {{ substr($hash, 0, 4) . '...' . substr($hash, -4) }}
```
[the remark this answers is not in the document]
wescopeland: Removed in latest.

resources/views/platform/components/manage-hashes/hash-entity.blade.php
```diff
@@ -0,0 +1,159 @@
...
+                    >
+                        RAPatches URL
+                        <x-fas-info-circle class="text-sm" />
+                    </label>
+
+                    <input
+                        id="{{ 'HASH_' . $hashEntity->MD5 . '_InternalPatchURL' }}"
+                        class="w-full"
+                        value="{{ $hashEntity->internal_patch_url }}"
+                        placeholder="https://github.com/RetroAchievements/RAPatches/blob/main/NES/Subset/5136-CastlevaniaIIBonus.zip"
```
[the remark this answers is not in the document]
wescopeland: Agreed, it is not currently very accessible.

Updated in latest with both dark and light mode considerations here:
https://github.com/wescopeland/RAWeb/blob/manage-hashes-blade/resources/css/utilities.css

resources/views/platform/components/manage-hashes/hash-entity.blade.php
```diff
@@ -0,0 +1,159 @@
...
+            >
+                Unlink
+            </button>
+        </div>
+    </div>
+
+    <div class="p-4">
+        @if ($hashEntity->User)
+            <p class="mb-4">
+                Linked by {!! userAvatar($hashEntity->User, icon: false) !!}
```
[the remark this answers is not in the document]
wescopeland: Done in latest - there is some special logic to see if there's both `User` and `Created`, or only `Created`. It's also possible both of them are null.

app/Helpers/database/hash.php
```diff
@@ -142,40 +143,21 @@ function removeHash(string $user, int $gameID, string $hash): bool
...
-    $dbResult = mysqli_query($db, $query);
-
-    if (!$dbResult) {
-        log_sql_fail();
-    }
-
-    addArticleComment("Server", ArticleType::GameHash, $gameID, $hash . " updated by " . $user . ". File Name: \"" . $name . "\". Label: \"" . $labels . "\"");
-
-    return $dbResult != null;
+    return $affectedRows > 0;
```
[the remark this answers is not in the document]
wescopeland: Agreed - I've moved this to `GameHashController::update` and deleted _update.php_. I've also removed _delete.php_ and migrated its functionality to `GameHashController::destroy`.

app/Platform/Models/GameHash.php
```diff
@@ -39,9 +39,12 @@ class GameHash extends BaseModel
         'hash',
         'Label',
         'MD5',
+        'internal_status',
         'Name',
         'system_id',
         'User',
+        'source',
+        'internal_patch_url',
```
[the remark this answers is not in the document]
wescopeland: Changed to `patch_url` and `compatibility` in latest.

app/Platform/RouteServiceProvider.php
```diff
@@ -59,6 +60,7 @@ protected function mapWebRoutes(): void
             // Route::get('game/{game}/assets', [GameAssetsController::class, 'index'])->name('game.asset.index');
             // Route::get('game/{game}/players', [GamePlayerController::class, 'index'])->name('game.player.index');
             Route::get('game/{game}/dev-interest', GameDevInterestController::class)->name('game.dev-interest');
+            Route::get('game/{game}/hashes/manage', ManageHashesController::class)->name('game.manage-hashes');
```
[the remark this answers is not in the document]
wescopeland: I did quite a bit of research on this and landed somewhere other than `index`.

It seems the standard conventions are mirroring RESTful architecture. The manage hashes page is somewhere at the boundary of that, largely because it allows the user to see multiple hashes (so it doesn't fit neatly with `::edit()`) but only manage one per API call.

Based on what I've studied, `index` is more appropriate for the "Supported Game Files" page.
`edit` would be appropriate if the page only showed a single resource for editing.
In the absence of a standard conventional name, one can be invented that is most appropriate for the use case; I chose `manage` and changed the name to "game.hashes.manage".

public/comments.php
```diff
@@ -51,7 +51,7 @@
         $commentsLabel = "Hash Comments";
         $navPath = [
             '_GamePrefix' => renderGameBreadcrumb($game->ID),
-            'Manage Hashes' => '/managehashes.php?g=' . $game->ID,
+            'Manage Hashes' => '/game/' . $game->ID . '/hashes/manage',
```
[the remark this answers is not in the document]
wescopeland: Updated in latest, however see https://github.com/RetroAchievements/RAWeb/pull/2110#discussion_r1429273725 for more context.

public/request/game-hash/update.php
```diff
@@ -8,15 +10,48 @@
...
-$input = Validator::validate(Arr::wrap(request()->post()), [
+$input = Validator::make(Arr::wrap(request()->post()), [
     'game' => 'required|integer|exists:GameData,ID',
     'hash' => 'required|string',
     'name' => 'required|string',
     'labels' => 'required|string',
-]);
+    'internal_patch_url' => 'nullable|url|regex:/github\.com\/RetroAchievements\/RAPatches\/blob\/main\/.*\.zip$/i',
+    'source' => 'nullable|url',
+])->validate();
```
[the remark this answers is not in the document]
wescopeland: There is - it grants additional flexibility to the caller to manage custom rules and messages - none of which is applicable here. Updated in latest.

resources/views/platform/components/game/link-buttons/index.blade.php
```diff
@@ -0,0 +1,98 @@
...
+$canSeeSetRequestors = $me && $me->Permissions >= Permissions::Registered && $gameAchievementsCount === 0;
+
+$numOpenTickets = countOpenTickets(
+    !$isViewingOfficial,
+    TicketFilters::Default,
+    null,
+    null,
+    null,
+    $gameId,
+)
```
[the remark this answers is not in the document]
wescopeland: Updated in latest to:

```php
if (in_array('tickets', $allowedLinks) && $canSeeOpenTickets) {
    $numOpenTickets = countOpenTickets(
        !$isViewingOfficial,
        TicketFilters::Default,
        null,
        null,
        null,
        $gameId,
    );
}
```

resources/views/platform/components/game/link-buttons/index.blade.php
```diff
@@ -0,0 +1,98 @@
...
+?>
+
+<ul class="flex @if ($variant === 'stacked') flex-col @endif gap-2">
+    @if (in_array('forum-topic', $allowedLinks) && $doesForumTopicExist)
```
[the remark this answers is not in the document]
wescopeland: Per your feedback in https://github.com/RetroAchievements/RAWeb/pull/2110#discussion_r1429182241, this should now be a little easier to reason about.

resources/views/platform/components/game/link-buttons/index.blade.php
```diff
@@ -0,0 +1,98 @@
...
+        <x-game.link-buttons.game-link-button
+            icon="💬"
+            href="{{ '/viewtopic.php?t=' . $gameForumTopicId }}"
+        >
+            Official Forum Topic
+        </x-game.link-buttons.game-link-button>
+    @endif
+
+    @if (in_array('forum-topic', $allowedLinks) && $canCreateForumTopic)
```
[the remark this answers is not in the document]
wescopeland: Done in latest.

resources/views/platform/components/manage-hashes/hashes-list.blade.php
```diff
@@ -0,0 +1,19 @@
+@props([
+    'gameId' => 1,
+    'hashes' => null, // Collection<GameHash>
+    'myUsername' => '',
+])
+
+<div class="flex flex-col gap-y-4">
+    @if ($hashes->isEmpty())
+        There are currently no hashes associated with this game.
```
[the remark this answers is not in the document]
wescopeland: Agreed, removed the empty state.

resources/views/platform/manage-hashes-page.blade.php
wescopeland: I agree, that would make more sense. Changed in latest to:

"Help with Hash Management for {{ gameTitle }}"

wescopeland:
> The new fields (patch/resource) don't appear outside of the manage hashes page. I would expect them to at least appear in the Supported Game Files page. I'm guessing that's coming as a separate future PR?

Yes - this PR is already quite large and touches several domains including the DB, REST API, and front-end. I didn't want to make it much larger than this.

app/Platform/RouteServiceProvider.php
```diff
@@ -24,6 +26,13 @@ public function boot(): void
         Route::pattern('game', '[0-9]{1,17}');
         Route::pattern('game_hash', '[a-zA-Z0-9]{1,32}');
         Route::pattern('system', '[0-9]{1,17}');
+
+        /*
+         * Don't reference hash identifiers by their raw ID
+         */
+        Route::bind('gameMd5Hash', function ($value) {
+            return GameHash::where('MD5', $value)->firstOrFail();
+        });
```
luchaos, line 35: The idea is to migrate all "active" hashes to the `hash` field and use `md5` for the hash that's provided by external sources.
Maybe we could move into that direction by calling the route key `gameHash`?
```php
Route::bind('gameHash', function ($value) {
    return GameHash::where('MD5', $value)->firstOrFail();
});
```

app/Platform/RouteServiceProvider.php
```diff
@@ -60,13 +69,16 @@ protected function mapWebRoutes(): void
             // Route::get('game/{game}/players', [GamePlayerController::class, 'index'])->name('game.player.index');
             Route::get('game/{game}/dev-interest', GameDevInterestController::class)->name('game.dev-interest');
 
+            // Route::get('game/{game}/hashes', [GameHashController::class, 'index'])->name('game.hashes'); "Supported Game Files"
```
luchaos, line 72: Heads up: initially the idea was to list all hashes, patches, save games at `game/{game}/assets` - not sure if a separate hashes page will be needed anymore then?

app/Platform/RouteServiceProvider.php
```diff
@@ -60,13 +69,16 @@ protected function mapWebRoutes(): void
...
+            Route::get('game/{game}/hashes/manage', [GameHashController::class, 'manage'])->name('game.hashes.manage');
```
luchaos, line 73: Management pages require an authenticated user.
To make sure users are prompted to log in and get redirected to this route after the fact I think this (and the CRUD routes below) should be moved into the auth middleware block below.
It needs adjustment though because the `verified` middleware won't work just yet (I'm working on migrating our custom email verification to framework defaults):
```php
Route::group([
    'middleware' => ['auth'], // TODO 'verified'
], function () {
    Route::get('game/{game}/hashes/manage', [GameHashController::class, 'manage'])->name('game.hashes.manage');
    Route::resource('game-hash', GameHashController::class)->parameters(['game-hash' => 'gameMd5Hash'])->only(['update', 'destroy']);
```

database/migrations/2023_12_16_000000_update_gamehashlibrary_table.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('GameHashLibrary', function (Blueprint $table) {
+            $table->string('compatibility')->nullable()->after('file_name_md5'); // Unverified, In Progress, Verified, Problematic
```
luchaos, line 13: Will those values be added as an Enum?
When stored in db it would be nice if they follow the same convention as types for achievements did; lowercase, snake- or kebap-case like unverified, in-progress, verified, problematic

app/Platform/RouteServiceProvider.php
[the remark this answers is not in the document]
luchaos: For resources and sub-resources in route names (not the URL, just the name) i think all the others are singular: `game.hash.manage`

app/Platform/Models/GameHash.php
```diff
@@ -29,19 +29,22 @@ class GameHash extends BaseModel
     // TODO drop game_hashes_md5_unique
     protected $table = 'GameHashLibrary';
 
-    protected $primaryKey = 'ID';
+    protected $primaryKey = 'id';
```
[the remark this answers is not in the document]
wescopeland: Addressed in latest.

app/Platform/RouteServiceProvider.php
wescopeland: Understood, adjustment made in latest.

app/Platform/RouteServiceProvider.php
wescopeland: Commented route removed in latest.

app/Platform/RouteServiceProvider.php
wescopeland: Makes sense - adjusted in latest.

database/migrations/2023_12_16_000000_update_gamehashlibrary_table.php
wescopeland: _HashCompatibility.php_ enum added in latest.

app/Platform/Controllers/GameHashController.php
```diff
@@ -29,15 +34,100 @@ public function show(GameHash $gameHash): void
...
 
-    public function update(Request $request, GameHash $gameHash): void
+    public function update(Request $request, GameHash $gameHash): JsonResponse
     {
+        $this->authorize('update', $this->resourceClass());
+
+        $input = $request->validate([
+            'name' => 'required|string',
+            'labels' => 'required|string',
+            'patch_url' => 'nullable|url|regex:/github\.com\/RetroAchievements\/RAPatches\/blob\/main\/.*\.zip$/i',
```
Jamiras, line 62: Is `blob` the preferred path? @televandalist provided me a link that uses `raw` here: https://discord.com/channels/310192285306454017/1179551921263104112/1179821069461368842

app/Platform/Enums/HashCompatibility.php
```diff
@@ -0,0 +1,31 @@
...
+
+declare(strict_types=1);
+
+namespace App\Platform\Enums;
+
+abstract class HashCompatibility
+{
+    public const Unverified = 'unverified';
+
+    public const InProgress = 'in_progress';
```
Jamiras, line 11: What does "in_progress" mean? Is compatibility being added or tested? Should we have separate values for both of those states? 

Wouldn't a hash be unverified until it was tested, at which point it transitions to verified or problematic?

app/Platform/Models/GameHash.php
```diff
@@ -29,19 +29,20 @@ class GameHash extends BaseModel
...
     public const CREATED_AT = 'Created';
 
     protected $fillable = [
         'description',
         'GameID',
         'hash',
-        'Label',
+        'Labels',
         'MD5',
+        'internal_status',
```
Jamiras, line 40: Wasn't this changed to `compatibility`?

resources/views/platform/components/game/link-buttons/index.blade.php
```diff
@@ -0,0 +1,103 @@
...
+$canCreateForumTopic = !$doesForumTopicExist && $me && $me->Permissions >= Permissions::Developer;
+
+$canSeeForumLink = in_array('forum-topic', $allowedLinks);
+$canSeeSupportedGameFiles = in_array('game-files', $allowedLinks) && $me && $me->Permissions >= Permissions::Registered;
+$canSeeCodeNotes = in_array('code-notes', $allowedLinks) && $me && $me->Permissions >= Permissions::Registered;
+$canSeeGuide = in_array('guide', $allowedLinks) && $gameGuideUrl;
+$canSeeOpenTickets = in_array('tickets', $allowedLinks) && $me && $me->Permissions >= Permissions::Registered;
+$canSeeSetRequestors = in_array('set-requestors', $allowedLinks) && $me && $me->Permissions >= Permissions::Registered && $gameAchievementsCount === 0;
+
+if (in_array('tickets', $allowedLinks) && $canSeeOpenTickets) {
```
Jamiras, line 33: `$canSeeOpenTickets` encapsulates `in_array('tickets, $allowedLinks)` now.

app/Platform/Controllers/GameHashController.php
wescopeland: I think `blob` is indeed more accurate. Try navigating to: https://github.com/RetroAchievements/RAPatches/tree/main/NES/FDSConversion. Then right click on the file and copy the URL. The result should be:

`https://github.com/RetroAchievements/RAPatches/blob/main/NES/FDSConversion/14870-NewCluCluLand-FDSConversion.zip`

app/Platform/Enums/HashCompatibility.php
wescopeland: I brought these over from https://discord.com/channels/310192285306454017/1085433484048343050

To quote @televandalist:

>🟡 Unverified: The hash has not been verified by an admin (or "entry manager" whenever site perms get an overhaul).
🔵 In Progress: The hash is in the current batch of hashes being verified.
🟢 Verified: An admin/entry manager has confirmed the hash is legitimate and info/metadata is accurate.
🔴 Problematic (or whatever term works best here): This would basically be like opening a ticket on a hash. The most common cause of this would be if an admin/entry manager is unable to reproduce the hash. A bad base could have been used, there could have been an issue creating the patch, etc.

app/Platform/Models/GameHash.php
wescopeland: Good catch, fixed in latest.

resources/views/platform/components/game/link-buttons/index.blade.php
wescopeland: Thanks - changed to just `if ($canSeeOpenTickets)`.

app/Platform/Enums/HashCompatibility.php
Jamiras: Oh, that's not compatibility, that's actually validity.

I was thinking:
* Unverified = Untested
* In Progress = Being tested (or being made compatible)
* Verified = Testing complete, version associated to hash is compatible
* Problematic = Testing complete, version is not compatible and should not be used to earn achievements.

We may want to track both, but I think we can combine them:
* Unverified = This hash hasn't been looked at
* Unknown = We couldn't figure out where this hash came from
* Untested = Known hash, but compatibility hasn't been tested (or set doesn't exist)
* Compatible = Known hash, works with the set
* Incompatible = Known hash, does not work with the set

I think "In Progress" is unnecessary - even if multiple people are trying to identify hashes independently, they'll probably do so one at a time, so the user could change it over to Unknown while they are doing the identification.

televandalist: When I'm working on hash/patch stuff, this is my usual process:
- Download an updated .csv of [this query](https://metabase.retroachievements.org/question/234-hash-database) and import into my hash sheet.
- Compare hashes in my hash sheet to the new query to see if anything has been removed.
- Compare hashes in the new query to see what I already have logged and delete those rows. Leftovers are hashes that have been linked since my last update.
- Compare new hashes to ones I have listed in the [RA Hashes repository](https://github.com/RetroAchievements/RAHashes). I start with No Intro and Redump, then work my way down the line. If file names and labels check out, then I'll move those to the "Verified" tab in my sheet.
- Leftovers from that are moved to a "ToDo" tab. These are what I consider "Unverified" and are almost always homebrews, unlicensed, pirate, and ROMs that require a patch of some sort.
  - For those that require patches, I either obtain or make the required patches and put them in the RAPatches repo. Once I reproduce a linked hash, I adjust the file name and label(s) as needed. I then update that hash in my sheet and move it to an "RAPatches" tab, which I also consider "Verified"
  - For select homebrews and other "aftermarket" games, I periodically submit information to No Intro so they end up in their DATs. I've done this for several hundred hashes at this point 😅

So, *to me:*
- **Verified** = A hash's filename and label(s) match those of a DAT by No Intro, Redump, Final Burn Neo, TOSEC, etc. If it requires a patch, then the patch is in the RAPatches repo and I have confirmed the hash is reproduceable. Over 34,000 hashes in our DB match this criteria.
- **Unverified** = The hash does not match one of the above and I have not tried to reproduce. There are currently a little over 200 of these. A small handful of these are just subset patches that haven't been submitted yet.
- **In Progress** = The hash does not match one of the above. I was able to reproduce, but have not added a patch to the RAPatches repo or submitted to No Intro's DATs yet. A lot of these are just GoodTools and NonGood hashes that need to be converted to No Intro + RAPatches. There are currently around 900 of these.
- **Problematic** = I was not able to reproduce the hash. This is very rare and I think there are only a small handful at any given time. These are typically versions of newer hacks or homebrews that I didn't obtain a patch for before an update or just something completely batshit.

Anyway, in the overall scope of things, I agree with Jamiras that "In Progress" is probably unnecessary. I have a tendency to think about this stuff in regards to my own workflow since I've been doing 99% of it for a few years now. 

As for "Compatible" and "Incompatible," I'm not sure. If something is incompatible it shouldn't be linked at all. Or are we planning to store hash info even for hashes that aren't linked? I'd love that, personally.

Jamiras: I'd like to be able to link incompatible hashes, which could be leveraged to return a pseudo-set to the client that would inform the user that they're using a known, but incompatible ROM instead of just saying "no achievements found" or "unknown game" like it does currently.

See #717 and #789.

app/Platform/Controllers/GameHashController.php
Jamiras: But if you actually go to that URL: https://github.com/RetroAchievements/RAPatches/blob/main/NES/FDSConversion/14870-NewCluCluLand-FDSConversion.zip

It takes you to a github page, not the download itself. 

That has a link to the raw zip flile:
https://github.com/RetroAchievements/RAPatches/raw/main/NES/FDSConversion/14870-NewCluCluLand-FDSConversion.zip

Are we linking to the github page or the zip file directly?

app/Platform/Enums/HashCompatibility.php
televandalist: Would we go about that by making dummy entries like the "Cheated ROM" ones? Elmo and I discussed doing that a few years ago, but I think we were waiting until the hash database to be in a better spot. I'd say it's manageable enough now. Could even rename the cheat entries and continue using those. I think people get freaked out when loading them as is 😆 

I could run our hash list against a P/C DAT and knock out a ton of them pretty quick.

Though my usual method of mass linking hashes (typically for rollouts and to create hub entries) leads to a 403 error now....

Jamiras: I'm thinking more along the lines of:

| Hash | Compatibility | Label |
| ---- | ---- | ---- |
| abcdef1234567890 | Compatible | Game (E).rom |
| 1a2b3c4d5e6f7890 | Incompatible | Game (J).rom |
| 88347abcdea9128a | Untested | Game (U).rom |
| 2abbabbab2babab2 | Incompatible | Game (E) [Infinite lives cheat].rom |

We might even want to go so far as to have a Blocked compatibility that would behave like Incompatible, but not be shown on the linked hashes page.

wescopeland: Updated in latest:

```php
public const Unverified = 'unverified';
public const Unknown = 'unknown';
public const Untested = 'untested';
public const Compatible = 'compatible';
public const Incompatible = 'incompatible';
```

app/Platform/Controllers/GameHashController.php
wescopeland: You're right - I should have actually checked where those links were going. Set to `raw` in latest.

Jamiras (approved):

app/Platform/Enums/HashCompatibility.php
Jamiras: Does there need to be a query/command to initialize all of the existing records to 'unverified'? Or will null naturally map to that later? We can probably address that in whatever PR adds the ability to edit these.

wescopeland: Indeed, I plan on taking care of working with this value in a subsequent PR.

PR #2114, opened 2023-12-18
perf: auth gate historyexamine.php and claimslist.php

wescopeland:
Redirect to /login if unauthenticated.


wescopeland:
@luchaos Do you think this is still worth doing? Happy to close the PR if not.

public/claimlist.php
```diff
@@ -10,7 +10,10 @@
 use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Str;
 
-authenticateFromCookie($user, $permissions, $userDetails);
+if (!authenticateFromCookie($user, $permissions, $userDetails, Permissions::Registered)) {
+    header("Location: /login");
+    exit;
```
Jamiras, line 15: This should probably just be:
```
return redirect(route('login'));
```

There's probably even a way to pass it the current page so it'll come back after login.

wescopeland: Changed to use a `redirect()` call in latest, however I'm not able to seem to get the `intended_url` value to work correctly.

Jamiras: It looks like `redirect()` automatically does that. After pasting the URL into an incognito window, getting redirected to the login page, and entering credentials, it took me back to the pasted URL.

Jamiras (approved):
Not sure if we want to wait for feedback from @luchaos about the desirability of this change.

wescopeland:
Doesn't seem like we can get an answer. Closing.

PR #2120, opened 2023-12-23
chore: add deprecation and authentication comments to dorequest

luchaos:
to be discussed

### DIFF
--- a/public/dorequest.php
+++ b/public/dorequest.php
@@ -137,18 +137,21 @@ switch ($requestType) {
     /*
      * Global, no permissions required
      */
+
+    // TODO deprecate?
     case "allprogress":
         $consoleID = (int) request()->input('c');
         $response['Response'] = GetAllUserProgress($username, $consoleID);
         break;
 
+    // TODO deprecate - next image ID should be provided by the server
     case "badgeiter":
         // Used by RALibretro achievement editor
         $response['FirstBadge'] = 80;
         $response['NextBadge'] = (int) FilenameIterator::getBadgeIterator();
         break;
 
-    // TODO: Deprecate - not used anymore
+    // TODO deprecate - not used anymore
     case "codenotes":
         if (!getCodeNotes($gameID, $codeNotesOut)) {
             return DoRequestError("FAILED!");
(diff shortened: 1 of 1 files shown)


public/dorequest.php
```diff
@@ -137,18 +137,21 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
      * Global, no permissions required
      */
+
+    // TODO deprecate?
     case "allprogress":
         $consoleID = (int) request()->input('c');
         $response['Response'] = GetAllUserProgress($username, $consoleID);
         break;
 
+    // TODO deprecate - next image ID should be provided by the server
```
Jamiras, line 147: This is used by the DLL to determine the maximum value allowed for manually specifying a badge.

We could remove the validation and just allow the user to enter any number, but I'm not sure how well the client would handle the 404 response.

public/dorequest.php
```diff
@@ -162,25 +165,30 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
         $response['CodeNotes'] = getCodeNotesData($gameID);
         $response['GameID'] = $gameID;
         break;
+
     case "gameid":
         $md5 = request()->input('m') ?? '';
         $response['GameID'] = getGameIDFromMD5($md5);
         break;
 
+    // TODO require authentication?
```
Jamiras, line 179: This is [exposed through rcheevos](https://github.com/RetroAchievements/rcheevos/wiki/rc_api_init_fetch_games_list_request) without authentication. Clients would have to upgrade, though I'm not sure if any clients besides the DLL use it.

public/dorequest.php
```diff
@@ -162,25 +165,30 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
     case "gameslist":
         $consoleID = (int) request()->input('c', 0);
         $response['Response'] = getGamesListDataNamesOnly($consoleID);
         break;
 
+    // TODO require authentication?
```
Jamiras, line 185: I'm not aware of any official uses for this. If it's being used by unofficial clients, we should provide a public API for this instead.

public/dorequest.php
```diff
@@ -162,25 +165,30 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
     case "officialgameslist":
         $consoleID = (int) request()->input('c', 0);
         $response['Response'] = getGamesListDataNamesOnly($consoleID, true);
         break;
 
+    // TODO deprecate?
```
Jamiras, line 191: the functionality that used to rely on this was removed from the DLL in 0.76 (2019)

public/dorequest.php
```diff
@@ -277,6 +285,7 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
         $response['AchievementID'] = $achIDToAward;
         break;
 
+    // TODO deprecate
```
Jamiras, line 288: The DLL still uses this to show friend activity in the clients. It's buried in the overlay. I don't think I've ever used it. We could probably deprecate it.

public/dorequest.php
wescopeland: This is indeed being used by third-party front-ends.

PR #2122, opened 2023-12-24
feat(api): add API_GetUserCompletionProgress endpoint

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/discussions/2081#discussioncomment-7937696, https://github.com/RetroAchievements/RAWeb/discussions/2081#discussioncomment-7821888, https://github.com/RetroAchievements/RAWeb/discussions/2081#discussioncomment-7802277, https://discord.com/channels/310192285306454017/1187330228071190588, and a big chunk of https://github.com/RetroAchievements/RAWeb/discussions/2081#discussioncomment-7794378.

This PR adds a new endpoint: `API_GetUserCompletionProgress`, which returns results identical to what is shown on the Completion Progress page (`/user/{username}/progress`). This includes a game list sorted by last unlock date, number of unlocks per single game in both hardcore and softcore modes, and the user's highest award status for the game (and when it was granted).

The endpoint also supports pagination. Ultimately, this endpoint is far more efficient than using the existing `API_GetUserCompletedGames` endpoint.

A few other things of note:
* Pagination is supported with `o` and `c` parameters. A max of 100 games can be returned at once, similar to the Completion Progress page itself.
* Timestamps are sent to consumers as ISO8601 strings.

```
GET https://retroachievements.org/API/API_GetUserCompletionProgress.php?z=WCopeland&y=...&u=Jamiras
```

```json
{
  "Count": 100,
  "Total": 428,
  "Results": [
    {
	"GameID": 14062,
	"Title": "Combat",
	"ImageIcon": "\/Images\/068385.png",
	"ConsoleID": 25,
	"ConsoleName": "Atari 2600",
	"MaxPossible": 15,
	"NumAwarded": 13,
	"NumAwardedHardcore": 13,
	"MostRecentAwardedDate": "2023-10-26T00:28:24+00:00",
	"HighestAwardKind": null,
	"HighestAwardDate": null
    },
    {
	"GameID": 24767,
	"Title": "Melli's Retro Land",
	"ImageIcon": "\/Images\/076470.png",
	"ConsoleID": 80,
	"ConsoleName": "Uzebox",
	"MaxPossible": 22,
	"NumAwarded": 22,
	"NumAwardedHardcore": 22,
	"MostRecentAwardedDate": "2023-10-24T18:51:34+00:00",
	"HighestAwardKind": "mastered",
	"HighestAwardDate": "2023-10-24T18:51:34+00:00"
    },
    // ...
  ]
}
```


public/API/API_GetUserCompletionProgress.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+/*
+ *  API_GetUserCompletionProgress - gets the entire completion progress (eg: /user/{username}/progress) for a user
+ *                                  similar to `GetUserCompletedGames`, but only includes a single record for each game
+ *                                  and also includes the game's current award level as shown on the "Completion Progress" page.
+ *    u : username
+ *    o : offset - number of entries to skip (default: 0)
+ *    c : count - number of entries to return (default: 100, max: 500)
```
Jamiras, line 9: The feature description says the max is 100.
> Pagination is supported with o and c parameters. A max of 100 games can be returned at once, similar to the Completion Progress page itself.

public/API/API_GetUserCompletionProgress.php
```diff
@@ -0,0 +1,93 @@
...
+        return 1;
+    } elseif (!$dateB) {
+        return -1;
+    } else {
+        return $dateB <=> $dateA;
+    }
+});
+
+// Apply pagination.
+$pagedResults = array_slice($filteredAndJoinedGamesList, $offset, $count);
```
Jamiras, line 70: This probably applies to the beaten game page too, so I'm not expecting any changes here, but why doesn't the `getUserCompletedGamesAndMax` function accept the offset and count to limit the query?

I'm guessing it has something to do with the additional level of sorting being applied a few lines above.

tests/Feature/Api/V1/UserCompletionProgressTest.php
```diff
@@ -0,0 +1,272 @@
...
+    {
+        Carbon::setTestNow(Carbon::now());
+
+        /**
+         * Set up a user with 10 games played:
+         * - 2 unfinished
+         * - 3 beaten
+         * - 1 beaten (softcore)
+         * - 3 completed (one of which is also beaten softcore)
+         * - 4 mastered (two of which are also beaten hardcore)
```
Jamiras, line 50: if a game is completed (softcore) and beaten (hardcore), which is reported?

public/API/API_GetUserCompletionProgress.php
wescopeland: I think some wires got crossed in my brain.

However, I'm wondering if it's okay for the endpoint to have a 500 max limit (the current implementation). Looks like the response size is ~134KB.

public/API/API_GetUserCompletionProgress.php
wescopeland: Indeed, the sort happens after the entire list is constructed (this is also true for the Completion Progress page).

tests/Feature/Api/V1/UserCompletionProgressTest.php
wescopeland: Similar to the Completion Progress page, the Completion (softcore) will be reported as opposed to the Beaten (hardcore) award.

public/API/API_GetUserCompletionProgress.php
Jamiras: Given that we're fetching the entire list before sorting, the limit only applies to the bandwidth, not the query or post-processing. This is probably fine.

Jamiras (approved):

PR #2124, opened 2023-12-25
feat: update api documentation links

wescopeland:
Removes the on-site API docs page the footer currently links to. Instead, it's better if we take the user directly to our API docs home at https://api-docs.retroachievements.org.

This PR also adds another link in the control panel near the user's API key:

Screenshot 2023-12-25 at 4 42 47 PM


app/Platform/RouteServiceProvider.php
```diff
@@ -63,8 +62,6 @@ protected function mapWebRoutes(): void
             // Route::get('create', CreateController::class)->name('create');
             // Route::resource('developers', DeveloperController::class)->only('index');
 
-            Route::get('docs/api', [ApiDocsController::class, 'index'])->name('docs.api');
```
luchaos, line 66: This page was meant for both web api and connect api docs in the future. But I'm not sure we'll ever need it 👍

luchaos (approved):

PR #2126, opened 2023-12-28
feat(api): add an optional `f` (flag) parameter to API_GetGameExtended

wescopeland:
This PR adds a new optional `f` parameter to API_GetGameExtended. By default, this parameter is set to 3 and the behavior of the endpoint does not change. However, if the parameter is set to 5, the game's list of Unofficial achievements is returned instead of the Core achievements.

There is a tool in the ecosystem that needs unofficial achievement data, and currently there's not a great way to get it short of scraping the site.


Jamiras (approved):
While testing this, I noticed that new fields are bleeding through because the API just dumps the object returned by the function. We should address that in a future PR.

* `players_total` is redundant with `NumDistinctPlayers`
* `achievement_set_version_hash` is not used for anything yet, so I'm surprised to see it being returned at all
* `achievements_published` and `points_total` reflect the published data even when queried with `f=5`.

wescopeland:
https://github.com/RetroAchievements/RAWeb/issues/2134

PR #2128, opened 2023-12-30
feat: add Standalone system [xi]

wescopeland:
Adds a new system ID, `102`, for Standalone.

To insert into the prod DB on deploy:

```sql
INSERT INTO Console (ID, Name) VALUES (102, 'Standalone');
```


app/Platform/Commands/AddStandaloneSystem.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $doesStandalonesSystemExist = DB::table('Console')->where('ID', $standalonesConfig['id'])->exists();
+        if (!$doesStandalonesSystemExist) {
+            DB::table('Console')->insert([
+                'ID' => $standalonesConfig['id'],
+                'Name' => $standalonesConfig['name'],
+            ]);
+        }
+    }
+}
```
Jamiras, line 67: Wouldn't it be better to do this with a migration?

Is there going to be another ID other than 102?

config/systems.php
```diff
@@ -578,4 +578,11 @@
         'manufacturer' => null,
         'active' => true,
     ],
+    102 => [
+        'name' => 'Standalones',
+        'name_full' => 'Standalones',
+        'name_short' => 'Indie',
+        'manufacturer' => '',
+        'active' => false,
+    ],
```
Jamiras, line 587: This causes the tooltip to show Standalones (maybe it should be singular)

image

and Indie is shown in the recent games list.

image

Also, I'm not sold on 'Indie' being the short form of Standalones, but I don't have any better ideas at this time.

app/Platform/Commands/AddStandaloneSystem.php
wescopeland: To the best of my knowledge, none of our migrations write rows to the DB. I vaguely recollect a conversation that happened on another PR where we decided not to take that approach, though I can't remember the PR anymore.

As I understand our implementation plan, the new system will only have a single system ID.

Jamiras: You're thinking about the Messages PR. That was a hefty chunk of logic that had to be executed. This is just inserting a single row into the DB. For previous new systems, that's always just been done by hand some time before deploying the code.

wescopeland: Migration added in latest and command removed.

config/systems.php
wescopeland: Changed the short name to "N/A" and set the full name to singular "Standalone".

database/migrations/2024_01_03_000000_add_standalone_system.php
```diff
@@ -0,0 +1,57 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\App;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        if (App::environment('testing')) {
+            return;
+        }
```
wescopeland, line 14: I must confess I'm really not in love with this.

However, there are numerous tests that rely on only a certain number of console IDs existing in the test database.

Part of me still feels like it'd be better if this write operation were not in a migration.

Jamiras: I think we should just craft an insert statement and run it directly against the prod DB.

As long as the seeder populates it for general use, any tests relying on this specific console can add it themselves.

wescopeland:
I think this should be all set.

* I've updated the OP with a crafted SQL statement to add the console to prod.
* I've added a commit co-authored by @oddgoo to include the new Standalone console icon (please give some credit to oddgoo when crafting future release notes).
* I've removed the migration that was previously in the PR.

Jamiras (approved):

Jamiras:
I've run the command on the prod and stage DBs.

PR #2129, opened 2024-01-01
feat(connect): add support for standalones and user event delegation [xi]

wescopeland:
This PR enables standalones support in the Connect API.

## How it works
* A new global parameter, `k`, has been added to _dorequest.php_. This optional parameter expects a username.
* New logic, driven by `k`, has been added to _dorequest.php_ to support "user delegation".
* This new logic allows a caller to delegate an event on behalf of another RetroAchievements user. For example, User X can `ping`, `startsession`, and `awardachievement` on behalf of User Y.

This is controlled via an `$allowsGenericDelegation` array, which limits what functions are allowed to be delegated:

```php
$allowsGenericDelegation = [
    "awardachievement",
    "ping",
    "startsession",
];
if (in_array($requestType, $allowsGenericDelegation)) {
  // ... do some delegation stuff
}
```

* If a function which allows generic delegation is called with a valid `k` parameter, delegation will be attempted.
* For `ping` and `startsession`, delegation is only allowed if the game belongs to the Standalones system ID and the caller has authored some official achievements.
* For `awardachievement`, delegation is only allowed if the game belongs to the Standalones system ID and the caller has authored the target achievement ID.
* `awardachievement` with delegation also requires a `v` parameter, which checks against `unlockValidationHash()`.
* A new function: `awardachievements`, has been added. Like `awardachievement`, this function accepts an `a` query parameter. However, the parameter can be a comma-separated list.
* When `awardachievements` is called, it will attempt to trigger delegated unlocks for all given achievement IDs, following the same logic as `awardachievement`.
* `awardachievements` uses a comma-separated list for the `v` parameter, which ensures all unlock requests validate against `unlockValidationHash()`.
* `awardachievements` returns a `SuccessfulIDs` array in its response, which only contains the IDs that were unlocked as part of the call. Another array, `ExistingIDs`, is returned which can tell a caller which achievements they've tried to sync that are already recognized as unlocked in our system.
* All delegated calls require `POST` instead of `GET`.

## Delegated `startsession`

```
POST http://localhost:64000/dorequest.php?u=User&t=Token&r=startsession&g=GameID&k=DelegatedUsername
```

```json
{
	"Success": true,
	"HardcoreUnlocks": [
		{
			"ID": 141,
			"When": 1591132445
		}
	],
	"ServerNow": 1704076711
}
```

https://github.com/RetroAchievements/RAWeb/assets/3984985/bc6397e3-cfff-41e0-8889-335b1db3518a

## Delegated `ping`

```
POST http://localhost:64000/dorequest.php?u=User&t=Token&r=ping&g=GameID&k=DelegatedUsername
```

```json
{
	"Success": true
}
```

https://github.com/RetroAchievements/RAWeb/assets/3984985/c08a05e5-9d6b-4c65-aec6-d17abd4da171

## Delegated `awardachievement`

```
POST http://localhost:64000/dorequest.php?u=User&t=Token&r=awardachievement&k=DelegatedUsername&h=1&a=AchievementID&v=MD5
```

```json
{
	"Success": true,
	"AchievementsRemaining": 5,
	"Score": 22866,
	"SoftcoreScore": 5,
	"AchievementID": 150
}
```

https://github.com/RetroAchievements/RAWeb/assets/3984985/3a53e5f3-0db1-48f2-8a8d-a53ce94fa0ff

## `awardachievements`

```
POST http://localhost:64000/dorequest.php?u=User&t=Token&r=syncachievements&k=DelegatedUser&h=1&a=AchievementIDsAsCSV&v=MD5
```

```json
{
	"Success": true,
	"Score": 22890,
	"SoftcoreScore": 5,
	"ExistingIDs": [
		141,
		147
	],
	"SuccessfulIDs": [
		142,
		145,
		146
	]
}
```

https://github.com/RetroAchievements/RAWeb/assets/3984985/8188df3e-405c-4a14-a181-36bc72d05565


app/Platform/Models/Achievement.php
```diff
@@ -123,6 +123,13 @@ public function getCanonicalUrlAttribute(): string
         return route('achievement.show', [$this, $this->getSlugAttribute()]);
     }
 
+    public function getCanSendConnectUpdatesForUsers(User|string $user): bool
```
Jamiras, line 126: This method name doesn't feel right - especially on the achievement. Maybe `getCanUnlockBeDelegatedByUser`?

public/dorequest.php
```diff
@@ -114,6 +155,25 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
     return DoRequestError("You do not have permission to do that.", 403, 'access_denied');
 }
 
+/*
+ * It is possible for some calls to be made on behalf of another user.
+ * This generally happens with a "Standalone" integration.
```
Jamiras, line 160: Suggest:
```
* This is only currently supported for "Standalone" integrations
```

public/dorequest.php
```diff
@@ -388,6 +448,56 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
         }
         break;
 
+    case "syncachievements":
```
Jamiras, line 451: This implies the achievement definitions will be synced with some external source. Recommend "awardachievements". "syncunlocks" would also be acceptable, but still implies syncing entire records, not generating new ones.

public/dorequest.php
```diff
@@ -388,6 +448,56 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+        if (!$foundTargetUser) {
+            return DoRequestError("The target user couldn't be found.", 400);
+        }
+
+        $achievementIdsArray = explode(',', $achievementIdsInput);
+        $idsToAttemptAward = array_filter($achievementIdsArray, function ($id) {
+            return filter_var($id, FILTER_VALIDATE_INT) !== false;
+        });
+
+        if (count($idsToAttemptAward) > 800) {
```
Jamiras, line 470: I would hope we don't try to award more than a couple dozen achievements at a time. The initial unlock of all previously earned achievements might have to be batched.

public/dorequest.php
```diff
@@ -114,6 +155,25 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+ * achievement ID values. It doesn't use the generic `delegateUserAction()` function.
+ */
+$allowsGenericDelegation = [
+    "awardachievement",
+    "ping",
+    "startsession",
+];
+if (in_array($requestType, $allowsGenericDelegation)) {
+    try {
+        [$user, $username] = delegateUserAction($user, $delegateTo, $gameID, $achievementID);
```
Jamiras, line 171: Since `delegateUserAction` is only called here, why not just embed it?
```
if (in_array($requestType, $allowsGenericDelegation)) {
    $foundTargetUser = User::firstWhere('User', $targetUsername);
    if (!$foundTargetUser) {
        return DoRequestError("The target user couldn't be found.", 403, 'access_denied');
    }
    ...
```

That would also eliminate the need to raise and catch exceptions.

public/dorequest.php
wescopeland: > The initial unlock of all previously earned achievements might have to be batched.

Can you explain this another way? There will almost certainly be a case where an initial "sync" contains 100+ achievements, and I should ensure how we expect this to be handled is well-documented.

Jamiras: > Can you explain this another way?

For normal use, I would only expect a handful of achievements to unlocked at a time - kill boss, kill boss damageless, kill boss using weapon X, complete level in under Y seconds, etc. Having 800 unlocks at the same time would be very spammy.

I understand that this is meant primarily for backfilling achievements earned before integrating with RetroAchievements, but it could also be used to batch several new unlocks together in a single server call - there could even be desire to do that for non-standalone emulators.

800 unlocks in a single request (even for backfilling) seems excessive. Does that imply that the standalone sets will have more than 800 achievements? That just seems unmanageable. If there needs to be an upper limit, it should be something more in line with the expected use cases.

If 1000 unlocks is a valid backfill use case, then the client should batch that into smaller chunks to fit whatever we determine to be a reasonable constraint. 800 doesn't feel like a reasonable constraint - you might as well leave it unbounded.

app/Platform/Models/Achievement.php
wescopeland: I definitely get what you're feeling here. I've renamed them in both `Game` and `Achievement`... maybe it's a little better now.

`Game::getCanDelegateActivity()`
`Achievement::getCanDelegateUnlocks()`

public/dorequest.php
wescopeland: Done in latest.

public/dorequest.php
wescopeland: Renamed to "awardachievements".

public/dorequest.php
wescopeland: The upper bound has been removed. The new logic makes a query to `player_achievements` to efficiently determine which achievements an unlock should be attempted for.

public/dorequest.php
wescopeland: Good idea -- originally I duplicated the logic in all the cases, then I lifted it up into the function to be a bit more elegant. Direct embedding is another good step beyond that. Done in latest.

public/dorequest.php
```diff
@@ -114,6 +120,52 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+    "ping",
+    "startsession",
+];
+if (
+    in_array($requestType, $allowsGenericDelegation)
+    && $delegateTo !== null
+    && ($gameID || $achievementID)
+) {
+    if (request()->method() !== 'POST') {
+        return DoRequestError('Access denied.', 403, 'access_denied');
```
Jamiras, line 140: I believe this should be 405 (Method Not Allowed). The code string can probably remain `access denied`.

public/dorequest.php
```diff
@@ -114,6 +120,52 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+    }
+
+    $foundDelegateToUser = User::firstWhere('User', $delegateTo);
+    if (!$foundDelegateToUser) {
+        return DoRequestError("The target user couldn't be found.", 403, 'access_denied');
```
Jamiras, line 145: Is 403 the best response for this? Maybe 404/not_found?

Same comment for missing game/achievements below.

public/dorequest.php
```diff
@@ -277,6 +337,78 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+            return DoRequestError("You must specify a target user.", 400);
+        }
+
+        $foundTargetUser = User::firstWhere('User', $delegateTo);
+        if (!$foundTargetUser) {
+            return DoRequestError("The target user couldn't be found.", 400);
+        }
+
+        $achievementIdsArray = explode(',', $achievementIdsInput);
+        $validationHashesArray = explode(',', $validationHashesInput);
```
Jamiras, line 361: Recommend generating one validation hash for the CSV, rather than sending in a parallel CSV of validation hashes.

public/dorequest.php
```diff
@@ -277,6 +337,78 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+
+        $idsToAttemptAward = array_filter($achievementIdsArray, function ($id) {
+            return filter_var($id, FILTER_VALIDATE_INT) !== false;
+        });
+
+        // Fetch the IDs of achievements already awarded to the user.
+        $alreadyAwardedIds = PlayerAchievement::where('user_id', $foundTargetUser->id)
+            ->whereIn('achievement_id', $idsToAttemptAward)
+            ->pluck('achievement_id')
+            ->all();
```
Jamiras, line 375: I'm still uncertain how standalones will handle hardcore vs. softcore, but this logic will not acknowledge an achievement being unlocked in hardcore if it has previously unlocked in softcore.

public/dorequest.php
```diff
@@ -264,7 +316,15 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
          */
         $response = array_merge($response, unlockAchievement($user, $achIDToAward, $hardcore));
 
-        if (Achievement::where('ID', $achIDToAward)->exists()) {
+        $foundAchievement = Achievement::where('ID', $achIDToAward)->first();
+        if ($foundAchievement !== null) {
+            if (
+                $delegateTo
+                && request()->input('v') !== $foundAchievement->unlockValidationHash($foundDelegateToUser, (int) $hardcore)
```
Jamiras, line 323: You may want this to be a case insensitive comparison as the validation hash can contains alphabetic characters and the client hashing algorithm may not generate the same casing the server expects. 

Alternate, we could put the onus on the client to always generate lowercase hashes. That would have to be documented somewhere.

public/dorequest.php
wescopeland: Fixed in latest.

public/dorequest.php
wescopeland: 404 is probably a better fit. Updated in latest.

public/dorequest.php
wescopeland: That makes sense.

This function now accepts a single validation hash, constructed using all the given achievement IDs.

public/dorequest.php
wescopeland: I've updated the code to account for this. It does add a bit more complexity to the `awardachievements` logic.

public/dorequest.php
wescopeland: Agreed - the verification is now case insensitive.

public/dorequest.php
```diff
@@ -257,14 +313,26 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+        if ($foundAchievement !== null) {
+            if (
+                $delegateTo
+                && strcasecmp(
+                    $validationHash,
+                    $foundAchievement->unlockValidationHash($foundDelegateToUser, (int) $hardcore)
+                ) !== 0
+            ) {
+                return DoRequestError('Access denied.', 403, 'access_denied');
+            }
```
Jamiras, line 334: You should move this above line 322. `unlockAchievement` has some side-effects if it thinks the achievement will be unlocked.

public/dorequest.php
```diff
@@ -277,6 +345,104 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
 
+    // This is only currently supported for "Standalone" integrations.
+    case "awardachievements":
+        if (request()->method() !== 'POST') {
+            return DoRequestError('Access denied.', 405, 'access_denied');
+        }
+
+        $achievementIdsInput = request()->input('a', '');
+        $hardcore = (bool) request()->input('h', 0);
+        $validationHash = request()->input('v');
```
Jamiras, line 356: Recommend changing these `request()->input` calls to `request->post()` to ensure the data is in the POST body and not the query parameters.

public/dorequest.php
```diff
@@ -277,6 +345,104 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+                    if (!$hardcore) {
+                        $alreadyAwardedIds[] = $foundPlayerAchievement->achievement_id;
+
+                        return false;
+                    }
+
+                    if (!$foundPlayerAchievement->achievement->getCanDelegateUnlocks($user)) {
+                        return false;
+                    }
+                }
```
Jamiras, line 413: recommend adding a `break` here after finding the matching record

public/dorequest.php
```diff
@@ -277,6 +345,104 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+            // If no PlayerAchievement record exists for this ID, it's always eligible for awarding.
+            return true;
+        });
+
+        $awardableAchievements = Achievement::whereIn('ID', $filteredAchievementIds)
+            ->with('game')
+            ->get()
+            ->filter(function ($achievement) use ($user) {
+                return $achievement->getCanDelegateUnlocks($user);
+            });
```
Jamiras, line 425: `$achievement->getCanDelegateUnlocks($user)` is handled by the `filteredAchievementIds` logic above. This can be removed.

public/dorequest.php
wescopeland: Good point, moved. To retain the same error reporting, I needed to add an `else` block:

```php
        } else {
            $response['Error'] = "Data not found for achievement {$achIDToAward}";
            $response['Success'] = false;
        }
```

public/dorequest.php
wescopeland: Not a problem, changed and test cases revised.

public/dorequest.php
wescopeland: I went a bit further and opted to eliminate the `foreach` entirely, leaning more heavily into collections with `firstWhere()`.

public/dorequest.php
wescopeland: Removed and refactored the filtering logic just a bit.

Jamiras (approved):
There's a conflict to be resolved, but I think everything else looks okay.

wescopeland:
Conflict resolved in latest. I'm not sure how we want to time this one. On one hand, even if merged it's likely our partner won't be ready for a bit. On the other hand, there's likely no harm in this going out in the upcoming release.

Jamiras:
> On one hand, even if merged it's likely our partner won't be ready for a bit. On the other hand, there's likely no harm in this going out in the upcoming release.

If it's merged, it'll be available on the staging site and we can have the partner do their integration testing there.

PR #2130, opened 2024-01-03
Add more detail to game lists on hub pages

Jamiras:
Modifies the hub pages to use a game list similar to the Developer Sets page, including showing the player's progress for sets in the hub. Additionally, adds buttons to quickly add/remove items from the user's Want to Play list.

image

Other hubs are now removed from the list and displayed in a separate category below the games.

Games are no longer grouped such that games with achievements appear before games without achievements, but a filter option is present to only show games with achievements.

Additionally, games may be sorted by the same criteria as available on the Developer Sets page, including grouping by console:

image

There is a bunch of duplicated code between the two pages, but because of slight differences between them, I'm not sure how feasible it is to eliminate that duplication. If you have any brilliant ideas, let me know.


resources/views/platform/components/game/related-games-table.blade.php
```diff
@@ -0,0 +1,187 @@
...
+                <th style='width:10%; cursor: help' class='text-right'
+                    title='The number of points associated to achievements in the set'>Points</th>
+                <th style='width:10%; cursor: help' class='text-right'
+                    title='As estimate of rarity for achievements in the set'>RetroRatio</th>
+                <th style='width:10%; cursor: help' class='text-right'
+                    title='The number of leaderboards in the set'>Leaderboards</th>
+                <th style='width:8%; cursor: help' class='text-right'
+                    title='The number of users who have played the set'>Players</th>
+                <th style='width:8%; cursor: help' class='text-right'
+                    title='The number of open tickets for achievements in the set'>Tickets</th>
```
wescopeland, line 66: IMHO, in the context of a hub, we shouldn't show this column unless the user is a Dev or higher.

resources/views/platform/components/game/related-games-table.blade.php
```diff
@@ -0,0 +1,187 @@
...
+                        title='Whether or not the game is on your want to play list'>Backlog</th>
+                @endif
+            </tr>
+            <?php $count = $achievementCount = $pointCount = $leaderboardCount = $ticketCount = 0; ?>
+            @foreach ($games as $game)
+                @if ($filterOptions['console'] && $game['ConsoleID'] != $console['ID'])
+                    @continue
+                @endif
+                <tr>
+                    <td>
```
wescopeland, line 80: The multiline avatars gives the appearance of the rows being fairly "snug".

```suggestion
                    <td class="py-2">
```

public/gameInfo.php
wescopeland, line 1: The mobile experience is objectively less cluttered on prod:

Prod:
Screenshot 2024-01-04 at 3 12 13 PM

Branch:
Screenshot 2024-01-04 at 3 12 26 PM

I don't think this warrants any action in the PR other than being noted. Ultimately, mobile is going to need an entirely different layout to make sense (similar to Completion Progress).

public/gameInfo.php
```diff
@@ -1363,8 +1364,34 @@ class="mt-1"
...
         if (!$isFullyFeaturedGame) {
             if (!empty($relatedGames)) {
-                RenderGameSort($isFullyFeaturedGame, $flagParam, $officialFlag, $gameID, $sortBy);
-                RenderGameAlts($relatedGames);
+                $controller = new RelatedGamesTableController();
+                $view = $controller(request());
+                echo $view->render();
+
+                // RenderGameSort($isFullyFeaturedGame, $flagParam, $officialFlag, $gameID, $sortBy);
+                // RenderGameAlts($relatedGames);
```
wescopeland, line 1372: nit: Do we still need these lines?

resources/views/platform/components/developer/sets-page.blade.php
```diff
@@ -114,25 +141,54 @@
...
+                            ?>
+                            <div role="progressbar" aria-valuemin="0" aria-valuemax="100"
+                                title="{{ $achievementsUnlocked }} of {{ $game['achievements_published'] }} unlocked"
+                                class="w-full h-1 bg-embed rounded flex">
+                                <div style="width: {{ $hardcoreProgressBarWidth }}%"
+                                        class="bg-[#cc9900] h-full {{ $hardcoreProgressBarWidth > 0 ? 'rounded-l' : '' }}"></div>
+                                <div style="width: {{ $softcoreProgressBarWidth }}%"
+                                        class="bg-[rgb(11,113,193)] h-full {{ $hardcoreProgressBarWidth === 0 ? 'rounded-l' : '' }}"></div>
+                            </div>
+                            </td>
```
wescopeland, line 166: The amount of PHP here suggests this may be worth breaking out as a child component that lives in a separate Blade file.

resources/views/platform/components/game/related-games-meta-panel/filters.blade.php
wescopeland, line 1: nit: There is a casing difference between the options: "console" vs "Achievements".
If I had to choose, I'd say the word "Achievements" should be lowercased instead of "Console" being uppercased.

resources/views/platform/components/game/related-games-meta-panel/sort-order.blade.php
wescopeland, line 1: Do you think it may be worth the column headers also being clickable to initiate these sorts? I found myself naturally reaching for them rather than the "Sort by" select.

resources/views/platform/components/game/related-games-table.blade.php
```diff
@@ -0,0 +1,187 @@
...
+            <div><table class='table-highlight mb-4'><tbody>
+
+            <tr>
+                <th style='width:28%'>Title</th>
+                <th style='width:12%; cursor: help' class='text-right'
+                    title='The number of achievements in the set'>Achievements</th>
+                <th style='width:10%; cursor: help' class='text-right'
+                    title='The number of points associated to achievements in the set'>Points</th>
+                <th style='width:10%; cursor: help' class='text-right'
+                    title='As estimate of rarity for achievements in the set'>RetroRatio</th>
```
wescopeland, line 60:
```suggestion
                    title='An estimate of rarity for achievements in the set'>RetroRatio</th>
```

There is a minor typo here. I'm wondering if we should also remark that the higher the number is, the rarer unlocks for the set are (not sure what a great wording for this might be).

resources/views/platform/components/game/related-games-table.blade.php
```diff
@@ -0,0 +1,187 @@
...
+                            <div role="progressbar" aria-valuemin="0" aria-valuemax="100"
+                                title="{{ $achievementsUnlocked }} of {{ $game['achievements_published'] }} unlocked"
+                                class="w-full h-1 bg-embed rounded flex">
+                                <div style="width: {{ $hardcoreProgressBarWidth }}%"
+                                        class="bg-[#cc9900] h-full {{ $hardcoreProgressBarWidth > 0 ? 'rounded-l' : '' }}"></div>
+                                <div style="width: {{ $softcoreProgressBarWidth }}%"
+                                        class="bg-[rgb(11,113,193)] h-full {{ $hardcoreProgressBarWidth === 0 ? 'rounded-l' : '' }}"></div>
+                            </div>
+                            </td>
+                        @endif
```
wescopeland, line 149: This seems like it should be a reusable component.

resources/views/platform/components/game/related-games-table.blade.php
wescopeland, line 1: Take a look at `/game/3273`.

Screenshot 2024-01-04 at 3 24 58 PM

If every set in a hub has 0 / `null` of something, should we consider omitting that column entirely?

wescopeland (commented):
Nice work. This first round is just me poking around with the UI/UX. I haven't spent too much time looking into the controller code yet.

resources/views/platform/components/game/related-games-table.blade.php
Jamiras: I think it's more important that the display is consistent across pages. We could maybe blank out the 0 values explicitly, like it done when the set has no achievements, but I think in this case, a set with 9 achievements worth 0 points is worth saying it's worth 0 points.

Instead, the logic should be tied to the fact that the games in question belong to the Events console, which should never have points. I've extracted events into a separate "Related Events" group similar to "Related Hubs", as events won't have points and you shouldn't be able to "Want to Play" events.

resources/views/platform/components/game/related-games-table.blade.php
Jamiras: Done

resources/views/platform/components/game/related-games-table.blade.php
Jamiras: Made the progress bar a component (inner else). the outer two levels of ifs stay here, along with some logic for extracting the values from $userProgress.

resources/views/platform/components/game/related-games-meta-panel/sort-order.blade.php
Jamiras: Having the dropdown allows the user to sort either direction with one click. The column headers require sometimes require two clicks. One for the default sort order, then again to get the reversed sort order. Avoiding the extra click might be desirable.

resources/views/platform/components/game/related-games-table.blade.php
wescopeland: Doesn't that page follow similar logic as I'm proposing here?

```php
$showTickets = (isset($user) && $permissions >= Permissions::Developer);
```

When I'm logged out, I don't see the Tickets column on a page like https://retroachievements.org/gameList.php?c=4.

resources/views/platform/components/game/related-games-meta-panel/sort-order.blade.php
wescopeland: I agree - there are merits to both approaches. Maybe it makes sense for both to be a feature.

resources/views/platform/components/game/related-games-table.blade.php
wescopeland: > I think it's more important that the display is consistent across pages. We could maybe blank out the 0 values explicitly, like it done when the set has no achievements, but I think in this case, a set with 9 achievements worth 0 points is worth saying it's worth 0 points.

You're right, this makes sense.

> I've extracted events into a separate "Related Events" group similar to "Related Hubs", as events won't have points and you shouldn't be able to "Want to Play" events.

Cool, I think that's a great solution.

resources/views/platform/components/game/related-games-meta-panel/sort-order.blade.php
Jamiras: The filter control also eliminates some sorts that have little value - like "Fewest Achievements", "Lowest RetroRatio", or "Lowest Progress".

app/Platform/Controllers/DeveloperSetsController.php
```diff
@@ -6,16 +6,12 @@
...
-use App\Platform\Models\Game;
-use App\Platform\Models\PlayerGame;
-use App\Platform\Models\System;
 use App\Site\Models\User;
 use Illuminate\Contracts\View\View;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\DB;
 
-class DeveloperSetsController extends Controller
+class DeveloperSetsController extends GameListControllerBase
```
wescopeland, line 14: Class extension is a classical OOP design pattern. However, given the context in which the pattern is used here, I suggest refactoring to use dependency injection (DI) instead of class extension/inheritance. Here are my reasons:

1. **DI will reduce tight coupling.**
`extends` is creating a tight coupling between `GameListControllerBase` and its subclasses. This tight coupling can lead to unintended consequences. Changes in `GameListControllerBase` could inadvertently affect all inheriting classes. It also becomes more challenging to modify or replace parts of the code in this inheritance tree independently, leading to a less modular system. In contrast, DI promotes loose coupling. We can more easily make changes without impacting classes that depend on what's injected.

2. **We can avoid complex class hierarchies.**
While the current first-party inheritance tree is only one layer deep, this can set a precedent for further nested inheritance, which can complicate the codebase over time, resulting in less modularity and more obscurity over the origin of methods and control flow in a class. DI, on the other hand, simplifies relationships between classes. The dependencies of the class are clearly defined and managed externally, which maintains clarity and simplicity in the class structure. This makes the code not only easier to understand and maintain in its current state, but also increases the chances it remains manageable as the project grows and evolves.

3. **We can make testing easier.** With class inheritance, swapping implementations or mocking dependencies for testing can be very challenging due to the hard-coded nature of the class's dependencies. DI allows for easier mocking or substitution of the dependencies, significantly simplifying the testing process.

I believe there are indeed merits to extension in certain scenarios (eg: `extends Controller`), however, in this specific case, I believe the advantages of DI outweigh those merits.

luchaos: agreed, Laravel allows to favor "composition over inheritance" with its DI container capabilities. Except for rare cases (Controller, BaseModel, BasePivot) I try to not extend wherever possible for the very reasons laid out above.

Jamiras: What should this dependency be called and where should it live? I was just trying to eliminate some duplicate code:
* `GameListControllerBase::getUserProgress` - returns a mapping of game_ids to achievement unlock counts (hardcore/softcore) for the logged-in user.
* `GameListControllerBase::getGameList` - returns an array of game data for display in the results table from a list of game IDs generated by the subclass.
* `GameListControllerBase::mergeWantToPlay` - injects a boolean column into the results table data for each game indicating whether or not it's already on the user's want to play list
* `GameListControllerBase::sortGameList` - sorts the game list based on one of the selectable filter types.

The subclass is then only responsible for building a list of gameIDs and calling these helper functions to turn that into something that can be displayed in the view.

Which brings me to my next question (which was also mentioned at in the top-most post of this PR):
> There is a bunch of duplicated code between the two pages, but because of slight differences between them, I'm not sure how feasible it is to eliminate that duplication. If you have any brilliant ideas, let me know.

The `related-games-meta-panel` has four files that are basically identical to the `sets-meta-panel`. The only real difference is changing the "Sole developer" checkbox to "Only with achievements". How to I minimize duplicating that? Similarly, `related-games-table.blade.php` is 90% the same as `sets-page.blade.php`. The biggest differences being that the sets page can show "X of Y" values (dev authored 3 of 16 achievements) instead of just "Y" values, and the column header tooltips were modified slightly to reflect context ("The number of achievements created by [user] in the set" vs. "The number of achievements in the set").

In my mind, all of this would be in a singular `GameListController.php`/`game-list-table.blade.php` with a lot of toggle-able features and multiple entry points for routing. But it seems like it's more desirable to have everything constructed from smaller bits instead of have one large thing that's highly customizable. I'm having trouble doing that without large chunks of copy/pasted code.

wescopeland: Your goal of eliminating duplicate code is good. All that's in contention is the design pattern employed to eliminate it.

I've created a demo commit in my fork at https://github.com/wescopeland/RAWeb/commit/f192f7fd78b8cd100635be5e2459122495bf88dc.

In this commit:
* A new service, `GameListService`, has been created in _app/Platform/Services_.
* `RelatedGamesTableController` now depends on `GameListService` through injection.
* `GameListControllerBase::getUserProgress` has been migrated to `GameListService::getUserProgressForGameIds`.

As for the meta panels, you are right that we do have plenty of these things floating around. My personal opinion on these is even though their UIs are similar, their requirements and implementations could vary wildly depending on their context. It seems like it may be fitting for us to create a common "meta surface", which accepts some kind of slot to render filter/sorting controls. We can genericize those controls too so we don't need to keep reinventing the wheel, while also allowing for enough flexibility to integrate a custom control. Those filter/sorting options can be driven by the host component that is mounting the meta surface (which is not too dissimilar from DI / inversion of control).

Jamiras: Thanks for the example. That helps a lot.

Jamiras: I've made the suggested changes.

The base class has been extracted to `GameListService`. I've chosen to store some state in the service instance to simplify several calls. As such, I'm not sure it should be a Service anymore - it's more of a Builder, but I don't know how that fits into our framework.

I've also updated the meta-panel and list to be dynamically generated from data provided by the Controller.

wescopeland:
I'm seeing what I think are a few bugs with the progress bars.

First, I went to the hub at http://localhost:64000/game/3881. I noticed the top game only has 7 achievements. I unlocked 4 of them for myself in hardcore mode. The progress bar is taking on the hardcore color.

When I `@dump()` the softcore progress and hardcore progress variables, it looks like they're both equal to 4. Hardcore progress is shown even though I have none:

Screenshot 2024-01-07 at 5 00 00 PM

Next, I unlocked one hardcore achievement for the set. I'm still only seeing one color, and the `title` for the progress bar is only reporting one mode of progress.

When a progress bar is 100% full, the right side is missing a border radius (`rounded-r` class name):

Screenshot 2024-01-07 at 4 52 38 PM

Jamiras:
> When a progress bar is 100% full, the right side is missing a border radius (rounded-r class name)

I'm pretty sure that logic was copied from here:
https://github.com/RetroAchievements/RAWeb/blob/master/resources/views/platform/components/game/achievements-list/list-item-global-progress.blade.php#L58-L61

I'll add logic to address the 100% case.

Jamiras:
> When I @dump() the softcore progress and hardcore progress variables, it looks like they're both equal to 4.

copy/paste error. I've corrected it.

I've also updated the progress-bar component to round both sides of a 100% bar

app/Platform/Services/GameListService.php
```diff
@@ -0,0 +1,519 @@
...
+            $this->consoles = $this->consoles->filter(function ($console) {
+                foreach ($this->games as $game) {
+                    if ($game['ConsoleID'] == $console['ID']) {
+                        return true;
+                    }
+                }
+
+                return false;
+            });
+        }
```
wescopeland, line 135: For larger datasets, collection methods will be more efficient:

```php
if ($countAfter < $countBefore) {
    $allConsoleIds = collect($this->games)->pluck('ConsoleID')->unique();

    $this->consoles = $this->consoles->filter(function ($console) use ($allConsoleIds) {
        return $allConsoleIds->contains($console->ID);
    });
}
```

app/Platform/Services/GameListService.php
```diff
@@ -0,0 +1,519 @@
...
+    private function renderNumberOrBlank(array $game, string $field): void
+    {
+        if ($game[$field] == 0) {
+            echo '<td></td>';
+        } else {
+            echo '<td class="text-right">';
+            echo localized_number($game[$field]);
+            echo '</td>';
+        }
+    }
```
wescopeland, line 301: These methods are highly functional, but also mix the presentational logic with the service logic.

The service should return data and let the Blade renderer handle presentation. For example:

```php
private function getTitleColumn(array $filterOptions): array
{
    return [
        'header' => 'Title',
        'render' => function ($game) use ($filterOptions) {
            return view('components.title-column', ['game' => $game, 'filterOptions' => $filterOptions])->render();
        }
    ];
}
```

IMO, each of these should request some Blade component to be rendered, and that component should handle its view concerns such as DOM template structure and what classnames to apply. Otherwise we are back to our legacy pattern of `echo`ing DOM.

app/Platform/Controllers/DeveloperSetsController.php
wescopeland: I think it's much better after this refactor.

Indeed, several of the methods in the service have functional impurities, but I think what you've done with the latest changes is much more maintainable for the long-term.

app/Platform/Services/GameListService.php
wescopeland: For cases where we're rendering just a single number or blank space, these existing methods are probably OK. It's likely just the `getBacklogColumn()` content that should be migrated to a Blade component. If the other columns become more complex at a later point in time, maybe that's the appropriate time for them to be migrated rather than now.

Jamiras: I've moved the blacklog column into a blade component and left the simple text fields as echo statements.

wescopeland (approved):
Great work, I'm happy with where we've landed here. Thank you for working earnestly through some of my feedback, I hope it wasn't too much of a bother.

I have two additional thoughts, but they don't bar approval:
* It is likely desirable to add a filter for award status.
* It may be desirable to add a bottom row for totals to these tables.

Issue #2131, opened 2024-01-04
Untracked users are appearing on the achievementInfo.php unlocks list


wescopeland:
**To Reproduce**
1. Go to https://retroachievements.org/achievement/79442
2. Observe that untracked users appear in the unlocks list.

**Expected behavior**
Untracked users should not be shown.

PR #2132, opened 2024-01-06
fix: don't return users for an achievement if they're untracked

MSGoodman:
Addresses #2131. Simply filters those users out in `player-achievement.php`. 

Seems a little paltry. Feels like it should have a test? Any input on the best way to do this? I don't see any tests for anything in `Helpers` right away, and am unsure of best practices here, but in general am more than happy to improve coverage.

| **Before** | **After** | 
|---------------|--------------|
| image | image |


wescopeland:
Looks like `getAchievementUnlocksData()` has three potential callers, two of which are API endpoints.

I'm almost wondering if, for defensive purposes, we should add another arg to the function that defaults to true, eg: `$areUntrackedUsersIncluded`, and for _achievementInfo.php_ in particular that arg is set to `false`.

Unfortunately, _achievementInfo.php_ is another one of those legacy pages floating around (this will be a recurring theme), this unless the functionality is ported completely over to a Controller or a Blade component, testing it is going to be quite difficult.

Helpers are in the process of being phased out - they're a relic from a time before the app used Laravel (or any framework at all). The helpers used to make raw non-ORM SQL calls to the DB.

MSGoodman:
> Looks like `getAchievementUnlocksData()` has three potential callers, two of which are API endpoints.
> 
> I'm almost wondering if, for defensive purposes, we should add another arg to the function that defaults to true, eg: `$areUntrackedUsersIncluded`, and for _achievementInfo.php_ in particular that arg is set to `false`.
> 
> Unfortunately, _achievementInfo.php_ is another one of those legacy pages floating around (this will be a recurring theme). This means unless the functionality is ported completely over to a Controller or a Blade component, testing it is going to be quite difficult to the point I'd wager it provides negative value.
> 
> Helpers are in the process of being phased out - they're a relic from a time before the app used Laravel (or any framework at all). The helpers used to make raw non-ORM SQL calls to the DB.

I'd be more than happy to update that method to accept an additional `areUntrackedUsersIncluded`. But do we _want_ to ever return untracked users? Admittedly I don't know the purpose of being untracked, but I imagine if I want to not appear on the site I probably also don't want to appear in API responses?

wescopeland:
> do we want to ever return untracked users?

Normally, I'd say no. However, because this function is called by API endpoints which are being used by third-parties, I'm hesitant to make any kind of behavior/logic change to underlying data they may be relying upon or using.

MSGoodman:
> > do we want to ever return untracked users?
> 
> Normally, I'd say no. However, because this function is called by API endpoints which are being used by third-parties, I'm hesitant to make any kind of behavior/logic change to underlying data they may be relying upon or using.

Makes sense! I've changed it to only do this filter in this spot and not from the API.

wescopeland:
Functionally this implementation is elegant and works.

I did some last-minute testing and noted a performance issue, though.

**`master`**
Screenshot 2024-01-06 at 5 31 26 PM

**Branch**
Screenshot 2024-01-06 at 5 31 17 PM

It seems filtering by Untracked users adds a significant amount of time to the query response if it's an achievement with tons of unlocks, to the amount of a 20x-40x performance degradation 😔 

This may require an alternative solution. I wonder if in _achievementInfo.php_ itself, in the loop responsible for iterating the users, if we can just not render users who are Untracked. This may be worth exploring and is likely the way to go.

MSGoodman:
> Functionally this implementation is elegant and works.
> 
> I did some last-minute testing and noted a performance issue, though.
> 
> **`master`** Screenshot 2024-01-06 at 5 39 16 PM
> 
> **Branch** Screenshot 2024-01-06 at 5 31 17 PM
> 
> It seems filtering by Untracked users adds a significant amount of time to the query response if it's an achievement with tons of unlocks, to the amount of a 20x-40x performance degradation 😔
> 
> This may require an alternative solution. I wonder if in _achievementInfo.php_ itself, in the loop responsible for iterating the users, if we can just not render users who are Untracked. This may be worth exploring and is likely the way to go.

Wow, good catch, thank you!

Unfortunately, after playing with this for a minute, I'm not sure this `achievmentInfo.php` solution will work. I created 60 achievement unlocks for an achievement, and had all but one of those users marked as untracked. With this solution, what I ended up with from `getAchievementUnlocksData` was a collection of records I completely filtered out, because the most recent 50 were all untracked:

image

Without this filter being on that database level, we end up with this problem. 

I'm no expert on MariaDB, but could we take advantage of an index or indexed view or something here? I'm open to any other suggestions! Or perhaps the edge case I mentioned above isn't a big deal because we don't expect _that_ many users to be untracked and it's not a huge deal if we're not always showing _exactly_ the 50 most recent unlocks.

wescopeland:
I have another idea:

We can proceed with letting `$unlocks` be initialized as-is (no changes from `master`).

Immediately afterwards, we pluck all usernames and query the DB for a list of those users which are not untracked. This is a much less intensive query than doing the Untracked filtering in `getAchievementUnlocksData()` itself.

Then, we can filter `$unlocks` by the response. Something like this:

```php
$unlocks = getAchievementUnlocksData(
    $achievementID,
    $user,
    $numWinners,
    $numWinnersHardcore,
    $numPossibleWinners,
    $parentGameID,
    0,
    50,
    false
);

$trackedUnlocksUsers = User::whereIn('User', $unlocks->pluck('User')->unique())
    ->where('Untracked', false)
    ->pluck('User');

$unlocks = $unlocks->filter(function ($unlock) use ($trackedUnlocksUsers) {
    return $trackedUnlocksUsers->contains($unlock['User']);
});
```

This seems to add less than 1ms of processing to the page.

MSGoodman:
> I have another idea: ...

I've implemented this idea in my newest commit, thank you. It definitely works! 

I think the concerns in [my message above](https://github.com/RetroAchievements/RAWeb/pull/2132#issuecomment-1879903521), for the other solution, still apply here: we only grab 50, and then filter out of _that_, so we won't have a consistent count on this page doing it this way (and _theoretically_ could have an empty list if the 50 most recent unlockers were all untracked). If that's not a concern then I think this is fine!

wescopeland:
> we only grab 50, and then filter out of that, so we won't have a consistent count on this page doing it this way (and theoretically could have an empty list if the 50 most recent unlockers were all untracked). If that's not a concern then I think this is fine!

No concerns on this from me, I think this should be perfectly fine.

wescopeland:
Requesting a review from @Jamiras. I feel my approval is probably a bit biased on its own.

MSGoodman:
> Requesting a review from @Jamiras. I feel my approval is probably a bit biased on its own as I was fairly hands-on with the implementation.

Yeah to that end, you basically wrote this code, if there's an easy way to add you as the author in here please let me know / do so.

Jamiras (approved):
The end results is less than 50 rows being displayed, but I doubt anyone would really notice.

Issue #1909, opened 2023-10-09
Set request limit can be exceeded


cdxemu:
**Describe the bug**
It is possible to exceed the set request limit (even accidentally) by opening multiple game pages when below the limit and requesting all of those sets without refreshing those pages.

This is because while the front-end hides the "Request Set" button when the game page is loaded (if the user was at their limit), it does not appear to check again upon activation of the button.

**To Reproduce**
1. Have one available set request remaining.
2. Open two game pages in separate tabs (games with no achievements that you have not already requested)
3. Navigate to both tabs and click the "Request Set" button on both game pages.
4. See the "User Requests Remaining: -1" on the 2nd game's page and the "[X+1] of [X] Requests Made" on your user page and Set Request List.

**Expected behavior**
Set requests over the user's limit should be blocked.

**Screenshots**

![game_page](https://github.com/RetroAchievements/RAWeb/assets/15841485/0c838c3c-09df-4a10-9c4f-7bba487c93dc)
![set_request_list](https://github.com/RetroAchievements/RAWeb/assets/15841485/4e9689d9-c163-4d0c-ab06-dfa8667c1da5)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 118.0.1

PR #2133, opened 2024-01-07
fix: stop user from requesting sets when they've hit their max

MSGoodman:
Addresses #1909. 

After manually testing I realized that the count derived from `scopeWithoutAchievements` was off, due to what I believe to be a tiny mistake in the query. I verified this by adding a third set request in the automated test and confirming it failed. Then I made the tiny change you see here, which made it pass again.

Please let me know if there's anything else I ought to do here!

As a note, the whole [local scope](https://laravel.com/docs/10.x/eloquent#local-scopes) thing is pretty weird. Do y'all have any particular tools/plugins in your IDEs you use to help with these? PhpStorm has no clue that it's a thing:
image


Jamiras (approved):
Good find. When a set has no achievements, `Achievements.GameID` will be null, so grouping by `Achievements.GameID` will cause all games with no achievements to be consolidated into a single row.

wescopeland (approved):

PR #2135, opened 2024-01-07
feat(home): make some slight tweaks to Players Online chart presentation

wescopeland:
This PR slightly modifies a few visual elements in the Players Online chart on the site homepage.

Screenshot 2024-01-07 at 12 39 20 PM

* The chart is now slightly wider. Currently in prod there's a noticeable amount of right-hand margin in the component.
* Numbers are now localized and include number separators.
* The tooltip no longer redundantly uses the word "online" twice in the same row.


resources/views/community/components/user/online-count-chart.blade.php
```diff
@@ -78,7 +80,7 @@ function drawCharts() {
             },
             legend: { position: 'none' },
             chartArea: {
-                'width': '85%',
+                'width': '93%',
```
Jamiras, line 83: The left margin is no longer wide enough.
image

```
for value in {200..260}; do var=$(($value*10)); echo $var; done > storage/logs/playersonline.log
```

wescopeland: Thanks for the command, that worked great. Updated in latest.

Screenshot 2024-01-08 at 6 14 03 PM

Screenshot 2024-01-08 at 6 14 12 PM

Jamiras: Might want a bit more padding on the right side too:
image

Jamiras: And if I make my browser window as small as possible:
image

wescopeland: Thanks, I've tweaked the left and right padding by just a few % which seems to remediate the visual artifacting across all breakpoints on Chrome.

In terms of small window size, IMO it's probably not worth optimizing for anything smaller than iPhone SE (this is also the resolution I test everything at XS):
Screenshot 2024-01-08 at 7 42 02 PM

Jamiras: I'm still seeing some awkward truncations at some sizes:
image

wescopeland: I think is caused by the X-axis deciding to render more labels, but only sometimes. I've updated the chart config to keep the X-axis consistent across window widths.

Jamiras (approved):

PR #2137, opened 2024-01-07
fix: Replace static image icons with fontawesome SVG icons

### LINKED ISSUE #2111: Replace images in public/assets/images/icons with blade icons
See https://github.com/RetroAchievements/RAWeb/tree/master/public/assets/images/icon

Some of them aren't even used anymore and can be deleted. All the other ones should be replaced by fontawesome solid blade ui icons: https://blade-ui-kit.com/blade-icons?set=9

grablair:
Replaced all references in code to `icon/.*\.(png|gif)`, except one reference to the loading spinner `.gif` file, which is in a javascript file. I didn't see any fontawesome JS dependency and I didn't want to add one unless I get feedback saying so.

Tested locally by viewing all modified pages on a dev stack.

Resolves #2111

<details>
<summary>Website Images</summary>

Screenshot 2024-01-07 at 13 54 57
Screenshot 2024-01-07 at 13 55 01
Screenshot 2024-01-07 at 13 55 18
Screenshot 2024-01-07 at 13 55 25
Screenshot 2024-01-07 at 13 55 31
Screenshot 2024-01-07 at 13 55 43
Screenshot 2024-01-07 at 13 55 55

</details>


app/Helpers/render/comment.php
wescopeland, line 1: Please import `Illuminate\Support\Facade\Blade` into this file.

app/Helpers/render/comment.php
```diff
@@ -113,7 +113,7 @@ function RenderArticleComment(
     $deleteIcon = '';
 
     if ($user && $user === $localUser || $allowDelete) {
-        $img = "<img src='" . asset('assets/images/icon/cross.png') . "' width='16' height='16' alt='delete comment'/>";
+        $img = Blade::render('<x-fas-xmark class="h-5 w-5" style="color: red" />');
```
wescopeland, line 116: There is no reason to use an inline style here. Please use a Tailwind classname.

app/Helpers/render/comment.php
wescopeland, line 1: This change breaks accessibility of both buttons by removing their `title` attributes.

public/controlpanel.php
```diff
@@ -43,6 +43,18 @@ function RenderUserPref(
...
+    $(iconRoot).find('.loadingicon-spinner').show();
+    $(iconRoot).find('.loadingicon-done').hide();
+    $(iconRoot).fadeTo(100, 1.0);
+}
+
+function ShowDoneIcon(iconRoot) {
+    $(iconRoot).find('.loadingicon-spinner').hide();
+    $(iconRoot).find('.loadingicon-done').show();
+    $(iconRoot).delay(750).fadeTo('slow', 0.0);
+}
```
wescopeland, line 56: We are slowly phasing out jQuery.

This should be achievable with a combination of vanilla JS and TailwindCSS classnames.

public/controlpanel.php
```diff
@@ -218,7 +230,10 @@ function confirmEmailChange(event) {
                 </tr>
                 </tbody>
             </table>
-            <img id='loadingicon-1' style='opacity: 0; float: right;' src='<?= asset('assets/images/icon/loading.gif') ?>' width='16' height='16' alt='loading icon'/>
+            <span id="loadingicon-1" style='opacity: 0; float: right;'>
```
wescopeland, line 233: While changing this to a span, we should also clean up these inline styles.

public/controlpanel.php
```diff
@@ -536,9 +571,10 @@ function ResetProgressForSelection() {
             </div>
             <div style="margin-bottom: 10px">
                 <input type="file" name="file" id="uploadimagefile" onchange="return UploadNewAvatar();">
-                <img id="loadingiconavatar" style="opacity: 0; float: right;"
-                     src="<?= asset('assets/images/icon/loading.gif') ?>"
-                     width="16" height="16" alt="loading">
+                <span id="loadingiconavatar" style='opacity: 0; float: right;'>
```
wescopeland, line 574: We should be able to remove these inline styles.

public/controlpanel.php
wescopeland, line 1: Please import `Illuminate\Support\Facade\Blade` into this file.

public/createtopic.php
wescopeland, line 1: Please import `Illuminate\Support\Facade\Blade` into this file.

public/rss.php
wescopeland, line 1: Please import `Illuminate\Support\Facade\Blade` into this file.

public/submitnews.php
wescopeland, line 1: Please import `Illuminate\Support\Facade\Blade` into this file.

public/submitnews.php
```diff
@@ -79,7 +79,11 @@
                         <img src="<?= $newsImage ?>" width="470" alt="News header image preview">
                     </div>
                     <input type="file" name="file" id="uploadimagefile" onchange="return UploadImage();">
-                    <img id="loadingicon" style="opacity: 0;" alt="loading icon" src="<?= asset('assets/images/icon/loading.gif') ?>">
+                    <?= Blade::render('<x-fas-spinner class="icon-spin-fast" :id="$id" :alt="$alt" :style="$style" />', [
+                        'id' => 'loadingicon',
+                        'alt' => 'loading icon',
+                        'style' => 'opacity: 0',
```
wescopeland, line 85: `opacity-0` ?

public/viewtopic.php
```diff
@@ -239,14 +239,18 @@ class="w-full mb-2"
             >$defaultMessage</textarea>
         HTML;
 
-        $loadingIconSrc = asset('assets/images/icon/loading.gif');
+        $loadingIcon = Blade::render('<x-fas-spinner class="icon-spin-fast" :id="$id" :alt="$alt" :style="$style" />', [
+            'id' => 'preview-loading-icon',
+            'alt' => 'Loading...',
+            'style' => 'opacity: 0',
```
wescopeland, line 245: Should be able to handle this with a Tailwind classname.

resources/css/icon.css
wescopeland, line 1: None of this should be necessary. Tailwind ships out of the box with a spin animation, and we shouldn't need an explicit classname to set something to the color green.

wescopeland (changes requested):

grablair:
Thanks for all the feedback @wescopeland! This work is far away from my area of expertise (distributed systems) so I greatly appreciate the thorough feedback and your patience as I ramp up. 

I'll go over your review, dig into the references you've supplied, and fix the issues you've pointed out.

wescopeland:
No problem, thanks for contributing your first PR. Please let me know if any of my feedback is ambiguous, confusing, or its contents spurs further questions or ideas.

app/Helpers/render/comment.php
wescopeland: Yep, apologies for the typo.

docs/CONTRIBUTING.md
```diff
@@ -37,6 +37,8 @@ Before submitting a pull request:
 
 - **Send coherent history** - Make sure each individual commit in your pull request is meaningful. If you had to make multiple intermediate commits while developing, please [squash them](https://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages) before submitting.
 
+- **Use [Tailwind CSS](https://tailwindcss.com/docs) for styling whenever possible** - We try to avoid inline CSS or custom styling wherever possible. The Tailwind CSS framework covers most cases.
```
grablair, line 40: Feel free to suggest alternative phrasing or tell me to nix this. I just know it would have been helpful for me as I looked into contributing. :)

app/Helpers/render/comment.php
```diff
@@ -113,7 +114,7 @@ function RenderArticleComment(
     $deleteIcon = '';
 
     if ($user && $user === $localUser || $allowDelete) {
-        $img = "<img src='" . asset('assets/images/icon/cross.png') . "' width='16' height='16' alt='delete comment'/>";
+        $img = Blade::render('<x-fas-xmark class="text-red-700 h-5 w-5" alt="delete comment"/>');
```
wescopeland, line 117: nit: Recommend bumping up to `red-600` for improved contrast across color light/dark/black color schemes.

wescopeland, line 117: Apologies. It looks like Lighthouse is still complaining:

Screenshot 2024-01-08 at 5 35 56 PM

It seems like it'll only be happy if we remove the `alt` from the icon itself and instead use a `title` attribute on the `<a />` tag.

Note, I believe this a11y check has been failing this entire time. I'm only raising this because IMO it's critical to clean up while we're giving these pieces of the UX some love.

app/Helpers/render/comment.php
```diff
@@ -163,7 +164,7 @@ function RenderCommentInputRow(string $user, int $articleTypeId, int $articleId)
 {
     sanitize_outputs($user, $formStr);
     $commentId = "art_{$articleTypeId}_{$articleId}";
-    $submitImageUrl = asset('assets/images/icon/submit.png');
+    $submitImage = Blade::render('<x-fas-square-plus class="w-5 h-5" alt="Submit"/>');
```
wescopeland, line 167: Similarly, Lighthouse is unhappy here:

Screenshot 2024-01-08 at 5 37 53 PM

I think for this to pass the a11y audit, we'll need to remove the `alt` attribute from the icon and instead use `aria-label` and `title` on the `<button />` tag (both properties should be on the button tag and have the same value).

Note, I believe this a11y check has been failing this entire time. I'm only raising this because IMO it's critical to clean up while we're giving these pieces of the UX some love.

public/controlpanel.php
wescopeland, line 1: I think this looks good, it's just hard to see. Can we bump up the size of the spinner and check icons by about 10-15%?

Screenshot 2024-01-08 at 5 41 43 PM

public/createtopic.php
```diff
@@ -63,7 +64,7 @@ class="w-full"
     <?php
     echo "</td></tr>";
 
-    $loadingIconSrc = asset('assets/images/icon/loading.gif');
+    $loadingIcon = Blade::render('<x-fas-spinner id="preview-loading-icon" class="animate-spin opacity-0 transition-all duration-200" alt="Loading..." />');
```
wescopeland, line 67: `alt` appears to not be a valid attribute for SVG elements.
Based on what I'm reading, it seems like the best course of action here may be to remove the `alt` attribute and hardcode an `aria-hidden="true"` to the element. This should work because these loading spinners are almost always short-lived. We should probably apply this to all the new loading spinners.

public/viewforum.php
```diff
@@ -134,7 +134,7 @@
 
         echo "<tr>";
 
-        echo "<td class='p-1'><img src='" . asset('assets/images/icon/forum-topic-unread.gif') . "' width='20' height='20' title='No unread posts' alt='No unread posts'></td>";
+        echo "<td class='p-1'>" . Blade::render('<x-fas-arrow-alt-circle-right class="h-4 w-4" :altText="$altText"/>', ['altText' => "No unread posts"]) . "</td>";
```
wescopeland, line 137: `alttext` is not a valid HTML attribute. Recommend instead using `title` on the `<td />` element.

wescopeland (commented):
Looking much better - just a bit more feedback.

public/createtopic.php
grablair: Makes sense!

public/viewforum.php
grablair: Sounds good.

For the record: as is, the code seems to use the term `No unread posts` as the `alt` for this icon on all topics, regardless of whether they are actually read or unread. I went into the repo history and it's been there ever since the birth of the repo (first commit that included the viewforum.php file: https://github.com/RetroAchievements/RAWeb/commit/83abe6097135bdef24bd2a0740444460b49772f3)

We should probably use a different `title`, unless you know somewhere in the code that modifies these via JS or something (me searches came up empty).

`title="Forum topic"` and `title="Forum` for `viewforum.php` and `forum.php`, respectively?

Thoughts?

wescopeland: Good catch. In that case, it may actually be harmful to ascribe any semantic meaning to these icons. Maybe we should just set them to `aria-hidden="true"`.

grablair: Yeah I had that thought as well. Sounds good!

wescopeland (approved):

public/editpost.php
```diff
@@ -90,7 +90,7 @@ class="w-full"
                     <div class="textarea-counter text-right" data-textarea-id="commentTextarea"></div>
 
                     <div class="flex gap-2">
-                        <img id="preview-loading-icon" src="$loadingIconSrc" style="opacity: 0;" width="16" height="16" alt="Loading..." class="w-4 h-4">
+                        $loadingIcon
```
Jamiras, line 93: This icon doesn't seem to be centered correctly.
image

Jamiras (commented):
Seem okay to me. I'm not sold on the new submit post button, but I don't have any better suggestions at this time.

public/editpost.php
grablair: Updated

Screenshot 2024-01-09 at 18 43 45

grablair:
> Seem okay to me. I'm not sold on the new submit post button, but I don't have any better suggestions at this time.

I have the same feeling, @Jamiras. What if we didn't have an icon at all? A simple "Submit" looks cleaner, IMO.

Screenshot 2024-01-09 at 19 31 41

Jamiras:
> What if we didn't have an icon at all? A simple "Submit" looks cleaner, IMO.

I'm fine with that, and it's more consistent with the forum and message pages.

Your image shows it the same height as the field. What happens if the field is stretched?

grablair:
> > What if we didn't have an icon at all? A simple "Submit" looks cleaner, IMO.
> 
> I'm fine with that, and it's more consistent with the forum and message pages.
> 
> Your image shows it the same height as the field. What happens if the field is stretched?

I set its height explicitly, so it remains the same size and does not move.

Screenshot 2024-01-10 at 11 56 00

app/Helpers/render/comment.php
```diff
@@ -163,7 +164,7 @@ function RenderCommentInputRow(string $user, int $articleTypeId, int $articleId)
 {
     sanitize_outputs($user, $formStr);
     $commentId = "art_{$articleTypeId}_{$articleId}";
-    $submitImageUrl = asset('assets/images/icon/submit.png');
+    $submitImage = Blade::render('<x-fas-square-plus class="w-5 h-5" />');
```
Jamiras, line 167: `$submitImage` is no longer used.

Jamiras (approved):

Issue #2118, opened 2023-12-22
Achievement tooltips cut off with long word in title


laytonloztew:
Achievements that have a long word in their title cause display issue on tooltip.

Example achievements:
https://retroachievements.org/achievement/252218
https://retroachievements.org/achievement/230516
https://retroachievements.org/achievement/343861
https://retroachievements.org/achievement/88981
https://retroachievements.org/achievement/222830
https://retroachievements.org/achievement/184185
https://retroachievements.org/achievement/310450
https://retroachievements.org/achievement/366061

**Screenshots**
![image](https://github.com/RetroAchievements/RAWeb/assets/12983268/7a345b39-20de-47ec-9c1a-bf5e6d7d041e)
![image](https://github.com/RetroAchievements/RAWeb/assets/12983268/dabe7238-887d-4cf5-a2cf-ad90856ef9c4)

**Desktop:**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 120.0.6099.130

PR #2138, opened 2024-01-07
fix: remediate overflow issue with long achievement titles in tooltips

wescopeland:
This PR fixes a text overflow issue with achievement titles in some achievement cards, for example `[ach=252218]`.

**Before**
Screenshot 2024-01-07 at 6 53 10 PM

**After**
Screenshot 2024-01-07 at 6 53 02 PM

Resolves https://github.com/RetroAchievements/RAWeb/issues/2118.


Jamiras (approved):
Verified this fixes the reported issue of long words not wrapping, but I did notice an issue with apostrophes in the title (not caused by this change). Tooltips for [ach=230516] and [ach=366061] as reported in the original issue do not display at all.

wescopeland:
The fix was substantial enough to warrant a separate PR: https://github.com/RetroAchievements/RAWeb/pull/2142

PR #2139, opened 2024-01-08
fix: change site-awards to grid layout

### LINKED ISSUE #1855: Badges descentralized

drisc:
Resolves #1855

Changes site awards to use a grid layout with specific numbers of columns at each breakpoint, 5 by default for XS (works best for mobile devices), 8 at SM, 12 at MD and 5 at LG/XL for the sidebar.

There is some additional padding set to try and keep the grid the same size at XL as it is at LG since the sidebar is slightly wider.

SM:
image

MD:
image

LG:
image

XL:
image


wescopeland:
There seems to be a substantial increase in padding in the component on XL between all elements.

**Prod**
Screenshot 2024-01-07 at 8 57 49 PM

**Branch**
Screenshot 2024-01-07 at 8 57 58 PM

wescopeland:
Now that I see this live, I'm not totally convinced it's the best solution.

XS:
Screenshot 2024-01-07 at 9 03 38 PM

This is tricky, indeed. It's almost as though the number of columns should be unbound and agnostic to the window's current breakpoint, being driven purely by wrapping in the container itself.

drisc:
I've added an extension to the tailwind config using `gridTemplateColumns` that will create a grid that fills the available container. This should solve the visual issue of large gaps at certain screen sizes. I've also adjusted the padding to try and bring it in line with how prod looks.

wescopeland:
I like the look of every breakpoint except XS:

Screenshot 2024-01-08 at 5 07 20 PM

This still seems like too much space between the badges. I think what might be better is to center the badge container itself, similar to what XL is doing.

drisc:
I've added a max width to the grid at XS and centered it with `flex`, this is what it looks like at iPhone SE size in the Firefox dev tools:
image

The max width of XS is 639px but the only thing that I can think of that would hit this size is a small, thin window on desktop, most phones would appear similar to the SE screenshot or have enough width to hit the MD breakpoint..

tailwind.config.js
```diff
@@ -62,6 +62,9 @@ module.exports = {
           '75%': { transform: 'rotate(-1.5deg)' },
         },
       },
+      gridTemplateColumns: {
+        'auto-fill-award-grid': 'repeat(auto-fill, minmax(52px, 1fr))',
+      },
```
wescopeland, line 67: While functional, this is not utility-first, so it goes against Tailwind's internal design principles. To better illustrate, this would be utility first if `award-grid` were a themeable unit of measurement, but this is more targeted towards a domain concern.

Instead, because it's a domain concern, JIT values are a better fit.

```
class="grid-cols-[repeat(auto-fill,minmax(52px,1fr))]"
```

app/Helpers/render/site-award.php
```diff
@@ -139,8 +139,8 @@ function RenderAwardGroup(array $awards, string $title, string $awardsOwnerUsern
 
     echo "<div id='" . strtolower(str_replace(' ', '', $title)) . "'>";
     echo "<h3 class='flex justify-between gap-2'><span class='grow'>$title</span>$counters</h3>";
-    echo "<div class='flex justify-center'>";
-    echo "<div class='component flex flex-wrap gap-2 justify-start bg-embed w-full xl:rounded xl:py-2 xl:px-5'>";
+    echo "<div class='max-sm:flex max-sm:justify-center'>";
+    echo "<div class='component w-full bg-embed gap-2 grid grid-cols-auto-fill-award-grid max-sm:grid-cols-5 max-sm:max-w-xs max-sm:px-4 max-sm:py-2 lg:gap-x-0 xl:py-2 xl:px-4'>";
```
wescopeland, line 143: Something is adding roughly 0.6rem of bottom margin to the `<h3>`, only on XS. Let's remove this for now so the spacing between the header and content is visually consistent across all breakpoints.

tailwind.config.js
drisc: I didn't realise we could do JIT, very cool. Moved the auto-fill out of the config.

app/Helpers/render/site-award.php
drisc: I assume this is being caused by the max-sm:py-2 that I added to match the XL breakpoint. This has now been removed.

wescopeland (approved):
I approve the implementation details, but I'd like a second opinion on how things look at the XS breakpoint.

Jamiras:
> Changes site awards to use a grid layout with specific numbers of columns at each breakpoint, 5 by default for XS (works best for mobile devices), 8 at SM, 12 at MD and 5 at LG/XL for the sidebar.

I'm seeing 5 at XS, LG, and XL. MD still changes from 12-16 and SM changes from 9-11 (depending on available width). As long as they're not fixed, why should XS be fixed?

Other than the limiting XS to 5 icons, the only change I see is the alignment when not in the sidebar. SM and MD justify the icons so they spread to both sides. XS centers them. 

#1855, which this claims to fix, only points out the the icons are left-aligned in the bottom panel, and center aligned in the side bar. I don't understand what all the other changes are trying to fix.

drisc:
> As long as they're not fixed, why should XS be fixed?

If there is not a hard constraint to 5 columns at XS then the icons spread out again since there are only 4 at the width of the container that I set up to center the icons at XS. If I remove that container width and the `flex` that centers it then it will expand to the size that Wes noticed in the image below.

image

I constrained and centered the container to remove the obvious gaps that were caused and brought it inline with how the sidebar looks at XL as per Wes' suggestion.

Without that constraint then it expands the same as SM and MD but the gaps look wide at iPhone SE size. I can remove that constraint if you think that will look better/be more consistant.

> #1855, which this claims to fix, only points out the the icons are left-aligned in the bottom panel, and center aligned in the side bar. I don't understand what all the other changes are trying to fix.

I was essentially trying to clean up the layout of the awards at each screen size and try to improve how it looks at certain breakpoints. I added extra padding to some areas to help keep the icons from spreading due to the way that css grid lays things out sometimes (this was something that I did at XL since the container widens, but only enough that the gaps between the awards look weird).

Wes' previous attempt to centralise the awards container led to the icons being centered in the container if there were not enough to fill it, I converted to a grid to combat that issue since they will follow the template that is defined with an auto-fill definition and then that resulting container can be padded or centered with the awards flowing to fit it.

drisc:
Made some more changes to try and bring it inline with the issue that it was intended to fix. The grid has uniform spacing across all breakpoints and is centered in it's container which fixes the slight alignment issue on desktop and mobile. This also removes the breakpoint specific padding at XL since the grid is centered anyway.

wescopeland (commented):
I again have no concerns with the current implementation, however would prefer to defer final judgment to @Jamiras.

Jamiras (approved):

Issue #2063, opened 2023-11-27
Errors displaying achievements tooltips with quotes in forum posts


Mindhral:
**Describe the bug**
Forum posts with achievements links ([ach] tags) provoke javascript warnings and errors in the browser console, and the tooltip doesn't display, when the game title or the achievement title contains single quotes.

**To Reproduce**
1. Go to [Achievement of the Week 2023](https://retroachievements.org/viewtopic.php?t=19685) topic for example
2. Open the browser console: it's full of warnings and errors
3. Hover on a link such as "Masked Jungle Warrior": the tooltip doesn't display and a bunch of other errors appears in the console (because of the quote in the game title (it's "Majora's Mask")
4. Hover on a link such as "Warhawk's Rooftop": same result because of the quote in achievement's title

**Expected behavior**
Tooltips on all achievements, no JS errors.

**Screenshots**
While hovering over the link:
![image](https://github.com/RetroAchievements/RAWeb/assets/148589639/a0aeb6f0-abbd-43a3-a033-58bccb755f8a)

Example of warning:
```
Alpine Expression Error: missing ) after argument list

Expression: "tooltipComponent($el, { staticHtmlContent: useCard('achievement', '55510', '', '<div class=\'tooltip-body flex items-start gap-2 p-2\' style=\'max-width: 400px\'><img src=\'https://media.retroachievements.org/Badge/119641.png\' width=\'64\' height=\'64\' /><div><div><b>Witch's Brew</b></div><div class=\'mb-1\'>Obtain a bottle from the Old Hag.</div><div>2 Points</div><div><i>Legend of Zelda, The: Majora's Mask</i></div></div></div>') })"

 <span class="inline" x-data="tooltipComponent($el, { …></div></div></div>') })" @mouseover="showTooltip($event)" @mouseleave="hideTooltip" @mousemove="trackMouseMovement($event)">
app-ee02857a.js:1:4982
```

HTML code of the corresponding element:
```
[...]

PR #2142, opened 2024-01-08
fix(achievement): remediate tooltip card rendering on some achievement titles

wescopeland:
Resolves the achievement tooltip card issue raised here: https://github.com/RetroAchievements/RAWeb/pull/2138#pullrequestreview-1808020900

> ... I did notice an issue with apostrophes in the title (not caused by this change). Tooltips for [ach=230516] and [ach=366061] as reported in the original issue do not display at all.

This PR slightly modifies the rendering function for achievement tooltip cards such that cards for achievements such as 230516 can now render correctly.


MSGoodman:
I don't think I have permissions to review or approve anything, but I can attest that this code is clear to me and that testing it locally it works. 

Perhaps worth noting, for record-keeping: I believe this will close #2063

Jamiras (approved):

app/Helpers/render/achievement.php
```diff
@@ -91,12 +94,15 @@ function renderAchievementCard(int|string|array $achievement, ?string $context =
...
 
+    $titleClassName = '';
     $renderedType = null;
     if ($type) {
         $renderedType = str_replace("\n", '', Blade::render(
             '<x-achievement.thin-type-indicator :type="$type" />',
             ['type' => $type]
         ));
+
+        $titleClassName = 'max-w-[190px]';
```
Jamiras, line 105: Is 190 perhaps too conservative?

image

wescopeland: Updated in latest to take the indicator label type into account when setting the width.

Issue #2098, opened 2023-12-11
`GetUserSummary` Awarded return type varies based on criteria


LaunchBox-C-Beats:
When using the GetUserSummary end point the return type for the `Awarded` property isn't consistent. When passing a value of `5` for both count parameters `Awarded` is returned as a dictionary. If you pass `0` for those same parameters the property is returned as an empty array.

wescopeland:
Transferred to RAWeb.

LaunchBox-C-Beats:
RecentAchievements property does the same

LaunchBox-C-Beats:
I am doing some string replace logic in my code to compensate, so this isn't critical for us at the moment (in case you need this information to correctly triage where to place your efforts).

MSGoodman:
Could this be as simple as changing the relevant lines in `API_GetUserSummary.php` from

```
// if no games were requested, initialize empty arrays for Awarded and RecentAchievements
$retVal['Awarded'] = [];
$retVal['RecentAchievements'] = [];
```

to 

```
// if no games were requested, initialize empty arrays for Awarded and RecentAchievements
$retVal['Awarded'] = json_encode(new stdClass());
$retVal['RecentAchievements'] = json_encode(new stdClass());
```

and updating the related test in `UserSummaryTest`? 

Or is there a more elegant solution?

Also, as much as I like consistency, are we worried that this might break other applications that expect this to return an empty array instead of an empty object?

wescopeland:
> are we worried that this might break other applications that expect this to return an empty array instead of an empty object?

Yep -- most of our API changes are additive (new data or new endpoints altogether) rather than tweaking existing stuff. Eventually we'll roll out a V2 of the API that hopefully is a bit more resilient and allows users to more properly structure their expected response in the request (ie: GraphQL, OData, trpc, etc).

Unfortunately even though the request here is perfectly reasonable, it will likely cause some breakage out there in the ecosystem.

Jamiras:
> Unfortunately even though the request here is perfectly reasonable, it will likely cause some breakage out there in the ecosystem.

I think this change is perfectly reasonable. The client shouldn't have to look for both a dictionary and/or an array for a specific field. A similar fix was made for another API (see #484)

https://github.com/Jamiras/RAWeb/blob/fc2e030c13e460930d47769d9807e125dbb3b123/public/API/API_GetGameInfoAndUserProgress.php#L59-L61

wescopeland:
[...]

Issue #484, opened 2020-05-19
API_GetGameInfoAndUserProgress, Achievement property isn't consistent type


CBeatt13:
Having an issue with the API_GetGameInfoAndUserProgress.php end point. If a game has achievements the Achievements property returned via the JSON is a dictionary. If a game does not have any achievements the type returned is an array. Because of this deserialization in C# throws first-chance exceptions. Would it be possible to ensure the return type stays consistent on the end point to avoid this?

claytonsulby:
I have the same issue. When I was creating structs for the JSON information some empty dicts are decoded as arrays. I think this has something to do with PHP. If you still need help there's an online tool that creates decode-ready structs for many languages here

https://app.quicktype.io/#l=c#

Also, I know you are using C#, but here's a modified version of the autogenerated enum that allows me to access the internal typed array. You might find this useful
```swift
    public enum GameProgressAchievements_DTO: Codable, Equatable  {
        
        case achievementMap([String: Achievement_DTO])
        case anythingArray([JSONAny])

        public init(from decoder: Decoder) throws {
            let container = try decoder.singleValueContainer()
            if let x = try? container.decode([JSONAny].self) {
                self = .anythingArray(x)
                return
            }
            if let x = try? container.decode([String: Achievement_DTO].self) {
                self = .achievementMap(x)
                return
            }
            throw DecodingError.typeMismatch(Achievement_DTO.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Wrong type for Achievements"))
        }

        public func encode(to encoder: Encoder) throws {
            var container = encoder.singleValueContainer()
            switch self {
            case .achievementMap(let x):
                try container.encode(x)
            case .anythingArray(let x):
                try container.encode(x)
            }
        }
        
        public static func == (lhs: PHPHelper.GameProgressAchievements_DTO, rhs: PHPHelper.GameProgressAchievements_DTO) -> Bool {
            return lhs.array == rhs.array
        }
        
        public var array:[Achievement_DTO] {
            
            switch self {
            case .achievementMap(let x):
                let achievements = x.map({ (key:String, achievement: Achievement_DTO) -> Achievement_DTO in
                    achievement
                })
                return achievements
            case .anythingArray(_):
                return []
            }

        }
        
    }

```

PR #2143, opened 2024-01-09
fix: return empty objects for user summary API even if no games requested

MSGoodman:
Addresses #2098 by always returning an object(/empty dictionary) rather than an array for the properties `RecentAchievements` and `Awarded`.

This is a super small PR, but people seem to have opinions about doing things this way vs. `json_decode('{}')` or something with `JSON_FORCE_OBJECT`. It's unclear to me what the most idiomatic solution is but I'm open to anything.


public/API/API_GetUserSummary.php
```diff
@@ -121,9 +121,9 @@
     unset($retVal['LastGame']['RichPresencePatch']);
     unset($retVal['LastGame']['system']);
 } elseif ($recentGamesPlayed === 0) {
-    // if no games were requested, initialize empty arrays for Awarded and RecentAchievements
-    $retVal['Awarded'] = [];
-    $retVal['RecentAchievements'] = [];
+    // if no games were requested, initialize empty objects for Awarded and RecentAchievements
+    $retVal['Awarded'] = json_encode(new stdClass());
+    $retVal['RecentAchievements'] = json_encode(new stdClass());
```
wescopeland, line 126: This returns strings:

```
GET http://localhost:64000/API/API_GetUserSummary.php?u=WCopeland&g=0&z=WCopeland&y=...
```

```json
	"Awarded": "{}",
	"RecentAchievements": "{}",
```

MSGoodman: This is hilarious and dumb, I didn't realize, thank you for catching this. Will fix.

Jamiras: Consider using `ArrayObject` as was done for #484: https://github.com/Jamiras/RAWeb/blob/fc2e030c13e460930d47769d9807e125dbb3b123/public/API/API_GetGameInfoAndUserProgress.php#L59-L61

MSGoodman: Thank you very much, that seems to have done it! Still picking up some of the quirks of PHP!

Jamiras (approved):

wescopeland (approved):

PR #2148, opened 2024-01-12
feat(hub): allow filtering by progression status

wescopeland:
This PR adds a new "Status" filter to Hubs, which is nearly identical to the filter that already exists on the Completion Progress page. Users can now filter hubs by their progression status of all games within the hub.

https://github.com/RetroAchievements/RAWeb/assets/3984985/5335e15b-cbf1-4719-ac57-e1fdf3c29279

Additionally:
* Fixed an a11y issue where the table heading was not wrapped in a `<thead>`.
* Added a reusable empty state component.
* Removed a dead filter from the Completion Progress page.


app/Platform/Controllers/PlayerCompletionProgressController.php
```diff
@@ -279,9 +279,6 @@ private function useGameStatusFilter(string $statusValue, array $filteredAndJoin
             'unawarded' => fn ($game) => !isset($game['HighestAwardKind']),
             'awarded' => fn ($game) => isset($game['HighestAwardKind']),
 
-            'pristine-mastered' => fn ($game) => isset($game['HighestAwardKind'])
-                && ($game['HighestAwardKind'] === 'mastered' && $game['PctWonHC'] == 1),
```
wescopeland, line 283: This was unused.

resources/views/components/meta-panel/filters.blade.php
wescopeland, line 1: Renamed to _checkbox-filters.blade.php_.

app/Platform/Controllers/RelatedGamesTableController.php
```diff
@@ -54,23 +58,53 @@ public function __invoke(Request $request): View
...
             'populated' => 'Only with achievements',
         ];
+        $availableSelectFilters = [
+            [
+                'kind' => 'status',
+                'label' => 'Status',
+                'options' => [
+                    'all' => 'All games',
+                    'unstarted' => 'No progress',
+                    'unawarded' => 'Unfinished',
```
Jamiras, line 83: Should `unawarded` also return `unstarted`? Or maybe the label could be clarified?

app/Platform/Controllers/RelatedGamesTableController.php
```diff
@@ -54,23 +58,53 @@ public function __invoke(Request $request): View
...
+                    'eq-beaten-softcore' => 'Beaten (softcore)',
+                    'eq-beaten-hardcore' => 'Beaten',
```
Jamiras, line 85: Should `beaten` include `mastered`?

app/Platform/Controllers/RelatedGamesTableController.php
```diff
@@ -54,23 +58,53 @@ public function __invoke(Request $request): View
...
+                    'eq-completed' => 'Completed',
+                    'eq-mastered' => 'Mastered',
+                    'any-beaten' => 'Beaten, either hardcore or softcore',
```
Jamiras, line 88: This should be moved up next to the individual Beaten selections.

app/Platform/Controllers/RelatedGamesTableController.php
```diff
@@ -54,23 +58,53 @@ public function __invoke(Request $request): View
...
+                    'gte-completed' => 'Completed or mastered',
+                    'awarded' => 'Games with any award',
+                    'eq-revised' => 'Games with awards for revised sets',
+                    'gte-beaten-softcore' => 'Games with softcore awards',
+                    'gte-beaten-hardcore' => 'Games with hardcore awards',
+                    'any-softcore' => 'Games with any softcore progress',
```
Jamiras, line 94: This and `any-hardcore` throw an exception:
> Trying to access array offset on value of type null

resources/views/components/meta-panel/index.blade.php
```diff
@@ -1,44 +1,60 @@
...
-    $hasFilters = count($availableFilters) > 0;
+    $hasSorts = !empty($availableSorts);;
+    $hasCheckboxFilters = !empty($availableCheckboxFilters);
+    $hasSelectFilters = !empty($availableSelectFilters);
+
     if ($hasSorts) {
-        $header = $hasFilters ? "Sorts and filters" : "Sorts";
-    } else if ($hasFilters) {
+        $header = ($hasCheckboxFilters || $hasSelectFilters) ? "Sorts and filters" : "Sorts";
+    } else if ($hasCheckboxFilters) {
```
Jamiras, line 16: This should also allow for only select filters.

resources/views/platform/components/game/game-list.blade.php
```diff
@@ -1,22 +1,34 @@
...
+);
+?>
+
 <div>
-    @if (count($consoles) < 1)
-        <p>{{ $noGamesMessage }}</p><br/>
+    @if (count($consoles) < 1 && $areFiltersPristine)
+        <div class="mb-12">
+            <x-empty-state>{{ $noGamesMessage }}</x-empty-state>
+        </div>
```
Jamiras, line 25: This logic doesn't seem to be working: http://localhost:64000/game/451

I'm assuming it's meant to hide the "no games" message if no filters are applied?

Or maybe this is just a special case where no filters exist.

app/Platform/Controllers/RelatedGamesTableController.php
wescopeland: In the latest commit, I've tried to make the filter options a little more clear:

Screenshot 2024-01-14 at 11 08 41 AM

`unawarded` internally has been renamed to `lt-beaten-softcore`. The label has changed to "Has progress, but no award".

app/Platform/Controllers/RelatedGamesTableController.php
wescopeland: Renamed internally to `eq-beaten-softcore-or-beaten-hardcore`. The label has been changed to "Beaten, but still missing achievements".

I feel like now with the other changes, this should probably stay put, but will defer to your opinion if you still feel it should be moved:

Screenshot 2024-01-14 at 11 08 41 AM

app/Platform/Controllers/RelatedGamesTableController.php
wescopeland: `any-hardcore` removed entirely in latest, however `any-softcore` has been retained and the error has been fixed.

resources/views/components/meta-panel/index.blade.php
wescopeland: Fixed in latest.

resources/views/platform/components/game/game-list.blade.php
wescopeland: Sorry, I am having trouble understanding. Can you explain this a different way?

For better context, there is a bug currently in prod where if you filter a list down to where there is no content, you can't clear your filter. To repro, visit:

https://retroachievements.org/game/163

Then filter by "Only with achievements". You'll get an empty state, but no way to get back.

Try the same on this branch and you'll have an empty state with the filter options still visible.

app/Platform/Controllers/RelatedGamesTableController.php
Jamiras: My brain says all the beaten stuff should be together and all the completed/mastered stuff should be together. "Beaten, but still missing achievements" falls between those two classifications.

I'd probably also move "Has any softcore progress" down to keep the "revised" option near the other completed/mastered options.

app/Platform/Controllers/RelatedGamesTableController.php
```diff
@@ -54,23 +58,48 @@ public function __invoke(Request $request): View
...
+                    'gte-beaten-softcore' => 'Beaten (softcore) or greater',
+                    'gte-beaten-hardcore' => 'Beaten or greater',
+                    'gte-completed' => 'Completed or mastered',
+                    'eq-mastered' => 'Mastered',
+                    'eq-beaten-softcore-or-beaten-hardcore' => 'Beaten, but still missing achievements',
+                    'any-softcore' => 'Has any softcore progress',
+                    'revised' => 'Completed or mastered, but the set was revised',
+                ],
+            ],
+        ];
```
Jamiras, line 93: These should be hidden when not logged in (`$loggedInUser === null` or `$this->gameListService->userProgress === null`)

resources/views/platform/components/game/game-list.blade.php
Jamiras: I just noticed that visiting a Central hub show a big banner saying there's no games.
image

It doesn't have any filter widgets, therefore it's impossible to have filters applied.

I imagine the intent of this code was to not show the banner if there are no games in the unfiltered state, which would be true for these centralized hubs.

app/Platform/Controllers/RelatedGamesTableController.php
wescopeland: Adjusted in latest, hopefully this is a bit better:

Screenshot 2024-01-14 at 3 30 16 PM

resources/views/platform/components/game/game-list.blade.php
wescopeland: OK, I think I understand now. I've changed the outer conditional logic to:

```php
@if (!$areFiltersPristine || count($consoles) > 0)
```

With this change, if a hub doesn't have any games (eg: 451), an empty state and meta surface won't display.
However, if a hub does have games and it's filtered to a state where games don't render, the empty state and meta surface both display.

app/Platform/Controllers/RelatedGamesTableController.php
wescopeland: Resolved in latest.

Jamiras (approved):

PR #2149, opened 2024-01-12
feat(hub): make table header rows sticky

wescopeland:
Makes the table header row stick to the top of the page as you scroll down. Also affects developer sets. The functionality is intact even when grouping tables by console.

https://github.com/RetroAchievements/RAWeb/assets/3984985/29856947-529a-4e91-ac17-18b293bd489a

### DIFF
--- a/resources/css/layout.css
+++ b/resources/css/layout.css
@@ -42,7 +42,6 @@ header {
 }
 
 main {
-  overflow-x: hidden;
   padding: 0;
   position: relative;
   min-height: 450px;
@@ -68,7 +67,6 @@ article {
   align-self: start;
   background: var(--box-bg-color);
   background: linear-gradient(180deg, var(--box-bg-color) 0%, var(--bg-color) 100%);
-  overflow: auto; /* 1 */
 }
 
 aside {
--- a/resources/views/platform/components/game/game-list.blade.php
+++ b/resources/views/platform/components/game/game-list.blade.php
@@ -40,76 +40,75 @@
                 <?php foreach ($totals as $key => $value) { $totals[$key] = 0; } ?>
             @endif
 
-            <div><table class='table-highlight mb-4'><tbody>
+            <div>
+                <table class='table-highlight mb-4'>
+                    <thead>
+                        <tr class="do-not-highlight sticky top-[42px] z-10 bg-box">
+                            @foreach ($columns as $index => $column)
+                                @php
+                                    $styles = [];
+                                    if (array_key_exists('width', $column)) {
+                                        $styles[] = 'width:' . $column['width'] . '%';
+                                    }
+                                    if (array_key_exists('tooltip', $column)) {
+                                        $styles[] = 'cursor:help';
+                                    }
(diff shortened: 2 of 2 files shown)


resources/css/layout.css
wescopeland, line 1: These styles I removed in order to make this work are much too greedy. Whatever pages are breaking overflow, it should be fixed on the page itself rather than having a global patch applied.

Jamiras: Do you have any idea what might be affected by this? Should we just merge it and cross our fingers?

wescopeland: I poked around quite a bit across games, users, achievements, forums, and lists but was unable to find anything that was actually affected by this across any breakpoint from XS to XL.

I assume it "fixed" something at some point in time. It's possible whatever it impacted was changed when we enabled the XL breakpoint or migrated some aspect of the UI to Blade/Tailwind.

It's probably reasonable to consider this could cause some minor CSS regression on some obscure page or feature somewhere. If it does, it'll be a trivial fix that can be scoped to that specific area of the app rather than needing these styles applied app-wide. These styles are akin to setting a global variable that's only used by 1 helper function somewhere.

Worst case scenario: somewhere in the app has a horizontal scrollbar where content was previously truncated.

resources/views/platform/components/game/game-list.blade.php
```diff
@@ -40,76 +40,75 @@
...
+                                    if (array_key_exists('width', $column)) {
+                                        $styles[] = 'width:' . $column['width'] . '%';
+                                    }
+                                    if (array_key_exists('tooltip', $column)) {
+                                        $styles[] = 'cursor:help';
+                                    }
+                                    $alignment = $column['align'] ?? 'left';
+                                    $class = $alignment !== 'left' ? "text-$alignment" : '';
+                                    $class .= ($index === 0) ? ' rounded-tl' : '';
+                                    $class .= ($index === count($columns) - 1) ? ' rounded-tr' : '';
```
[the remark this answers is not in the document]
wescopeland: Sorry, this is cruft that I forgot to remove. Deleted in latest.

resources/views/platform/components/game/game-list.blade.php
```diff
@@ -40,76 +40,75 @@
                 <?php foreach ($totals as $key => $value) { $totals[$key] = 0; } ?>
             @endif
 
-            <div><table class='table-highlight mb-4'><tbody>
+            <div>
+                <table class='table-highlight mb-4'>
+                    <thead>
+                        <tr class="do-not-highlight sticky top-[42px] z-10 bg-box">
```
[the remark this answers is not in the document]
wescopeland: Good catch. This should now be resolved in latest. This is the change:

https://github.com/RetroAchievements/RAWeb/pull/2149/commits/8690117bf65790e84a1b12e2ad0ae927d5352263#diff-77a1114bd788d9cf7334b5e2938094ce30837d04fd875980627f26a2953cdac1R7

Jamiras (approved):

PR #2150, opened 2024-01-12
feat(controlpanel): add copy to clipboard functionality for web api key

wescopeland:
The web API key is now a giant button, which the user can click to copy the key to the clipboard. There is a true copy to clipboard button added next to the key too (to signify there's some kind of interaction here).

https://github.com/RetroAchievements/RAWeb/assets/3984985/b00fae9f-2add-4d89-97ce-1fc58e06fb4f


public/controlpanel.php
```diff
@@ -279,9 +279,17 @@ function confirmEmailChange(event) {
             echo "<tr>";
             echo "<td class='align-top'>Web API Key</td>";
             echo "<td>";
-            echo "<input class='mb-1' size='60' readonly value='$apiKey'>";
+
+            $copyIcon = Blade::render('<x-fas-copy />');
 
             echo <<<HTML
+                <div role="button" x-init="{}" @click="copyToClipboard('$apiKey')" class="flex items-center gap-x-2 bg-embed max-w-fit rounded pl-2 mb-2">
```
Jamiras, line 286: Is there a point to `x-init={}`?

public/controlpanel.php
```diff
@@ -279,9 +279,17 @@ function confirmEmailChange(event) {
...
+                    <p class="font-mono text-neutral-200 light:text-neutral-700">$apiKey</p>
+                    <button class="btn transition-transform lg:active:scale-95" @click="copyToClipboard('$apiKey')" title="Copy web API key to clipboard" aria-label="Copy web API key to clipboard">
+                        $copyIcon
+                    </button>
```
Jamiras, line 290: The button border blends in with the background and makes it seems unaligned with the text field
image

public/controlpanel.php
wescopeland: Yes. Without this directive, Alpine.js doesn't have an execution context for the component.

public/controlpanel.php
wescopeland: Screenshot 2024-01-13 at 7 00 41 PM

I've made an adjustment in latest such that it's no longer ambiguous where the button is. There's now a single button, with the key and icon contained therein.

Jamiras (approved):
Hovering over it makes it change color, but it's not entirely clear what will happen when clicking. Might want to add a tooltip.

wescopeland:
That is a good point. In latest, I've added both a `title` and `aria-label` attribute to meet the gamut of any accessibility concerns that could surface with this button.

Jamiras (approved):

PR #2156, opened 2024-01-14
fix(user-card): dont show empty motto

wescopeland:
Resolves an issue where if a motto is empty, the space for the motto in the user card still appears:

Screenshot 2024-01-13 at 9 34 07 PM


resources/views/components/community/user-card.blade.php
wescopeland, line 1: While I was in here, I changed the comments to use `{{-- --}}` syntax as opposed to `<!-- -->` syntax.
The former syntax results in the comments to not be sent down to the browser.

Jamiras (approved):

PR #2158, opened 2024-01-14
feat(forum): migrate post history to blade & improve mobile-friendliness

wescopeland:
This PR migrates the inner content of `/forumposthistory.php` to Blade, specifically reusing the post history component already leveraged on the homepage.

Resolves https://discord.com/channels/310192285306454017/1183823595978362880.

**Before**
Screenshot 2024-01-14 at 6 37 16 PM

Screenshot 2024-01-14 at 6 37 25 PM

**After**
Screenshot 2024-01-14 at 6 36 39 PM

Screenshot 2024-01-14 at 6 35 47 PM

### DIFF
--- a/public/forumposthistory.php
+++ b/public/forumposthistory.php
@@ -88,7 +88,6 @@ RenderContentStart("Forum Post History");
                 :postedAt="$postedAt"
                 :summary="$summary"
                 :tooltipLabel="$tooltipLabel"
-                :variant="$variant"
                 :viewLabel="$viewLabel"
                 :viewHref="$viewHref"
                 :view2Label="$view2Label"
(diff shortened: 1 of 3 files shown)


Jamiras (approved):
My only comment here is that only 12 items fit on my desktop vs. 20 with the old layout. Maybe we should reduce the page size to 10 or 15?

wescopeland:
> Maybe we should reduce the page size to 10 or 15?

How strongly do you feel about this proposed reduction? I lean more towards keeping it as-is (otherwise this could be perceived as a feature impairment).

Jamiras:
About 2/3 of the rows are visible on initial page load currently. Less than half are visible on initial page load with this change.

In my opinion less scrolling is better. I'd prefer not having to scroll at all - just scan the list and click the Next button.

But I don't feel strongly enough to argue over the change. We can deploy like this and see if users have feedback.

wescopeland:
I've made a slight adjustment to the PR to shrink these down to 52px in height on desktop. On master, they're 48px. Hopefully this helps a bit.

Jamiras:
I think I prefer the three-line variant over having the title after the date. It looks really awkward with that gap. Particularly since things don't align from item to item.
image

It would probably look better if you put the forum title before the date:
image

But I feel as if the forum title is more important than the user. So maybe user could be moved down?
image

That separates the user from the date. None of these really work.

As much as I dislike the scrolling, I think the three-line version is preferable over anything I can come up with two lines.

wescopeland:
Reverted the most recent commit & temporarily converted to draft. I'm not in love with where desktop is here and will try to come up with a solution that's more agreeable for both mobile and desktop.

PR #2159, opened 2024-01-15
Add game suggestions page

Jamiras:
Adds an option to the user's menu
image

That opens a page listing several potential games the user might want to play.
image

Also adds a link to game pages that lists potential games to play if the user enjoyed a specific game (https://discord.com/channels/310192285306454017/1193986525256499300).
image

image

For the user's suggested games, a weighted pool is created between completely random suggestions, games on the user's want to play list, and games related to games that the user has mastered. As such, players with more things on their want to play list will see more suggestions related to the want to play list, and players with more masteries will see more things related to previously mastered games.

For the similar games list, the pool only contains games related to the source game.

When a related game is selected from the weighted pool, it's further randomized to return one of the following:
* A game specifically marked as similar
* A random game from a hub the related game is associated to
* A random game from an author of achievements in the related game
* The most commonly mastered game amongst a random subset of players who mastered the related game.

Once a candidate game has been selected (either from the random list, the want to play list, or the related list), we check to see if it's already in the list. If not, and the user has some progress in the game, the game may be discarded based on that progress (i.e. a game with 10% completion has a 10% chance to be discarded). If the game is still eligible at this point, it's added to the list.

Once the list is full, or a maximum number of attempts has been made, it's sorted and presented to the user.


resources/views/components/menu/account.blade.php
```diff
@@ -54,6 +54,7 @@
         <x-dropdown-item :link="route('user.completion-progress', $user)">Completion Progress</x-dropdown-item>
         
         @if($user->Permissions >= Permissions::Registered)
+            <x-dropdown-item :link="route('game.suggest')">Game Suggestions</x-dropdown-item>
             <x-dropdown-item :link="url('gameList.php?t=play')">Want to Play Games</x-dropdown-item>
```
wescopeland, line 58: An autogenerated list seems like it should have slightly lower priority than the list of entities explicitly chosen by the user. Recommend keeping Want to Play Games above Game Suggestions.

resources/views/platform/components/player/suggest-game-page.blade.php
wescopeland, line 1: Is this feature intended to primarily be used for game discovery?

If so, I can't help but wonder if the list should not include any game which contains progress. If the user is looking for something new to play, there's probably a reason why they aren't wanting to return to something that already has progress. I think the only edge case is the suggestion which informs the user about a revision.

Could there be something I'm not considering?

I propose:
* With the exception of revisions, do not return games that have progress.
* Remove the Progress column.

resources/views/platform/components/game/link-buttons/index.blade.php
```diff
@@ -100,4 +101,13 @@
             Set Requestors
         </x-game.link-buttons.game-link-button>
     @endif
+
+    @if ($canSeeSuggestedGames)
+        <x-game.link-buttons.game-link-button
+            icon="🕹️"
+            href="{{ route('game.suggest-for-game', $gameId) }}"
+        >
+            Find something similar to play
```
wescopeland, line 110: nit: I can't help but notice the casing on this makes it stick out from the rest of the buttons.

Screenshot 2024-01-15 at 6 15 41 PM

Because the button is so close to the "Similar Games" header, I wonder if users might find this ambiguous or duplicative.

Screenshot 2024-01-15 at 6 18 12 PM

The other buttons all seem to be nouns. Maybe something like "Suggested Games" would work better as a label here?

resources/views/platform/components/game-list-item/suggest-reason.blade.php
```diff
@@ -0,0 +1,61 @@
...
+            :gameId="$relatedGameId"
+            :gameTitle="$relatedGameTitle"
+            :gameImageIcon="$relatedGameIcon"
+            :iconSize="16"
+        />
+    @else
+        In Similar Games collection
+    @endif
+@elseif ($selectionMethod == 'common-hub')
+    Shares {{ $relatedSubject }}
```
wescopeland, line 24: Is the name of the hub actually important? I think we can improve readability by changing this to "hub". Right now it's ambiguous that `$relatedSubject` is a hub, and some hub titles are quite long.

**Before**
Screenshot 2024-01-15 at 6 23 58 PM

**After**
Screenshot 2024-01-15 at 6 22 43 PM

resources/views/platform/components/player/suggest-game-page.blade.php
wescopeland, line 1: It's possible to land on the page and have most or all of the recommendations be randomly selected:

Screenshot 2024-01-15 at 6 42 44 PM

I'd wager this will happen a lot of the time with accounts out of the top 1,000 in site rank.

Maybe the max # of randomly selected entities should be capped at 2 or 3. If we're reaching for stuff, maybe we can check the user's activity and find similar games based on that?

resources/views/platform/components/game-list-item/suggest-reason.blade.php
wescopeland, line 1: What if I'm a player who only plays softcore (no masteries) or only pursues beaten awards? Should we also derive recommendations from those statuses?

resources/views/platform/components/game-list-item/suggest-reason.blade.php
```diff
@@ -0,0 +1,61 @@
...
+        with mastered game
+        <x-game.avatar
+            :gameId="$relatedGameId"
+            :gameTitle="$relatedGameTitle"
+            :gameImageIcon="$relatedGameIcon"
+            :iconSize="16"
+        />
+    @endif
+@elseif ($selectionMethod == 'common-author')
+    Shares set developer {!! userAvatar($relatedSubject, icon: false) !!}
```
wescopeland, line 35: Similar feedback to naming the hub. Letting the user know a recommendation is due to a common author seems very valuable, but I'm not so sure most users will initially be interested in the name of the dev (which is also present at the top of the game page).

I feel this should simply read "Shares set developer with mastered game:".

resources/views/platform/components/game-list-item/suggest-reason.blade.php
wescopeland, line 1: Initially, my brain did not make the connection between the reasoning label and the game underneath. Recommend suffixing the labels with colons, eg:

Screenshot 2024-01-15 at 6 41 09 PM

wescopeland (commented):
Very cool. First pass on UX - I haven't looked too deeply into implementation details yet. Wanted to get a UX pass out of the way first.

resources/views/platform/components/player/suggest-game-page.blade.php
Jamiras: It's meant to solve the issue of "I'm bored, but I don't know what to play".

I'd argue that a player is more likely to be upset by having a game recommended that they purposefully reset all of their achievements on then if they happen to get something that they did 80% of and decided that was enough.

There's already logic to discard things based on the amount of progress a user has made. In the case above where the user has 80% of the achievements for a game, there's an 80% chance the recommendation will be discarded entirely.

If you feel games with any progress should be devalued, we can always skew that calculation (if completion > 0, discard if random(1, 100) - 30 < completion - that would automatically discard everything with 70% or more completion and only have a 69% chance of not discarding something with 1% completion).

Jamiras: As for the progress column itself, I have a lot of things on my Want to Play list in various states of progress because I want to revisit them at some time to try for the achievements I missed.

Additionally, if a user does get recommended something with progress, the bar would help them remember that they've already played it and prevent them from thinking it might be interesting and tossing it on their Want to Play list.

Jamiras: The default weighting of random games may be too much for early users, but a user without much history doesn't have much related stuff to pick from, so they will inevitably have more random selections anyway. 

After 30 attempts to find 10 games, the list is filled with random items. On a user with 3000 hardcore points and no Want to Play list, I still ended up with an average of 5 random recommendation when the weighting for random games was set to 0 - meaning that the normal randomization logic averaged 5 valid games out of 30 attempts to find suggestions.

I still want random games to appear for users who have hundreds of Mastered games or things in their Want to Play queue. I'll see if I can come up with a better algorithm for injecting random games.

Jamiras: I've settled on the random weight being 1/8 of the total of all the other weights (making it a roughly 11% chance of occurring before the filling that happens after 30 attempts). This seems to work well for players with few alternatives to recommend from and still includes some randomness for players with lots of alternatives.

resources/views/platform/components/game-list-item/suggest-reason.blade.php
Jamiras: I've modified the logic to correctly state "completed" vs. "mastered", and added logic to also randomly select games related to "beaten" games.

resources/views/platform/components/game-list-item/suggest-reason.blade.php
Krylan: I think that keeping name of the hub in recommendations could be better option. I personally would be interested in seeing what type of hub is associated with the game. Some hubs could be less relevant for the player, and other more (a specific game developer or game mechanic). Though, word "hub" could still be before or after the name to make it more obvious.

resources/views/platform/components/player/suggest-game-page.blade.php
wescopeland: > I'd argue that a player is more likely to be upset by having a game recommended that they purposefully reset all of their achievements on then if they happen to get something that they did 80% of and decided that was enough.

This is a valid point that I had not considered.

> If you feel games with any progress should be devalued, we can always skew that calculation (if completion > 0, discard if random(1, 100) - 30 < completion - that would automatically discard everything with 70% or more completion and only have a 69% chance of not discarding something with 1% completion).

I do lean towards this preference, but I'm unable to quantify it with anything other than conjecture. It would make for an interesting A/B test in the future. For now I'm happy to defer to your judgment on what is best.

resources/views/platform/components/game/link-buttons/index.blade.php
wescopeland: It's a tough one for sure. I notice in latest that the capitalization is now at least consistent with the rest of the buttons - that's probably good enough for now.

resources/views/platform/components/game/link-buttons/index.blade.php
wescopeland, line 1: In searching for potential edge cases, I landed on a game called "Animal Breeder 2", which does not have a set and has extremely limited metadata. The game ID is 4116.

If I try to get suggestions for this game, I'm taken to a blank page:

Screenshot 2024-01-20 at 6 11 12 PM

I wonder if there should be some threshold for when we show the button on game pages. Only show it if there are achievements that are promoted to Core? That's just a naive idea, but there could be some smarter conditions.

Jamiras: The problem with this particular game is that none of the games in any of the related hubs have achievements. That would be difficult to detect to hide the button. If a player is looking for a similar game to something without achievements, there's a possibility they won't find anything. And I don't think it's something that would happen frequently enough to matter.

This should have been displaying a "No suggestions found" widget, but there was a logic error in game-list.blade.php. I've corrected that.

resources/views/platform/components/game-list-item/suggest-reason.blade.php
Jamiras: I've added the 'hub' keyword before the hub name.

resources/views/platform/components/player/suggest-game-page.blade.php
Jamiras: I've added a 10% skew to the discard logic. Anything with 90%+ completion will automatically be discarded, and anything with 1% completion will have an 11% chance to be discarded.

resources/views/platform/components/game-list-item/suggest-reason.blade.php
wescopeland: The hub keyword before the hub name helps a lot.

resources/views/platform/components/game/suggest-game-page.blade.php
```diff
@@ -0,0 +1,36 @@
...
+    pageTitle="Game Suggestions - {{ $game->Title }}"
+    pageDescription="A list of random games that a user might want to play if they enjoyed {{ $game->Title }}"
+>
+    <x-game.breadcrumbs 
+        :targetConsoleId="$game->system->ID"
+        :targetConsoleName="$game->system->Name"
+        :targetGameId="$game->ID"
+        :targetGameName="$game->Title"
+        currentPageLabel="Game Suggestions"
+        />
```
wescopeland, line 20: nit: Fixed indentation.

```suggestion
    />
```

wescopeland (approved):
Verified locally. I don't have any concerns about the implementation, including _SuggestGameController.php_. The code is straightforward enough to follow, even for someone mathematically-challenged like me.

As a minor nit, I recommend converting loose equality `==`/`!=` to strict equality `===`/`!==` wherever possible. There are several instances of this in the controller and some of the template conditionals, however this is not a gating factor for me.

PR #2153, opened 2024-01-13
refactor(user): migrate Completion Progress to blade and show beaten indicators

wescopeland:
This PR rewrites/migrates the Completion Progress component on user profiles to Blade. The progress bars now show award indicators for beaten (softcore), beaten, completed, and mastered that mirror the Completion Progress page.

Implementation details:
* `<x-hardcore-progress />` has been renamed to `<x-game-progress-bar />`.
* `<x-game-progress-bar />` now optionally supports showing award indicators for those 4 aforementioned award states as well as unfinished. This is not currently enabled on any tables using the bars.

https://github.com/RetroAchievements/RAWeb/assets/3984985/3631fd9b-2aae-438d-9fcb-6a33284c427c

PR #2161, opened 2024-01-15
feat: show progress on console game lists

wescopeland:
This PR migrates console game lists to the new reusable datagrid functionality that was recently introduced on hubs.

## UI/UX Changes
* The new route is `/system/{systemId}/games`. Old routes will redirect to the new route.
* Console game lists now support showing user progress and have an add/remove to backlog button.
* Console game lists now show player count for all games, and users can sort by this value similar to hubs.
* Users can now filter by progress status (ported from https://github.com/RetroAchievements/RAWeb/pull/2148).
* When viewing the list of games without achievements or the entire game list, the results are no longer paginated. Thanks to data denormalization from V5, even with progress and backlog support enabled, this results in sub-50ms queries.
* SEO metadata has been greatly improved.
* The "Last Updated" column has, again, been removed. There will likely be resistance to this - we should educate users why it's being removed (because it's not considering achievements and is only considering game metadata like release date, etc).
* It is no longer possible to support by clicking on column headers. _It is likely this enhancement should come in a future PR._
* The points column no longer shows RetroPoints.

**Before**
Screenshot 2024-01-15 at 4 43 06 PM

**After**
Screenshot 2024-01-15 at 4 43 17 PM

## Future Improvement
* Column headings should likely be sortable.
* We should add award indicators to the progress bars once https://github.com/RetroAchievements/RAWeb/pull/2153 is merged.
* We should add a user setting which simulates the color script behavior. I'll take this up in a separate PR.
* We should dramatically improve the mobile UX. This will require a separate PR.


app/Platform/Services/GameListService.php
```diff
@@ -102,10 +102,12 @@ public function initializeGameList(array $gameIds): void
...
-            if (substr($game['Title'], 0, 1) == '~') {
-                $tilde = strrpos($game['Title'], '~');
-                $game['SortTitle'] = trim(substr($game['Title'], $tilde + 1) . ' ' . substr($game['Title'], 0, $tilde + 1));
+            $game['SortTitle'] = mb_strtolower($game['Title']);
+            if (substr($game['SortTitle'], 0, 1) == '~') {
+                $tilde = strrpos($game['SortTitle'], '~');
+                $game['SortTitle'] = trim(substr($game['SortTitle'], $tilde + 1) . ' ' . substr($game['SortTitle'], 0, $tilde + 1));
+            } else {
+                $game['SortTitle'] = " " . $game['SortTitle'];
             }
```
wescopeland, line 111: I changed the `SortTitle` assignment logic.

"Tagged" games (eg: unlicensed, prototype, hack, etc) will now be moved to the bottom of the list in these datagrids. This mirrors the current _gameList.php_ behavior and seems desirable.

app/Platform/Controllers/SystemController.php
```diff
@@ -76,16 +83,112 @@ public function destroy(System $system): void
...
+
+        $availableSelectFilters = [];
+        if ($loggedInUser) {
+            $availableSelectFilters = [
+                [
+                    'kind' => 'status',
+                    'label' => 'Status',
+                    'options' => [
+                        'all' => 'All games',
+                        'unstarted' => 'No achievements earned',
```
Jamiras, line 152: I don't think it's specifically related to this PR, but when I apply the "No achievements earned" filter to my ArduBoy list, Ardynia disappears.

All Games:
image

No achievements earned:
image

app/Platform/Controllers/SystemController.php
```diff
@@ -76,16 +83,112 @@ public function destroy(System $system): void
...
+                        'lt-beaten-softcore' => 'Has progress, but no award',
+                        'gte-beaten-softcore' => 'Beaten (softcore) or greater',
+                        'gte-beaten-hardcore' => 'Beaten or greater',
+                        'eq-beaten-softcore-or-beaten-hardcore' => 'Beaten, but still missing achievements',
+                        'gte-completed' => 'Completed or mastered',
+                        'revised' => 'Completed or mastered, but the set was revised',
```
Jamiras, line 158: Also likely unrelated: I have a completed game that's mostly hardcore and hasn't been revised, but it's showing up under "revised":
image

app/Platform/Services/GameListService.php
Jamiras: Rather than prefixing a space on "good" titles to move them up, it might be more efficient to only prefix the modified titles with a "~" to move them down?

app/Platform/Services/GameListService.php
```diff
@@ -375,7 +377,7 @@ private function getTitleColumn(array $filterOptions): array
         return [
             'header' => 'Title',
             'render' => function ($game) use ($filterOptions) {
-                if (!$filterOptions['console']) {
+                if (isset($filterOptions['console']) && !$filterOptions['console']) {
```
Jamiras, line 380: I had a slightly different solution to this problem in #2159: https://github.com/RetroAchievements/RAWeb/pull/2159/files#diff-c4abbf0b8b7d6b454423e86a2bb4115262798cf69424713559b116f0ba4c74f3R378

public/gameList.php
```diff
@@ -23,6 +23,10 @@
     $offset = 0;
 }
 
+if ($consoleIDInput > 0) {
+    return redirect()->route('system.game.index', ['system' => $consoleIDInput]);
+}
+
```
Jamiras, line 29: I notice "All Games" hasn't been converted. I'm guessing that's intentional?

app/Platform/Services/GameListService.php
wescopeland: Good idea, addressed in latest.

app/Platform/Controllers/SystemController.php
wescopeland: Good catch - in latest I've hunted down the bug that was causing this problem.

app/Platform/Controllers/SystemController.php
wescopeland: Yep, another logic error here. Should be fixed in latest.

app/Platform/Services/GameListService.php
wescopeland: I attempted to replace this line with `if (!($filterOptions['console'] ?? false)) {`, however, for some reason it's causing the console labels to reappear:

Screenshot 2024-01-20 at 10 54 58 AM

public/gameList.php
wescopeland: Intentional, as we currently don't support pagination and I figured such an implementation could be hairy enough to justify its own PR.

app/Platform/Services/GameListService.php
wescopeland: I agree with your gut feeling.

Addressed in latest. I've added a new public member to `GameListService` called `withSystemNames`, which defaults to `true`.

`SystemController` flips this value to `false`.

I've adjusted `GameListService::getTitleColumn()` to:

```php
private function getTitleColumn(array $filterOptions): array
{
    $consoleName = null;
    if ($this->withConsoleNames) {
        $consoleName = $filterOptions['console'] ?? null;
    }

    return [
        'header' => 'Title',
        'render' => function ($game) use ($consoleName) {
            echo '<td>';
            echo Blade::render('
                <x-game.multiline-avatar
                    :gameId="$ID"
                    :gameTitle="$Title"
                    :gameImageIcon="$ImageIcon"
                    :consoleName="$consoleName"
                />', array_merge($game, ['consoleName' => $consoleName])
            );
            echo '</td>';
        },
    ];
}
```

Jamiras (approved):

app/Platform/Services/GameListService.php
```diff
@@ -372,29 +379,24 @@ private function initializeUserGames(User $user, array $gameIds): void
...
+        $consoleName = null;
+        if ($this->withConsoleNames) {
+            $consoleName = $filterOptions['console'] ?? null;
+        }
+
         return [
             'header' => 'Title',
-            'render' => function ($game) use ($filterOptions) {
-                if (!$filterOptions['console']) {
-                    echo '<td class="py-2">';
```
Jamiras, line 379: You requested I add this "py-2" because the multi-line rows felt "snug": https://github.com/RetroAchievements/RAWeb/pull/2130#discussion_r1442207108

If we don't need that, we could move the <td> elements out of render and put them directly in the sets-page.blade.php.

wescopeland: I think I screwed up here; I'm seeing a regression.

wescopeland: Screenshot 2024-01-23 at 5 56 10 PM

Digging into this now.

app/Platform/Services/GameListService.php
Jamiras: This solution doesn't seem to be working quite right. The console names aren't visible on the hub lists when not grouping by console, and when grouping by console they all just say "1":
image

app/Platform/Services/GameListService.php
wescopeland: Sorry about that - what a mess.

There was both a style and a logic regression here. I think they're both cleaned up in latest.

`py-2` is now reapplied if `$consoleName` is truthy.

The `$consoleName` assignment logic has been adjusted to:

```php
$consoleName = null;
if ($this->withConsoleNames && empty($filterOptions['console'])) {
    $consoleName = $game['ConsoleName'] ?? null;
}
```

Now I correctly see console names again on hubs if I'm not grouping by console.

Jamiras (approved):

app/Platform/Services/GameListService.php
wescopeland: Thanks, I think we both noticed this around the same time. Addressed in latest, and further expounded upon here: https://github.com/RetroAchievements/RAWeb/pull/2161#discussion_r1464102596

app/Platform/Controllers/SystemController.php
```diff
@@ -76,16 +83,113 @@ public function destroy(System $system): void
...
+        $totalUnfilteredCount = count($gameIds);
+
+        $this->gameListService->initializeUserProgress($loggedInUser, $gameIds);
+        $this->gameListService->initializeGameList($gameIds);
+
+        if (isset($filterOptions['populated'])) {
+            $this->gameListService->filterGameList(function ($game) use ($filterOptions) {
+                return $this->usePopulatedFilter($game, $filterOptions['populated']);
+            });
+        }
```
luchaos, line 120: this fails for me:
```
 ------ -------------------------------------------------------------------------------------------------------------
  Line   app/Platform/Controllers/SystemController.php
 ------ -------------------------------------------------------------------------------------------------------------
  116    Offset 'populated' on array{populated: mixed, status?: mixed} in isset() always exists and is not nullable.
 ------ -------------------------------------------------------------------------------------------------------------
 ```

PR #2162, opened 2024-01-16
feat(home): use live Total Points Earned in global statistics

wescopeland:
Reimplements the "Points earned since March 2nd, 2013" using Livewire 3. The number now counts up in real-time as users unlock achievements and `StaticData` updates.

https://github.com/RetroAchievements/RAWeb/assets/3984985/61315292-db7a-475e-adcc-0c6e9ba612ae

To test this PR, in _TotalPointsEarned.php_, replace the first line in `render()` with `$this->totalPointsEarned += 18;`.

### DIFF
--- a/resources/views/community/components/global-statistics/total-points-earned-javascript.blade.php
+++ b/resources/views/community/components/global-statistics/total-points-earned-javascript.blade.php
@@ -1,0 +1,15 @@
+{{--
+    This stuff shouldn't be in the Livewire component because
+    the template is sent to the browser on each re-render.
+--}}
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/countup.js/1.7.0/countUp.min.js"></script>
+
+<script>
+document.addEventListener('livewire:initialized', () => {
+    const counter = new CountUp('points-counter', {{ $totalPointsEarned }});
+    Livewire.on('updatePoints', (totalPoints) => {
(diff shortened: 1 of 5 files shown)


resources/views/community/components/global-statistics/total-points-earned-javascript.blade.php
```diff
@@ -0,0 +1,15 @@
+{{--
+    This stuff shouldn't be in the Livewire component because
+    the template is sent to the browser on each re-render.
+--}}
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/countup.js/1.7.0/countUp.min.js"></script>
```
wescopeland, line 6: This is <2KB.

Jamiras:
I'm going to defer judgment on this.

I've verified the functionality, but I don't think we should have the main page making a server call every second and a half to update a value that's neat to look at once-in-a-while, but not something I want to actively watch change.

At least it's smart enough to not call the server when it's offscreen. Unfortunately, it's onscreen at the same time as the Recent Players widget, which is what I imagine most people would be looking at if they're just sitting at the main page.

wescopeland:
Though it was fun to build, I'm not entirely convinced this is desirable either. Maybe on some smaller-scale pages such as the developer feed it would be more appropriate.

PR #2163, opened 2024-01-16
fix(tooltip): fill small gap at the rounded corners of tooltips

drisc:
There is a small gap at the corners of the tooltip, this PR adds a background to the border so that the gap is filled by the borders current themed colour.


resources/js/alpine/tooltipComponent/utils/renderTooltip.ts
```diff
@@ -52,7 +52,7 @@ export function renderTooltip(
   );
 
   if (!options?.isBorderless) {
-    store.tooltipEl.classList.add('border', 'border-embed-highlight');
+    store.tooltipEl.classList.add('border', 'border-embed-highlight', 'bg-embed-highlight');
```
wescopeland, line 55: Does this make the `border` and `border-embed-highlight` classnames redundant?
In other words, could they be removed?

drisc: Did a test and yes they can be removed. I have noticed that the option `isBorderless` doesn't seem to do anything though. Setting it to true or false makes no difference.

wescopeland: You will see it take effect on any card which has a loading indicator, such as game tooltips.

drisc: I see the change now, my expectation was that it would effect all tooltips not just the loading spinner.

wescopeland (approved):

Jamiras (approved):
I don't really see a "gap":
image

But the border does appear to have changed thickness:
image

drisc:
The gap is more apparent when it overlaps with a bright color (and may be easier to see on HiDPI screens):
opencorners

The fix also stops some tooltips being double bordered which is why they look thinner.

wescopeland:
> (and may be easier to see on HiDPI screens):

Indeed - I am not able to see the gap on my desktop display, but I am able to see it on my laptop display which has a much higher PPI.

PR #2165, opened 2024-01-19
feat(manage): add Filament admin panel & permission matrix preparations

luchaos:
Add basic management features for users, systems, and achievements.

Introduce permission matrix without affecting existing permissions (yet, will follow as soon as new roles have been assigned).

Introduce audit logging to keep track of model/role changes.

**Seed roles and test users**

```sql
SET FOREIGN_KEY_CHECKS = 0;
TRUNCATE TABLE auth_model_roles;
TRUNCATE TABLE auth_roles;
SET FOREIGN_KEY_CHECKS = 1;
```

```shell
php artisan db:seed
```

**Assign administrator role by user ID to get started**

```sql
INSERT INTO auth_model_roles(role_id, model_type, model_id) VALUES(2, 'user', <USER_ID>);
```

As administrator you can go to user management and assign additional roles. Update `storage/settings.json` according to `storage/settings.dist.json` and/or go to `/manage ´` directly.

Each role has a seeded test user. Password is the same as the username (e.g. `HubManager:HubManager`).

## Why

To finally get the role-based, non-linear permission matrix in place and to allow for better management features without building everything from scratch. I chose [Filament](https://filamentphp.com/) as the admin panel as it comes with a plethora of features and I had very positive experiences with it in the past. Not a single line of HTML, JavaScript, or CSS is needed to build panels/features.

Upcoming PRs:
- All remaining references to legacy permissions will be replaced with policy checks
- Admin/Dev dropdowns on pages will be replace with Filament panels
- Manage dropdown features (news, admin tools, ...) will all move to Filament eventually
- More resources have been prepared and will be added in separate PRs (all the way to event management)


app/Community/Policies/TicketPolicy.php
```diff
@@ -0,0 +1,52 @@
...
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class TicketPolicy
+{
+    use HandlesAuthorization;
+
+    public function manage(User $user): bool
+    {
+        return $user->hasAnyRole([
+            Role::DEVELOPER,
```
Jamiras, line 19: What functionality is "manage" associated to. It seems like other policies use it for moderation. Does this need to be/include `DEVELOPER_STAFF`?

app/Community/Policies/TriggerTicketPolicy.php
```diff
@@ -15,7 +16,13 @@ class TriggerTicketPolicy
...
     public function manage(User $user): bool
     {
-        return $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper;
+        return $user->hasAnyRole([
+            Role::HUB_MANAGER,
+            Role::DEVELOPER_STAFF,
+            Role::DEVELOPER,
+            Role::DEVELOPER_JUNIOR,
+        ])
+            || $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper;
```
Jamiras, line 25: What's the difference between TicketPolicy and TriggerTicketPolicy?

lang/en/permission.php
```diff
@@ -5,31 +5,39 @@
...
+
+        Role::FOUNDER => __('Founder'),
         Role::ARCHITECT => __('Architect'),
         Role::ENGINEER => __('Engineer'),
-        Role::FOUNDER => __('Founder'),
+        Role::BETA => __('Beta'),
+
+        // vanity roles assigned by admins
+
+        Role::DEVELOPER_VETERAN => __('Veteran'),
```
Jamiras, line 41: Should this label be "Veteran Developer"?

config/roles.php
```diff
@@ -4,237 +4,171 @@
...
-    Role::RELEASE_MANAGER,
     Role::ADMINISTRATOR,
+    Role::ARCHITECT,
     Role::BETA,
+    Role::ENGINEER,
+    Role::FOUNDER,
+    Role::RELEASE_MANAGER,
 ];
 
 $adminAssignable = [
```
Jamiras, line 15: As an admin, assigning a valid role to another admin (like Developer) causes the other admin to lose their admin role.

Jamiras (commented):
Two general questions, since I can't figure out where the related code lives:
1) I see audit_log entries being written. How do I view them?
2) When clicking on a user from the user list, it automatically takes me to the edit page. Why does the edit page have a link to the same edit page?

app/Community/Policies/TicketPolicy.php
luchaos: `manage` ability checks are used for all Filament resources and in blade views to check for general visibility.
Updated the TicketPolicy to match TriggerTicketPolicy, it's not in use yet.

app/Community/Policies/TriggerTicketPolicy.php
luchaos: Similar to `Comment` vs `AchievementComment`, `TriggerTicket` is a more specific form of `Ticket`. tickets are meant to be migrated to be polymorphic, where tickets can not only be written for triggers (achievements, leaderboards) but for other resources, too.

lang/en/permission.php
luchaos: good call, updated

config/roles.php
luchaos: fixed in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/7790b8f44921b44e78253b72ab0acd78e2079af8 - this was not quite finished yet

luchaos:
> Two general questions, since I can't figure out where the related code lives:
> 
> 1. I see audit_log entries being written. How do I view them?
> 2. When clicking on a user from the user list, it automatically takes me to the edit page. Why does the edit page have a link to the same edit page?

Audit logs are visible for users now (systems will follow in another commit, achievements will have their own versioning system and won't rely on audit logs).

The sub-navigation has been improved for users, systems and achievements will follow tomorrow.

config/roles.php
Jamiras: As a root user, I added admin to another user and was no longer able to add or remove roles from that user. As such, I'm unable to test the original problem.

I see the same behavior as a non-root admin (cannot modify another admin), but I'd expect root to be able to.

I would also expect a non-root admin be able to add/remove lesser roles on another admin.

app/Filament/Resources/UserResource/Pages/EditUser.php
```diff
@@ -0,0 +1,33 @@
...
+        $record = $this->getRecord();
+
+        return '[' . $record->ID . '] ' . $record->User;
+    }
+
+    protected function getHeaderActions(): array
+    {
+        return [
+        ];
+    }
```
Jamiras, line 32: Changes to the Legacy permissions level are not captured in the audit log, and do not trigger the related functionality (i.e. changing the value to Banned does not clear out the authentication fields [Password, AppToken, etc]).

app/Filament/Resources/UserResource/Pages/EditUser.php
```diff
@@ -0,0 +1,33 @@
...
+}
```
Jamiras, line 33: Any user with the Administrator role can change Legacy permissions to any level (including Moderator). The old logic restricts things such that the new value must be lower than the logged-in user's current value.

app/Filament/Resources/UserResource/Pages/ViewUser.php
```diff
@@ -0,0 +1,35 @@
...
+use Filament\Resources\Pages\ViewRecord;
+
+class ViewUser extends ViewRecord
+{
+    protected static string $resource = UserResource::class;
+
+    public static function getNavigationLabel(): string
+    {
+        return __('Details');
+    }
```
Jamiras, line 19: Would it be possible to get an explicit "None" to show up if the user has no roles?
image

app/Filament/Resources/AchievementResource/Pages/EditAchievement.php
```diff
@@ -0,0 +1,36 @@
...
+        return __('Edit');
+    }
+
+    public function getSubheading(): ?string
+    {
+        /** @var Achievement $record */
+        $record = $this->getRecord();
+
+        return '[' . $record->ID . '] ' . $record->Title;
+    }
```
Jamiras, line 27: Editing an achievement's published state/points does not trigger the aggregate data update jobs.

Jamiras: Neither does changing the associated game.

app/Filament/Resources/SystemResource/Pages/ListSystems.php
```diff
@@ -0,0 +1,21 @@
...
+class ListSystems extends ListRecords
+{
+    protected static string $resource = SystemResource::class;
+
+    protected function getHeaderActions(): array
+    {
+        return [
+            Actions\CreateAction::make(),
+        ];
+    }
```
Jamiras, line 20: Is there a way to sort the systems list by name? Would it be reasonable to make that the default sort?

public/assets/images/system/super-cassette-vision.png
wescopeland, line 1: Should this be removed?

luchaos: removed in latest

app/Filament/Resources/UserResource/Pages/ViewUser.php
luchaos: hidden if empty in latest

app/Filament/Resources/UserResource/Pages/EditUser.php
```diff
@@ -0,0 +1,33 @@
...
+        /** @var User $record */
+        $record = $this->getRecord();
+
+        return '[' . $record->ID . '] ' . $record->User;
+    }
+
+    protected function getHeaderActions(): array
+    {
+        return [
+        ];
```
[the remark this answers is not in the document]
luchaos: asterisk usually implies that it's a required field

app/Filament/Resources/RoleResource/Pages/ViewRole.php
```diff
@@ -0,0 +1,34 @@
...
+
+        return $record->name;
+    }
+
+    public function getHeaderActions(): array
+    {
+        return [
+            // TODO Actions\EditAction::make(),
+        ];
+    }
```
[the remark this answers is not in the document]
luchaos: Both order and whether it should be hidden (when 0).
Not used yet but will inform the presentation on the profile page

app/Filament/Resources/RoleResource/Pages/ViewRole.php
```diff
@@ -0,0 +1,34 @@
...
+        return __('Details');
+    }
+
+    public function getSubheading(): ?string
+    {
+        /** @var Role $record */
+        $record = $this->getRecord();
+
+        return $record->name;
+    }
```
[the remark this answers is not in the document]
luchaos: Good call, added avatars to the list in latest.
It's possible to reuse a resource's table definition so it looks exactly the same as on a resource's listing without having to define all the columns again.
Unfortately there's a bug with filament table actions in reused tables that has to be fixed first.

app/Filament/Resources/UserResource/Pages/EditUser.php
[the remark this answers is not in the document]
luchaos: https://filamentphp.com/docs/3.x/forms/fields/getting-started#marking-a-field-as-required

app/Filament/Resources/SystemResource/Pages/ListSystems.php
luchaos: Sortable by ID and full name (default) in latest

app/Filament/Resources/UserResource/Pages/EditUser.php
luchaos: good point - maybe making it read-only in the admin panels should be enough?
the plan is to migrate away from those very soon.

luchaos: Removed `Permissions` form input and removed it from activity logs again. Just remembered that it wasn't included in the list of logged fields as it won't be around for very long.

app/Filament/Resources/AchievementResource/Pages/EditAchievement.php
luchaos: good call, adjusted in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/075fc40c293c2d729898e34a5f80828a4918be93

Introduced a new `AchievementMoved` event which dispatches game metrics updates for both the new and the original games.

app/Filament/Resources/UserResource/Pages/EditUser.php
luchaos: Removed input, see https://github.com/RetroAchievements/RAWeb/pull/2165#discussion_r1465694276

Banning users will be a new action in the admin panels which does not rely on `Permissions` anymore but on a `banned_at` timestamp which will behave similarly to the soft-delete `deleted_at` timestamp (global eloquent model scope will be introduced for that).

app/Helpers/database/achievement.php
```diff
@@ -4,11 +4,6 @@
 use App\Platform\Enums\AchievementFlag;
 use App\Platform\Enums\AchievementPoints;
 use App\Platform\Enums\AchievementType;
-use App\Platform\Events\AchievementCreated;
-use App\Platform\Events\AchievementPointsChanged;
-use App\Platform\Events\AchievementPublished;
-use App\Platform\Events\AchievementTypeChanged;
-use App\Platform\Events\AchievementUnpublished;
```
luchaos, line 11: These events do not have to be dispatched explicitly anymore - moved to model events which guarantee consistent behaviour regardless of where models were created/changed: https://github.com/RetroAchievements/RAWeb/pull/2165/files#diff-af5d573c5bcfbd0ed76895c2569ebc277ce30f5b862d9d6e2cbe8b1f6102c014R111-R142

app/Helpers/database/achievement.php
```diff
@@ -416,18 +399,10 @@ function updateAchievementFlag(int|string|array $achID, int $newFlag): void
...
     foreach ($updatedAchievements as $achievement) {
-        if ($newFlag === AchievementFlag::OfficialCore) {
-            AchievementPublished::dispatch($achievement);
-        }
-
-        if ($newFlag === AchievementFlag::Unofficial) {
-            AchievementUnpublished::dispatch($achievement);
-        }
+        $achievement->Flags = $newFlag;
+        $achievement->save();
```
luchaos, line 405: Saving model individually will trigger model events. Batch updates like `$achievements->update(['Flags' => $newFlag]);` previously do not trigger model events.

app/Helpers/database/achievement.php
luchaos: I'll add accompanying tests for these changes

luchaos: added tests in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/ebcb8c31b7b1ca6a1314591bc09bc17fc86845ae

app/Filament/Resources/AchievementResource/Pages/EditAchievement.php
Jamiras: I see this spawning a job, but it fails:
```
  2024-01-26 15:50:26 App\Platform\Listeners\DispatchUpdateGameMetricsJob  RUNNING
  2024-01-26 15:50:26 App\Platform\Listeners\DispatchUpdateGameMetricsJob  5.54ms FAIL
```

Looks like you have a typo:
```
Undefined property: App\\Platform\\Events\\AchievementMoved::$origianlGame at /var/www/html/app/Platform/Listeners/DispatchUpdateGameMetricsJob.php:49)
```

config/roles.php
Jamiras: I'm still seeing this behavior.

A Root user can add the Administrator role to another user, but once that has occurred, the root user cannot modify that user at all. The Administrator role cannot be removed, and no other roles can be added or removed.

Similarly, a non-Root Administrator cannot add or remove any roles from another non-Root Administrator.

storage/app/settings.dist.json
```diff
@@ -2,6 +2,11 @@
     "sync": false,
     "system.alert": false,
     "tools": [
+        {
+            "label": "Manage",
+            "route": "filament.admin.pages.dashboard",
+            "abilities": "accessManagementTools"
+        },
```
Jamiras, line 9: I've noticed that my menu no longer contains any items except this one. I don't see any reason why this feature should prevent access to Logs/Horizon Queues/Mailpit/etc.
image

fwiw, it's blank if I don't make this change, so it's not because of this change.
image

luchaos: forgot to add those to my todo list, thanks for the reminder! everything was visible to moderators previously because of the Gate::before hook ehich has been removed.

I'd like to make those root only, what do you think? that's the role sysadmins should have

app/Filament/Resources/AchievementResource/Pages/EditAchievement.php
luchaos: fixed in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/cadd52475b5f643089765989e370895af6052f2c

storage/app/settings.dist.json
Jamiras: Maybe Architect or Engineer (not exactly sure what those imply). 

Logs and Horizon Queue are definitely geared towards the technical people, but that doesn't mean we should limit it to Root. Maybe Release Manager?

Sentry and Stats could probably be open to any Administrator, maybe even Release Manager.

Metabase should be available to Administrators and Event Managers.

config/roles.php
luchaos: fix root role privileges fro admins in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/a5a889fc17518e487cee500ea213fa6e85fdcd5c

non-Root Administrator to non-Root Administrator should not be necessary as admins can edit their own roles at any time. however, slightly improved in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/e1760a66f351ea1a543038a48142d4e9a6d5a495

storage/app/settings.dist.json
luchaos: adjusted in https://github.com/RetroAchievements/RAWeb/pull/2165/commits/7c17198384c34d88478df04fc05e61c7636e5475

please change `viewLogs` to `viewLogViewer` in `storage/app/settings.json` according to the `settings.dist.json` file.

luchaos: `Architect` and `Engineer` are vanity roles and have no permissions attached.

`Release Manager` is meant to be focused on client/integration releases.

Sentry and Stats are open to admins now.

I wouldn't want to give Event Managers access to Metabase, get rid of metabase rather and build the tools needed into Filament. Metabase is becoming a liability, only meant to give tools into some people's hands until purpose-built tools are available. Too many raw queries in there...

Jamiras: I think I'm still hung up on what an Administrator is. In the past, they had access to everything. It would make sense to limit these tools to that Role assuming that's the scope of the Role.

My mind would probably connect things better if the role were renamed to System Administrator or something similar.

I do see the menu items returned, but they don't seem to be affected by roles yet. I assumed they would given the fact that you had me rename `viewLogs` to `viewLogViewer`. I'm fine if you want to do that in a later PR.

luchaos: Those are all good points. It's loosely inspired by the [original permission matrix](https://docs.google.com/spreadsheets/d/17YROKCpqinnyYG3cE8Q1mLDxzOAARCFzvjGltuADZb0/edit?usp=sharing) - however, I'm not sure all of it is applicable.

We should probably discuss this with the team to figure out where our roles should be heading at.

`Root` is basically `System Administrator`. Not sure what to do about the `Administrator` role at this point. I initially planned to get rid of it altogether. That's from a time where the admin role in this community had a very bad reputation. So instead of making it a catch-all role it'd be split up into dedicated task-focused roles. Hence all the manager roles.

> I do see the menu items returned, but they don't seem to be affected by roles yet. I assumed they would given the fact that you had me rename viewLogs to viewLogViewer. I'm fine if you want to do that in a later PR.

`viewLogViewer` is the gate that the log viewer package expects itself - it's checking for the `tool` ability which is currently bound to root and admin. I'm not sure I understand correctly - in which way are they not affected by roles yet? If you mean applying role checks on individual pages, those will follow in another PR which will reduce/get rid of the usage of `UserAccounts.Permissions`.

Jamiras: When my account has no roles:
```
MariaDB [retroachievements-web]> select * from auth_model_roles where model_id=43495;
Empty set (0.001 sec)
```
I still have the full menu:
image

I have to drop my non-role Permissions all the way down to 1 to make them disappear - at which point the entire Manage button disappears.

luchaos: oh! that's because of the local environment check. it always displays tools locally.

luchaos: note that the legacy Permissions still is in effect, too. this makes it hard to reason about the actual permissions in some cases. a follow-up PR will hopefully resolve most of that.

Jamiras (approved):

PR #2166, opened 2024-01-20
add additional leaderboard formats

Jamiras:
Allows the use of new formats defined in https://github.com/RetroAchievements/rcheevos/pull/305 in leaderboards.

Consolidates labels and formatting functions into helpers on enum.


wescopeland (approved):
Verified locally. The tests are excellent and I have no concerns with the implementation.

PR #2167, opened 2024-01-20
refactor(user): migrate Recent Progress chart to blade

wescopeland:
Migrates the user profile Recent Progress chart to Blade. Chart colors have been changed to mirror the distribution chart on game pages (gold for hardcore, gray for softcore). Other than that, implementation details are unchanged.

**Before**
Screenshot 2024-01-20 at 5 41 22 PM

**After**
Screenshot 2024-01-20 at 5 41 14 PM

### DIFF
--- a/resources/views/community/components/user/recent-progress.blade.php
+++ b/resources/views/community/components/user/recent-progress.blade.php
@@ -1,0 +1,85 @@
+@props([
+    'username' => '',
+    'userScoreData' => [],
+])
+
+<?php
+use Illuminate\Support\Carbon;
+?>
+
+<script defer src="https://www.gstatic.com/charts/loader.js"></script>
+<script>
+document.addEventListener('DOMContentLoaded', function() {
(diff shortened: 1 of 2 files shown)


resources/views/community/components/user/recent-progress.blade.php
```diff
@@ -0,0 +1,85 @@
+@props([
+    'username' => '',
+    'userScoreData' => [],
+])
+
+<?php
+use Illuminate\Support\Carbon;
+?>
+
```
Jamiras, line 9: Now might be a good time to inject a "missing data" statement if there is no recent score data.

image

instead of 

image

wescopeland: Good idea, this also keeps us from unnecessarily loading all that JS.

I've gone with "No points earned in the last 14 days." as the label.

Jamiras (approved):

PR #2172, opened 2024-01-22
feat(systems): clean up system names & images

luchaos:
Menu, config, and database entries for systems should be in line and not use more than two sources for names. Eventually system names will all be sourced from the database and editable via the upcoming admin panel.

Run `(sail) php artisan db:seed` to update you `Console` table.


config/systems.php
```diff
@@ -566,22 +567,22 @@
...
+        'name_full' => 'Events',
         'name_short' => 'Events',
         'manufacturer' => null,
         'active' => true,
     ],
     102 => [
         'name' => 'Standalone',
         'name_full' => 'Standalone',
-        'name_short' => 'N/A',
+        'name_short' => 'Standalone',
```
wescopeland, line 585: `name_short` is what appears in the Last N Games Played components on the user profile as well as the completion progress page. This is quite long - can we please revert this or change it to something else?

luchaos: `Standalone` is its shortest form yet represents its full name properly.
Presentation has to follow imo.
We have `Super Cassette Vision` as short name in there as well.
What do you suggest?

luchaos: Maybe `EXE` as that's what's on the icon?

luchaos: CV and SCV for (super) casette vision sound good to me, i'll adjust the config accordingly.
I'd favor EXE instead of N/A personally to make sure each system has an identifiable name. I had something different in mind for presentation when introducing short names, i can see how super short names are preferrable in the current uses though

wescopeland: Sounds good, EXE seems like a better short name than N/A to me and I agree it's much more identifiable.

config/systems.php
```diff
@@ -425,28 +425,29 @@
...
-        'name' => 'N-Gage',
+        'name' => 'Nokia N-Gage',
         'name_full' => 'Nokia N-Gage',
         'name_short' => 'N-Gage',
         'manufacturer' => 'Nokia',
         'active' => false,
     ],
     62 => [
-        'name' => '3DS',
+        // 'name' => '3DS',
```
Jamiras, line 435: leftover comment should be removed

config/systems.php
luchaos: updated to EXE, ECV (Epoch Casette Vision), and eSCV (Epoch Super Casette Vision) in https://github.com/RetroAchievements/RAWeb/pull/2172/commits/61bfa0c59c2ef772b337aea75a0f93a64eda9a99

config/systems.php
```diff
@@ -47,15 +47,15 @@
...
+        'name' => 'NES/Famicom',
+        'name_full' => 'Nintendo Entertainment System/Famicom',
         'name_short' => 'NES',
         'manufacturer' => 'Nintendo',
         'active' => true,
     ],
     8 => [
-        'name' => 'PC Engine / TurboGrafx',
-        'name_full' => 'NEC PC Engine / TurboGrafx',
+        'name' => 'PC Engine/TurboGrafx16',
```
Jamiras, line 57: There should be a hyphen in `TurboGrafx-16`: https://en.wikipedia.org/wiki/TurboGrafx-16

config/systems.php
luchaos: oh, `eSCV` is the emulator. i'll change it to `ESCV`.

config/systems.php
luchaos: thanks! https://github.com/RetroAchievements/RAWeb/pull/2172/commits/cdd2d9f383eae7b445b317e8503764b1ad86845e

config/systems.php
luchaos: good catch: [TurboGrafx-16](https://github.com/RetroAchievements/RAWeb/pull/2172/commits/cdd2d9f383eae7b445b317e8503764b1ad86845e)

Jamiras (approved):

PR #2174, opened 2024-01-25
fix(award-badge): remove un-needed flex and justify

drisc:
While working on solutions for the awards grid I set flex and justify-center on a couple of divs in the RenderAward function. They are no longer needs and cause alignment issues on pages that use the function such as the reordering page so they have now been removed.


wescopeland:
> They are no longer needs and cause alignment issues on pages that use the function such as the reordering page so they have now been removed.

For documentation purposes, would you mind sharing a before & after in this PR?

drisc:
Because the award icon becomes a flex object with justification it centers itself in any container that it is inside of.

I noticed that the badges were no longer left aligned on the reorder page while checking something.
Before:
image

After:
image

Jamiras (approved):
I see the reported before/after behavior (thanks for pictures). I don't understand what changed to cause the issue though.

drisc:
While testing methods to center the award at different breakpoints I added flex and justify-center to the award rendering function itself, this caused them to be applied wherever it is used. I made the change and didn't realise untile testing something else.

wescopeland (approved):

PR #2175, opened 2024-01-26
feat: redesign user profiles

### LINKED ISSUE #1430: Claims on Developer Profile wrap wrong on long Game Titles
**Describe the bug**
It seems that if the title of a game entry is very long the wrapping on the User Profile behaves a bit weird. As seen in the screenshot the site make a full new line just for the `*` which is weird considering that it normally should just be behind the console name.

**To Reproduce**
1. Claim a game with a very long name
2. Visit the user profile
3. See error

**Expected behavior**
The `*` should be directly behind the console name as seen on the last claim for the Game Boy Color game.

**Screenshots**
image

 - Browser: Opera GX (95.0.4635.74)

wescopeland:
This PR introduces a massive facelift to the primary container of the user profile.

**Why?**
* Currently, much of this primary container is a wall of text that does not present well on mobile.
* With incoming _Final Fantasy XI_ support, the user profile will leave a lasting first impression on an influx of new users.
* With the user profile's current data presentation strategy, it is difficult to enhance or change the profile.

No stats have been removed from the user profile. Several new stats have been added, which is now afforded by the improved spatial efficiency of the page.

### Use Case 1: New Users
When a user first creates an account, they receive an empty profile.

Screenshot 2024-01-25 at 7 36 45 PM

Even in this minimal state, we can preview most of the new components.

The top meta closely mirrors the UX of the rewritten user cards introduced in https://github.com/RetroAchievements/RAWeb/pull/1642.

Screenshot 2024-01-25 at 7 39 20 PM

The social interaction controls have received a revamp:

Screenshot 2024-01-25 at 7 39 47 PM

Though empty, we can also see the new "stat panels":

Screenshot 2024-01-25 at 7 40 22 PM

Secondary stats are buried under a "see more" button:

https://github.com/RetroAchievements/RAWeb/assets/3984985/11600714-3f53-4d4f-9993-c9a5a92e870a

### Use Case 2: Softcore-only Players

If a player prefers to play on softcore mode and has no hardcore progress, all the stats on their profile will flip to showing softcore as the more prominent mode.

Screenshot 2024-01-25 at 7 44 07 PM

This adjustment includes the top meta on the profile, which also flips to softcore (identical to the user tooltip card behavior).

Screenshot 2024-01-25 at 7 46 34 PM

### Use Case 3: Hardcore-only Players

If a player prefers to play on hardcore mode and has no softcore progress, all the stats on their profile will show hardcore as the more prominent mode. This is also the default mode until the user unlocks their first achievement.

Screenshot 2024-01-25 at 7 49 23 PM

Note that there are several new stats available in the "Player stats" section:

Screenshot 2024-01-25 at 7 50 01 PM

"Started games beaten" represents any game with at least 1 achievement unlocked (either mode) and has any kind of associated award.
"Points earned in the last X days" is also irrespective of mode. It is derived from whichever mode the user has more progress in.

### Use Case 4: Mixed-progress Players

If a player has both hardcore and softcore progress, both modes are displayed, with the primary mode being whichever one the user has more points in (identical to the user card).

Screenshot 2024-01-25 at 7 55 00 PM

### Use Case 5: Developers

Users with development permissions (JrDev, Dev, and Moderator) or past development contributions have another panel appear on their profile: "Developer stats".

Screenshot 2024-01-25 at 7 52 27 PM

This panel includes a few new or hard-to-find stats and shows all of the user's claims. Additionally, this resolves #1430.

### Use Case 6: Mobile

The profile is now mobile-first.

Screenshot 2024-01-25 at 7 56 38 PM

---

Other situations which are handled properly:
* Untracked players
* Players that follow the currently logged in user

---

Feedback for this UX was collected from multiple users across 4 different teams.

ref:
https://discord.com/channels/476211979464343552/1152683553348407409/1200146454333886534
https://discord.com/channels/310192285306454017/511718348178849792/1200170876230967307


database/migrations/platform/2024_01_24_000000_update_codenotes_table.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('CodeNotes', function (Blueprint $table) {
+            $table->index('AuthorID');
```
luchaos, line 13: To match [upcoming schema changes](https://github.com/RetroAchievements/RAWeb/blob/master/app/Platform/Models/MemoryNote.php#L16-L22) ahead of time:
```php
$table->index('AuthorID', 'memory_notes_user_id_index');
```

database/migrations/platform/2024_01_24_000000_update_codenotes_table.php
```diff
@@ -0,0 +1,23 @@
...
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('CodeNotes', function (Blueprint $table) {
+            $table->dropIndex('AuthorID');
```
luchaos, line 20:
```php
$table->dropIndex('memory_notes_user_id_index');
```

resources/views/community/components/user/profile/developer-claims-list.blade.php
```diff
@@ -0,0 +1,29 @@
...
+    'numAllowedClaims' => null, // ?int
+])
+
+<div>
+    <p class="text-2xs mb-1">
+        <span class="font-bold">{{ $label }}</span>
+
+        @if ($numAllowedClaims !== null && count($claims) <= $numAllowedClaims)
+            ({{ count($claims) }} of {{ $numAllowedClaims }} allowed)
+        @endif
```
Jamiras, line 13: why is this restricted when `claims` exceed `numAllowedClaims`? Wouldn't it be better to show "2 of 1 allowed" instead of not showing anything?

resources/views/community/components/user/profile/mode-player-stat-cells.blade.php
```diff
@@ -0,0 +1,95 @@
...
+        @else
+            <a href="{{ '/globalRanking.php?t=2&o=' . $softcoreRankMeta['rankOffset'] . '&s=2' }}">
+                #{{ localized_number($softcoreRankMeta['rank']) }}
+            </a>
+            of {{ localized_number($softcoreRankMeta['numRankedUsers']) }}
+        @endif
+    </x-user.profile.stat-element>
+
+    <x-user.profile.stat-element label="Achievements unlocked (softcore)">
+        <span class="{{ $totalSoftcoreAchievements > 0 ? 'font-bold' : 'text-muted' }}">
```
Jamiras, line 85: Should these be italic to match the text-muted rank? (applies to all muted 0s)
image

resources/views/community/components/user/profile/developer-stats.blade.php
```diff
@@ -0,0 +1,120 @@
...
+                @else
+                    <span>0</span>
+                @endif
+            </x-user.profile.stat-element>
+        </div>
+
+        <div class="flex flex-col gap-y-1">
+            <x-user.profile.stat-element label="Code notes created">
+                @if ($developerStats['totalAuthoredCodeNotes'] > 0)
+                    <a href="{{ '/individualdevstats.php?u=' . $username }}" class="font-bold">
```
Jamiras, line 65: Should we add an anchor to the code notes section of the dev stats page for this link?

resources/views/community/components/user/profile/last-seen-in.blade.php
```diff
@@ -0,0 +1,62 @@
...
+
+@if ($sessionGame)
+    <div class="mb-6">
+        <div class="flex w-full items-center gap-x-1.5">
+            <p role="heading" aria-level="2" class="mb-0.5 text-2xs font-bold">
+                Most recently played
+
+                @if ($mostRecentSession?->rich_presence_updated_at)
+                    <p class="smalldate min-w-auto">
+                        {{ Carbon::parse($mostRecentSession->rich_presence_updated_at)->diffForHumans() }}
```
Jamiras, line 36: This should be given a tooltip with the raw date like we do for forum posts.

resources/views/community/components/user/profile/last-seen-in.blade.php
```diff
@@ -0,0 +1,62 @@
...
+    ?? $userMassData['RichPresenceMsg']
+    ?? null
+);
+?>
+
+@if ($sessionGame)
+    <div class="mb-6">
+        <div class="flex w-full items-center gap-x-1.5">
+            <p role="heading" aria-level="2" class="mb-0.5 text-2xs font-bold">
+                Most recently played
```
Jamiras, line 32: the timestamp is slightly misaligned with this text
image

app/Community/Components/UserProfileMeta.php
```diff
@@ -0,0 +1,227 @@
...
+            ->get();
+
+        $pointsLast30Days = 0;
+        $pointsLast7Days = 0;
+
+        $now = now();
+
+        foreach ($achievements as $playerAchievement) {
+            $achievementDate = $playerAchievement->{$dateColumn};
+            $daysAgo = $now->diffInDays($achievementDate, false);
```
Jamiras, line 215: This logic is backwards. If I print "$daysAgo", it's always negative.

The problem is fairly obvious with our test data as every user appears to have been active in the last week.

app/Community/Components/UserProfileMeta.php
```diff
@@ -0,0 +1,227 @@
...
+
+            if ($daysAgo <= 30) {
+                $pointsLast30Days += $playerAchievement->achievement->points;
+            }
+            if ($daysAgo <= 7) {
+                $pointsLast7Days += $playerAchievement->achievement->points;
+            }
```
Jamiras, line 222: Since the query only returns entries for the last 30 days, this could be simplified:
```
if ($daysAgo <= 7) {
    $pointsLast7Days += $playerAchievement->achievement->points;
} else {
    $pointsLast30Days += $playerAchievement->achievement->points;
}
```

resources/views/community/components/user/profile/primary-meta.blade.php
```diff
@@ -0,0 +1,81 @@
...
+                :username="$username"
+            />
+
+            {{-- Last Activity --}}
+            <p>
+                <span class="font-bold">Last Activity:</span>
+                <span class="cursor-help" title="{{ getNiceDate(strtotime($userMassData['LastActivity'])) }}">
+                    {{ Carbon::parse($userMassData['LastActivity'])->diffForHumans() }}
+                </span>
+            </p>
```
Jamiras, line 70: The [old implementation](https://github.com/RetroAchievements/RAWeb/pull/2175/files#diff-98777a5d0a4ac46a7efe8cada6437732a761b079f5b2f8d6221cf473b65cf7deL205) hid this field if a user didn't have any activity yet (see profile for newest created user)

resources/views/community/components/user/profile-meta.blade.php
```diff
@@ -0,0 +1,68 @@
...
+        :userMassData="$userMassData"
+        :username="$username"
+    />
+
+    <div class="mb-2 sm:ml-[142px] md:ml-0 md:absolute md:top-0.5 md:right-0 lg:relative lg:ml-[142px] xl:absolute xl:ml-0">
+        <x-user.profile.social-interactivity :username="$username" />
+    </div>
+</div>
+
+<hr class="border-embed-highlight mb-2" />
```
Jamiras, line 36: This `hr` is nearly imperceptable.

image

Contrast greatly increased in highlighted region.

app/Community/Components/UserProfileMeta.php
```diff
@@ -0,0 +1,227 @@
...
+        // Format and return the result to 2 decimal places.
+        return number_format($averageFinishedGames, 2, '.', '');
+    }
+
+    private function calculateAveragePointsPerWeek(string $userMemberSince, int $points = 0): int
+    {
+        $memberSince = Carbon::createFromFormat('Y-m-d H:i:s', $userMemberSince);
+        $now = Carbon::now();
+
+        $weeksOfMembership = $memberSince->diffInWeeks($now);
```
Jamiras, line 185: Should this use `$memberSince->diffInWeeks($user->lastLogin)`? 

If a user hasn't logged in for many years, should their average points per week reflect that? 

If they did come back, it would obviously tank the value.

app/Community/Components/UserProfileMeta.php
```diff
@@ -0,0 +1,227 @@
...
+            ->select(DB::raw('COUNT(LeaderboardDef.ID) AS TotalAuthoredLeaderboards'))
+            ->value('TotalAuthoredLeaderboards');
+
+        $totalAuthoredCodeNotes = $user->authoredCodeNotes()->where('Note', '!=', '')->count();
+
+        $openTickets = null;
+        // FIXME: Uses legacy roles.
+        if ($user->getAttribute('Permissions') >= Permissions::Registered) {
+            $openTickets = array_sum(countOpenTicketsByDev($user->User));
+        }
```
Jamiras, line 130: banned users can have open tickets. see Euclide. Suggest basing this on "ContribCount" instead.

resources/views/community/components/user/profile-meta.blade.php
```diff
@@ -0,0 +1,68 @@
...
+    <x-user.profile.primary-meta
+        :hardcoreRankMeta="$hardcoreRankMeta"
+        :softcoreRankMeta="$softcoreRankMeta"
+        :userMassData="$userMassData"
+        :username="$username"
+    />
+
+    <div class="mb-2 sm:ml-[142px] md:ml-0 md:absolute md:top-0.5 md:right-0 lg:relative lg:ml-[142px] xl:absolute xl:ml-0">
+        <x-user.profile.social-interactivity :username="$username" />
+    </div>
```
Jamiras, line 33: This widget should be hidden for users with permissions <= 0. (Improvement over past behavior).

Jamiras: It may be desirable to add a Banned/Spam indicator instead as that information is no longer visible in the title card.

Before:
image

app/Community/Components/UserProfileMeta.php
```diff
@@ -0,0 +1,227 @@
...
+
+        $ticketsClosedResolved = getNumberOfTicketsClosedForOthers($user->User);
+        $ticketsResolvedForOthers = 0;
+        foreach ($ticketsClosedResolved as $ticketData) {
+            $ticketsResolvedForOthers += $ticketData['ResolvedCount'];
+        }
```
Jamiras, line 136: These don't appear to be used. Recommend avoiding the calculation.

app/Platform/Concerns/ActsAsDeveloper.php
```diff
@@ -26,6 +27,14 @@ public function authoredAchievements(): HasMany
         return $this->hasMany(Achievement::class, 'Author', 'User');
     }
 
+    /**
+     * @return HasMany<MemoryNote>
+     */
+    public function authoredCodeNotes(): HasMany
+    {
+        return $this->hasMany(MemoryNote::class, 'AuthorID', 'ID');
```
Jamiras, line 35: I'd probably put the "where not empty" here, as empty notes represent deleted notes.

database/migrations/platform/2024_01_24_000000_update_codenotes_table.php
Jamiras: Is this even necessary? I didn't run the migrate step and the page seems to be working.

resources/views/community/components/user/profile-meta.blade.php
```diff
@@ -0,0 +1,68 @@
...
+    :averagePointsPerWeek="$averagePointsPerWeek"
+    :hardcoreRankMeta="$hardcoreRankMeta"
+    :recentPointsEarned="$recentPointsEarned"
+    :softcoreRankMeta="$softcoreRankMeta"
+    :totalHardcoreAchievements="$totalHardcoreAchievements"
+    :totalSoftcoreAchievements="$totalSoftcoreAchievements"
+    :userMassData="$userMassData"
+/>
+
+@if ($userMassData['ContribCount'] > 0 || $userMassData['Permissions'] >= $jrDevPermission)
```
Jamiras, line 56: This logic already exists in the controller to populate the developer stats. Would it make sense to just use `if !empty($developerStats)` here?

resources/views/community/components/user/profile-meta.blade.php
```diff
@@ -0,0 +1,68 @@
...
+    />
+
+    <div class="mb-2 sm:ml-[142px] md:ml-0 md:absolute md:top-0.5 md:right-0 lg:relative lg:ml-[142px] xl:absolute xl:ml-0">
+        <x-user.profile.social-interactivity :username="$username" />
+    </div>
+</div>
+
+<hr class="border-embed-highlight mb-2" />
+
+<x-user.profile.follows-you-label :username="$username" />
```
Jamiras, line 38: This seems out of place where it's at. Could it be moved up near the follow widget?

image

database/migrations/platform/2024_01_24_000000_update_codenotes_table.php
wescopeland: Index renamed in latest to `memory_notes_user_id_index`.

> Is this even necessary? I didn't run the migrate step and the page seems to be working.

I encountered some performance issues without adding this index. For some users with a lot of code notes, the query added 100ms+, and with the index this was brought down to <5ms.

database/migrations/platform/2024_01_24_000000_update_codenotes_table.php
wescopeland: Thanks, resolved in latest.

resources/views/community/components/user/profile/developer-claims-list.blade.php
wescopeland: Removed `count($claims) <= $numAllowedClaims` from the conditional.

resources/views/community/components/user/profile/mode-player-stat-cells.blade.php
wescopeland: Yes - fixed in latest.

resources/views/community/components/user/profile/developer-stats.blade.php
wescopeland: Not a bad idea, done in latest.

resources/views/community/components/user/profile/last-seen-in.blade.php
wescopeland: Done in latest.

resources/views/community/components/user/profile/last-seen-in.blade.php
wescopeland: Good catch, fixed in latest.

app/Community/Components/UserProfileMeta.php
wescopeland: Thanks, I've adjusted this logic in latest and hopefully it's in a better place now.

app/Community/Components/UserProfileMeta.php
wescopeland: Yep, had some cruft leftover from something else I was trying in here. Simplified in latest.

resources/views/community/components/user/profile/primary-meta.blade.php
wescopeland: Thanks, resolved in latest.

database/migrations/platform/2024_01_24_000000_update_codenotes_table.php
Jamiras: Bleh. I thought that was an aggregated column, not an index. An index makes sense.

resources/views/community/components/user/profile-meta.blade.php
wescopeland: It is tough to see on dark and black and easy to see on light mode. I'm indifferent on whether or not it should stay; I don't think it's hurting anything.

app/Community/Components/UserProfileMeta.php
wescopeland: I lean towards no but will defer to what decision you believe is best. My "no" leaning is only because this sounds like something we'd need to explain to people, which probably makes it a difficult UX to reason about.

app/Community/Components/UserProfileMeta.php
wescopeland: Fixed in latest.

resources/views/community/components/user/profile-meta.blade.php
wescopeland: Good idea, adjusted in latest. The legacy role indicator now also appears for all legacy roles that are not `Registered` or `Unregistered`.

app/Community/Components/UserProfileMeta.php
wescopeland: Thanks. I forgot to delete this. Removed in latest.

app/Platform/Concerns/ActsAsDeveloper.php
wescopeland: Yes, agreed. Done in latest.

resources/views/community/components/user/profile-meta.blade.php
wescopeland: Yep, that works great. Done in latest.

resources/views/community/components/user/profile-meta.blade.php
wescopeland: I've spent some time tweaking this label and the 3 social buttons across all breakpoints. Hopefully it's a little less out of place now.

resources/views/community/components/user/profile-meta.blade.php
Jamiras: It's definitely not hurting anything. It's just one of those things that I didn't see for a while, so when it did catch my eye I wondered if it was a bug.

app/Community/Components/UserProfileMeta.php
Jamiras: I don't have strong opinions one way or the other. It might be weird to see Average Points per Week decay on a retired user, but would someone even notice?

resources/views/community/components/user/profile/last-seen-in.blade.php
```diff
@@ -0,0 +1,64 @@
...
+    $sessionGame = Game::with('system')->find($userMassData['LastGame']['ID']);
+}
+
+$mostRecentRichPresenceMessage = (
+    $mostRecentSession?->rich_presence
+    ?? $userMassData['RichPresenceMsg']
+    ?? null
+);
+
+$parsedDate = Carbon::parse($mostRecentSession->rich_presence_updated_at);
```
Jamiras, line 27: this needs a `?->` too. Try visiting an account that hasn't been used in a while - like Brian.

wescopeland: Thanks, addressed in latest.

app/Community/Components/UserProfileMeta.php
wescopeland: Barring any objections, for now I'll keep this as-is.

resources/views/community/components/user/profile-meta.blade.php
wescopeland: Unless you're opposed, I'd like to keep this for now. On one of my monitors, I can't see the `<hr />` while in dark mode. On my work monitor (higher quality / color accuracy), it subtly adds some visual distinction between the two sections that my eye picks up on.

app/Community/Components/UserProfileMeta.php
Jamiras: I misled you here. The else should be eliminated. Last 30 days should include the last 7 days.

resources/views/community/components/user/profile/primary-meta.blade.php
```diff
@@ -0,0 +1,88 @@
...
+    'softcoreRankMeta' => [],
+    'userMassData' => [],
+    'username' => '',
+])
+
+<?php
+use App\Site\Enums\Permissions;
+use Illuminate\Support\Carbon;
+
+$hasVisibleRole = $userMassData['Permissions'] !== Permissions::Registered && $userMassData['Permissions'] !== Permissions::Unregistered;
```
Jamiras, line 12: I wonder if we shouldn't show Unregistered (at least for admins). When a new user complains about not being able to unlock stuff, it's often because they haven't verified their email yet.

Admittedly, that information is already available in the Admin panel, but a big tag would make it obvious.

luchaos: Spam, Banned, Unregistered, and Registered should completely be stripped from presentation on public profile information. Those are all accessible in the admin panels and considered relevant only for moderation imo.

Spam and Banned will result in a 404, hidden from the public eye.
Registered and Unregistered are just distinctive to email verification which is a moderation issue and should be considered private information. not having a verified email address should inform the features a user can use. Personally, I'd restrict that heavily. Anyways, non of the public's business imo.

luchaos: > Admittedly, that information is already available in the Admin panel, but a big tag would make it obvious.

We can think about adding a moderation card on profiles to have quick access to a user's activities and status, but should follow the introduction of the admin panel to make sure that's the primary interface to interact with a user on a moderation level.

Jamiras: Things will probably change as we move to Roles, but the Spam/Banned/Unregistered information that is currently available is being eliminated by this PR:
image

Spam and Banned do result in 404 for non-Admins, so having those accounts show the Spam/Banned tag is useful information - an Admin visiting a profile may not realize the account has been moderated.

I agree that the average user doesn't need to know the difference between whether or not another user is Registered or Unregistered, but it is useful information for diagnosing a specific common issue for new users. 

As I mentioned earlier, the user's Permissions setting is available in the Admin panel, as Moderators are allowed to change the value from there. It's just that the previously visible information is now hidden behind a page scroll and a click.

luchaos: Got it, that makes perfectly sense now. I interpreted that incorrectly :)

wescopeland: Logic adjusted in latest to show all `Permissions` strings for admins except "Registered".

app/Community/Components/UserProfileMeta.php
wescopeland: Fixed in latest.

Jamiras: You misunderstood me. The `else` should be eliminated, not the entire `else` clause:
```
if ($daysAgo <= 7) {
    $pointsLast7Days += $playerAchievement->achievement->points;
} 

$pointsLast30Days += $playerAchievement->achievement->points;
```

wescopeland: https://github.com/RetroAchievements/RAWeb/pull/2175/commits/c872060b51469a306fe1981d40f2bf386b721e93

wescopeland:
Rebased against the latest master with Filament and resolved the one merge conflict.

Jamiras (approved):

wescopeland:
Barring any additional feedback from @luchaos, I'd like to merge this before end of day.

app/Community/Components/UserProfileMeta.php
```diff
@@ -0,0 +1,218 @@
...
+        int $totalSoftcoreAchievements,
+    ) {
+        $this->averageCompletionPercentage = $averageCompletionPercentage;
+        $this->user = $user;
+        $this->userJoinedGamesAndAwards = $userJoinedGamesAndAwards;
+        $this->userMassData = $userMassData;
+        $this->userClaims = $userClaims;
+        $this->totalHardcoreAchievements = $totalHardcoreAchievements;
+        $this->totalSoftcoreAchievements = $totalSoftcoreAchievements;
+    }
```
luchaos, line 42:
```suggestion
    public function __construct(
        private string $averageCompletionPercentage = '0.00',
        private User $user,
        private array $userJoinedGamesAndAwards = [],
        private array $userMassData = [],
        private ?array $userClaims = null,
        private int $totalHardcoreAchievements = 0,
        private int $totalSoftcoreAchievements = 0,
    ) {
    }
```

luchaos (approved):
Unfortunately I don't have enough time to go deeper but it looks good to merge to me 👍

app/Community/Components/UserProfileMeta.php
wescopeland: Thanks, I didn't realize this would work. I'll get into a better habit of doing this -- wonder if there's even a possible linting rule w/ an autofixer? Adjusted in latest.

luchaos: Rector can do such refactoring but it's unfortunately very resource intensive.

PR #2177, opened 2024-01-27
fix(profile): centralise achievement grid within recently played game dropdowns

drisc:
Changes display layout from flex to grid and centers the grid of icons using grid-cols auto-fill.

Demo:

https://github.com/RetroAchievements/RAWeb/assets/2449110/5d8a3a8e-c8da-4df3-b1a2-13b6f9fd05d9


wescopeland:
At XS, the left-hand side of the container element has a padding difference of 8 pixels which is throwing off the centering.

drisc:
The container had pl-2 and pr-4 set which is why it's padding want equal the whole way around, that has been fixed in latest.

Before:
image


After:
image

wescopeland (approved):
Verified locally on desktop and mobile through the macOS iPhone simulator.

PR #2178, opened 2024-01-27
fix(game-list): remediate mobile scrolling issues

wescopeland:
Resolves a mobile regression introduced in #2149.

Currently, on mobile it is not possible to horizontally scroll on the new game lists.

**Root Cause**
`sticky` inherently cannot work on mobile because mobile needs to scroll the grid horizontally (`overflow-x-auto`), which disables sticky positioning.

This PR resolves the issue by moving sticky to being desktop only.

In an ideal world, mobile doesn't need to horizontally scroll on these pages at all, however getting to that point is going to be a significantly larger piece of work.


Jamiras (approved):
I still didn't need a horizontal scrollbar at MD on desktop, but I'll defer to your judgment for mobile.

I did notice that even at SM and XS, I didn't have a horizontal scrollbar.
image
image

Jamiras:
Unrelated: Should the "local (master)" actually reflect the branch I was on? What builds that label?

wescopeland:
> I did notice that even at SM and XS, I didn't have a horizontal scrollbar.

Not sure if this is helpful, but if you're using Chrome you can simulate swiping by enabling device simulation in the devtools like this:


https://github.com/RetroAchievements/RAWeb/assets/3984985/e6b4beef-dfce-47d4-966f-e3901ce87593

---

> Unrelated: Should the "local (master)" actually reflect the branch I was on? What builds that label?

In _body-end.blade.php_, I see:
```html
<b class="text-danger text-capitalize">{{ app()->environment() }} ({{ config('app.branch') }})</b>
```

Which leads me to _app.php_:
```php
'branch' => env('APP_BRANCH', 'master'),
```

Looks like it's coming from an environment variable.

Jamiras:
> Not sure if this is helpful, but if you're using Chrome you can simulate swiping by enabling device simulation in the devtools like this:

Good to know, but I was also pointing out that there's no scrollbar in desktop mode at a sufficiently small size. My desktop doesn't really do swiping.

wescopeland:
Ah, I see what you're saying now.

The table itself is what has the overflow (which allows for horizontal swiping on mobile), however because the table has such a great height, the scrollbar likely appears at the bottom of the table, which is all the way down the page. It will be quite impractical to scroll horizontally while using a desktop browser at XS or SM (unless you hold down the shift key and use the mouse wheel to scroll horizontally).

The ultimate fix here is to make these tables either mobile-first or more responsive across all breakpoints. This will be a much larger task, though.

PR #2179, opened 2024-01-27
perf(connect): put gameslist, officialgameslist, and hashlibrary behind redis

wescopeland:

PR #2182, opened 2024-01-28
perf(connect): optimize gameslist and officialgameslist routines

wescopeland:
Implements @Jamiras's suggestion from https://github.com/RetroAchievements/RAWeb/pull/2179#discussion_r1468890568:

> [gameslist and officialgameslist] could be hugely improved by just querying the ID/Title instead of doing a full data dump and extracting the Title.

**`gameslist` Before (ms)**
1300ms
1230ms
1190ms
1170ms
1190ms
Average: 1216ms

**`gameslist` After (ms)**
359ms
358ms
352ms
357ms
355ms
Average: 356ms (around a 3.5x speed increase)

### DIFF
--- a/app/Helpers/database/game.php
+++ b/app/Helpers/database/game.php
@@ -559,7 +559,7 @@ function getGamesListDataNamesOnly(int $consoleId, bool $officialFlag = false): 
             return $query->where('GameData.achievements_published', '>', 0);
         })
         ->orderBy('Console.Name')
-        ->orderBy('GameData.title')
+        ->orderBy('GameData.Title')
         ->select('GameData.Title', 'GameData.ID')
         ->pluck('GameData.Title', 'GameData.ID')
         ->toArray();
--- a/app/Helpers/database/game.php
+++ b/app/Helpers/database/game.php
@@ -548,17 +548,21 @@ function getGamesList(?int $consoleID, ?array &$dataOut, bool $officialFlag = fa
     return count($dataOut);
 }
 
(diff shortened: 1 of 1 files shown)


app/Helpers/database/game.php
```diff
@@ -548,17 +548,21 @@ function getGamesList(?int $consoleID, ?array &$dataOut, bool $officialFlag = fa
...
-
-    $data = getGamesListData($consoleID, $officialFlag);
-
-    foreach ($data as $element) {
-        $retval[$element['ID']] = utf8_encode($element['Title']);
-    }
-
-    return $retval;
+    return Game::with('system')
+        ->join('Console', 'GameData.ConsoleID', '=', 'Console.ID')
```
luchaos, line 554: `with('system')` sounds like it's already doing the join? 🤔 
can this be omitted or is it required for something else?

luchaos (approved):

app/Helpers/database/game.php
wescopeland: Safe to omit, removed in latest.

Jamiras (approved):
Actually looking at `getGamesListData`, it's not fetching as much information as I thought. The gain is going to be coming from using the aggregated column. That could be applied directly in `getGamesListData`, but the only other thing calling that is  achievementinspector.php via `getGamesList`, so it might be worthwhile to create a second PR to just inline the query there and eliminate `getGamesList`/`getGamesListData`.

PR #2183, opened 2024-01-28
fix(ticket-manager): fix link failing to add unofficial flag for ticket manager

### LINKED ISSUE #2173: Open Unofficial Tickets displays core tickets
**Describe the bug**
Viewing Unofficial achievements for a game and clicking the "Open Unofficial Tickets" button opens the Ticket Manager filtered to Core achievements, not Unofficial.

**To Reproduce**
Click the "Open Unofficial Tickets" button on a game with unofficial achievements that have tickets
Example: [https://retroachievements.org/game/5360?f=5](https://retroachievements.org/game/5360?f=5) (at time of posting this set has 5 core tickets, these are displayed instead of the 16 unofficial tickets)

**Expected behavior**
I expect the Unofficial achievement state filter to be applied in this case

drisc:
Resolves #2173.

Adds the unofficial flag to the url for the button when viewing unofficial achievements.

### DIFF
--- a/resources/views/platform/components/game/link-buttons/index.blade.php
+++ b/resources/views/platform/components/game/link-buttons/index.blade.php
@@ -85,12 +85,21 @@ if ($canSeeOpenTickets) {
     @endif
 
     @if ($canSeeOpenTickets)
+        @if ($isViewingOfficial)
         <x-game.link-buttons.game-link-button
             icon="🎫"
             href="{{ '/ticketmanager.php?g=' . $gameId }}"
         >
-            Open @if (!$isViewingOfficial) Unofficial @endif Tickets ({{ $numOpenTickets }})
+            Open Tickets ({{ $numOpenTickets }})
         </x-game.link-buttons.game-link-button>
+        @else
+        <x-game.link-buttons.game-link-button
+            icon="🎫"
+            href="{!! '/ticketmanager.php?g=' . $gameId . '&f=5' !!}"
+        >
+            Open Unofficial Tickets ({{ $numOpenTickets }})
+        </x-game.link-buttons.game-link-button>
+        @endif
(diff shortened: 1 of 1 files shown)


resources/views/platform/components/game/link-buttons/index.blade.php
wescopeland, line 1: I think I agree more logically with the original implementation. This PR essentially hard forks the button into two different buttons, which may be error-prone if this component ever gets refactored.

What may be better is an additional conditional in the `href` attribute.

drisc: I couldn't find a way to inline a condtional with the `href` so I created a new if statment with the required logic. The string needs to be within {!! !!} to escape the &. Otherwise it renders out as `&amp;f=5`.

wescopeland: `{!! !!}` should be okay here since we're in full control of the data.

I propose an idiomatic approach for the logic:

```php
// untested pseudocode
$ticketManagerUrlParams = [
    'g' => $gameId,
    'f' => $isViewingOfficial ? null : AchievementFlag::Unofficial,
];
$ticketManagerUrl = url('/ticketmanager.php') . '?' . http_build_query($ticketManagerUrlParams);
```

This combines Laravel's URL construction helper and `http_build_query()`. It also eliminates a magic number.

Now all the logic for the construction of this URL is in one place and will be easy to change when this eventually becomes a named route.

drisc: I've implemented the changes in your feedback in latest.

wescopeland (approved):

Jamiras (approved):

PR #2185, opened 2024-01-30
feat(user): display modern roles

wescopeland:
_**Not ready to be merged until users have cut over from legacy roles.**_

This PR migrates 3 areas of site UI to display modern roles: the user card, user profile, and forum posts.
The user list still needs to be converted, probably in a separate PR.

### How it works
A new method has been added to the `User` model: `->getVisibleRoles()`:

```php
public function getVisibleRoles(): Collection
{
    $roles = $this->roles()
        ->where('display', '>', 0)
        ->orderBy('display')
        ->orderBy('name')
        ->get();

    return $roles;
}
```

On the user profile, all the visible roles are shown on a new line (the redundant points row has been removed, as it's already in the Player stats box).

On the user card and forum posts UI, the first role in the list (if any) is displayed.

**User Profile**
Screenshot 2024-01-29 at 9 45 19 PM

**User Card**
Screenshot 2024-01-29 at 9 46 25 PM

**Forum Posts**
Screenshot 2024-01-29 at 9 46 58 PM


app/Models/User.php
```diff
@@ -326,6 +327,23 @@ public function getActivitylogOptions(): LogOptions
...
+    /**
+     * @return DbCollection<int, Role>
+     */
+    public function getVisibleRoles(): Collection
+    {
+        /** @var DbCollection|Role[] $roles */
+        $roles = $this->roles()
+            ->where('display', '>', 0)
+            ->orderBy('display')
+            ->orderBy('name')
```
Jamiras, line 341: I wonder if there's a better solution to this. `Developer` shares the same display order as `Artist` and `Writer`, but an Artist/Developer would always show Artist, and a Writer/Developer would always show Developer.

I'm also uncertain how the promotion from `Junior Developer` to `Developer` to `Veteran Developer` works. If the old role is not removed, `Developer` would display instead of `Veteran Developer`.

Jamiras (approved):
Left a comment about display order, but I don't think it should block this PR.

app/Models/User.php
wescopeland: The best solution is probably more granular display orders. I think we probably have a rough mental model for how roles should work from a hierarchy point of view.

In the examples given, I'd argue for both Artist/Developer and Writer/Developer, the Artist or Writer roles should be removed and the Developer role should remain, as those two roles will have a subset of the Developer role's abilities.

I'm also unsure how the promotion from Junior Developer to Developer to Veteran Developer works. It seems like we need some kind of "promotion" control in Filament that can intelligently handle this for users responsible for doing promotions. For example, if a Writer is promoted to Developer, the Writer role can be removed.

wescopeland:
Closing - I think this would be better-handled during the React migration.

PR #2189, opened 2024-01-31
feat(permissions): store new permissions related timestamps

luchaos:
Allows to use `$user->isForumVerified()`, `$user->hasVerifiedEmail()`, and `$user->isBanned()` in upcoming PRs.

Preparation for `TODO drop ManuallyVerified in favor of forum_verified_at`.

Backfill timestamps - low precision is acceptable in this case:

```sql
UPDATE
    UserAccounts
SET
    email_verified_at = Created
WHERE
    Permissions > 0
    AND email_verified_at IS NULL;

UPDATE
    UserAccounts
SET
    forum_verified_at = Created
WHERE
    ManuallyVerified = 1
    AND forum_verified_at IS NULL;

UPDATE
    UserAccounts
SET
    banned_at = Updated
WHERE
    Permissions < 0
    AND banned_at IS NULL;
```

### DIFF
--- a/app/Helpers/database/user-permission.php
+++ b/app/Helpers/database/user-permission.php
@@ -152,7 +152,7 @@ function setAccountForumPostAuth(string $sourceUser, int $sourcePermissions, str
 
     if (!$authorize) {
         // This user is a spam user: remove all their posts and set their account as banned.
-        $query = "UPDATE UserAccounts SET ManuallyVerified = 0, forum_verified_at = NOW(), Updated=NOW() WHERE User='$user'";
+        $query = "UPDATE UserAccounts SET ManuallyVerified = 0, forum_verified_at = null, Updated=NOW() WHERE User='$user'";
         $dbResult = s_mysql_query($query);
         if (!$dbResult) {
             return false;
--- a/app/Helpers/database/user-permission.php
+++ b/app/Helpers/database/user-permission.php
@@ -152,7 +152,7 @@ function setAccountForumPostAuth(string $sourceUser, int $sourcePermissions, str
 
     if (!$authorize) {
         // This user is a spam user: remove all their posts and set their account as banned.
-        $query = "UPDATE UserAccounts SET ManuallyVerified = 0, Updated=NOW() WHERE User='$user'";
+        $query = "UPDATE UserAccounts SET ManuallyVerified = 0, forum_verified_at = NOW(), Updated=NOW() WHERE User='$user'";
         $dbResult = s_mysql_query($query);
         if (!$dbResult) {
             return false;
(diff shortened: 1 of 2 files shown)


app/Helpers/database/user-permission.php
```diff
@@ -152,7 +152,7 @@ function setAccountForumPostAuth(string $sourceUser, int $sourcePermissions, str
 
     if (!$authorize) {
         // This user is a spam user: remove all their posts and set their account as banned.
-        $query = "UPDATE UserAccounts SET ManuallyVerified = 0, Updated=NOW() WHERE User='$user'";
+        $query = "UPDATE UserAccounts SET ManuallyVerified = 0, forum_verified_at = NOW(), Updated=NOW() WHERE User='$user'";
```
Jamiras, line 155: Shouldn't `forum_verified_at` be explicitly set to NULL when blocking a user?

Jamiras: I blocked a user and got this error:
```
Attempt to read property "Deleted" on null
```

This didn't happen on master, so I assume it's somehow related to this?

luchaos: > Shouldn't forum_verified_at be explicitly set to NULL when blocking a user?

addressed in https://github.com/RetroAchievements/RAWeb/pull/2189/commits/e1538037c16d1e5e1c670194867d9163b9ca90b2

> Attempt to read property "Deleted" on null

I can't replicate this, even without the fix. Where exactly did that error come up, which file?

Jamiras: I just clicked on one of the unauthorized posts from (http://localhost:64000/viewforum.php?f=0) and then clicked the Block button. I switched to master and tried the same without an error there. I did not reconfirm on the branch.

It seems to require the post being blocked not be the only post on the page. If it is, the user is redirected to a different page and the error does not occur.

With this information I _was_ able to reproduce on master.

Jamiras: It seems to be more complicated than that. If I add a post to an existing topic from a user who has not been verified and then go back to my user and click the Block button, the error doesn't occur.

The error seems to indicate that `$commentData->user` is null here:
https://github.com/RetroAchievements/RAWeb/blob/28a032d8f491be1f1ee68066f79ae2592bbe0818/resources/views/community/components/forum/post.blade.php#L29-L31

Which was changed in #2181. The old code would have returned null values via the join, so I think the solution is to coalesce to null:
```
    $commentAuthorDeletedDate = $commentData->user->Deleted ?? null;
    $commentAuthorJoinDate = $commentData->user->Created ?? null;
    $commentAuthorPermissions = $commentData->user->Permissions ?? null;
```

Jamiras:
I thought I posted this earlier, but apparently not:

> Backfill timestamps - low precision is acceptable in this case:

We could try to be smarted and pull some of that information from the audit comments.
image

app/Helpers/database/user-permission.php
Jamiras: I made the proposed change here: https://github.com/RetroAchievements/RAWeb/commit/75146b3d6314f6ca431c91819f7f094cfbba4a97

luchaos: awesome, great find - lgtm

Issue #2134, opened 2024-01-07
API_GetGameExtended returns undesirable fields


wescopeland:
**Describe the bug**
https://github.com/RetroAchievements/RAWeb/pull/2126#pullrequestreview-1801252883

This endpoint should not return `players_total`, `achievement_set_version_hash`, `achievements_published`, or `points_total`.

**Additional context**
This is likely occurring because the endpoint is using a helper function(s). We should detach it from that.

jacobguinther:
Are you proposing a new helper function similar to `getGameMetadata` or to just replicate the useful logic from that helper into `API_GetGameExtended`?

These fields are introduced when `getGameMetadata` calls `getGameData`.

We could just unset them in `API_GetGameExtended`

Are there large portions of `getGameMetadata` that are unnecessary? The `$user` and `$user2` parameters are not used in `API_GetGameExtended` but that's the only logic I noticed that wouldn't need to be replicated.

wescopeland:
> We could just unset them in API_GetGameExtended.

The web API should generally be detached from internal helper functions and rely more on Eloquent ORM for data retrieval. 

Calling `unset()` does indeed solve the issue, but not the root problem. Any time these internal helper functions change, the shape of data returned by the web API will also change.

PR #2193, opened 2024-02-01
refactor(api): migrate GetGame and GetGameExtended to Eloquent ORM

drisc:
Removed all calls to old functions that relied on crafted queries to retrieve data for these two endpoints.

They both now use data from Models.


wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/2134.

public/API/API_GetGame.php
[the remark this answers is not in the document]
drisc: I've applied these reductions to latest.

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,118 @@
...
+            return null;
+        } else {
+            $matchValue = $result->Title;
+            $remainingValue = preg_replace('/(?:^|\s)\[.*Subset - .*\](?:\s|$)/', '', $matchValue);
+            $remainingResult = Game::where('Title', $remainingValue)->first();
+
+            return (int) $remainingResult->ID ? $remainingResult->ID : null;
+        }
+    }
+}
```
wescopeland, line 82: I feel fairly strongly that this should live somewhere in the `Game` model class.

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,118 @@
...
+use App\Platform\Models\Game;
+use Carbon\Carbon;
 
 $gameID = (int) request()->query('i');
 $flag = (int) request()->query('f', (string) AchievementFlag::OfficialCore);
-getGameMetadata($gameID, null, $achData, $gameData, flag: $flag, metrics: true);
 
-if ($gameData === null) {
+$game = Game::with('system')->find($gameID);
+$gameSetClaims = AchievementSetClaim::find($gameID);
```
wescopeland, line 63: Does this query still run if `gameID` is 999999?

drisc: Yes, it still runs when $game is null, I'll prevent that.

public/API/API_GetGameExtended.php
drisc: Agreed, I have moved this to a public function in the Game model

app/Platform/Models/Game.php
```diff
@@ -161,6 +161,24 @@ public function shouldBeSearchable(): bool
...
     }
 
+    // == temp subset detection
+
+    public function getParentGameId(string $title, int $consoleID, int $gameID): ?int
+    {
+
+        $result = Game::where('Title', 'LIKE', '%' . $title . '%')->where('ConsoleID', '=', $consoleID)->first();
+
+        if ($result->Title = $title) {
```
wescopeland, line 171: Is this a logic error?

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,99 @@
...
+    ['NumDistinctPlayers' => count($game->players)],
+    ['NumAchievements' => count($gameAchievements)],
+    ['Achievements' => $gameAchievements],
+    ['Claims' => $gameClaims],
+    ['NumDistinctPlayersCasual' => count($game->players)], // Deprecated - Only here to maintain API V1 compat
+    ['NumDistinctPlayersHardcore' => count($game->players)], // Deprecated - Only here to maintain API V1 compat
+);
 
-return response()->json($gameData);
+return response()->json($getGameExtended);
```
luchaos, line 148: local variable can be inlined
```php
return response()->json(array_merge(
    $gameData,
    ['ConsoleName' => $game->system->Name],
    ['ParentGameID' => $game->getParentGameId($game->Title, $game->ConsoleID, $gameID)],
    ['NumDistinctPlayers' => count($game->players)],
    ['NumAchievements' => count($gameAchievements)],
    ['Achievements' => $gameAchievements],
    ['Claims' => $gameClaims],
    ['NumDistinctPlayersCasual' => count($game->players)], // Deprecated - Only here to maintain API V1 compat
    ['NumDistinctPlayersHardcore' => count($game->players)], // Deprecated - Only here to maintain API V1 compat
));
```

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,99 @@
...
+if ($gameSetClaims) {
+    $gameClaims =
+        [[
+            'User' => $gameSetClaims->User,
+            'SetType' => $gameSetClaims->SetType,
+            'GameID' => $gameSetClaims->GameID,
+            'ClaimType' => $gameSetClaims->ClaimType,
+            'Created' => Carbon::parse($gameSetClaims->Created)->format('Y-m-d H:i:s'),
+            'Expiration' => Carbon::parse($gameSetClaims->Finished)->format('Y-m-d H:i:s'),
+        ]];
```
luchaos, line 131:
```suggestion
    $gameClaims = [
        [
            'User' => $gameSetClaims->User,
            'SetType' => $gameSetClaims->SetType,
            'GameID' => $gameSetClaims->GameID,
            'ClaimType' => $gameSetClaims->ClaimType,
            'Created' => Carbon::parse($gameSetClaims->Created)->format('Y-m-d H:i:s'),
            'Expiration' => Carbon::parse($gameSetClaims->Finished)->format('Y-m-d H:i:s'),
        ]
    ];
```

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,99 @@
...
+            'ImageBoxArt' => $game->ImageBoxArt,
+            'Publisher' => $game->Publisher,
+            'Developer' => $game->Developer,
+            'Genre' => $game->Genre,
+            'Released' => $game->Released,
+            'IsFinal' => $game->IsFinal,
+            'RichPresencePatch' => md5($game->RichPresencePatch),
+            'GuideURL' => $game->GuideURL,
+            'Updated' => $game->Updated,
+        ];
```
luchaos, line 86: To match indentation:
```suggestion
    $gameData = [
        'ID' => $game->ID,
        'Title' => $game->Title,
        'ConsoleID' => $game->ConsoleID,
        'ForumTopicID' => $game->ForumTopicID,
        'Flags' => null, // Always '0', this is different in the extended endpoint test for some reason
        'ImageIcon' => $game->ImageIcon,
        'ImageTitle' => $game->ImageTitle,
        'ImageIngame' => $game->ImageIngame,
        'ImageBoxArt' => $game->ImageBoxArt,
        'Publisher' => $game->Publisher,
        'Developer' => $game->Developer,
        'Genre' => $game->Genre,
        'Released' => $game->Released,
        'IsFinal' => $game->IsFinal,
        'RichPresencePatch' => md5($game->RichPresencePatch),
        'GuideURL' => $game->GuideURL,
        'Updated' => $game->Updated,
    ];
```

public/API/API_GetGame.php
```diff
@@ -22,29 +22,35 @@
...
+            'ImageTitle' => $game->ImageTitle,
+            'ImageIngame' => $game->ImageIngame,
+            'ImageBoxArt' => $game->ImageBoxArt,
+            'Publisher' => $game->Publisher,
+            'Developer' => $game->Developer,
+            'Genre' => $game->Genre,
+            'Released' => $game->Released,
+        ];
+
+    return response()->json($gameData);
```
luchaos, line 53: local variable can be inlined, and to match indentation:
```suggestion
    return response()->json([
        'Title' => $game->Title,
        'GameTitle' => $game->Title,
        'ConsoleID' => $game->ConsoleID,
        'ConsoleName' => (string) $game->system->Name,
        'Console' => (string) $game->system->Name,
        'ForumTopicID' => $game->ForumTopicID,
        'Flags' => 0, // Always '0'
        'GameIcon' => $game->ImageIcon,
        'ImageIcon' => $game->ImageIcon,
        'ImageTitle' => $game->ImageTitle,
        'ImageIngame' => $game->ImageIngame,
        'ImageBoxArt' => $game->ImageBoxArt,
        'Publisher' => $game->Publisher,
        'Developer' => $game->Developer,
        'Genre' => $game->Genre,
        'Released' => $game->Released,
    ]);
```

public/API/API_GetGame.php
```diff
@@ -22,29 +22,35 @@
...
+} else {
```
luchaos, line 54: `else` can be omitted by inverting the condition (`if (!$game)`) and make it an early return.

app/Platform/Models/Game.php
luchaos: `===` 🙏

public/API/API_GetGame.php
luchaos: Alternatively: `$game->only(['Title', 'ConsoleID'])`, then map the few remaining fields, as the response already matches most of those field names.

luchaos: Planned for a long time, it might be worth introducing "Api V1" resource responses: https://laravel.com/docs/10.x/eloquent-resources#resource-responses

app/Platform/Models/Game.php
drisc: I'm guessing this needs updating elsewhere as well?

public/API/API_GetGame.php
drisc: Using map() adds a few ms to the response time (as I found in the Extended endpoint initially), just `return`ing the data is currenly faster.

app/Platform/Models/Game.php
luchaos: Seems to be the only occurrence in this PR - haven't checked the other files yet. I hope we didn't miss any 😅

public/API/API_GetGame.php
drisc: Good catch, updated

app/Platform/Models/Game.php
```diff
@@ -161,6 +161,24 @@ public function shouldBeSearchable(): bool
         return false;
     }
 
+    // == temp subset detection
+
+    public function getParentGameId(string $title, int $consoleId, int $gameId): ?int
```
wescopeland, line 166:
```
GET http://localhost:64000/API/API_GetGameExtended.php?z=...&y=...&i=24511
```

`ParentGameID` is null on this branch, but is 24453 in `master`.

wescopeland (changes requested):

app/Platform/Models/Game.php
wescopeland: The design of the function is worth thinking about some more:

* `getParentGameId()` should not require any arguments - it's already attached to the game that its given arguments are coming from.
* I think the `Game::where()` call can be entirely eliminated, which should improve performance because this query uses a `LIKE`, which skips database indexes.

public/API/API_GetGame.php
```diff
@@ -22,29 +22,31 @@
...
+
 $gameID = (int) request()->query('i');
 
-$game = getGameData($gameID);
+$game = Game::with('system')->find($gameID);
 
-if ($game === null) {
+if (!$game) {
     return response()->json();
+} else {
```
wescopeland, line 33: nit: This `else` is redundant.

public/API/API_GetGame.php
```diff
@@ -22,29 +22,31 @@
  *  string     Released                   release date information for the game
  */
 
+use App\Platform\Models\Game;
+
 $gameID = (int) request()->query('i');
```
wescopeland, line 27: nit: Might be a good time to change this to `$gameId` like the other endpoint.

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,97 @@
...
+use App\Platform\Models\Game;
+use Carbon\Carbon;
 
-$gameID = (int) request()->query('i');
+$gameId = (int) request()->query('i');
 $flag = (int) request()->query('f', (string) AchievementFlag::OfficialCore);
-getGameMetadata($gameID, null, $achData, $gameData, flag: $flag, metrics: true);
 
-if ($gameData === null) {
+if ($game = Game::with('system')->find($gameId)) {
```
wescopeland, line 62: Assignment operators within conditionals can be cumbersome to understand. Please assign `$game` outside of an `if`, then perform the truthiness check within.

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,97 @@
...
+    $gameSetClaims = AchievementSetClaim::where('GameID', $gameId)->get();
+}
+
+if (!$game) {
     return response()->json();
+} else {
```
wescopeland, line 68: nit: This `else` is redundant.

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,97 @@
...
+            'GameID' => $gc->GameID,
+            'ClaimType' => $gc->ClaimType,
+            'Created' => Carbon::parse($gc->Created)->format('Y-m-d H:i:s'),
+            'Expiration' => Carbon::parse($gc->Finished)->format('Y-m-d H:i:s'),
+        ];
+    });
+}
 
-return response()->json($gameData);
+return response()->json(array_merge(
```
wescopeland, line 136: nit: There's no need to instantiate 8 arrays here. `$gameData` is an array, and all the other fields can be in another array.

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,97 @@
...
+/* Achievements Map
+    The maps here are return the data in the order that the v1 api expects, originally
+    the endpoint used function calls that made queries with the data in the
+    order that the endpoint returns but the Eloquent collections order the
+    data by how the columns are laid out in the database. This adds a few ms to the
+    response time, though not enough that it has any real impact.
+*/
+
+if (!$game->achievements->where('Flags', $flag)->isEmpty()) {
+    $gameAchievements = $game->achievements->where('Flags', $flag)->keyBy('ID')->map(function ($am) {
```
wescopeland, line 99: I like this functional approach, but I've discovered this currently suffers from an N+1 query problem. This is largely because of `Achievement::find`, which is happening in the loop.

There is a big optimization opportunity here, which can be found by lifting the achievement queries out of the loop and grabbing all the needed achievements with a single query, perhaps with `::findMany()`.

public/API/API_GetGame.php
drisc: Without this else, the endpoint returns 1 instead of an empty json response

wescopeland: I'm experiencing something different:

```php
if (!$game) {
    return response()->json();
}

return response()->json([
    'Title' => $game->Title,
    'GameTitle' => $game->Title,
    'ConsoleName' => $game->system->Name,
    'Console' => $game->system->Name,
    'ForumTopicID' => $game->ForumTopicID,
    'Flags' => 0, // Always '0'
    'GameIcon' => $game->ImageIcon,
    'ImageIcon' => $game->ImageIcon,
    'ImageTitle' => $game->ImageTitle,
    'ImageIngame' => $game->ImageIngame,
    'ImageBoxArt' => $game->ImageBoxArt,
    'Genre' => $game->Genre,
    'Released' => $game->Released,
]);
```

```
GET http://localhost:64000/API/API_GetGame.php?z=...&y=...&i=9999999
200

[]
```

drisc: Ah I see what I did wrong, I removed the whole empty return.

public/API/API_GetGameExtended.php
drisc: The query now takes place outside of the map and passes a collection to the map to be keyed and returned in the correct order.

public/API/API_GetGameExtended.php
drisc: Array fields compacted, return now only includes $gameData and a second array for the nested and individual stuff

public/API/API_GetGameExtended.php
drisc: All further responses now rely on $game returning data

app/Platform/Models/Game.php
drisc: New title check code by @wescopeland implemented, legacy db lookup removed from getParentGameIdFromGameTitle() in favor of result passed from new function in Game model

app/Platform/Models/Game.php
```diff
@@ -161,6 +161,28 @@ public function shouldBeSearchable(): bool
...
+    {
+        // Use regular expression to check if the title includes a subset pattern and extract the base title.
+        if (preg_match('/(.+)\[Subset - .+\]/', $this->Title, $matches)) {
+            // Trim to ensure no leading/trailing spaces.
+            $baseSetTitle = trim($matches[1]);
+
+            // Attempt to find a game with the base title and the same console ID.
+            $game = Game::where('Title', $baseSetTitle)
+                 ->where('ConsoleID', $this->ConsoleID)
+                 ->first();
```
wescopeland, line 176:
```suggestion
                ->where('ConsoleID', $this->ConsoleID)
                ->first();
```

nit: There's some extra whitespace here that should be removed to keep with 4-space indentation.

public/API/API_GetGame.php
wescopeland: This feedback was erroneously marked as resolved.

Note: I do like marking the feedback I leave as resolved when I go in for a re-review. It helps me quickly take inventory of what I need to check when following up on a review.

public/API/API_GetGameExtended.php
wescopeland: The `else` can still be removed.

wescopeland (commented):
I don't foresee having any further feedback beyond these comments.

drisc:
I’ll open a new PR for a subset test this weekend

app/Helpers/database/game.php
```diff
@@ -24,17 +24,11 @@ function getGameData(int $gameID): ?array
...
 
 // If the game is a subset, identify its parent game ID.
-function getParentGameIdFromGameTitle(string $title, int $consoleID): ?int
+function getParentGameIdFromGameTitle(string $title, int $consoleId): ?int
 {
-    if (preg_match('/(.+)(\[Subset - .+\])/', $title, $matches)) {
-        $baseSetTitle = trim($matches[1]);
-        $query = "SELECT ID FROM GameData WHERE Title = :title AND ConsoleID = :consoleId";
-        $result = legacyDbFetch($query, ['title' => $baseSetTitle, 'consoleId' => $consoleID]);
+    $foundGame = Game::where('Title', $title)->where('ConsoleID', $consoleId)->first();
```
Jamiras, line 29: This adds an extra query to every place that was previously calling this function. The "[Subset -" guard should be re-added.

app/Models/Game.php
```diff
@@ -160,6 +160,28 @@ public function shouldBeSearchable(): bool
...
+            // Trim to ensure no leading/trailing spaces.
+            $baseSetTitle = trim($matches[1]);
+
+            // Attempt to find a game with the base title and the same console ID.
+            $game = Game::where('Title', $baseSetTitle)
+                ->where('ConsoleID', $this->ConsoleID)
+                ->first();
+
+            // If a matching game is found, return its ID.
+            return $game?->id ?? null;
```
Jamiras, line 178: This should probably return the whole game model, as both existing calls to `getParentGameIdFromGameTitle` proceed to fetch additional information from the related record.

drisc: Changing this to return a collection rather than an int will require refactoring all the areas that take this as input since they are only expecting a game id not a whole collection. Would that be better as a seperate PR?

app/Helpers/database/game.php
drisc: Subset guard re-added in latest

drisc: Adding the subset guard around $foundGame and it's return causes UploadAchievementTest to fail with this error: 
```
Call to a member function getParentGameId() on null
--- Expected
+++ Actual
@@ @@
 '{\n
-    "AchievementID": 2,\n
-    "Error": "",\n
-    "Success": true\n
+    "Error": "Call to a member function getParentGameId() on null",\n
+    "Status": 500,\n
+    "Success": false\n
 }'
```

What else do I need to do here to make sure it returns the correct data?

Jamiras: Without seeing your changes, I can't say for certain, but the error implies that you're still calling `$foundGame->getParentGameId()` and `$foundGame` is null. You could change the `->` to `?->`, but it would be simpler to just wrap the logic in a condtional:
```
    if (strstr('[Subset - ', $title) !== false) {
        $foundGame = Game::where('Title', $title)->where('ConsoleID', $consoleId)->first();
        return $foundGame->getParentGameId() ?? null;
    }

    return null;
```

app/Models/Game.php
Jamiras: I'm seeing things like this:
```
    $parentGameId = getParentGameIdFromGameTitle($game->Title, $game->system_id);
    if ($parentGameId) {
        $parentGame = Game::firstWhere('ID', $parentGameId);
```

Which could just be:
```
    $parentGame = getParentGameIdFromGameTitle($game->Title, $game->system_id);
    if ($parentGame) {
```

And:
```
    $parentGameId = getParentGameIdFromGameTitle($gameDataOut['Title'], $gameDataOut['ConsoleID']);

    $query = "SELECT players_total AS NumDistinctPlayers FROM GameData WHERE ID=" . ($parentGameId ?? $gameID);
    $gameMetrics = legacyDbFetch($query);
    $gameDataOut['ParentGameID'] = $parentGameId;
    $gameDataOut['NumDistinctPlayers'] = $gameMetrics['NumDistinctPlayers'] ?? 0;
```

Could be:
```
    $parentGame = getParentGameIdFromGameTitle($gameDataOut['Title'], $gameDataOut['ConsoleID']);
    $gameDataOut['ParentGameId'] = $parentGame?->id;
    $gameDataOut['NumDistinctPlayers'] = $parentGame?->players_total ?? 0;
```
 
In the other three cases where only the ID is necessary, just pull it from the model: `$parentGame?->id`.

I don't think these changes are complex enough to require a separate PR, but if you don't feel comfortable making the changes, someone else could do them later.

wescopeland:
There is some feedback that has been outstanding for a couple of weeks. How can I help get this PR over the finish line?

drisc:
I'm going to give this some attention today, I've been busy with IRL stuff for the past couple of weeks.

app/Models/Game.php
drisc: These suggestions have been applied to latest.

app/Helpers/database/game.php
drisc: This suggestion has been applied to latest.

app/Models/Game.php
```diff
@@ -160,6 +160,28 @@ public function shouldBeSearchable(): bool
         return false;
     }
 
+    // == temp subset detection
+
```
wescopeland, line 164:
```suggestion
    // TODO remove when game_achievement_sets is ready
```

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,100 @@
...
+];
+
+/* Maps
+    The maps here are return the data in the order that the v1 api expects, originally
+    the endpoint used function calls that made queries with the data in the
+    order that the endpoint returns but the Eloquent collections order the
+    data by how the columns are laid out in the database. This adds a few ms to the
+    response time, though not enough that it has any real impact.
+*/
+
```
wescopeland, line 98:
```suggestion
// Use maps to structure the data with how legacy API consumers might expect it to be returned.
```

public/API/API_GetGameExtended.php
```diff
@@ -50,28 +50,100 @@
...
+    'ImageIngame' => $game->ImageIngame,
+    'ImageBoxArt' => $game->ImageBoxArt,
+    'Publisher' => $game->Publisher,
+    'Developer' => $game->Developer,
+    'Genre' => $game->Genre,
+    'Released' => $game->Released,
+    'IsFinal' => $game->IsFinal,
+    'RichPresencePatch' => md5($game->RichPresencePatch),
+    'GuideURL' => $game->GuideURL,
+    'Updated' => $game->Updated,
```
wescopeland, line 88: If I'm not mistaken, this is a Carbon date too. While dumping it natively works, we may want to manually use `->format()` just to be safe.

wescopeland: Close, the format being output by the latest code revision is:

"2024-02-24 17:33:55"

To match prod, we need:

"2024-04-02T19:52:54.000000Z"

Might just be a matter of `->format('Y-m-d\TH:i:s.u\Z')`.

drisc: Did some googling and found that `toIso8601ZuluString()` is also a potential solution though it doesn't add the extra 0's between the Seconds and the Z so I'll go with the one that you suggested.

wescopeland (approved):

PR #2206, opened 2024-02-01
don't suggest subsets through shared hubs/developers

Jamiras:


app/Platform/Controllers/SuggestGameController.php
```diff
@@ -293,7 +293,7 @@ private function getReasonColumn(): array
         ];
     }
 
-    public function isGameEligible(int $gameId, array $gameIds): bool
+    public function isGameEligible(int $gameId, array $gameIds, array $why = []): bool
```
luchaos, line 296: i think `$context` would be a more fitting name for this parameter. not that important though

luchaos (approved):

PR #2210, opened 2024-02-01
perf(game): set box art dimensions ahead of time on the server

wescopeland:
This PR resolves a huge layout shift currently happening on game pages:

https://github.com/RetroAchievements/RAWeb/assets/3984985/4d1fd54e-c9a7-4312-a4cd-5ae2414c9bef

This is not a V6 regression; it has been happening for as long as I can remember.

**Root Cause**
Because box arts are dynamically sized images that sit above static content, they currently do not have defined `width` and `height` attributes. The browser by default assigns these values to 0, and a layout shift occurs when the browser finally loads the image. On fast connections this is imperceptible. On slow connections and mobile, the shift is quite obvious.

**Resolution**
These image dimensions can be set ahead of time by the server before loading the page by using PHP's builtin `getimagesize()`. Once the dimensions are determined, they can be cached long-term in Redis. The cache can be busted when new box art is uploaded for the game.

### DIFF
--- a/public/gameInfo.php
+++ b/public/gameInfo.php
@@ -15,7 +15,9 @@ use App\Platform\Services\GameListService;
 use App\Site\Enums\Permissions;
 use App\Site\Enums\UserPreference;
 use App\Site\Models\User;
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Blade;
+use Illuminate\Support\Facades\Cache;
 
 $gameID = (int) request('game');
 if (empty($gameID)) {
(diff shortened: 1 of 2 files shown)


public/gameInfo.php
```diff
@@ -1179,8 +1181,30 @@ function resize() {
     <?php view()->share('sidebar', true) ?>
     <aside>
         <?php
+        $boxArtImgSrc = media_asset($gameData['ImageBoxArt']);
+
+        $boxArtImgSize = Cache::remember(
+            'game:box-art-dimensions:' . $gameID,
+            Carbon::now()->addDays(7),
+            function () use ($boxArtImgSrc) {
+                $boxArtImgSize = @getimagesize($boxArtImgSrc);
```
Jamiras, line 1190: While this is probably good for production, it's very impactful on my dev site. Hitting random pages takes (on average) 1.2 seconds longer than on master - presumably because it has to download the asset from the real server to calculate the size.
```
MEDIA_URL=https://media.retroachievements.org
```

I'm guessing even on production, `$boxArtImgSrc` will be a URL so it'll be more complex than just reading a file from the host machine.

luchaos: the new media system will include metadata like dimensions in a model which will make it easier to set sizes explicitly. disk i/o is very expensive.
i'd favor waiting a bit longer to address the layout shift until we have the data in place?

wescopeland: Not a problem - though I do wonder if we could start storing this dimension data somehow now in the DB for these media assets which are causing the layout shifts. We could capture the data on upload for new assets and find some clever way to backfill data for existing assets.

luchaos: the media management will do just that. see `media` table.
all media that currently exists will be imported there, i'd like to adjust a few other things first though as it impacts all areas of the service.

wescopeland: This makes sense to me 👍 I will close this PR.

PR #2211, opened 2024-02-02
perf: add .animate-spin just-in-time

wescopeland:
Force `.animate-spin` to only appear when loading spinner SVGs are actually visible.


Jamiras (approved):

Alexdatadestroyer:
Thank you for this one! I guess that CPU hanging issue is over now 👍

PR #2220, opened 2024-02-02
feat: allow leaderboard publishing for not yet validated systems

wescopeland:
https://discord.com/channels/310192285306454017/1076603222489301002/1203001821963227196


app/Helpers/database/leaderboard.php
```diff
@@ -25,6 +26,14 @@ function SubmitLeaderboardEntry(
     if ($dbResult !== false) {
         $lbData = mysqli_fetch_assoc($dbResult);
 
+        $foundGame = Game::find($lbData['GameID']);
+        if ($foundGame?->ConsoleID && !isValidConsoleId($foundGame->ConsoleID)) {
+            $retVal['Success'] = false;
+            $retVal['Error'] = "Cannot insert the value $newEntry into leaderboard with ID: $lbID because the game's system is not yet validated.";
```
Jamiras, line 32: Might want a shorter error message for this.
image

The one for unofficial achievement unlocks is "Unofficial achievements cannot be unlocked."

Maybe "Cannot submit entry for unsupported console"

Jamiras: Ignore the fact that the error message says achievement. That's a bug with the new DLL.

wescopeland: No problem, updated in latest to your suggestion.

Jamiras (approved):

PR #2233, opened 2024-02-06
feat(user): migrate mod tools to blade and change their position on the page

wescopeland:
This PR moves the admin tools placement on user profiles to more neatly align with the new layout of the page.

With the new page layout, the admin tools are much further down the page and harder to see. This PR attempts to rectify the issue.

* Migrates the admin tools to blade and renames them to "Moderator" tools to better reflect modern roles.
* The "Moderator" box now lives near the user's primary meta near the top of the page.
* Clicking the Moderator box causes the tools to expand in the immediate space below the button.

Screenshot 2024-02-05 at 9 57 57 PM

Screenshot 2024-02-05 at 9 58 06 PM


wescopeland:
This is a stacked PR:

* `master`
  * #2229 
    * #2233 👈

resources/views/components/user/profile-meta.blade.php
```diff
@@ -19,6 +20,11 @@
 $jrDevPermission = Permissions::JuniorDeveloper;
 
 $isUserStatsDefaultExpanded = request()->cookie('prefers_hidden_user_profile_stats') !== 'true';
+$me = Auth::user() ?? null;
+$amIModerator = false;
+if ($me) {
+    $amIModerator = $me->getAttribute('Permissions') >= Permissions::Moderator;
+}
```
luchaos, line 27: can be shortened to something like this:
`$userIsModerator = request()->user()?->getAttribute('Permissions') >= Permissions::Moderator;`

I find the term "me" counter-intuitive in these context. As usually there's no need to reference "me" as most abilities are checked on the acting "user" attached to the request, which in turn exposes `can()` or acting on `request()->user()` 🤔 

Anyways, I think this could be inlined as with the new permissions/roles in place it would probably be rewritten to a blade directive, something like this:

```php
@can('manage', User::class)
    <x-user.profile.moderation-tools :targetUser="$user" />
@endcan
```

resources/views/components/user/profile/moderation-tools.blade.php
```diff
@@ -0,0 +1,134 @@
+@props([
+    'targetUser' => null, // User
+])
+
+<?php
+use App\Community\Enums\ArticleType;
+use App\Community\Enums\UserAction;
+use App\Enums\Permissions;
```
luchaos, line 8: To match the other components after the refactoring, these use statements should be at the very top of the file.

resources/views/components/user/profile/primary-meta.blade.php
```diff
@@ -101,4 +101,25 @@ class="rounded-sm h-[64px] w-[64px] sm:max-h-[128px] sm:max-w-[128px] sm:min-w-[
             <x-user.profile.follows-you-label :username="$username" />
         </div>
     </div>
+
+    @if ($amIModerator)
+        <button class="absolute bottom-0 right-0 btn" onclick="toggleModeratorTools()">
+            Moderator ▼
```
luchaos, line 107: In reference to maybe linking to Filament admin panels in the future, I had those initially labeled in verbs, like in this case it would've been "Moderate". Does this apply here, what do you think?

luchaos (commented):
I like the move of the dropdown and how the button moved to the right.

In the prototype there were "Manage" and "Edit" buttons in the very top right of the main column, on the same level as "section headings", in this case that'd be where other action buttons are placed atm as well.

Screenshot 2024-02-06 at 21 24 21

Are we still planning on moving those buttons there to then link into Filament where the moderation features should live in the future? We still can have "quick" actions on detail pages but more intricate moderation tools would be living in a dedicated UI.

resources/views/components/user/profile-meta.blade.php
wescopeland: Here's a good illustration for why I went with the "me" nomenclature:

```php
@if ($userIsModerator)
    <x-user.profile.moderation-tools :targetUser="$user" />
@endif
```

With `$userIsModerator`, it's ambiguous who the `targetUser` prop is targeting. While skimming, I'd infer this to mean we're only referring to a single user, when in fact we're talking about two different users.

The only other way I can think of to make this less ambiguous would be to inline the `request()->...` in the `@if` directive.

resources/views/components/user/profile/moderation-tools.blade.php
wescopeland: Fixed in latest.

While working on this, I couldn't help but notice that a `@use()` directive has been added to Laravel recently:

```php
@use('App\Community\Enums\ArticleType')
@use('App\Community\Enums\UserAction')
@use('App\Enums\Permissions')
```

Is this permissible in RAWeb?

resources/views/components/user/profile/primary-meta.blade.php
wescopeland: No issues here on changing this, updated to "Moderate" in latest.

resources/views/components/user/profile-meta.blade.php
luchaos: Ah, I see - that's what i usually referred to as `$actingUser` or similar in such cases.
Makes more sense now that you reminded me 👍

resources/views/components/user/profile/moderation-tools.blade.php
luchaos: I think that'd mess with the variety of editors we have in use as those wouldn't properly detect those as imports. Does VS Code understand those directives?
Personally, I'd prefer native php tags at the very top for now 🤔

luchaos: In regards to a potential introduction of bladestan or similar this would also mean that  eventually this php block should contain `declare(strict_types=1);` which i don't think blade has an equivalent to. At least i couldn't find anything on that after a quick search.

luchaos: Looking into `@use()` more on a machine where have more IDEs than Sublime Text available, it seems not even PHPStorm with the paid Laravel Idea plugin handles those properly. 
That'd render PHPStorm's refactoring features slightly less useful and pose some additional risks if an incorrect namespace is left.
With that in mind i think `<?php ?>` is the safer route to go until the directive finds better support.

wescopeland: No problem at all, let's stick with `<?php ?>`.

luchaos (approved):

resources/views/components/user/profile-meta.blade.php
Jamiras: I'd probably do `$canModerate` or something similar, which leans into the `@can('manage', User::class)` logic, but I'm fine with this for the time being.

resources/views/components/user/profile-meta.blade.php
```diff
@@ -37,6 +43,10 @@
     @endif
 </div>
 
+@if ($amIModerator)
+    <x-user.profile.moderation-tools :targetUser="$user" />
+@endif
```
Jamiras, line 48: The existing Admin button (and related code) still exists lower in the page.

wescopeland: Whoops, removed in latest. Thanks.

Jamiras (approved):

PR #2234, opened 2024-02-07
feat: skip over comments from banned users

wescopeland:
Updates `RenderCommentsComponent()` to not render comments from banned users if the current user has less than Moderator permissions.

Because `$numComments` is being set as an arg, there is an edge case from this change where if the only visible comment is from a banned user, the "no comments" message won't render. I don't think this is too important, but figured I'd preemptively point out the behavior.


wescopeland:
This is a stacked PR:

* `master`
  * #2229 
    * #2234 👈

Jamiras (approved):

PR #2237, opened 2024-02-09
fix(connect): block unlocking of unofficial achievements

wescopeland:


public/dorequest.php
```diff
@@ -333,8 +333,10 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
              */
             $response = array_merge($response, unlockAchievement($user, $achIDToAward, $hardcore));
 
-            dispatch(new UnlockPlayerAchievementJob($user->id, $achIDToAward, $hardcore))
-                ->onQueue('player-achievements');
+            if ($foundAchievement->is_published) {
```
Jamiras, line 336: Would it make sense to check `$response['success']`, which would also prevent firing the job if the player had previously unlocked the achievement?

public/dorequest.php
```diff
@@ -430,7 +432,7 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
         foreach ($awardableAchievements as $achievement) {
             $unlockAchievementResult = unlockAchievement($targetUser, $achievement->id, $hardcore);
 
-            if (!isset($unlockAchievementResult['Error'])) {
+            if (!isset($unlockAchievementResult['Error']) && $achievement->is_published) {
                 dispatch(new UnlockPlayerAchievementJob($targetUser->id, $achievement->id, $hardcore))
                     ->onQueue('player-achievements');
```
Jamiras, line 437: I wonder if the logic should be in the `UnlockPlayerAchievement` action? Are there other places that call the action? Will there be?

The job also updates the session, maybe we still want that to happen?

wescopeland: Good point. I've removed `$achievement->is_published` from the conditional and updated `UnlockPlayerAchievement` to instead use:

```php
if ($achievement->is_published) {
    $unlock->save();
}
```

public/dorequest.php
wescopeland: Changed the conditional to simply `$response['Success']` in latest.

app/Models/Achievement.php
```diff
@@ -282,6 +282,18 @@ public function getPointsWeightedAttribute(): int
...
+    {
+        $this->Flags = AchievementFlag::OfficialCore;
+        $this->save();
+    }
+
+    public function demote(): void
+    {
+        $this->Flags = AchievementFlag::Unofficial;
+        $this->save();
+    }
```
Jamiras, line 295: I see these were added for testing. I'd recommend against having them be part of the official model, as it implies that's all that needs to happen when an achievement is promoted/demoted. In reality, the player and game metrics need to be updated as well.

app/Platform/Actions/UnlockPlayerAchievement.php
```diff
@@ -77,7 +77,9 @@ public function execute(
         }
 
         // commit the unlock
-        $unlock->save();
+        if ($achievement->is_published) {
+            $unlock->save();
+        }
 
         // post the unlock notification
         PlayerAchievementUnlocked::dispatch($user, $achievement, $hardcore);
```
Jamiras, line 85: Recommend stifling the event as well. If the unlock wasn't committed, it didn't really happen. There's no reason to rebuild the player, game, or developer contribution metrics, and anything else that might be watching won't be able to find the unlock if they query for it.

app/Models/Achievement.php
wescopeland: That's a fair point I hadn't considered. Removed.

app/Platform/Actions/UnlockPlayerAchievement.php
wescopeland: No problem, done in latest.

Jamiras (approved):

PR #2238, opened 2024-02-09
feat(octane): enable Octane reloads in Docker

luchaos:
As a reminder: with our Laravel Sail setup, Octane runs at http://localhost:64008/ by default.

This adds a watcher to restart the Octane process after file changes.

An additional item is displayed in the bottom debug area:

Screenshot 2024-02-09 at 19 05 44

### DIFF
--- a/docker/8.2/supervisord.conf
+++ b/docker/8.2/supervisord.conf
@@ -14,8 +14,7 @@ stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
 
 [program:octane]
-# command=/usr/bin/php -d variables_order=EGPCS /var/www/html/artisan octane:start --watch --server=swoole --host=0.0.0.0 --port=8888
-command=/usr/bin/php -d variables_order=EGPCS /var/www/html/artisan octane:start --server=swoole --host=0.0.0.0 --port=8888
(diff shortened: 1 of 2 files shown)


wescopeland (approved):
Verified Octane and hot reload locally.

Is running the UI through 64008 the preferred way to test/debug?

luchaos:
> Is running the UI through 64008 the preferred way to test/debug?

Not quite yet - it's mainly to verify that everything is working with Octane eventually. I tested a few things and to my surprise I can't find any of the previous issues anymore (which is fantastic!)

As soon as everything is running with Octane on the server we might consider changing the default 64000 port to it and remove the other ports. That should be a later step imo.

wescopeland:
It appears with Octane running, hrefs are visually different. Not sure if this is something to worry about, but probably worth noting.

Screenshot 2024-02-09 at 1 50 00 PM

luchaos:
> It appears with Octane running, hrefs are visually different. Not sure if this is something to worry about, but probably worth noting.

This is due to the `APP_URL` in your local `.env` file - changing it to `http://localhost:64008` would fix the detection of external links.

wescopeland:
Got it, that makes perfect sense 👍

PR #2241, opened 2024-02-09
feat(user): linkify the profile motto

wescopeland:
Adds linkification support to user profile mottos, eg:

Screenshot 2024-02-09 at 4 19 34 PM

This is very difficult to do on the server and there isn't a good package in the PHP/Laravel ecosystem to handle this. Therefore, I built a thin Alpine.js directive around [linkifyjs](https://github.com/Hypercontext/linkifyjs), which is well-maintained, popular, and small.

It takes a moment for the profile's client-side JS to kick in. To compensate for this, the text color for hyperlinks gradually crossfades from `text-text` to `text-link` via TailwindCSS animation.


luchaos:
What do you think of adding a redirect page in-between to make sure we have a potential safeguard step in-between? I'd like people to know they're leaving the site into depths of the internet.

Inspired by Steam it's been prepared but is not used yet: https://retroachievements.org/redirect?url=https://github.com
It's meant to replace links in all user-generated texts.

wescopeland:
Good idea. I'll move this to draft and built that out this weekend.

wescopeland:
For reference: https://i.stack.imgur.com/sQF82.png

wescopeland:
Updated in latest to use the `/redirect` route. I've restyled this route to be mobile-first and similar to Steam's external URL gateway.


https://github.com/RetroAchievements/RAWeb/assets/3984985/408c136c-3540-4db7-a598-f4b33e375e24

package.json
```diff
@@ -16,6 +16,7 @@
     "dependencies": {
         "@floating-ui/dom": "^1.5.1",
         "js-cookie": "^3.0.5",
+        "linkify-html": "^4.1.3",
```
Jamiras, line 19: Do I need to do something other than `composer install` to get this? `sail npm run build` fails:
```
[vite]: Rollup failed to resolve import "linkify-html" from "/var/www/html/resources/js/alpine/linkifyDirective/linkifyDirective.ts".
```

wescopeland: Try running `npm install` in the RAWeb directory (or possibly `sail npm install`)

resources/views/components/user/profile/primary-meta.blade.php
```diff
@@ -55,8 +55,10 @@ class="rounded-sm h-[64px] w-[64px] sm:max-h-[128px] sm:max-w-[128px] sm:min-w-[
 
         {{-- Motto --}}
         @if (!empty($userMassData['Motto']))
-            <div class="rounded bg-box-bg px-2 py-1 max-w-fit italic text-2xs hyphens-auto mb-3">
-                <p style="word-break: break-word;">{{ $userMassData['Motto'] }}</p>
+            <div x-data="{}" class="rounded bg-box-bg px-2 py-1 max-w-fit italic text-2xs hyphens-auto mb-3">
+                <p x-linkify class="[&>a]:text-text [&>a]:transition-colors [&>a]:duration-700 [&>a]:ease-out" style="word-break: break-word;">
+                    {{ $userMassData['Motto'] }}
+                </p>
```
Jamiras, line 61: It's probably a limitation of the library, but it's trying to linkify part of "file.mp3".
image

resources/js/alpine/linkifyDirective/linkifyDirective.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function linkifyDirective(
+  el: HTMLElement,
+  _: DirectiveData,
+  { effect }: DirectiveUtilities,
+) {
+  effect(() => {
+    el.innerHTML = linkifyHtml(el.innerHTML, {
+      formatHref: {
+        url: (href) => `/redirect?url=${href}`,
```
Jamiras, line 12: redirect is misleading when the link is pointing at retroachievements.org: http://localhost:64000/user/TheUnwanted

image

http://localhost:64000/user/TheJoh4nx
image

resources/views/components/user/profile/primary-meta.blade.php
Jamiras: Also oddly linkifying this: 
image

resources/js/alpine/linkifyDirective/linkifyDirective.ts
Jamiras, line 12: We should probably also add the redirect intermediary page to `[url=]` shortcode tags (and linkified raw URLs in shortcode).

resources/views/components/user/profile/primary-meta.blade.php
wescopeland: I've updated the directive to exclude *.mp and *.ar top level domains.

resources/js/alpine/linkifyDirective/linkifyDirective.ts
wescopeland: Good catch. In latest, internal links will no longer pass through the redirect gateway.

wescopeland: Updated `[url=]`, `[link]` and auto-url embeds in latest.

app/Support/Shortcode/Shortcode.php
```diff
@@ -171,12 +171,40 @@ private function parse(string $input, array $options = []): string
...
+
+        // Is this an external link? If so, it needs to go through the redirect gateway.
+        $finalUrl = $prefixedUrl;
+        if (strpos($prefixedUrl, 'retroachievements.org') === false) {
+            $finalUrl = route('redirect', ['url' => $prefixedUrl]);
+        }
+
+        // Finally, build and return the needed anchor tag.
+        // HTML escape the visible URL to prevent XSS vulnerabilities.
+        return sprintf('<a href="%s">%s</a>', $finalUrl, htmlspecialchars($prefixedUrl, ENT_QUOTES, 'UTF-8'));
```
Jamiras, line 188: This is no longer using the text for the link:
```
Go to [url=http://google.com]Google[/url] then [url=http://retroachievements.org/]RetroAchievements[/url]!
```
image

instead of
image

resources/js/alpine/linkifyDirective/linkifyDirective.ts
```diff
@@ -0,0 +1,38 @@
...
+        },
+      },
+
+      validate: {
+        url: (text) => {
+          const sanitized = text.toLowerCase();
+
+          if (sanitized.endsWith('.mp') || sanitized.endsWith('.ar')) {
+            return false;
+          }
```
Jamiras, line 27: While this does address specific cases I found, it feel like a kludge. I'd rather have a few unexpected links then have to maintain this.

app/Support/Shortcode/Shortcode.php
wescopeland: Resolved in latest.

resources/js/alpine/linkifyDirective/linkifyDirective.ts
wescopeland: I've removed this `verify` logic in latest - all TLDs will now be linkified.

Jamiras (approved):

PR #2242, opened 2024-02-10
chore(RedirectController): add several 3rd party allowed domains

wescopeland:
Directly related to https://github.com/RetroAchievements/RAWeb/pull/2241, but decided to open as a separate PR in case it warrants more discussion unrelated to the core of that feature itself.

Our "redirect gateway" is modeled closely after Steam's (https://github.com/RetroAchievements/RAWeb/pull/2241#issuecomment-1936629001). Steam also has a list of allowed domains which the gateway doesn't warn about.

I've gone through the DB to look for common links that people put in their mottos and figured this was a good first pass at a reasonable 3rd party allowlist. In other words, `/redirect?url=twitch.tv/asdfasdfasdf` will take a user directly to the destination rather than throwing up the warning page.


wescopeland:
@luchaos I would be interested in your feedback here. Do you believe these allowed domains are acceptable? Are we perhaps missing any?

luchaos:
> Do you believe these allowed domains are acceptable? Are we perhaps missing any?

Tbh I initially had no allowlists in mind other than our own subdomains. Reasoning being that it would be an additional maintenance overhead to check if each of those sites are still what they claimed to be at a certain time.

I have no strong opinions on the list :)

app/Http/Controllers/RedirectController.php
```diff
@@ -12,8 +12,27 @@
...
+        'exophase.com',
+        'github.com',
+        'howlongtobeat.com',
+        'infinitebacklog.net',
+        'infinitebacklog.nl',
+        'psnprofiles.com',
+        'steamcommunity.com',
+        'twitch.tv',
+        'twitter.com',
+        'youtube.com',
```
Jamiras, line 35: What about `www.google.com`? Official forum posts link to google searches for GameFAQs, longplays, and wikipedia

Jamiras: Possibly `discord.com`? I see several posts linking to stuff on discord - particularly for events.

Jamiras: Do we need explicit subdomains? I see both "https://youtube.com" and "https://www.youtube.com" links.

RALordAndrew: I don't think Google should be on the allow list. The "I'm Feeling Lucky" button and google.com/url can redirect users to arbitrary websites.

wescopeland: I've added `discord.com` to the allowlist. No strong feelings one way or the other towards `google.com`. 

@Jamiras

> Do we need explicit subdomains?

Fortunately, `www` as an explicit subdomain is not required. On this branch, try navigating to the following URLs and you should see the same result:

http://localhost:64000/redirect?url=https://github.com
http://localhost:64000/redirect?url=https://www.github.com

Jamiras: I understand that github.com will resolve to www.github.com. My question is whether we need to put both github.com and www.github.com in the list so both URLs will be trusted and avoid the redirect warning.

wescopeland: Unless I'm misinterpreting the question, having only `github.com` in the list results in both `github.com` and `www.github.com` being trusted.

This is demonstrable by having it in the list and navigating to either of the two routes given above. A redirect warning will never be shown; the user will be sent directly to GitHub.

Jamiras: Nope. That was my question. I just saw all the subdomains for retroachievements.org and assumed www would have to be explicitly called out.

luchaos: Not sure if I added those subdomain variants for brevity or ignoring my own code which explicitly omits subdomains 😅 
https://github.com/RetroAchievements/RAWeb/blob/master/app/Http/Controllers/RedirectController.php#L37

wescopeland: Resolved merge conflict & removed all first-party subdomains. I've confirmed those work fine without needing to be included.

Jamiras (approved):

PR #2245, opened 2024-02-11
migrate gameCompare to controller

Jamiras:
Before:
image

After:
image

* Uses the standard game subpage header now
* Admin expander renamed to Moderate and moved into header
* "Change game" dropdown removed
* Sort widget added
  image
* Comparison list replaced with standard grid - column headers sticky on scroll
* Column order changed to more quickly compare user with target user, and still allow the most direct access to the achievement description when looking at information for the target user
* Sidebar widget now shows achievements earned instead of points (and is right-aligned)
* Footer also updated - icon no longer repeated, points removed, and mixed mode progress listed in a single row.

image

image


app/Platform/Controllers/CompareUnlocksController.php
```diff
@@ -0,0 +1,154 @@
...
+class CompareUnlocksController extends Controller
+{
+    public function __invoke(Request $request, Game $game, User $user): View
+    {
+        $activeUser = $request->user();
+        if ($activeUser === null) {
+            abort(403);
+        }
+
+        $this->getFollowedUsersCompletion($activeUser, $game);
```
wescopeland, line 24: Dumb question, is this actually doing anything?

resources/views/components/game/compare-progress.blade.php
```diff
@@ -0,0 +1,83 @@
...
+                    <td>{!! userAvatar($friend, iconSize: 20) !!}</td>
+                    <td class='text-right'>
+                        <a href="{!! route('game.compare-unlocks', ['game' => $game, 'user' => $friend->User]) !!}">
+                            {{ $completion['achievements_unlocked'] }}/{{ $completion['achievements_total'] }}
+                        </a>
+                    </td>
+                </tr>
+            @endforeach
+            </tbody></table>
+
```
wescopeland, line 73:
```suggestion
```

nit

resources/views/components/game/compare-progress.blade.php
```diff
@@ -0,0 +1,83 @@
...
+        @endif
+
+        <p class='mt-3'>Compare with any user:</p>
+
+        <input size='24' name='compareuser' type='text' class='searchboxgamecompareuser' placeholder='Enter User...' />
+        &nbsp;
+        <button class='btn' onclick='selectSearchBoxUser()'>Select</button>
+
```
wescopeland, line 81:
```suggestion
```

nit

resources/views/components/hidden-controls-toggle-button.blade.php
```diff
@@ -0,0 +1,7 @@
+@props([
+    'label' => 'Moderate', // string
+])
+
+<button id="hidden-controls-toggle-button" class="absolute bottom-2 right-0 mb-2 btn" onclick="toggleHiddenControls()">
+    {{ $label }} ▼
```
wescopeland, line 6:
```suggestion
    {{ $slot }} ▼
```

Using a slot instead of `$label` aligns more closely with how HTML buttons semantically work.

```html
<!-- native button -->
<button>Moderate</button>

<!-- custom button -->
<x-button>Moderate</x-button>

<!-- another custom button -->
<x-hidden-controls-toggle-button>Moderate</x-hidden-controls-toggle-button>
```

resources/views/components/hidden-controls.blade.php
```diff
@@ -0,0 +1,28 @@
...
+                    buttonEl.innerHTML = buttonEl.innerText.substring(0, buttonEl.innerText.length-1) + "▲";
+                }
+            } else {
+                contentEl.classList.add('hidden');
+                if (buttonEl) {
+                    buttonEl.innerHTML = buttonEl.innerText.substring(0, buttonEl.innerText.length-1) + "▼";
+                }
+            }
+        }
+    }
```
wescopeland, line 18:
```suggestion
    function toggleHiddenControls() {
        const buttonEl = document.getElementById('hidden-controls-toggle-button');
        const contentEl = document.getElementById('hidden-controls-content');
        
        if (contentEl && buttonEl) {
            contentEl.classList.toggle('hidden');
            buttonEl.innerText = buttonEl.innerText.slice(0, -1) + (contentEl.classList.contains('hidden') ? "▼" : "▲");
        }
    }
```

A bit more concise. I believe this is functionally equivalent, though the `buttonEl` conditional is lifted up higher.

resources/views/components/game/compare-progress.blade.php
```diff
@@ -0,0 +1,83 @@
...
+  });
+});
+
+function selectSearchBoxUser() {
+  var $searchBoxCompareuser = $('.searchboxgamecompareuser');
+  window.location = "{!! $baseUrl !!}" + $searchBoxCompareuser.val();
+}
+</script>
+
+<div id='gamecompare' class='component'>
```
luchaos, line 46: Minor nit: due to how the pages were written previously our html attributes use single quotes. It's common practice to use double quotes for html attributes, for newly written or ported blade views I'd like to aim for this cleanup step if possible (I'll probably do another chore PR just for that at some point).

resources/views/components/hidden-controls.blade.php
Jamiras: This was based on the function you added for the Moderate button on the profile page:
https://github.com/wescopeland/RAWeb/blob/fd5d5c789a90c7e179ff5f8f08ce5bd45ae11430/resources/views/components/user/profile/primary-meta.blade.php#L114-L124

I'll make the suggested changes.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -999,7 +999,15 @@ function resize() {
         }
 
         if ($user !== null && $numAchievements > 0) {
-            RenderGameCompare($user, $gameID, $friendScores, $totalPossible);
+            $gameModel = Game::find($gameID);
+            $followedUserCompletion = CompareUnlocksController::getFollowedUsersCompletion($userModel, $gameModel);
```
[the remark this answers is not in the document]
Jamiras: As indicated above, this function just performs a query that's needed by the `<x-game.compare-progress>` control. Is is acceptable to just put that query in the control?

resources/views/pages/game/[game]/achievements/compare/[user].blade.php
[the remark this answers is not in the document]
Comment (Jamiras) [reply resources/views/pages/game/[game]/achievements/compare/[user].blade.php:1]: I tried putting the column in all three positions. Having it first makes the most sense from a presentation standpoint, but functionally, this page is primarily used to look at the target player's times. I guess I could reverse the order of all the columns so it's "Achievement, Other User, Me".

resources/views/pages-legacy/gameInfo.blade.php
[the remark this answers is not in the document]
luchaos: i think moving the query into the component is ok in this case 👍

app/Platform/RouteServiceProvider.php
```diff
@@ -71,6 +72,7 @@ protected function mapWebRoutes(): void
             // Route::get('game/{game}/players', [GamePlayerController::class, 'index'])->name('game.player.index');
             Route::get('game/{game}/dev-interest', GameDevInterestController::class)->name('game.dev-interest');
             Route::get('game/{game}/suggest', [SuggestGameController::class, 'forGame'])->name('game.suggest');
+            Route::get('game/{game}/achievements/compare/{user}', CompareUnlocksController::class)->name('game.compare-unlocks');
```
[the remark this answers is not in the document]
luchaos: I think I'd go with something like `/user/{user}/game/{game}/compare/{targetUser}` as it's about two users where the game is related to the first as a sub-resource.

app/Platform/Controllers/CompareUnlocksController.php
```diff
@@ -0,0 +1,154 @@
...
+            ->whereIn('user_id', $followedUserIds)
+            ->select($fields)
+            ->orderBy('achievements_unlocked_hardcore', 'DESC')
+            ->orderBy('achievements_unlocked', 'DESC')
+            ->get()
+            ->toArray();
+
+        return $playerGames;
+    }
+}
```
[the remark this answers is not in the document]
luchaos: i think moving the query into the component is ok in this case as well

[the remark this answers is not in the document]
Jamiras: done

resources/views/pages-legacy/gameInfo.blade.php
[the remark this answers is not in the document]
Jamiras: done

app/Platform/RouteServiceProvider.php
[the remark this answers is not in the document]
Jamiras: The first user is implied as the logged in user. Do we want to allow arbitrarily comparing any two users?

`/user/me/game/1/compare/them`
`/user/them/game/1/compare/someoneelse`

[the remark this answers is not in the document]
luchaos: Good question. I guess it's enough to only ever compare the currently logged-in user with the user specified via URL.

So it's just the inversion of the current route: `/user/{user}/game/{game}/compare`
If I'm not mistaken it would allow to keep the current implementation of `CompareUlocksController` as-is?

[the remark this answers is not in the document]
Jamiras: There's a few things that had to be altered (particularly the path to the file and the breadcrumb), but I've made the suggested changes.

wescopeland:
Mobile still needs some work, likely due to the table layout. However, I am only posting this so it is noted. If you'd like, I'm happy to take a pass at the mobile UI after this is approved and merged.

Screenshot 2024-03-03 at 9 46 47 AM

app/Platform/Controllers/CompareUnlocksController.php
```diff
@@ -0,0 +1,120 @@
...
+                $achievements[$unlock->achievement_id]['otherUserTimestampRaw'] = $unlock->unlocked_hardcore_at;
+                $achievements[$unlock->achievement_id]['otherUserTimestamp'] = Carbon::parse($unlock->unlocked_hardcore_at)->format('d M Y, g:ia');
+                $achievements[$unlock->achievement_id]['otherUserHardcore'] = true;
+            } elseif ($unlock->unlocked_at) {
+                $achievements[$unlock->achievement_id]['otherUserTimestampRaw'] = $unlock->unlocked_at;
+                $achievements[$unlock->achievement_id]['otherUserTimestamp'] = Carbon::parse($unlock->unlocked_at)->format('d M Y, g:ia');
+                $achievements[$unlock->achievement_id]['otherUserHardcore'] = false;
+            }
+        }
+
```
wescopeland, line 72: I can't help but notice the code for managing `$activeUserUnlocks` and `$otherUserUnlocks` is largely duplicative.

I think we may be able to get away with something like:

```php
processUserUnlocks($activeUser, $gameId, $achievements, 'user');
processUserUnlocks($otherUser, $gameId, $achievements, 'otherUser');
```

```php
// pseudocode
function processUserUnlocks($user, $gameId, &$achievements, $prefix)
{
  $unlocks = ... make the Eloquent ORM query ...

  foreach ($unlocks as $unlock) {
    if (!array_key_exists(...) { continue; }

    $timestampKey = "{$prefix}TimestampRaw";
    $formattedTimestampKey = "{$prefix}Timestamp";
    $hardcoreKey = "{$prefix}Hardcore";

    if ($unlock->unlocked_hardcore_at) {
      $achievements[$unlock->achievement_id][$timestampKey] = $unlock->unlocked_hardcore_at;
      // ... more stuff ...
    } elseif ($unlock->unlocked_at) {
      // ...
    }
  }
}
```

app/Platform/Controllers/CompareUnlocksController.php
```diff
@@ -0,0 +1,120 @@
...
+                return 1;
+            } elseif ($bTimestamp == null) {
+                return -1;
+            } else {
+                return $aTimestamp <=> $bTimestamp;
+            }
+        }
+
+        return $a['DisplayOrder'] <=> $b['DisplayOrder'];
+    }
```
wescopeland, line 119:
```php
private function sortByUnlockTimestamps(array $a, array $b, string $field): int
{
    // Force null values to be sorted last.
    $nullTimestampValue = '9999-01-01 12:00:00';

    $aTimestamp = $a[$field] ?? $nullTimestampValue;
    $bTimestamp = $b[$field] ?? $nullTimestampValue;

    if ($aTimestamp !== $bTimestamp) {
        return $aTimestamp <=> $bTimestamp;
    }

    // If timestamps are equal, fall back to the DisplayOrder value.
    return $a['DisplayOrder'] <=> $b['DisplayOrder'];
}
```

Functionally equivalent, and whittles the implementation down from 4 possible returns to 2.

app/Platform/Controllers/CompareUnlocksController.php
```diff
@@ -0,0 +1,120 @@
...
+            'display' => function ($a, $b) {
+                return $a['DisplayOrder'] <=> $b['DisplayOrder'];
+            },
+            'title' => function ($a, $b) {
+                return $a['Title'] <=> $b['Title'];
+            },
+        };
+
+        usort($achievements, $sortFunction);
+    }
```
wescopeland, line 102: Closures from the `match` expression can be directly returned:

```php
private function sortList(array &$achievements, string $sortOrder): void
{
    usort($achievements, match ($sortOrder) {
        'otherUnlocks' => fn ($a, $b) => $this->sortByUnlockTimestamps($a, $b, 'otherUserTimestampRaw'),
        'display' => fn ($a, $b) => $a['DisplayOrder'] <=> $b['DisplayOrder'],
        'title' => fn ($a, $b) => $a['Title'] <=> $b['Title'],
        default => fn ($a, $b) => $this->sortByUnlockTimestamps($a, $b, 'userTimestampRaw'),
    });
}
```

resources/views/components/game/compare-progress.blade.php
wescopeland, line 1: The white space to the right of the button catches the eye:

Screenshot 2024-03-03 at 10 17 50 AM

This might solve it:

```html
<div class="w-full flex items-center gap-x-2">
  <input name="compareuser" type="text" class="searchboxgamecompareuser w-full" placeholder="Enter User..." />
  <button class="btn" onclick="selectSearchBoxUser()">Select</button>
</div>
```

resources/views/components/game/compare-progress.blade.php
```diff
@@ -0,0 +1,112 @@
...
+}
+</script>
+
+<div id="gamecompare" class="component">
+    <h2 class="text-h3">Compare Progress</h2>
+    <div class="nicebox">
+        @if ($followedUserCompletion === null)
+            <p>RetroAchievements is a lot more fun with others.</p>
+            @if ($user === null)
+                <a href="/createaccount.php">Create an account</a> or login and start earning achievements today!<br>
```
wescopeland, line 81: Dumb question, how do I get this to appear? I'm trying to view a game page while unauthenticated and that doesn't seem to do it.

resources/views/pages/user/[user]/game/[game]/compare.blade.php
```diff
@@ -0,0 +1,179 @@
...
+                    href="/usergameactivity.php?ID={{ $game->ID }}&f={{ $otherUser->User }}"
+                >
+                    View User Game Activity
+                </x-game.link-buttons.game-link-button>
+            </ul>
+        </div>
+    </x-hidden-controls>
+@endif
+
+@if (@$numAchievements === 0)
```
Comment (wescopeland) [inline resources/views/pages/user/[user]/game/[game]/compare.blade.php:90]: Is the `@` directly before `$numAchievements` a typo? Maybe just `@if (!$numAchievements)` would suffice.

Comment (wescopeland) [reply resources/views/pages/user/[user]/game/[game]/compare.blade.php]: `@if` simply performs a truthiness evaluation of the given expression. The expression `!$numAchievements` is close to functionally equivalent given that 0 is a falsy value (though this expression also checks for stuff like `null`). 

However, the difference between the two is usually pedantic and explicitly checking for `$numAchievements === 0` is also perfectly ok.

resources/views/components/game/compare-progress.blade.php
wescopeland: No worries 👍

resources/views/components/game/compare-progress.blade.php
wescopeland: I think I see the problem. The `input` tag should change too. In my code snippet, I removed the explicit `size` attribute and added a `w-full` classname.

wescopeland (approved):

PR #2247, opened 2024-02-12
refactor(forms): use existing components

luchaos:
Some of the prototype components were already included in V2 but never used. In preparation for porting the remaining components this focuses on forms. Gets rid of table layouts and makes forms responsive.

Examples were added to demo page.

Screenshot 2024-02-12 at 10 34 19
Screenshot 2024-02-12 at 10 34 28
Screenshot 2024-02-12 at 10 34 36

**Rich-text editors / shortcode**

`injectShorcode()` now has a `target` parameter to allow for multiple rich-text editors on the same page.

The existing `<x-input.textarea />` component combines all the features that were previously in separate components (character counter, preview, shortcode buttons). Adding the `richText` attribute on textareas enables rich-text controls.

The submit buttons have been removed from textarea as that does not belong there semantically and moved to the left. I'd like us to move to the UX/UI paradigm which suggests that submit buttons on full page forms should be left aligned to follow the flow of content; in dialogs, inline forms and on small screens they're either full width or right aligned. Unfortunately i can't find the reference for this anymore but i'd prefer it over moving everything to the right on full page forms. (Update: https://ux.stackexchange.com/a/13539, https://uxplanet.org/buttons-alignment-policy-a26de4ce0c70)

**Preparation for `<x-section>` layout components**

To prepare for the next PR which addresses the page headers' base components this PR introduces `<x-section>` and `<x-section-header>` components. The idea is to allow us to move away from boxing everything in the main and sidebar columns into one element, but separate sections visually. In addition, sections are simple slotted components which can have headers (usually headings) with action buttons aligned to the right and wrap on smaller screens as can be seen in the screenshot above where the submit button was moved to the heading's height for illustration.

A similar approach will be introduced in a later PR which makes use of the currently unused `page-header` area (again, see demo page for a hint). Some cleanup left on that end.

**Related changes in existing pages**

Refactor messages views (messages create and message-thread detail) and forum topic legacy pages (create topic, view post bottom comment area, edit post) to use section/form components.

Apply message policies in controllers and views to display forms and buttons depending on permission via policy.

Includes a fix to make `message/create` read `messages/create` as that required moving the file. This better reflects the current REST URL structure.


resources/views/pages/messages.blade.php
luchaos, line 1: This moves the thread title to the first column as I found myself constantly being linked away to user pages instead of opening the thread. I think the left-most link in a table should refer to the related resource in the given context.

This change should not have been included in this PR - I can move it if necessary.

resources/views/pages/demo/index.blade.php
```diff
@@ -144,36 +147,111 @@
...
+                </x-slot>
+            </x-section-header>
+            <x-input.text name="text" requiredSilent inline help="Some helpful text describing this input" />
+            <x-input.user requiredSilent label="{{ __res('user', 1) }}" value="Scott" inline help="Scott preselected" />
+            <x-input.user requiredSilent inline help="Search user" />
+            <x-input.textarea label="{{ __res('message', 1) }}" requiredSilent inline maxlength="20" help="Some helpful text describing this input" />
+            <x-input.checkbox label="Checkbox" checked inline requiredSilent />
+            <x-form-actions inline />
+        </x-section>
+    </x-form>
```
luchaos, line 225: Usually `<x-section />` wrap forms but in this case it's the other way around to leverage alpine's scope.

resources/views/components/form.blade.php
```diff
@@ -1,5 +1,19 @@
...
+    'validate' => false,
+    'enctype' => 'multipart/form-data',
+])
+
+<form
+    action="{{ $action }}"
+    method="{{ $method }}"
+    enctype="{{ $enctype }}"
+    x-data="{ isValid: {{ $validate ? 'false' : 'true' }}, isSending: false }"
+    x-on:submit="isSending = true"
```
luchaos, line 13: Each `<x-form>` has its own alpine scope.

app/Community/Controllers/MessageController.php
```diff
@@ -6,28 +6,24 @@
...
 use App\Models\User;
 use Illuminate\Http\RedirectResponse;
-use Illuminate\Http\Request;
-use Illuminate\Support\Arr;
-use Illuminate\Support\Facades\Validator;
 
 class MessageController extends Controller
 {
-    public function store(Request $request): RedirectResponse
+    public function store(MessageRequest $request): RedirectResponse
```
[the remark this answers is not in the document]
wescopeland: Good to know. Going forward, should `Request` be wrapped for these form handling functions in controllers?

resources/views/components/form-field.blade.php
```diff
@@ -1,33 +1,41 @@
...
-        @if(!empty($icon))
-            <div class="input-group">
-                <span class="input-group-prepend">
-                    <span class="input-group-text">
-                        {{ svg('fas-'.$icon, 'icon') }}
-                    </span>
-                </span>
+        @if($icon || $prepend)
+            <div class="flex flex-row">
+                <div class="form-control-prepend flex items-center">
```
wescopeland, line 22: All of the icons seem purely decorative. Is it possible to set `aria-hidden="true"` on this div when an icon is used?

resources/views/components/form-field.blade.php
```diff
@@ -1,33 +1,41 @@
...
+                </div>
                 {{ $slot }}
             </div>
         @else
             {{ $slot }}
         @endif
-        @if(!empty($help))
-            <p class="help-block text-secondary mb-0">
+        @if($help)
+            <p class="help-block text-muted">
```
wescopeland, line 31: Similar to error messages, this also needs an ID, eg:

```html
id="help-{{ $id }}"
```

resources/views/components/form-field.blade.php
```diff
@@ -1,33 +1,41 @@
...
-        <div class="{{ ($fullWidth ?? false) ? '' : (($inline ?? false) ? 'lg:w-36 lg:pr-3 lg:text-right' : '') }} pt-1 whitespace-nowrap {{ $errors && $errors->has($attribute ?? null) ? 'text-danger' : '' }}">
-            <label for="{{ $fieldId ?? $attribute ?? null }}">
-                {{ $label }} {{ !empty($required) ? '*' : '' }}
-            </label>
-        </div>
+@props([
+    'help' => null,
+    'hidden' => null,
+    'icon' => null,
+    'id' => null,
```
wescopeland, line 5: If no ID is given, should one be autogenerated like the child components?

resources/views/components/input/text.blade.php
```diff
@@ -1,12 +1,63 @@
...
+
+    <input
+        autocomplete="off"
+        class="form-control {{ $fullWidth ? 'w-full' : '' }} {{ $name && $errors && $errors->has($name) ? 'is-invalid' : '' }}"
+        id="{{ $id }}"
+        maxlength="{{ $maxlength }}"
+        name="{{ $name }}"
+        type="{{ $type }}"
+        value="{{ $name ? old($name, $model?->getAttribute($name) ?? $value) : $value }}"
+        @if($name && $errors && $errors->has($name))aria-describedby="error-{{ $id }}"@endif
```
wescopeland, line 57: This is a great use of `aria-describedby`, however it is also necessary to use this attribute when `help` is rendered. eg:

```html
aria-describedby="{{ $errors->has($name) ? 'error-'.$id : ($help ? 'help-'.$id : '') }}"
```

resources/views/components/input/checkbox.blade.php
```diff
@@ -1,19 +1,51 @@
...
             value="1"
-            {{ old($attribute, !empty($model) ? $model->getAttribute($attribute) : null) ? 'checked' : '' }}
+            {{ $name ? (old($name, $model?->getAttribute($name) ?? $checked) ? 'checked' : '') : ($checked ? 'checked' : '') }}
+            @if($name && $errors && $errors->has($name))aria-describedby="error-{{ $id }}"@endif
+            {{ $disabled ? 'disabled' : '' }}
+            {{ ($required || $requiredSilent) ? 'required' : '' }}
         >
-        <label class="custom-control-label" for="{{ $attribute }}">
-            {{ $label ?? __('validation.attributes.'.$attribute) }} {{ !empty($required) ? '*' : '' }}
+        <label class="" for="{{ $id }}">
```
wescopeland, line 47:
```suggestion
        <label for="{{ $id }}">
```

resources/views/components/input/text.blade.php
```diff
@@ -1,12 +1,63 @@
...
+<x-form-field
+    :help="$help"
+    :hidden="$type === 'hidden'"
+    :icon="$icon"
+    :id="$id"
+    :inline="$inline"
+    :model="$model"
+    :name="$name"
+>
+    @if($label && $type !== 'hidden')
```
wescopeland, line 43: It seems like `$type === 'hidden'` should set "sr-only" rather than removing the label from the DOM.

resources/views/components/input/select.blade.php
```diff
@@ -1,33 +1,73 @@
...
-            {{ $label }}
-        </div>
-        <input type="hidden" id="{{ $attribute }}" name="{{ $attribute }}" value="{{ $value }}">
-    @elseif($options->count() > 1 || empty($required))
-        <select class="form-control" id="{{ $attribute }}" name="{{ $attribute }}">
+    @if($label)
+        <x-slot name="label">
+            {{ $label === true ? __('validation.attributes.' . strtolower($name)) : $label }} {{ $required && !$requiredSilent ? '*' : '' }}
+        </x-slot>
+    @endif
```
wescopeland, line 45: Similar to other feedback, `$label` should likely always be present.

app/Community/Controllers/MessageController.php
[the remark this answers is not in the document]
luchaos: I think so, yes, when applicable. Prior to refactoring a form to Livewire; any form that submits data to a controller's store or update method _can_ reference a specialized form request class that includes all the validation rules. There are arguments for not using form request classes but calling the validator explicitly in the controller method instead (as can be seen in legacy pages and the MessageController prior to this PR). Same goes for calling `authorize()` in a controller vs. applying a middleware on the route level. It's pretty much a matter of taste of the end of the day. As in, where do you expect to find information about how the request is handled, locally or on a lower level.
It is kinda related to writing single-purpose controllers with `__invoke()` vs. defining multiple CRUD methods per controller ("resource controller").

From what I can tell we will use Folio's file-base routing and more Livewire components. Both of these developments imply that authorization will be localized (Folio's middleware() and calling authorize() in controllers, yet, defining abilities in route middleware for legacy pages) and validation in request classes. However, it's up to you and @Jamiras to tell me which one you prefer - I'm familiar enough with all of those approaches to look for them when in doubt naturally. Otherwise I'd go with form request classes until further notice.

resources/js/utils/injectShortcode.ts
```diff
@@ -3,11 +3,12 @@
  * selected text with the forum shortcode. If there's no selection, the
  * shortcode is inserted at the end of the message.
  *
+ * @param target The id of the target textarea.
  * @param start The opening tag of the shortcode.
  * @param end The closing tag of the shortcode. Defaults to an empty string.
  */
-export function injectShortcode(start: string, end = ''): void {
-  const commentTextarea = document.getElementById('commentTextarea') as HTMLTextAreaElement;
+export function injectShortcode(target: string, start: string, end = ''): void {
```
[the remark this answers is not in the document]
luchaos: thank you! i actually had that in mind after inspecting the frontend code you've written so far, liking the way you named variables referencing elements, but completely forgot to revisit this eventually. Addressed in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/eca58ca51055f1edd0fc3e0289c83c365b7a64aa

resources/views/components/form-field.blade.php
```diff
@@ -1,33 +1,41 @@
...
+    'inline' => false,
+    'label' => false,
+    'name' => null,
+    'prepend' => null,
+    'required' => false,
+])
+
+<div class="{{ $inline ? 'lg:flex' : '' }} {{ $hidden && !($name && $errors && $errors->has($name)) ? '' : 'mb-3 last-of-type:mb-0' }}">
```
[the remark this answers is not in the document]
luchaos: Not at all - I think that's a fine approach. While losing the implicit spacing of form elements regardless of context, adding a container to drive spacing explicitly sounds good to me. Flex is a cleaner approach here, agreed. Do we want to introduce a `<x-form.group>` component as an immediate container of form elements to consolidate this or would that be superfluous, not worth being dry in this case?
```html
<x-form>
    <x-form-group> <!-- has class="flex flex-col gap-y-3" -->
        <x-input />
        <x-input />
        <x-input />
    </x-form>
</x-form>
```

resources/views/components/input/text.blade.php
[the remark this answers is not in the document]
luchaos: that's what it was actually called before i copied it into the text input component :D good call, i'll rename it to `input` to better reflect the relevant native element.

resources/views/components/form.blade.php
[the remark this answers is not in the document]
luchaos: Sounds good - I think it was called `form` in one of the prototype's iterations even. Good reasoning. Maybe it's best to incorporate this change when moving all the base components? Would make it a "pure" refactor PR without any feature work.

resources/views/components/form-field.blade.php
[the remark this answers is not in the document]
wescopeland: Very interesting question. The presence of a form-group component implies there could also be multiple form groups within a single form. I don’t think I have a strong opinion on this one way or the other, but my mind does wander to considering the implications of multiple form groups 🤔

resources/views/components/input/checkbox.blade.php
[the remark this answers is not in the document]
luchaos: I'm not sure yet but I think I prefer just the checkbox and the actual text to be clickable. Wrapping the input and the label within the `label` tag has the side-effect to make a lot of whitespace apart from the gap clickable, too, which I find to be unexpected 🤔

resources/views/components/form.blade.php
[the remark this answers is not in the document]
luchaos: Renaming it to `form` actually loses some context to the underlying `<input>` element.
E.g. `<x-base.form.checkbox />`, `<x-base.form.date />`, `<x-base.form.datetime-local />`.
I think `<x-base.input.datetime-local />` better describes that it is in fact a datetime-local input?

resources/views/components/input/checkbox.blade.php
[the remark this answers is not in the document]
wescopeland: It's a bug.

Screenshot 2024-02-18 at 9 20 22 AM

The checkbox tap target will always be too small for mobile users. Compare this to any checkbox on user profiles and you won't find the same issue exhibited:

Screenshot 2024-02-18 at 9 22 19 AM

None of the whitespace around the checkbox is clickable, because the checkbox is not influencing layout in any way. That's left purely to the discretion of the container element.

resources/views/components/form.blade.php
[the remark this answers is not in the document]
wescopeland: I agree it does cause this problem in some circumstances, but creates more ambiguity in others 🤔

I think this nuance comes down to these components:
* search.blade.php
* select.blade.php
* textarea-rich-text-controls.blade.php
* textarea.blade.php
* user.blade.php

Every other component appears to be syntactic sugar around _text.blade.php_. So I can't help wonder if this structure makes sense:

```
/form
  /input
    checkbox.blade.php
    code.blade.php
    date.blade.php
    datetime-local.blade.php
    email.blade.php
    file.blade.php
    image.blade.php
    number.blade.php
    password-confirmed.blade.php

  input.blade.php (this is no longer called text.blade.php)
  search.blade.php
  select.blade.php
  textarea-rich-text-controls.blade.php
  textarea.blade.php
  user.blade.php
```

This leads to usages like:
```html
<x-base.form.input />
<x-base.form.input.checkbox />
<x-base.form.input.email />
<x-base.form.input.date />

<x-base.form.textarea />
<x-base.form.select />
```

This is a tiny bit more verbose, but the relationships between the fields (and their HTML element counterparts) seems explicitly clear. What do you think?

resources/views/components/input/checkbox.blade.php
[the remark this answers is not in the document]
luchaos: I missed that inline flex would solve the problem i had. Addressed in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/41c88377c97bc403d384bbd16a9dcc98d5ca02d0

resources/views/components/form-field.blade.php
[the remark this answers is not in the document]
luchaos: Yeah, not sure either. I'd like to postpone this particular change for another PR if that's ok for you?

resources/views/components/form.blade.php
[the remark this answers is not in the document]
luchaos: Moved in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/ba2b53fe7bb5c2985e0bb76d7dc1ffb8c9565491

Should I take care of the other base components while at it or do it in a separate PR?
Some of the other base components might need some additional cleanup (not using props etc)

resources/views/components/input/text.blade.php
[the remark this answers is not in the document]
luchaos: Addressed in https://github.com/RetroAchievements/RAWeb/commit/ba2b53fe7bb5c2985e0bb76d7dc1ffb8c9565491

resources/views/components/form-field.blade.php
```diff
@@ -1,33 +1,41 @@
...
+    @if($label)
```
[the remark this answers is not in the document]
luchaos: Good point. The only reason it's been written like this is for cases like checkboxes where the layout is not driven by the label. I made it more explicit to omit rendering labels elements for checkboxes and hidden inputs - everything else will have its label rendered.
I don't think we need another case where the label should be sr-only yet.

https://github.com/RetroAchievements/RAWeb/pull/2247/commits/659b1830882b86d77028687d5e6acb68067f7816

resources/views/components/form-field.blade.php
luchaos: good catch, addressed in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/28fb4f08eb56fc20fbc89888b8fe46f89f4acb7d

resources/views/components/input/text.blade.php
luchaos: I assumed hidden inputs to not require labels 🤔

resources/views/components/form-field.blade.php
luchaos: Added in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/a2f129ebbf5c5197aa4d72235ab69fbdb7f05115

resources/views/components/input/text.blade.php
luchaos: That's awesome - added in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/a2f129ebbf5c5197aa4d72235ab69fbdb7f05115

resources/views/components/input/select.blade.php
luchaos: Addressed in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/7aaaa88ce62aa676d6a2b56cbf78db0a530a4a37

resources/views/components/form.blade.php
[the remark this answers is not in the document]
luchaos: Moved button (https://github.com/RetroAchievements/RAWeb/pull/2247/commits/91cad42726bfac331decc00226a8297fb7e8a989) and form (https://github.com/RetroAchievements/RAWeb/pull/2247/commits/f950ecc575c677d2ada2f3612dade9d6bd34bdfa) components to base directory.

The remaining components will follow in a separate PR.

resources/views/components/form-field.blade.php
luchaos: Good idea - just to be safe. Added in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/666665af0c4796773e9ceb420619c82e600c797b

resources/views/components/form.blade.php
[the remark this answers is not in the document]
wescopeland: This is much more intuitive, thank you!

resources/views/components/form-field.blade.php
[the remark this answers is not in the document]
wescopeland: Thanks. I've poked around with the form controls and confirmed the DX becomes acceptably disagreeable when omitting the label, enough for the developer to know they've probably made a mistake 👍

resources/views/components/form-field.blade.php
```diff
@@ -1,33 +1,41 @@
...
-        </div>
+@props([
+    'help' => null,
+    'hidden' => null,
+    'icon' => null,
+    'id' => null,
+    'inline' => false,
+    'label' => false,
+    'name' => null,
+    'prepend' => null,
```
wescopeland: This seems to always be `null`, no matter what (unless there's a reference I haven't found). Is this a dead prop? One reason I'm wondering is because if this accepts a string, it could cause an a11y issue due to the new `aria-hidden="true"` attribute. May be best to remove.

resources/views/pages/demo/index.blade.php
wescopeland, line 1: I'd like to propose removing the styled form inputs section from the bottom of the page, as these components actually have baked in them some not-great a11y practices compared to the standardized form components.

resources/views/components/input/text.blade.php
wescopeland: Apologies, I actually misinterpreted in my original review what `$type` was doing here. You're right, input fields with type="hidden" do not require labels.

resources/views/pages/demo/index.blade.php
luchaos: removed in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/3618ea26697b016bce3d3be42d4388e13fc87206

I only kept those around to see how the tailwind form plugin would behave - that'S from when I aimed for DaisyUI compatibility. Not needed anymore.

resources/views/components/form-field.blade.php
luchaos: good point, it's not needed except for icon prepends. removed prop in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/84f3880c0d8ec294cec6be61d7c05f705397eb77

resources/views/pages/messages.blade.php
luchaos: Is this change ok for you, too @Jamiras ?

Before
Screenshot 2024-02-27 at 01 46 53

After
Screenshot 2024-02-27 at 01 47 00

public/js/all.js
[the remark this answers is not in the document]
luchaos: Still needed

wescopeland (approved):
Thanks for working through my feedback.

I think the only thing that's still a pain point for me is `mb-3 last-of-type:mb-0`. Layout concerns shouldn't belong to these components, it should be the responsibility of their container.

Container-driven layout ultimately increases the reusability of components ensuring they work in all contexts. It also increases the child components' maintainability, as the components can focus simply on their content and functionality.

It's OK if we push this to a follow-up PR, but if we decide to do that, IMO it should be resolved immediately after this PR is merged (I'm happy to do this).

resources/views/components/form-field.blade.php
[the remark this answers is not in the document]
luchaos: As discussed: introduced containers as suggested in https://github.com/RetroAchievements/RAWeb/pull/2247/commits/20e85e996258905c3268e1319b8b63c2642d5136

wescopeland (approved):

PR #2253, opened 2024-02-18
feat(developer-stats): add pagination

wescopeland:
Adds pagination support to the /developerstats.php route. Currently it is impossible to see beyond the first 100 results.

**Before**
Screenshot 2024-02-18 at 11 24 24 AM

**After**
Screenshot 2024-02-18 at 11 24 12 AM


resources/views/pages-legacy/developerstats.blade.php
```diff
@@ -109,4 +128,20 @@
     }
     echo "</tbody></table></div>";
     ?>
+
+    @if ($previousPageHref || $nextPageHref)
+        <div class="flex flex-col gap-y-4 items-center justify-between sm:flex-row sm:justify-end md:gap-x-4">
+            <div class="flex gap-x-4 items-center">
+                @if ($previousPageHref)
+                    <a class="btn transition-transform lg:active:scale-95" href="{{ $previousPageHref }}">< Previous</a>
```
Jamiras, line 136: Filters (such as only viewing inactive developers) are lost when changing pages.

resources/views/pages-legacy/developerstats.blade.php
```diff
@@ -8,16 +8,35 @@
...
-                                            & ~TicketFilters::StateClosed;
+$offset = requestInputSanitized('o', 0, 'integer');
+$resolvedForOthersTicketFilter = (
+    TicketFilters::AllFilters
+    & ~TicketFilters::StateOpen
+    & ~TicketFilters::StateRequest
+    & ~TicketFilters::StateClosed
+);
+
+$devStatsList = GetDeveloperStatsFull(100, $offset, $type, $devFilter);
```
Jamiras, line 19: Suggest lowering the per-page count to 50 or maybe even 25. With pagination, scrolling should be limited.

wescopeland: Lowered to 50 - happy to go further down to 25 if you feel this is still excessive.

Jamiras: I think lower is probably better. Maybe it's just because the only way to see what page your on is to scroll all the way down.

app/Helpers/database/user.php
```diff
@@ -282,7 +314,7 @@ function GetDeveloperStatsFull(int $count, int $sortBy, int $devFilter = 7): arr
                   WHERE $stateCond
                   GROUP BY ua.ID
                   ORDER BY $order
-                  LIMIT $count";
+                  LIMIT $offset, $count";
```
Jamiras, line 317: Offset doesn't seem to be working when sorting by tickets resolved:
http://localhost:64000/developerstats.php?t=4&f=4&o=100

Or open tickets:
http://localhost:64000/developerstats.php?t=3&f=4&o=50

Or active claims:
http://localhost:64000/developerstats.php?t=7&f=1&o=50

resources/views/pages-legacy/developerstats.blade.php
wescopeland: Not a problem at all. Lowered to 25 in latest.

Jamiras (approved):

PR #2256, opened 2024-02-23
fix session classification for manual unlocks

Jamiras:
Fixes an issue where an achievement unlocked between two sessions doesn't appear in either session. Originally reported here https://discord.com/channels/476211979464343552/1207063525248344095/1210362941921038387 with a manually unlocked achievement.

Also updated the display to report when an achievement was manually unlocked.


app/Helpers/database/user-activity.php
```diff
@@ -350,7 +351,7 @@ function getUserGameActivity(string $username, int $gameID): array
 
                     return;
                 }
-                $possibleSession = $session;
+                $possibleSession = &$session;
```
Jamiras, line 354: This was the original problem. Not grabbing the child by reference didn't propagate the changes back to the caller.

wescopeland (approved):

Issue #1723, opened 2023-08-07
The results field doesn't follow the navbar / search input


w400pedro:
#### Search for something in the input, then the results will be shown, and after that, scroll up or down to see the issue.

I will attach this video recorded on Google Chrome to provide a better description:

https://www.youtube.com/watch?v=niMlP61HnCY

wescopeland:
I believe this has to do with the search box's current implementation using jQuery UI Autocomplete, which is not very flexible and difficult to extend.

It would probably be ideal to consider migrating this away from the jQuery UI component to something like Floating UI. The tricky part is getting accessibility right, as this would all need to be reimplemented from scratch.

WAI-ARIA compliance guidelines on such a component can be found here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/

PR #2257, opened 2024-02-23
 refactor(search): make search results follow input when scrolling

jacobguinther:
I rewrote the main website search bar using Alpine and Floating UI. This fixes an issue where if the user scrolled down while the search results were open the search results would not follow the search input.

I followed the WAI-ARIA compliance guidelines:  
[https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/)

I didn't style the focus or hover state because I thought it might look weird if it was the only element styled to spec when tabbing through. (2px focus ring with 2px space between ring and combobox)

I did add some custom CSS outside of Tailwind for the dropdown item selected state. The combobox spec said to replace borders with padding instead of making them transparent. Tailwind uses rem for padding and px for borders and I needed to use px for both.

RA-Search

Fixes #1723

### DIFF
--- a/resources/css/search.css
+++ b/resources/css/search.css
@@ -34,6 +34,22 @@
   padding-left: 10px;
 }
 
+#search-listbox>li {
+  padding: 3px 6px;
+  font-family: Segoe UI, Arial, sans-serif;
+}
+
+#search-listbox>li:hover {
+  padding: 1px 4px;
+}
+
+#search-listbox>li.listbox-item--hover {
+  background: rgb(253 230 138);
+  border: 2px solid rgb(161 98 7);
+  border-radius: 0.5rem;
+  padding: 1px 4px;
+}
+
 @media only screen and (max-width: 374px) {
   .searchbox-top .searchboxinput:focus,
   .searchbox-top .searchboxinput:not(:placeholder-shown) {
--- a/resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
+++ b/resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
@@ -1,0 +1,129 @@
+import { fetcher } from '@/utils';
+
+type SearchResult = {
+  label: string,
+  mylink: string,
+}
+
+type NavbarSearchComponentType = {
+  showSearchResults:boolean,
+  searchText:string,
+  results: SearchResult[],
+  selectedIndex:number,
+  activeDecendentId: string,
+  handleClickSearchResult: (label: string) => void,
+  handleKeyDown: (e:KeyboardEvent) => void,
+  handleKeyUp: (e:KeyboardEvent, el:HTMLInputElement) => void,
+}
+
+export function navbarSearchComponent(): NavbarSearchComponentType {
+  return {
+    showSearchResults: false,
+    searchText: '',
+    results: [],
+    selectedIndex: -1,
+
+    get activeDecendentId() {
+      if (this.selectedIndex !== -1) {
+        return this.results[this.selectedIndex - 2].mylink.slice(1).replace('/', '-');
+      }
+      return '';
+    },
+
+    handleClickSearchResult(label: string) {
+      const input = document.querySelector<HTMLInputElement>('.searchboxinput');
+      this.searchText = label;
+      if (input) {
+        input.value = label;
+        this.showSearchResults = false;
+        input.scrollLeft = input.scrollWidth * -1;
+      }
+    },
+
+    handleKeyDown(e:KeyboardEvent) {
+      if (e.key === 'ArrowUp' || e.key === 'Enter') {
+        e.preventDefault();
+      }
+    },
+
+    async handleKeyUp(e:KeyboardEvent, el: HTMLInputElement) {
+      if (e.key === 'ArrowLeft'
+          || e.key === 'ArrowRight'
+          || e.key === 'Shift'
+          || e.key === 'Control'
+          || e.key === 'Alt'
+          || e.key === 'Meta'
+      ) return;
+
+      const {
+        value
+      } = el;
+
+      if (value.length < 2) {
+        this.showSearchResults = false;
+        this.selectedIndex = -1;
+        return;
+      }
+      const SearchBoxTopDropdown = document.querySelector('#search-listbox');
+      const SearchBoxTopDropdownOptions = SearchBoxTopDropdown?.childNodes;
+      if (SearchBoxTopDropdownOptions === undefined) return;
+
+      switch (e.key) {
+        case 'ArrowUp':
+          if (this.showSearchResults) {
+            if (this.selectedIndex === -1 || this.selectedIndex === 2) {
+              this.selectedIndex = SearchBoxTopDropdownOptions.length - 2;
+            } else {
+              this.selectedIndex--;
+            }
+          }
+
+          return;
+        case 'ArrowDown':
+          if (this.showSearchResults) {
+            if (this.selectedIndex === -1 || this.selectedIndex === SearchBoxTopDropdownOptions.length - 2) {
+              this.selectedIndex = 2;
+            } else {
+              this.selectedIndex++;
+            }
+          } else {
+            this.showSearchResults = true;
+          }
+          return;
+        case 'Enter':
+          if (this.showSearchResults) {
+            this.showSearchResults = false;
+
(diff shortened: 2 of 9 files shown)


resources/views/layouts/components/head.blade.php
```diff
@@ -43,6 +43,9 @@
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.8.1/jquery.min.js"></script>
     <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.8.23/jquery-ui.min.js"></script>
 
+    <script src="https://cdn.jsdelivr.net/npm/@floating-ui/core@1.6.0"></script>
+    <script src="https://cdn.jsdelivr.net/npm/@floating-ui/dom@1.6.1"></script>
```
wescopeland, line 47: I would strongly prefer that we did not ship Floating UI to consumers through both our Vite bundle and the document head.

wescopeland (changes requested):

resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
wescopeland, line 1: Search needs to be debounced to not fire immediately on every keypress after the minimum character count has been reached.

resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
```diff
@@ -0,0 +1,149 @@
...
+        });
+      }
+    },
+
+    get activeDecendentId() {
+      if (this.selectedIndex !== -1) {
+        return this.results[this.selectedIndex - 2].mylink.slice(1).replace('/', '-');
+      }
+      return '';
+    },
```
wescopeland, line 51: Is it possible that due to `- 2`, `this.results` could be accessed at a negative index? We may want to have some comments describing why this getter is functioning the way it does; currently my brain is having difficulty reasoning about the logic.

resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
```diff
@@ -0,0 +1,149 @@
...
+  showSearchResults:boolean,
+  searchText:string,
+  results: SearchResult[],
+  selectedIndex:number,
+  activeDecendentId: string,
+  init: (formEl: ReferenceElement, ulEl:HTMLElement)=>void,
+  handleClickSearchResult: (label: string) => void,
+  handleKeyDown: (e:KeyboardEvent) => void,
+  handleKeyUp: (e:KeyboardEvent, el:HTMLInputElement) => void,
+}
```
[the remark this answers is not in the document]
wescopeland: nit: Let's also rename from `NavbarSearchComponentType` to `NavbarSearchComponentProps`. `Type` is probably redundant given a `type` or `interface` keyword, however `Props` does accurately describe the intention for how the model should be used.

resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
```diff
@@ -0,0 +1,149 @@
...
+type NavbarSearchComponentType = {
+  showSearchResults:boolean,
+  searchText:string,
+  results: SearchResult[],
+  selectedIndex:number,
+  activeDecendentId: string,
+  init: (formEl: ReferenceElement, ulEl:HTMLElement)=>void,
+  handleClickSearchResult: (label: string) => void,
+  handleKeyDown: (e:KeyboardEvent) => void,
+  handleKeyUp: (e:KeyboardEvent, el:HTMLInputElement) => void,
```
wescopeland, line 18: There is a subtle type mismatch here. `handleKeyUp` is async, thus it will always return a Promise under the hood:

```suggestion
  handleKeyUp: (e:KeyboardEvent, el:HTMLInputElement) => Promise<void>,
```

resources/css/search.css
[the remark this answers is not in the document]
jacobguinther: Let me clarify why I originally put them in `search.css`

The class `listbox-item--hover` is used to style the options when they are selected using the directional keys. So I think that needs to stay in `search.css`. As for the `li` styling, the fact that Tailwind uses `rem` for padding and `px` for borders makes the following difficult:

WAI-ARIA Combobox
>Note: Because transparent borders are visible on some systems with operating system high contrast settings enabled, transparency cannot be used to create a visual difference between the element that is focused and other elements. Instead of using transparency, the focused element has a thicker border and less padding. When an element receives focus, its border changes from zero to two pixels and padding is reduced by two pixels. When an element loses focus, its border changes from two pixels to two and padding is increased by two pixels.

```sh
# With the site base font at 13px
# If we use:
border-2 border-width: 2px;

# Then we are left with:
p-0.5	padding: 0.125rem; 1.625px
p-1     padding: 0.25rem;  3.25px
p-1.5   padding: 0.375rem; 4.875px
p-2     padding: 0.5rem;   6.5px
```

 Am I missing something?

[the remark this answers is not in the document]
wescopeland: > The class `listbox-item--hover` is used to style the options when they are selected using the directional keys. So I think that needs to stay in `search.css`.

Isn't it just a classname that's applied when `selectedIndex - 2 === i`? If so, can't we apply any classname, including Tailwind classnames, when this condition is satisfied?

> As for the `li` styling, the fact that Tailwind uses `rem` for padding and `px` for borders makes the following difficult

Those are Tailwind's defaults, but we can coerce Tailwind into using whatever we want with its JIT functionality, including using hard pixel values, eg:

```
class="border-2 rounded-lg py-px px-[4px]"

maps to:
-> border-width: 2px;
-> border-radius: 0.5rem;
-> padding-top: 1px; padding-bottom: 1px;
-> padding-left: 4px; padding-right: 4px;
```

app/Helpers/database/user-activity.php
wescopeland, line 1: It looks like there are at least 4 unrelated commits in the branch that are associated with other units of work. I believe these would be included in a squashed changeset, which could make `git blame` unclear.

wescopeland: No problem at all, it happens. Thanks for resolving.

resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
jacobguinther: I did a small refactor that should make this easier to understand.

resources/js/alpine/navbarSearchComponent/navbarSearchComponent.ts
wescopeland: The latest revision is better and elegant.

There is compelling Human-Computer Interface research floating around which suggests the optimal debounce time for search inputs sits somewhere between 200ms and 300ms. This appears to be based on the median reaction times that sampled people tend to have in various studies.

Based on this research, Alpine.js itself appears to default debounces to 250ms. It's probably safe for us to also push this down from 500 to Alpine's default (which can be done just by excluding a time after `.debounce`).

resources/views/layouts/components/head.blade.php
```diff
@@ -51,4 +51,4 @@
     {{-- END v1 compat --}}
 
     @stack('head')
-</head>
+</head>
```
wescopeland, line 54: nit: Whatever changed here can probably be reverted so it doesn't make it into this PR's changeset.

wescopeland:
Thanks for working diligently through my feedback. I promise it's to ensure we can make this feature the highest quality it can be.

While doing my final round of testing, I noticed two behavior changes.

1: The input is no longer synchronized with the selected item while doing keyboard navigation.

2: Very minor, but there appears to have been a subtle change to z-indexing. This can be observed by hovering over the "Manage" menu when the dropdown is visible.

**Prod**
Screenshot 2024-03-03 at 10 42 40 AM

**This branch**
Screenshot 2024-03-03 at 10 42 29 AM

jacobguinther:
The behavior changes should be fixed now. I tested the z-index on the 'games' menu instead of 'manage' because I am not logged in so you might want to double-check that one again.

wescopeland (approved):

wescopeland:
Hi @jacobguinther 

There is one merge conflict to resolve, then we should be ready to merge. Thanks!

PR #2258, opened 2024-02-23
style: modernize the eslint rules

wescopeland:
This PR adjusts some lint rules in _.eslintrc.yaml_ and then executes `npm run lint:fix` across the project's TypeScript code.

The major changes are:
* Removed `airbnb-base`, as it does not play nicely with TypeScript. This also allows for the removal of `import` rule exceptions.
* Added `eslint:recommended`, `plugin:prettier/recommended`, and `simple-import-sort`.

**Why?**
Our current ESLint configuration does not enforce stylistic formatting.

As I've developed and maintained the project's TypeScript files, I have always had a _.prettierrc_ in my local repository. I've been manually formatting code, as well as sorting import and export orders, by hand. I've been maintaining this formatting in the TypeScript files from the beginning.

Now that other developers are working in the files, it makes sense to keep this consistent.

With this PR's change, Prettier style rules are enforced purely using ESLint and ESLint fixers. Developers don't need to run or execute Prettier commands.

Prettier is fairly ubiquitous in the JS ecosystem, and it's more unusual to see projects that _don't_ use it. I know there is a desire to migrate to Livewire, but as long as there's TypeScript in the project it makes sense to have consistent formatting rules, which our current ESLint configuration does not provide.


luchaos:
airbnb is referenced in https://github.com/RetroAchievements/RAWeb/blob/master/docs/CONTRIBUTING.md?plain=1#L28

luchaos (approved):

wescopeland:
Will merge after https://github.com/RetroAchievements/RAWeb/pull/2257 is resolved.

PR #2259, opened 2024-02-23
fix(game): use correct subset player count if parent game has no players

wescopeland:
This PR resolves an issue that can currently be observed here:

https://retroachievements.org/game/15689

Screenshot 2024-02-23 at 5 15 33 PM

The site suggests that there are 0 players, when in fact there are many.

**Root Cause**
This is occurring because the parent game ID has 0 players, and that player count is being inherited. To resolve this, in the rare circumstance that we detect there are 0 players in the parent set, fallback to using the original set's player count statistics.


app/Helpers/database/game.php
```diff
@@ -108,6 +108,15 @@ function getGameMetadata(
...
         $gameDataOut['NumDistinctPlayers'] = $gameMetrics['NumDistinctPlayers'] ?? 0;
 
+        // Handle the rare situation where there is a parent game ID that doesn't have a set.
+        // We should only fall into this block if there's a subset linked to a parent game ID
+        // that actually has no players. In this case, use the subset's true player count.
+        if ($parentGameId && $gameDataOut['NumDistinctPlayers'] === 0) {
+            $query = "SELECT players_total AS NumDistinctPlayers FROM GameData WHERE ID=" . $gameID;
+            $gameMetrics = legacyDbFetch($query);
+            $gameDataOut['NumDistinctPlayers'] = $gameMetrics['NumDistinctPlayers'] ?? 0;
+        }
```
Jamiras, line 118: Is this block actually necessary? The UpdateGameMetrics code should set the subset's player_total to always be non-zero.

wescopeland: You're right - this isn't necessary. The UpdateGameMetrics change should naturally resolve the issue. Not sure why I decided to add this and didn't make that connection. Anyways, I've reverted these _game.php_ changes in latest.

Jamiras (approved):

PR #2260, opened 2024-02-24
feat(multiset): add sync script for achievement sets

wescopeland:
This PR introduces foundational work required in the database layer to support the nuanced management of multiple achievement sets per game.

## Abstract
We aim to allow players to simultaneously pursue multiple achievement sets within a single game session. This initiative addresses the existing limitation of being confined to a single achievement set per ROM hash.

Multiset functionality is predicated on several key considerations that underpin the system's design and user interaction with achievement sets. From the user perspective, there will be three types of sets: Core, Bonus, and Challenge.

* **Core and Bonus Sets.** Bonus sets are designed to directly complement and supplement Core sets. When a user loads a ROM hash for a Core set, we aim to allow them to unlock achievements from Bonus sets too. This can be considered a "core to subset" relationship.
* **Challenge Sets.** Unlike Bonus sets, Challenge sets will continue to require loading a unique hash. However, players will be permitted to earn achievements from the Core set and any applicable Bonus sets simultaneously when the Challenge set hash is loaded. This can be considered a "subset to core" relationship.

---

This is a Phase One PR which satisfies:
* Build a Sync Script
* Supported `type` values

A new sync script has been developed to synchronize three V2 tables within the database:
* `game_achievement_sets`
* `achievement_sets`
* `achievement_set_achievements`

In RAWeb currently, every achievement set regardless of its intended function is actually treated as a Core set. The sync script honors this behavior when populating the `type` values of `game_achievement_sets`: they are initially all set to "core".

**How to test this PR**
You can queue all the jobs to synchronize your database tables by executing:
```bash
sail artisan ra:platform:game:populate-core-set
```

You can target specific game IDs with an optional parameter:
```bash
sail artisan ra:platform:game:populate-core-set 1
```

After jobs are finished executing, you should be able to observe that the data for those three aforementioned tables in your database is correct.


wescopeland:
Putting this in draft. https://github.com/RetroAchievements/RAWeb/pull/2283 should likely be prioritized first, as we have a requirement to build a set/hash compatibility matrix.

PR #2265, opened 2024-02-28
feat(api): add API_GetRecentGameAwards endpoint

wescopeland:
This PR adds a new endpoint, API_GetRecentGameAwards, to the app.

**Why?**
Currently without making numerous calls to user endpoints, it's difficult to determine if a tracked list of users has been awarded a game award, which can be helpful for things like Discord announcements. With this new endpoint, it's possible to watch just this single list, and if any users the consumer cares about has been granted an award, an announcement can be driven off that information.

**How?**
```
https://retroachievements.org/API/API_GetRecentGameAwards.php?y=...&z=...
```

| Param | Required? | Description                                                                                                                                                                |
| :--- | :-------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `z`  | Yes       | Your username.                                                                                                                                                             |
| `y`  | Yes       | Your web API key.                                                                                                                                                          |
| `d`  |           | Starting date (YYYY-MM-DD) (default: now).                                                                                                                                 |
| `o`  |           | Offset, number of entries to skip (default: 0).                                                                                                                            |
| `k`  |           | A comma-separated list of desired game award kinds. Possible values are "beaten-softcore", "beaten-hardcore", "completed", and "mastered" (default: all game award kinds). |

**Response**
```json
{
  "Count": 25,
  "Total": 172318,
  "Results": [
    {
      "User": "renanbrj",
      "AwardKind": "mastered",
      "AwardDate": "2022-01-01T23:48:04+00:00",
      "GameID": 14284,
      "GameTitle": "Batman Returns",
      "ConsoleID": 15,
      "ConsoleName": "Game Gear"
    },
    // ...
  ]
}
```


wescopeland:
Documented in https://github.com/RetroAchievements/api-docs/pull/39.

public/API/API_GetRecentGameAwards.php
```diff
@@ -0,0 +1,117 @@
...
+                });
+            }
+        }
+    });
+}
+
+// Now, actually make the fetch, which includes both a limit and the offset.
+$fetchedGameAwards = (clone $baseQuery)->orderBy('AwardDate', 'desc')
+    ->skip($offset)
+    ->limit(25)
```
Jamiras, line 82: This restriction should be documented in the header comments somewhere. Maybe the Count return value:
```
 *  int         Count                       number of game awards returned in the response (maximum: 25)
```

Would it be desirable to have a count parameter that defaults to 25 with some reasonable upper limit?

public/API/API_GetRecentGameAwards.php
```diff
@@ -0,0 +1,117 @@
...
+    'd' => ['sometimes', 'date', 'before_or_equal:today'],
+    'o' => ['sometimes', 'integer', 'min:0', 'nullable'],
+    'k' => ['sometimes', 'string'],
+]);
+
+$offset = $input['o'] ?? 0;
+$targetDate = $input['d'] ?? null;
+$awardKindsCsv = $input['k'] ?? '';
+
+// Any given kind which isn't included in $validAwardKinds will just be ignored.
```
Jamiras, line 42: If all values are invalid, should it return 0 records instead of matching all types?

public/API/API_GetRecentGameAwards.php
wescopeland: > Would it be desirable to have a count parameter that defaults to 25 with some reasonable upper limit?

Good idea. I've added a `c` param which defaults to 25 and has a maximum of 100. If a user gives a value above the maximum, rather than throw an error, we do:

```php
$count = min($count, 100)
```

which forces it down to 100.

Due to this change, I've documented the restriction on the `c` param.

public/API/API_GetRecentGameAwards.php
wescopeland: Yes, good idea. That would honor the principle of least astonishment. I've made this adjustment in latest:

```php
$validAwardKinds = ['beaten-softcore', 'beaten-hardcore', 'completed', 'mastered'];
$awardKinds = explode(',', $awardKindsCsv);
$awardKinds = array_filter($awardKinds, fn ($kind) => in_array($kind, $validAwardKinds, true));

// If award kinds are provided but they're all invalid, return 0 records.
if (!empty($awardKindsCsv) && empty($awardKinds)) {
    return response()->json([
        'Count' => 0,
        'Total' => 0,
        'Results' => [],
    ]);
}
```

Now if I filter by something like "car", I get the 0 records response.

Jamiras (approved):

PR #2266, opened 2024-02-29
feat(api): add new filter options and fields to API_GetConsoleIDs endpoint

WedgeSparda:
Added a couple of options in order to filter the response from `API_GetConsoleIDs`:
- `a`: If 1, only active systems will return. Default to 0.
- `g`: If 1, only game systems will return (not Hubs, Events, etc). Default to 0.

Also, added two new fields to the response indicating the value of these properties for each element in the response:

```json
[
  {
    "ID": 1,
    "Name": "Mega Drive",
    "IconURL": "https://static.retroachievements.org/assets/images/system/md.png",
    "Active": true,
    "IsGameSystem": true
  }
  // ...
]

### CODE REVIEW

File: app/Helpers/database/system.php
Comment (wescopeland) [inline app/Helpers/database/system.php:1]: You couldn't have known this, but we're actually in the process of moving away from all helper functions.

This functionality is thin enough that it should live in the endpoint itself.

File: app/Platform/Enums/SystemFlag.php
Comment (wescopeland) [inline app/Platform/Enums/SystemFlag.php:1]: This may be redundant. How about `isValidConsoleId()`?

File: app/Helpers/helpers.php [hunk 1]
```diff
@@ -22,6 +22,7 @@
 require_once __DIR__ . '/database/site-award.php';
 require_once __DIR__ . '/database/static.php';
 require_once __DIR__ . '/database/subscription.php';
+require_once __DIR__ . '/database/system.php';
```
wescopeland, line 25:
```suggestion
```

wescopeland (commented):
Cursory glance to help steer in the right direction

app/Helpers/database/system.php
WedgeSparda: Good to know. I'll move that logic out to the endpoint.

app/Platform/Enums/SystemFlag.php
WedgeSparda: After my last push I was thinking I that enum is needed at all. So, I'm thinking about removing it and just a plain int instead. What do you think?

wescopeland: Unfortunately, it may not work. `active` exists on the table, but in prod all values are `NULL`. 🫣

wescopeland: Correction: `active` is no longer `NULL` in prod on this table.

WedgeSparda: Is casted to bool in the response of the endpoint, is that what you mean?

wescopeland: Yeah. The column's data type is `tinyint(1)`, which is synonymous with a boolean. The endpoint is probably more ergonomic to use if we cast it up front for consumers rather than them knowing only 0 and 1 are the possible return values.

```php
$myVar = 1;
echo (bool) $myVar; // "true"
```

public/API/API_GetConsoleIDs.php
wescopeland, line 1: With these changes, it's probably time to also add more coverage to _ConsoleIDsTest.php_.

Some test cases that might be good:
* Happy path without new query params (add new fields to the existing test case)
* Validation on the new query params
* Correct outputs given specific query params

app/Platform/Enums/SystemFlag.php
wescopeland: Note: With all this in mind, `isValidConsoleId()` is still the source of truth for this information currently in the code. We should probably continue to rely on that for this new field for the time being.

app/Helpers/database/system.php
WedgeSparda: Done

public/API/API_GetConsoleIDs.php
WedgeSparda: Added some tests as you said. Let me know if I'm missing some case.

app/Platform/Enums/SystemFlag.php
WedgeSparda: But I see `isValidConsoleId()` is returning `true` for `Hubs` and `Events`, which are not real consoles. What I meant with this change is to filter out those "fake" consoles.

wescopeland: `isValidConsoleId()` should be returning `false` for Hubs:

https://github.com/RetroAchievements/RAWeb/blob/master/app/Helpers/database/release.php#L96

Events is currently a valid console, as its "games" can appear on user profiles and users can master event "games".

WedgeSparda: My mistake, `Hubs` was not returned since is `false`.
My point is I want this endpoint to get a list exactly like the one under the `Games` menu (as seen [here](https://api-docs.retroachievements.org/v1/get-console-ids.html)), but since `Events` is not there, I though using `isGameSystem` function was the best approach.

tests/Feature/Api/V1/ConsoleIDsTest.php
```diff
@@ -32,18 +32,74 @@ public function testGetConsoleIds(): void
...
+        $this->get($this->apiUrl('GetConsoleIDs', ['g' => 1]))
+            ->assertSuccessful()
+            ->assertJsonCount(1)
+            ->assertJson([
+                [
+                    'ID' => $system1->ID,
+                    'Name' => $system1->Name,
+                    'IconURL' => $system1->icon_url,
+                    'Active' => boolval($system1->active),
+                    'IsGameSystem' => System::isGameSystem($system1->ID),
```
wescopeland, line 102: I think we should remove the `::isGameSystem()` calls from these `assertJson`s and explicitly check for `true` or `false`. This will cause a test failure if someone makes an actual behavior change.

Right now, if the Events system ID were to somehow change, tests would continue passing while API consumers might actually break. I think a test failure in that circumstance is desirable.

app/Platform/Enums/SystemFlag.php
WedgeSparda: @wescopeland 
I'm not sure if its a better way to distinct `Events` from the rest of real consoles, apart from using `System::isGameSystem`

wescopeland: `System::isGameSystem` seems fine to me. It's more `->active` that I'm still concerned about. The source of truth in the code currently for a system being active is not the `active` value (even though it's populated in prod), it's `isValidConsoleId()`.

WedgeSparda: Ah, I see what you mean now. I'll change it

WedgeSparda: So, in order to have the more accurate data, instead of fetching systems from the database using the “active” flag, is better to fetch all and then filter with isValidConsoleId() ?

luchaos: Ideally we'll make the database value the source of truth - that's been the goal at least.

WedgeSparda: So, do you think I should change the code to rely on isValidConsoleId() for now or leave it with “active” flag sin is the path to follow in the future?

luchaos: please use `isValidConsoleId()` for now as wes suggested. the move to the database flag will follow at a later time as it needs more tests.

WedgeSparda:
I'm facing a problem when running `phpstan` before pushing.

```
 ------ ------------------------------------------------------------------------------------------------------------------------ 
  Line   app/Providers/AppServiceProvider.php                                                                                    
 ------ ------------------------------------------------------------------------------------------------------------------------ 
  108    mysqli_connect(): php_network_getaddresses: getaddrinfo for mysql failed: nodename nor servname provided, or not known  
  108    mysqli_connect(): php_network_getaddresses: getaddrinfo for mysql failed: nodename nor servname provided, or not known  
 ------ ------------------------------------------------------------------------------------------------------------------------
```

But I didn't change anything from `AppServiceProvider`. Any idea?

wescopeland:
You should be good to open an accompanying PR in `api-docs` for these changes now.

WedgeSparda:
@wescopeland Docs PR opened https://github.com/RetroAchievements/api-docs/pull/42

public/API/API_GetConsoleIDs.php
```diff
@@ -4,20 +4,36 @@
...
+ *    g : only game systems - 1 for only game systems, 0 for all system types (Events, Hubs, etc) (default: 0)
  *
  *  array
  *   object    [value]
- *    string    ID                  unique identifier of the console
+ *    int       ID                  unique identifier of the console
  *    string    Name                name of the console
  *    string    IconURL             system icon URL
+ *    bool      Active              boolean value indicating if the console is active in RA
+ *    bool      IsGameSystem        boolean value indicating if is a game system (not Events, Hubs, etc.)
```
wescopeland, line 16: nit: These lines don't need to respecify that the value is a boolean. The datatype is already listed.

```suggestion
 *    bool      Active              indicates if the system is active in RA
 *    bool      IsGameSystem        indicates if the system is a game system (not Events, Hubs, etc.)
```

public/API/API_GetConsoleIDs.php
```diff
@@ -4,20 +4,36 @@
...
+$onlyActive = (bool) request()->query('a', '0');
+$onlyGameConsoles = (bool) request()->query('g', '0');
+
+$systems = System::all()
+    ->filter(function ($system) use ($onlyActive) {
+        return $onlyActive ? isValidConsoleId($system->ID) : true;
+    })
+    ->filter(function ($system) use ($onlyGameConsoles) {
+        return $onlyGameConsoles ? System::isGameSystem($system->ID) : true;
+    });
```
wescopeland, line 28: It's more efficient to do this filtering at the query level.

WedgeSparda: Since there is no database fields to query around, how it would be? Sorry, don't have much experience with Laravel.

wescopeland: Ah, I was mistaken. You're right. Apologies.

wescopeland (approved):

WedgeSparda:
Any ETA for this to be merged?

wescopeland:
We don't have a strict release schedule, however I don't believe we plan on doing a release this week.

PR #2278, opened 2024-03-03
feat(api): add TrueRatio field to recent achievement endpoints

wescopeland:
This PR adds the `TrueRatio` field to the following endpoint responses:
* API_GetAchievementsEarnedBetween
* API_GetUserRecentAchievements


tests/Feature/Api/V1/UserRecentAchievementsTest.php
```diff
@@ -44,6 +44,11 @@ public function testGetUserRecentAchievements(): void
         $unlock3Date = $now->clone()->subMinutes(1);
         $this->addSoftcoreUnlock($this->user, $achievement3, $unlock3Date);
 
+        // Refresh the achievements.
+        $achievement1 = Achievement::findOrFail($achievement1->id);
```
luchaos, line 48: just $achievement->fresh() should work, too, if i'm not mistaken.

luchaos (approved):

tests/Feature/Api/V1/UserRecentAchievementsTest.php
wescopeland: Interesting! I didn't know about this. It appears there are two methods to choose from: `fresh()` and `refresh()`.

`fresh()` fetches a new instance and assigns it, eg:
```php
$achievement = $achievement->fresh();
```

`refresh()` is even more minimal. It updates the current instance in-place, eg:
```php
$achievement->refresh();
```

Latest has been updated to use `refresh()`.

wescopeland:
Documented at https://github.com/RetroAchievements/api-docs/pull/40

PR #2283, opened 2024-03-10
refactor(multiset): convert GameHashLibrary to game_hashes

wescopeland:
This is a preparation/precursor PR to creating a new `achievement_set_game_hashes` pivot table.

This PR takes care of almost all TODOs in the `GameHash` model:
* `GameHashLibrary` has been renamed to `game_hashes`
* Several columns have been renamed:
  * `MD5` -> `md5`
  * `Name` -> `name`
  * `GameID` -> `game_id`
  * `Created` -> `created_at`
  * `Labels` -> `labels`
* `game_id` is now an actual foreign key, linked to the `GameData` table.

Additionally, in the interest of being forward-facing, a new column has been added: `user_id`. After the migration runs, the following SQL query will populate the column:
```sql
UPDATE game_hashes gh
JOIN UserAccounts ua on gh.User = ua.User
SET gh.user_id = ua.ID;
```

A follow-up PR can drop the `User` column.

This PR has no impact on `game_hash_sets` or `game_hash_set_hashes`. Those two tables will be dropped in a subsequent PR.


app/Helpers/database/hash.php
```diff
@@ -1,142 +1,91 @@
...
+        ->select('md5', 'name', 'labels', 'user_id')
+        ->orderBy('name')
+        ->orderBy('md5')
+        ->get()
+        ->map(function ($hash) {
+            return [
+                'Hash' => $hash->md5,
+                'Name' => $hash->name,
+                'Labels' => $hash->labels,
+                'User' => $hash->user ? $hash->user->User : null,
```
luchaos, line 41: Can be shortened to `'User' => $hash->user?->User,`

app/Helpers/database/hash.php
```diff
@@ -1,142 +1,91 @@
...
-    return $retVal;
+    $hashes = $game->hashes()
+        ->with('user')
+        ->select('md5', 'name', 'labels', 'user_id')
+        ->orderBy('name')
+        ->orderBy('md5')
+        ->get()
+        ->map(function ($hash) {
+            return [
+                'Hash' => $hash->md5,
```
luchaos, line 38: The idea was to replace the `md5` column usage with the `hash` column in `game_hashes` as `md5` might be ambiguous. Sources like `nointro` also provide md5 hashes which are not the same as the "RA" hash for game detection. `hash` would become the main reference regardless of what was hashed and how, depending on the system.

app/Helpers/database/hash.php
wescopeland: Thanks, adjusted in latest 👍

app/Helpers/database/hash.php
wescopeland: Thanks for this additional background info. For full disclosure, this is an Eloquent ORM reimplementation of the original query:

```php
$query = "SELECT MD5 AS Hash ...
```

I figured it made sense to go ahead and migrate all of these helpers to Eloquent ORM while I was picking away in here. Agreed that migration to the `hash` field should be left to future work.

Maybe I should add a TODO to the `GameHash` model which says something like `md5` will eventually become `hash`?

luchaos: Sounds good. To be sure - the `md5` column would stay but repurposed to hold external `md5` values as soon as all contents and usage of it have been moved over to `hash`.

wescopeland: How does this sound? (not committed yet)

```php
// TODO migrate functionality from md5 to hash. ensure md5 retains external md5 values and hash is the primary reference for game detection
```

luchaos: That's perfect!

wescopeland: Great! Committed in latest.

luchaos (approved):
The command can be removed after release. We went the route of just providing sql statements in the past as well. Would this suffice in this case, too?

wescopeland:
Agreed. In the latest commit, I've removed the command and I've updated the OP with the SQL query to execute after the migration runs.

Jamiras:
```
UPDATE game_hashes gh
JOIN UserAccounts ua on gh.user_id = ua.ID
SET gh.user_id = ua.ID;
```

This query seems wrong. It doesn't make sense to update the column being used for the join. I suspect you wanted to join on `gh.User=ua.User`.

Jamiras (approved):

wescopeland:
@Jamiras Yep, you're right. Thanks for catching that. Updated in OP to:

```sql
UPDATE game_hashes gh
JOIN UserAccounts ua on gh.User = ua.User
SET gh.user_id = ua.ID;
```

Issue #1451, opened 2023-03-30
Game Compare page Table Overflow


BahamutVoid:
**Describe the bug**
Game Compare pages can have the Achievements column too wide and makes the page require vertical scrolling.

**To Reproduce**
1. Go to a Game Compare page with long achievements descriptions, [for example](https://retroachievements.org/gamecompare.php?ID=8302&f=Nezperdian).

**Expected behaviour**
All information should be visible without vertical scrolling.

**Screenshots**
![image](https://user-images.githubusercontent.com/18756430/228840177-823563af-42c2-48f9-841c-1a185a15d6c3.png)

**Desktop:**
 - Windows 11
 - Vivaldi

wescopeland:
Before attempting to resolve this I want to be sure I understand and can properly reproduce the issue. Here is what I see on the URL provided:

<img width="1019" alt="Screenshot 2023-04-15 at 9 42 13 AM" src="https://user-images.githubusercontent.com/3984985/232227689-5584481e-2f59-48de-869c-6df067cce4fb.png">

While the styles are indeed jank, to the best of my knowledge it doesn't appear any data is missing or truncated. There is a horizontal scrollbar, but it only moves the table ~6 pixels and doesn't seem to be revealing any hidden information.

As of the latest site release, is this also what you're seeing? Or are you still seeing data/images being truncated?

anegoda1995:
Confirm it.
To reproduce it, please use Dev Console in your browser and use a width in the range 1024 - 1279
The page needs some tweaks for the width range 1024 - 1279
<img width="1104" alt="image" src="https://github.com/RetroAchievements/RAWeb/assets/10788396/e01a4827-5b0e-4a04-a9b5-b8f88f6f627d">
<img width="1326" alt="image" src="https://github.com/RetroAchievements/RAWeb/assets/10788396/a343ac7d-110e-4b03-a3b4-fd28c7554ab4">

wescopeland:
Resolved by #2245.

PR #2284, opened 2024-03-11
fix: overflow-x when the table is so huge

gboquizosanchez:
This PR going to solve or put a little fix into https://github.com/RetroAchievements/RAWeb/issues/1451

Now it's seen:

image

With the fix:

image

The overflow fixed the content to the box. And solve possible problems in others displays.


wescopeland:
This may conflict with https://github.com/RetroAchievements/RAWeb/pull/2245.

gboquizosanchez:
Then I close it

PR #2286, opened 2024-03-12
feat(link-preview): add description and image to game links in mature hub

drisc:
Following discussion [here](https://discord.com/channels/310192285306454017/1216580587884908604), added mature hub icon to link preview for games in mature hub and added description to show amount of achievements.


resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -295,6 +295,8 @@
 @if ($gate)
     <x-app-layout
         :pageTitle="$pageTitle"
+        :pageDescription="$pageDescription ?? null"
+        :pageImage="media_asset('/Images/044930.png')" {{-- hardcoded to use the mature hub icon --}}
```
wescopeland, line 299: I think this should be hosted locally rather than live on as a media asset.

wescopeland: Either that, or it should pull the Adult Hub `Game` entity and fetch the badge from there since it is the ultimate source of truth. Whichever approach we take, the value shouldn't be hardcoded.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -293,10 +293,14 @@
 <?php endif ?>
 
 @if ($gate)
+    <?php
+        $matureHubIcon = Game::where('Title', '[Theme - Mature]')->get(['ImageIcon']);
```
wescopeland, line 297: This is overfetching for the entire model and then narrowing it to a single field. `value()` will perform better.

drisc: Change applied in latest.

wescopeland (approved):
Verified locally.

Screenshot 2024-03-12 at 4 54 03 PM

PR #2292, opened 2024-03-15
fix: remediate layout issues with forums/messages reply ui

wescopeland:
Modifies the reply UI for forums & messages to resolve some layout issues, especially on mobile.

**Before**
Screenshot 2024-03-15 at 5 35 30 PM

Screenshot 2024-03-15 at 5 35 21 PM

**After**
Screenshot 2024-03-15 at 5 34 13 PM

Screenshot 2024-03-15 at 5 34 25 PM

### DIFF
--- a/resources/views/components/base/button/submit.blade.php
+++ b/resources/views/components/base/button/submit.blade.php
@@ -4,7 +4,6 @@
     ]) }}
     :disabled="!isValid || isSending"
 >
-    {{ svg('fas-' . ($icon ?? 'paper-plane'), 'icon') }}
     <span x-show="isSending" x-cloak>{{ __('Sending...') }}</span>
     <span x-show="!isSending">{{ (string) $slot ?: __('Submit') }}</span>
 </button>
--- a/resources/views/components/base/form-actions.blade.php
+++ b/resources/views/components/base/form-actions.blade.php
@@ -10,7 +10,7 @@
         {{ $submitLabel }}
     </x-base.button.submit>
 
-    @if($hasRequiredFields)
+    @if ($hasRequiredFields)
         <div class="{{ $largeSubmit ? 'mt-2' : 'ml-3' }} whitespace-nowrap">
             * {{ __('Required') }}
         </div>
--- a/resources/views/components/base/form-field.blade.php
+++ b/resources/views/components/base/form-field.blade.php
@@ -4,6 +4,7 @@
     'id' => null,
     'inline' => false,
     'inputType' => null,
+    'isLabelVisible' => true,
     'label' => null,
     'name' => null,
     'required' => false,
@@ -14,13 +15,13 @@ $id = $id ?: 'input_' . Str::random();
 ?>
 
 <div class="{{ $inline ? 'lg:flex' : '' }}">
-    @if($inputType !== 'hidden' && $inputType !== 'checkbox')
-        <label for="{{ $id }}" class="block {{ $inline ? 'lg:w-36 lg:pr-3 lg:text-right lg:pt-1' : 'mb-1' }} {{ $name && $errors && $errors->has($name) ? 'text-danger' : '' }}">
+    @if ($inputType !== 'hidden' && $inputType !== 'checkbox')
+        <label for="{{ $id }}" class="block {{ $inline ? 'lg:w-36 lg:pr-3 lg:text-right lg:pt-1' : 'mb-1' }} {{ $name && $errors && $errors->has($name) ? 'text-danger' : '' }} {{ $isLabelVisible ? '' : 'sr-only' }}">
             {{ $label ?: __('validation.attributes.' . strtolower($name)) }} {{ $required && !$requiredSilent ? '*' : '' }}
         </label>
     @endif
     <div class="grow">
-        @if($icon)
+        @if ($icon)
             <div class="flex flex-row">
                 <div class="form-control-prepend flex items-center" aria-hidden="true">
                     {{ $icon ? svg('fas-' . $icon, 'icon') : '' }}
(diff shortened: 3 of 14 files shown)


resources/views/components/base/button/submit.blade.php
```diff
@@ -4,7 +4,6 @@
     ]) }}
     :disabled="!isValid || isSending"
 >
-    {{ svg('fas-' . ($icon ?? 'paper-plane'), 'icon') }}
```
wescopeland, line 7: Removed this icon for the time being. Maybe it makes sense to re-add an icon slot or something later. The icon was artificially increasing the button height.

resources/views/components/base/form/textarea.blade.php
```diff
@@ -79,18 +82,30 @@
...
+                @endif
+            </div>
+
+            @if ($formActions)
+                {{ $formActions }}
             @endif
         </div>
     </div>
-
-    <div id="post-preview-{{ $id }}"></div>
```
wescopeland, line 95: Moved this preview stuff out of the textarea component. I agree it's desirable to have it in one place, but not as part of the field or the form. This was causing some layout issues when the preview did appear in select areas of the UI.

resources/views/components/base/form-field.blade.php
```diff
@@ -4,6 +4,7 @@
     'id' => null,
     'inline' => false,
     'inputType' => null,
+    'isLabelVisible' => true,
```
wescopeland, line 7: There should always be a label for non-hidden form fields, but it's also desirable to be able to manipulate the visibility of the label. Setting this prop to `false` puts the label in `sr-only` mode.

luchaos (approved):

resources/views/components/base/form/textarea.blade.php
luchaos: Couldn't this be just the default and be externalized when needed?
It's mainly about the id for me - being forced to assign ids to form elements manually isn't the best DX, i usually try to avoid that as it's error prone.

resources/views/components/base/form-field.blade.php
luchaos: Do we have use-cases where we actually want to hide the labels?
The way it is written now the label _has_ to be visible because the inputs are not meant to be used without them. I'm still not fully convinced that we should allow for labels to be hidden like that.

resources/views/components/base/button/submit.blade.php
luchaos: Is it about the icon or the layout? I preferred the button with the icon.

resources/views/components/base/form-field.blade.php
luchaos: note that to me all the forms that ever should exist for users already exist and the all require labels. i'd like to have most of the dev/mod dropdowns to migrated out of the frontend eventually which reduces the need for custom form components.

luchaos: just realized that it's about the reply textarea label - i left that visible on purpose. hiding it like this makes sense if that's preferred 👍

resources/views/components/base/button/submit.blade.php
wescopeland: It's about the layout. I like the icon.

Moving the spinner to the button is a good idea, as the spinner itself also messes with the layout.

The problem is if you have a button with an icon and a button without an icon sitting next to each other, the button with the icon is taller because the icon is increasing its height. Here's a contrived example:

Screenshot 2024-03-16 at 8 45 51 AM

I've tried wrapping the icon in a span and using all sorts of classes like h-*, max-h-*, text-* to change its height, but the icon component appears to be resistant to these efforts.

resources/views/components/base/form/textarea.blade.php
wescopeland: Agreed, that sounds like a good compromise to me 👍 

These forum/message fields are unique in that their previews depend on being disjointed from the field (as they are themselves trying to simulate the layout of how the post will appear on the page). Other textareas probably won't need that unique requirement.

_textarea.blade.php_ updated with this recommendation in https://github.com/RetroAchievements/RAWeb/pull/2292/commits/d2da07c76d6fddc70f6f6ef532faae290f83589a.

resources/views/components/base/form-field.blade.php
wescopeland: Yep, you nailed it 👍 This all comes down to the reply textarea label, which uniquely makes sense not to be visible.

resources/views/components/base/button/submit.blade.php
luchaos: got it. so if i undrrstood correctly with a slot in a separate PR we could make it so all buttons have the same height regardless of whether an icon is displayed or not?

wescopeland: I think so. At a minimum, we'll need to track down what specifically is causing the icon to artificially increase the button height. For some reason, all the classnames I threw at it in an attempt to resolve the issue had no effect.

The nuclear option would be to explicitly tweak the button height itself, but that would definitely have a negative impact on the flexibility of the reusable component.

luchaos: yeah, makes sense. thanks for looking into it! let's address this in a separate step

resources/views/components/base/form-actions.blade.php
```diff
@@ -10,7 +10,7 @@
         {{ $submitLabel }}
     </x-base.button.submit>
 
-    @if($hasRequiredFields)
+    @if ($hasRequiredFields)
```
luchaos, line 13: Interesting, blade templates are mixed in the docs with spaces. i'm not used to those, is this a preset in a formatter?

wescopeland: Just me, force of habit 😅
Happy to revert to `@if(...condition)` if preferred. The docs do seem to favor a space between for `@if`, but not for other directives. I have to confess I don't understand the nuance.

https://laravel.com/docs/11.x/blade#if-statements

luchaos: Same, only now did I notice that the docs are not consistent in that regard.
Nevermind, let's keep it like this and address this with a linter for blade files as soon as a useful/working tool and preset come along.

PR #2294, opened 2024-03-16
feat(achievement): add double writes for `user_id` column

wescopeland:
A precursor to dropping the `Author` column from the `Achievements` table.

`user_id` is now written to in `UploadNewAchievement()`.

After the migration runs, the following SQL query can be run to populate the `user_id` column:
```sql
UPDATE Achievements ach
JOIN UserAccounts ua ON ach.user_id = ua.ID
SET ach.user_id = ua.ID;
```

The `Author` column is still what everything on the site reads from. A follow-up PR can drop this column from the DB, adjust all the read targets to instead use `user_id`, and delete the command.


app/Platform/Commands/UpdateAchievementsUserIds.php
```diff
@@ -0,0 +1,60 @@
...
+
+class UpdateAchievementsUserIds extends Command
+{
+    protected $signature = 'ra:platform:achievement:update-user-ids
+                            {achievementId?}';
+    protected $description = 'Update user IDs in the Achievements table. This is a precursor to dropping the User column from the table.';
+
+    public function __construct()
+    {
+        parent::__construct();
```
luchaos, line 20: could be omitted as it's redundant.

i think a raw query would be more efficient in this case than iterating over entries in code. same result though so not too important imo.

luchaos (approved):

app/Platform/Commands/UpdateAchievementsUserIds.php
wescopeland: Good call. Command removed and SQL query added to the OP of this PR.

app/Helpers/database/achievement.php
```diff
@@ -165,10 +165,12 @@ function UploadNewAchievement(
     $consoleID = $gameData['ConsoleID'];
     $consoleName = $gameData['ConsoleName'];
     $isEventGame = $consoleName == 'Events';
-    $userPermissions = getUserPermissions($author);
+
+    $author = User::where('User', $authorUsername)->first();
+    $authorPermissions = (int) $author->getAttribute('Permissions');
```
Jamiras, line 170: Recommend using `?->` or `??` in case `authorUsername` doesn't exist.

Jamiras (approved):

app/Helpers/database/achievement.php
wescopeland: In latest, adjusted to `?->` and changed the ORM query to use `firstWhere()`.

luchaos: Unrelated heads up: the reason that we have to access the permissions attribute like this instead of just `$author->Permissions` is because of the existing `->permissions()` relationship which causes a conflict. This will resolve itself as soon as the migration to roles is complete and this column removed.

PR #2296, opened 2024-03-16
refactor(Comment): rename UserID to user_id, add foreign key

wescopeland:
**BEFORE EXECUTING THE MIGRATION, YOU MUST RUN:**
```sql
DELETE FROM Comment WHERE NOT EXISTS (
  SELECT 1 FROM UserAccounts WHERE ID = Comment.UserID
);
```

---

This PR renames `UserID` to `user_id` in the `Comment` table. Additionally, `UserID` was missing a FK relationship to the `UserAccounts` table. The migration introduced in this branch adds this relationship.

### DIFF
--- a/database/migrations/community/2024_03_16_000001_update_comment_table.php
+++ b/database/migrations/community/2024_03_16_000001_update_comment_table.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->renameColumn('UserID', 'user_id');
+        });
+
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->foreign('user_id')->references('ID')->on('UserAccounts')->onDelete('set null');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->dropForeign(['user_id']);
+        });
+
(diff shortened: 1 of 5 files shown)


database/migrations/community/2024_03_16_000001_update_comment_table.php
```diff
@@ -0,0 +1,31 @@
...
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->renameColumn('UserID', 'user_id');
+        });
+
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->foreign('user_id')->references('ID')->on('UserAccounts')->onDelete('set null');
```
Jamiras, line 17: This failed for me:
```
SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key
constraint fails (`retroachievements-web`.`#sql-alter-1-f8`, CONSTRAINT `comment_user_id_foreign` 
FOREIGN KEY (`user_id`) REFERENCES `UserAccounts` (`ID`) ON DELETE SET NULL) (Connection: mysql, 
SQL: alter table `Comment` add constraint `comment_user_id_foreign` foreign key (`user_id`) references
`UserAccounts` (`ID`) on delete set null)
```

I have several comments from users that don't exist:
```
select distinct c.user_id from Comment c left join UserAccounts u on u.ID=c.user_id where u.ID is null;
 ```

wescopeland: Thanks - I should have checked beyond my local.

I played around on Preview and confirmed there are roughly 272 orphaned comments in that env. I've updated the original post of this PR with a query that, when executed before the migration, purges the orphaned comment records.

Jamiras (approved):

PR #2297, opened 2024-03-17
refactor(Ticket): rename reporter/resolver user columns and add foreign keys

wescopeland:
**BEFORE RUNNING THIS MIGRATION, YOU MUST EXECUTE:**
```sql
-- Command One (this resolves around ~13 rows with integrity issues)
UPDATE Ticket
LEFT JOIN UserAccounts ON Ticket.ReportedByUserID = UserAccounts.ID
SET Ticket.ReportedByUserID = NULL
WHERE UserAccounts.ID IS NULL;

-- Command Two (this probably won't do anything)
UPDATE Ticket
LEFT JOIN UserAccounts ON Ticket.ResolvedByUserID = UserAccounts.ID
SET Ticket.ResolvedByUserID = NULL
WHERE UserAccounts.ID IS NULL;
```

This PR renames two columns on the `Ticket` table:
* `ReportedByUserID` -> `reporter_id`
* `ResolvedByUserID` -> `resolver_id`

Additionally, foreign key relationships are added between these columns to `UserAccounts.ID`.

`User` and `Ticket` models are updated so Eloquent ORM is aware of the relationships. I shuffled some existing code around in `ActsAsPlayer` just to properly organize/alphabetize things.

### DIFF
--- a/app/Platform/Concerns/ActsAsPlayer.php
+++ b/app/Platform/Concerns/ActsAsPlayer.php
@@ -44,7 +44,7 @@ trait ActsAsPlayer
 
     public function hasPlayed(Game $game): bool
     {
-        return $this->playerGame($game)->exists();
+        return $this->playerGames()->where('game_id', $game->id)->exists();
     }
 
     public function playerGame(Game $game): ?PlayerGame
--- a/app/Platform/Concerns/ActsAsPlayer.php
+++ b/app/Platform/Concerns/ActsAsPlayer.php
@@ -11,6 +11,7 @@ use App\Models\PlayerAchievement;
 use App\Models\PlayerBadge;
 use App\Models\PlayerGame;
 use App\Models\PlayerSession;
+use App\Models\Ticket;
 use App\Models\User;
 use Carbon\Carbon;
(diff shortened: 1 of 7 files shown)


app/Platform/Concerns/ActsAsPlayer.php
```diff
@@ -39,6 +40,18 @@ public function rollConnectToken(): void
         $this->save();
     }
 
+    // == instance methods
+
+    public function hasPlayed(Game $game): bool
+    {
+        return $this->playerGame($game)->exists();
```
Jamiras, line 47: is it safe to call `->exists()` on `null`?

wescopeland: For full disclosure, `hasPlayed()` already existed on the model. In this PR, I relocated it in the file (it was listed in the "relations" section but is more of an internal method than a relation).

However, your observation is astute. No, this usage of `->exists()` is not safe.

I've rewritten `hasPlayed()`:

```php
public function hasPlayed(Game $game): bool
{
    return $this->playerGames()->where('game_id', $game->id)->exists();
}
```

Jamiras (approved):

PR #2299, opened 2024-03-17
refactor: add user_id to SiteAward table with double writes

wescopeland:
**AFTER EXECUTING THE MIGRATION, RUN:**
```sql
-- 1, remove awards from users that don't exist anymore
DELETE FROM SiteAwards
WHERE User NOT IN (SELECT User FROM UserAccounts);

-- 2, write all SiteAwards.user_id values
UPDATE SiteAwards sa
JOIN UserAccounts ua on sa.User = ua.User
SET sa.user_id = ua.ID;
```

---

This PR adds a migration which:
* Adds a `user_id` column to the `SiteAwards` table.
* Establishes a foreign key relationship between this new column and `UserAccounts.ID`.

Additionally, everywhere a site award is granted, `user_id` is now being written to (alongside the pre-existing `User` value).


app/Helpers/database/site-award.php
```diff
@@ -172,12 +174,12 @@ function HasPatreonBadge(string $username): bool
     return mysqli_num_rows($dbResult) > 0;
 }
 
-function SetPatreonSupporter(string $username, bool $enable): void
+function SetPatreonSupporter(string $username, int $userId, bool $enable): void
```
Jamiras, line 177: Since you're already changing these signatures, how difficult would it be to pass a `User` object in?

wescopeland: Good point - not difficult at all. Addressed in https://github.com/RetroAchievements/RAWeb/pull/2299/commits/5e2b5b0a0fbb5f1765be19dec59565f8d653e0d5.

Now, a `User` entity is sent all the way down to `AddSiteAward()`.

Jamiras (approved):

PR #2300, opened 2024-03-17
feat: allow moderators to mute users

wescopeland:
This PR leverages the existing `UserAccounts.muted_until` field and Folio user management panel to allow moderators to mute users.

When a user is muted, they will no longer be able to comment, post or edit on the forum, or send DMs.

To mute a user, visit `http://localhost:64000/manage/users/[USERNAME_HERE]/edit`. There is already a field to set the `muted_until` value:

Screenshot 2024-03-17 at 5 47 23 PM

**Before**
Screenshot 2024-03-17 at 5 44 37 PM

**After**
Screenshot 2024-03-17 at 5 44 23 PM


televandalist:
We're gonna need to pull the "admin comments" soon. Got a lot of catching up to do. ;)

DawnBracken:
LuteFrontier2023 is shaking in his boots.

luchaos (approved):

luchaos:
Minor detail: it's `Filament` instead of `Folio` ;)
Thanks for wiring it all up! ❤️

wescopeland:
Terminology confusion 😵‍💫, thank you for the correction!

app/Helpers/render/comment.php
```diff
@@ -79,22 +80,22 @@ function RenderCommentsComponent(
...
-            $user,
+            $username,
             $articleTypeID,
             (int) $comment['ID'],
             $canDeleteComment
         );
     }
 
-    if (isset($user)) {
+    if (isset($user) && !$user->isMuted) {
```
Jamiras, line 98: I wonder if a muted user should be informed so they don't question why they can't add comments?

app/Helpers/database/user-auth.php
```diff
@@ -180,6 +180,7 @@ function authenticateFromCookie(
     }
 
     $userDetailsOut = $user->toArray();
+    $userDetailsOut['isMuted'] = $user->isMuted;
```
Jamiras, line 183: Why is this necessary? shouldn't it already be populated by the `toArray` call since it's in the `visible` list?

resources/views/pages-legacy/viewtopic.blade.php
```diff
@@ -195,7 +195,7 @@
     <?php
     $user = auth()->user();
     ?>
-    @if ($thisTopicID != 0 && $user?->hasVerifiedEmail())
+    @if ($thisTopicID != 0 && $user?->hasVerifiedEmail() && !$user?->isMuted)
         <x-section>
             <div class="flex w-full bg-embed p-2 rounded-lg">
                 @guest
```
Jamiras, line 201: Unrelated to this PR, but how can this `@guest` code be reached? A null user would probably not have a verified email, and definitely wouldn't be muted.

Jamiras (commented):
Should private messages also be muted? I assume we'd want to at least allow messaging RAdmin.

app/Helpers/database/user-auth.php
wescopeland: The necessity comes down to how Laravel handles attribute serialization for accessors. Although `isMuted` is indeed set up as an accessor in the model and is included in the `$visible` list, because the value is driven by an accessor and is dynamically computed, it also needs to be added to `$appends` for `->toArray()` to serialize it, eg:

```php
protected $appends = [
  'avatarUrl',
  'isMuted',
];
```

However, I'm afraid this can result in the following unexpected developer experience:

```php
$user = User::select("User", "Motto")->firstWhere("User", "Scott");

// ERROR: The attribute [muted_until] either does not exist or 
// was not retrieved for model [App\Models\User].
$user->toArray();
```

I think maybe we should just remove it from `$visible`.

app/Helpers/render/comment.php
wescopeland: Updated in latest:

Screenshot 2024-03-25 at 4 58 41 PM

resources/views/pages-legacy/viewtopic.blade.php
wescopeland: Good catch, it's unreachable. I went ahead and fixed it in latest by moving it outside the `@if` directive.

wescopeland:
@Jamiras 
> Should private messages also be muted? I assume we'd want to at least allow messaging RAdmin.

My opinion is loosely held and I could be swayed otherwise, but I think private messages (including to RAdmin) should be muted. I assume this kind of action is going to be taken against a user who has already been warned to stop. Especially now that the UI will show when the mute will be lifted, once a user is muted I'm not sure there will actually be much more to substantiate on.

resources/views/pages-legacy/viewtopic.blade.php
Jamiras: I see the new code at line 200, but the old (unreachable) code is still present here.

wescopeland: Whoops, removed in latest.

Jamiras (approved):

Jamiras:
I've noticed that there's no way for a moderator to tell that a user is muted without looking directly at the database. The field appears empty when editing the account a subsequent time, and there's no audit trail. 

Similarly, the `muted_until` field remains set even after the period has elapsed. The logic is still correct as it looks for times in the future, but it might be nice to clean that up after it's no longer needed.

wescopeland:
https://github.com/RetroAchievements/RAWeb/issues/2313

PR #2304, opened 2024-03-23
refactor(UserGameListEntry): drop User from the SetRequest table

wescopeland:
This PR adds a migration which drops `User` from the `SetRequest` table. The table already contains a populated `user_id` column with a proper foreign key relationship. This column now serves as the source of truth for all queries against the table.

Additionally:
* $user->gameList has been renamed to $user->gameLists due to this being a one-to-many relationship.
* Several helpers which use `s_mysql_query()` have been refactored to now use Eloquent ORM.
* On game pages, the `$user` model is no longer prop-drilled down to the add-to-list button component. `auth()->user()` context is used instead.


app/Community/Concerns/ActsAsCommunityMember.php
```diff
@@ -33,10 +33,19 @@ public function activities(): HasMany
...
     {
-        return $this->hasMany(UserGameListEntry::class, 'user_id', 'ID')
-            ->where('SetRequest.type', $type);
+        return $this->hasMany(UserGameListEntry::class, 'user_id', 'ID');
+    }
+
+    /**
+     * @return HasMany<UserGameListEntry>
+     */
+    public function gameLists(string $type): HasMany
```
Jamiras, line 44: This feels wrong. The name implies it would return multiple lists instead of multiple entries. Recommend overloading the first function:
```
    public function gameListEntries(?string $type = null): HasMany
    {
        $query = $this->hasMany(UserGameListEntry::class, 'user_id', 'ID');
        
        if ($type !== null)
        {
            $query->where('SetRequest.type', $type);
        }
        
        return $query;
    }
```

database/migrations/community/2024_03_22_00000_update_set_requests_table.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('SetRequest', function (Blueprint $table) {
+            $table->dropColumn('User');
+        });
```
Jamiras, line 14: This fails for me. And the column clearly does exist.
```
2024_03_22_00000_update_set_requests_table ............................................. 2ms FAIL

   Illuminate\Database\QueryException 

  SQLSTATE[42000]: Syntax error or access violation: 1072 Key column 'User' doesn't exist in table (Connection: mysql, SQL: alter table `SetRequest` drop `User`)
```

```
MariaDB [retroachievements-web]> desc SetRequest;
+------------+---------------------+------+-----+---------------------+----------------+
| Field      | Type                | Null | Key | Default             | Extra          |
+------------+---------------------+------+-----+---------------------+----------------+
| id         | bigint(20) unsigned | NO   | PRI | NULL                | auto_increment |
| user_id    | bigint(20) unsigned | YES  | MUL | NULL                |                |
| User       | varchar(32)         | NO   | MUL | NULL                |                |
```

Jamiras: Ah, there's a key that had to be dropped too.
```
UNIQUE KEY `user_game_list_entry_username_game_id_type_unique` (`User`,`GameID`,`type`),
```

```
alter table `SetRequest` drop key `user_game_list_entry_username_game_id_type_unique`;
```

app/Community/Concerns/ActsAsCommunityMember.php
wescopeland: Good idea. Updated in latest.

database/migrations/community/2024_03_22_00000_update_set_requests_table.php
wescopeland: Whoops, I missed that one. I've updated the migration to drop that unique key. This issue should be resolved now.

Jamiras (approved):

PR #2305, opened 2024-03-23
refactor(AchievementSetClaim): add user_id column and double writes

wescopeland:
**BEFORE EXECUTING THE MIGRATION, RUN:**

```sql
-- Fix ~5 data integrity issues.
DELETE FROM SetClaim
WHERE User NOT IN (SELECT User FROM UserAccounts);
```

**AFTER EXECUTING THE MIGRATION, RUN:**

```sql
-- Write all existing user IDs to the SetClaim table.
UPDATE SetClaim sc
JOIN UserAccounts ua on sc.User = ua.User
SET sc.user_id = ua.ID;
```

---

This PR:
* Adds a migration which:
  * Adds a `user_id` column (FK to `UserAccounts.ID`) to the `SetClaim` table.
  * Renames `GameID` on the `SetClaim` table to `game_id` and gives it a FK relationship to `GameData`.
* Migrates the `insertClaim()` helper function to use Eloquent ORM.
* Adds double writes to `user_id` for `SetClaim` insertions.

NOTE: I've ensured API_GetGameExtended, API_GetClaims, and API_GetUserClaims are unaffected by this change.


app/Helpers/database/set-claim.php
```diff
@@ -8,47 +8,58 @@
...
+        if ($primaryClaimFinishTime) {
+            $finishedAt = $primaryClaimFinishTime;
+            $special = ClaimSpecial::None;
+        }
+    } else {
+        // Different roles are allowed a maximum number of active primary claims.
+        // Does this user currently have fewer primary claims than that maximum?
+        $isUserAllowedToClaim = getActiveClaimCount($user->User, false) < permissionsToClaim($userPermissions);
+
+        if ($special === ClaimSpecial::None && !$isUserAllowedToClaim) {
```
Jamiras, line 44: `getActiveClaimCount` could be avoided in the `$special == ClaimSpecial::None` was moved up to the else (line 39)

database/migrations/community/2024_03_23_000000_update_setclaim_table.php
```diff
@@ -0,0 +1,47 @@
...
+            $table->dropForeign(['user_id']);
+            $table->dropColumn('user_id');
+        });
+
+        // [1] Rename GameID to game_id, establish foreign key relationship to GameData table.
+        Schema::table('SetClaim', function (Blueprint $table) {
+            $table->dropForeign(['game_id']);
+            $table->renameColumn('game_id', 'GameID');
+            $table->unsignedInteger('GameID')->change();
+        });
```
Jamiras, line 45: My rollback failed. Not sure why - this _looks_correct.
```
2024_03_23_000000_update_setclaim_table ............................................... 40ms FAIL

   Doctrine\DBAL\Schema\Exception\ColumnDoesNotExist 

  There is no column with name "gameid" on table "SetClaim".
```

app/Helpers/database/set-claim.php
wescopeland: Good catch, updated in latest.

database/migrations/community/2024_03_23_000000_update_setclaim_table.php
wescopeland: I had the same issue but assumed it was something up with my local machine, as the error is a bit bizarre.

I was able to fix this on my local after deconstructing every single operation into separate `Schema::table()` tasks and forcing Doctrine to identify `game_id` as an `unsignedBigInteger` before the rename runs. Now that rollback seems to work correctly:

```php
Schema::table('SetClaim', function (Blueprint $table) {
    $table->dropForeign(['game_id']);
});
Schema::table('SetClaim', function (Blueprint $table) {
    $table->unsignedBigInteger('game_id')->change();
});
Schema::table('SetClaim', function (Blueprint $table) {
    $table->renameColumn('game_id', 'GameID');
});
Schema::table('SetClaim', function (Blueprint $table) {
    $table->unsignedInteger('GameID')->change();
});
```

Jamiras (approved):

PR #2307, opened 2024-03-24
Convert user game activity page to service/blade

Jamiras:
Includes some minor changes to styling:
* Nested timestamps changed from "DD MM YYY, HH:MM" to "HH:MM:SS". Dates are still included for session header entries.
* Entry description text is muted

Before:
image

After:
image


config/missing-page-redirector.php
```diff
@@ -85,6 +84,8 @@
         '/userList.php' => '/users',
         '/history.php' => '/user/{u}/history',
         '/historyexamine.php' => '/user/{u}/history/{d}',
+        '/usergameactivity.php' => '/user/{f}/game/{ID}',
+        '/gamecompare.php' => '/user/{f}/game/{ID}/compare',
```
Jamiras, line 88: This change, or the changes in `recent-game-players.blade.php` are necessary before deployment. It was missed when applying feedback in #2245.

I can create a separate PR if it seems unlikely that this will be merged.

app/Platform/Services/PlayerGameActivityService.php
wescopeland, line 1: It appears there are several magic strings used for the session type (eg: `player-session`, `manual-unlock`, etc). Should these perhaps be class constants or something else that's reusable?

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -0,0 +1,298 @@
...
+            }
+        }
+
+        $this->sessions[$existingSessionIndex]['events'][] = $event;
+    }
+
+    private function insertEvent(array &$session, array $event): void
+    {
+        $session['events'][] = $event;
+        $this->sortEvents($session['events']);
```
wescopeland, line 121: This sort might be redundant, given that `$this->sortEvents($session['events'])` is called at the end of `initialize()`. If it's not redundant, I'm wondering if we can get away with each insertion not triggering a re-sort.

app/Platform/RouteServiceProvider.php
```diff
@@ -91,7 +92,7 @@ protected function mapWebRoutes(): void
             // Route::resource('user.achievements', PlayerAchievementController::class)->only('index')->names(['index' => 'user.achievement.index']);
             // Route::resource('user.games', PlayerGameController::class)->only('index')->names(['index' => 'user.game.index']);
             // Route::resource('user.game', PlayerGameController::class)->only('show');
-            // Route::get('user/{user}/game/{game}', [PlayerGameController::class, 'activity'])->only('user.game.activity');
+            Route::get('user/{user}/game/{game}', [PlayerGameController::class, 'activity'])->name('user.game.activity');
```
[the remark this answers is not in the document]
Jamiras: The route without the subclause could only be for viewing the user's progress in the game, which is already handled (indirectly) by the `/compare` path. 

I'm fine with either.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -0,0 +1,298 @@
...
+
+            $this->achievementsUnlocked++;
+        }
+
+        foreach ($this->sessions as &$session) {
+            $this->sortEvents($session['events']);
+        }
+    }
+
+    private function addUnlockEvent(object $playerAchievement, Carbon $when, bool $hardcore, bool $repeat = false): void
```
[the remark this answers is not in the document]
Jamiras: I don't remember what that was for. I've removed it.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -0,0 +1,298 @@
...
+                    // two events at same time should be sub-sorted by ID
+                    $diff = ($a['ID'] ?? 0) - ($b['ID'] ?? 0);
+                }
+            }
+
+            return $diff;
+        });
+    }
+
+    private function findSession(string $type, Carbon $when): int
```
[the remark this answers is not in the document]
Jamiras: It should replace the `player-session` in the if statement. Missed it when I was extracting the method.

app/Platform/Services/PlayerGameActivityService.php
Jamiras: There probably should be, but the enums are specific to this service/the associated blade. Would they go into the normal Enums folder?

app/Platform/Services/PlayerGameActivityService.php
Jamiras: This was supposed to insert at the appropriate index to maintain the sort, but it was easier to just append and re-sort. Without that logic, maintaining a sorted list while building it probably isn't necessary. I've removed it for the time-being.

app/Platform/Services/PlayerGameActivityService.php
wescopeland: Yeah, I think just plain _app/Enums_ is probably fine. It seems like we're moving away from concrete boundaries on our models so this doesn't seem too problematic to me.

Jamiras: done

PR #2311, opened 2024-03-27
don't exclude open tickets from TicketFilters::All

Jamiras:
Fixes "No other tickets found for this achievement" when only open tickets exist.

The basic problem is the where clause: "`ua.User <> ua2.User`", which filters out rows where `ua2.User` is null.


wescopeland (approved):
Verified locally on several tickets.

PR #2314, opened 2024-03-29
feat(game): adjust compare progress ux

wescopeland:
Makes some slight adjustments to the compare progress UX:

* Progress bars are now shown instead of just counts. This better highlights softcore, hardcore, and mixed progress.
* Users with 0 progress are no longer shown.

Additionally:
* Adjusted the component code so it adheres to 4-space indentation.
* Increased the spacing between the component and Achievement Distribution.

**Before**
Screenshot 2024-03-28 at 9 03 25 PM

**After**
Screenshot 2024-03-28 at 9 02 56 PM


resources/views/components/game/compare-progress.blade.php
```diff
@@ -86,27 +93,40 @@ function selectSearchBoxUser() {
             <p>None of your followed users have played this game</p>
```
Jamiras, line 93: This message may need to be modified.
> None of your followed users have unlocked any achievements for this game.

resources/views/components/game/compare-progress.blade.php
```diff
@@ -86,27 +93,40 @@ function selectSearchBoxUser() {
...
+                            href="{!! route('game.compare-unlocks', ['game' => $game, 'user' => $friend->User]) !!}"
+                        >
+                            <x-game-progress-bar
+                                :softcoreProgress="$completion['achievements_unlocked']"
+                                :hardcoreProgress="$completion['achievements_unlocked_hardcore']"
+                                :maxProgress="$completion['achievements_total']"
+                            />
+                            <p class="text-2xs">
+                                {{ $completion['achievements_unlocked']}} of {{ $completion['achievements_total'] }}
+                            </p>
```
Jamiras, line 117: I think it might be better without the sublabel. The tooltip only appears over the bar, and the label makes it hard to hover over the bar. The tooltip also shows the same information, but includes the hardcore/softcore breakdown. Not having this sublabel would also be more consistent with the same widget's use in the recent players list.

resources/views/components/game/compare-progress.blade.php
wescopeland: Adjusted in latest

resources/views/components/game/compare-progress.blade.php
wescopeland: Label removed. Added a way of increasing padding around the progress bar container so the tooltip is easier to surface.

Jamiras (approved):

PR #2316, opened 2024-03-30
chore(user): implement user ratings deletion todo

wescopeland:
Very minor housekeeping task as I work through the tables with user relations.

We can't write to these records anymore, but we do delete them on account deletion. This PR implements a TODO by associating `Rating` to the `User` model.

We should also populate the `user_id` records in stage & prod. They are currently unpopulated. I tested the following query on preview.

```sql
UPDATE Rating r
JOIN UserAccounts ua on r.User = ua.User
SET r.user_id = ua.ID;
```


Jamiras (approved):
Seems fine. Untested.

If we're not planning to re-enable the feature, shouldn't we just purge all the records and the related code?

wescopeland:
> shouldn't we just purge all the records and the related code?
It's probably the right thing to do. I'll clone the table data from prod and store it somewhere secure, just in case.

I'll close this PR in favor of a separate PR to drop the table, model, and related code.

PR #2320, opened 2024-03-30
chore(Message): drop UserTo and UserFrom

wescopeland:
Drops `UserTo` and `UserFrom` from the `messages` table. These fields are currently unused; all references are relying on the `author_id` foreign key.


app/Models/Message.php
```diff
@@ -12,6 +12,8 @@ class Message extends BaseModel
 {
     use Searchable;
 
+    protected $table = 'messages';
```
wescopeland, line 15: Redundant, but added to improve searchability (similar to `PlayerGame`, etc).

database/migrations/community/2024_03_30_000001_update_messages_table.php
```diff
@@ -0,0 +1,32 @@
...
+                $table->dropColumn('UserTo');
+            });
+        }
+
+        if (Schema::hasColumn('messages', 'UserFrom')) {
+            Schema::table('messages', function (Blueprint $table) {
+                $table->dropColumn('UserFrom');
+            });
+        }
+    }
```
Jamiras, line 23: There are still references to these columns in the MigrateMessages command, which is no longer needed. Please remove it as part of this PR.

wescopeland: Removed in latest.

It seems like we have a lot of dead code like this hanging around. At some point, we should probably sweep through and delete it all.

Jamiras (approved):

PR #2323, opened 2024-03-31
feat: improve user muting moderation ux

wescopeland:
This PR makes some improvements to the user mute feature for moderators:

1. Mutes are now recorded & visible in the Filament user panel audit log.

Screenshot 2024-03-31 at 11 13 33 AM

2. If a user is muted, the "Muted until" datepicker in the user management form pre-populates with the `muted_until` date.

Screenshot 2024-03-31 at 11 13 19 AM

3. If a user is muted, it's now visible in a new mod-only "Moderation Overview" section of their profile.

Screenshot 2024-03-31 at 11 13 12 AM

Additionally:
* The "Muted until" datepicker control now has a maxdate value to correspond to the `timestamp` type.
* The Moderation Overview section is driven entirely by modern roles and table values. `Permissions` and `Untracked` are both treated as nonexistent.

NOTE: This PR does not implement anything to auto-clear the `muted_until` value.


resources/views/components/user/profile/moderation-overview.blade.php
```diff
@@ -0,0 +1,46 @@
...
+    </div>
+
+    <div class="w-full p-2 bg-embed flex flex-col gap-y-2 rounded">
+        <div class="grid md:grid-cols-2 gap-x-12 gap-y-1">
+            @if ($targetUser->is_muted)
+                <x-user.profile.stat-element
+                    label="Muted until"
+                    value="{{ $targetUser->muted_until->format('Y-m-d') }}"
+                    :href="route('filament.admin.resources.users.edit', $targetUser)"
+                />
```
Jamiras, line 19: I assume it's a timezone thing, but this is showing a the value from the database:
image

Which happens to be in the past, so it probably shouldn't be shown.

But the picker shows the previous day:
image

Jamiras: Definitely seems to not be honoring the `is_muted` correctly:
image

app/Models/User.php
```diff
@@ -379,6 +390,16 @@ public function getIsMutedAttribute(): bool
...
 
+    public function getIsUnrankedAttribute(): bool
+    {
+        return $this->isUnranked();
+    }
+
+    public function getIsBannedAttribute(): bool
+    {
+        return $this->isBanned();
+    }
```
Jamiras, line 401: Should these be on ActsAsCommunityMember?

resources/views/components/user/profile-meta.blade.php
```diff
@@ -43,9 +38,16 @@
...
+@can('manage', App\Models\User::class)
     <x-user.profile.moderation-tools :targetUser="$user" />
-@endif
+@endcan
+
+@can('manage', App\Models\User::class)
+    @if ($user->isModerated())
+        <x-user.profile.moderation-overview :targetUser="$user" />
+    @endif
+@endcan
```
Jamiras, line 50: I have a feeling this is somehow tied to my muted dates from the past still being visible, but this section isn't appearing for me.

resources/views/components/user/profile/moderation-overview.blade.php
Jamiras: These are related to the manage page, not this element. I just searched for "Muted until".

resources/views/components/user/profile-meta.blade.php
wescopeland: Indeed, I think this is happening because your muted date is in the past (you're not currently muted) and there is no other moderation action on your account. The logic for `isModerated()` is:

```php
public function isModerated(): bool
{
    return
        $this->muted_until?->isFuture() // <-- !!
        || $this->unranked_at !== null
        || $this->banned_at !== null
        || $this->DeleteRequested !== null
    ;
}
```

app/Models/User.php
wescopeland: Went ahead and moved them there in latest. I kept `isModerated()` in `User`, as I figured it's plausible some moderation action has happened on the user that isn't necessarily tied to the concerns of `ActsAsCommunityMember`.

resources/views/components/user/profile/moderation-overview.blade.php
wescopeland: Should be resolved in latest:
* If `muted_until` is in the past, the value won't display in the Filament "view" panel.
* The datepicker now honors the UTC timezone for pre-populated values. This issue was caused by a client-side hydration issue with Filament's JS datepicker control.

Jamiras (approved):

PR #2326, opened 2024-04-01
refactor(LeaderboardEntry): add double writes in the helper functions

wescopeland:
Right now we maintain two tables for leaderboard entries: `LeaderboardEntry` and `leaderboard_entries`.

`leaderboard_entries` is being populated by a sync script which executes every minute.

This PR migrates several leaderboard helper functions to Eloquent ORM, and implements explicit double writes in all of the helpers responsible for writing to the legacy `LeaderboardEntry` table.

Ideally, after this is deployed to prod, we can disable the sync script and work towards dropping the `LeaderboardEntry` table altogether.


resources/views/pages-legacy/leaderboardinfo.blade.php
```diff
@@ -173,8 +173,6 @@
         // else
         //    echo "<a href='leaderboardinfo.php?i=$lbID&amp;c=$count&amp;f=1'>Show Friends Only</a> - <b>All Results</b><br><br>";
 
-        echo "<div class='larger'>$lbTitle: $lbDescription</div>";
```
wescopeland, line 176: Went ahead and removed this while I was messing with leaderboards. I noticed this is duplicative with content directly above it on the page.

app/Helpers/database/leaderboard.php
```diff
@@ -2,134 +2,165 @@
...
 
-    $userID = getUserIDFromUser($user);
-    if ($userID === 0) {
+    if (!$leaderboardEntry) {
         return false;
     }
 
-    $query = "DELETE FROM LeaderboardEntry
-              WHERE ( LeaderboardID = $lbID AND UserID = $userID )";
+    $score = ValueFormat::format($leaderboardEntry->score, $leaderboardEntry->leaderboard->Format);
```
Jamiras, line 141: I realize this is existing code, but `ValueFormat::format` returns a string. `$score` is an out `int`. The function signature should be updated.

app/Helpers/database/leaderboard.php
```diff
@@ -2,134 +2,165 @@
...
+                ->update(['Score' => $newEntry, 'DateSubmitted' => now()]);
+        }
+    } else {
+        // No existing leaderboard entry. Let's insert a new one.
+        LeaderboardEntryLegacy::create([
+            'LeaderboardID' => $leaderboard->id,
+            'UserID' => $user->id,
+            'Score' => $newEntry,
+            'DateSubmitted' => now(),
+        ]);
```
Jamiras, line 126:
```
$ curl "localhost:64000/dorequest.php?r=submitlbentry&u=Jamiras&t=[redacted]&i=74041&s=1444&m=8e3630186e35d477231bf8fd50e54cdd&v=60253d7cbe23a30f406b0be85e1ddf7e"
{"Success":false,"Error":"Add [LeaderboardID] to fillable property to allow mass assignment on [App\\Models\\LeaderboardEntryLegacy].","Status":500}
```

app/Helpers/database/leaderboard.php
```diff
@@ -2,134 +2,165 @@
...
+            // Update the player's entry.
+            $existingLeaderboardEntry->score = $newEntry;
+            $existingLeaderboardEntry->save();
+
+            $retVal['BestScore'] = $newEntry;
         } else {
-            $retVal['Success'] = false;
-            $retVal['Error'] = "Cannot insert the value $newEntry into leaderboard with ID: $lbID (unknown issue)";
+            // No change made.
+            $retVal['BestScore'] = $existingLeaderboardEntry->Score;
```
Jamiras, line 69: (same call as below, but record exists)
```
$ curl "localhost:64000/dorequest.php?r=submitlbentry&u=Jamiras&t=[redacted]&i=74041&s=1444&m=8e3630186e35d477231bf8fd50e54cdd&v=60253d7cbe23a30f406b0be85e1ddf7e" && echo
{"Success":false,"Error":"The attribute [Score] either does not exist or was not retrieved for model [App\\Models\\LeaderboardEntry].","Status":500}
```

app/Helpers/database/leaderboard.php
wescopeland: Good catch, I naively assumed this would be automatically caught and didn't think to check the function signature. Updated in latest.

app/Helpers/database/leaderboard.php
wescopeland: Looks like I accidentally didn't commit a model update which included that `$fillable` list. Resolved in latest.

app/Helpers/database/leaderboard.php
```diff
@@ -2,134 +2,165 @@
...
     }
+}
 
-    $score = ValueFormat::format((int) $data['Score'], $data['Format']);
+function removeLeaderboardEntry(User $user, int $lbID, ?string &$score): bool
+{
+    $leaderboardEntry = LeaderboardEntry::with('leaderboard')
+        ->where('leaderboard_id', $lbID)
+        ->where('user_id', $user->id)
+        ->first();
```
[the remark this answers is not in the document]
wescopeland: Looks like we need to perform hard deletes here. Updated in latest.

app/Helpers/database/leaderboard.php
```diff
@@ -743,22 +775,27 @@ function duplicateLeaderboard(int $gameID, int $leaderboardID, int $duplicateNum
...
 {
-    $query = "DELETE FROM LeaderboardDef WHERE ID = $lbID";
+    $leaderboard = Leaderboard::find($lbID);
 
-    $dbResult = s_mysql_query($query);
+    if (!$leaderboard) {
+        return false;
+    }
+
+    $leaderboard->delete();
```
Jamiras, line 798: This might need to be a `forceDelete` too. After deleting a leaderboard, it still appears in the list, but cannot be deleted again.

wescopeland: Yep, given the nature of the previous code that was there, you're right. Good catch, thanks.

Jamiras (approved):

PR #2328, opened 2024-04-03
feat(manage): add role-based privileges to achievement editing

wescopeland:
This PR introduces role-based privileges to `/manage/achievements/{achievementId}/edit`.

The privileges live in `AchievementPolicy::updateField()`:

* ROOT has the ability to edit everything.
* DEVELOPER and DEVELOPER_STAFF can edit everything except `game_id`.
* DEVELOPER_JUNIOR can edit everything except `game_id` if the achievement is not promoted and they have an active claim on the game ID.
* WRITER can edit the `title` and `description` fields.

If a user has any multiple of these roles, their field editing permissions are aggregated.

**Viewing the achievement edit panel as a Writer:**
Screenshot 2024-04-03 at 6 46 04 PM


Jamiras:
How do I test this? I've given myself the "Developer" role, but Achievements doesn't appear in the navigation menu, and directly hitting "http://localhost:64000/manage/achievements" returns a 403.

wescopeland:
@Jamiras Apologies for the confusion. I initially tested everything hardcoded against the Writer role and forgot to enable access to the panel for the three developer roles before pushing.

I've opened `/manage` and `/manage/achievements` to those three roles and I've done an additional round of testing with all three of them. Here's how I tested:

1. Create a new user.
2. From your Root or Administrator account, visit `/manage/users/{User}/role`. Assign the user the role you'd like to test; I started with Writer.

`WRITER`
The writer can access `/manage` and `/manage/achievements`.
They can click the "Edit" button and modify values for the Title and Description fields.

`DEVELOPER_JUNIOR`
The junior developer can access `/manage` and `/manage/achievements`.
They can click the "Edit" button and can modify all fields with the exception of "Game" for achievements that are unpublished on an actively-claimed game.

`DEVELOPER` / `DEVELOPER_STAFF`
The developer can access `/manage` and `/manage/achievements`.
They can click the "Edit" button and can modify all fields with the exception of "Game" for any achievement.

`ROOT`
Your root user has the ability to update the value of the "Game" field.

---

One important nuance: System comments are not created when performing edits from the management panel. We'll need to decide on what our best approach to handle this should be. I see three possible options:
* Write system comments when updating achievement records from the management panel.
* Don't write system comments - rely entirely on the audit log (this is ideally the long-term goal anyway). Instead, update the comments UI on the achievements page to pull from the audit log and show both the log and system comments all-in-one.
* Don't write system comments - make the audit log public similar to the current set of system comments.

Jamiras:
Unrelated to this specific PR, but it's very difficult for me to add roles when the dropdown gets truncated:
image

Jamiras:
> I see three possible options:
> * Write system comments when updating achievement records from the management panel.
> * Don't write system comments - rely entirely on the audit log (this is ideally the long-term goal anyway). Instead, update the comments UI on the achievements page to pull from the audit log and show both the log and system comments all-in-one.
> * Don't write system comments - make the audit log public similar to the current set of system comments.

I think the second solution is probably the best. I doubt users are interested when the display order changes, but they are definitely interested in when the logic changes. The other fields I could go either way on. Some might care that the type changed, others might care that the points or description changed. I expect publishing/demoting and title changes are less important.

Jamiras (approved):

wescopeland:
@Jamiras
> ... it's very difficult for me to add roles when the dropdown gets truncated ...

I noticed this as well and started picking away at it. It looks like there's some bug going on with the dialog margin/padding, as well as a potential logic error with the autocomplete filtering itself. I'll make a note to look into both and resolve as a separate PR.

PR #2332, opened 2024-04-04
feat: don't show rich presence strings from muted users

wescopeland:
This PR makes an adjustment to rich presence strings for muted users. While their strings will continue to be stored in the DB, they are now displayed to end users as "Playing {Game Title}".

This impacts the active players component on the home page and the dev feed, as well as the players list on game pages. Additionally, the RP string is removed from the "last seen in" component on the user profile.


Jamiras:
Before I spend any time looking at this, is it really desirable? It would be impractical (and nearly impossible) to use rich presence to communicate with others. It would be a lot easier to just edit their motto.

wescopeland:
I assumed it was desirable based on a conversation I recently read from another admin regarding recent RP policy changes, but now I can't seem to find that conversation. No worries.

redwizard42:
> I assumed it was desirable based on a conversation I recently read from another admin regarding recent RP policy changes, but now I can't seem to find that conversation. No worries.

I think tele was thinking of it as a solution for allowing name entry into the RP again. I'd want at least the set dev to still be able to see it. (In some fashion)

PR #2333, opened 2024-04-04
Capture user agent for player sessions

Jamiras:
Simplified version is displayed on detailed user activity page. Full string is in tooltip.

image


database/migrations/platform/2024_04_03_000000_update_player_sessions.php
```diff
@@ -0,0 +1,25 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('player_sessions', function (Blueprint $table) {
+            $table->string('user_agent', 255)->nullable()->after('duration');
+            $table->string('ip_addr', 40)->nullable()->after('user_agent');
```
wescopeland, line 14: nit: I can't find anything else in our schemas that's truncated. Can we name this `ip_address` instead of `ip_addr`?

app/Platform/Services/UserAgentService.php
```diff
@@ -0,0 +1,183 @@
...
+
+    private function trimOperatingSystem(string $os): string
+    {
+        $index = strpos($os, ';');
+        if ($index !== false) {
+            $os = substr($os, 0, $index);
+        }
+
+        return trim($os);
+    }
```
wescopeland, line 182: Logically equivalent, but much more concise:
```php
return strtok(trim($os), ';');
```

app/Platform/Services/UserAgentService.php
```diff
@@ -0,0 +1,183 @@
...
+                    $c = substr($version, $index + 1, 1);
+                    if (is_numeric($c)) {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        return false;
+    }
```
wescopeland, line 172: I believe this entire function can be implemented in 1 line with a regular expression:

```php
return preg_match('/^\d+(?:\.\d+)+/', $version) === 1;
```

Your test suite is quite comprehensive, and they all still seem to pass.

app/Platform/Services/UserAgentService.php
```diff
@@ -0,0 +1,183 @@
...
+    {
+        if (array_key_exists($userAgent, $this->cache)) {
+            return $this->cache[$userAgent];
+        }
+
+        $data = $this->parseUserAgent($userAgent);
+        $this->cache[$userAgent] = $data;
+
+        return $data;
+    }
```
wescopeland, line 21: This one line is logically equivalent, and due to `??=` the cache value is inserted if it isn't present:
```php
return $this->cache[$userAgent] ??= $this->parseUserAgent($userAgent);
```

Jamiras: done

database/migrations/platform/2024_04_03_000000_update_player_sessions.php
Jamiras: done

app/Platform/Services/UserAgentService.php
Jamiras: done. but the token should be trimmed, not the input.

app/Platform/Services/UserAgentService.php
Jamiras: done, but I've kept the minimum length and `is_numeric` check on the first character to avoid compiling the regex when not necessary (most strings passed to this function won't start with a number).

wescopeland (approved):

luchaos (approved):
very nice - this will help a lot!

PR #2338, opened 2024-04-06
refactor(ForumTopic): drop Author and read from author_id

wescopeland:
This PR drops the `Author` column from the `ForumTopic` table and switches all reads to instead use `author_id`.

Additionally, many places that were fetching a forum topic actually only needed to do an existence check. This has been optimized.

Several related helper functions have been migrated to Eloquent ORM.


app/Helpers/database/user-permission.php
```diff
@@ -154,40 +154,38 @@ function getUserForumPostAuth(string $user): bool
...
+    $targetUser->ManuallyVerified = 1;
+    $targetUser->forum_verified_at = now();
+    $targetUser->save();
 
-    if (getAccountDetails($user, $userData)) {
-        addArticleComment('Server', ArticleType::UserModeration, $userData['ID'],
-            $sourceUser . ' authorized user\'s forum posts'
-        );
-    }
+    authorizeAllForumPostsForUser($targetUser->User);
```
Jamiras, line 184:
```
authorizeAllForumPostsForUser(): Argument #1 ($user) must be of type App\Models\User, string given, called in /var/www/html/app/Helpers/database/user-permission.php on line 184
```

resources/views/pages-legacy/forumposthistory.blade.php
```diff
@@ -58,12 +59,18 @@
...
+    $fetchedUserCache = [];
+
     foreach ($recentPosts as $topicPostData) {
         $postMessage = $topicPostData['ShortMsg'];
         if ($topicPostData['IsTruncated']) {
             $postMessage .= '...';
         }
-        $postAuthor = $topicPostData['Author'];
+        
+        $authorId = $topicPostData['author_id'];
```
Jamiras, line 70:
```
Undefined array key "author_id"
```

app/Helpers/database/forum.php
```diff
@@ -3,7 +3,9 @@
 use App\Community\Enums\ArticleType;
 use App\Community\Enums\SubscriptionSubjectType;
 use App\Enums\Permissions;
+use App\Models\ForumTopic;
 use App\Models\Game;
+use App\Models\User;
```
Jamiras, line 8:
```
Unknown column 'Author' in 'field list'
```
error on line 182 trying to create a new forum topic

app/Helpers/database/user-permission.php
wescopeland: Thanks, fixed in latest.

app/Helpers/database/forum.php
wescopeland: Resolved in latest.

resources/views/pages-legacy/forumposthistory.blade.php
wescopeland: Thanks, I wasn't seeing this during testing. It took a little bit of trial and error but I managed to repro. Resolved in latest.

Jamiras (approved):

PR #2339, opened 2024-04-07
fix prepopulated fields in new messages

Jamiras:
Such as "Contact Reporter" on ticket page and "Contact RAdmin" link on manage hashes


wescopeland (approved):
Verified locally.

PR #2341, opened 2024-04-07
fix(achievement): record "causer" to the activity log correctly

wescopeland:
This PR:
* Adds `MemAddr` and `user_id` to `Achievement::getActivitylogOptions()`.
* Fixes a bug where the activity logger does not correctly attach the causer on `UploadNewAchievement()` function calls.

As a result, achievement changes caused by Connect API function calls now correctly have users attached to their activitylog records.

Logic diffs are now also recorded in the activitylog.
 
Screenshot 2024-04-07 at 6 20 52 PM


Jamiras (approved):
It's probably intentional, but long MemAddrs don't wrap, so the feature is practically unusable.
image

(not that you'd even be able to eyeball a change in a multiline definition)

You can see the whole value in the view source, so it's not totally useless, and it's good to have the history stored somewhere.

I know luchaos talked about keeping versioned triggers somewhere. If we just had old versions in the database, we wouldn't need to duplicate the changes in the log - the log would just say "definition changed from 5 to 6" or something similar.

wescopeland:
@Jamiras You're right - the `triggers` table is where this should live, not the activity log. This appears to the the table where versioned achievement/leaderboard/etc logic can live.

I've rolled back the activity log changes, however I've kept the "causer" changes so updates from the Connect API are ascribed to the proper user.

Jamiras (approved):

PR #2345, opened 2024-04-11
fix(getParentGameFromGameTitle): check game title string correctly

wescopeland:
The current conditional never evaluates truthy.

### DIFF
--- a/app/Helpers/database/game.php
+++ b/app/Helpers/database/game.php
@@ -27,7 +27,7 @@ function getGameData(int $gameID): ?array
 // If the game is a subset, identify its parent game.
 function getParentGameFromGameTitle(string $title, int $consoleId): ?Game
 {
-    if (strstr('[Subset - ', $title) !== false) {
+    if (mb_strpos($title, '[Subset') !== false) {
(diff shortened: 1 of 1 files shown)


Jamiras (approved):

app/Helpers/database/game.php
```diff
@@ -27,7 +27,7 @@ function getGameData(int $gameID): ?array
 // If the game is a subset, identify its parent game.
 function getParentGameFromGameTitle(string $title, int $consoleId): ?Game
 {
-    if (strstr('[Subset - ', $title) !== false) {
+    if (mb_strpos($title, '[Subset') !== false) {
```
gboquizosanchez, line 30: https://php.watch/versions/8.0/str_contains

wescopeland: It was changed to `mb_strpos()` to be consistent with all other usages in the code. `getParentGameFromGameTitle()` itself will be going away soon as we build out the multiset feature.

gboquizosanchez: Perfect, I got the argument. But I'm a little bit allergic to strpos equalized to false, sorry haha

PR #2357, opened 2024-04-13
fix(ForumTopic): on delete, remove any associated game ForumTopicID values

wescopeland:
Normally, I'd reach for a foreign key with `onDelete("set null")`, but that might not be the correct approach given this comment on the `Game` model:

```php
// TODO drop ForumTopicID, migrate to forumable morph
```

At a minimum, we should be preserving integrity of the game records on topic delete.

### DIFF
--- a/public/request/forum-topic/delete.php
+++ b/public/request/forum-topic/delete.php
@@ -2,6 +2,7 @@
 
 use App\Enums\Permissions;
 use App\Models\ForumTopic;
+use App\Models\Game;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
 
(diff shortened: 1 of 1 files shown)


public/request/forum-topic/delete.php
```diff
@@ -13,8 +14,18 @@
...
-$topic->delete();
+$foundTopic = ForumTopic::find($forumTopicId);
+if ($foundTopic) {
+    $foundTopic->delete();
+
+    $foundAssociatedGame = Game::firstWhere('ForumTopicID', $forumTopicId);
+    if ($foundAssociatedGame) {
+        $foundAssociatedGame->ForumTopicID = null;
+        $foundAssociatedGame->save();
+    }
```
Jamiras, line 28: shouldn't this happen automatically if we create a foreign key that has `->onDelete('set null')`?

wescopeland: It should. I wasn't sure if we wanted to add a foreign key here, per the PR description:

> Normally, I'd reach for a foreign key with onDelete("set null"), but that might not be the correct approach given this comment on the Game model:

However, it seems like if we're going to store this data, that's probably the best approach. There's no timeline for when we'll have a forumable graph.

PR #2360, opened 2024-04-14
fix(game-compare): remediate link button size issue on LG and lower

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1229087996075315351.

Additionally, modifies some of the link button styles so the icon and text are now better aligned along the center of the button.

**Before**
Screenshot 2024-04-14 at 1 30 53 PM

**After**
Screenshot 2024-04-14 at 1 30 43 PM

### DIFF
--- a/resources/views/pages/user/[user]/game/[game]/compare.blade.php
+++ b/resources/views/pages/user/[user]/game/[game]/compare.blade.php
@@ -74,7 +74,7 @@ $canModerate = ($user->Permissions >= Permissions::Moderator);
 
 @if ($canModerate)
     <x-hidden-controls>
-        <div style="width:20%">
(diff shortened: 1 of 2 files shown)


resources/views/pages/user/[user]/game/[game]/compare.blade.php
```diff
@@ -74,7 +74,7 @@
 
 @if ($canModerate)
     <x-hidden-controls>
-        <div style="width:20%">
+        <div class="grid md:grid-cols-3 lg:grid-cols-4 xl:grid-cols-5">
```
Comment (wescopeland) [inline resources/views/pages/user/[user]/game/[game]/compare.blade.php:77]: This div was the root cause of the issue.

Rather than a hardcoded width, it's better to reach for a grid layout with specified column counts at certain breakpoints (generally XS and SM just have 1 column, which is the default).

Jamiras (approved):

PR #2361, opened 2024-04-14
fix(forum): remediate /game subpage shortcode normalization issues

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1229130726901809192.

The shortcode normalizer is bit too aggressive on game page URLs now that we have routes like "/game/{id}/{subpage}".

**Before**
Screenshot 2024-04-14 at 3 19 32 PM

**After**
Screenshot 2024-04-14 at 3 19 38 PM

Reproducible with string `https://retroachievements.org/game/5236/hashes`.


app/Helpers/shortcode.php
```diff
@@ -113,6 +113,13 @@ function normalize_user_shortcodes(string $value): string
 
 function normalize_game_shortcodes(string $value): string
 {
+    // If the original string contains a subpath beyond the game ID,
+    // eg: "https://retroachievements.org/game/5236/hashes", return the
+    // original string rather than attempting to normalize it.
+    if (preg_match("~https?://[\w\-\.]+/game/\d+/.+~si", $value)) {
+        return $value;
+    }
```
Jamiras, line 121: Why isn't this also a problem with users?
image

And should we protect against similar behavior for achievements and tickets in case they get trailing paths at some point in the future?

Jamiras: Suggest adding a unit test for this logic.

wescopeland: Users already had a lookahead in the regex, which led me to implement a much more elegant solution in latest.

I've trimmed down three of those normalization functions into a single reusable function. I've left some TODOs around the ticket normalization function, which may be useful for the work happening in https://github.com/RetroAchievements/RAWeb/pull/2375.

Additionally, I've added several tests to cover the behavior.

tests/Unit/ShortcodeTest.php
```diff
@@ -166,6 +166,90 @@ public function testStripAndClampCombinedWordsAndActualShortcodes(): void
...
+
+        $this->assertEquals(
+            '[ach=9]',
+            $normalized
+        );
+    }
+
+    public function testNormalizeAchievementShortcodesWithSubpaths(): void
+    {
+        $rawString = 'https://retroachievements.org/user/achievement/9/subpath';
```
Jamiras, line 231: `/user/achievement/`?

tests/Unit/ShortcodeTest.php
```diff
@@ -166,6 +166,90 @@ public function testStripAndClampCombinedWordsAndActualShortcodes(): void
...
+
+        $this->assertEquals(
+            '[game=1]',
+            $normalized
+        );
+    }
+
+    public function testNormalizeGameShortcodesWithSubpaths(): void
+    {
+        $rawString = 'https://retroachievements.org/user/game/1/hashes';
```
Jamiras, line 207: `/user/game/`?

app/Helpers/shortcode.php
```diff
@@ -59,86 +58,43 @@ function url2shortcode(string $url): string
     return $link;
 }
 
-function normalize_shortcodes(string $value): string
+function normalize_shortcodes(string $input): string
```
Jamiras, line 61: Minor: I noticed that the preview path doesn't go through this code.

image

tests/Unit/ShortcodeTest.php
wescopeland: Typo, fixed in latest

tests/Unit/ShortcodeTest.php
wescopeland: Typo, fixed in latest

app/Helpers/shortcode.php
wescopeland: _preview.php_ in latest has been adjusted so it also executes this function before calling the shortcode helper.

PR #2369, opened 2024-04-18
Use first achievement as a reference for the average points per week

Zekromaster:
As discussed on the Discord, I altered the calculation of the average points per week so that instead of using the join date as a starting date, they use the date of the first achievement on the profile.


app/Helpers/database/user.php
wescopeland, line 1: We're in the process of moving away from these helpers. I think this should instead go in `UserProfileMeta::calculateAveragePointsPerWeek()` itself.

In `UserProfileMeta`, have access to `User $user`, which should allow us to better leverage the `HasMany` relationship to `->playerAchievements()`.

wescopeland (commented):
Thanks for submitting your first PR to RAWeb!

Confirmed this query:

```sql
SELECT 
    id,
    achievement_id,
    unlocked_at
FROM 
    player_achievements
WHERE 
    user_id = 1
    AND unlocked_at IS NOT NULL
ORDER BY 
    unlocked_at ASC
LIMIT 1;
```

Takes ~1ms on prod so not too concerned about back-end performance.

I am currently on break until May 1, so my responses/re-review may be much more delayed than usual.

app/Community/Components/UserProfileMeta.php
```diff
@@ -425,12 +435,17 @@ private function calculateAverageFinishedGames(array $userJoinedGamesAndAwards):
...
+        if (is_null($startingDateStr)) {
+            return 0;
+        }
+
+        $startingDate = Carbon::createFromFormat('Y-m-d H:i:s', $startingDateStr);
+
         $now = Carbon::now();
 
-        $weeksOfMembership = $memberSince->diffInWeeks($now);
+        $weeksOfMembership = $startingDate->diffInWeeks($now);
```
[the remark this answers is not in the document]
Zekromaster: Agreed. Will do.

app/Helpers/database/user.php
Zekromaster: I've changed the code to make use of the playerAchievements() relationship instead of relying on a helper.

Good to know what the recommended standardised way to access the DB is, it got confusing with all the different ways used throughout the codebase - guess it comes with the territory when a project is old enough.

app/Community/Components/UserProfileMeta.php
```diff
@@ -425,12 +435,17 @@ private function calculateAverageFinishedGames(array $userJoinedGamesAndAwards):
         return number_format($averageFinishedGames, 2, '.', '');
     }
 
-    private function calculateAveragePointsPerWeek(string $userMemberSince, int $points = 0): int
+    private function calculateAveragePointsPerWeek(?string $startingDateStr, int $points = 0): int
```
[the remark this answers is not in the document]
Zekromaster: The function's signature was completely changed by my more recent commit. The "str" at the end was probably an holdover of some kind.

wescopeland:
@Zekromaster I like to mark my comments as resolved, otherwise I lose track of what I need to review when reapproaching a PR. Thanks for your understanding on this.

Zekromaster:
> @Zekromaster I like to be the one who marks my comments as resolved, otherwise I lose track of what I need to review when reapproaching a PR. Thanks for your understanding on this. For the time being, I've "unresolved" the given comments.

No problem, sorry for the confusion!

app/Community/Components/UserProfileMeta.php
```diff
@@ -425,19 +426,30 @@ private function calculateAverageFinishedGames(array $userJoinedGamesAndAwards):
         return number_format($averageFinishedGames, 2, '.', '');
     }
 
-    private function calculateAveragePointsPerWeek(string $userMemberSince, int $points = 0): int
+    private function calculateAveragePointsPerWeek(User $user, bool $hardcore, int $points = 0): int
```
wescopeland, line 429:
```suggestion
    private function calculateAveragePointsPerWeek(User $user, bool $doesUserPreferHardcore, int $points = 0): int
```

nit: For a reader unfamiliar with the code, it may be unclear why the boolean arg would ever be set to true. I think renaming the arg could improve maintainability of this function.

app/Community/Components/UserProfileMeta.php
```diff
@@ -425,19 +426,30 @@ private function calculateAverageFinishedGames(array $userJoinedGamesAndAwards):
...
     {
-        $memberSince = Carbon::createFromFormat('Y-m-d H:i:s', $userMemberSince);
+        $field = $hardcore ? "unlocked_hardcore_at" : "unlocked_at";
+
+        $startingDate = $user
+            ->playerAchievements()
+            ->whereNotNull($field)
+            ->orderBy($field)
+            ->first()[$field] ?? null;
```
wescopeland, line 437: This could be further optimized. `[$field]` kind of clues in that Eloquent ORM is doing `SELECT *` under the hood. What we really want is `SELECT $field`.

```php
$startingDate = $user->playerAchievements()
    ->whereNotNull($field)
    ->orderBy($field, 'asc')
    ->value($field);
```

A minor nit. This is a strange convention coming from another language. Typically in this codebase, FP chains are formatted like:

```php
$result = $subject->initialFunction()
    ->functionB()
    ->functionC()
    ->functionD();
```

It took me a while to get used to this. Unfortunately, our automated tooling doesn't fix this formatting.

app/Community/Components/UserProfileMeta.php
```diff
@@ -425,19 +426,30 @@ private function calculateAverageFinishedGames(array $userJoinedGamesAndAwards):
...
+
+        if (is_null($startingDate)) {
+            return 0;
+        }
+
         $now = Carbon::now();
 
-        $weeksOfMembership = $memberSince->diffInWeeks($now);
+        $weeksSinceFirstUnlock = $startingDate->diffInWeeks($now);
```
wescopeland, line 445: This is unrelated to your change, but we might as well clean this up while we're in here.

I think I may have played with this algorithm a bit when designing the component, and at one point during the design phase there were multiple references to `$now`. However, I believe `$now` is currently an unnecessary abstraction, given it only aliases to `Carbon::now()` and it's only used in one place.

```php
$weeksSinceFirstUnlock = $startingDate->diffInWeeks(Carbon::now());
```

resources/views/pages-legacy/userInfo.blade.php
```diff
@@ -113,6 +113,7 @@
         username: $userPage
     );
 }
+
```
wescopeland, line 116: nit: Unless we're making a change to this page's template, let's revert this.

app/Community/Components/UserProfileMeta.php
Zekromaster: Oh, I should really have read the Eloquent documentation more carefully. Oh, well, I learnt something today.

Zekromaster:
I'm assuming all commits will be squashed before merge, so I'm just pushing another "apply code review" commit. I manually reverted what had to be reverted and tested squashing locally, now UserProfileMeta.php is the only file that would get touched by squash-merging.

app/Community/Components/UserProfileMeta.php
```diff
@@ -425,19 +425,27 @@ private function calculateAverageFinishedGames(array $userJoinedGamesAndAwards):
         return number_format($averageFinishedGames, 2, '.', '');
     }
 
-    private function calculateAveragePointsPerWeek(string $userMemberSince, int $points = 0): int
+    private function calculateAveragePointsPerWeek(User $user, bool $doesUserPreferHardcore, int $points = 0): int
```
Jamiras, line 428: As best as I can tell, this is only called from the one place above. There's no reason to have `$points` be optional. If there is another case where points wasn't being provided, adding an extra parameter here would have broken it.

Jamiras (commented):
From a cursory glance, these seem like reasonable changes. I'll leave approval to Wes as he's already been heavily involved.

app/Community/Components/UserProfileMeta.php
Zekromaster: I think I simply left points optional because it was also optional before my changes even though it was always provided then too.

wescopeland (approved):

PR #2373, opened 2024-04-19
feat(notifications): split ticket notifications into a separate nav item

drisc:
Splits ticket notifications out from the main notification nav item for users that have the manage permissions for tickets.

Before:
CleanShot 2024-04-19 at 5  14 55@2x

After:
CleanShot 2024-04-19 at 5  16 36@2x


app/Providers/AppServiceProvider.php
```diff
@@ -89,6 +89,7 @@ public function boot(): void
          * Register Livewire components
          */
         Livewire::component('notification-icon', NotificationIcon::class);
+        Livewire::component('notification-ticket', TicketIcon::class);
```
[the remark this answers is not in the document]
drisc: Next commit will have components and blade templates renamed in this better style. The terms you suggested fit well.

app/Components/NotificationIcon.php
```diff
@@ -29,31 +29,16 @@ public function render(): View
...
-            if ($openTicketsData[TicketState::Open]) {
-                $notifications->push([
-                    'link' => url('/ticketmanager.php?u=' . $user->User . '&t=' . (TicketFilters::Default & ~TicketFilters::StateRequest)),
-                    'title' => $openTicketsData[TicketState::Open] . ' ' . __res('ticket', (int) $openTicketsData[TicketState::Open]) . ' for you to resolve',
-                    'class' => 'text-danger',
-                ]);
-            }
-            if ($openTicketsData[TicketState::Request]) {
+        // Ticket feedback for users without manage permissions
+        if ($user->cannot('manage', Ticket::class)) {
```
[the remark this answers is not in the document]
drisc: Makes sense, next commit will have this changes.

resources/views/components/notification/ticket.blade.php
```diff
@@ -0,0 +1,23 @@
...
+        @endif
+    </x-slot>
+    @if($notifications->isNotEmpty())
+        @foreach($notifications as $notification)
+            <x-dropdown-item :link="$notification['link']" :class="$notification['class'] ?? ''">
+                {{ $notification['title'] }}
+            </x-dropdown-item>
+        @endforeach
+    @else
+        <div class="flex whitespace-nowrap text-muted px-3 py-2">{{ __('No new notifications') }}</div>
```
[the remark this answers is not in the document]
drisc: Changed in latest.

app/Components/TicketIcon.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function render(): View
+    {
+        /** @var User $user */
+        $user = request()->user();
+
+        $notifications = collect();
+
+        // Open ticket notifications
+        if ($user->can('manage', Ticket::class)) {
```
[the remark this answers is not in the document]
drisc: Changed in latest.

resources/css/utilities.css
```diff
@@ -52,3 +52,15 @@ span.smalldate {
...
+  color: var(--dot-color-low);
+}
+
+.dot-warning {
+  color: var(--dot-color-warning);
+}
+
+.dot-urgent {
+  color: var(--dot-color-urgent);
+}
```
wescopeland, line 66: Using global CSS for this is probably overkill. Tailwind utilities alone are enough to get by.

However, I don't believe this is desirable. Ascribing severity to the dot based solely on count of the items may miss the mark. There may be only 1 item, but it's a critical issue. We also really shouldn't convey such information using color- it's an accessibility violation.

For the time being, I suggest we keep things simple and have the dot represent the presence of notifications.

app/Policies/AchievementSetClaimPolicy.php
wescopeland, line 1: I think this change as-is is fine. However, we'll likely need to revisit this policy in the future.

An entity's `manage()` policy conventionally has some subtle nuances that lean more into the administrative realm, eg: the ability to manage _any_ claim. Junior Developers also need to have more tight access controls around their role (see: https://github.com/RetroAchievements/RAWeb/pull/2328).

I would anticipate anyone with `AchievementSetClaim::manage()` level access to essentially have "god mode" over all claims in the system, up to and including the ability to force-drop any of them. This leads me to believe the current implementation for this in `master` is not correct (although there is no harm in this, the policy is currently unused).

resources/css/utilities.css
drisc: The colour is based on the `max()` value of the new priority key that is found in the `$notifications` collection, not the count of notifications found. So if you had 12 messages it would be blue, add one Ticket awaiting feedback and the dot would be Orange and if you then got a message with priority 2 (say a message from an admin account) then the dot would go red.

That said your right about it being an accessability issue so I'll go ahead and roll that back.

app/Policies/AchievementSetClaimPolicy.php
drisc: The only place that currently uses this new manage policy is the notification area and it's only checking for if the user is able to see claim expiry. It's not on any management facing areas at the moment but maybe moving to the view function in the policy rather than manage is the way forward for this use case.

wescopeland: Got it, I see the code now.

Given the context, I'm inclined to believe the correct policy for this logic is `AchievementSetClaim::create()`. This policy should control whether a user can claim a set for a game, which should also imply if they'll receive notifications about their created claims that are about to expire.

I believe what may be best for now is to duplicate your `manage()` code into `create()` for that policy, then point the conditional to "create".

resources/views/components/notifications/ticket.blade.php
```diff
@@ -0,0 +1,32 @@
+@props([
+    'trigger' => '',
+    'count' => null, //int
```
wescopeland, line 3:
```suggestion
    'count' => 0,
```
0 seems like a sane default, which also allows us to get rid of the type comment.

Jamiras (commented):
Seems functional. I'll let Wes provide final approval as he's been providing the feedback.

wescopeland (approved):
Thanks for getting this knocked out and working through the feedback.

PR #2374, opened 2024-04-21
feat(opengraph): change mature hub game preview icon to actual game icon, add hub icon

drisc:
This PR changes the mature game icon shown via the gate to the sets actual icon. Several people have pointed out that we don't allow NSFW imagery in set icons anyway so it makes little sense to obscure it. When a user clicks through the link they will be presented with the mature gate if they are not logged in or do not have the mature webpref set.

This PR also adds Hub images to link previews as currently they just show the default RA logo.


resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -278,6 +278,13 @@
 );
 ?>
 
+<?php if (!$isFullyFeaturedGame): ?>
+    <?php
+        $pageType = 'retroachievements:hub';
+        $pageImage = media_asset($gameData['ImageIcon']);;
+    ?>
+<?php endif ?>
```
wescopeland, line 286: Is there a reason why new code would use `<?php if (): ?>` instead of `@if ()`?

drisc: Not really, I just made it the same style as the other if statement below it, I've updated both of them.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -278,7 +278,14 @@
 );
 ?>
 
-<?php if ($isFullyFeaturedGame): ?>
+@if (!$isFullyFeaturedGame)
```
wescopeland, line 281: There's enough duplication here that I think it may be worth making it more concise. We can probably get away with just lifting it up into the PHP block directly above, as with this reduction we no longer need directives of any kind.

```php
sanitize_outputs(...);

$pageType = $isFullyFeaturedGame ? 'retroachievements:game' : 'retroachievements:hub';
$pageImage = media_asset($gameData['ImageIcon']);

if ($isFullyFeaturedGame) {
    $pageDescription = generateGameMetaDescription(
        $gameTitle,
        $consoleName,
        $numAchievements,
        $totalPossible,
        $isEventGame
    );
}
?>
```

drisc: The psudocode that you provided is sound, added to latest.

Jamiras (approved):

PR #2375, opened 2024-04-21
Convert ticket editor to blade

Jamiras:
Uses a newer layout similar to the recent profile changes.

Detailed list of changes:
* Link for "other tickets for game" moved into breadcrumb.
* Page title changed from "Ticket Manager - N Open Tickets" to "Achievement Name (Report Type)".
* Ticket Row data, and many display fields moved into two summary chunks - one for the ticket, and one for the related achievement.
* Notes moved down to be part of comments block.
* Action moved below comments block - you should add a comment before performing an action.
* "Demote achievement to unofficial" moved into action menu.
* "Times earned since reported" is not shown for resolved tickets
* When available, client information from User-Agent is listed in Player history.
* Comment section now fills full width of editor.

Before (open ticket):
image

After (open ticket):
image

Before (closed ticket):
image

After (closed ticket):
image


suspect15:
I wonder if color coding the ticket status would be helpful.  Perhaps something like red for Open, maybe yellow for Requested, green for Closed/Resolved?

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -0,0 +1,423 @@
...
+                @if (empty($userAgentLinks))
+                    <span class="text-muted">No client data available</span>
+                @else
+                    <span>
+                        <span class="font-bold">Clients used:</span>
+                        <span>{!! implode(', ', $userAgentLinks) !!}</span>
+                    </span>
+                @endif
+
+                    <button id="unlockHistoryButton" class="absolute bottom-0 right-0 btn"
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:222]: Changing "bottom-0 right-0" to "-top-2 -right-2" will let the button sit flush against the container edge.

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -0,0 +1,423 @@
...
+
+    <div class="mt-2">
+        <div class="flex w-full justify-between border-embed-highlight items-center">
+            <p role="heading" aria-level="2" class="mb-0.5 text-2xs font-bold">
+                Comments
+            </p>
+            @if ($permissions >= Permissions::Developer)
+                <div class="flex flex-col gap-y-1">
+                    <a class="btn py-2 block transition-transform lg:active:scale-[97%]" href="{!! $contactReporterUrl !!}">
+                        Contact the reporter
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:313]: Recommend changing this to "Message the reporter" so it's consistent with the canonical term we use for the feature: "Messages".

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -0,0 +1,423 @@
...
+            </div>
+        </div>
+    @endif
+
+    <div class="mt-2">
+        <div class="flex w-full justify-between border-embed-highlight items-center">
+            <p role="heading" aria-level="2" class="mb-0.5 text-2xs font-bold">
+                Comments
+            </p>
+            @if ($permissions >= Permissions::Developer)
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:310]: We should definitely use a policy here with `@can()`. Maybe:
```php
@can('manage', App\Models\Ticket::model)
```

The only thing that gives me pause is in `TicketPolicy::manage()`, I see:

```php
{
    return $user->hasAnyRole([
        Role::HUB_MANAGER,
        Role::DEVELOPER_STAFF,
        Role::DEVELOPER,
        Role::DEVELOPER_JUNIOR,
    ])
        || $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper;
}
```

The `@if` currently restricts against Jr Devs. Maybe it shouldn't? Or maybe we should have a new function in our ticket policy that excludes Jr Devs from the ability to view this data?

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -0,0 +1,423 @@
...
+                @endif
+            </tbody></table>
+        </div>
+    </div>
+
+    <div>
+        <form method="post" action="/request/ticket/update.php">
+            {!! csrf_field() !!}
+            <input type="hidden" name="ticket" value="{{ $ticket->ID }}">
+            @if ($permissions >= Permissions::Developer)
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:358]: This seems like it should be `@can('manage, App\Models\User::class)`. We probably need to update `UserPolicy::requireAdministrativePrivileges()` to use a `Permissions` fallback similar to our other policies, eg:

```php
// !! untested pseudocode !!

private function requireAdministrativePrivileges(User $user, ?User $model = null): bool
{
    if (!$model) {
        return $user->hasAnyRole([
            Role::ROOT,
            Role::ADMINISTRATOR,
            Role::MODERATOR,
        ])
            || $user->getAttribute('Permissions') >= Permissions::Moderator;
    }

    // ...
}
```

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -0,0 +1,423 @@
...
+                @endif
+            @elseif ($user?->id === $ticket->reporter->id)
+                @if (TicketState::isOpen($ticket->ReportState))
+                    <input type="hidden" name="action" value="{{ TicketAction::ClosedMistaken }}">
+                    <button class='btn'>Close as mistaken report</button>
+                @endif
+            @endif
+        </form>
+
+        @if ($permissions >= Permissions::JuniorDeveloper)
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:399]: Ideally this would be driven by something in our achievement policy, but #2328 is likely a hold-up here.

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -0,0 +1,423 @@
...
+                        onclick="toggleExpander('achievementLogicButton', 'achievementLogicContent')">Achievement Logic ▼</button>
+            </div>
+
+            <div id="achievementLogicContent" class="hidden devboxcontainer">
+                <div style='clear:both;'></div>
+                <li>Achievement ID: {{ $ticket->achievement->id }}</li>
+                <div>
+                    <li>Mem:</li>
+                    <code>{{ $ticket->achievement->MemAddr }}</code>
+                    <li>Mem explained:</li>
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:411]: I know this pattern is used elsewhere, but this HTML is technically invalid.

An `<li>` cannot exist outside of a `<ul>` or `<ol>`. It's a list item that doesn't belong to a list.

I assume in the legacy codebase, an `<li>` was chosen purely for the sake of adding a dot element next to the text. I restructured the HTML a bit here to make it valid:

```php
<div id="achievementLogicContent" class="hidden devboxcontainer">
    <ul class="list-disc ml-4 mb-2">
        <li>Achievement ID: {{ $ticket->achievement->id }}</li>
        <li>
            Mem:
            <code>{{ $ticket->achievement->MemAddr }}</code>
        </li>
    </ul>

    <p>Mem explained:</p>
    <div>
        @php
            $triggerDecoderService = new TriggerDecoderService();
            $groups = $triggerDecoderService->decode($ticket->achievement->MemAddr);
            $triggerDecoderService->addCodeNotes($groups, $ticket->achievement->GameID);
        @endphp
        <x-trigger.viewer :groups="$groups" />
    </div>
</div>
```

Note: this does cause longer code samples to immediately break to the next line.

app/Helpers/render/comment.php
```diff
@@ -61,38 +62,34 @@ function RenderCommentsComponent(
 
     echo "<table id='feed' class='table-highlight'><tbody>";
 
-    $lastID = 0;
-    $lastKnownDate = 'Init';
+    // moderators can delete any comment
+    // users can delete any comment off of their wall
+    $canDeleteAnyComment =
+        $permissions >= Permissions::Moderator
+        || ($articleTypeID === ArticleType::User && $user?->id === $articleID);
```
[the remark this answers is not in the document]
Jamiras: Ideally, yes, but `$commentData` is just an array built from direct SQL, so we don't actually have `Comment` models at this point. I've added the logic to `CommentPolicy::delete`, but actually making that change is a much larger refactoring project and shouldn't be part of this PR.

app/Helpers/render/comment.php
```diff
@@ -61,38 +62,34 @@ function RenderCommentsComponent(
...
-            $comment['CommentPayload'],
-            // TODO no unix timestamp here
-            (int) $comment['Submitted'],
-            $username,
-            $articleTypeID,
-            (int) $comment['ID'],
-            $canDeleteComment
-        );
+        // users can delete their own comments
+        $canDeleteComment = $canDeleteAnyComment || $comment['User'] === $username;
```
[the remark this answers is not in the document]
Jamiras: See previous comment.

resources/views/components/comment/item.blade.php
```diff
@@ -0,0 +1,77 @@
...
+            <div>
+                <span class="smalldate">{{ $when?->format('j M Y H:i') }}</span>
+            </div>
+
+            <div style="word-break: break-word">
+                {!! $payload !!}
+            </div>
+        </td>
+    </tr>
+@elseif ($author && $author->banned_at && !request()->user()?->can('manage', $author))
```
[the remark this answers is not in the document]
Jamiras: `CommentPolicy::manage` would make sense, but see my previous comment about having access to a `Comment` at this point.

This represents [the following code](https://github.com/RetroAchievements/RAWeb/pull/2375/files#diff-4378c3dc8567f97e42908e6d86dde70e80460cad1a06afba69ae39a671ecb97fL68-L69):
```
        if (isset($comment['banned_at']) && $permissions < Permissions::Moderator) {
            continue;
```

resources/views/components/comment/item.blade.php
```diff
@@ -0,0 +1,77 @@
...
+    {{-- banned user comments are only visible to moderators --}}
+@else
+    @if ($commentId > 0)
+    <tr class="comment" id="comment_{{ $commentId }}_highlight">
+    @else
+    <tr class="comment">
+    @endif
```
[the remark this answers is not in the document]
Jamiras: done.

resources/views/pages/ticket/[ticket].blade.php
Comment (Jamiras) [reply resources/views/pages/ticket/[ticket].blade.php:223]: I like it flush with the expanded stuff more than the outer container.

Current:
image

image

Your proposal:
image

image

resources/views/pages/ticket/[ticket].blade.php
Comment (Jamiras) [reply resources/views/pages/ticket/[ticket].blade.php]: This just hides the "Contact the Reporter" button from non-developers. I don't see any reason why it shouldn't be available for Junior Developers.

resources/views/pages/ticket/[ticket].blade.php
Comment (Jamiras) [reply resources/views/pages/ticket/[ticket].blade.php]: done.

resources/views/pages/ticket/[ticket].blade.php
Comment (Jamiras) [reply resources/views/pages/ticket/[ticket].blade.php:359]: This is just preventing a user from changing the ticket state. Junior Developers have to have their solutions reviewed, so this should be restricted to Developers+. I've added an `updateState` method to `TicketPolicy`.

resources/views/pages/ticket/[ticket].blade.php
Comment (Jamiras) [reply resources/views/pages/ticket/[ticket].blade.php]: done

resources/views/pages/ticket/[ticket].blade.php
Comment (wescopeland) [reply resources/views/pages/ticket/[ticket].blade.php:223]: That reasoning makes sense, though it seems inconsistent with how we use this UI pattern. I'd argue the component design shouldn't change because of some other element somewhere on the page, for these reasons:

1. It makes it more difficult to wrap this up into something reusable, as there's a one-off modification on a page somewhere.
2. It's plausible the "Award Hardcore" button won't be there one day, in which case the shifted position of the "Unlock History" button won't make sense.

I don't feel strongly enough about it to express more concern beyond this, as I can already think of counterarguments to these two points.

resources/views/pages/ticket/[ticket].blade.php
Comment (wescopeland) [reply resources/views/pages/ticket/[ticket].blade.php]: Yes, my focus was more focused on the `manage()` policy itself, which implies broader reach than what I think we might want to allow for Junior Developers.

Comment (wescopeland) [reply resources/views/pages/ticket/[ticket].blade.php]: After rethinking this, it's probably not an issue.

wescopeland (approved):

PR #2377, opened 2024-04-22
refactor(forum): use role-based access policies

wescopeland:
This PR migrates most (all?) forum functionality to use access policies and modern roles, with a legacy `Permissions` fallback in place to preserve existing functionality.

Functionally, in this branch, a user account should be able to interact with the forum identically to how it would in prod, unless otherwise stated.

Summary of changes:
* `ForumTopicCommentPolicy` and `ForumCommentPolicy` are the meat of the PR - this is where all the rules live.
* `User::hasVerifiedEmail()` no longer relies on `Permissions`. It instead checks if `email_verified_at` is a truthy datetime value which occurred in the past.
* ~~Users can now delete their own forum topics.~~
* Default permissions checks in `authenticateFromCookie()` associated with the forum have largely been removed. Policies are checked against instead.
* Props have been updated on the forum's post component to accept a native `ForumTopicComment` model.

**How I tested my changes:**
* First, I browsed the forum and various topics to ensure nothing exploded while unauthenticated.
* Next, I set up 4 users: a new unverified user, a verified user, a muted user, and a moderator.
* To the best of my ability, with each user, I attempted to:
  * Post a reply to an existing topic.
  * Post a new topic.
  * Edit/delete my topic.
  * Edit any of my posts.
  * Edit another user's posts.
  * Delete another user's posts.
* I also tested verifying a user's posts to ensure that wouldn't explode.


resources/views/pages-legacy/viewtopic.blade.php
```diff
@@ -149,39 +155,35 @@
...
-        $isSubscribed
-    );
+    if ($user) {
+        RenderUpdateSubscriptionForm(
+            "updatetopicsubscription",
+            SubscriptionSubjectType::ForumTopic,
+            $thisTopicID,
+            $isSubscribed
+        );
+    }
```
wescopeland, line 165: This conditional fixes a subtle bug. Currently, for unauthenticated users, they still see the "Subscribe" button when viewing topics. Clicking it always returns an error. For now, we should probably just hide the button if the user isn't logged in.

app/Models/User.php
```diff
@@ -439,7 +439,7 @@ public function getPointsWeightedAttribute(): int
 
     public function hasVerifiedEmail(): bool
     {
-        return (int) $this->getAttribute('Permissions') >= Permissions::Registered;
+        return $this->email_verified_at?->isPast() ?? false;
```
Jamiras, line 442: FYI, There are ~40 users in the database with `Permissions == Registered` and a null `email_verified_at`.

The most recently used ones appear to be the result of being manually set to Registered by a moderator. Since they're going to be seeing the "please verify your email" banner, it seems like it's a non-issue.

app/Policies/ForumTopicPolicy.php
```diff
@@ -36,23 +39,32 @@ public function view(?User $user, ForumTopic $topic): bool
...
     public function delete(User $user, ForumTopic $topic): bool
     {
-        if ($user->hasAnyRole([
-            Role::MODERATOR,
-        ])) {
-            return true;
-        }
-
-        return $user->is($topic->user);
+        return $this->manage($user) || $user->is($topic->user);
```
Jamiras, line 67: This seems to be the intent from the left side, but giving the OP permission to delete any topic they've created seems like a dangerous option - particularly when it comes to Official Forum Topics. This might be something to discuss with the moderation team.

wescopeland: Agreed, let's retain the current behavior. I've adjusted this so only moderators can delete topics.

Jamiras (approved):

PR #2380, opened 2024-04-23
refactor(UserRelationship): migrate to eloquent ORM, read from user ids

wescopeland:
This PR switches all user relationship reads from `User` and `Friend` to `user_id` and `related_user_id`. To mitigate risk/downtime, a migration to drop those columns will come later.

Most of the user relationship helper functions have been migrated to Eloquent ORM. Areas I tested:
* User profiles
* Global points ranking
* The friends page
* Private messages


app/Community/Concerns/ActsAsCommunityMember.php
```diff
@@ -79,14 +81,21 @@ public function followerUsers(): BelongsToMany
         return $this->inverseRelatedUsers()->where('Friendship', '=', UserRelationship::Following);
     }
 
-    public function isFollowing(string $username): bool
+    public function getRelationship(User $user): int
```
wescopeland, line 84: This has been migrated from the static `UserRelationship::getRelationship()`. It seems more appropriate to have here on the `ActsAsCommunityMember` trait.

app/Helpers/database/user-relationship.php
```diff
@@ -71,69 +71,73 @@ function changeFriendStatus(User $senderUser, User $targetUser, int $newStatus):
     }
 }
 
-function isUserBlocking(string $user, ?string $possibly_blocked_user): bool
+// jamiras if you see this comment it means i forgot to delete it before submitting
+// a PR. this function adds 30ms of execution time over the original function for the
+// worst case scenario user.
+function GetFriendList(User $user): array
```
wescopeland, line 74: Note: this executes roughly 30ms slower than the raw SQL version.

app/Helpers/database/user-relationship.php
```diff
@@ -71,69 +71,70 @@ function changeFriendStatus(User $senderUser, User $targetUser, int $newStatus):
     }
 }
 
-function isUserBlocking(string $user, ?string $possibly_blocked_user): bool
+function GetFriendList(User $user): array
```
wescopeland, line 74: Note: this executes roughly 30ms slower than the original SQL version.

Jamiras: With only four friends, my query time changed from 260us to 750us.

Using someone following 1800 people (which seems to be the max in prod right now, and makes the following page unresponsive), my query time actually went down (from 90ms to 21ms).

Additionally, as best as I can tell, this is only called from `dorequest.php?r=getfriendslist`, which I don't believe is exposed through rcheevos, so it should only be getting called if the user navigates to the friends page of the DLL overlay.

Unless I'm missing something, I don't think this is something to worry about.

Jamiras: You might be running in to whatever caused [this comment](https://github.com/RetroAchievements/RAWeb/pull/2380/files#diff-08ba1639618d6064911bf201eb28ef06ea28bf0ed974101a116d4d716da9b5dfR132-R135) that would affect other relations functions: 
```
    // TODO: why is it so much faster to run this query and build the IN list
    //       than to use it as a subquery? i.e. "AND aw.User IN ($friendsSubquery)"
    //       local testing took over 2 seconds with the subquery and < 0.01 seconds
    //       total for two separate queries
```

Not sure why I'm not seeing that behavior.

app/Helpers/database/user-relationship.php
```diff
@@ -71,69 +71,70 @@ function changeFriendStatus(User $senderUser, User $targetUser, int $newStatus):
...
+    $friendList = $user->followedUsers()
+        ->where('Permissions', '>=', Permissions::Unregistered)
+        ->whereNull('Deleted')
+        ->orderBy('LastActivityID', 'DESC')
+        ->get()
+        ->map(function ($friend) {
+            return [
+                'Friend' => $friend->User,
+                'RAPoints' => $friend->points,
+                'LastSeen' => empty($friend->RichPresenceMsg) ? 'Unknown' : strip_tags($friend->RichPresenceMsg),
```
Jamiras, line 85: The old logic appears to have been converting empty and "Unknown" to `_`. I'm not sure why - this seems better and is consistent with `GetExtendedFriendsList`.

resources/views/pages-legacy/userInfo.blade.php
```diff
@@ -176,7 +176,7 @@
     if ($userWallActive) {
         // passing 'null' for $user disables the ability to add comments
         RenderCommentsComponent(
-            !isUserBlocking($userPage, $user) ? $user : null,
+            !$userPageModel->isBlocking(request()->user()) ? $user : null,
```
Jamiras, line 179:
```
App\Models\User::isBlocking(): Argument #1 ($user) must be of type App\Models\User, null given, called in /var/www/html/storage/framework/views/4d718b16de0564107bea562ee3b2c8ee.php on line 218
```
if not logged in.

wescopeland: Thanks, updated latest to also include a call to `request()->user()` in the conditional.

The inner call to `request()->user()` (which is cached) is still there. I considered lifting all this up to another variable, but wasn't sure what to name the variable that wouldn't make it ambiguous with all the other stuff happening around the user page.

app/Helpers/database/user-relationship.php
wescopeland: So strange, thanks for sharing your findings. The user I tested against was the same one who is following ~1800 people. My response time went up from 90ms to 130ms. Since this is only called from the Connect API under rare circumstances, you're likely right that this is nothing worth worrying about.

Jamiras (approved):

PR #2385, opened 2024-04-27
refactor(SiteAwards): swap all User reads to user_id

wescopeland:
This PR changes all `SiteAwards`/`PlayerBadge` `User` reads to instead read from the new `user_id` column. To mitigate risk, a migration to drop `User` has not been included. This can come in a cleanup PR that happens after this makes its way to prod.

There should be no functional changes included in this branch.


app/Helpers/database/site-award.php
```diff
@@ -211,80 +195,84 @@ function SetCertifiedLegend(User $user, bool $enable): void
...
+        )
+        ->orderBy('AwardDate', 'DESC')
+        ->offset($offset)
+
+        // We don't want to show multiple rows per user for the same game if they earned
+        // multiple awards very quickly (ie: beat a game and also mastered it simultaneously).
+        // Unfortunately, using window functions like ROW_NUMBER() with Eloquent ORM is
+        // a huge hassle and arguably makes this code much more difficult to maintain. Rather than
+        // try to use a window function, we'll slightly overfetch from the database, then manually
+        // filter the records down later.
```
wescopeland, line 221: Want to call out this change specifically - I spent about an hour fiddling around with getting `ROW_NUMBER()` to work with this Eloquent ORM migration. Ultimately, I decided the complexity trade-off really wasn't worth it, so I fell back to this overfetching solution.

resources/views/components/cards/game.blade.php
wescopeland, line 1: Some mild cleanup in here I decided to do while fixing the game card. FWIW, there is a functional difference in these comment styles. `<!-- -->` gets shipped to the end user's browser. `{{-- --}}` doesn't, as it gets expunged by the server, so some bytes are saved over the wire.

app/Helpers/database/site-award.php
Jamiras: Does this work correctly with paging? If It takes 60 records to display 50 on the first page, will going to page 2 offset by 60 or 50?

wescopeland: It doesn't - I made an oversight.

Getting this function to work with Eloquent ORM has proven to be quite challenging. I've reverted it back to the SQL implementation and adjusted the read to be from `user_id` instead of `User`. We can handle the Eloquent ORM migration later as a separate piece of work.

Jamiras (approved):

PR #2389, opened 2024-04-29
feat(permissions): rename Hub Manager, add Cheat Investigator and Game Editor

wescopeland:
This PR makes some slight tweaks to the permissions matrix:

1. _Hub Manager_ has been renamed to _Game Hash Manager_.
Reasoning: this is an administrative/staff role that covers much more than the management of hubs. This role is basically @televandalist.

2. A new role, _Game Editor_ has been added.
Reasoning: there isn't a role that neatly covers the responsibilities of site cleanup (management of game metadata and hub curation). Game Editor will fill this gap. This PR doesn't implement any behaviors for the Game Editor role - that will come in a future change. This role is beneficial because there is at least one user performing site cleanup duties whom is not a full developer.

3. _Tester_ has been renamed to _Play Tester_.
Reasoning: at the time these roles were conceived, I don't believe there was a formal QAM team. Renaming this role to Play Tester helps remove ambiguity over whether this role represents Play Testing or QAM responsibilities (for which no role currently exists).

4. A new role, _Cheat Investigator_ has been added.
Reasoning: RACheats needs the ability to use a subset of moderation tooling. This hidden role, assignable only by staff, has no public-facing UX similar to Admin and Root. This PR grants users who have this role to access a desirable subset of moderation tooling.

---

**SQL Edits**
No change is required to the DB seeder. However, we'll need to create and update a few values in the `auth_roles` table.

```sql
-- rename hub-manager to game-hash-manager
UPDATE auth_roles
SET name = 'game-hash-manager',
    updated_at = NOW()
WHERE name = 'hub-manager';

-- rename tester to play-tester
UPDATE auth_roles
SET name = 'play-tester',
    updated_at = NOW()
WHERE name = 'tester';

-- add cheat-investigator
INSERT INTO auth_roles (name, display, guard_name, created_at, updated_at)
VALUES ('cheat-investigator', 0, 'web', NOW(), NOW());

-- add game-editor
INSERT INTO auth_roles (name, display, guard_name, created_at, updated_at)
VALUES ('game-editor', 5, 'web', NOW(), NOW());
```


config/roles.php
```diff
@@ -14,18 +14,20 @@
 
 $adminAssignable = [
     Role::ARTIST,
-    Role::DEVELOPER,
     Role::DEVELOPER_JUNIOR,
     Role::DEVELOPER_STAFF,
     Role::DEVELOPER_VETERAN,
+    Role::DEVELOPER,
```
Jamiras, line 20: I think these were meant to be sorted alphabetically.

config/roles.php
```diff
@@ -64,7 +66,7 @@
     // creator roles assigned by admin
 
     [
-        'name' => Role::HUB_MANAGER,
+        'name' => Role::GAME_HASH_MANAGER,
```
Jamiras, line 69: Do we need a migration for these?

resources/views/pages/user/[user]/game/[game]/compare.blade.php
```diff
@@ -67,12 +64,12 @@
...
 
-        @if ($canModerate)
+        @can('viewSessionHistory', App\Models\PlayerGame::class)
             <x-hidden-controls-toggle-button>Moderate</x-hidden-controls-toggle-button>
-        @endif
+        @endcan
     </div>
 
-@if ($canModerate)
+@can('viewSessionHistory', App\Models\PlayerGame::class)
```
Comment (Jamiras) [inline resources/views/pages/user/[user]/game/[game]/compare.blade.php:72]: Is there some reason to prefer calling `can` twice versus storing the result in a variable?

config/roles.php
wescopeland: Thanks, fixed in latest.

config/roles.php
wescopeland: I don't think so - we're not going to be modifying any table structures. We just need to run the SQL commands in the PR description before the release.

resources/views/pages/user/[user]/game/[game]/compare.blade.php
Comment (wescopeland) [reply resources/views/pages/user/[user]/game/[game]/compare.blade.php:72]: Yes. It's worth noting that using `@can()` twice in this way does indeed call the policy function each time it's used in the template, but this is a measured choice and reflects a preference for reinforcing the declarative nature of the Blade template.

My opinion is that template languages should emphasize clarity and simplicity. By using `@can()` directly, the template makes it immediately clear to the reader what permissions are required for the relevant sections. The maintainer doesn't need to cross-check an imperative variable.

This is a weakly-held opinion though. Happy to revert if you prefer the `$canModerate` approach.

Comment (Jamiras) [reply resources/views/pages/user/[user]/game/[game]/compare.blade.php:72]: I generally prefer to minimize function calls for performance reasons. What if the policy requires a DB query?

In this case it probably doesn't matter.

Jamiras (approved):

PR #2390, opened 2024-04-30
fix(ui): tooltips initialize to proper position (no first frame jank)

### LINKED ISSUE #2155: Tooltip card position skips for 1 frame
**Describe the bug**
When landing on a tooltipped element and immediately stopping the cursor, then resuming motion, you will observe that the tooltip "jumps" to the correct position. This can sometimes result in a disorienting experience.

**To Reproduce**
1. Go to a user profile.
2. Move your cursor to an achievement badge and immediately stop movement.
3. Resume movement.

**Expected behavior**
Tooltip motion should be fluid and seamless, always following the current cursor position.

https://github.com/RetroAchievements/RAWeb/assets/3984985/3e31b981-4ddf-4bed-97f4-f93fba925a4f

christianlegge:
Fixes #2155.

Tooltips are created on a 70ms delay from when the user hovers over an anchor. Previously, the timeout callback captured the mouse position, so they would be created at the position of the initial mouseover, resulting in a slight flicker (or, if you deliberately stopped moving the mouse before 70ms passed, it would stick to the incorrect position). This change fixes the tooltip functions to take an offset instead, leaving the mouse events free to update the base position during the 70ms delay.

I had a quick look to see if this was inside the scope for unit tests and I'm not sure if there's a good way to test in that context; seems like more of an integration/E2E issue, but I'm not the most experienced with that. If we want a unit test for this I can try to figure out how to do it.


wescopeland (approved):
Thanks, very much appreciated. This has been driving me crazy for a while now.

PR #2391, opened 2024-04-30
refactor(Achievement): switch several Author reads to user_id

wescopeland:
This PR switches several `Achievement` table reads that rely on `Author` to instead query by `user_id`. There should be no functional changes in this PR.

There are several calls elsewhere in the codebase that still rely on the `Author` value. I'll tackle those in a separate PR.


app/Helpers/database/achievement.php
```diff
@@ -27,13 +27,15 @@ function getAchievementsList(
...
-    $innerJoin = "";
+    $innerJoinPlayerAchievements = "";
     $withAwardedDate = "";
     if ($params > 0 && $user) {
-        $innerJoin = "LEFT JOIN player_achievements AS pa ON pa.achievement_id = ach.ID AND pa.user_id = " . $user->id;
+        $innerJoinPlayerAchievements = "LEFT JOIN player_achievements AS pa ON pa.achievement_id = ach.ID AND pa.user_id = " . $user->id;
         $withAwardedDate = ", COALESCE(pa.unlocked_hardcore_at, pa.unlocked_at) AS AwardedDate";
     }
 
+    $innerJoinUserAccounts = "LEFT JOIN UserAccounts AS ua ON ua.ID = ach.user_id";
```
Jamiras, line 37: Why use a variable for this? There's other `INNER JOIN`s directly in the query.

resources/views/components/game/claim-info.blade.php
```diff
@@ -72,7 +72,7 @@
         } else {
             // player_sessions only exist after 14 Oct 2023
             $achievement = Achievement::where('GameID', $gameId)
-                ->where('Author', $claim['User'])
+                ->where('user_id', $playerGame->user->id)
```
Jamiras, line 75: This fails if the claimant has never played the game:
```
Attempt to read property "id" on null
```

app/Helpers/database/achievement.php
wescopeland: Good catch. This is leftover cruft from refactoring work. Fixed in latest.

resources/views/components/game/claim-info.blade.php
wescopeland: Thanks. Resolving this proved to be quite the rabbit hole.

The component currently suffers from an N+1 query problem in `master`. This is easily observable on game 20255.

I fixed in the error in latest, but I also rewrote nearly all of these queries to rely on eager-loaded model relationships in order to resolve the N+1 query issue.

resources/views/components/game/devbox-claim-management.blade.php
```diff
@@ -37,11 +36,11 @@
 $userHasClaimSlot = false;
 $openTickets = null;
 
+$userPermissions = $user?->getAttribute('Permissions') ?? 0;
```
Jamiras, line 39: Recommend using enum as fallback value: `Permissions::Unregistered`

wescopeland: Agreed, done in latest.

Jamiras (approved):

PR #2392, opened 2024-05-01
fix(leaderboardinfo): remediate pagination regression

wescopeland:
Remediates a regression accidentally introduced in https://github.com/RetroAchievements/RAWeb/pull/2376.

After that PR's changes, game leaderboards no longer render user ranks correctly after pagination (the count always starts at 1).

In this branch, I've dialed back some of that PR's Eloquent ORM changes. _leaderboardinfo.php_ now again relies on `GetLeaderboardData()`, however, this function now reads from the new `leaderboard_entries` table.


Jamiras:
Not sure how it specifically applies to the changes here, but the Next button is missing on my SMB 1-1 leaderboard. It appears as expected on several other leaderboards I've looked at.

Prod:
image

master:
image

this branch:
image

wescopeland:
@Jamiras 

To be sure I'm checking the right thing, are you referring specifically to `leaderboardinfo.php?i=74041`?

Here is what I see:

**Prod**
Prod


**This branch**
This Branch

Jamiras:
> are you referring specifically to `leaderboardinfo.php?i=74041`?

Yes. And I am ranked (644 per the screenshot above). Can you try a my user?

It appears to be because of this line:
```
if ($numActualEntries == $count) {
```
`numActualEntries` is 51. `count` is 50.

wescopeland:
That worked. `$numEntries` was not being set correctly, which was causing `$lastEntry` to never be truthy, which ultimately resulted in `$numActualEntries` not being tallied as it should've been.

Fixed in latest.

Jamiras (approved):

PR #2393, opened 2024-05-02
Guided ticket creation

Jamiras:
Converts the create ticket page to blade and adds a "phone tree" in front of it. The "Report Issue" button on achievement pages now goes to a page with several options for types of issues to report.

If you've never played the game, there's a limited number of options - all of which are redirected to site accounts for dealing with them as documented in the Contact Us page:
image

If you have unlocked the achievement, an additional option appears for creating a Triggered at the Wrong Time ticket:
image

If you haven't unlocked the achievement, two more options appear for creating a Did not Trigger ticket and for requesting a manual unlock (through RAdmin):
image

When you click on one of the Create Ticket buttons, you'll be taken to the existing Create Ticket page, with a few minor changes:
* The Game and Achievement are no longer listed as fields. They appear in the header/breadcrumb.
* A large block of text was added near the Description field to better indicate what sort of information would be useful to a developer.
* The issue/emulator/version/core/mode fields will be initially populated with any available information, but all will remain editable in case the best guesses need to be adjusted.
image


resources/views/pages/achievement/[achievement]/report-issue.blade.php
```diff
@@ -0,0 +1,148 @@
...
+$unlockedSoftcore = false;
+$unlockedManually = false;
+$hasHardcoreUnlocks = false;
+$hasSoftcoreUnlocks = false;
+$hasSession = false;
+
+$playerAchievements = request()->user()->playerAchievements()
+    ->join('Achievements', 'player_achievements.achievement_id', '=', 'Achievements.ID')
+    ->where('Achievements.GameID', $achievement->game->id)
+    ->get();
```
Comment (wescopeland) [inline resources/views/pages/achievement/[achievement]/report-issue.blade.php:28]: It may be better to rely on the model relations over using an explicit join:

```php
// example
$user
  ->playerAchievements()
  ->whereHas("achievement.game", function ($query) use ($achievement) {
    $query->where("id", $achievement->game->id);
  })
  ->with("achievement.game")
  ->get();
```

Note in this example, no column names are actually used (even `GameID` is abstracted away). This means when a column gets renamed to implement one of the many TODOs in our models, this won't break.

The response shape of this query is going to be slightly different than the current implementation.

resources/views/pages/achievement/[achievement]/report-issue.blade.php
```diff
@@ -0,0 +1,148 @@
...
+    $hasSoftcoreUnlocks |= $playerAchievement->unlocked_at !== null;
+    $hasSession = true;
+}
+
+if ($unlockedHardcore || ($unlockedSoftcore && !$hasHardcoreUnlocks)) {
+    $ticketType = TicketType::TriggeredAtWrongTime;
+}
+
+if (!$hasSession) {
+    $hasSession = request()->user()->playerGames()->where('game_id', $achievement->game->id)->exists();
```
Comment (wescopeland) [inline resources/views/pages/achievement/[achievement]/report-issue.blade.php:49]: Is this a good use case for `ActsAsPlayer::hasPlayed()`?

```php
request()->user()->hasPlayed($achievement->game);
```

resources/views/pages/achievement/[achievement]/report-issue.blade.php
```diff
@@ -0,0 +1,148 @@
...
+        The achievement type (progression/win/missable) is not correct.
+    </x-ticket.guide-link>
+
+    @php
+        $subject = urlencode("Unwelcome Concept: $achievement->title ({$achievement->game->title})");
+        $message = urlencode("I'd like to report an unwelcome concept in [ach=$achievement->id]:\n(Describe the issue here)");
+    @endphp
+    <x-ticket.guide-link text="Message DevCompliance"
+                         link="{{ route('message.create') }}?to=DevCompliance&subject={{ $subject }}&message={{ $message }}">
+        <span>The achievement contains an <a href="https://docs.retroachievements.org/Unwelcome-Concepts/">unwelcome concept</a>.</span>
```
Comment (wescopeland) [inline resources/views/pages/achievement/[achievement]/report-issue.blade.php:136]: Recommend removing this `<span>` and wrapping the `{{ $slot }}` itself in the component with a `<p>` tag:

```php
<div class="flex items-center">
    <p>{{ $slot }}</p>
</div>
```

This does intentionally enforce that the description text must be simply text.

config/missing-page-redirector.php
```diff
@@ -58,7 +58,7 @@
         /*
          * achievement tickets
          */
-        '/reportissue.php' => '/achievement/{i}/tickets/create',
+        '/reportissue.php' => '/achievement/{i}/report-issue',
```
wescopeland, line 61: Was there an intention for us at one point to support an `/achievement/{id}/tickets` route?

resources/views/pages/achievement/[achievement]/report-issue.blade.php
Comment (Jamiras) [reply resources/views/pages/achievement/[achievement]/report-issue.blade.php:28]: added a `scopeForGame(Game)` to both `PlayerAchievement` and `PlayerGame`.

resources/views/pages/achievement/[achievement]/report-issue.blade.php
Comment (Jamiras) [reply resources/views/pages/achievement/[achievement]/report-issue.blade.php:49]: yep ;)

resources/views/pages/achievement/[achievement]/report-issue.blade.php
Comment (Jamiras) [reply resources/views/pages/achievement/[achievement]/report-issue.blade.php:136]: done.

app/Platform/RouteServiceProvider.php
```diff
@@ -74,6 +75,9 @@ protected function mapWebRoutes(): void
             Route::get('game/{game}/dev-interest', GameDevInterestController::class)->name('game.dev-interest');
             Route::get('game/{game}/suggest', [SuggestGameController::class, 'forGame'])->name('game.suggest');
 
+            Route::get('achievement/{achievement}/create-ticket', [AchievementController::class, 'createTicket'])->name('achievement.create-ticket');
```
Jamiras, line 78: Given your question about the old tickets route, I've changed this to `achievement/{achievement}/tickets/create`

resources/views/components/ticket/guide-link.blade.php
```diff
@@ -0,0 +1,15 @@
+@props([
+    'link' => '', // string
```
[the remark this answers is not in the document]
wescopeland: Thank you for highlighting that - I didn't realize this pattern was this prevalent in the codebase.

I generally hold strong opinions about renaming standard HTML attributes in props, as I've found it leads to a poor developer experience and general confusion (it seems we also have some props named `$url` which serve the same purpose). To address this, I plan to open a follow-up PR to rename most, if not all, of these props in our other components to `$href` to align more closely to conventional HTML terminology.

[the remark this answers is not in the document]
wescopeland: Related: https://github.com/RetroAchievements/RAWeb/pull/2403

wescopeland (approved):
Looks good. I think this could be a big help.

PR #2394, opened 2024-05-02
refactor(forum): migrate more stuff to Eloquent ORM, remove last `Author` read

wescopeland:
This is another PR that should introduce no functional changes.

More functions, specifically around topic creation, have been migrated to Eloquent ORM. Additionally, the final `Author` read from the forum feature has been removed in favor of `author_id`.


app/Helpers/database/forum.php
```diff
@@ -335,23 +295,24 @@ function getTopicCommentCommentOffset(int $forumTopicID, int $commentID, int $co
...
-    if ($game->ForumTopicID > 0 && ForumTopic::where('ID', $game->ForumTopicID)->exists()) {
-        // valid topic already exists
-        return false;
+    // If a valid forum topic already exists for the game, bail.
+    if ($game->ForumTopicID > 0 && ForumTopic::find($game->ForumTopicID)->exists()) {
+        return null;
     }
 
-    $forumID = match ($game->system->ID) {
+    // TODO we probably can't get away with hardcoding this indefinitely.
```
Jamiras, line 314: This is broken in more ways than just being hard-coded. Everything that's not in the list goes straight to the Mega Drive forum.

Once we have a system editor, that can be used to associate/create forums for each system. See #650.

There's also plans to allow each game to have it's own forum (instead of just an official forum topic). See #661. At that point, it may be moot to have system-level forums.

public/request/game/generate-forum-topic.php
```diff
@@ -12,8 +13,12 @@
     'game' => 'required|integer|exists:GameData,ID',
 ]);
 
-if (generateGameForumTopic($user, (int) $input['game'], $forumTopicID)) {
-    return redirect(url("/viewtopic.php?t=$forumTopicID"))->with('success', __('legacy.success.create'));
+$userModel = User::firstWhere('User', $user);
```
Jamiras, line 16: I've been assuming anything validated by `authenticateFromCookie` is just `request()->user()`. Is that a bad assumption?

Jamiras (approved):

public/request/game/generate-forum-topic.php
wescopeland: It's a safe assumption based on `authenticateFromCookie`'s implementation details:

```php
/** @var ?User $user */
$user = auth()->user();
```

`$userDetailsOut` is just:

```php
$userDetailsOut = $user->toArray();
```

Eventually we'll remove `authenticateFromCookie()` completely, but a good step forward might be dropping that `->toArray()` call so we don't have to instantiate a `$userModel` value all over the place.

wescopeland: Related: https://github.com/RetroAchievements/RAWeb/pull/2404

PR #2395, opened 2024-05-02
chore(downloads): point RetroArch 'Cores Compatibility List' link to RetroAchievements documentation

Hexadigital:
The current linked Cores Compatibility List for RetroArch points towards the libretro documentation, which is out of date and also recommend cores like Snes9x 2010, which is explicitly marked on our end as "PLEASE STOP USING THIS FOR ACHIEVEMENTS!!!!"

This small PR updates the link used to point towards our documentation, instead, which is both more up to date as well as under our control.


wescopeland (commented):
I'm not opposed to us doing this. However, I can't help but feel the libretro documentation should be properly updated. If it's claiming to support things that aren't actually supported, this seems problematic.

Hexadigital:
I'm open to creating a PR over there as well, but it looks like it hasn't been updated for more than three years (aside from adding new cores as not supported).

wescopeland:
> I still believe that it would be best to point towards our own documentation rather than external documentation

That is the intent of the documentation links. The only ones that point to our internal documentation are for the RA* emulators which we provide for download.

I would like a 2nd opinion, but changing this link specifically seems inconsistent with what the rest of the page is doing. It might be best to put a catch-all somewhere on the page which specifically directs users to our internal documentation.

Hexadigital:
For what it's worth, the actual Documentation button under RetroArch currently points towards our documentation as well.

I'd be fine with an alternate proposal - as long as it has the end result of making sure players aren't reading wildly out of date documentation and using the wrong cores, which leads to either tickets that could have been avoided, or players not unlocking achievements to begin with and then leaving RA.

wescopeland:
> the actual Documentation button under RetroArch currently points towards our documentation as well.

You're right - I misspoke.

Given that, this seems like a reasonable change to me.

wescopeland (approved):

PR #2400, opened 2024-05-03
Convert ticketmanager to blade

Jamiras:
Splits the massive singular view (with various filters combinations) into separate pages sharing logic in a service and blade template.

General changes:
* Ticket Status now only has three options (All, Open, and Resolved). Open includes Request tickets, and Resolved includes Closed tickets. You can no longer mix and match across all the supported ticket statuses.
* Hash, Emulator, and Progression Only filters are no longer present.
* Achievement State now has a third option (All). Previously only Core and Unofficial were available, and they were mutually exclusive.
* You cannot specifically enter a username to view tickets assigned to the user. You can access a list of tickets assigned to a user from their profile page.
* You cannot specifically enter a username to view tickets reported by a user. If this is desirable functionality, we can find somewhere to put it.

Main Ticket Manager (`ticketmanager.php`):
image
* Achievement State default is now All (was Core)

Tickets assigned to user (accessed from user profile page, user profile menu, and ticket notification icon) (`ticketmanager.php?u=`):
image
* Developer filter is not present. All tickets will be assigned to the same user.

Tickets assigned to user in Request state (accessed from ticket/general notification icons) (`ticketmanager.php?p=`):
image
* No filters are shown on this page.

Tickets resolved for others (accessed from Developer stats page) (`ticketmanager.php?r=`):
image
* Ticket status filter is not present. Only resolved tickets will be shown.
* It is no longer possible to view both tickets resolved for self and tickets resolved for others on the same page. To view tickets resolved for self, view the Tickets assigned to user page and change the Ticket Status to Resolved.

Tickets for a game (accessed from game page) (`ticketmanager.php?g=`):
image

Tickets for an achievement (accessed from achievement page) (`ticketmanager.php?a=`):
image
* Developer filter is not present. Only one developer exists for an achievement.
* Achievement Type filter is not present. An achievement cannot be both Core and Unofficial at the same time.


app/Platform/Services/TicketListService.php
```diff
@@ -0,0 +1,214 @@
...
+        return [
+            'status' => $validatedData['filter']['status'] ?? 'unresolved',
+            'type' => (int) ($validatedData['filter']['type'] ?? 0),
+            'achievement' => $validatedData['filter']['achievement'] ?? 'all',
+            'mode' => $validatedData['filter']['mode'] ?? 'all',
+            'developerType' => $validatedData['filter']['developerType'] ?? 'all',
+        ];
+    }
+
+    public function getSelectFilters(bool $showStatus = true, bool $showDevType = true, bool $showAchievementType = true): array
```
wescopeland, line 48: Not entirely sure where it's all coming from, but the first select filter specifically has too much left padding.

**Ticket manager**
Screenshot 2024-05-06 at 4 56 46 PM

**Game lists**
Screenshot 2024-05-06 at 4 57 28 PM

This may be resolvable by modifying the classnames for a specific div in _components/meta-panel/index.blade.php_:

```html
{{-- i added the last classname --}}
<div class="grid sm:flex gap-y-4 sm:divide-x-2 divide-embed-highlight [&>*:first-child]:!pl-0">
```

This classname forces the first child element to always have a left padding value of 0.

resources/views/pages/tickets/index.blade.php
wescopeland, line 1: When I change my filter option, I'm not currently being sent back to page 1.

resources/views/pages/tickets/most-reported-games.blade.php
```diff
@@ -0,0 +1,64 @@
...
+    ->join('Achievements', 'Achievements.ID', '=', 'Ticket.AchievementID')
+    ->select('GameID', DB::raw('count(*) AS TicketCount'))
+    ->groupBy('GameID')
+    ->orderBy('TicketCount', 'DESC')
+    ->take(100)
+    ->get();
+
+@endphp
+
+<x-app-layout pageTitle="Ticket Manager">
```
wescopeland, line 24: Recommend changing the page title to "Most Reported Games"

resources/views/pages/tickets/most-reported-games.blade.php
```diff
@@ -0,0 +1,64 @@
...
+                    <th>Game</th>
+                    <th class="text-right">Number of Open Tickets</th>
+                </tr>
+            </thead>
+
+            <tbody>
+                @foreach ($ticketedGames as $ticketedGame)
+                    @php $game = Game::firstWhere('ID', $ticketedGame->GameID) @endphp
+                    <tr>
+                        <td>
```
wescopeland, line 48: A little bit of padding here does seem to help make the list more readable.

```suggestion
                        <td class="py-2">
```

resources/views/pages/tickets/[achievement].blade.php
Comment (wescopeland) [inline resources/views/pages/tickets/[achievement].blade.php:1]: With [how Folio route parameters work](https://laravel.com/docs/11.x/folio#route-parameters), _[achievement].blade.php_, _[game].blade.php_, and _[user].blade.php_ are all in conflict with each other because they reference the same dynamic parameter.

The issue doesn't surface because a controller is handling the data fetching. In the event that the Folio `render()` hook was used instead, I believe these routes would all break.

The _pages_ folder should follow the app routing structure identically. In other words, for the route:

`game/1/tickets`

There should be a _resources/views/pages/game/[game]/tickets.blade.php_ file.

resources/views/components/meta-panel/select-filter.blade.php
```diff
@@ -10,9 +10,7 @@
 ?>
 
 <script>
-function handleSelectFilterChanged(event) {
-    const kind = '{{ $kind }}';
-
+function handleSelectFilterChanged(kind, event) {
```
wescopeland, line 13: nit: Conventionally in JS ecosystem handler functions, `event` is always the first arg.

wescopeland (commented):
I'm about 70% done, but need to step away for a little while. I'll finish the remaining 30% tomorrow, but felt it'd be productive to go ahead and share where I'm at.

resources/views/pages/tickets/most-reported-games.blade.php
Jamiras: Yep. I got the header below, but not the HTML title.

resources/views/pages/tickets/most-reported-games.blade.php
Jamiras: done

resources/views/pages/tickets/[achievement].blade.php
Comment (Jamiras) [reply resources/views/pages/tickets/[achievement].blade.php:1]: I wondered about that, but felt better leaving them all together as they have basically the same template and it would be more obvious which files need to change if we ever decided to make a uniform change to them.

resources/views/components/meta-panel/select-filter.blade.php
Jamiras: done

resources/views/pages/tickets/index.blade.php
Jamiras: fixed

app/Platform/Services/TicketListService.php
Jamiras: It's a styling difference between Sorts and Filters:
https://github.com/RetroAchievements/RAWeb/blob/1689d8cd943551457a0c45ca5eed15f440e98487/resources/views/components/meta-panel/index.blade.php#L32-L33
https://github.com/RetroAchievements/RAWeb/blob/1689d8cd943551457a0c45ca5eed15f440e98487/resources/views/components/meta-panel/index.blade.php#L41-L43

wescopeland: Thanks for sharing. That is indeed the problem.

I believe I have an elegant solution to this problem.

The solution is to remove all the padding classes on those child divs so they all read as:
```html
<div class="grid gap-y-1">
```

and change the parent div to:
```html
<div
    class="
        grid sm:flex gap-y-4 sm:divide-x-2 divide-embed-highlight
        sm:*:px-4 xl:*:px-8 first:*:pl-0 last:*:pr-0
    "
>
```

_However_, this `*:` syntax is only supported with Tailwind v3.4, which is a few minor releases above our version. There are no breaking changes from our version to v3.4.

To upgrade, open _package.json_, and adjust line 49 to:
```
"tailwindcss": "^3.4.3",
```
Then, run `npm install` in your terminal while in the RAWeb directory. Then restart `npm run dev`.

app/Community/Controllers/TicketController.php
wescopeland, line 1: `TicketListService` should be instantiated via DI. It makes sense to create new instances of these things in legacy Blade files, but in other classes we should lean more into Laravel's DI system.

```php
public function __construct(
    protected TicketListService $ticketListService,
) {
}
```

_UserProgressionStatus.php_ is a good example of a class that follows a similar pattern.

public/API/API_GetTicketData.php
```diff
@@ -209,36 +209,70 @@
...
 }
 
+$getTicketsInfo = function (Builder $tickets, int $offset, int $count): array {
+    $result = [];
+
+    /** @var Ticket $ticket */
+    foreach ($tickets->orderBy('ReportedAt', 'DESC')->offset($offset)->take($count)->get() as $ticket) {
+        $ticket->loadMissing(['achievement', 'reporter', 'resolver']);
+        $ticket->achievement->loadMissing('game');
+        $ticket->achievement->game->loadMissing('system');
```
wescopeland, line 224: I'm particularly concerned about this as it's inducing an N+1 query problem. We should perform these loads outside the loop if possible.

```php
$tickets = $tickets->with([
    'achievement.game.system',
    'reporter',
    'resolver',
])
    ->orderBy('ReportedAt', 'DESC')
    ->offset($offset)
    ->take($count)
    ->get();
```

app/Community/Controllers/TicketController.php
Jamiras: TicketListService contains state for the current request. Is it safe to have an instance registered in the controller? Will separate controller instances be created per request? In that case, shouldn't the service only be instantiated for the methods that require it?

wescopeland: This is only a concern if `TicketListService` state is reused across multiple subsequent requests. As best as I can tell, it isn't.

Each time one of these page views is requested, Laravel is going to create an entirely new instance of the controller, and along with it, a new instance of `TicketListService` will get injected through the constructor. The instance will be used for that specific page view and then discarded when the response is sent back to the client.

app/Platform/Services/TicketListService.php
Jamiras: done

public/API/API_GetTicketData.php
Jamiras: done

app/Community/Controllers/TicketController.php
Jamiras: done

package.json
wescopeland, line 1: Looks like we may be missing the accompanying _package-lock.json_ change that would have automatically occurred after `npm install`. Do you have a _package-lock.json_ file with local changes?

Jamiras: > Do you have a package-lock.json file with local changes?

No. And from the error messages I get when running `sail npm install` (where sail is an alias for `sudo vendor/bin/sail`), it seems like that's something managed by docker. 
```
npm ERR! code EACCES
npm ERR! syscall open
npm ERR! path /var/www/html/package-lock.json
npm ERR! errno -13
npm ERR! Error: EACCES: permission denied, open '/var/www/html/package-lock.json'
npm ERR!  [Error: EACCES: permission denied, open '/var/www/html/package-lock.json'] {
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'open',
npm ERR!   path: '/var/www/html/package-lock.json'
npm ERR! }
npm ERR! 
npm ERR! The operation was rejected by your operating system.
npm ERR! It is likely you do not have the permissions to access this file as the current user
npm ERR! 
npm ERR! If you believe this might be a permissions issue, please double-check the
npm ERR! permissions of the file and its containing directories, or try running
npm ERR! the command again as root/Administrator.
```

Since the alias is already using sudo, I don't know how to appease the error. It doesn't seem to affect anything, so I've been ignoring it for months.

wescopeland: Interesting; I execute `npm install` and `npm run dev` outside of sail, just literally those commands with no `sail` prefix.

If `npm install` in the RAWeb directory doesn't update _package-lock.json_, don't worry about it. It does need to be kept current to ensure stable builds, but we can push an update for it after this PR is merged to keep it current.

Jamiras: I may have messed up my folder permissions doing it wrong. `npm install` generates a similar error, but `sudo npm install` does seem to work without generating errors and does update the package.json

wescopeland (approved):

PR #2401, opened 2024-05-03
add dorequest?r=gameinfolist

Jamiras:
Supports https://github.com/RetroAchievements/RAIntegration/pull/1080

To display the recently played games in the DLL overlay, we need the game title and image for all recently played games. Normally, we pull this from the cached patch data, but that expires after 30 days. For older records, we effectively fetch the patch data again. If we have to do this for a lot of games, the server sends back 429 Too Many Requests errors.

Instead of making N requests for patch data, this allows for a single request to be made for all of the needed game titles/icons, and avoids the overhead of also returning the achievement/leaderboard/rich presence data.

### DIFF
--- a/public/dorequest.php
+++ b/public/dorequest.php
@@ -222,6 +222,7 @@ switch ($requestType) {
         $response['CodeNotes'] = getCodeNotesData($gameID);
         $response['GameID'] = $gameID;
         break;
+
     case "gameid":
         $md5 = request()->input('m') ?? '';
         $response['GameID'] = getGameIDFromMD5($md5);
@@ -232,9 +233,18 @@ switch ($requestType) {
         $response['Response'] = getGamesListDataNamesOnly($consoleID);
         break;
 
-    case "officialgameslist":
+    case "officialgameslist": // TODO: is this used anymore? It's not used by the DLL.
         $consoleID = (int) request()->input('c', 0);
         $response['Response'] = getGamesListDataNamesOnly($consoleID, true);
+        break;
+
+    case "gameinfolist":
+        $gamesCSV = request()->input('g', '');
+        if (empty($gamesCSV)) {
+            return DoRequestError("You must specify which games to retrieve info for", 400);
+        }
+        $response['Response'] = Game::whereIn('ID', explode(',', $gamesCSV))
+            ->select('Title', 'ID', 'ImageIcon')->get()->toArray();
         break;
 
     case "hashlibrary":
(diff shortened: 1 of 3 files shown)


public/dorequest.php
```diff
@@ -232,11 +233,20 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
         $response['Response'] = getGamesListDataNamesOnly($consoleID);
         break;
 
-    case "officialgameslist":
+    case "officialgameslist": // TODO: is this used anymore? It's not used by the DLL.
```
wescopeland, line 236: It's (unfortunately) used by 3rd party front-ends tapping into the Connect API, eg: Batocera.

wescopeland: Maybe we can reimplement in the web API somehow and have this fall through to the web API to avoid a breaking change?

Jamiras: I would think that `API_GetGameList?f=1` would be preferred. It returns the ID/Title for each game of a console, along with the image, and numbers of achievements, leaderboards, and points.

wescopeland: For reference: https://github.com/batocera-linux/batocera-emulationstation/blob/3959d0742f1b3fb9e3b86e56f365d5eea99eaf5d/es-app/src/RetroAchievements.cpp#L449

Jamiras: Even better. `API_GetGameList?f=1&h=1` returns a list of ID/Titles for each game in a console, plus the known associated hashes. 

The only thing that would have to change is they'd have to query each console individually, or we'd have to open up `API_GetGameList` to support the "0" (all) console.

Jamiras: I can't help wonder how useful this functionality even is for them, given how few of our consoles do full file hashes.

public/dorequest.php
```diff
@@ -232,11 +233,20 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
         break;
 
+    case "gameinfolist":
+        $gamesCSV = request()->input('g', '');
+        if (empty($gamesCSV)) {
+            return DoRequestError("You must specify which games to retrieve info for", 400);
+        }
+        $response['Response'] = Game::whereIn('ID', explode(',', $gamesCSV))
+            ->select('Title', 'ID', 'ImageIcon')->get()->toArray();
+        break;
```
wescopeland, line 248: Should we consider adding an upper bound to the count of game IDs allowed in the call?

Even with nginx's default maximum URL length of 8,192 characters, someone could craft a GET request which fetches data for ~1,500 games per request.

Jamiras: The intended client (the DLL) won't be requesting any more than 20. I'm fine with adding an upper limit. The public APIs seem to use 100 for similar requests

wescopeland (approved):

PR #2406, opened 2024-05-07
refactor: read from more user id fields

wescopeland:
Another PR with no functional changes. This is a PR which adjusts all the helpers currently reading from `Author` columns to read from user ID values instead.


app/Helpers/database/achievement.php
```diff
@@ -65,9 +65,9 @@ function getAchievementsList(
         $query .= "AND pa.unlocked_at IS NULL ";
     }
 
-    if (isValidUsername($developer)) {
-        $bindings['author'] = $developer;
-        $query .= "AND ach.Author = :author ";
+    if (isValidUsername($developer?->username)) {
```
Jamiras, line 68: This can probably just be `if ($developer)`. I wouldn't expect an invalid username to resolve into a model.

app/Helpers/database/achievement.php
```diff
@@ -65,9 +65,9 @@ function getAchievementsList(
...
+        $bindings['author'] = $developer->username;
+        $query .= "AND ua.User = :author ";
```
Jamiras, line 70: The point of this ticket is to move away from matching the string. Recommend:
```
$query .= "AND ach.user_id = $developer->id";
```

app/Helpers/database/ticket.php
```diff
@@ -516,22 +509,20 @@ function getNumberOfTicketsClosedForOthers(string $user): array
...
               LEFT JOIN UserAccounts as ua2 ON ua2.ID = t.resolver_id
-              LEFT JOIN Achievements as a ON a.ID = t.AchievementID
+              LEFT JOIN Achievements as ach ON ach.ID = t.AchievementID
               WHERE t.ReportState IN (" . TicketState::Closed . "," . TicketState::Resolved . ")
-              AND ua.User NOT LIKE '$user'
-              AND a.Author LIKE '$user'
-              AND a.Flags = '3'
+              AND ua.ID != {$user->id}
+              AND ach.user_id = {$user->id}
+              AND ach.Flags = '3'
```
Jamiras, line 525: You updated the other lines to use the enum. Recommend doing it here too.

app/Helpers/database/user.php
```diff
@@ -327,7 +328,7 @@ function GetDeveloperStatsFull(int $count, int $offset = 0, int $sortBy = 0, int
     $query = "SELECT ua.ID, COUNT(*) AS OpenTickets
               FROM Ticket tick
               INNER JOIN Achievements ach ON ach.ID=tick.AchievementID
-              INNER JOIN UserAccounts ua ON ua.User=ach.Author
+              INNER JOIN UserAccounts ua ON ua.ID = ach.user_id
```
Jamiras, line 331: I think this `ua` could be eliminated too.

app/Helpers/database/user.php
```diff
@@ -341,7 +342,7 @@ function GetDeveloperStatsFull(int $count, int $offset = 0, int $sortBy = 0, int
               INNER JOIN UserAccounts as ua ON ua.ID = tick.resolver_id
               INNER JOIN Achievements as ach ON ach.ID = tick.AchievementID
               WHERE tick.resolver_id != tick.reporter_id
-              AND ach.Author != ua.User
+              AND ach.user_id != ua.ID
```
Jamiras, line 345: I think this `ua` could be eliminated too.

app/Helpers/database/achievement.php
wescopeland: Yep, good point. Fixed in latest.

app/Helpers/database/achievement.php
wescopeland: Updated in latest to:

```php
$bindings['userId'] = $developer->id;
$query .= "AND ach.user_id = :userId ";
```

app/Helpers/database/ticket.php
```diff
@@ -516,22 +509,20 @@ function getNumberOfTicketsClosedForOthers(string $user): array
...
               FROM Ticket AS t
               LEFT JOIN UserAccounts as ua ON ua.ID = t.reporter_id
               LEFT JOIN UserAccounts as ua2 ON ua2.ID = t.resolver_id
-              LEFT JOIN Achievements as a ON a.ID = t.AchievementID
+              LEFT JOIN Achievements as ach ON ach.ID = t.AchievementID
               WHERE t.ReportState IN (" . TicketState::Closed . "," . TicketState::Resolved . ")
-              AND ua.User NOT LIKE '$user'
-              AND a.Author LIKE '$user'
-              AND a.Flags = '3'
+              AND ua.ID != {$user->id}
```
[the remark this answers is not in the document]
wescopeland: Done in latest

app/Helpers/database/ticket.php
wescopeland: Done in latest

app/Helpers/database/user.php
wescopeland: Yep, good call. Done in latest

app/Helpers/database/user.php
wescopeland: This was one a bit trickier, but now is also eliminated in latest

Jamiras (approved):

PR #2409, opened 2024-05-08
feat: add CommentPolicy::create()

wescopeland:
https://discord.com/channels/476211979464343552/1002689485005406249/1237681025233059890

Muted devs can still comment on their own tickets.


resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -317,37 +317,39 @@ function AwardManually(hardcore) {
...
-                @foreach ($commentData as $comment)
-                    @php
-                        $when = Carbon::createFromTimestamp($comment['Submitted']);
-                        $commentUser = 
-                            ($comment['User'] === $ticket->reporter->User) ? $ticket->reporter :
-                            (($comment['User'] === $user->User) ? $user :
-                                User::firstWhere('User', $comment['User']));
-                    @endphp
+            <table id="feed" class="table-highlight">
+                <tbody>
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:321]: This is all just fixed indention. `<table>` and `<tbody>` were indented the same, so I moved this `<tbody>` to the next line.

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -317,37 +317,39 @@ function AwardManually(hardcore) {
...
+                            articleType="{{ ArticleType::AchievementTicket }}"
+                            :articleId="$ticket->ID"
+                            :commentId="$comment['ID']"
+                            :allowDelete="$allowDelete"
+                        />
+                    @endforeach
+
+                    @can('create', [App\Models\Comment::class, $ticket])
+                        {!! RenderCommentInputRow($user->User, ArticleType::AchievementTicket, $ticket->ID) !!}
+                    @endcan
```
Comment (wescopeland) [inline resources/views/pages/ticket/[ticket].blade.php:350]: This is the actual logical change in this file.

app/Policies/CommentPolicy.php
```diff
@@ -26,6 +28,37 @@ public function view(?User $user, Comment $comment): bool
         return true;
     }
 
+    public function create(User $user, ?Model $commentable): bool
+    {
+        if ($user->isMuted()) {
+            // Even when muted, developers may still comment on tickets for their own achievements.
```
Jamiras, line 34: Should this be limited to unresolved tickets?
`TicketState::isOpen($commentable->ReportState)`

I've had users thank me after resolving a ticket, so I assume we'd want to allow two-way communication.

app/Helpers/render/comment.php
```diff
@@ -92,8 +92,7 @@ function RenderCommentsComponent(
             ]);
     }
 
-    if (isset($user) && !$user->is_muted) {
-        // User comment input:
+    if ($user?->can('comment', [App\Models\Comment::class])) {
```
Jamiras, line 95: I'm unable to comment on anything _except_ tickets. I think `comment` should be `create`, but that causes exceptions:
```
Too few arguments to function App\Policies\CommentPolicy::create(), 1 passed in /var/www/html/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 811 and exactly 2 expected
```

app/Policies/CommentPolicy.php
Jamiras: Similarly, should a muted user be allowed to comment on an achievement they authored, or the related game?

app/Policies/CommentPolicy.php
```diff
@@ -26,6 +28,37 @@ public function view(?User $user, Comment $comment): bool
         return true;
     }
 
+    public function create(User $user, ?Model $commentable): bool
```
Jamiras, line 31: There appear to be separate policies for each of the supported model types. Not sure how those should (or do) interact with this.
* AchievementCommentPolicy
* GameCommentPolicy
* NewsCommentPolicy
* UserCommentPolicy

Maybe this is just for the other types: like ticket, hashes, and claims.

app/Policies/CommentPolicy.php
wescopeland: Frankly, I think we're already being pretty generous just with this change consider a target user will actively be under moderation.

I very much support the idea of locking this down further to only tickets that are still open and have made that change in latest. Personally, I don't think we should extend this to authored achievements or associated games unless the team feels it's an absolute necessity.

app/Helpers/render/comment.php
wescopeland: Sorry about that, I incorrectly tested this code path. Fixed in latest.

app/Policies/CommentPolicy.php
wescopeland: That was my assumption too - `CommentPolicy` is just a catch-all for "everything else". The other policies shouldn't have any bearing on this one.

Jamiras (approved):

PR #2410, opened 2024-05-09
fix: update link to avoid PCSX2 redirect in docs

falsepopsky:
The previous link leads to a redirect URL handled by the PCSX2 site. It's better to update the link here to avoid relying on their redirect, which could be removed without warning and cause a 404 error.


falsepopsky:
failing tests are unrelated to this change, also if something can take a look at this question that i made in the discord group : https://discord.com/channels/310192285306454017/1237238010366656595

wescopeland (approved):

PR #2411, opened 2024-05-09
(tickets) don't show recent history region for deleted user

Jamiras:
For soft-deleted user (http://localhost:64000/ticket/12345) and null foreign key (http://localhost:64000/ticket/12531)


wescopeland (approved):
Verified locally.

PR #2412, opened 2024-05-09
fix error redirecting ticketmanager with no parameters

Jamiras:
Fixes " Undefined array key 0 " error trying to access ticketmanager.php with no parameters.


wescopeland (approved):
Verified locally.

PR #2415, opened 2024-05-09
fix(leaderboard-listing): exclude untracked users

wescopeland:
http://localhost:64000/leaderboardinfo.php?i=3547

**Before**
Screenshot 2024-05-09 at 6 02 09 PM

**After**
Screenshot 2024-05-09 at 6 02 03 PM

### DIFF
--- a/resources/views/components/game/leaderboards-listing/leaderboards-list-item.blade.php
+++ b/resources/views/components/game/leaderboards-listing/leaderboards-list-item.blade.php
@@ -19,10 +19,6 @@ use App\Platform\Enums\ValueFormat;
 
         @php
             $bestScore = $leaderboard->entries()
-                ->whereHas('user', function ($query) {
-                    $query->where('Untracked', '!=', 1)
-                        ->whereNull('unranked_at');
-                })
                 ->orderBy("score", $leaderboard->rank_asc ? 'ASC' : 'DESC')
                 ->first();
         @endphp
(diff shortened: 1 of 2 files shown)


resources/views/components/game/leaderboards-listing/leaderboards-list-item.blade.php
```diff
@@ -19,6 +19,10 @@ class="leading-3"
 
         @php
             $bestScore = $leaderboard->entries()
+                ->whereHas('user', function ($query) {
+                    $query->where('Untracked', '!=', 1)
+                        ->whereNull('unranked_at');
+                })
```
Jamiras, line 25: Shouldn't this just be part of `entries()`? Would we ever want to return untracked users?

It looks like `ActsAsPlayer::scopeTracked()` and the `GetLeaderboardData` query filter out untracked users unless they're the logged in user. I think we should revisit that logic too. If a user is untracked, we could potentially show where they would be in the leaderboard, but the leaderboard rankings for tracked users shouldn't be shifted when an untracked user views the list.

wescopeland: Agreed, doing it on `entries()` is probably the right call. I've updated `entries()` in latest to account for this. It also accepts an optional boolean arg in the event we do want those untracked records.

For showing where a user would theoretically be placed on the leaderboard, I don't have a terribly strong opinion. It might be easier to implement after we migrate leaderboardinfo.php wholesale from the legacy pages.

Jamiras (approved):

PR #2417, opened 2024-05-09
fix(Shortcode): prevent potential crash from malformed link

wescopeland:
If `[link]` is malformed, return an empty string.

Unfortunately, we have no way to access the original `$input` to use that as a fallback. This is a stopgap fix to prevent the page from crashing outright.

### DIFF
--- a/app/Support/Shortcode/Shortcode.php
+++ b/app/Support/Shortcode/Shortcode.php
@@ -197,7 +197,7 @@ final class Shortcode
         $url = $shortcode->getParameter('url') ?: $shortcode->getContent();
 
         if (empty($url)) {
-            return '';
+            return '[broken link]';
         }
 
         // Ensure the correct protocol prefix (http/https) is being used.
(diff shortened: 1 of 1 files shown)


app/Support/Shortcode/Shortcode.php
```diff
@@ -196,6 +196,10 @@ private function renderLink(ShortcodeInterface $shortcode): string
         // Determine the URL from the shortcode's 'url' parameter or content.
         $url = $shortcode->getParameter('url') ?: $shortcode->getContent();
 
+        if (empty($url)) {
+            return '';
```
Jamiras, line 200: Would it be better to return something indicating a problem occurred? i.e. "[unsupported link]" or "[broken link]"?

Jamiras (approved):

app/Support/Shortcode/Shortcode.php
wescopeland: No objections here. Updated in latest to return "[broken link]".

Jamiras (approved):

PR #2420, opened 2024-05-10
chore: drop LeaderboardEntry

wescopeland:
This PR switches the remaining leaderboard reads over to `leaderboard_entries` and drops the legacy `LeaderboardEntry` table.

We're no longer writing to the table. I've archived a backup of the table just in case we need it later.


database/migrations/platform/2024_05_10_000000_drop_leaderboardentry_table.php
```diff
@@ -0,0 +1,29 @@
...
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::dropIfExists('LeaderboardEntry');
```
Jamiras, line 12: There's still some references to this table in `getLeaderboardsForGame`.

app/Platform/Controllers/DeveloperFeedController.php
```diff
@@ -73,15 +73,15 @@ private function attachRecentLeaderboardEntryRowsMetadata(mixed $entryRows): mix
         $entryRows->transform(function ($row) use ($userData, $gameData, $consoleData) {
             $game = $gameData[$row->GameID] ?? null;
 
-            $row->User = $userData[$row->UserID]->User ?? null;
-            $row->Untracked = $userData[$row->UserID]->Untracked ?? null;
+            $row->User = $userData[$row->user_id]->User ?? null;
```
Jamiras, line 76: Users aren't appearing in recent leaderboard entries on the developer feed.

wescopeland: Thanks, I saw the rows populating but I didn't look closely enough to realize users were missing. The `pluck()` call needed to be changed to target the right field. Resolved in latest.

database/migrations/platform/2024_05_10_000000_drop_leaderboardentry_table.php
wescopeland: Thanks. I believe this may actually turn out to be the most impactful change in the PR.

It appears `getLeaderboardsForGame()` is only being used to populate a count value called `$numLeaderboards` on game pages. In latest I've simply refactored this to `$gameModel->leaderboards()->visible()->count()` and deleted the function.

The SQL query in `getLeaderboardsForGame()` was accounting for _half the loading performance_ of several game pages on my local, clocking in at roughly 100-300ms. The new Eloquent query takes less than 1ms, a tremendous improvement.

Jamiras (approved):

Issue #2378, opened 2024-04-22
Screwed up ">" signs when editing forum posts


PopovMark:
Same thing as #2343.
Changes this symbol to 

&
g
t
; 

when you hit Edit. Spaced it with line breaks because it's shown as the "greater than" symbol otherwise.

wescopeland:
Rather than continuing to band-aid this, I'm going to migrate the page out of _pages-legacy_.

Alfex3:
Now, when writing a comment, it makes the text after the < symbol disappear. I found out when I looked at tickets for which I provided a fix and comments were cut where that simbol was. I think it's treated like an HTML tag.

wescopeland:
https://github.com/RetroAchievements/RAWeb/pull/2421

PR #2421, opened 2024-05-10
refactor: migrate edit post view to Folio

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/2378.

Migrates _editpost.php_ out of _pages-legacy_ and into a Folio view. No functional changes.


Jamiras:
I'm going to need some guidance on this. Doing my normal checkout branch, composer install, npm install, and npm run build, I can't even access the forums page:
```
Laravel\Folio\FolioRoutes::get(): Argument #2 ($arguments) must be of type array, int given, called in /var/www/html/vendor/laravel/folio/src/FolioServiceProvider.php on line 75
```
And what's with all the empty `.getkeep` files?

wescopeland:
Apologies for that. When I was testing, I was hitting the route directly, but I forgot to update the `route()` call for this page on the topic view. This is now fixed in latest.

I've also removed the .gitkeep files -- these were mostly just placeholders for stuff in _missing-page-redirector.php_.

You should now be able to view topics normally again.

This is the block of code that does things a little differently (and inspired the mess of PRs I opened today):

```php
render(function (View $view, ForumTopicComment $forumTopicComment) {
    if (!Auth::user()->can('update', $forumTopicComment)) {
        return abort('401');
    }
    
    return $view->with([
        'category' => $forumTopicComment->forumTopic->forum->category,
        'forum' => $forumTopicComment->forumTopic->forum,
        'forumTopic' => $forumTopicComment->forumTopic,
    ]);
});
```

This `render()` hook [builds all the data for the page](https://laravel.com/docs/11.x/folio#render-hooks) without needing to connect anything to a `RouteServiceProvider`. It also has access to DI, so we can inject services directly into these (indeed, in some of the PRs I just converted existing controllers to services).

When a `RouteServiceProvider` is used, it actually overrides/disables all of Folio's built-in features, including middleware. My rush of PRs today was to migrate all of our Folio routes to actually use Folio and load data via `render()` hooks.

config/missing-page-redirector.php
```diff
@@ -108,14 +108,14 @@
         /*
          * forums
          */
-        '/forum.php' => '/forums/category/{c}',
+        '/forum.php' => '/forums/category/{c}', // TODO what if no category is specified?
```
Jamiras, line 111: The way it's implemented, if `c` is not in the query params, it won't get replaced, so it'll try to open `/forums/category/{c}` and fail.

The redirector is not used as long as the legacy blade still exists, so it's not a problem that needs to be addressed now, but the solution is to implement a multi-redirect like I did for ticketmanager.php:
```
        '/forum.php' => [
            'c' => '/forums/category/{c}',
            '' => '/forums',
        ],
```

resources/views/pages/forums/post/[ForumTopicComment]/edit.blade.php
```diff
@@ -45,16 +37,16 @@
...
-                <x-base.form.input label="{{ __res('forum-topic', 1) }}" readonly :value="$topic->title" inline />
+                <x-base.form.input label="{{ __res('forum', 1) }}" readonly value="{{ $forum->title }}" inline :fullWidth="false" />
+                <x-base.form.input label="{{ __res('author', 1) }}" readonly value="{{ $forumTopic->user?->display_name ?? 'Deleted user' }}" inline :fullWidth="false" />
+                <x-base.form.input label="{{ __res('forum-topic', 1) }}" readonly :value="$forumTopic->title" inline />
 
                 <x-base.form.textarea
                     id="input_compose"
                     label="{{ __res('message', 1)}} "
-                    :value="$foundPost->Payload"
+                    :value="$forumTopicComment->body"
```
Comment (Jamiras) [inline resources/views/pages/forums/post/[ForumTopicComment]/edit.blade.php:49]: How does `forumTopicComment` get passed here? Doesn't it have to be in the `$view->with([` list?

Comment (wescopeland) [reply resources/views/pages/forums/post/[ForumTopicComment]/edit.blade.php:57]: Folio treats all dynamic fields (`[thing]`) as globally-available context values. They can be "peeked" into in `render()`. More docs here: https://laravel.com/docs/11.x/folio#route-model-binding

```php
// [ForumTopicComment] --> $forumTopicComment
render(function (View $view, ForumTopicComment $forumTopicComment) { ... });
```

However, there's no harm in also putting it in `with()`. I also re-added `@props()` for improved documentation.

Jamiras (approved):

PopovMark:
The issue with the > signs is still happening. Please fix.

wescopeland:
@PopovMark You're going to need to be more specific if I'm going to debug the issue.

How exactly is the issue reproduced? When I naively edit posts, I do not see the &lt; or &gt; appearing on either edit or save.

PopovMark:
Nvm, the signs don't break anymore when I submit it like that, they still show up properly in the posts.

PR #2425, opened 2024-05-11
refactor(games): migrate 'pages' routes to use Folio-native routing and render hooks

wescopeland:
Related to https://github.com/RetroAchievements/RAWeb/pull/2424, this PR migrates:

`/games/suggest`
`/game/{game}/dev-interest`
`/game/{game}/suggest`

to use Folio-native routing.

A few other game-related routes have been updated to pass data to the view using Folio's `render()` hook.


app/Platform/Services/SuggestGamesService.php
```diff
@@ -467,4 +352,117 @@ public function selectSimilarGame(int $gameId): int
 
         return 0;
     }
+
+    private function initializeUserProgress(?User $user): void
```
wescopeland, line 356: No functional changes in this file starting at this line. I just moved all the private functions so they'd be grouped together.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -541,7 +541,7 @@ function resize() {
                     $interestedUsers = UserGameListEntry::where('type', UserGameListType::Develop)
                         ->where('GameID', $gameID)
                         ->count();
-                    echo "<div><a class='btn btn-link' href='" . route('game.dev-interest', $gameID) . "'>View Developer Interest ($interestedUsers)</a></div>";
+                    echo "<div><a class='btn btn-link' href='" . route('game.dev-interest', ['game' => $gameID]) . "'>View Developer Interest ($interestedUsers)</a></div>";
```
Jamiras, line 544: There must still be a broken route somewhere on this page. I get the "`Argument #2 ($arguments) must be of type array, int given`" error trying to view a game.

resources/views/pages/game/[game]/dev-interest.blade.php
```diff
@@ -1,57 +1,67 @@
...
+use Illuminate\View\View;
+
+use function Laravel\Folio\{middleware, name, render};
 
 middleware(['auth', 'can:view,game', 'can:develop']);
 name('game.dev-interest');
 
-?>
+render(function (View $view, Game $game, GameDevInterestPageService $pageService) {
+    // TODO use a policy -- lift it up to the Folio page middleware. "can:viewDeveloperInterest,game"?
```
Comment (Jamiras) [inline resources/views/pages/game/[game]/dev-interest.blade.php:15]: I believe `can:develop` should be sufficient, but I can't figure out where that logic lives to validate that assumption. Maybe it should be `can:develop,game` to validate set claim for jr. devs?

Jamiras (dismissed):

resources/views/pages-legacy/gameInfo.blade.php
wescopeland: Found it - fixed in latest.

resources/views/pages/game/[game]/dev-interest.blade.php
Comment (wescopeland) [reply resources/views/pages/game/[game]/dev-interest.blade.php]: Took me a while to hunt that one down too. It lives in _app/Providers/AuthServiceProvider.php_:

```php
Gate::define('develop', fn (User $user) => $user->hasAnyRole([
    Role::DEVELOPER,
    Role::ARTIST,
    Role::WRITER,
])
    // TODO remove as soon as permission matrix is in place
    || $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper);
```

Given this implementation, I wasn't too sure.

Comment (Jamiras) [reply resources/views/pages/game/[game]/dev-interest.blade.php]: That doesn't look like it would allow Jr. Devs access to their claimed sets. It probably should, but that can be addressed as a separate PR.

Jamiras (approved):

PR #2427, opened 2024-05-11
refactor(messages): use Folio-native routing and render hooks

wescopeland:
Related to #2425 and #2424. Migrates messages routes to use Folio-native routing and `render()` hooks. Right now all the Folio stuff is being overwritten by `RouteServiceProvider`.

There should be no functional changes.


resources/views/pages/message-thread/[messageThread].blade.php
```diff
@@ -1,53 +1,37 @@
...
 use App\Support\Shortcode\Shortcode;
-use Illuminate\Support\Carbon;
-use function Laravel\Folio\{middleware, name};
+use Illuminate\Support\Facades\Auth;
+use Illuminate\View\View;
 
-middleware(['auth', 'can:view,messageThread']);
+use function Laravel\Folio\{middleware, name, render};
+
+middleware(['auth', 'can:view,messageThread']); // TODO add 'verified' middleware
```
Comment (Jamiras) [inline resources/views/pages/message-thread/[messageThread].blade.php:11]: I can view my Messages list, but attempting to open any of them returns a 401.

resources/views/pages/messages/create.blade.php
```diff
@@ -1,18 +1,28 @@
 <?php
 
 use App\Models\Message;
-use function Laravel\Folio\{middleware, name};
+use Illuminate\View\View;
 
-middleware(['auth', 'can:create,' . Message::class]);
+use function Laravel\Folio\{middleware, name, render};
+
+middleware(['auth', 'can:create,' . Message::class]); // TODO add 'verified' middleware
```
Jamiras, line 8: I get a `Argument #2 ($arguments) must be of type array, int given` error after submitting a newly created message.

Jamiras (changes requested):

resources/views/pages/messages/create.blade.php
wescopeland: Resolved in latest.

resources/views/pages/message-thread/[messageThread].blade.php
Comment (wescopeland) [reply resources/views/pages/message-thread/[messageThread].blade.php]: Strange, I am not able to repro this locally after trying numerous threads. However, `MessageThreadPolicy::view()` is just a stub, so I've removed `can:view,messageThread` from the middleware call on the page.

Comment (Jamiras) [reply resources/views/pages/message-thread/[messageThread].blade.php]: Now I get `Laravel\Folio\InlineMetadataInterceptor::{closure}(): Argument #2 ($messageThread) must be of type App\Models\MessageThread, string given, called in /var/www/html/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php on line 36` trying to view individual messages

Comment (wescopeland) [reply resources/views/pages/message-thread/[messageThread].blade.php]: This one is a bit perplexing. I did find a broken route that I just pushed a commit to fix for, but I'm not sure if it's related. 

Screenshot 2024-05-16 at 6 09 00 PM

That error message does seem to be complaining about the 2nd arg of the `render()` hook on the message thread page.

On the latest commit, if you're still seeing the error, can you try running:

```
sail artisan cache:clear
sail artisan view:clear
```

Comment (Jamiras) [reply resources/views/pages/message-thread/[messageThread].blade.php]: Same error. Tried as a couple other users too.

Comment (wescopeland) [reply resources/views/pages/message-thread/[messageThread].blade.php]: Still can't repro on my end no matter what I try 🙁. I've tried various users and whatnot at this point, no luck.

The error seems to be suggesting that a string is being given for `$messageThread` instead of an actual `MessageThread` model.

Latest commit has a couple more fixes to try:
* Pass in a whole `$message` entity to the route instead of the ID
* Rename the Blade file to point directly to the `id` field.

If this still doesn't work, the nuclear option might be to read `request()->getPathInfo()` and query for the thread itself. But I'm very puzzled why I'm unable to repro this issue. When I land on the page from an inbox, I very clearly see Folio kicking in and loading the model:

Screenshot 2024-05-16 at 8 22 14 PM

Comment (Jamiras) [reply resources/views/pages/message-thread/[messageThread].blade.php]: I backtracked through your commits, and the one that mattered was renaming the file.
`[messageThread].blade.php` generates the error.
`[MessageThread:id].blade.php` works.

For what it's worth, `[MessageThread:id].blade.php` generates a not-authorized error with the `'can:view,messageThread'` middleware clause. Not sure if that helps.

It seems like this is not a desirable change. What can I do to further identify the problem?

Comment (wescopeland) [reply resources/views/pages/message-thread/[messageThread].blade.php]: I'm beginning to wonder if there could possibly be something OS-specific going on. My suspicion is driven by a line from the Folio documentation, which seems to suggest there is some OS-based logic in Folio itself:

---

Screenshot 2024-05-17 at 5 15 03 PM

---

`[MessageThread:id]` should be synonymous with `[messageThread]`. Using `[MessageThread:id]` isn't desirable due to inconsistency, but should be functionally equivalent with just `[messageThread]`.

In latest, based on a hunch, I've added `protected $table = 'message_threads';` to the `MessageThread` model. There's nothing wrong with that model inferring the name of its table... but I'm wondering if there's something in Folio that's looking for `$table`.

With this change applied in the latest commit, can you rename [MessageThread:id] back to [messageThread] and see if the error still happens? If not, we can probably just proceed with [MessageThread:id].

Comment (Jamiras) [reply resources/views/pages/message-thread/[messageThread].blade.php]: The comment there is because ":" is not a valid character in Windows paths, which is yet another reason we shouldn't name the file `[MessageThread:id]`.

fwiw, `[messagethread].blade.php` generates a 404, so having the capital T is causing it to try to do _something_.

This also works, but shouldn't be necessary:
```
render(function (View $view, string $messageThread, MessageThreadService $pageService) {
    $user = Auth::user();
    $currentPage = (int) (request()->input('page.number') ?? 1);

    $messageThread = MessageThread::find($messageThread);

    return $view->with($pageService->buildForMessageThreadViewData($user, $messageThread, $currentPage));
});
```

Comment (wescopeland) [reply resources/views/pages/message-thread/[messageThread].blade.php]: I'm particularly nervous about different behavior happening in different environments. The only way we see the same behavior on both of our machines is if I coerce Folio into pointing directly at the `id` field for the `MessageThread` model on this page.

To align with the doc and per your feedback regarding Windows paths, I've updated the filename to `[MessageThread-id].blade.php`. This still works great on my end - if it works on yours I think we're probably safe. I'm very puzzled why this problem is happening at all - I suspect it is a bug with Folio itself.

Comment (Jamiras) [reply resources/views/pages/message-thread/[messageThread].blade.php]: I can confirm `[MessageThread-id].blade.php` also works on my VM.

[This](https://stackoverflow.com/questions/43654601/laravel-eloquent-route-model-binding-not-working-with-multi-word-table-name) and [this](https://stackoverflow.com/questions/72241894/laravel-implicit-binding-routing-with-more-than-two-word) suggest it's a casing thing with multi-word model names, but I couldn't make that work.

Jamiras (approved):

PR #2429, opened 2024-05-11
 refactor(system.game.index): migrate to Folio-native routing and render hook

wescopeland:
No functional changes.

Migrates `SystemController::games()` to `SystemGamesPageService::buildViewData()`. Removes reference from `RouteServiceProvider`. Adds `render()` hook to the Folio page view.


app/Platform/Services/SystemGamesPageService.php
```diff
@@ -0,0 +1,154 @@
...
+use Illuminate\Http\Request;
+
+class SystemGamesPageService
+{
+    public function __construct(
+        protected GameListService $gameListService,
+    ) {
+    }
+
+    public function buildViewData(Request $request, System $system): array
```
Jamiras, line 18: You could argue that this should just be part of `GameListService`, as that builds a large portion of the view data, then this just tweaks it a bit.

Jamiras (approved):

app/Platform/Services/SystemGamesPageService.php
wescopeland: That is indeed arguable - I do like the present separation of concerns though. It's conceivable `GameListService` could receive even more functionality in the future.

PR #2436, opened 2024-05-12
fix: use ->developer reads on Achievement model

wescopeland:
Adjusts all remaining instances where something is accessing the `Author` attribute of the `Achievement` model. These now point to the `developer` relation.

This PR also fixes a bug where the "OP" indicator on forums is not displaying correctly on reply posts.


resources/views/components/ticket/ticket-list.blade.php
```diff
@@ -74,7 +74,7 @@
                                     :consoleName="$game->system->Name"
                                 />
                             </td>
-                            <td>{!! userAvatar($ticket->achievement->author) !!}</td>
+                            <td>{!! userAvatar($ticket->achievement->developer->display_name) !!}</td>
```
Jamiras, line 77: `userAvatar` accepts a `User`. Suggest just passing `$ticket->achievement->developer` to eliminate re-fetching all the developer UserAccount records.

resources/views/pages/ticket/[ticket].blade.php
```diff
@@ -181,7 +181,7 @@
                 <div class="relative w-full p-2 bg-embed rounded">
                     <x-ticket.stat-element label="Achievement">{!! achievementAvatar($ticket->achievement, iconSize:16) !!}</x-ticket.stat-element>
                     <x-ticket.stat-element label="Game">{!! gameAvatar($ticket->achievement->game, iconSize:16) !!}</x-ticket.stat-element>
-                    <x-ticket.stat-element label="Author">{!! userAvatar($ticket->achievement->author, iconSize:16) !!}</x-ticket.stat-element>
+                    <x-ticket.stat-element label="Author">{!! userAvatar($ticket->achievement->developer->display_name, iconSize:16) !!}</x-ticket.stat-element>
```
Comment (Jamiras) [inline resources/views/pages/ticket/[ticket].blade.php:184]: Same comment about passing developer. Less impact with a singular record.

resources/views/components/ticket/ticket-list.blade.php
wescopeland: Good point - that slipped my mind. Addressed in latest.

resources/views/pages/ticket/[ticket].blade.php
Comment (wescopeland) [reply resources/views/pages/ticket/[ticket].blade.php]: Resolved in latest.

Jamiras (approved):

PR #2437, opened 2024-05-12
feat(multiset): create the achievement_set_game_hashes pivot table

wescopeland:
To support the compatibility matrix between ROM hashes and non-Core achievement sets, this PR creates a new pivot table named achievement_set_game_hashes. This table associates supported ROM hashes with Bonus and Specialty sets, allowing for the exclusion of incompatible subsets until they are made compatible.

The `achievement_set_game_hashes` table has the following columns:
* `achievement_set_id` (foreign key referencing the achievement_sets table)
* `game_hash_id` (foreign key reference the game_hashes table)
* `compatible` (boolean)

This pivot table enables the platform to manage the compatibility between ROM hashes and subsets, ensuring that incompatible sets do not load during the user’s play session for their specified ROM.

Additionally, this PR adds a few multiset-related model files, and adds a seeder for game achievement sets.

With all this combined, we can start to do some interesting things.

After running the migration, you can execute the individual seeder by running:
```
sail artisan db:seed --class=AchievementSetSeeder
```

The seeder will grab 10 random games in your database that have more than 6 hashes. From those 10 random games, it'll create a healthy mix of core sets, bonus sets, and specialty sets that are associated with those games.

---

Once our database is migrated and seeded, we can then run through a few interesting behaviors.
```php
// In my case, game 65 was seeded with a core set, some bonus sets, and a specialty set.
$game = Game::with("gameAchievementSets.achievementSet")->find(65);

// "This game has 5 sets."
dump("This game has {$game->gameAchievementSets->count()} sets.");

foreach ($game->gameAchievementSets as $gameAchievementSet) {
  dump("{$gameAchievementSet->achievementSet->id} {$gameAchievementSet->type}");
  /**
   * "323 core"
   * "324 bonus"
   * "325 bonus"
   * "326 bonus"
   * "327 specialty"
   */
}

// Given a hash, get its compatible game achievement sets. I picked some hashes after
// poking around in the game_achievement_sets and achievement_set_game_hashes tables.
$normalGameMd5 = "4ada3f1cb2306565035a6a6ecbe915d5";
$specialtyGameMd5 = "6ddca3ea05ff5b42a52240020713d5f6";

// Let's dump all the compatible achievement set IDs for $normalGameMd5.
dump("Loading game hash {$normalGameMd5}");
$foundNormalGameHash = GameHash::with([
  "achievementSets" => function ($query) {
    $query->wherePivot("compatible", true);
  },
  "achievementSets.gameAchievementSets"
])->firstWhere("md5", $normalGameMd5);

/**
 * "Loading game hash 4ada3f1cb2306565035a6a6ecbe915d5"
 * "Found 4 compatible sets."
 * "Achievement Set: [323:core]"
 * "Achievement Set: [324:bonus]"
 * "Achievement Set: [325:bonus]"
 * "Achievement Set: [326:bonus]"
 */
dump("Found {$foundNormalGameHash->achievementSets->count()} compatible sets.");
foreach ($foundNormalGameHash->achievementSets as $achievementSet) {
  foreach ($achievementSet->gameAchievementSets as $gameAchievementSet) {
    dump(
      "Achievement Set: [{$achievementSet->id}:{$gameAchievementSet->type}]"
    );
  }
}

// Let's dump all the compatible achievement set IDs for $specialtyGameMd5.
dump("Loading game hash {$specialtyGameMd5}");
$foundSpecialtyGameHash = GameHash::with([
  "achievementSets" => function ($query) {
    $query->wherePivot("compatible", true);
  },
  "achievementSets.gameAchievementSets"
])->firstWhere("md5", $specialtyGameMd5);

/**
 * "Loading game hash 6ddca3ea05ff5b42a52240020713d5f6"
 * "Found 5 compatible sets."
 * "Achievement Set: [323:core]"
 * "Achievement Set: [324:bonus]"
 * "Achievement Set: [325:bonus]"
 * "Achievement Set: [326:bonus]"
 * "Achievement Set: [327:specialty]"
 */
dump(
  "Found {$foundSpecialtyGameHash->achievementSets->count()} compatible sets."
);
foreach ($foundSpecialtyGameHash->achievementSets as $achievementSet) {
  foreach ($achievementSet->gameAchievementSets as $gameAchievementSet) {
    dump(
      "Achievement Set: [{$achievementSet->id}:{$gameAchievementSet->type}]"
    );
  }
}
```


database/factories/AchievementSetFactory.php
```diff
@@ -0,0 +1,30 @@
...
+ * @extends Factory<AchievementSet>
+ */
+class AchievementSetFactory extends Factory
+{
+    protected $model = AchievementSet::class;
+
+    public function definition(): array
+    {
+        return [
+            'user_id' => User::factory(),
```
Jamiras, line 21: Not sure I understand why a set would have a user. A set is a group of achievements, which can each have their own unique author.

Also, this creates new users. Shouldn't it just pick a random developer from the database, or preferably, a user associate to one of the achievements in the set?

database/factories/GameAchievementSetFactory.php
```diff
@@ -0,0 +1,41 @@
...
+        ];
+    }
+
+    public function type(string $type): self
+    {
+        return $this->state(function (array $attributes) use ($type) {
+            return [
+                'type' => $type,
+                'title' => $type === AchievementSetType::Core ? null : $this->faker->words(3, true),
+                'order_column' => $type === AchievementSetType::Core ? 0 : $this->faker->numberBetween(1, 10),
```
Jamiras, line 37: This field/column is weirdly named. Are you sure this is how it was meant to be used?

Jamiras (approved):
I think I understand the intent of these changes, but I'm having trouble visualizing the validity of the data without any associated UI.

database/factories/GameAchievementSetFactory.php
wescopeland: Yes. There are TODOs in all models with a `DisplayOrder` field to rename that field to `order_column`.

`order_column` for `GameAchievementSet` will likely just represent the order in which each set is rendered on a game page, similar to the ordering of achievements.

database/factories/AchievementSetFactory.php
wescopeland: > Not sure I understand why a set would have a user. A set is a group of achievements, which can each have their own unique author.

I agree. I didn't question why `user_id` was on the table, but I don't think a set should necessarily be ascribed to a user unless it's some kind of "community-made set" (??).

I've updated the seeder to respect that `user_id` is nullable, and it will no longer set a value for that field.

PR #2440, opened 2024-05-13
enforce order by title for gameslist response

Jamiras:
Fixes https://discord.com/channels/310192285306454017/533411674162593812/1239126267052298240


tests/Feature/Connect/GamesListTest.php
```diff
@@ -28,109 +29,135 @@ public function testGamesList(): void
...
-            ]);
+            ])
+            // assertExactJson doesn't enforce order. this ugly code pulls the array keys from
+            // the Response subobject and validates the order of them.
+            ->assertJson(function (AssertableJson $json) {
+                $json->has("Response", function (AssertableJson $json2) {
+                    $this->assertEquals(array_keys($json2->toArray()), [5, 1, 7]);
+                    $json2->etc(); // prevent "Unexpected properties" error for not directly interacting with "Response"
+                })->etc(); // prevent "Unexpected properties" error for not directly interacting with "Success"
+            });
```
wescopeland, line 69: I'm wondering if it makes sense to create a reusable assertion helper - this seems like it could be a use case that comes up again in the future.

This seems like it could be a use case that comes up again in the future. I'm wondering if it makes sense to create a reusable assertion helper.

I fiddled around for a little while and here's what I came up with.

Example usage in _GamesListTest.php_:
```php
// all games for console 1
$this->assertExactOrderedJson(
    $this->get($this->apiUrl('gameslist', ['c' => $system1->id])),
    [
        'Success' => true,
        'Response' => [
            '5' => 'Five',
            '1' => 'One',
            '7' => 'Seven',
        ],
    ]
);
```

Added util to _TestCase.php_:
```php
protected function assertExactOrderedJson(TestResponse $response, array $expectedJson): void
{
    $response->assertExactJson($expectedJson);

    $expectedOrder = $this->extractOrder($expectedJson);
    $responseOrder = $this->extractOrder($response->json());

    $this->assertEquals($expectedOrder, $responseOrder);
}

private function extractOrder(array $array, array $path = []): array
{
    $order = [];

    foreach ($array as $key => $value) {
        if (is_array($value)) {
            $order = array_merge($order, $this->extractOrder($value, array_merge($path, [$key])));
        } else {
            $order[] = array_merge($path, [$key]);
        }
    }

    return $order;
}
```

Example usage:
```php
// all games for console 1
$this->assertExactJsonWithOrder(
    $this->get($this->apiUrl('gameslist', ['c' => $system1->id])),
    [
        'Success' => true,
        'Response' => [
            '5' => 'Five',
            '1' => 'One',
            '7' => 'Seven',
        ],
    ]
);
```

Jamiras: I don't like that your proposal enforces strict ordering for the entire response. We don't really case if 'Success' appears after 'Response'. We only care that the children of 'Response' are in a particular order, which is not really a feature of JSON, but we have existing clients depending on it.

We could potentially encapsulate the logic in a helper function:
```
            ->assertJson(fn (AssertableJson $json) => 
                TestCase::assertJsonKeyOrder($json, "Response", [8, 4, 10, 2]));
```

But until we have another case that requires the logic, there's no reason to add extra stuff in the base class.

wescopeland: Something like this could work with a minor tweak to the util:

```php
$response = $this->get($this->apiUrl('gameslist', ['c' => $system1->id]));
$this->assertEquals(true, $response['Success']);
$this->assertExactOrderedJson($response['Response'], [
    '5' => 'Five',
    '1' => 'One',
    '7' => 'Seven',
]);
```

Nevertheless, I agree with your point - this probably isn't necessary unless we find ourselves with an additional use case for comparing exact key ordering.

wescopeland (approved):

PR #2442, opened 2024-05-15
feat: migrate code notes to Folio and Livewire 3

wescopeland:
This PR migrates the code notes page from _pages-legacy_ to a Folio page powered by Livewire 3.

**Summary of changes**
* Memory addresses are now sticky with the user scroll position.
* The page now uses the permissions matrix / `MemoryNotePolicy`.
* Nearly all TODOs for the `MemoryNote` model have been implemented in a new migration.
* There is still a single instance of `RenderCodeNote()` in the codebase -- all editing logic has been removed from this function.
* The new route is `/game/{game}/notes`.

https://github.com/RetroAchievements/RAWeb/assets/3984985/0310b6c8-8651-4b41-91a7-d9b7c640d703


app/Helpers/database/code-note.php
```diff
@@ -28,33 +28,7 @@ function getCodeNotesData(int $gameID): array
     return $codeNotesOut;
 }
 
-function getCodeNotes(int $gameID, ?array &$codeNotesOut): bool
```
wescopeland, line 31: This function was redundant with `getCodeNotesData()`. I deleted it and migrated all usages to `getCodeNotesData()`.

app/Platform/Livewire/Forms/MemoryNoteForm.php
```diff
@@ -0,0 +1,40 @@
...
+    #[Locked]
+    public ?MemoryNote $memoryNote;
+
+    public function setMemoryNote(MemoryNote $memoryNote): void
+    {
+        $this->memoryNote = $memoryNote;
+        $this->body = $memoryNote->body;
+    }
+
+    public function update(): void
```
wescopeland, line 24: This is the function that executes upon saving a code note on the new page.

composer.json
```diff
@@ -37,6 +37,7 @@
         "laravel/tinker": "^2.8",
         "league/flysystem-aws-s3-v3": "^3.3",
         "livewire/livewire": "^3.0",
+        "livewire/volt": "^1.6",
```
wescopeland, line 40: Probably should've installed this when we began the Folio transition. Volt enables Livewire to play nicely with Folio and use a Folio-like functional syntax. [docs](https://livewire.laravel.com/docs/volt)

resources/views/pages/game/[game]/notes.blade.php
Comment (wescopeland) [inline resources/views/pages/game/[game]/notes.blade.php:1]: This is the entrypoint for the new page.

app/Policies/MemoryNotePolicy.php
```diff
@@ -25,36 +28,55 @@ public function viewAny(User $user): bool
...
+        if ($user->hasAnyRole([Role::DEVELOPER_STAFF, Role::DEVELOPER])) {
+            return true;
+        }
+
+        // Users with the DEVELOPER_JUNIOR role are allowed to update their own notes.
+        if ($user->hasRole(Role::DEVELOPER_JUNIOR) && $user->is($memoryNote->user)) {
+            return true;
+        }
+
+        return false;
```
Jamiras, line 57: Are you intentionally ignoring $user->Permissions? The Edit button is not available for existing Devs currently.

resources/views/pages-legacy/leaderboardList.blade.php
```diff
@@ -26,7 +26,7 @@
 $gameData = getGameData($gameID);
 $codeNotes = [];
 if ($permissions >= Permissions::JuniorDeveloper) {
-    getCodeNotes($gameID, $codeNotes);
+    $codeNotes = getCodeNotesData($gameID);
 }
```
Jamiras, line 30: codeNotes does not appear to be used on this page. these four lines (27-30) could be eliminated.

resources/views/livewire/memory-notes/memory-notes-list-item.blade.php
```diff
@@ -0,0 +1,91 @@
...
+                        Cancel
+                    </button>
+                    
+                    <button
+                        class="btn"
+                        x-show="!isEditing"
+                        x-on:click="isEditing = true"
+                    >
+                        Edit
+                    </button>
```
Jamiras, line 86: Not sure where it's coming from, or if it even matters, but every row is generating an invisible Save button:
```
            <button
                class="btn"
                x-on:click="window.beginSave(window.Livewire.find('YeBoRjAofI7XmqZ8YrOb'), )"
                disabled
                wire:dirty.remove.attr="disabled"
            >
                Save
            </button>
```

app/Policies/MemoryNotePolicy.php
wescopeland: Not intentional, this was an oversight on my part. Addressed in latest.

resources/views/livewire/memory-notes/memory-notes-list-item.blade.php
wescopeland: It could definitely be more efficient. The offending code was here:

```html
<div x-show="isEditing" x-cloak>
    <livewire:memory-notes.memory-note-edit-form
        :memoryNote="$memoryNote"
    />
</div>
```

In a Livewire component, we can conditionally render stuff in real-time with our standard Blade directives such as `@if ()`. However, this is quite slow. Since this is driven by a client-side action, it's better to handle this with Alpine.js.

`x-show` is an Alpine directive that applies an inline style of "display: none;" unless its given condition is met. What this means though is that its DOM nodes are always present.

Here's my updated implementation:
```html
<div x-show="isEditing">
    <template x-if="isEditInitialized">
        <livewire:memory-notes.memory-note-edit-form
            :memoryNote="$memoryNote"
        />
    </template>
</div>
```

Now we're combining `x-show` and `x-if`. `x-if` is very similar to `x-show`, but it directly adds and removes elements from the DOM. The DOM elements aren't present, they're stored efficiently in memory as a `DocumentFragment` (and are still going to be visible in most browser devtools, but will likely be labeled differently).

The nuance here is we can't rely on `x-if` alone to remove those DOM nodes when the user exits Edit Mode. Unfortunately, this breaks Livewire's data binding. With the combination of `x-if` and `x-show`, after the first render in Edit Mode, we return to pure client-side visibility control with CSS.

Hopefully this explanation makes sense.

Jamiras: Something in this change is causes the submissions (delete and update) to not refresh the page/row.

The first note _has been_ deleted. The second _has been_ saved.
image

wescopeland: For several hours today, I attempted to mitigate this issue while retaining the `x-if` directive, but I haven't had any luck.

I've reverted the `x-if` change. Having those hidden buttons probably isn't a big deal for the time being.

One other observation I've made: some routes such as `/game/2721/notes` crash my page locally. This happens on `master` without this PR's changes as well. At some point, for huge lists of notes, we may want to implement pagination & search.

resources/views/livewire/memory-notes/memory-notes-list-item.blade.php
```diff
@@ -0,0 +1,91 @@
...
+        <p class="font-mono" x-show="!isEditing">
+            {!! nl2br(e($memoryNote->body)) !!}
+        </div>
+    </td>
+    
+    <td>
+        <div class="flex justify-center py-1">
+            {!! 
+                userAvatar(
+                    $memoryNote->user->display_name,
```
Jamiras, line 58: http://localhost:64000/game/19173/notes
> Attempt to read property "display_name" on null

http://localhost:64000/game/501/notes
> Allowed memory size of 134217728 bytes exhausted (tried to allocate 56885960 bytes)

wescopeland:
I'm wondering at this point if it would be better to just migrate the entire page to Filament, which gets us things like pagination, searching, and editing almost for free. I'm going to noodle on this a bit and put the PR into draft for the time being.

PR #2444, opened 2024-05-16
refactor: migrate forumposthistory.php to Folio

wescopeland:
Another attempt at https://github.com/RetroAchievements/RAWeb/pull/2158.

This PR migrates _forumposthistory.php_ to two Folio routes:
* `/forums/posts`
* `/user/{user}/posts`

Desktop is basically identical to what's already in prod:

Screenshot 2024-05-15 at 8 50 18 PM

Mobile has been redesigned:

Screenshot 2024-05-15 at 8 50 48 PM


app/Community/Services/ForumRecentPostsPageService.php
```diff
@@ -0,0 +1,94 @@
...
+            )
+                ->toArray();
+        } else {
+            $postsList = getRecentForumTopics(
+                $currentOffset,
+                $maxPerPage,
+                $currentUserPermissions,
+                260,
+            );
+        }
```
wescopeland, line 90: The maximum character count for these calls have both been set to 260.

Rather than rely purely on DB truncation, this is now handled a bit more elegantly in CSS with Tailwind utilities `truncate` and `line-clamp-*`.

app/Helpers/database/forum.php
```diff
@@ -448,6 +450,7 @@ function getRecentForumTopics(int $offset, int $count, int $permissions, int $nu
             WHERE ftc.Authorised=1 AND DateCreated >= DATE_SUB(NOW(), INTERVAL 7 DAY)
             GROUP BY ftc.ForumTopicId
         ) AS d7 ON d7.ForumTopicId = ft.ID
+        LEFT JOIN UserAccounts AS ua ON ua.ID = lftc.author_id
```
wescopeland, line 453: Resolves an N+1 query problem that was previously on _forumposthistory.php_.

resources/views/pages/user/[user]/posts.blade.php
```diff
@@ -0,0 +1,71 @@
...
+render(function (View $view, User $user, ForumRecentPostsPageService $pageService) {
+    $currentUser = Auth::user();
+
+
+    if (!$currentUser->can('viewUserPosts', [\App\Models\ForumTopicComment::class, $user])) {
+        // Traditionally this would probably be a 401. However, 
+        // a 401 might imply the target user has blocked the current user.
+        // Returning a 404 is a bit less presumptuous.
+        abort(404);
+    }
```
Comment (wescopeland) [inline resources/views/pages/user/[user]/posts.blade.php:23]: I want to call out this piece, as it's a minor behavior change.

resources/views/pages/user/[user]/posts.blade.php
```diff
@@ -0,0 +1,79 @@
...
+
+use function Laravel\Folio\{name, middleware, render};
+
+name('user.posts');
+
+render(function (View $view, User $user, ForumRecentPostsPageService $pageService) {
+    $currentUser = Auth::user();
+
+
+    if (!$currentUser->can('viewUserPosts', [\App\Models\ForumTopicComment::class, $user])) {
```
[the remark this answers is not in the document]
Comment (wescopeland) [reply resources/views/pages/user/[user]/posts.blade.php]: Thanks - fixed in latest.

resources/views/pages/forums/posts.blade.php
```diff
@@ -0,0 +1,60 @@
...
+
+use App\Community\Services\ForumRecentPostsPageService;
+use App\Support\Shortcode\Shortcode;
+use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\View\View;
+
+use function Laravel\Folio\{name, middleware, render};
+
+name('forum.posts');
```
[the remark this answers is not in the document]
wescopeland: Not a problem on the name - changed to `forum.recent-posts` in latest.

The only thing I'm less sure about is our models and DB structure seem to imply we actually have multiple forums (plural). For example, my local database has 20 `Forum` rows. This is why I chose the route `/forums/...`.

Maybe this means our route name should be "forums.posts". What do you think?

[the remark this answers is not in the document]
Jamiras: We generally pluralize the type for lists, but singularize it when passing an ID. So `/forums` would be fine, but it should be `/forum/[forum]/recent-posts`. In this case, I think `/forums/recent-posts` makes sense as it applies to all forums.

resources/views/pages/user/[user]/posts.blade.php
```diff
@@ -0,0 +1,79 @@
...
+use App\Community\Services\ForumRecentPostsPageService;
+use App\Models\User;
+use App\Support\Shortcode\Shortcode;
+use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\View\View;
+
+use function Laravel\Folio\{name, middleware, render};
+
+name('user.posts');
```
Comment (Jamiras) [inline resources/views/pages/user/[user]/posts.blade.php:12]: I wonder if this should also be `recent-posts` for consistency. In this case, a user's "recent" posts may not be all that recent.

Jamiras (approved):

resources/views/pages/user/[user]/posts.blade.php
Comment (wescopeland) [reply resources/views/pages/user/[user]/posts.blade.php:12]: I had the same thought, and that ambiguity is what led me to lean towards "user.posts" instead of "user.recent-posts".

I could go either way - do you think it's worth changing? If so, I don't mind.

Comment (wescopeland) [reply resources/views/pages/user/[user]/posts.blade.php:12]: I'll keep this as `user.posts` for the time being, for the reasons you've stated - a user's "recent" posts could have happened months or even years ago. Recency for this specific view is just the sort order (which happens to be hardcoded), but it's not the primary viewing context.

Not a strong opinion though, we can always change it later if it makes more sense to be `user.recent-posts`.

PR #2447, opened 2024-05-19
update connect API to return shared ranks

Jamiras:
Fixes an issue where multiple users with the same score would report unique ranks:
```
1 UserA    20
2 UserB    19
3 UserC    19
4 UserD    19
5 UserE    18
```

Instead, it should be:
```
1 UserA    20
2 UserB    19
2 UserC    19
2 UserD    19
5 UserE    18
```


app/Helpers/database/leaderboard.php
```diff
@@ -75,9 +76,60 @@ function SubmitLeaderboardEntry(
...
+        $score = null;
+        foreach ($entries as &$entry) {
+            if ($entry['Score'] !== $score) {
+                $score = $entry['Score'];
+                $rank = $index;
+            }
+
+            $entry['Rank'] = $rank;
+            $index++;
+        }
```
Jamiras, line 121: This `foreach` is the actual fix. Previously `$entry['Rank']` was just assigned to the index.

wescopeland (approved):
Verified locally.

PR #2448, opened 2024-05-19
feat: denormalize and restore the Followed Users Ranking component

wescopeland:
This PR restores the Followed Users Ranking component on user profiles. The component was previously replaced with a call-to-action due to how intensive it was in regards to DB performance. Now, player daily and weekly points summaries are denormalized into the `player_stats` table so the DB calls for the component are now essentially free, even for users that are following 1000+ other users.

**This PR is stacked on top of https://github.com/RetroAchievements/RAWeb/pull/2446.**

**Before**
Screenshot 2024-05-19 at 2 17 52 PM

**After**
Screenshot 2024-05-19 at 2 17 35 PM

**How it works:**
* A new job has been written, `UpdatePlayerPointsStats`, which inserts denormalized points summary values for a user's daily/weekly hardcore, weighted, and softcore points into the `player_stats` table.
* The job has been configured to dispatch on `PlayerMetricsUpdated` and on `PlayerRankedStatusChanged`.
* For existing records only, the job has been configured to dispatch every hour to keep them up to date.
* When a record `value` hits 0, it is hard deleted from `player_stats` by a tiny command: `DeleteStalePlayerPointsStatsEntries`. This can't be inside `UpdatePlayerPointsStats` itself without causing race conditions.

**How to test this PR:**
`UpdatePlayerPointsStats` is exposed through a new Artisan Console command. The command also accepts mock dates to improve testing.

```bash
## For async stuff, be sure to restart Horizon after pulling down this branch.

# Synchronously write points summaries for one user
sail artisan ra:platform:player:update-points-stats Jamiras

# Synchronously write points summaries for one user for a specific date
sail artisan ra:platform:player:update-points-stats Jamiras --date="2023-11-05"

# Kick off jobs to write points summaries for all users on a specific date
sail artisan ra:platform:player:update-points-stats --date="2023-11-05"

# Kick off jobs to update only existing points summaries for the current date
sail artisan ra:platform:player:update-points-stats --existing-only

# Kick off jobs to write points summaries for all users for the current date
sail artisan ra:platform:player:update-points-stats
```


app/Platform/Actions/UpdatePlayerPointsStats.php
```diff
@@ -0,0 +1,149 @@
...
+            return $playerAchievement->unlocked_hardcore_at !== null;
+        });
+        $softcoreAchievements = $recentPlayerAchievements->filter(function ($playerAchievement) {
+            return $playerAchievement->unlocked_hardcore_at === null;
+        });
+
+        // "day" will be over the last 24 hours.
+        // "week" will be the beginning of the week (server time).
+        $statIntervals = [
+            'day' => Carbon::now()->subDay(),
```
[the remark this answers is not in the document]
wescopeland: No strong feelings on this. In latest, I've set it to `->startOfDay()` for consistency.

docker-compose.yml
```diff
@@ -53,7 +53,7 @@ services:
         ports:
             - '${FORWARD_DB_PORT:-3306}:3306'
         volumes:
-            - 'mariadb-data:/var/lib/mysql'
+            - 'mariadb-data_prod_05192024:/var/lib/mysql'
```
Jamiras, line 56: OMG! I restarted my machine with this branch checked out and thought my DB got nuked. So I rebuilt my VM and it happened again. Then I spent like two hours trying to figure out how to get to the data that was clearly still taking up space on my HD. I finally saw the "data_prod_05192024" in one of the messages and grep'd to find this.

app/Platform/EventServiceProvider.php
```diff
@@ -101,6 +103,7 @@ class EventServiceProvider extends ServiceProvider
             DispatchUpdateGameMetricsJob::class, // dispatches GameMetricsUpdated
         ],
         PlayerMetricsUpdated::class => [
+            DispatchUpdatePlayerPointsStatsJob::class, // dispatches PlayerPointsStatsUpdated
```
Jamiras, line 106: This doesn't seem to be working for me. If I curl a dorequest to unlock an achievement, my Daily stats don't get updated. I did `sail artisan horizon:terminate` and `sail artisan horizon` to restart the service, and I see the "UpdatePlayerMetrics" job go through Horizon, but not this. When I tried resetting my VM (to ensure Horizon was restarted), my database became empty.

docker-compose.yml
wescopeland: 😓 I am so sorry for that. I really need to get a handle on gitignoring changes that don't belong so I'm not constantly committing/not committing stuff that doesn't belong.

I've fixed the branch, and I commit to you I'll get a better handle on this.

app/Platform/EventServiceProvider.php
wescopeland: Do you see a similar result when you use the admin tool to manually unlock an achievement for a user? This is what I see:

Screenshot 2024-06-03 at 8 08 55 PM

Jamiras: With latest, the player points job is getting run, and the points show up if I click the more link:
image

But they're still not present on the profile page itself:
image

I've verified the rows are not present in the database, which implies the queries for the more page are not using the denormalized data.

If I run the job manually: `sail artisan ra:platform:player:update-points-stats Jamiras`, then the rows do get created, but the player still doesn't show up in the daily/weekly tabs on the profile page.

Jamiras: It looks like it's just not including the current player on the daily/weekly tabs. If I manually unlock an achievement for a followed user, and manually run their associated update-points-stats job, _they_ appear in the daily/weekly tabs:
image
image

wescopeland: Fixed a few odds and ends in latest. I believe I hunted down why the job wasn't dispatching when doing a curl to the Connect API with `r=awardachievement`. I've also made adjustments to include the current user in the Daily and Weekly view, as well as fixed a bug with the rank # for the current user.

Screenshot 2024-06-05 at 7 57 45 PM

app/Platform/Actions/UpdatePlayerPointsStats.php
```diff
@@ -0,0 +1,149 @@
...
+        int $points,
+    ): void {
+        $attributes = ['user_id' => $user->id, 'type' => $playerStatType];
+        $existingPlayerStat = PlayerStat::where($attributes)->first();
+
+        if ($existingPlayerStat) {
+            $existingPlayerStat->update(['value' => $points]);
+        } elseif ($points !== 0) {
+            PlayerStat::create(array_merge($attributes, ['value' => $points]));
+        }
```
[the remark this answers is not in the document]
wescopeland: Yes, this was indeed my original approach. However, I found this approach would occasionally lead to race conditions where some values just were never deleted. By creating a separate action for deletion, the problem went away.

app/Platform/Jobs/UpdatePlayerPointsStatsJob.php
```diff
@@ -0,0 +1,68 @@
...
+    use Queueable;
+    use SerializesModels;
+
+    public function __construct(
+        private readonly int $userId,
+        private readonly ?string $mockCurrentDate = null, // For testing only, use with the UpdatePlayerPointsStats command.
+    ) {
+    }
+
+    public int $uniqueFor = 3600;
```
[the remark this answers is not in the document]
wescopeland: `$uniqueFor` is indeed a Laravel thing - it defines the duration for which a job's uniqueness should be maintained. The job will not be processed again within this time period, thus ensuring only one instance of the job with the same ID is processed at any given time, regardless of how many may be dispatched/queued up.

ref: https://laravel.com/docs/11.x/queues#unique-jobs

app/Platform/AppServiceProvider.php
```diff
@@ -107,6 +113,8 @@ public function boot(): void
             /** @var Schedule $schedule */
             $schedule = $this->app->make(Schedule::class);
 
+            $schedule->command(UpdatePlayerPointsStats::class, ['--existing-only'])->hourly();
+            $schedule->command(DeleteStalePlayerPointsStatsEntries::class)->hourly();
```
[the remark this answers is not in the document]
wescopeland: An approach that leans into simplicity seems best for this use case. I went with setting it to `weekly()`.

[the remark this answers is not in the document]
Jamiras: I think we'll still want to limit the query to X days in the past, or we'll still have the flush/recreate on the week boundary.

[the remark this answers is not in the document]
wescopeland: OK, I think this should now be addressed in latest.

The command now accepts an optional `days` param, which defaults to 7. The query has a `->where('created_at', '<', $dateThreshold)` added to ensure we're deleting records older than the specified number of days.

app/Platform/Commands/DeleteStalePlayerPointsStatsEntries.php
```diff
@@ -0,0 +1,36 @@
...
+            PlayerStatType::PointsHardcoreDay,
+            PlayerStatType::PointsHardcoreWeek,
+            PlayerStatType::PointsSoftcoreDay,
+            PlayerStatType::PointsSoftcoreWeek,
+            PlayerStatType::PointsWeightedDay,
+            PlayerStatType::PointsWeightedWeek,
+        ];
+        PlayerStat::whereIn('type', $relevantPlayerStatTypes)
+            ->where('value', 0)
+            ->where('created_at', '<', $dateThreshold)
```
Jamiras, line 33: 'updated_at'?

If we're recycling entries, it should be trivial to have created dates a week in the past.

wescopeland: Changed from `created_at` to `updated_at` in latest.

Jamiras (approved):

wescopeland:
Preferably would like to see this one not included in 6.7.0, as there's already quite a bit of things merged for 6.7.0 and this one should be watched closely.

If no objections, let's push this one to the release afterwards.

PR #2449, opened 2024-05-19
fix: hide random white flashing dot that appears on every page

wescopeland:
Introduced in https://github.com/RetroAchievements/RAWeb/pull/2257.

On every page, there is a tiny flash of unstyled content which appears until JavaScript is finished loading/parsing. Resolved by sticking a `x-cloak` onto the DOM node.

https://github.com/RetroAchievements/RAWeb/assets/3984985/e7c76bc5-1635-41c6-9767-a42a9821367b


resources/views/components/menu/search.blade.php
```diff
@@ -53,6 +57,7 @@ class="flex-1 searchboxinput"
         class="p-0.5 w-fit absolute top-0 left-0 rounded-lg bg-yellow-100 z-20"
         x-ref="searchListbox"
         x-show="showSearchResults"
+        x-cloak
```
wescopeland, line 60: This is the actual fix.

resources/views/layouts/app.blade.php
wescopeland, line 1: This is all just whitespace modifications.

Jamiras (approved):

PR #2455, opened 2024-05-20
feat(api): add fortify login route to retrieve api key

wescopeland:
The PR adds a new V2 web API route:

```
POST https://localhost:64000/api/v2/login
```
```
# Request Payload
{
  "username": "MyUsername"
  "password": "MyPassword"
}
```
```
# Response Payload
{
  "webApiKey": "XXXXXXXXXXXXXXXXXXXXXXXXXX"
}
```

This route will conceivably allow emulation front-ends to exchange user credentials for a user's web API key, ideally to ween them off the Connect API (for which they're currently exchanging credentials for a connect token).

**Security Considerations:**
* Error messages are intentionally obtuse.
* The route uses Fortify, not `authenticateFromPassword()`.
* The route uses the same rate limiter as the site's login page.

### DIFF
--- a/app/Api/Controllers/WebApiController.php
+++ b/app/Api/Controllers/WebApiController.php
@@ -37,7 +37,7 @@ class WebApiController extends Controller
         ], 501);
     }
 
-    public function requestWebApiKey(Request $request): JsonResponse
+    public function login(Request $request): JsonResponse
     {
         $request->validate([
             'username' => 'required|string',
--- a/app/Api/Controllers/WebApiController.php
+++ b/app/Api/Controllers/WebApiController.php
@@ -7,6 +7,8 @@ namespace App\Api\Controllers;
 use App\Http\Controller;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+use Laravel\Fortify\Fortify;
 
 class WebApiController extends Controller
 {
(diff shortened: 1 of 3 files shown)


wescopeland:
Documented at https://github.com/RetroAchievements/api-docs/pull/46.

app/Api/Controllers/WebApiController.php
```diff
@@ -34,4 +36,25 @@ public function users(Request $request): JsonResponse
...
+            'password' => $request->input('password'),
+        ];
+
+        if (Auth::attempt($credentials)) {
+            $user = Auth::user();
+
+            return response()->json(['webApiKey' => $user->api_token]);
+        }
+
+        return response()->json(['error' => 'Something went wrong'], 400);
```
Jamiras, line 58: I understand the necessity to not reveal too much, but would it be correct to return "Invalid user/password" here [like we do for connect](https://github.com/RetroAchievements/RAWeb/blob/master/tests/Feature/Connect/LoginTest.php#L100-L204)?

app/Api/Controllers/WebApiController.php
```diff
@@ -34,4 +36,25 @@ public function users(Request $request): JsonResponse
             'data' => $request->input(),
         ], 501);
     }
+
+    public function login(Request $request): JsonResponse
```
Jamiras, line 40: Is it possible to add unit tests for this? I would really like everything in the V2 API to have good test coverage.

app/Api/Controllers/WebApiController.php
Jamiras: I know @luchaos wants to move to a response structure more in line with the [official JSON API docs](https://jsonapi.org/format/#error-objects): https://discord.com/channels/476211979464343552/757767535293890682/1132370669938671758.

I think a response formatted in that manner would look like:
```
{
    "errors": [
        {
            "status": "401",
            "code": "invalid_credentials",
            "title": "Invalid username/password",
        }
    ]
}
```

I would really like to get his feedback on anything we're going to call V2 of the API.

Jamiras: Connect also has a secondary "access denied" error if the credentials are valid, but the account's email address hasn't yet been verified.

wescopeland: An alternative to putting this in the V2 API namespace might be to return something like `APIKey` from `r=login2`. I would be happy to put this in the V1 Web API, but it's designed to only accept unauthenticated GET requests.

app/Api/Controllers/WebApiController.php
wescopeland: Yes - certainly. Pending if we decide to keep this in here, I'm happy to add code coverage.

PR #2456, opened 2024-05-21
feat(ticket): create and use `ticketable_author_id` column

wescopeland:
Implements [the following requirements](https://discord.com/channels/476211979464343552/1241906810689028137/1242234958693666886):
* When a ticket is opened, the owner will be whoever the author of the ticketable entity is at that moment in time.
* That owner is permanent, even when reauthorship occurs. This prevents reauthorship from affecting dev stats.
* Every ticket must have an owner.

This PR introduces a migration to add a `ticketable_author_id` column to the `Ticket` table. All queries to determine author / dev stats are now pointed at this value rather than the achievement `user_id`.

After running the migration, execute the following SQL query:
```sql
UPDATE Ticket
JOIN Achievements ON Ticket.AchievementID = Achievements.ID
SET Ticket.ticketable_author_id = Achievements.user_id;
```

Approximately 6 rows will be left that don't have a `ticketable_author_id`. Optionally, we can assign these to the System user:
```sql
UPDATE Ticket
SET Ticket.ticketable_author_id = 14188
WHERE Ticket.ticketable_author_id IS NULL;
```


app/Helpers/database/ticket.php
```diff
@@ -302,40 +320,33 @@ function countRequestTicketsByUser(?User $user = null): int
 
 function countOpenTicketsByDev(User $dev): array
 {
-    $cacheKey = CacheKey::buildUserOpenTicketsCacheKey($dev->User);
-
-    return Cache::remember($cacheKey, Carbon::now()->addHours(20), function () use ($dev) {
```
wescopeland, line 307: This query is so fast (~3ms worst case I observed) it doesn't seem like there's any point in caching it.

Jamiras: It definitely mattered six months ago (see #1928). It looks like the only thing in the [eloquent](https://github.com/RetroAchievements/RAWeb/pull/1928/files#diff-33c2242e14e47d19d23e3afcebf5459b73b9ed4d1f0f07c0ce57605d9df67d77L434-L443) that changed is the filter on `achievement.Author` -> `achievement->user_id`. Maybe there was an index added?

I do still see the query being duplicated, but there's only one query, and it takes ~0.5ms.

app/Helpers/database/ticket.php
```diff
@@ -88,76 +89,93 @@ function submitNewTicket(User $user, int $achID, int $reportType, int $hardcore,
...
-    $ticketUrl = route('ticket.show', ['ticket' => $ticketID]);
     $bugReportDetails = "
-Achievement: $achTitle
-Game: $gameTitle
-Problem: $problemTypeStr
-Comment: $note
+Achievement: {$achievement->title}
+Game: {$game->title}
+Problem: {$problemTypeStr}
+Comment: {$ticket->body}
```
Jamiras, line 104: I tested creating a new ticket and the Comment did not show up in the email.

> Jamiras would like to report a bug with an achievement you've created:
>
> Achievement: Who Let the Dogs Out
> Game: Legend of Zelda, The: Ocarina of Time
> Problem: Doesn't trigger
> Comment:
>
> This ticket will be raised and will be available for all developers to inspect and manage at the following URL:
http://localhost:64000/ticket/72901

app/Helpers/database/ticket.php
wescopeland: > Maybe there was an index added?

Digging a little deeper into this, `Author` had an index too. However, I believe the perf discrepancy comes down to the fact that `Author` wasn't a foreign key, whereas `user_id` is. When the DB engine knows about a foreign key relationship, the query optimizer can more efficiently execute the join and oftentimes avoid a full table scan.

app/Helpers/database/ticket.php
wescopeland: Good catch - looks like I missed checking in a model update. Updated in latest.

PR #2459, opened 2024-05-21
refactor(LeaderboardDef): drop `Author`

wescopeland:
Removes the last `Author` reads from `LeaderboardDef` and drops the column. Additionally, eliminates some dead legacy code.


app/Helpers/database/leaderboard.php
```diff
@@ -537,84 +539,28 @@ function UploadNewLeaderboard(
...
         $errorOut = "Internal error updating leaderboard.";
 
         return false;
     }
 
-    if ($originalAuthor != '') {
+    if (isset($originalAuthor)) {
         addArticleComment("Server", ArticleType::Leaderboard, $idInOut,
-            "$author edited this leaderboard.", $author
+            "{$authorModel->display_name} edited this leaderboard.", $authorModel->username
```
Jamiras, line 557: Not sure if it matters, but edits from Connect use this message
image

And edits from the webpage use this one
image

Jamiras (approved):

app/Helpers/database/leaderboard.php
wescopeland: Probably not a huge deal if our long-term goal is to move over to the activitylog, but no harm in getting these consistent. Changed the webpage edits to match "edited this leaderboard".

PR #2461, opened 2024-05-22
refactor(SiteAwards): drop `User`

wescopeland:
Drops `User` from the `SiteAwards` table.

This is the final PR for dropping user / author columns and switching everything over to rely on IDs. After this and the other PRs are merged, no currently-used table will reference users by username anymore.


app/Platform/Services/DeveloperFeedService.php
```diff
@@ -106,7 +106,7 @@ private function attachRecentAwardRowsMetadata(mixed $awardRows): mixed
             $row->GameTitle = $game->Title ?? null;
             $row->GameIcon = $game->ImageIcon ?? null;
 
-            $row->Untracked = $userData[$row->User]->Untracked ?? null;
+            $row->Untracked = $userData[$row->user->User]->Untracked ?? null;
```
Jamiras, line 109: You should be able to eliminate the `$userData` query on line 92. On line 91 just `loadMissing('user')` and change this to `$row->user->Untracked ?? null`.

database/migrations/platform/2024_05_21_000000_update_siteawards_table.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('SiteAwards', function (Blueprint $table) {
+            $table->dropColumn('User');
+        });
```
Jamiras, line 14: I'm confused:
```
SQLSTATE[42000]: Syntax error or access violation: 1072 Key column 'User' doesn't exist in table (Connection: mysql, SQL: alter table `SiteAwards` drop `User`)
```
```
MariaDB [retroachievements-web]> desc SiteAwards;
+----------------+---------------------+------+-----+---------+----------------+
| Field          | Type                | Null | Key | Default | Extra          |
+----------------+---------------------+------+-----+---------+----------------+
| id             | bigint(20) unsigned | NO   | PRI | NULL    | auto_increment |
| AwardDate      | datetime            | NO   |     | NULL    |                |
| User           | varchar(50)         | NO   | MUL | NULL    |                |

```

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -41,7 +41,7 @@
 $userModel = null;
 $defaultSort = 1;
 if (isset($user)) {
-    $userModel = User::firstWhere('User', $user);
+    $userModel = User::with('playerBadges')->find($userID);
```
Jamiras, line 44: Does this fetch all of the user's badges? Shouldn't `getUserGameProgressionAwards` just fetch the singular badge associated to the user/game combination?

app/Platform/Services/DeveloperFeedService.php
wescopeland: Fixed in latest

database/migrations/platform/2024_05_21_000000_update_siteawards_table.php
wescopeland: This came down to the unique constraint we used to have before the table actually had a primary key. Fixed in latest.

resources/views/pages-legacy/gameInfo.blade.php
wescopeland: You're right - this is not optimal. Adjusted in latest so `getUserGameProgressionAwards` only targets the given game ID.

Jamiras (approved):

database/migrations/platform/2024_05_21_000000_update_siteawards_table.php
```diff
@@ -0,0 +1,35 @@
...
+            $table->index(['user_id', 'AwardData', 'AwardType', 'AwardDataExtra']);
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('SiteAwards', function (Blueprint $table) {
+            $table->dropIndex(['user_id', 'AwardData', 'AwardType', 'AwardDataExtra']);
+            $table->string('User', 50)->after('AwardDate');
+            $table->unique(['User', 'AwardData', 'AwardType', 'AwardDataExtra'], 'siteawards_user_awarddata_awardtype_awarddataextra_unique');
```
Jamiras, line 31: This is failing for me. Presumably because User is null for every record.

wescopeland: This probably can't be resolved in the migration itself and requires manual intervention.

```sql
UPDATE SiteAwards sa
JOIN UserAccounts ua ON sa.user_id = ua.ID
SET sa.User = ua.User;
```

Jamiras: Agreed, but having an exception is bad. Recommend just not recreating the index.

wescopeland: Got it - I wasn't sure if that was a viable approach. Updated in latest to not recreate the index.

PR #2463, opened 2024-05-23
refactor: migrate random set page to folio

wescopeland:
Migrates `/random.php` to `/game/random`.

Inlines the helper function to retrieve a random game with achievements, and leverages denormalized game metrics rather than making a join on the Achievements table (which was not possible when the helper was written).


resources/views/pages/game/random.blade.php
```diff
@@ -0,0 +1,39 @@
...
+// With 25k rows in the GameData table, and 6k games with achievements, the chance of any
+// individual query failing is roughly 75%. The chance of three queries in a row failing is
+// 42%. At ten queries, the chance is way down at 6%, and we're still 40+ times faster than
+// the alternate solutions.
+$getRandomGameWithAchievements = function(): Game {
+    $maxId = Game::max('ID');
+
+    do {
+        $gameId = random_int(1, $maxId);
+        $gameWithAchievements = Game::where('ID', $gameId)
```
Jamiras, line 26: Since we're leveraging the denormalized data now, can't the loop be eliminated in favor of:
```
$gameWithAchievements = Game::inRandomOrder()->where...
```

resources/views/pages/game/random.blade.php
```diff
@@ -0,0 +1,39 @@
...
+            ->where('ConsoleID', '<', 100)
```
Jamiras, line 27: suggest:
```
->whereNotIn('ConsoleID', System::getNonGameSystems())
```

resources/views/pages/game/random.blade.php
wescopeland: Good idea - that's much simpler. Done in latest.

resources/views/pages/game/random.blade.php
wescopeland: Much better - suggestion applied in latest. Thanks.

resources/views/pages/game/random.blade.php
Jamiras: doesn't look like this got pushed.

resources/views/pages/game/random.blade.php
Jamiras: doesn't look like this got pushed.

resources/views/pages/game/random.blade.php
Jamiras: You need to add back in a clause to only return games with achievements.

wescopeland: Whoops - thanks. Updated in latest.

resources/views/pages/game/random.blade.php
wescopeland: Should be fixed now. The query now reads:
```php
$randomGameWithAchievements = Game::whereNotIn('ConsoleID', System::getNonGameSystems())
    ->where('achievements_published', '>=', 6)
    ->inRandomOrder()
    ->firstOrFail();
```

Jamiras (approved):

Issue #2399, opened 2024-05-02
Wrong softcore users count in Achievement Distribution graph


Mindhral:
**Describe the bug**
In the Achievement Distribution graph on a game's page, the bars give the hardcore players count in yellow and the softcore players count in gray for each possible number of achievements (or interval).
It works as long as there are only pure hardcore and pure softcore players, but the softcore user count is wrong if there is at least one player with unlocks in both modes.

**To Reproduce**
As this behavior is conditioned by changing data, it's possible my example becomes obsolete, but it shouldn't happen too soon.

1. Go to https://retroachievements.org/game/23766
2. At the time this ticket is made, the fastest way to see there is a problem is to look at the hidden table bellow the graph (CSS selector "#chart_distribution table")
3. The data for 16 achievements shows 1 hardcore player and -1 softcore player, which is obviously wrong.

**Expected behavior**
Either real softcore players counts, or (better in my opinion) switching the second bar to "Total players" (or a better wording).

**Screenshots**
![image](https://github.com/RetroAchievements/RAWeb/assets/148589639/86c0d496-1952-4d8c-986b-825e9df74652)

![image](https://github.com/RetroAchievements/RAWeb/assets/148589639/9dcdb388-3aeb-4817-bdcc-70d03cb8119f)

**Additional context**
The data returned by the DB query is the number of hardcore players and the total number of players for each number of achievements. The code in https://github.com/RetroAchievements/RAWeb/blob/b0ce38383b62ad04c629654958b9a1e0fb13a163/app/Helpers/render/game.php#L354 calculates the number of softcore players as the subtraction of the 2 numbers.

But a player who has mixed unlocks doesn't necessarily count in the same bucket for hardcore and total, so you can't remove them from one to get a complement of the other (I know it's not very clear, sorry).
Generally speaking, we can't determine the number of softcore players with the data returned by the DB query (I can give a simple example with 3 players and 2 achievements where there can be 2 possible situations leading to the same query result).

Event if the DB result was changed, hardcore players and softcore players should not be drawn the way they are in my opinion, as their addition doesn't mean anything when there are players with mixed unlocks (imagine a player with 2 unlocks in hardcore and 2 unlocks in softcore). Which is why I suggest to change the label and number shown, to reflect exactly what is returned by the query: hardcore and total.

TwosomesUP:
[...]

PR #2466, opened 2024-05-24
fix(game): show correct softcore users count in Achievement Distribution graph

TwosomesUP:
Initial commit and composer fix
Update achievement distribution selections and migration to Eloquent
Update calculateBuckets and handleAllAchievementsCase to accommodate selection update
Update V1Test to accommodate selection update

Resolves: #2399


TwosomesUP:
I made some minor optimizations to the queries, and they seem to be executing faster, but I'll need to set up some volume testing on my end to be sure.

image

wescopeland:
Thanks for taking a look. Just checked on my end with latest from the branch, seems about the same 😟

Screenshot 2024-05-24 at 4 53 35 PM

wescopeland:
Sorry, same problem on latest 😔

**Branch**
Screenshot 2024-05-24 at 8 24 22 PM


**master**
Screenshot 2024-05-24 at 8 24 45 PM

TwosomesUP:
No worries! I'm going to break this off into another branch to play around with. Hopefully I can figure something out that's more performant. 😁

TwosomesUP:
Ok, I was able to insert about 25k rows into player_games for game_id 1 with a variety of unlock combinations, and I'm definitely seeing the performance hit now!

image

image

This should allow me to run better unit tests with this selection. 😄

TwosomesUP:
Hey @wescopeland !

I did some extensive testing/tinkering with the Achievement Distribution query and as the data is currently structured, I'm not sure we'd be able to get the selection much faster than it currently is. The fastest I could get it was 50ms for 25k rows on the `player_games` table.

That said, there is a different approach we could take to speed this up, but it would require altering the `player_games` table and adding a new stored Generated Column for softcore unlocks, and an index. This has a much bigger impact, so I'll leave it up to RA if you all would like to consider this route, but here are my results from local testing:

Legacy
Legacy

PR Aggregate (Eloquent)
Aggregate

Alter + Index (Eloquent)
WithAlterAndIndex_Eloquent

Alter + Index (DB::select)
WithAlterAndIndex_DBSelect

I'm not sure if there is a configuration difference with my instance of MySQL but the Alter + Index seems to have equivalent performance compared to the legacy query. For your reference, here is the SQL I used for the alter and the index build:

```
alter table player_games
add column achievements_unlocked_softcore int unsigned generated always as (greatest(achievements_unlocked - achievements_unlocked_hardcore, 0)) stored;

alter table player_games add index player_games_game_id_achievements_unlocked_softcore_index (game_id, achievements_unlocked_softcore)
```

Again, this can have a big impact since it alters a table with potentially a million+ rows in production, and it would calculate the softcore unlocks for each one of those (one time). Once altered, the column would be calculated for a row upon insert or recalculated upon update. 

With that in mind, it's completely understandable if this is something RA would not want to pursue at the moment.

wescopeland:
I had a feeling needing a new denormalized column would be how things might shake out 😆

That said, I would be fine with pursuing this. Before moving forward, I think another maintainer should also sign off on the idea though (cc @RetroAchievements/web-maintainers). I can think back to other times where I've needed to perform a calculation to peek into a user's mixed progress. Just having the raw value would've been more helpful and performant.

Going down this road is a more significant lift and is much more fraught with peril. It would need to occur in two phases (releases): a write phase and a read phase.

For the write phase, we would need a PR/changeset which includes:
* The migration itself, in _database/migrations/platform_.
* Changes made to the _UpdatePlayerGameMetrics_ action so these new values get populated.
* Updated tests, etc.

I personally haven't triggered a full resync of player game metrics on the server, but I know we've done it once or twice in the past when we first began experimenting with the performance benefits of denormalized data. Once the code is deployed, we'd trigger a full resync, which would queue up probably about 1,000,000 jobs. This would take quite some time to finish.

For the read phase, it'd essentially be a PR containing what you have now, but instead reading from these new denormalized values.

TwosomesUP:
Great! I can start researching the write phase steps this week. I may reach out for guidance on a few things if I hit a wall, but I'll be sure to let you know.

TwosomesUP:
@wescopeland a PR for the Write Phase (#2477) has been submitted. If/when the new PR is approved and merged, I can update this PR to accommodate the changes. Just let me know if there's anything I may have missed, or if you have any questions/concerns.

Thanks!

TwosomesUP:
@wescopeland circling back to this and after some testing, it seems like the automatic "Soft Delete" check when selecting from the `player_games` table using Eloquent and the PlayerGame model is impacting performance. Specifically when the `deleted_at` column is included in the selection. Here are the timings for selections with and without the soft delete check:

### **Eloquent with Soft Delete check:**
```
$countQuery = PlayerGame::selectRaw("$countColumn as AwardedCount, count(*) as NumUniquePlayers")
            ->where("$countColumn", ">", 0)
            ->where('player_games.game_id', $gameID)
            ->groupBy('AwardedCount')
            ->orderByDesc('AwardedCount');
```
image

### **Eloquent without Soft Delete check (code same as above):**
image
*This one removes the `use SoftDeletes` from the PlayerGame model. Probably don't want to do this. :)

### **DB::table() select without Soft Delete check:**
```
$countQuery = DB::table('player_games')
            ->selectRaw("$countColumn as AwardedCount, count(*) as NumUniquePlayers")
            ->where("$countColumn", ">", 0)
            ->where('player_games.game_id', $gameID)
            ->groupBy('AwardedCount')
            ->orderByDesc('AwardedCount');
```
image
*This one bypasses the PlayerGame model and does a direct DB select

wescopeland:
@TwosomesUP Sorry for my late reply here. I'm in the process of setting up some production data to test with, which should be done later today. As soon as that is set up, getting you a review will be my highest priority 👍

TwosomesUP:
I appreciate it Wes, but no rush! I still need to update this pull request with the new selects that account for the new `achievements_unlocked_softcore` column. 

I've been on vacation, so I haven't had a chance to get to it yet, but I'll try to get that pushed either today or tomorrow.

app/Helpers/database/player-achievement.php
```diff
@@ -279,74 +279,82 @@ function getUnlocksInDateRange(array $achievementIDs, string $startTime, string
...
+
+        $countQuery = DB::table("player_games")
+            ->selectRaw("$countColumn as AwardedCount, count(*) as NumUniquePlayers")
+            ->where("$countColumn", ">", 0)
+            ->where('player_games.game_id', $gameID)
+            ->groupBy('AwardedCount')
+            ->orderByDesc('AwardedCount');
+
+        if ($joinUsers) {
+            $countQuery->join("useraccounts", "player_games.user_id", "=", "useraccounts.ID")
```
wescopeland, line 312: nit: All `useraccounts` and `achievements` direct references should be changed to the table names `UserAccounts` and `Achievements`.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -33,6 +33,10 @@
 $officialFlag = AchievementFlag::OfficialCore;
 $unofficialFlag = AchievementFlag::Unofficial;
 $flagParam = requestInputSanitized('f', $officialFlag, 'integer');
+
+//for testing
+//$flagParam = requestInputSanitized('f', $unofficialFlag, 'integer');
+
```
wescopeland, line 39: This should be removed.

app/Helpers/database/player-achievement.php
```diff
@@ -279,74 +279,82 @@ function getUnlocksInDateRange(array $achievementIDs, string $startTime, string
...
-        $query = "SELECT pg.$countColumn AS AwardedCount, COUNT(*) AS NumUniquePlayers
-                    FROM player_games AS pg
-                    $joinStatement
-                    WHERE pg.game_id = :gameId AND pg.$countColumn > 0
-                    $requestedByStatement
-                    GROUP BY AwardedCount
-                    ORDER BY AwardedCount DESC";
+        $countColumn = $isHardcore ? 'player_games.achievements_unlocked_hardcore' : 'player_games.achievements_unlocked_softcore';
+
+        $countQuery = DB::table("player_games")
```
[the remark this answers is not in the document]
TwosomesUP: I mentioned it earlier in the thread, but I found that using the PlayerGame model added an implicit check for soft deletes that reduced performance. Directly querying the table seemed to fix that:

https://github.com/RetroAchievements/RAWeb/pull/2466#issuecomment-2241292804

Just let me know if there might be a better approach!

app/Helpers/database/player-achievement.php
TwosomesUP: I'm not sure I understand. Do you mean spelling the table names using capitals (e.g. spelling it `Achievements` instead of `achievements`)?

wescopeland: Yes - just to match the existing table names in the database. This may help with a case-sensitive search all across the project when we later rename the tables.

app/Helpers/database/player-achievement.php
```diff
@@ -279,74 +279,82 @@ function getUnlocksInDateRange(array $achievementIDs, string $startTime, string
...
-        $joinStatement = 'INNER JOIN UserAccounts AS ua ON ua.ID = pg.user_id';
-        $joinStatementUnofficial = 'INNER JOIN UserAccounts AS ua ON ua.ID = pa.user_id';
-        $requestedByStatement = 'AND (ua.Untracked = 0';
-        if ($requestedBy) {
-            $bindings['requestedBy'] = $requestedBy;
-            $requestedByStatement .= ' OR ua.User = :requestedBy';
-        }
-        $requestedByStatement .= ')';
-    }
+    $joinUsers = $numPlayers < 100;
```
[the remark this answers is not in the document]
TwosomesUP: Updated to `$shouldJoinUsers`

app/Helpers/database/player-achievement.php
TwosomesUP: Table names updated to use the proper case

resources/views/pages-legacy/gameInfo.blade.php
TwosomesUP: Comment removed

wescopeland (approved):

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -2,9 +2,7 @@
 
 use App\Community\Enums\ArticleType;
 use App\Community\Enums\ClaimSetType;
-use App\Community\Enums\ClaimStatus;
 use App\Community\Enums\ClaimType;
-use App\Community\Enums\SubscriptionSubjectType;
```
Jamiras, line 7: This still seems to be needed:
```
Class "SubscriptionSubjectType" not found
```

Jamiras: Same for ClaimStatus:
```
Class "ClaimStatus" not found
```

TwosomesUP: Thanks for catching that! Looks like pint was removing those on me. Added back.

wescopeland:
Hi @TwosomesUP! Looks like we may have a small merge conflict. Then, assuming @Jamiras approves, we can merge to get this in the next release.

Jamiras (approved):

PR #2469, opened 2024-05-26
feat(manage): add Game resource to filament

wescopeland:
This PR adds a basic `GameResource` to the Filament management panel. Games are an essential atom of the platform, so the sooner we get this resource up and going the easier it will be to migrate management utilities over to Filament & the perms matrix.

For now, only three fields are editable. This is mostly just foundational work for the game resource.

This moves us closer to being able to eliminate the games devbox altogether, which will be a great help when doing a Folio migration of the game page.

**Index**
Screenshot 2024-05-26 at 11 51 54 AM

**Details**
Screenshot 2024-05-26 at 11 52 13 AM

**Edit**
Screenshot 2024-05-26 at 11 52 29 AM


app/Models/Game.php
```diff
@@ -99,12 +104,21 @@ protected static function newFactory(): GameFactory
...
         return LogOptions::defaults()
-            ->logOnlyDirty();
+            ->logOnly([
+                'Title',
+                'ForumTopicID',
+                'GuideURL',
+                'Publisher',
+                'Developer',
+                'Genre',
+            ])
```
wescopeland, line 119: In a follow-up PR, I'll need to update the legacy .php endpoint functions to correctly record the causer for these.

app/Policies/AchievementPolicy.php
```diff
@@ -115,10 +114,10 @@ public function forceDelete(User $user, Achievement $achievement): bool
...
     {
         $roleFieldPermissions = [
-            Role::DEVELOPER_JUNIOR => ['title', 'description', 'flags', 'type', 'points', 'display_order'],
-            Role::DEVELOPER => ['title', 'description', 'flags', 'type', 'points', 'display_order'],
-            Role::DEVELOPER_STAFF => ['title', 'description', 'flags', 'type', 'points', 'display_order'],
-            Role::WRITER => ['title', 'description'],
+            Role::DEVELOPER_JUNIOR => ['Title', 'Description', 'Flags', 'type', 'Points', 'DisplayOrder'],
+            Role::DEVELOPER => ['Title', 'Description', 'Flags', 'type', 'Points', 'DisplayOrder'],
+            Role::DEVELOPER_STAFF => ['Title', 'Description', 'Flags', 'type', 'Points', 'DisplayOrder'],
+            Role::WRITER => ['Title', 'Description'],
```
wescopeland, line 120: To improve DX, these all now map 1:1 to the column names in the DB.

package.json
```diff
@@ -44,6 +44,7 @@
         "laravel-echo": "^1.15.1",
         "laravel-vite-plugin": "^0.8.0",
         "postcss": "^8.4.31",
+        "postcss-nesting": "^12.1.5",
```
wescopeland, line 47: This is a Filament dependency.

app/Filament/Resources/GameResource.php
```diff
@@ -0,0 +1,340 @@
...
+
+                        Infolists\Components\TextEntry::make('title'),
+
+                        Infolists\Components\TextEntry::make('forumTopic.id')
+                            ->label('Forum Topic')
+                            ->url(fn (?int $state) => url("viewtopic.php?t={$state}")),
+
+                        Infolists\Components\TextEntry::make('system')
+                            ->formatStateUsing(fn (System $state) => "[{$state->id}] {$state->name}")
+                            ->url(fn (System $state) => SystemResource::getUrl('view', ['record' => $state->id])),
```
Jamiras, line 87: This link gets generated even if I don't have access.
image

app/Filament/Resources/GameResource.php
```diff
@@ -0,0 +1,340 @@
...
+    public static function infolist(Infolist $infolist): Infolist
+    {
+        return $infolist
+            ->schema([
+                Infolists\Components\ImageEntry::make('badge_url')
+                    ->label('')
+                    ->size(config('media.icon.lg.width')),
+
+                Infolists\Components\Section::make('Metadata')
+                    ->columns(['md' => 2, 'xl' => 3, '2xl' => 4])
```
Jamiras, line 74: Not sure if there's an easy way to do it, but published achievements is arguably more important that ForumTopicId for the fourth column.

app/Filament/Resources/GameResource.php
wescopeland: Good catch - I didn't think of that. Fixed in latest.

app/Filament/Resources/GameResource.php
wescopeland: Very easy - done in latest.

app/Filament/Resources/GameResource.php
Jamiras: The link still exists in the View page:
image

app/Filament/Resources/GameResource.php
Jamiras: Looks like the column disappeared entirely?
image

wescopeland: Hrm, this is what I see:

Screenshot 2024-06-07 at 8 28 27 PM

Does anything change if you do `sail artisan view:clear`?

app/Filament/Resources/GameResource.php
wescopeland: Forgot about that one. Updated in latest.

app/Filament/Resources/GameResource.php
Jamiras: Seems to have been a cookie or something. I logged in incognito mode and it was present, so I logged out and logged back in normal mode and it's present there too now.

PR #2470, opened 2024-05-26
feat(user): attach/detach user roles on Permissions update

wescopeland:
This PR makes an update to the legacy .php endpoint responsible for updating a user's `Permissions` value. Now when the value is set, the user's attached roles are updated accordingly. This is basically a form of double-writes so we can ultimately move towards dropping the `Permissions` field from the table.

Eventually this endpoint will go away altogether, with this sort of user management stuff happening in the Filament management app. 

We'll need to execute some SQL queries to get the `auth_model_roles` table current:

```sql
-- attach all junior developer roles
INSERT INTO auth_model_roles (role_id, model_type, model_id)
SELECT 7, 'user', ID
FROM UserAccounts
WHERE Permissions = 2;

-- attach all developer roles
INSERT INTO auth_model_roles (role_id, model_type, model_id)
SELECT 6, 'user', ID
FROM UserAccounts
WHERE Permissions = 3;
```

### DIFF
--- a/public/request/user/update.php
+++ b/public/request/user/update.php
@@ -3,6 +3,7 @@
 use App\Community\Enums\ArticleType;
 use App\Community\Enums\UserAction;
 use App\Enums\Permissions;
+use App\Models\Role;
 use App\Models\User;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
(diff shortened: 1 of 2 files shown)


public/request/user/update.php
```diff
@@ -28,13 +29,36 @@
...
+        }
+        if ($value === Permissions::Developer) {
+            $foundTargetUser->removeRole(Role::DEVELOPER_JUNIOR);
+            $foundTargetUser->removeRole(Role::MODERATOR);
+
+            $foundTargetUser->assignRole(Role::DEVELOPER);
+        }
+        if ($value === Permissions::Moderator) {
+            $foundTargetUser->assignRole(Role::MODERATOR);
+        }
```
Jamiras, line 61: This code doesn't revoke the Moderator role when a moderator is demoted, but it's currently impossible to promote or demote a moderator from the UI as a user's current rank and new rank must both be below the rank of the user making the change.

public/request/user/update.php
```diff
@@ -28,13 +29,36 @@
...
+        // Auto-apply forums permissions.
+        if ($value >= Permissions::JuniorDeveloper && !$foundTargetUser->ManuallyVerified) {
+            setAccountForumPostAuth($foundSourceUser, $permissions, $foundTargetUser, authorize: true);
+        }
+
+        // Adjust attached roles.
+        if ($value <= Permissions::Unregistered) {
+            $foundTargetUser->roles()->detach();
+        }
+        if ($value === Permissions::Registered) {
```
Jamiras, line 42: These should be elseifs or a switch statement. None of the cases overlap.

public/request/user/update.php
wescopeland: I had forgotten about that restriction. Given that this legacy endpoint is tightly coupled to said UI, it seems like what may be best is to just exclude a condition for the `Moderator` permission altogether.

public/request/user/update.php
wescopeland: Went with elseifs in latest, as the `<=` case is not as clean to handle with a switch.

Jamiras: I was thinking the `<=` could be the default cause, but I can see why that wouldn't necessarily be desired.

Jamiras (approved):

wescopeland:
I executed the SQL commands on stage. Will still need to be run on prod.

Issue #2471, opened 2024-05-26
Wiping leaderboard entries should soft delete them


wescopeland:
Currently, when deleting a `LeaderboardEntry` record, the value is hard deleted from the database. The model uses soft deletes, but we aren't currently utilizing this after the cutover from `LeaderboardEntryLegacy`.

We should swap over to using soft deletes, while also ensuring any queries pulling from `leaderboard_entries` ignore those trashed records.

PR #2472, opened 2024-05-26
fix: hard delete leaderboard entries on leaderboard reset

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1244081094983159961.

Related: https://github.com/RetroAchievements/RAWeb/issues/2471


wescopeland:
Closed after rethinking this - we should just resolve the root cause rather than continuing to band-aid this.

PR #2474, opened 2024-05-28
fix(auth): remediate auto-login failure on password reset

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1243907569299361863/1245018616596791366.

Currently, we're attempting to auto-login users on password reset, and this is virtually guaranteed to fail because login is not leveraging Fortify.

Now, on password reset, users will correctly be auto-logged in via `Auth::login()`, and they'll be redirected to the home page with a success message.


public/request/auth/reset-password.php
```diff
@@ -9,19 +11,17 @@
...
-// Perform auto-login:
-if (authenticateFromCookie($user, $permissions, $userDetails)) {
-    generateAppToken($user, $tokenInOut);
-}
+// Change the user's password and automatically log them in.
+changePassword($targetUser->username, $newPass);
+Auth::login($targetUser);
 
-return back()->with('success', __('legacy.success.password_change'));
+return redirect()->route('home')->with('success', __('legacy.success.password_change'));
```
Jamiras, line 27: Really not sure what's going on, or if it's even worth investigating, but if you cause two validation errors when trying to change your password, a successful change will redirect to the login page instead of the home page.

1. Click link from reset password email
2. Click "Change Password" - get Password required error
3. Click "Change Password" - get Password required error
4. Enter a valid password twice. Click "Change Password" - get redirected to login page.

This also happens if you enter two passwords that don't match twice, or two blanks and two mismatches. It does not occur if you only generate a single validation error.

Jamiras (approved):

public/request/auth/reset-password.php
wescopeland: I appreciate your thoroughness, as always. Weird, I can't seem to repro this issue. Wonder if it's env related? Tried a few different things but can't seem to get it to surface:

https://github.com/RetroAchievements/RAWeb/assets/3984985/1bbf19dd-2298-4295-851e-9708699ddbdd


https://github.com/RetroAchievements/RAWeb/assets/3984985/9d9206b5-342a-48f3-9fba-816753866315

Jamiras: Seems to be working fine for me now too.

PR #2477, opened 2024-05-30
Player Games Database and Metrics Updates

TwosomesUP:
This update includes a DB migration that adds a new column to the `player_games` tabled named achievements_unlocked_softcore. This column will be used to track a player's softcore unlocks for a game, which will eventually be utilized by the Achievement Distribution graph to more accurately represent hardcore vs. softcore unlock distribution.

The PR also includes changes to the UpdatePlayerGameMetrics action in order to populate this new column, as well as the UpdatePlayerGameMetrics command to allow for batch processing of the `player_games` table. In order to update the softcore unlocks for all players, the following command would need to be run with no parameters:

`php artisan ra:platform:player:update-game-metrics`

Due to the volume of rows in the `player_games` table in production, this could take a while to run. Running it against 25000 rows locally took about 10 minutes.

Finally, a new Platform test has been added to validate the changes made related to this PR.

If this PR is approved and merged, #2466 will need to be updated to accommodate the changes to the `player_games` table.


Jamiras:
Why is this necessary? `player_games` already has `achievements_unlocked` and `achievements_unlocked_hardcore`. To get the softcore count just subtract the two:
```
SELECT achievements_unlocked, 
       achievements_unlocked_hardcore,
       achievements_unlocked - achievements_unlocked_hardcore AS achievements_unlocked_softcore
  FROM player_games
   ...
```

TwosomesUP:
I agree, subtracting them in a SQL select should be all that is needed, and was what I attempted in my original PR (#2466) but there were performance concerns. Especially once the player count for a game reaches large numbers.

This update would make the SQL used for the distribution selection equivalently performant to the existing selection.

database/migrations/platform/2024_05_25_000001_update_player_games_table.php
```diff
@@ -0,0 +1,26 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+            $table->unsignedInteger('achievements_unlocked_softcore')
+                ->nullable();
```
Jamiras, line 14: please add an `->after('achievements_unlocked_hardcore')` to put the column near the related existing columns.

app/Platform/Commands/UpdatePlayerGameMetrics.php
```diff
@@ -28,40 +31,63 @@ public function handle(): void
...
+            $query->chunkById(100, function ($rows) use ($progressBar) {
+                $jobs = $rows->map(fn ($row) => (new UpdatePlayerGameMetricsJob($row->user_id, $row->game_id))->onQueue('player-game-metrics')
+                )->all();
+
+                Bus::batch($jobs)->onQueue('player-game-metrics')->dispatch();
 
-        if ($isSync) {
-            $this->updatePlayerMetrics->execute($user);
+                $progressBar->advance(count($rows));
+            });
```
Jamiras, line 90: This took about an hour on my local machine, and it's probably not something we want to do beyond the initial seeding.

We could just seed with a query (eliminating this code), but even that took 7+ minutes, and I don't know if a single 7 minute query is better for the server than several million individual queries.
```
> UPDATE player_games SET achievements_unlocked_softcore = achievements_unlocked - achievements_unlocked_hardcore;
Query OK, 10914277 rows affected (7 min 6.703 sec)
Rows matched: 11641973  Changed: 10914277  Warnings: 0
```

tests/Feature/Platform/Action/UpdatePlayerGameMetricsTest.php
```diff
@@ -0,0 +1,97 @@
...
+                'achievements_unlocked_softcore' => 2,
+            ],
+            $playerGames->only(
+                [
+                    'achievements_unlocked',
+                    'achievements_unlocked_hardcore',
+                    'achievements_unlocked_softcore',
+                ]
+            )
+        );
```
Jamiras, line 95: Nice test flow, but if you're going to implement the test, you should fully implement it and validate the other things updated y the action - like points and timestamps

app/Platform/Actions/UpdatePlayerGameMetrics.php
```diff
@@ -90,6 +91,7 @@ public function execute(PlayerGame $playerGame, bool $silent = false): void
             'points_weighted_total' => $game->TotalTruePoints,
             'points_weighted' => $pointsWeighted,
             'created_at' => $createdAt,
+            'achievements_unlocked_softcore' => $achievementsUnlockedSoftcoreCount,
```
Jamiras, line 94: recommend moving up so it's grouped with the other `achievements_unlocked*` fields.

app/Platform/Commands/UpdatePlayerGameMetrics.php
TwosomesUP: That's a good point. If we don't imagine having to run this against the entire table again, then removing this code and doing the update with straight SQL might be best. However, running a single update against such a large table could lock it up during the duration of the update's execution. 

To mitigate this, we may be able to batch the update. Let me do some research, and I can get back to you.

tests/Feature/Platform/Action/UpdatePlayerGameMetricsTest.php
TwosomesUP: Good call! I can work on including validations for the rest of the updates made in the action.

app/Platform/Commands/UpdatePlayerGameMetrics.php
TwosomesUP: Ok, I tested a few things and if we wanted to use SQL for this update we could utilize a temporary stored procedure that batches rows in chunks and commits them in transactions. It's still a lot of data to process, but at least it wouldn't lock quite as much of the table all at once. Here's a sample of what I ran on my local:

```
DROP PROCEDURE IF EXISTS UpdateSoftcoreUnlocks;
SELECT COUNT(*) AS "Rows to Process" FROM player_games WHERE achievements_unlocked_softcore IS NULL;

DELIMITER //

CREATE PROCEDURE UpdateSoftcoreUnlocks()
BEGIN
    DECLARE done INT DEFAULT FALSE;
    DECLARE batch_size INT DEFAULT 10000;
    DECLARE last_id INT DEFAULT 0;
    DECLARE total_rows INT;

    DECLARE cur CURSOR FOR
        SELECT id
        FROM player_games
        WHERE achievements_unlocked_softcore IS NULL
        ORDER BY id
        LIMIT batch_size;

    DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;

    SET total_rows = (SELECT COUNT(*) FROM player_games WHERE achievements_unlocked_softcore IS NULL);

    WHILE total_rows > 0 DO
        START TRANSACTION;

        OPEN cur;

        read_loop: LOOP
            FETCH cur INTO last_id;

            IF done THEN
                LEAVE read_loop;
            END IF;

            UPDATE player_games
            SET achievements_unlocked_softcore = achievements_unlocked - achievements_unlocked_hardcore
            WHERE id = last_id;

        END LOOP;

        CLOSE cur;

        COMMIT;

        SET total_rows = (SELECT COUNT(*) FROM player_games WHERE achievements_unlocked_softcore IS NULL AND id > last_id);
        SET done = FALSE;
    END WHILE;
END //

DELIMITER ;

CALL UpdateSoftcoreUnlocks();
DROP PROCEDURE UpdateSoftcoreUnlocks;

SELECT COUNT(*) AS "Rows Processed" FROM player_games WHERE achievements_unlocked_softcore IS NOT NULL;
```
Just let me know which direction you'd like to go in. We could run something like this, just execute a straight SQL update, or utilize artisan and queue a ton of jobs to do individual updates.

tests/Feature/Platform/Action/UpdatePlayerGameMetricsTest.php
TwosomesUP: Updates have been made to the test to cover all updates being made in the UpdatePlayerGameMetrics action. Just let me know if I missed anything!

app/Platform/Commands/UpdatePlayerGameMetrics.php
Jamiras: In the past, I've used bash to iterate. From my notes:
```
for i in $(seq 1 758500); do php artisan ra:platform:player:update-metrics $i; done
```

We could probably do something similar here too.

However this pans out, I want to merge this and run the queries in a separate release from #2466 so the data is ready before the code tries to use it.

TwosomesUP: That sounds good to me. If this PR goes through I'll be sure to update #2466 so it utilizes the new softcore column once everything is ready.

I do like the idea of running a bash loop for the command. Especially since it uses the existing code. As soon as I get a chance, I'll pull the changes to this file.

tests/Feature/Platform/Action/UpdatePlayerGameMetricsTest.php
```diff
@@ -0,0 +1,322 @@
...
+    use RefreshDatabase;
+
+    use TestsPlayerAchievements;
+    use TestsPlayerBadges;
+
+    public function testMetrics(): void
+    {
+        User::factory()->count(1)->create();
+
+        $user = User::first();
```
wescopeland, line 30:
```suggestion
        $user = User::factory->create();
```
nit: Less ambiguous to combine these. Some tests are seeded with an initial user (`$this->user`) which could make the intent here for a future maintainer unclear.

wescopeland (dismissed):

app/Platform/Commands/UpdatePlayerGameMetrics.php
TwosomesUP: @Jamiras the UpdatePlayerGameMetrics command has been reverted to its original state. I think using bash or SQL is the best option, since this is a one time update. Just let me know if you'd like to discuss how we're going to update the softcore column in `player_games` any further, but resolving the conversation for now.

Jamiras: The bash command was processing about 100 users per minute on my VM, meaning it would take ~5 days to fully populate the data. Prod should be faster, but I still expect it to take several hours.

The stored procedure was processing approximately 1000 player_games per second, which would still take about 3 hours.

The best overall approach seems to be to take the 7 minute query listed above and run it in chunks:
```
UPDATE player_games SET achievements_unlocked_softcore=achievements_unlocked-achievements_unlocked_hardcore WHERE id IN (SELECT id FROM player_games
WHERE achievements_unlocked_softcore IS NULL AND id <= 200000);

UPDATE player_games SET achievements_unlocked_softcore=achievements_unlocked-achievements_unlocked_hardcore WHERE id IN (SELECT id FROM player_games
WHERE achievements_unlocked_softcore IS NULL AND id > 200000 AND id <= 400000);
...
```

Each query took about 5 seconds, and we can let the server breathe between them. With ~12m player_games entries, it will take 60 individual queries to update all the records.

tests/Feature/Platform/Action/UpdatePlayerGameMetricsTest.php
```diff
@@ -0,0 +1,320 @@
...
+                    'beaten_percentage_hardcore',
+                    'beaten_at',
+                    'beaten_hardcore_at',
+                    'completed_at',
+                    'completed_hardcore_at',
+                    'completion_percentage',
+                    'completion_percentage_hardcore',
+                ]
+            )
+        );
```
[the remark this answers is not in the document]
TwosomesUP: That makes sense to me. I'll make the updates this week.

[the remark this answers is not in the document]
TwosomesUP: Individual lines for each assert added

app/Platform/Commands/UpdatePlayerGameMetrics.php
TwosomesUP: That sounds like a great idea to me. Just let me know if you need me to generate the script.

Jamiras (approved):

PR #2480, opened 2024-05-31
collapse user agents that share client/version

Jamiras:
Prevents duplicate entries from appearing in the Clients Used list on the ticket page.

For example, this would have listed "RANes (1.1.0.0)" twice, with separate tooltips. Now it lists it once with a compound tooltip.
image


app/Platform/Services/UserAgentService.php
```diff
@@ -15,6 +15,13 @@ public function decode(string $userAgent): array
 
     private function parseUserAgent(string $userAgent): array
     {
+        if (empty($userAgent) || $userAgent === '[not provided]') {
```
Jamiras, line 18: Special handling for [when a heartbeat call is made without a User Agent](https://github.com/RetroAchievements/RAWeb/blob/master/app/Platform/Events/PlayerSessionHeartbeat.php#L35). Also, short-circuit for null/empty entries that were previously being handled indirectly later.

wescopeland (approved):

PR #2482, opened 2024-06-04
don't send null author to client

Jamiras:
Prevents crash reported in https://github.com/libretro/RetroArch/issues/16648.

Also fixes exception "Missing required parameter for [Route: user.show]" when trying to view the unofficial achievements for the game.

There were 24 achievements associated to a user whose account was hard-deleted prior to 2018. About half of them were associated to the TMNT2 set referenced in the associated issue. When #2406 changed the query to use the foreign key, there was no associated account to retrieve, so the author was null. The client was not expecting null, resulting in the reported crash, and the webpage was not expecting null, reporting an exception.

NOTE: To address this without requiring this PR be merged/deployed (and to maintain the historical data that the non-foreign key had captured), I've created an empty account for the deleted user, and immediately marked it as deleted.

Before:
image

After:
image


wescopeland (approved):
Looks good. Thank you for hunting this one down.

PR #2483, opened 2024-06-06
feat: allow users to only accept communications from people they follow

wescopeland:
**Post-Deploy SQL to Execute**

```sql
-- add the 'team-account' role
INSERT INTO auth_roles (name, display, guard_name, created_at, updated_at)
VALUES('team-account', 1, 'web', NOW(), NOW());
```

---

Adds a new privacy setting:

Screenshot 2024-06-06 at 6 07 10 PM

When enabled:
* If a non-follower attempts to create a message thread with a user, they will receive a permissions error.
* If a non-follower attempts to reply with in an existing message thread with a user, they will receive a permissions error.
* The non-follower can no longer comment on the user's wall.

Junior Developers and up are excluded from the message threads rules.


app/Community/Controllers/MessageController.php
```diff
@@ -38,9 +38,20 @@ public function store(MessageRequest $request): RedirectResponse
...
+                }
+            }
+
             (new AddToMessageThreadAction())->execute($thread, $user, $input['body']);
         } else {
             $recipient = User::firstWhere('User', $input['recipient']);
+
+            if (!$user->can('createForRecipient', [MessageThread::class, $recipient])) {
+                return back()->withErrors(__('legacy.error.permissions'));
+            }
```
Jamiras, line 53: Do we really need two separate policies for this? Should this also just check `$user->can('sendToRecipient', [Message::class, $recipient]`?

app/Policies/MessagePolicy.php
```diff
@@ -46,4 +48,25 @@ public function forceDelete(User $user, Message $message): bool
...
     }
+
+    public function sendToRecipient(User $user, User $targetUser): bool
+    {
+        $isUserDeveloper = $user->hasAnyRole([
+            Role::DEVELOPER,
+            Role::DEVELOPER_JUNIOR,
+            Role::DEVELOPER_STAFF,
+        ])
+            || $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper;
```
Jamiras, line 59: RACheats, WritingTeam, RANews, and TheUnwanted don't have developer permissions, so won't be able to reply. We probably want to attribute all of the shared accounts with a "POINT_OF_CONTACT" role and add that to this list.

app/Policies/MessagePolicy.php
```diff
@@ -46,4 +48,25 @@ public function forceDelete(User $user, Message $message): bool
...
+
+        /**
+         * TODO check user privacy settings
+         */
```
Jamiras, line 63: Are there other user privacy settings I'm not aware of, or is this in reference to some future feature?

app/Policies/MessagePolicy.php
```diff
@@ -46,4 +48,25 @@ public function forceDelete(User $user, Message $message): bool
     {
         return false;
     }
+
+    public function sendToRecipient(User $user, User $targetUser): bool
+    {
+        $isUserDeveloper = $user->hasAnyRole([
+            Role::DEVELOPER,
+            Role::DEVELOPER_JUNIOR,
+            Role::DEVELOPER_STAFF,
```
Jamiras, line 57: Should this also include MODERATOR?

app/Community/Controllers/MessageController.php
```diff
@@ -38,9 +38,20 @@ public function store(MessageRequest $request): RedirectResponse
                 return back()->withErrors(__('legacy.error.error'));
             }
 
+            foreach ($thread->users as $threadUser) {
+                if (!$threadUser->is($user) && !$user->can('sendToRecipient', [Message::class, $threadUser])) {
+                    return back()->withErrors(__('legacy.error.permissions'));
```
Jamiras, line 43: This error is misleading. "Insufficient permissions". Something along the lines of "Unable to complete request" might work better.

Jamiras, line 43: This error is misleading. "Insufficient permissions". Something along the lines of "Unable to complete request" might work better.

app/Policies/MessagePolicy.php
wescopeland: It's a reference to a future feature; I found numerous references to the feature elsewhere in the codebase (search for "check user privacy settings"). I assume this data will be held in `UserAccounts.preferences`.

app/Community/Controllers/MessageController.php
wescopeland: Reduced down to `MessagePolicy::sendToRecipient()` in latest.

app/Policies/MessagePolicy.php
wescopeland: Added a new role in latest:
Screenshot 2024-06-07 at 6 41 01 PM

Also updated the PR description with some SQL needed to add it to the DB. After deployment, I can attach the role to:
* RAdmin
* RACheats
* DevCompliance
* QATeam
* WritingTeam
* RANews
* RAEvents
* TheUnwanted
* DevQuest
* RAArtTeam

app/Policies/MessagePolicy.php
wescopeland: It should, as well as a few others probably. Updated in latest to:
```php
$canUserAlwaysSend = $user->hasAnyRole([
    Role::ADMINISTRATOR,
    Role::DEVELOPER_JUNIOR,
    Role::DEVELOPER_STAFF,
    Role::DEVELOPER,
    Role::EVENT_MANAGER,
    Role::FORUM_MANAGER,
    Role::MODERATOR,
    Role::POINT_OF_CONTACT,
])
    || $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper;
```

app/Community/Controllers/MessageController.php
wescopeland: I went ahead and went with:
> This user isn't accepting messages right now.

This seems like the right balance of not giving away too much information, but also not being overly vague on what is happening. I'm afraid "unable to complete request" is inevitably going to lead to people filing bug reports.

wescopeland: https://github.com/RetroAchievements/RAWeb/pull/2483#discussion_r1631761741

app/Community/Controllers/MessageController.php
Jamiras: Does `MessageThread::class` need to be changed to `Message::class`?

app/Policies/CommentPolicy.php
```diff
@@ -57,11 +57,8 @@ public function create(User $user, ?Model $commentable = null): bool
             return false;
         }
 
-        // users cannnot comment on the wall of a user who has them blocked
         if ($commentable !== null && $commentable instanceof User) {
-            if ($commentable->isBlocking($user)) {
-                return false;
-            }
+            return (new UserCommentPolicy())->create($user, $commentable);
```
Jamiras, line 61: Shouldn't this be `$user->can('create', [UserComment::class, $commentable])` or something similar?

lang/en/permission.php
```diff
@@ -36,6 +36,7 @@
         Role::FOUNDER => __('Founder'),
         Role::ARCHITECT => __('Architect'),
         Role::ENGINEER => __('Engineer'),
+        Role::POINT_OF_CONTACT => __('Team Account'),
```
Jamiras, line 39: If "Team Account" is a better name for the role, now would be the time to change it. Once it's in the DB, it's a lot harder.

app/Community/Controllers/MessageController.php
wescopeland: Yes, thanks. Updated in latest.

app/Policies/CommentPolicy.php
wescopeland: It should. It didn't even occur to me that we had a `UserComment` class. Updated in latest.

lang/en/permission.php
wescopeland: Agreed - done.

Jamiras (approved):

PR #2490, opened 2024-06-10
refactor: migrate viewtopic.php to the pages dir

wescopeland:
This PR migrates _viewtopic.php_ to `/forums/topic/[forumTopic]`.

The quick reply form and Options controls have been migrated to Livewire. Functionally, there should be no changes.

Screenshot 2024-06-09 at 8 20 23 PM


wescopeland:
Moving to draft. If we're going to be migrating the UI to React, it's probably a better use of time to review the Inertia PRs.

PR #2496, opened 2024-06-13
chore(deps): prepare inertia

wescopeland:
No functional changes, just hoping to eliminate all the lockfile noise that adding this stuff to the project is inevitably going to cause. We're behind on some critical dependencies anyway so it makes sense to knock this out.

To test this PR, both `npm run build` and `npm run test` should pass. There should be no JS errors on the home page.

The important updates are:
* Bumped major versions for Vite, Vitest, and laravel-vite-plugin and updated some types.
* Switched the project to ESM (ECMAScript Modules) by adding `type: "module"` to the _package.json_ file, a necessary prerequisite for Inertia. This necessitated a minor update to our PostCSS config file.
* Set "private" to true in our _package.json_ file because RAWeb is not a package or a library.
* Deleted _jsconfig.json_ (unused) and swapped _tsconfig.json_ to use Laravel Breeze's sane default.
* Moved all the existing Alpine.js TALL stack stuff to a dedicated _tall-stack_ directory under _resources/js_ to prevent conflicts with Inertia.


Jamiras:
> To test this PR, both npm run build and npm run test should pass. There should be no JS errors on the home page.

`npm run build` and `npm run test` both pass, but I can't access the home page:

image

wescopeland:
Should be fixed in latest - I also encountered this issue in #2502 and forgot to backport the fix to this branch. Vite needs a slight config tweak to not blow away the client-side JS build.

Jamiras (approved):
Did some random spot-checking and didn't see any errors.

PR #2501, opened 2024-06-14
fix(forum): remove unnecessary sanitization from forum editing

wescopeland:
Resolves: https://discord.com/channels/476211979464343552/1026595325038833725/1251247318917841046

This sanitization is no longer necessary now that the entire UI is handled by Blade and Livewire. There are no other legacy callers.

### DIFF
--- a/app/Helpers/database/forum.php
+++ b/app/Helpers/database/forum.php
@@ -152,10 +152,6 @@ function setLatestCommentInForumTopic(int $topicID, int $commentID): bool
 
 function editTopicComment(int $commentID, string $newPayload): bool
 {
-    $newPayload = str_replace("'", "''", $newPayload);
-    $newPayload = str_replace("<", "&lt;", $newPayload);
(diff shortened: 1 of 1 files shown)


app/Helpers/database/forum.php
```diff
@@ -152,10 +152,6 @@ function setLatestCommentInForumTopic(int $topicID, int $commentID): bool
 
 function editTopicComment(int $commentID, string $newPayload): bool
 {
-    $newPayload = str_replace("'", "''", $newPayload);
```
Jamiras, line 155: Seems like this like is still required. Maybe the query should use a binding?

wescopeland: 1 moment, it's simple enough I can just convert it to Eloquent ORM.

wescopeland: Resolved in latest

Jamiras (approved):

PR #2502, opened 2024-06-15
feat: migrate `/forums/recent-posts` to Inertia.js+React

wescopeland:
This PR migrates the `/forums/recent-posts` to Inertia.js+React. Rather than delete the PHP code, it still exists. The React route can be accessed at `/forums/recent-posts2`. This is intended to help compare and contrast the existing page vs the rewritten page.

In a follow-up PR, the PHP page can be removed and replaced with the new one.

**How to test this PR:**
* `composer install` and `npm install`
* `sail up`
* `npm run build`
* With Sail running in the background, `sail artisan inertia:start-ssr`
* Navigate to http://localhost:64000/forums/recent-posts2

`sail artisan inertia:start-ssr` causes the React code to render on the server. For development purposes, it should not be used and the conventional `npm run dev` should be used instead.

Screenshot 2024-06-15 at 2 01 10 PM

There should only be two functional changes - quite minor.

**1. Breadcrumbs are slightly different.**
Screenshot 2024-06-15 at 2 01 46 PM

The breadcrumbs are built with [shadcn-ui/Breadcrumb](https://ui.shadcn.com/docs/components/breadcrumb). These are battle-tested and accessible. We can style these however we like and can match the existing style if we so choose, but I think this newer style works much better for both light and dark modes.

**2. Pagination is slightly different.**
Screenshot 2024-06-15 at 2 03 15 PM

Icons are being used instead of keyboard angle brackets (`<` and `>`).

---

The React components have very high test coverage. `npm run test` has been modified to be similar to `composer test`, using a dot reporter. You can find the tests in the project by searching for any file with the pattern _*.test.tsx_.

---

**Next steps:**
Assuming this PR is approved and merged, in a follow-up PR:
* The PHP code for `/forums/recent-posts` will be deleted.
* The query to build all the props for the page will be migrated to Eloquent ORM and likely called via an action.
* All the temporary code/TODOs in `ForumTopicCommentController::showRecentPosts()` will be removed.


resources/js/forums/components/RecentPostsMainRoot/AggregateRecentPostLinks/AggregateRecentPostLinks.test.tsx
wescopeland, line 1: Here's an example of a test suite. This suite is specifically testing this UI component:

Screenshot 2024-06-15 at 3 59 09 PM

Without a full-stack E2E test framework like Dusk or Playwright, this would be very difficult to test as a Blade component, and it would be quite fragile as well. The React tests are more lightweight, less code to write, and this entire test suite runs in just a few milliseconds.

wescopeland (commented):
A general tour of the code.

resources/js/common/components/UserAvatar/UserAvatar.tsx
wescopeland, line 1: If you've never seen a React component before, this is a good one to review. Here's an annotated version:

```tsx
// Almost every React component will use the `FC` type. This is shorthand
// for "function component". In React, components literally are functions. 
// They take in some arguments (called props) and with those props, they 
// output HTML.

// JSX/TSX itself is just syntactic sugar for dressing up function calls.
// For example: `<p>Hello, world</p>` will be compiled to 
// `React.createElement('p', null, 'Hello, world!');`.
import type { FC } from 'react';

// "use" is a magic reserved prefix in React. It designates that something is a "hook".
// Just like components are reusable UI, hooks are reusable behavior. 
import { useCardTooltip } from '@/common/hooks/useCardTooltip';

// This is analogous to Blade's @props() directive, only this version has
// true strict typing. The build will fail if these types are violated.
interface UserAvatarProps {
  displayName: string | null;

  // In TypeScript, if a field has a question mark suffixing its name, then
  // it isn't required for the caller to set its value.
  hasTooltip?: boolean;
  size?: 8 | 16 | 24 | 32 | 64 | 128;
}

export const UserAvatar: FC<UserAvatarProps> = ({ displayName, size = 32, hasTooltip = true }) => {
  // Hooks must be declared before any template logic and before any conditional logic.
  const { cardTooltipProps } = useCardTooltip({ dynamicType: 'user', dynamicId: displayName });

  // The FC type requires that the function either returns a template or `null`.
  // Because this is actually just JavaScript, you can use real ternaries in the template
  // to drive conditional logic. If a prop is set to be undefined, it will be omitted
  // entirely from the template.
  return (
    <a
      href={displayName ? route('user.show', [displayName]) : undefined}
      className="flex items-center gap-2"
      {...(hasTooltip && displayName ? cardTooltipProps : undefined)}
    >
      <img
        loading="lazy"
        decoding="async"
        width={size}
        height={size}
        src={`http://media.retroachievements.org/UserPic/${displayName}.png`}
        alt={displayName ?? 'Deleted User'}
        // There's a gotcha here. You cannot use `class` in JSX/TSX. This is because
        // we're still in JavaScript, and `class` is a reserved keyword.
        className="rounded-sm"
      />

      {displayName ? <span>{displayName}</span> : null}
    </a>
  );
};
```

app/Community/Controllers/ForumTopicCommentController.php
```diff
@@ -83,4 +92,46 @@ protected function destroy(ForumTopicComment $comment): RedirectResponse
...
+
+        // TODO after POC: remove this code
+        if (isset($pageProps['previousPageUrl'])) {
+            $pageProps['previousPageUrl'] = str_replace('recent-posts', 'recent-posts2', $pageProps['previousPageUrl']);
+        }
+        if (isset($pageProps['nextPageUrl'])) {
+            $pageProps['nextPageUrl'] = str_replace('recent-posts', 'recent-posts2', $pageProps['nextPageUrl']);
+        }
+
+        return Inertia::render('forums/recent-posts', $pageProps);
```
wescopeland, line 135: We'll probably want some kind of middleware to camelcase any response that's sent to the Inertia/React app.

resources/js/features/forums/components/RecentPostsMainRoot/RecentPostsPagination/RecentPostsPagination.tsx
```diff
@@ -0,0 +1,39 @@
...
+          <BasePaginationItem>
+            <BasePaginationPrevious href={previousPageUrl}>
+              Previous {maxPerPage}
+            </BasePaginationPrevious>
+          </BasePaginationItem>
+        ) : null}
+
+        {nextPageUrl ? (
+          <BasePaginationItem>
+            <BasePaginationNext href={nextPageUrl}>Next {maxPerPage}</BasePaginationNext>
```
Jamiras, line 33: Stylistically, shouldn't this be three lines to match up with the Previous implementation above?

resources/js/common/components/UserAvatar/UserAvatar.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { faker } from '@faker-js/faker';
+
+import { render, screen } from '@/test';
+
+import { UserAvatar } from './UserAvatar';
+
+describe('Component: UserAvatar', () => {
```
Jamiras, line 7: How are these tests run? I don't see any changes to add it to CI or pre-push

resources/js/common/components/+vendor/BaseBreadcrumb.tsx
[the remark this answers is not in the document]
wescopeland: Intentional. It's a tiny QoL thing to pin the "vendor" path near the top of the folder tree.

resources/js/features/forums/components/RecentPostsMainRoot/RecentPostsPagination/RecentPostsPagination.tsx
wescopeland: This formatting is left entirely at the mercy of ESLint & Prettier. This is beneficial, as we don't need to really focus on code formatting & style. Prettier can take the wheel.

When editing JS/TS, I always leave VSCode's "format on save" setting enabled:


https://github.com/user-attachments/assets/a90112e4-889f-4687-99ef-e435b8b43feb

resources/js/common/components/UserAvatar/UserAvatar.test.tsx
wescopeland: It all works seamlessly with our existing Vitest setup. Try `npm run test`.

resources/js/features/forums/components/RecentPostsMainRoot/RecentPostsPagination/RecentPostsPagination.tsx
wescopeland: Another example of Prettier auto-formatting. It's pretty normal to code something as a stream of consciousness and then hit save and let Prettier do all the work. Note that these Prettier rules have been in place since https://github.com/RetroAchievements/RAWeb/pull/2258.


https://github.com/user-attachments/assets/ae64d77c-28ae-4086-b7ae-6de7d2b7bed9

resources/js/features/forums/components/RecentPostsMainRoot/PostTimestamp/PostTimestamp.tsx
```diff
@@ -0,0 +1,20 @@
...
+  postedAt: string;
+  asAbsoluteDate: boolean;
+}
+
+export const PostTimestamp: FC<PostTimestampProps> = ({ postedAt, asAbsoluteDate }) => {
+  if (asAbsoluteDate) {
+    return dayjs.utc(postedAt).format('DD MMM YYYY, HH:mm');
+  }
+
+  return dayjs.utc(postedAt).fromNow();
```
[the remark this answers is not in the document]
wescopeland: In https://github.com/RetroAchievements/RAWeb/pull/2502/commits/66c5f6817876bc4ef650a85e674aa7b8468b2034, I've attempted to build a `diffForHumans()` util that, while not a perfect 1:1 match to Carbon's implementation, should be fairly close to how it's rounding the divides for certain time thresholds.

resources/js/app.tsx
luchaos, line 1: Seeing that ts files are indented with two spaces I think the `.editorconfig` should be updated as well:
```
[*.{css,js,yml,yaml,neon,rc,ts}]
indent_size = 2
```

app/Community/Controllers/ForumTopicCommentController.php
luchaos: This is a lot of code for manually mapping props 😅 I see that you added a comment to migrate/thin it out.

A general note:

I think the goal should be to receive a `LengthAwarePaginator` instance from the service which includes our custom pagination logic.

The translation/mapping can be done by using Resource classes and/or view models that take care of changing the case. Ideally we use the exact same attribute names as in the database in the frontend.

There are ways to generate types from models by using something like https://spatie.be/docs/typescript-transformer/v2/introduction

app/Http/Kernel.php
```diff
@@ -48,6 +48,10 @@ class Kernel extends HttpKernel
             // TODO \Illuminate\Routing\Middleware\ThrottleRequests::class.':web',
         ],
 
+        'inertia' => [
+            Middleware\HandleInertiaRequests::class,
+        ],
```
luchaos, line 53: It might be enough to add a `middlewareAliases` below if we do not plan on adding more to this group. I don't think it's necessary to create a group.

resources/js/app.tsx
wescopeland: Ah, I missed this. Good point.

In https://github.com/RetroAchievements/RAWeb/pull/2502/commits/1d7107f541cec7a22ce66ee3f6ac69ccc70ac83f, I've added both `ts` and `tsx` to the line.

app/Community/Controllers/ForumTopicCommentController.php
wescopeland: Thanks for this insight - this is great info to know.

Maybe the best place to address it will be in the follow-up PR that removes the old page in favor of the new Inertia one.

luchaos: Sounds good to me 👍 Ideally we can address this in a way that may act as a template for existing and new services alike. To get rid of most of the custom mappings and normalize pagination urls by using more Laravel features out of the box.

app/Http/Kernel.php
wescopeland: Good point, the `middlewareAliases` stuff below totally escaped me. Addressed in https://github.com/RetroAchievements/RAWeb/pull/2502/commits/0323a0b522942f22d464b0d4e9bf5e3175778c64.

resources/js/common/components/UserAvatar/UserAvatar.test.tsx
wescopeland: I just noticed that Sail was not using the correct version of Node.js. In https://github.com/RetroAchievements/RAWeb/pull/2502/commits/399dbd6aefc731b377d7a2b1fe76bae574145353 I've updated the `laravel.test` container to use Node 20.

Now you should be able to do:
```bash
sail npm install
sail npm run test
```

Jamiras (approved):
Seems functional. I don't have the skillset to review the code, or the experience to comment on the file layouts.

wescopeland:
Per @luchaos's comment [here](https://github.com/RetroAchievements/RAWeb/pull/2502#discussion_r1685744950), in https://github.com/RetroAchievements/RAWeb/pull/2502/commits/1e52954bb7aa011629a4135d7e37a4cb1750bc61 I've added missing dependencies, namely `spatie/laravel-typescript-transformer` and `spatie/laravel-data`.

These dependencies will give us the ability to build low-code data resources from our PHP models, as well as automatically generate TypeScript types from them.

PR #2505, opened 2024-06-16
refactor(game): migrate several helpers to eloquent orm

wescopeland:
This PR migrates several of the key game modification helper functions to use Eloquent ORM queries. Currently, since these are making raw SQL calls to the DB, most model updates made by developers are not showing up in the game entity's activitylog in the management panel.


app/Helpers/database/game.php
```diff
@@ -910,40 +920,36 @@ function submitNewGameTitleJSON(
...
+    $game->RichPresencePatch = $dataIn;
+    $game->save();
+
+    addArticleComment('Server', ArticleType::GameModification, $gameId, "{$username} changed the rich presence script");
 
     return true;
 }
 
-function getRichPresencePatch(int $gameID, ?array &$dataOut): bool
+function getRichPresencePatch(int $gameId, ?array &$dataOut): bool
```
Jamiras, line 943: $dataOut should be a string.

Jamiras (approved):

app/Helpers/database/game.php
wescopeland: Changed to string in latest. This one escaped me - has this been incorrectly typed the whole time?

PR #2507, opened 2024-06-17
Provide more detail in unlock history on ticket page

Jamiras:
Adds per-client usage rates to client summary on the ticket page (and the user game moderation page):
image

The unlock history is now grouped by session (showing the client used for each session):
image

Older tickets (submitted prior to capturing session information) show generated sessions. The "Generated Session" label has a tooltip to clarify what a generated session is:
image

Information showing when the ticket was created and when the associated achievement was unlocked is still injected into the history, including manual unlock information.
image

Also extracted a lot of logic from the ticket view page into a service.


app/Platform/Services/TicketService.php
```diff
@@ -0,0 +1,96 @@
...
+                    attributeEscape($hash->name) . '">' . $hash->md5 . '</a>';
+                $this->ticketNotes = str_ireplace($hash->md5, $replacement, $this->ticketNotes);
+            }
+        }
+    }
+
+    public function buildHistory(Ticket $ticket, User $user): void
+    {
+        $this->activity = new PlayerGameActivityService();
+        $this->userAgentService = new UserAgentService();
```
wescopeland, line 74: nit: These should probably be instantiated via DI in `__construct()`:

```php
public function __construct(
  protected PlayerGameActivityService $activity,
  protected UserAgentService $userAgentService,
) {
}
```

These will be spun up and torn down only during the lifetime of the current request.

Jamiras: done

app/Platform/Services/TicketService.php
```diff
@@ -0,0 +1,96 @@
...
+            $this->existingUnlock = PlayerAchievement::where('user_id', $ticket->reporter->id)
+                ->where('achievement_id', $ticket->achievement->id)
+                ->first();
+
+            $this->openTicketLinks = [];
+            $this->closedTicketLinks = [];
+            $achievementTickets = Ticket::where('AchievementID', $ticket->achievement->id);
+            foreach ($achievementTickets->get() as $otherTicket) {
+                if ($otherTicket->ID !== $ticket->ID) {
+                    $url = '<a href="' . route('ticket.show', ['ticket' => $otherTicket]) . '">' . $otherTicket->ID . '</a>';
```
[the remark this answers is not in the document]
Jamiras: done

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -280,4 +298,48 @@ public function summarize(): array
...
+                    $client .= ' (' . $decoded['clientVersion'] . ')';
+                }
+                if (array_key_exists('clientVariation', $decoded)) {
+                    $client .= ' - ' . $decoded['clientVariation'];
+                }
+
+                if (in_array($client, $clients)) {
+                    $clients[$client]['duration'] = $clients[$client]['duration'] + $session['duration'];
+                    if (!in_array($userAgent, $clients[$client]['agents'])) {
+                        $this->clients[$client]['agents'][] = $userAgent;
```
wescopeland, line 323: Possible typo and/or logic here? As best as I can tell, `$this->clients` is undefined.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -280,4 +298,48 @@ public function summarize(): array
             'totalPlaytime' => $totalTime,
         ];
     }
+
+    // returns array of ['agents' => [], 'duration' => 0, 'durationPercentage' => 0.0]
+    public function getClientBreakdown(UserAgentService $userAgentService): array
```
wescopeland, line 303: nit: Rather than accepting a `$userAgentService` as an arg, it's probably better to just instantiate one in PlayerGameActivityService in `__construct()` and use `$this->userAgentService->decode($userAgent)`.

app/Platform/Services/TicketViewService.php
```diff
@@ -0,0 +1,98 @@
...
+    public array $openTickets = [];
+    public array $closedTickets = [];
+    public string $contactReporterUrl = '';
+    public ?PlayerAchievement $existingUnlock = null;
+    public string $ticketNotes = '';
+
+    public function __construct(
+        public PlayerGameActivityService $activity = new PlayerGameActivityService(),
+        public UserAgentService $userAgentService = new UserAgentService(),
+    ) {
```
wescopeland, line 26:
```suggestion
    public function __construct(
        protected PlayerGameActivityService $activity,
        protected UserAgentService $userAgentService,
    ) {
```

Something external shouldn't be explicitly accessing `TicketViewService`'s instances of these two services. It should be okay to set them to `protected` and not need to `new` them up. This will cause Laravel DI to kick in.

wescopeland (commented):
Verified UI/UX. Just a few more comments on the code.

app/Platform/Services/PlayerGameActivityService.php
Jamiras: The service caches the decoded user agents, and this allows that cache to be leveraged by multiple places processing the user agents. The service may be poorly named. Maybe `UserAgentDecoder` or something similar would be better?

In general, classes named XXXService generally have too many responsibilities, but I was trying to conform to the existing codebase.

wescopeland: Ah, I think I understand now.

> Maybe `UserAgentDecoder` or something similar would be better?

Could be ... the design almost seems to imply a singleton pattern with lazy initialization 🤔 In Laravel, this may map to the concept of a service provider, but unfortunately I'm not too well-versed in this part of the framework.

I think what you have in the current head of the branch is fine. The most important thing was the clean separation of the application logic with the presentational layer, and that's all accounted for at this point.

wescopeland (approved):

PR #2520, opened 2024-06-22
fix: 2519

### LINKED ISSUE #2519: Cannot respond to tickets when muted
image

I was told that i would be able to comment on tickets but this is not the case.

wescopeland:
Resolves #2519.


Jamiras (approved):
Looks reasonable. Unable to test.

PR #2516, opened 2024-06-21
feat(manage): add achievement relation manager for games

wescopeland:
This PR ports most (all?) of the functionality from the achievement inspector to the management app.

Developers are able to:
* Drag and drop to reorder achievements.
* Toggle code columns.
* Bulk promote and demote.
* Bulk assign/remove types.

Additionally, actions are guarded using the `AchievementPolicy`, ensuring that Jr Devs can't update things they're not supposed to be able to.

The achievements are now listed on the game view page, eg: `/manage/games/1`.

I wish there was a way to pop some kind of "Success!" confirm on reordering. Unfortunately, it doesn't look like Filament supports this.

https://github.com/RetroAchievements/RAWeb/assets/3984985/1855cebb-ee1c-40ab-9ecf-a49ebde0d975

PR #2522, opened 2024-06-22
feat(manage): add game hash relation manager for games

wescopeland:
Accompanies, but is not dependent upon, #2516. After both are merged, we can put them into tabs on the page rather than have them load simultaneously.

This PR adds basic hash management to `/manage/games/{gameId}`. The intention is to mimic `/game/{gameId}/hashes/manage` in the player-facing app.

Screenshot 2024-06-22 at 6 49 52 PM

Screenshot 2024-06-22 at 7 01 37 PM

Screenshot 2024-06-22 at 7 04 30 PM


app/Models/GameHash.php
```diff
@@ -54,7 +57,8 @@ public function getRouteKeyName(): string
     public function getActivitylogOptions(): LogOptions
     {
         return LogOptions::defaults()
-            ->logOnlyDirty();
+            ->logOnlyDirty()
+            ->dontSubmitEmptyLogs();
```
Jamiras, line 61: I'm not seeing any new entries in the `audit_log` table.

Jamiras (approved):

app/Models/GameHash.php
wescopeland: Good catch -- updated in latest.

Jamiras (approved):

wescopeland:
It actually appears Filament is smart enough to put the relation managers in tabs automatically when they're both present:

Screenshot 2024-07-21 at 12 16 53 PM

PR #2523, opened 2024-06-23
fix(MessagePolicy): don't allow incoming DMs from blocked users

wescopeland:
If a given user X has user Y blocked, Y should not be able to send a direct message to X for any reason.


app/Policies/MessagePolicy.php
```diff
@@ -60,8 +59,11 @@ public function sendToRecipient(User $user, User $targetUser): bool
             Role::FORUM_MANAGER,
             Role::MODERATOR,
             Role::TEAM_ACCOUNT,
-        ])
-            || $user->getAttribute('Permissions') >= Permissions::JuniorDeveloper;
+        ]);
+
+        if ($targetUser->isBlocking($user)) {
+            return false;
+        }
```
Jamiras, line 66: It's probably desirable for TEAM_ACCOUNT (and maybe MODERATOR) to still be able to message the user, even if they're blocked.

Developer permission was primarily for the "Contact Reporter" functionality, but the conversation could still (presumably) take place in the ticket itself. Do we really even need a "Contact Reporter" button? Should the Developer items be removed from the canAlwaysSend logic?

wescopeland: Good thoughts.

I've renamed `$canAlwaysSend` to `$canUserAlwaysPierceNoContactPreference` and moved it closer to its respective conditional. I still tend to agree with the reasons why we allow devs and some others to pierce that preference, but I don't necessarily think the same subset of users need to be able to get through a block.

The block conditional logic now reads:
```php
$canUserSendWhileBlocked = $user->hasAnyRole([
    Role::ADMINISTRATOR,
    Role::MODERATOR,
    Role::TEAM_ACCOUNT,
]);
if ($targetUser->isBlocking($user) && !$canUserSendWhileBlocked) {
    return false;
}
```

Jamiras (approved):
Looks logical. Unable to test.

PR #2524, opened 2024-06-23
test(connect): add basic coverage to lbinfo

wescopeland:


tests/Feature/Connect/LeaderboardInfoTest.php
```diff
@@ -0,0 +1,56 @@
...
+            ->assertExactJson([
+                'Success' => true,
+                'LeaderboardData' => [
+                    'ConsoleID' => $system->id,
+                    'ConsoleName' => $system->name,
+                    'Entries' => [],
+                    'GameID' => $game->id,
+                    'GameIcon' => $game->ImageIcon,
+                    'GameTitle' => $game->title,
+                    'ForumTopicID' => $game->ForumTopicID,
```
Jamiras, line 42: rcheevos [doesn't use the game or system fields](https://github.com/RetroAchievements/rcheevos/blob/4357a1420db9f8c45353c365b6e1e126ba549144/src/rapi/rc_api_info.c#L193-L199). Maybe we could use this opportunity to remove them? I don't think anything is calling the endpoint directly.

tests/Feature/Connect/LeaderboardInfoTest.php
```diff
@@ -0,0 +1,56 @@
...
+        $leaderboard = Leaderboard::factory()->create(['GameID' => $game->id]);
+
+        $this->get($this->apiUrl('lbinfo', ['i' => $leaderboard->id]))
+            ->assertStatus(200)
+            ->assertExactJson([
+                'Success' => true,
+                'LeaderboardData' => [
+                    'ConsoleID' => $system->id,
+                    'ConsoleName' => $system->name,
+                    'Entries' => [],
```
Jamiras, line 38: I'd like to see a second test that validates Entries. It's good to have an empty test and a non-empty test.

Ideally, we'd also have a test where Entries is truncated, and one where the `u` parameter is passed, but I understand that this is is just "basic coverage".

tests/Feature/Connect/LeaderboardInfoTest.php
```diff
@@ -0,0 +1,56 @@
...
+                    'LBAuthor' => $leaderboard->developer?->User,
+                    'LBCreated' => $now->format('Y-m-d H:i:s'),
+                    'LBDesc' => $leaderboard->description,
+                    'LBFormat' => $leaderboard->format,
+                    'LBID' => $leaderboard->id,
+                    'LBMem' => $leaderboard->Mem,
+                    'LBTitle' => $leaderboard->title,
+                    'LBUpdated' => $now->format('Y-m-d H:i:s'),
+                    'LowerIsBetter' => (int) $leaderboard->rank_asc,
+                    'TotalEntries' => 0,
```
Jamiras, line 52: rcheevos doesn't reference this field at all. Was it added recently? Seems like something that might be useful to expose through. I'll make the changes there.

wescopeland: As best as I can tell, `TotalEntries` has been present for at least 3 years: https://github.com/RetroAchievements/RAWeb/pull/646/files

Agreed, seems like there could be some utility in that value.

tests/Feature/Connect/LeaderboardInfoTest.php
wescopeland: Makes sense. Went ahead and did this in the `lbinfo` routine rather than `GetLeaderboardData()`, as the helper function is also shared by the leaderboard info page.

`lbinfo` now has:
```php
// Remove fields that are unused by rc_client.
unset(
    $leaderboardData['GameTitle'],
    $leaderboardData['ConsoleID'],
    $leaderboardData['ConsoleName'],
    $leaderboardData['ForumTopicID'],
    $leaderboardData['GameIcon'],
);
```
Which is called if a leaderboard with `$lbID` is indeed found.

tests/Feature/Connect/LeaderboardInfoTest.php
wescopeland: No problem - this coverage was mostly a reaction to the formatting issue that came up a few weeks ago (ie: `'LBCreated' => $now->format('Y-m-d H:i:s')`).

However, adding another test for entries was no big deal. Added in latest.

tests/Feature/Connect/LeaderboardInfoTest.php
```diff
@@ -0,0 +1,123 @@
...
+        /** @var System $system */
+        $system = System::factory()->create();
+        /** @var Game $game */
+        $game = Game::factory()->create(['ConsoleID' => $system->id]);
+        /** @var Leaderboard $leaderboard */
+        $leaderboard = Leaderboard::factory()->create([
+            'GameID' => $game->id,
+            'LowerIsBetter' => false,
+            'Format' => ValueFormat::Score,
+        ]);
```
Jamiras, line 68: Suggest continuing the first test here to avoid rebuilding the database between tests. Start with the empty leaderboard fetch, add some entries, then fetch it again.

tests/Feature/Connect/LeaderboardInfoTest.php
Jamiras: They don't appear to be getting used there either. I commented out the code setting them in leaderboard.php and the leaderboard info page loaded without errors. It's actually getting that data directly from the model: https://github.com/RetroAchievements/RAWeb/blob/85b02885dbca9bfb484f45f14c3fb7c85eba5935/resources/views/pages-legacy/leaderboardinfo.blade.php#L41-L46

In fact, it looks like the only thing following `$lbData[` in the file is `'Entries'`.

tests/Feature/Connect/LeaderboardInfoTest.php
wescopeland: Done in latest. I am hoping when we upgrade to Laravel 11, we won't need to do this anymore ([ref](https://laravel.com/docs/11.x/releases#database-performance)).

To better separate the two test cases, I have included `#region` markers so they can at least be folded.

Screenshot 2024-07-17 at 8 27 32 PM

tests/Feature/Connect/LeaderboardInfoTest.php
wescopeland: Got it- I naively assumed they were used there. Removed in latest and reverted the `lbinfo` changes.

Jamiras (approved):

PR #2528, opened 2024-06-23
fix(user): enforce uniqueness on display names

wescopeland:
**SQL to Run _Before_ Migration/Deployment:**
```sql
-- reassign all display_name values to match User, which is already unique
UPDATE `UserAccounts`
SET `display_name` = `User`;
```

---

This PR adds a uniqueness constraint on `UserAccounts.display_name`. I don't believe it's desirable for two users to have the same `display_name`.

Also, the user profile route now supports navigation by `display_name` in addition to `User`.


resources/views/pages-legacy/userInfo.blade.php
```diff
@@ -23,7 +23,7 @@
     abort(400);
 }
 
-$userPageModel = User::firstWhere('User', $userPage);
+$userPageModel = User::where('User', $userPage)->orWhere('display_name', $userPage)->first();
```
Jamiras, line 26: This doesn't actually work. `getUserPageInfo` on the next line down tries to fetch the model where `User=$userPage`, and 404s.

wescopeland: This does seem to work if the `display_name` doesn't have special chars:
Screenshot 2024-07-20 at 10 19 55 AM

However, I am able to repro the issue if I try a name that contains non-ASCII chars. I've pushed some small changes to account for this.

We're going to probably need to fix the validation issue at both the code (`isValidUsername`) and the DB level. There are some crazy `display_name` values currently stored in the DB that I'm going to go out on a limb and say we should not support:

* ᜇᜒᜌᜒᜆ᜔ᜇᜓ
* COSTA༬
* Exe³³³
* ??ØØ?

In latest, I've updated `isValidUsername` to add what I think is a reasonable regex-based validation pattern. We can query against this pattern directly in the DB with:
```sql
SELECT * FROM UserAccounts
WHERE display_name NOT REGEXP '^[[:alpha:][:digit:][:space:]_.-]+$';
```

On my local, this yields 6 results. We should remove the display name for all of those users via:
```sql
DELETE FROM UserAccounts
WHERE display_name NOT REGEXP '^[[:alpha:][:digit:][:space:]_.-]+$';
```

app/Helpers/util/string.php
```diff
@@ -29,11 +28,20 @@ function attributeEscape(?string $input): string
...
 function isValidUsername(?string $username): bool
 {
+    /**
+     * Allows:
+     *   - \p{L}: Any kind of letter from any language.
+     *   - \p{N}: Any kind of numeric character from any script.
+     *   - \s   : Spaces.
+     *   - \-_.  : Hyphens, underscores, and periods.
+     */
+    $allowedPattern = '/^[\p{L}\p{N}\s\-_.]*$/u';
```
Jamiras, line 38: I think we want to continue to restrict user names to ASCII only. If they want Unicode characters, they can put those in the display name.

resources/views/pages-legacy/userInfo.blade.php
Jamiras: It works for `Scotty` because there's actually a `Scotty` user, in which case, I'd expect _that_ profile to load.

Given the possible overlap of display names and username, I think you should abandon trying to support display names in the URL.

wescopeland: Fair enough - reverted URL support changes in latest. I've also updated the SQL query in the OP to match your suggestion:
```sql
UPDATE `UserAccounts`
SET `display_name` = `User`;
```

Jamiras: Long term, I think we want the URL to _always_ look up by display name, since that's the only thing that should be public. If I change my display_name to "UniqueDisplayName", I wouldn't expect anyone to know to go to `/user/Jamiras` to find me.

Jamiras: It's a conversation for another time, but we should also maintain some sort of "retired" display_name list. If a user changes their display_name after some period of time (maybe six months or a year), the old display_name gets reserved so someone else wouldn't be able to impersonate the original user.

wescopeland: > Long term, I think we want the URL to always look up by display name, since that's the only thing that should be public. If I change my display_name to "UniqueDisplayName", I wouldn't expect anyone to know to go to /user/Jamiras to find me.

That makes sense to me. If we're in a world where all `{user}` routes are looking up by `display_name` instead of `User`, I'm guessing we probably don't need to worry about enforcing no overlap between `display_name` and the username.

> we should also maintain some sort of "retired" display_name list

My understanding is this is what the `user_usernames` table is for.

Jamiras:
> I've also updated the SQL query in the OP to match your suggestion:
```
UPDATE `UserAccounts`
SET `display_name` = `User`;
```
Don't we still want to maintain those unicode values that already exist? (where not null, and resolve uniqueness)

wescopeland:
If we do, I still think we should throw out the ones that I categorized as "crazy" earlier:

```sql
DELETE FROM UserAccounts
WHERE display_name NOT REGEXP '^[[:alpha:][:digit:][:space:]_.-]+$';
```
then:
```sql
UPDATE `UserAccounts`
SET `display_name` = 'User'
WHERE `display_name` IS NOT NULL;
```

Jamiras:
@luchaos - [Since you went through the effort](https://discord.com/channels/476211979464343552/1002697125081124986/1167420029113999370) of transferring the non-ASCII usernames to the display_name field, what are your thoughts on sanitizing them?

There are currently six accounts that are incompatible with a unique key on display_name:
```
> select User, display_name, LastLogin from UserAccounts where display_name in (select display_name from UserAccounts where display_name is not null group by display_name having count(*) > 1) order by display_name;
```
Five of which haven't even been used as far as I can tell.

wescopeland:
All `display_name` values in prod have been set to null: https://discord.com/channels/476211979464343552/718091467893243937/1272243405333856347

Jamiras (approved):

wescopeland:
Going to go ahead and merge this to get it off our plate.

I'm not sure if we want to run this SQL query in any environment:
```sql
UPDATE `UserAccounts`
SET `display_name` = `User`;
```

MariaDB treats NULL as unique, and we currently don't have any code to double-write this value (and I'm not sure that's desirable). My thinking right now is we continue to leave all `display_name` values as NULL until we support username changes.

PR #2535, opened 2024-06-26
chore(database): add game_hash_id to player_achievements and leaderboard_entries

wescopeland:
`rc_client` is already sending the currently-used hash for the `awardachievement` and `submitlbentry` routines. However, the server isn't storing it anywhere.

While `game_hash_id` exists on `player_games` and `player_sessions`, this is actually a reference to the most recently-used hash by the player. It is a close approximation, but not an exact measurement, of the hash the player used to unlock X achievement or submit Y leaderboard entry.

For trustworthy analytics, we need more precise tracking of these values at the achievement and leaderboard entry level. This PR adds a migration to insert these columns into those two tables.

**WARNING: The migration takes approximately 40 minutes to complete.** Given this duration, it makes sense to separate the migration from the PR that includes any logic changes to record the values. They'll need to be deployed separately.

While the migration is running, MariaDB leaves both tables open for reads and writes. This should not cause any disruption on prod.

### DIFF
--- a/database/migrations/platform/2024_06_25_000000_update_player_hash_references.php
+++ b/database/migrations/platform/2024_06_25_000000_update_player_hash_references.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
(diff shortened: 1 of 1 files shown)


wescopeland:
Closing, as it appears we can figure this out using relations that already exist.

PR #2546, opened 2024-07-06
feat(manage): set up basic leaderboard and entries management

wescopeland:
Begins the process of migrating leaderboard & entry management to the management app.
Editing of leaderboards is not yet supported, however a user can search for leaderboards, view their entries, and delete entries.

https://github.com/RetroAchievements/RAWeb/assets/3984985/97d8abf9-0193-48a3-984e-4ad93773ac1a


app/Policies/LeaderboardEntryPolicy.php
```diff
@@ -14,6 +13,16 @@ class LeaderboardEntryPolicy
 {
     use HandlesAuthorization;
 
+    public function manage(User $user): bool
+    {
+        return $user->hasAnyRole([
+            Role::GAME_HASH_MANAGER,
```
Jamiras, line 19: Probably don't need game hash manager here. Maybe cheat investigator?

What's the difference between this and the delete logic below? How can an entry be managed other than through deletion?

app/Policies/LeaderboardPolicy.php
```diff
@@ -5,30 +5,45 @@
...
 use Illuminate\Auth\Access\HandlesAuthorization;
 
 class LeaderboardPolicy
 {
     use HandlesAuthorization;
 
+    public function manage(User $user): bool
+    {
+        return $user->hasAnyRole([
+            Role::GAME_HASH_MANAGER,
```
Jamiras, line 19: Hash manager?

app/Policies/LeaderboardEntryPolicy.php
wescopeland: Swapped for cheat investigator.

`manage()` is what Filament uses to determine if a resource is viewable for a user.

app/Policies/LeaderboardPolicy.php
wescopeland: Removed.

Jamiras (approved):

PR #2549, opened 2024-07-08
feat(api): add endpoint for user's Want to Play Games list

ioslife:
Trying my hand at adding an endpoint for getting a user's backlog/want to play list. 

This is a draft, would like to get Wes' opinion on if I am going down the right road before I get any further into this.


ioslife:
Helpers are being deprecated, so going to go a different route.

wescopeland:
My summarized feedback, just for public visibility:
* We're in the process of deprecating all helpers, so let's not add a new one.
* We should either reuse (or add reusable) model attributes, or keep the logic strictly in-line within the API endpoint file itself.
* The endpoint needs to be comprehensively tested and documented, both here in RAWeb and in api-docs.
* There is a user in prod who apparently has over 8000 games on their want to play list -- we should likely paginate this endpoint.
* We can't review PRs that are untested - please do not rely on CI alone to verify the code. Verify locally before the code hits CI.

ioslife:
api-docs PR: https://github.com/RetroAchievements/api-docs/pull/52

ioslife:
the phpunit step failed for something I haven't touched and haven't seen fail before. Is this something I need to fix for this PR?

tests/Feature/Api/V1/UserWantToPlayListTest.php
wescopeland, line 1: Please also verify that the `o` and `c` parameters are working.

ioslife: I have manually tested it, but I'll add unit tests for it as well.

ioslife: Added offset, count, and offsetAndCount test cases

ioslife: Ah didn't realize the migrations happen every time. Merged all my scenarios into one

tests/Feature/Api/V1/UserWantToPlayListTest.php
```diff
@@ -0,0 +1,315 @@
...
+                        'AchievementsPublished' => $gameTwo->achievements_published,
+                    ],
+                ],
+            ]);
+
+            $this->get($this->apiUrl('GetUserWantToPlayList', ['u' => $user->User, 'o' => 5, 'c' => 2]))
+            ->assertSuccessful()
+            ->assertJson([
+                'Count' => 2,
+                'Total' => 2,
```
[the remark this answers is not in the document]
ioslife: Ah, that's because I am handling the offset and count in the get query

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,74 @@
...
+    ->pluck('GameID')
+    ->toArray();
+
+$pagedResults = array_slice($wantToPlayGameIDs, $offset, $count);
+
+$results = [];
+
+if (!empty($pagedResults)) {
+    foreach ($pagedResults as $nextGameID) {
+        $game = Game::with('system')->find($nextGameID);
```
wescopeland, line 54: There is [an N+1 problem](https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it) lurking here. This endpoint can make over 100 calls to the database per invocation.

I think the optimal way to handle this will be to have two queries:
* The first query's job should be to determine `Total`. This can be very lightweight.
* The second query's job should be to fetch the data for the current page.

ioslife: Yeah, that is much cleaner. Done!

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,71 @@
...
+ *    o : offset - number of entries to skip (default: 0)
+ *    c : count - number of entries to return (default: 100, max: 500)
+ *  int         Count                       number of want to play game records returned in the response
+ *  int         Total                       number of want to play game records the user actually has overall
+ *  array       Results
+ *   object      [value]
+ *    int        ID                         unique identifier of the game
+ *    string     Title                      name of the game
+ *    int        ConsoleID                  unique identifier of the console associated to the game
+ *    string     ConsoleName                name of the console associated to the game
```
wescopeland, line 15: This isn't included in the response.

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+/*
+ *  API_GetUserWantToPlayList - returns a list of GameIDs that a user has saved on their WantToPlayList
```
wescopeland, line 4: This description probably needs to be updated.

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,71 @@
...
+    $gameData = [
+        'ID' => $game->ID,
+        'Title' => $game->Title,
+        'ConsoleID' => $game->ConsoleID,
+        'ImageIcon' => $game->ImageIcon,
+        'PointsTotal' => $game->points_total,
+        'AchievementsPublished' => $game->achievements_published,
+    ];
+    $results[] = $gameData;
+}
```
wescopeland, line 65: I think we can optimize this even further.

```php
// untested pseudocode

// Fire off a quick query for the total, using `COUNT(*)`.
$totalWantToPlayItems = UserGameListEntry::where('user_id', $user->id)
    ->where('type', UserGameListType::Play)
    ->count();

// Paginate at the DB level and then map the collection results.
$results = UserGameListEntry::where('user_id', $user->id)
    ->where('type', UserGameListType::Play)
    ->with(['game.system'])
    ->skip($offset)
    ->take($count)
    ->get()
    ->map(function ($entry) {
        $game = $entry->game;
        return [
            'ID' => $game->id,
            'Title' => $game->title,
            'ConsoleName' => $game->system->name,
            'ImageIcon' => $game->ImageIcon,
            'TotalPoints' => $game->points_total,
            'AchievementsPublished' => $game->achievements_published,
        ];
    });
```

tests/Feature/Api/V1/UserWantToPlayListTest.php
```diff
@@ -0,0 +1,315 @@
...
+                        "ImageIcon" => $gameTen->ImageIcon,
+                        "ConsoleID" => $system->ID,
+                        "PointsTotal" => $gameTen->points_total,
+                        'AchievementsPublished' => $gameTen->achievements_published,
+                    ],
+                ],
+            ]);
+            $this->get($this->apiUrl('GetUserWantToPlayList', ['u' => $user->User, 'o' => 5]))
+            ->assertSuccessful()
+            ->assertJson([
```
wescopeland, line 218: nit: Indentations on these and the ones below are inconsistent.

tests/Feature/Api/V1/UserWantToPlayListTest.php
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    public function testGetUserWantToPlayListUnknownUser(): void
+    {
+        $this->get($this->apiUrl('GetUserWantToPlayList', ['u' => 'nonExistant']))
+            ->assertSuccessful()
+            ->assertJson([]);
+    }
+
+    public function testGetUserWantToPlayList(): void
```
wescopeland, line 36: nit: Recommend cutting this from 10 games to 5. The validation will be just as effective and it'll make the test a little easier for a reader to skim.

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+/*
```
wescopeland, line 3: nit: In this doc block, make sure all documentation indentation is consistent. There is too much indentation in the last two items:

```
 *    string     ConsoleName                name of the console associated to the game
 *    string     ImageIcon                  site-relative path to the game's icon image
 *    int        TotalPoints                     total points able to be earned
 *    int        NumPossibleAchievements            total number of achievements to be unlocked
```

wescopeland: `NumPossibleAchievements` should also be renamed. Please do a one-over of the doc block to ensure it's consistent with the latest implementation details.

public/API/API_GetUserWantToPlayList.php
ioslife: Oh cool, this is where the Eloquent knowledge comes into play. Adjusted your code to work properly.

public/API/API_GetUserWantToPlayList.php
ioslife: Good catch, removed. I was hoping to include it in the response, but didn't have access on the tables I am querying.

public/API/API_GetUserWantToPlayList.php
ioslife: Adjusted indents and made sure docs look right

wescopeland:
The only thing I'm waiting on before doing a final review pass is outstanding feedback to be addressed in https://github.com/RetroAchievements/api-docs/pull/52.

ioslife:
> The only thing I'm waiting on before doing a final review pass is outstanding feedback to be addressed in [RetroAchievements/api-docs#52](https://github.com/RetroAchievements/api-docs/pull/52).

Fixed, still not sure why it was showing any diff between what appeared to be the same

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,70 @@
...
+    'o' => ['sometimes', 'integer', 'min:0', 'nullable'],
+    'c' => ['sometimes', 'integer', 'min:1', 'max:500', 'nullable'],
+]);
+
+$offset = $input['o'] ?? 0;
+$count = $input['c'] ?? 100;
+
+$user = User::firstWhere('User', request()->query('u'));
+if (!$user) {
+    return response()->json([]);
```
wescopeland, line 39: Recommend also attaching a 404 status code, similar to API_GetUserProfile.

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,70 @@
...
+ *    o : offset - number of entries to skip (default: 0)
+ *    c : count - number of entries to return (default: 100, max: 500)
+ *  int         Count                       number of want to play game records returned in the response
+ *  int         Total                       number of want to play game records the user actually has overall
+ *  array       Results
+ *   object      [value]
+ *    int        ID                         unique identifier of the game
+ *    string     Title                      name of the game
+ *    int        ConsoleID                  unique identifier of the console associated to the game
+ *    string     ImageIcon                  site-relative path to the game's icon image
```
wescopeland, line 15: ConsoleName is undocumented.

tests/Feature/Api/V1/UserWantToPlayListTest.php
```diff
@@ -0,0 +1,212 @@
...
+    {
+        Carbon::setTestNow(Carbon::now());
+
+        /** Set up a user with 10 games on Want to Play List: */
```
wescopeland, line 40:
```suggestion
        /** Set up a user with 5 games on Want to Play List: */
```

wescopeland (approved):

wescopeland:
Approved - I have no reservations with this implementation.

The only thing I think we should consider is whether or not there are any privacy concerns with this, as currently on the website the data is not publicly exposed. I don't think this data is particularly sensitive so it isn't something I'm worried about, but it feels worth noting.

Before merging, I'll give enough time so that any another maintainers can weigh in if they have an opinion regarding this.

Jamiras:
> The only thing I think we should consider is whether or not there are any privacy concerns with this, as currently on the website the data is not publicly exposed. 

Set requests are public, so I don't see any reason why want to play lists couldn't also be public. From an API perspective, I totally understand the desire to have that information in a client, but should it be limited to the user making the call? In my opinion, it's unlikely that a user would ever ask for the want to play list of another user. What would they do with it?

I know there have been several discussions around the want to dev list. Some people want to share their list so others can potentially collaborate, whereas others specifically don't want the list public so people don't pester them about actually starting development. The compromise was to make the list for a game visible to the claimant so they can seek collaborators after they claim it. People from the first group still want their list public as a user might not claim a game unless they had already arranged the collaboration.

wescopeland:
> should it be limited to the user making the call? In my opinion, it's unlikely that a user would ever ask for the want to play list of another user. What would they do with it?

I think this is a great point. I'm in favor of this endpoint returning a 401 if the caller is requesting a list that isn't their own.

wescopeland:
** But this isn't a terribly strong opinion, and not one I'd hold the PR up for, for the same reason you've stated:

> it's unlikely that a user would ever ask for the want to play list of another user. What would they do with it?

The only thing I can imagine where this sort of situation would come up is an app/integration reusing the same API key across multiple users, or an app that is using some form of connection pooling to bypass our rate limit.

ioslife:
Would this be achieved by checking username vs supplied API key?

The only thing I'd like to challenge with this is maybe allowing users that follow each other the ability to see each other's list. 
i.e. if I follow wes on the site and wes follows me, we could both see each other's list. 

I don't feel super strongly about that idea and not sure what the implementation would look like on the API side, but I could see a use case for it as an end user.

wescopeland:
> Would this be achieved by checking username vs supplied API key?

This sounds like authorization to me, which leads me to believe it should be achieved via an ability in a `UserGameListEntryPolicy`. There may be some nuance here as the model supports three different types of lists (see `UserGameListType` -- this should probably be renamed to `UserGameListEntryType`).

> The only thing I'd like to challenge with this is maybe allowing users that follow each other the ability to see each other's list.
i.e. if I follow wes on the site and wes follows me, we could both see each other's list.

I think I'd be fine with this as long as the authorization is enforced solely by the policy mentioned above.

ioslife:
> > Would this be achieved by checking username vs supplied API key?
> 
> This sounds like authorization to me, which leads me to believe it should be achieved via an ability in a `UserGameListEntryPolicy`. There may be some nuance here as the model supports three different types of lists (see `UserGameListType` -- this should probably be renamed to `UserGameListEntryType`).
> 
> > The only thing I'd like to challenge with this is maybe allowing users that follow each other the ability to see each other's list.
> > i.e. if I follow wes on the site and wes follows me, we could both see each other's list.
> 
> I think I'd be fine with this as long as the authorization is enforced solely by the policy mentioned above.

I'll have to look more into policies as that is definitely out of my realm of knowledge. 

This puts this PR on hold along with my PR for GetComments https://github.com/RetroAchievements/RAWeb/pull/2552

If anyone has some recommended reading for how to go about this, or would be interested in pairing with me to walk through how these work, let me know.

ioslife:
I got the policy and friend relationship working as expected. There is probably some optimization that can be done, but will get to that later. 

I am trying to get the tests to work as expected, but using this->actingAs($user) is not respecting using the call as $user.

ioslife:
Resolved the issue with actingAs($user) by no longer using BootstrapV1Api for setup. This was automatically setting a user as the one to use the webAPIKey which caused issues with how we were setting the user.

ioslife:
@Jamiras based on your feedback I have added a policy for only allowing users to see their own want to play list, with one exception. If two users are have a mutual following relationship (are friends), they are allowed to see the list.

app/Models/User.php
```diff
@@ -556,4 +558,24 @@ public function scopeVerified(Builder $query): Builder
...
+    public function friends(): BelongsToMany
+    {
+    return $this->belongsToMany(User::class, 'friends', 'user_id', 'related_user_id')
+        ->wherePivot('Friendship', '1')
+        ->whereExists(function ($query) {
+            $query->select(DB::raw(1))
+                  ->from('friends as f')
+                  ->whereRaw('f.user_id = UserAccounts.id AND f.related_user_id = ' . $this->id)
+                  ->where('Friendship', '1');
+        });
```
ioslife, line 574: This can likely be optimized, though I am not exactly sure how.

app/Policies/UserGameListEntryPolicy.php
```diff
@@ -0,0 +1,71 @@
...
+    //         ? Response::allow()
+    //         : Response::denyWithStatus(401);
+    // }
+
+    public function view(User $user, User $targetUser): Response
+    {
+        return ($user->id === $targetUser->id) || $user->isFriendsWith($targetUser)
+        ? Response::allow()
+        : Response::denyWithStatus(401);
+    }
```
ioslife, line 45: I wanted to have a way to override the `view(User, UserGameListEntry)` method, but I was getting yelled at by IDE for having two of the same method names. Any recommendations on this?

public/API/API_GetUserWantToPlayList.php
```diff
@@ -0,0 +1,80 @@
...
+if (!$user) {
+    return response()->json([], 404);
+}
+
+$policy = new UserGameListEntryPolicy();
+$response = $policy->view(Auth::user(), $user);
+
+if ($response->denied()) {
+    return response()->json([], 401);
+}
```
ioslife, line 50: This works, but not sure if instantiating the policy like this is the proper way to use a policy on this endpoint.

tests/Feature/Api/V1/UserWantToPlayListTest.php
```diff
@@ -0,0 +1,297 @@
...
+        $user = User::factory()->create(['User' => 'myUser']);
+        $this->user = $user;
+    }
+
+    protected function apiUrl(string $method, array $params = []): string
+    {
+        $params = array_merge(['y' => $this->user->APIKey], $params);
+
+        return sprintf('API/API_%s.php?%s', $method, http_build_query($params));
+    }
```
ioslife, line 38: I am overriding the `BootstrapV1Api` methods here to ensure that `myUser` is the one that is making the API calls.

app/Policies/UserGameListEntryPolicy.php
```diff
@@ -0,0 +1,71 @@
...
+
+    // public function view(User $user, UserGameListEntry $userGameListEntry): bool
+    // {
+    //     return $user->ID === $userGameListEntry->user_id
+    //         ? Response::allow()
+    //         : Response::denyWithStatus(401);
+    // }
+
+    public function view(User $user, User $targetUser): Response
```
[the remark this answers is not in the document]
ioslife: Yup, that is very valid. Adjusted.

app/Policies/UserGameListEntryPolicy.php
```diff
@@ -0,0 +1,69 @@
...
+
+        return false;
+    }
+
+    // public function view(User $user, UserGameListEntry $userGameListEntry): bool
+    // {
+    //     return $user->ID === $userGameListEntry->user_id
+    //         ? Response::allow()
+    //         : Response::denyWithStatus(401);
+    // }
```
wescopeland, line 38:
```suggestion
```

app/Policies/UserGameListEntryPolicy.php
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if ($user->hasAnyRole([
+            Role::MODERATOR,
+            Role::ADMINISTRATOR,
+        ])) {
+            return true;
+        }
+
+        return false;
```
wescopeland, line 30: nit:
`return $user->hasAnyRole([...])`

app/Policies/UserGameListEntryPolicy.php
```diff
@@ -0,0 +1,69 @@
...
+    // public function view(User $user, UserGameListEntry $userGameListEntry): bool
+    // {
+    //     return $user->ID === $userGameListEntry->user_id
+    //         ? Response::allow()
+    //         : Response::denyWithStatus(401);
+    // }
+
+    public function view(User $user, User $targetUser): bool
+    {
+        return $user->id === $targetUser->id || $user->isFriendsWith($targetUser);
```
wescopeland, line 42:
```suggestion
        return $user->is($targetUser) || $user->isFriendsWith($targetUser);
```

app/Models/User.php
```diff
@@ -556,4 +558,24 @@ public function scopeVerified(Builder $query): Builder
     {
         return $query->where('Permissions', '>', 0);
     }
+
```
wescopeland, line 561: These new functions all belong in _ActsAsCommunityMember.php_.

Use `relatedUsers()`, `followedUsers()`, etc as a reference to build the new relation.

ioslife: Hah, this makes it so much simpler.

ioslife: resolved with:
```
/**
     * Check if two users are mutually following each other.
     */
    public function isFriendsWith(User $user): bool
    {
        return $this->isFollowing($user) && $user->isFollowing($this);
    }
```

app/Community/Concerns/ActsAsCommunityMember.php
```diff
@@ -98,6 +98,14 @@ public function isBlocking(User $user): bool
         return $this->getRelationship($user) === UserRelationship::Blocked;
     }
 
+    /**
+     * Check if two users are mutually following each other.
+     */
```
wescopeland, line 103:
```suggestion
```
nit: Normally I'm in favor of as much documentation as possible, but this only describes verbatim the function implementation.

wescopeland (commented):
I only see one other thing in the current implementation that is worth commenting on. I'll do a final round of functional testing later tonight or tomorrow evening.

app/Community/Concerns/ActsAsCommunityMember.php
ioslife: Yeah, that's fine by me. I only had it since "friends" and "mutually following" are different, but you're right that the code speaks for itself.

wescopeland (approved):

PR #2551, opened 2024-07-09
feat(manage): migrate Most Reported Games to filament

wescopeland:
This PR migrates the Most Reported Games page to the management app. This does imply the page is now auth-gated, however it's arguable this page probably shouldn't be public-facing.

New columns include:
* Player Count
* Tickets Opened Since
* Ticketed Achievements

If a game has >=10 open tickets, the ticket count changes to the "danger" color.

**Before**
Screenshot 2024-07-08 at 8 01 17 PM

**After**
Screenshot 2024-07-08 at 8 00 59 PM


Jamiras:
> This does imply the page is now auth-gated, however it's arguable this page probably shouldn't be public-facing.

I agree with this. Users can still see how many tickets a single game has and determine whether or not they want to engage based on that. There's no reason for anyone not looking to repair a set to be looking for games with a lot of tickets. Also note that you can still sort the games lists for individual consoles by number of tickets.

Jamiras (approved):

PR #2552, opened 2024-07-10
feat(api): Add endpoint for GetComments

ioslife:
To address: https://github.com/RetroAchievements/RAWeb/discussions/2081#discussioncomment-9875779

```/*
*  API_GetComments - returns the comments associated to a game or achievement
*    i : game or achievement id
*    u : username
*    t : 1 = game, 2 = achievement, 3 = user
*
*  int         Count                       number of comment records returned in the response
*  int         Total                       number of comment records the game/achievement/user actually has overall
*  array       Results
*   object      [value]
*    int         User                      username of the commenter
*    string      Submitted                 date time the comment was submitted
*    string      CommentText               text of the comment
*/
```


public/API/API_GetComments.php
```diff
@@ -0,0 +1,76 @@
...
+/*
+*  API_GetComments - returns the comments associated to a game or achievement
+*    i : game or achievement id
+*    u : username
+*    t : 1 = game, 2 = achievement, 3 = user
+*    o : offset - number of entries to skip (default: 0)
+*    c : count - number of entries to return (default: 100, max: 500)
+*
+*  int         Count                       number of comment records returned in the response
+*  int         Total                       number of comment records the game/achievment/user actually has overall
```
ioslife, line 16:
```suggestion
*  int         Total                       number of comment records the game/achievement/user actually has overall
```

fix typo

ioslife:
I broke my tests before the last push. Fixing now.

wescopeland (changes requested):
This endpoint surfaces multiple security issues.

Comments for banned users are still returned even though they're invisible in the UI. Comments for users are still returned if they have their wall disabled.

ioslife:
I've converted this to a draft for now. After talking with Wes, there is a bigger effort going on to deprecate Permissions and start using Policies. 

```
resources/views/components/comment/item.blade.php first needs to be refactored to use policies.
This is nontrivial because there are multiple types of commentables and different policies mapping to each.
```
- comments from Wes

I will keep looking at this, but it's getting to a point that is out of my wheel house. I'd love for someone to work with me to make sure we get this done properly.

database/factories/CommentFactory.php
```diff
@@ -0,0 +1,32 @@
...
+            'Payload' => $this->faker->paragraph,
+            'Submitted' => $this->faker->dateTimeBetween('-1 year', 'now'),
+            'Edited' => $this->faker->dateTimeBetween('now', '+1 year'),
+            'user_id' => 1,
+            'commentable_id' => $this->faker->numberBetween(1, 100),
+            'commentable_type' => $this->faker->randomElement([
+                'App\Models\Article',
+                'App\Models\Post',
+                'App\Models\Product',
+            ]),
```
wescopeland, line 29: Review pending, but I need this factory for something else so leaving a comment here 😅

Let's omit these two from the factory for now as they're currently unused in production.

app/Models/Comment.php
wescopeland, line 1: For convention's sake, we typically add `protected static function newFactory()...` to our models when setting up factories for them.

database/factories/CommentFactory.php
ioslife: This was generated with `php artisan make:factory CommentFactory` iirc. I'll remove the commentables and add the Articles.

wescopeland:
Not entirely sure what I'm doing wrong:

```
http://localhost:64000/API/API_GetComments.php?z=...&y=...&i=1&t=1
```

```
{
	"message": "App\\Models\\User::getUsernameAttribute(): Return value must be of type string, null returned",
	"exception": "TypeError",
	"file": "/var/www/html/app/Models/User.php",
	"line": 412,
	"trace": [
		{
			"file": "/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php",
			"line": 658,
			"function": "getUsernameAttribute",
			"class": "App\\Models\\User",
			"type": "->"
		},
... a bunch of stuff
```

app/Policies/CommentPolicy.php
```diff
@@ -26,7 +26,7 @@ public function manage(User $user): bool
 
     public function view(?User $user, Comment $comment): bool
     {
-        return true;
+        return $user->isNotBanned();
```
wescopeland, line 29: In this case, is `$user` the currently logged in user, or is it the author of the comment?

I believe it's the currently logged in user. In which case, if it's a moderator or administrator, I'd expect them to be able to view any comment, even if the user is banned.

database/factories/CommentFactory.php
```diff
@@ -0,0 +1,31 @@
...
+    protected $model = Comment::class;
+
+    public function definition(): array
+    {
+        $user = User::inRandomOrder()->first();
+
+        return [
+            'Payload' => $this->faker->paragraph,
+            'Submitted' => $this->faker->dateTimeBetween('-1 year', 'now'),
+            'Edited' => $this->faker->dateTimeBetween('now', '+1 year'),
```
wescopeland, line 25: Can we give this random odds of being set? I did 1/12 odds in another branch:

https://github.com/RetroAchievements/RAWeb/blob/master/database/factories/CommentFactory.php#L20

public/API/API_GetComments.php
```diff
@@ -0,0 +1,85 @@
...
+if ($username) {
+    $user = User::firstWhere('User', $username);
+    if (!$user || !$user->UserWallActive) {
+        return response()->json(['Count' => 0, 'Total' => 0, 'Results' => []]);
+    }
+}
+
+$articleId = $user ? $user->ID : $gameOrAchievementId;
+
+$comments = Comment::withTrashed()
```
wescopeland, line 53: If a comment has been soft deleted, does this mean the endpoint exposes the soft deleted comment to the consumer?

app/Policies/CommentPolicy.php
ioslife: In this situation, `$user` is the author of the comment

```
$results = $comments->filter(function ($nextComment) use ($policy) {
    return $policy->view($nextComment->user, $nextComment);
})->map(function ($nextComment) {
    return [
        'User' => $nextComment->user->username,
        'Submitted' => $nextComment->Submitted,
        'CommentText' => $nextComment->Payload,
    ];
});
```

public/API/API_GetComments.php
ioslife: Sure does. I have updated the queries with `whereNull('deleted_at)`

public/API/API_GetComments.php
```diff
@@ -0,0 +1,87 @@
...
+use App\Models\User;
+use App\Policies\CommentPolicy;
+use App\Support\Rules\CtypeAlnum;
+use Illuminate\Support\Facades\Validator;
+
+/*
+*  API_GetComments - returns the comments associated to a game or achievement
+*    i : game or achievement id
+*    u : username
+*    t : 1 = game, 2 = achievement, 3 = user
```
wescopeland, line 14: Doing a final review now. This took a minute to put my thoughts together on - apologies for it coming so late in the review. I promise this is the last thing.

The final piece that's nagging at me is potential for improvement w/ the DX of these query params.

I've been thinking through how I'm going to implement this endpoint in the JS client lib, and the ergonomics of the `t` query param keeps leaping off the page at me.

I'm considering the shape of one of these calls:
```
GET /API/API_GetComments.php?u=iOSLife&t=3
```
This seems quite duplicative to me. Because `iOSLife` is a string and it can only be provided via `u`, I can't help but feel the endpoint should not need to be reminded of this by means of `t=3`.

Logically, an improvement would be to make `t` optional when `u` is given. However, this also begs the question of whether `u` and `i` can thus be coalesced into a single query parameter. I think the best possible DX looks something like:

```bash
GET /API/API_GetComments.php?i=iOSLife # i! this just works
GET /API/API_GetComments.php?i=iOSLife&t=3 # this also works
GET /API/API_GetComments.php?i=15502&t=1 # this works too
# no more u param. consumers can just work with i and t.
```

Even though it's not an ID, we absorb the username into `i`. `t` can still be provided (it's safe to assume some consumers will have libraries that do this eagerly), but is optional & implied if `i` is a string. `t` is only required if an integer is given for `i`, as when this occurs, the endpoint cannot reasonably determine what resource the user is trying to fetch.

app/Policies/CommentPolicy.php
wescopeland: I think I've confused myself with this and may need help understanding the implementation details.

Is this checking if the comment author can see their own comment?

public/API/API_GetComments.php
ioslife: I like this a lot and will work on the change. Thanks for working it out and as always, never a rush on the reviews.

app/Policies/CommentPolicy.php
ioslife: Hah, yep. It seems I have also confused myself with this! It should be `return $policy->view(Auth::user(), $nextComment);`

public/API/API_GetComments.php
ioslife: Just pushed up changes to address this. Let me know your thoughts.

I also caught that banned users could call this endpoint and get _some_ results back, so I added a check to ensure the calling user is not banned. I'm not sure if we invalidate API keys once a user is banned, if we do I'll revert that.

public/API/API_GetComments.php
```diff
@@ -0,0 +1,115 @@
...
+    $username = $query['i'];
+    $commentType = 3;
+}
+
+$user = null;
+
+if ($username) {
+    $user = User::firstWhere('User', $username);
+
+    if (!$user || !$user->UserWallActive || $user->banned_at) {
```
wescopeland, line 69: This seems like logic that would be best to ultimately be reusable. Perhaps this should be driven by `UserCommentPolicy::viewAny()`.

wescopeland (commented):
This is looking a lot better. I've spent some time trying to break things and I'm not able to. I have only one (minor) suggested refactor from this final review and then I think we'll be in good shape.

public/API/API_GetComments.php
ioslife: Added `!$user->UserWallActive || $user->banned_at` to `UserCommentPolicy::viewAny()`. Let me know if the way I did it is what you were thinking.

wescopeland: Looks good - I'll do a final manual test after work today and assuming it passes will hit approve. Feel free to go ahead and make any preparations/updates in https://github.com/RetroAchievements/api-docs/pull/59 for the new request shape we landed on. Thanks for working through all my feedback.

wescopeland (approved):

PR #2556, opened 2024-07-11
chore(connect): allow Dolphin leaderboard submissions in non-prod environments

wescopeland:


wescopeland:
This is purely for internal Dolphin testing. I plan on reverting this commit after they're done poking around in stage.

PR #2562, opened 2024-07-18
chore: update message policy

wescopeland:
https://discord.com/channels/476211979464343552/1263470180236005486/1263470180236005486

### DIFF
--- a/app/Policies/MessagePolicy.php
+++ b/app/Policies/MessagePolicy.php
@@ -25,7 +25,7 @@ class MessagePolicy
 
     public function create(User $user): bool
     {
-        return $user->isNotMuted();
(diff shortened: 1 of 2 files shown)


Jamiras:
This seems like a bandage that doesn't address the underlying issue. 

A banned user shouldn't be restricted from using certain parts of the site. They shouldn't be allowed to use the site at all. I've verified that a banner user cannot log in, but a user who was logged in at the point where they were banned can still wander around the site and do stuff like this. We need to somehow update the Auth service to not validate banned accounts through whatever mechanism is keeping them logged in.

wescopeland:
Agreed, I'm going to pursue a more fundamental approach and close this PR.

PR #2563, opened 2024-07-19
feat(api): add endpoint for fetching Game Leaderboards 

ioslife:
```
/*
 *  API_GetGameLeaderboards - returns a list of Leaderboards for the give GameID
 *    i : gameID
 *    o : offset - number of entries to skip (default: 0)
 *    c : count - number of entries to return (default: 100, max: 500)
 *  int         Count                       number of want to play game records returned in the response
 *  int         Total                       number of want to play game records the user actually has overall
 *  array       Results
 *   object      [value]
 *    int        ID                         unique identifier of the leaderboard
 *    string     Title                      name of the leaderboard
 *    int        Description                details about what the leaderboard is tracking
 */
```

PR #2564, opened 2024-07-19
feat(api): adding an endpoint for fetching leaderboard entries

ioslife:
Docs PR: https://github.com/RetroAchievements/api-docs/pull/56

```
/*
 *  API_GetLeaderboardEntries - returns entries for a Leaderboard for the given ID
 *    i : leaderboardID
 *  int         Count                       number of user entries returned in the response
 *  int         Total                       number of user entries the leaderboard actually has overall
 *  array       Results
 *   object      [value]
 *    int        Rank                       user's leaderboard rank
 *    string     User                       name of user
 *    int        Score                      raw value of the leaderboard entry's score
 *    string     FormattedScore             string value of the formatted leaderboard entry's score (reference GetGameLeaderboard for Format type)
 *    string     DateSubmitted              an ISO8601 timestamp string for when the entry was submitted
 */

### CODE REVIEW

File: public/API/API_GetLeaderboardSubmissions.php [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+ *  string      Title                       name of the leaderboard
+ *  int         Description                 details about what the leaderboard is tracking
+ *  int         Count                       number of user entries returned in the response
+ *  int         Total                       number of user entries the leaderboard actually has overall
+ *  array       Entries
+ *   object      [value]
+ *    int        Rank                       user's leaderboard rank
+ *    string     User                       name of user
+ *    string     Score                      string value of the proper ValueFormat of the leaderboard entry //this feels wrong
+ *    string     DateSubmitted              an ISO8601 timestamp string for when the entry was submitted
```
wescopeland, line 17: Recommend nesting entries metadata into a unique object like:

```ts
{
  "Entries": {
    "Count": 100,
    "Total": 512,
    "Results": [ ... ]
  }
}

```

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,62 @@
...
+    ->take($count)
+    ->get()
+    ->map(function ($entry) {
+        return [
+            'ID' => $entry->id,
+            'UserID' => $entry->user_id,
+            'Score' => $entry->score,
+            'DateSubmitted' => $entry->updated_at->toDateTimeString(),
+        ];
+    });
```
wescopeland, line 54: I think this can be more optimized:

```php
// untested pseudocode

$leaderboardData = Leaderboard::select("ID", "Title", "Description")
  ->where("ID", $leaderboardId)
  ->withCount("entries")
  ->first();

$entries = LeaderboardEntry::select("id", "user_id", "score", "updated_at")
  ->where("leaderboard_id", $leaderboardId)
  ->skip($offset)
  ->take($count)
  ->get();
```

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+/*
+ *  API_GetLeaderboardSubmissions - returns a info and submissions for a Leaderboard for the given ID
```
wescopeland, line 4:
```suggestion
 *  API_GetLeaderboardSubmissions - returns info and submissions for a Leaderboard for the given ID
```

public/API/API_GetLeaderboardSubmissions.php
ioslife: Done, but not sure that `$entries = new stdClass();` is the right way?

public/API/API_GetLeaderboardSubmissions.php
ioslife: Altered your psuedocode to do what I think is what you were going for

public/API/API_GetLeaderboardSubmissions.php
wescopeland, line 1: I've taken a little while to study this and https://github.com/RetroAchievements/RAWeb/pull/2563 and I've come up with a proposal for both endpoints.

I'm leaning heavily towards leaderboards being one of the few cases where it may make sense to reuse a helper function: `GetLeaderboardData()`.

I'm feeling this way for a few reasons:
* It has built-in support for rank, using both `ValueFormat` and `LowerIsBetter`.
* It has built-in support for pagination.
* It already knows how to properly sort entries.

I don't like coupling endpoints to helpers, but I feel like if we don't this could get pretty error-prone.

I think both of these endpoints should use `GetLeaderboardData()` and return different things. For the endpoint that fetches the leaderboard metadata, a response shape like this would make sense:

```json
{
    "GameID": 12345,
    "RankAsc": false, // LowerIsBetter
    "Title": "Foo",
    "Description": "Bar",
    "Format": "SCORE",
    "TotalEntries": 12345,
    "TopEntry": {
        "User": "user display_name",
        "Score": 12345,
    }
}
```

For the endpoint in this PR:
```json
// note that i removed all leaderboard metadata here. this is an entries endpoint & is solely focused on entries.
{
    "Count": 12345,
    "Total": 12345,
    "Results": [
        {
            "User": "User",
            "DateSubmitted": "probably an ISO8601 timestamp",
            "Score": 12345,
            "Rank": 1,
        }
    ]
}
```

wescopeland:
Based on the above comment, I'm going to mark my previous feedback as resolved as it's likely not relevant anymore.

public/API/API_GetLeaderboardSubmissions.php
wescopeland: Note: A slight modification may be required in `GetLeaderboardData()` such that if `$numToFetch` is zero, it won't query for entries (line 181).

ioslife: Looking into this today. Thanks for the feedback, Wes.

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,60 @@
...
+    'o' => ['sometimes', 'integer', 'min:0', 'nullable'],
+    'c' => ['sometimes', 'integer', 'min:1', 'max:500', 'nullable'],
+]);
+
+$offset = $input['o'] ?? 0;
+$count = $input['c'] ?? 100;
+
+$leaderboardId = request()->query('i');
+
+$leaderboardData = Leaderboard::where("ID", $leaderboardId)
```
wescopeland, line 34: nit: Recommend renaming this to `$leaderboard`, so it's not ambiguous with the `GetLeaderboardData()` helper.

wescopeland, line 34: Since we're leaning in to `GetLeaderboardData()`, withCount and first can probably be dropped. Recommend using `firstWhere()`.

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,60 @@
...
+$fetchedLeaderboardData = GetLeaderboardData($leaderboardData, null, $count, $offset);
+
+$results = [];
+foreach ($fetchedLeaderboardData['Entries'] as $entry) {
+    $results[] = [
+        'User' => $entry['User'],
+        'DateSubmitted' => date('Y-m-d H:i:s', $entry['DateSubmitted']),
+        'Score' => $entry['Score'],
+        'Rank' => $entry['Rank'],
+        ];
```
wescopeland, line 51: nit: There's some formatting weirdness going on here that should probably be corrected.

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+/*
+ *  API_GetLeaderboardSubmissions - returns submissions for a Leaderboard for the given ID
```
wescopeland, line 4: nit: The more I think about it, the more I think we should rename this to API_GetLeaderboardEntries to match the nomenclature of the database and ORM model.

tests/Feature/Api/V1/LeaderboardSubmissionsTest.php
```diff
@@ -0,0 +1,187 @@
...
+            'leaderboard_id' => $leaderboard->ID,
+            'user_id' => $userFive->ID,
+        ]);
+
+        $this->get($this->apiUrl('GetLeaderboardSubmissions', ['i' => $leaderboard->ID]))
+            ->assertSuccessful()
+            ->assertJson([
+                'Count' => 5,
+                'Total' => 5,
+                'Results' => [
```
wescopeland, line 90: Using the id to check for the "Rank" value adds brittleness to the test case.

I think it would be a good idea to add at least one entry in the test case that is a tie (to get coverage that Rank is actually working correctly). In the assert, it's probably a better practice to hardcode what the rank should be for each entry.

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,60 @@
...
+ *    i : leaderboardID
+ *    o : offset - number of entries to skip (default: 0)
+ *    c : count - number of entries to return (default: 100, max: 500)
+ *  int         Count                       number of user entries returned in the response
+ *  int         Total                       number of user entries the leaderboard actually has overall
+ *  array       Results
+ *   object      [value]
+ *    int        Rank                       user's leaderboard rank
+ *    string     User                       name of user
+ *    string     Score                      string value of the proper ValueFormat of the leaderboard entry //this feels wrong
```
wescopeland, line 14: Returning the raw score would probably be even more confusing, for example on speedrun leaderboards (check game ID 1).

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,60 @@
...
+ *    string     DateSubmitted              an ISO8601 timestamp string for when the entry was submitted
```
wescopeland, line 15: `Y-m-d H:i:s` is not an ISO8601 timestamp.

public/API/API_GetLeaderboardSubmissions.php
```diff
@@ -0,0 +1,60 @@
...
+        'Rank' => $entry['Rank'],
+        ];
+}
+
+return response()->json([
+    'Count' => count($fetchedLeaderboardData['Entries']),
+    'Total' => $fetchedLeaderboardData['TotalEntries'],
+    'Results' => usort($results, function ($a, $b) {
+        return $a['Rank'] - $b['Rank'];
+    }),
```
wescopeland, line 59: I haven't tested this locally - how are ties sorted?

app/Helpers/database/leaderboard.php
```diff
@@ -174,40 +174,43 @@ function GetLeaderboardData(
...
-    $rankScore = null;
-    $userFound = false;
-    $entries = $leaderboard->sortedEntries()->with('user')->skip($offset)->take($numToFetch);
-    foreach ($entries->get() as $entry) {
-        if ($entry->score !== $rankScore) {
-            if ($rankScore === null) {
-                $rank = $leaderboard->getRank($entry->score);
-            } else {
-                $rank = $index;
+    if ($numToFetch != 0) {
```
wescopeland, line 177: Would it be possible to early return instead of having a large conditional? eg:

```php
// untested

if ($numToFetch === 0) {
    return $retVal;
}
```

public/API/API_GetLeaderboardSubmissions.php
ioslife: Yeah, I thought so too.

public/API/API_GetLeaderboardSubmissions.php
ioslife: Agreed, as I have looked at it it's felt weirder and weirder

public/API/API_GetLeaderboardSubmissions.php
ioslife: Adjusted to using date('c', $var) instead of date('Y-m-d H:i:s', $var)

public/API/API_GetLeaderboardSubmissions.php
ioslife: I've adjusted the rank sort

public/API/API_GetLeaderboardSubmissions.php
ioslife: This part is tricky. I'm not sure what the proper thing to return is here. Do you have thoughts on it?

wescopeland: Perhaps we should rename "Score" to "FormattedScore" and keep letting ValueFormat do the work.

It can be on the consumer to figure out what to do with the formatted score based on the format returned from another endpoint about the given leaderboard ID.

public/API/API_GetLeaderboardSubmissions.php
wescopeland: I'm having trouble following why the sort is necessary.

`GetLeaderboardData()` should already be returning the entries list pre-sorted, right?

```php
$userFound = false;
$entries = $leaderboard->sortedEntries()->with('user')->skip($offset)->take($numToFetch);
foreach ($entries->get() as $entry) {
```

public/API/API_GetLeaderboardSubmissions.php
ioslife: Yeah, that's good with me.

public/API/API_GetLeaderboardSubmissions.php
ioslife: Hah! That's a case of over-engineering. You are very right. Removed.

public/API/API_GetLeaderboardEntries.php
```diff
@@ -0,0 +1,56 @@
...
+}
+
+$fetchedLeaderboardData = GetLeaderboardData($leaderboard, null, $count, $offset);
+
+$results = [];
+foreach ($fetchedLeaderboardData['Entries'] as $entry) {
+    $results[] = [
+        'User' => $entry['User'],
+        'DateSubmitted' => date('c', $entry['DateSubmitted']),
+        'FormattedScore' => $entry['Score'],
```
wescopeland, line 47: We need to call `ValueFormat::format()` on these.

wescopeland:
I'm good with the response shape at this point - we'll need an accompanying docs PR.

app/Helpers/database/leaderboard.php
```diff
@@ -157,10 +157,15 @@ function GetLeaderboardData(
         }
     }
 
+    if ($numToFetch == 0) {
```
wescopeland, line 160:
```suggestion
    if ($numToFetch === 0) {
```
Similar to JS, better to use triple equals with PHP.

tests/Feature/Api/V1/LeaderboardEntriesTest.php
```diff
@@ -0,0 +1,234 @@
...
+        ]);
+
+        $this->get($this->apiUrl('GetLeaderboardEntries', ['i' => $leaderboard->ID]))
+            ->assertSuccessful()
+            ->assertJson([
+                'Count' => 5,
+                'Total' => 5,
+                'Results' => [
+                    [
+                        "Rank" => $leaderboard->getRank($leaderboardEntryFour->score),
```
wescopeland, line 112: Recommend hardcoding all of these ranks in the tests.

If the implementation details of `$leaderboard->getRank()` change, the tests will be invalid but won't fail.

public/API/API_GetLeaderboardEntries.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+use App\Models\Leaderboard;
+use App\Platform\Enums\ValueFormat;
+use App\Support\Rules\CtypeAlnum;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'i' => ['required', 'min:1', new CtypeAlnum()],
```
wescopeland, line 26:
```suggestion
    'i' => ['required', 'min:1'],
```

public/API/API_GetLeaderboardEntries.php
```diff
@@ -0,0 +1,59 @@
...
+ *    int        Rank                       user's leaderboard rank
+ *    string     User                       name of user
+ *    int        Score                      raw value of the leaderboard entry's score
+ *    string     FormattedScore             string value of the formatted leaderboard entry's score (reference GetGameLeaderboard for Format type)
+ *    string     DateSubmitted              an ISO8601 timestamp string for when the entry was submitted
+ */
+
+use App\Models\Leaderboard;
+use App\Platform\Enums\ValueFormat;
+use App\Support\Rules\CtypeAlnum;
```
wescopeland, line 21:
```suggestion
```

PR #2566, opened 2024-07-19
fix exception viewing games list sorted by ticket count when not logged in

Jamiras:
Fixes "Undefined array key \"NumTickets\"" exception when accessing a bookmark of games for a system sorted by tickets when not logged in.


wescopeland (approved):
Verified locally.

PR #2572, opened 2024-07-23
feat(manage): add support for news management in filament

wescopeland:
This PR adds support for news management into Filament, with the ultimate goal of submitnews.php being removed from a subsequent release.

There are a few additional behavior changes that are worth noting:
* The home page news carousel no longer supports HTML of any kind.
* A modest character limit has been placed on the `Payload` field in Filament.
* `Link` is a required field in Filament.

These behavior changes are all transition measures to ultimately move us away from the carousel.

**How to test image uploads in this PR:**
Minio serves as a drop-in replacement to AWS S3. Sail automatically loads a Minio instance per our docker-compose file. In your _.env_ file, set:
```env
# make sure all other MEDIA_URL lines are disabled
MEDIA_URL=http://localhost:${FORWARD_MINIO_PORT}/local

AWS_BUCKET=local
AWS_DEFAULT_REGION=eu-west-1
AWS_USE_PATH_STYLE_ENDPOINT=false
AWS_MINIO=true
```

https://github.com/user-attachments/assets/2153a25a-c509-49b6-9bb0-768bb96c3806


app/Policies/NewsPolicy.php
```diff
@@ -37,27 +39,43 @@ public function view(?User $user, News $news): bool
     public function create(User $user): bool
     {
         return $user->hasAnyRole([
+            Role::ADMINISTRATOR,
+            Role::DEVELOPER_STAFF,
+            Role::DEVELOPER,
+            Role::EVENT_MANAGER,
+            Role::MODERATOR,
             Role::NEWS_MANAGER,
+            Role::TEAM_ACCOUNT,
```
Jamiras, line 48: Do we really want to allow any developer to post news items? I think EVENT_MANAGER and NEWS_MANAGER are probably all that's needed. If an administrator, moderator, or team account wants to create news postings, they can be given the NEWS_MANAGER role.

I could maybe see allowing MODERATOR to update.

app/Policies/NewsPolicy.php
```diff
@@ -37,27 +39,43 @@ public function view(?User $user, News $news): bool
...
     public function update(User $user, News $news): bool
     {
         return $user->hasAnyRole([
+            Role::ADMINISTRATOR,
+            Role::DEVELOPER_STAFF,
+            Role::DEVELOPER,
+            Role::EVENT_MANAGER,
+            Role::MODERATOR,
             Role::NEWS_MANAGER,
+            Role::TEAM_ACCOUNT,
```
Jamiras, line 61: This should maybe be limited to NEWS_MANAGER and MODERATOR unless the news item was created by the current user.

app/Filament/Resources/NewsResource.php
```diff
@@ -0,0 +1,149 @@
...
+                                        '<p>HTML is not supported.</p>' .
+
+                                        '<p>' .
+                                            strlen($state ?? '') . '/' . $maxLength .
+                                        '</p>' .
+                                    '</div>'
+                                );
+                            })
+                            ->reactive()
+                            ->required(),
```
Jamiras, line 75: Is there some way to make this box taller? It seems to default to two lines.

resources/views/components/news/carousel-item.blade.php
```diff
@@ -22,7 +22,7 @@ class="py-6 px-8 absolute top-0 left-0 w-full h-full"
         </h4>
 
         <p class="newstext shadowoutline lg:opacity-0 transition delay-300 duration-300 subpixel-antialiased {{ $index === 0 ? '!opacity-100' : '' }}">
-            {!! $news->Payload !!}
+            {{ $news->Payload }}
```
Jamiras, line 25: > The home page news carousel no longer supports HTML of any kind.

Do we need to migrate the existing data? At a minimum, `<br>` should be converted to newline. I'm not sure what to do about links.

image

Jamiras: A lot of existing data exceeds the 238 character limit.
```
MariaDB [retroachievements]> select count(*) from News where length(Payload) > 238;
+----------+
| count(*) |
+----------+
|     1448 |
+----------+
```

app/Policies/NewsPolicy.php
wescopeland: I don't think we want to allow any developer to post news items - this is mostly to retain current functionality until the legacy page is deleted.

app/Policies/NewsPolicy.php
wescopeland: Same comment as above -- I agree, this is only in place to preserve existing functionality until the legacy page is deleted.

resources/views/components/news/carousel-item.blade.php
wescopeland: We should indeed convert `<br>` to newline - I think all the others we just let go. Currently, news items past 10 are completely inaccessible anyway. When we have a real feed, we should have similar cutoffs (just like the filament panel itself doesn't let the user see posts past a certain date).

app/Filament/Resources/NewsResource.php
wescopeland: Bumped to 4 in latest.

resources/views/components/news/carousel-item.blade.php
Jamiras: This change might be jarring for some users. The blurb in the above image effectively just becomes this:
image

The forum link is redundant with the title link, and there's nowhere to actually link to the magazine. Instead, the user has to click the title link (which I admit I never knew was clickable), and then click on the magazine link from there.

Jamiras (dismissed):

wescopeland:
@Jamiras I've just made a handful of changes in latest. These changes can be viewed at this URL: https://github.com/RetroAchievements/RAWeb/pull/2572/commits/ec5e673da9824f5302988fd1d8b7c96092b2756d.

Most of the changeset is driven by your feedback from https://github.com/RetroAchievements/RAWeb/pull/2682#discussion_r1748971337, specifically:

> Where are "uploaded" files stored? Do they automatically get cleaned up?

This is now better handled in latest. Draft files are stored in Livewire's temp storage directly, which the server automagically cleans on a routine schedule. Additionally:

* `<br>` and `<a>` tags are once again supported to better smooth out the migration from legacy management to Filament.
* Fixed a bug where news images would vanish when correcting a typo in the `Payload` in Filament.
* Restored HTML support in the carousel. I plan on once again eliminating this when news is converted to React.

Jamiras (approved):

PR #2573, opened 2024-07-24
allow extra information to be passed through report issue phone tree

Jamiras:
A new `extra` parameter has been added to the report-issue page and passed through the tickets/create page to allow the DLL to pass additional information when creating tickets. The `extra` parameter is an encoded JSON package, so we can expand it in the future.

For example, append `?extra=eyJ0cmlnZ2VyUmljaFByZXNlbmNlIjoi8J%2BQukxpbmsg8J%2BXuu%2B4j0RlYXRoIE1vdW50YWluIOKdpO%2B4jzMvMyDwn5GlMS80IPCfp78wLzQg8J%2BRuzAvNjAg8J%2BQnDAvMjQg8J%2BSgDUg8J%2BVmTEyOjAwIEFN8J%2BMmSJ9` to the `/achievement/N/report-issue` url.

When the ticket is created, it'll include:
```
Rich Presence at time of trigger:
🐺Link 🗺️Death Mountain ❤️3/3 👥1/4 🧿0/4 👻0/60 🐜0/24 💀5 🕙12:00 AM🌙
```


wescopeland (approved):
Verified locally. Also confirmed that if invalid `extra` input is given, `json_decode(base64_decode($input['extra']))` is smart enough to return `null`.

PR #2582, opened 2024-07-29
fix: don't record hash analytics for multi-disc game systems

wescopeland:
Intended to address some feedback from @Jamiras regarding #2537.

For systems that support games with multiple discs/disks, each one has a unique hash. Swapping discs/disks will create a new session. For some of these systems, such as the Apple II, this can happen very frequently.

This PR adjusts session hash recording to only include systems which do not have games with disc/disk swapping in their libraries.


wescopeland:
My only additional question on this: would it be OK to limit this logic to just Apple II and DOS? I'm wondering if we may be losing some good data on PSX, PS2, and GC with how broad the net on this PR is.

app/Models/System.php
```diff
@@ -79,12 +79,36 @@ protected static function newFactory(): SystemFactory
...
+        return [
+            System::PlayStation,
+            System::GameCube,
+            System::Wii,
+            System::WiiU,
+            System::PlayStation2,
+            System::Xbox,
+            System::AppleII,
+            System::SegaSaturn,
+            System::Dreamcast,
```
Jamiras, line 108: rcheevos also supports m3u for CPC, C64, MSX, PC88, 3DO, PCE-CD, PCFX, and SegaCD.

Recommend sorting this alphabetically to make it easier to look at.

app/Platform/Actions/ResumePlayerSession.php
```diff
@@ -104,7 +107,7 @@ public function execute(
         $playerSession = new PlayerSession([
             'user_id' => $user->id,
             'game_id' => $game->id,
-            'game_hash_id' => $gameHash?->id,
+            'game_hash_id' => $gameHash && !$isForMultiDiscGamesSystem ? $gameHash->id : null,
```
Jamiras, line 110: It's probably still useful to capture the first hash loaded.

Jamiras:
> My only additional question on this: would it be OK to limit this logic to just Apple II and DOS? I'm wondering if we may be losing some good data on PSX, PS2, and GC with how broad the net on this PR is.

This is just a stop-gap until we have game_hash_sets. A multi-disc session should be associated to a hash set, instead of a specific hash. Though we're probably still a ways away from that.

My concern with allowing PSX/PS2/GC through has to do with how the games are booted from the emulator. If launching the game from the emulator always loads the first disc, there would still be a session created for that first disc for whatever period of time it takes for the player to load their save game and get prompted to insert the second disc.

Maybe we could restrict the functionality to ignore hashes containing "Disk" or "Disc" as part of their description instead of relying on the system?

wescopeland:
> Maybe we could restrict the functionality to ignore hashes containing "Disk" or "Disc" as part of their description instead of relying on the system?

I really like this idea. After doing a cursory check of several games/hashes in the database, I think this would get us pretty decent coverage.

I've ripped the previous changes out of the `System` model and added a function to the `GameHash` model:
```php
public function isMultiDiscGameHash(): bool
{
    return Str::contains(Str::lower($this->name), ['disk', 'disc']);
}
```

Now all the conditionals are driven by this logic rather than a check on the game system.

app/Platform/Actions/ResumePlayerSession.php
wescopeland: Adjusted in latest.

app/Models/System.php
wescopeland: Ripped out per https://github.com/RetroAchievements/RAWeb/pull/2582#issuecomment-2257082062.

app/Models/GameHash.php
```diff
@@ -128,6 +129,16 @@ public function addRegion(string $region): self
...
     // == accessors
 
+    /**
+     * Check if the game hash is related to a multi-disc game.
+     * These hashes nearly always have the word "disk" or "disc" in their label.
+     * This is temporary until we fully support game_hash_sets.
+     */
+    public function isMultiDiscGameHash(): bool
+    {
+        return Str::contains(Str::lower($this->name), ['disk', 'disc']);
```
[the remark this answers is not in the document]
wescopeland: Good call - updated in latest.

Jamiras (approved):

PR #2583, opened 2024-07-31
migrate top achievers widget to blade

Jamiras:
https://discord.com/channels/310192285306454017/1098073710394621994/1098073710394621994

Replaces the "High Scores" widget on the game page with a dynamic widget that shows "Most Points Earned" if there's less than 10 masteries, or "Latest Masters" if there's 10 or more.

Before:
image

After (less than 10 masteries):
image

After (10 or more masteries):
image

Where the "more..." link takes the user to the a paginated list of all users who have mastered the game.

No changes were made to accommodate pre-revision masteries. Masteries are still calculated as having the maximum number of hardcore points available for the set.

Games with 0 points (events) now show the number of unlocked achievements if at least 10 people haven't unlocked all achievements in the set. Here's a preview of "DevQuest 15 - Console Conquerer":

image


resources/views/components/game/top-achievers/latest-masters.blade.php
wescopeland, line 1: Recommend right-aligning this column. The lack of right-alignment seems to create an illusion of inconsistent padding on each row. The effect is especially pronounced at the LG breakpoint in particular. Here's a before and after on XL:

**Before**
Screenshot 2024-08-02 at 5 33 24 PM


**After**
Screenshot 2024-08-02 at 5 33 35 PM

This change may wind up looking kinda weird on the dedicated "masters" page. Maybe it's something we only do on game pages.

app/Platform/Services/GameTopAchieversService.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    private function convertPlayerGames(Collection $playerGames): array
+    {
+        $retval = [];
+        foreach ($playerGames as $playerGame) {
+            $retval[] = [
+                'user_id' => $playerGame->user_id,
+                'achievements_unlocked_hardcore' => $playerGame->achievements_unlocked_hardcore,
+                'points_hardcore' => $playerGame->points_hardcore,
+                'last_unlock_hardcore_at' => $playerGame->last_unlock_hardcore_at->unix(),
```
wescopeland, line 144: I seem to be getting an error on this line when I open a game that has only softcore unlocks and no hardcore unlocks. This does seem to imply the UI needs to support some kind of empty state (or just 404 the dedicated page and don't render the dedicated component on the game page).

Screenshot 2024-08-02 at 5 35 48 PM

resources/views/components/game/top-achievers/score-row.blade.php
```diff
@@ -0,0 +1,19 @@
...
+@else
+    <tr>
+@endif
+
+        <td class="text-right">{{ $rank }}</td>
+        <td class="whitespace-nowrap">
+            <x-user.avatar display="icon" icon-size="xs" :user="$user" />
+            <x-user.avatar hasHref="true" :user="$user" />
+        </td>
+        <td class="text-right">{{ $score }}</td>
```
wescopeland, line 18: We should probably use `localized_number()` on this value.

resources/views/components/game/top-achievers/mastery-row.blade.php
```diff
@@ -0,0 +1,20 @@
...
+@if ($masteryUser->id === request()->user()->id)
+    <tr style='outline: thin solid'>
+@else
+    <tr>
+@endif
+
+        <td class="text-right">{{ $rank }}</td>
+        <td class="whitespace-nowrap">
+            <x-user.avatar display="icon" :icon-size="$iconSize" :user="$masteryUser" />
+            <x-user.avatar hasHref="true" :user="$masteryUser" />
```
wescopeland, line 17: Tooltips don't seem to be working for these avatar components.

resources/views/components/game/top-achievers/latest-masters.blade.php
wescopeland: Per your screenshot here:
353648658-b8446e45-4522-4543-a16d-6984de4571e6

I also can't help but wonder if visually on the game page it'd be better to exclude the time in order to help maintain a single line of text for visual consistency. Perhaps it can be expressed either as a tooltip, or only available on the dedicated page.

wescopeland: Looking at this again with a fresh set of eyes, I actually don't mind it too much. I think the only thing that should be looked into here is possibly removing the times from the game page rows.

resources/views/components/game/top-achievers/mastery-row.blade.php
Jamiras: It looks like the `<x-avatar>` component isn't fully fleshed out yet: https://github.com/RetroAchievements/RAWeb/blob/c5d15e85961be6bbcb06a5d35a9cb08f5e62d3ce/resources/views/components/avatar.blade.php#L22-L34

Replaced with `userAvatar` function call.

resources/views/components/game/top-achievers/score-row.blade.php
Jamiras: done

app/Platform/Services/GameTopAchieversService.php
Jamiras: This also happens if there aren't enough hardcore players to fill the list. I've fixed the error and added a default message if the list is empty.

resources/views/components/game/top-achievers/most-points.blade.php
```diff
@@ -0,0 +1,53 @@
...
+use App\Models\User;
+
+@endphp
+
+<div class="component">
+    <h2 class="text-h3">Most Points Earned</h2>
+
+    <div class="max-h-[980px] overflow-y-auto">
+        @if (empty($highestPointEarners))
+        <tr colspan="3">No points have been earned yet.</tr>
```
wescopeland, line 17: This is inside of a table row, but outside of a table -- technically invalid HTML. Might be best to change this to a `<p>` tag?

wescopeland (commented):
Only one thing left I'm noticing and then I think this will be good to go.

wescopeland (approved):

resources/views/components/game/top-achievers/most-points.blade.php
Jamiras: I had it in the table first, but then moved it out. Fixed.

PR #2588, opened 2024-08-02
feat(api): use released_at and released_at_granularity on API_GetGame

wescopeland:
Will update other endpoint(s) in a separate PR.

### DIFF
--- a/public/API/API_GetGame.php
+++ b/public/API/API_GetGame.php
@@ -19,7 +19,8 @@
  *  string     Publisher                  publisher information for the game
  *  string     Developer                  developer information for the game
  *  string     Genre                      genre information for the game
- *  string     Released                   release date information for the game
+ *  string?    Released                   an ISO8601 timestamp of the game's earliest release date, or null. also see ReleasedAtGranularity.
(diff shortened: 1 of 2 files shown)


public/API/API_GetGame.php
```diff
@@ -48,5 +49,6 @@
     'Publisher' => $game->Publisher,
     'Developer' => $game->Developer,
     'Genre' => $game->Genre,
-    'Released' => $game->Released,
+    'Released' => $game->released_at?->format('Y-m-d H:i:s'),
```
Jamiras, line 52: Why include hours/minutes/seconds if the granularity doesn't go that far? It looks like the time portion comes from when the released_at was last updated:
```
"Released":"1985-09-13 21:29:49","ReleasedAtGranularity":"day"
```

```
+---------------------+---------------------+
| released_at         | updated             |
+---------------------+---------------------+
| 1985-09-13 21:29:49 | 2024-08-15 21:29:49 |
+---------------------+---------------------+
```

If it's desirable to keep a full timestamp, suggesting forcing to midnight.

wescopeland: Good point -- set to YYYY-MM-DD in latest.

Jamiras (approved):

PR #2589, opened 2024-08-02
feat(api): use released_at and released_at_granularity on extended game endpoints

wescopeland:
Sets `Released` to use `released_at` and adds `ReleasedAtGranularity` to API_GetGameExtended and API_GetGameInfoAndUserProgress.


public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -95,8 +99,15 @@
...
-unset($gameData['system']);
+// Don't expose these values.
+// TODO stop using getGameMetadata helper
 unset($gameData['achievement_set_version_hash']);
+unset($gameData['achievements_published']);
+unset($gameData['players_total']);
+unset($gameData['points_total']);
+unset($gameData['released_at_granularity']);
+unset($gameData['released_at']);
+unset($gameData['system']);
```
wescopeland, line 110: This is probably a breaking change, but these values are undocumented and shouldn't be exposed.

public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -44,8 +44,8 @@
  *  string     ImageBoxArt                site-relative path to the game's box art image
  *  string     Publisher                  publisher information for the game
  *  string     Developer                  developer information for the game
- *  string     Genre                      genre information for the game
```
ioslife, line 47: Don't think you meant to remove Genre

wescopeland: Fixed in latest. Thanks.

public/API/API_GetGameExtended.php
```diff
@@ -81,7 +82,8 @@
     'Publisher' => $game->Publisher,
     'Developer' => $game->Developer,
     'Genre' => $game->Genre,
-    'Released' => $game->Released,
+    'Released' => $game->released_at?->format('Y-m-d H:i:s'),
```
ioslife, line 85: Looking at the tests below, the format for `Released` used to be: `'M Y'`. This is _technically_ also a breaking change.

wescopeland: That was the format in the tests, but not in real life. Up until now, the field has been completely free-form text. Here's a few examples that are now hidden as of the latest release:

```
109,Oktober 1987
183,"Novermber 15, 1996 (UK)"
524,Never
542,Physically
626,"JP:July 16, 1993, NA:October 1993, EU:1993"
874,"JP:November 27, 1992 NA:December 1, 1993"
910,Nov 15th 2021 - Dec 15th 2021
```

ioslife: Cool, then this is much better.

Jamiras, line 85: see comment about including hour/minute/second: https://github.com/RetroAchievements/RAWeb/pull/2588#discussion_r1719026370

wescopeland: Agreed. Set to YYYY-MM-DD in latest.

Jamiras (approved):

PR #2591, opened 2024-08-04
chore: remove z param from the API

ioslife:
After recognizing that the username (z) param in the API calls doesn't affect anything, I started investigating why we weren't ensuring that the provided API key (y param) matches `User->APIKey`. I thought this to be a bug, but it appears that Laravel doesn't need the username.

It looks like the way that the auth guards are designed (by Laravel, not RA), providing an API token is all you need. https://laravel.com/docs/5.8/api-authentication

From what I am reading online, there doesn't seem to be any reason to include username and API token. From what I can tell, we aren't even ingesting the z= value anywhere in our API for auth purposes.

AuthServiceProvider.php seems like it used to with an override, but even that isn't set up right (it is also commented out).

TL;DR –  the z param on API requests seems to be unnecessary. The API token should be the sole authentication method.


app/Platform/AuthServiceProvider.php
```diff
@@ -5,15 +5,12 @@
...
+use Illuminate\Support\Facades\Auth;
 
 class AuthServiceProvider extends ServiceProvider
 {
     public function boot(): void
     {
-        // Auth::extend('api-token', function ($app, $name, array $config) {
-        //     return new TokenGuard(Auth::createUserProvider($config['provider']), $app->request, 'z', 'api_token');
-        // });
-        //
```
ioslife, line 16: I removed this commented out code. Removing the z param makes it to where even what this was doing before being commented out is pointless.

public/API/API_GetAchievementDistribution.php
```diff
@@ -16,7 +16,7 @@
 
 $gameID = (int) request()->query('i');
 $hardcore = (int) request()->query('h', (string) UnlockMode::Softcore);
-$requestedBy = request()->query('z');
+$requestedBy = request()->query('z'); // the z (username) param has been deprecated, but to ensure we don't introduce a breaking change we will keep this here
```
ioslife, line 19: `getAchievementDistribution()` takes a `$requestedBy` param, so I left this reference the the z param here in case people continue to call with the z param.

Now that I think about it though, this might be pointless. We could just do `$requested = Auth::user()->User` because as it stands now, the z param will _always*_ be present.

*unless someone is currently calling this endpoint with an empty username, which is possible. The endpoint wouldn't return expected results for them though.

wescopeland: I think doing `Auth::user()` is likely preferable to having consumers tack on the `z` query param.

app/Platform/AuthServiceProvider.php
wescopeland: Do we need the new `Auth` import on line 8?

ioslife: Nope, residual from me testing this comment.

wescopeland (approved):

app/Exceptions/Handler.php
```diff
@@ -72,7 +72,6 @@ protected function buildExceptionContext(Throwable $e): array
                     unset($params['p']);
                 }
             } elseif (str_contains($context['url'], '/API/')) {
-                unset($params['z']);
```
Jamiras, line 75: We should probably still exclude this from the exception log, as consumers will likely continue to pass it for quite some time.

ioslife: Yeah, that's a good point. Will revert.

Jamiras (approved):

ioslife:
There was a merge conflict here in a file I didn't touch, but seems to be resolved now

PR #2594, opened 2024-08-04
fix: handle PlayerAchievementLocked events in DispatchUpdateDeveloperContributionYieldJob

wescopeland:
Reverts the changes from https://github.com/RetroAchievements/RAWeb/pull/2545 and implements a fix for the root cause of the issue.

`DispatchUpdateDeveloperContributionYieldJob` is not handling the `PlayerAchievementLocked` case. This results in the contrib yield job being called with a null user.


Jamiras:
> `DispatchUpdateDeveloperContributionYieldJob` is not handling the `PlayerAchievementLocked` case. This results in the contrib yield job being called with a null user.

It seems like this code is/was trying to handle the null user, but there might be a order of operations issue in the logic:
```
        if (!$user instanceof User) {
            return;
        }
```

wescopeland:
Figured it out.

`findOrFail()` explodes if it's querying for a soft-deleted user. Checking for `PlayerAchievementLocked` turns out to not be necessary - I don't believe this is currently being dispatched anywhere.

Should be addressed in latest.

Jamiras (approved):
Did you have any insight onto whether `(!$user instanceof User)` is logically correct? Does it need to be `(!($user instanceof User))`?

wescopeland:
It's logically correct as-is. [`instanceof` has a higher operator precedence than logical not](https://www.php.net/manual/en/language.operators.precedence.php).

PR #2595, opened 2024-08-05
feat(manage): improve breadcrumbs usability

wescopeland:
This PR slightly alerts the management app's achievement and leaderboard page breadcrumbs so the game is always in the list of links.

**Before**
Screenshot 2024-08-05 at 7 05 46 PM

**After**
Screenshot 2024-08-05 at 7 07 20 PM

### DIFF
--- a/app/Filament/Resources/LeaderboardResource/Pages/Details.php
+++ b/app/Filament/Resources/LeaderboardResource/Pages/Details.php
@@ -23,7 +23,7 @@ class Details extends ViewRecord
             route('filament.admin.resources.leaderboards.index') => 'Leaderboards',
             route('filament.admin.resources.games.view', $game) => $game->title,
             route('filament.admin.resources.leaderboards.view', $leaderboard) => $leaderboard->title,
-            'Audit Log',
+            'View',
         ];
     }
 
--- a/app/Filament/Resources/LeaderboardResource/Pages/Details.php
+++ b/app/Filament/Resources/LeaderboardResource/Pages/Details.php
@@ -12,6 +12,20 @@ class Details extends ViewRecord
 {
     protected static string $resource = LeaderboardResource::class;
 
+    public function getBreadcrumbs(): array
+    {
+        /** @var Leaderboard $leaderboard */
+        $leaderboard = $this->record;
+        $game = $leaderboard->game;
+
+        return [
+            route('filament.admin.resources.leaderboards.index') => 'Leaderboards',
+            route('filament.admin.resources.games.view', $game) => $game->title,
+            route('filament.admin.resources.leaderboards.view', $leaderboard) => $leaderboard->title,
+            'Audit Log',
+        ];
+    }
+
     protected function getHeaderActions(): array
     {
         return [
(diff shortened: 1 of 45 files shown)


app/Filament/Resources/LeaderboardResource/Pages/Details.php
```diff
@@ -5,13 +5,28 @@
...
+    {
+        /** @var Leaderboard $leaderboard */
+        $leaderboard = $this->record;
+        $game = $leaderboard->game;
+
+        return [
+            route('filament.admin.resources.leaderboards.index') => 'Leaderboards',
+            route('filament.admin.resources.games.view', $game) => $game->title,
+            route('filament.admin.resources.leaderboards.view', $leaderboard) => $leaderboard->title,
+            'Audit Log',
```
Jamiras, line 26: This is not the audit log

app/Filament/Resources/LeaderboardResource/Pages/Edit.php
```diff
@@ -5,13 +5,28 @@
...
+    {
+        /** @var Leaderboard $leaderboard */
+        $leaderboard = $this->record;
+        $game = $leaderboard->game;
+
+        return [
+            route('filament.admin.resources.leaderboards.index') => 'Leaderboards',
+            route('filament.admin.resources.games.view', $game) => $game->title,
+            route('filament.admin.resources.leaderboards.view', $leaderboard) => $leaderboard->title,
+            'Audit Log',
```
Jamiras, line 26: This is not the audit log

Jamiras (commented):
Somewhat related: the leaderboard link on the game page goes directly to the edit sub-page.
image

wescopeland:
> Somewhat related: the leaderboard link on the game page goes directly to the edit sub-page.

This is a default Filament behavior for users who have the ability to edit the resource. If the user doesn't have the ability to edit the resource, they'll instead be linked to the resource's "Details" (view) page.

With that in mind, we should probably ensure both pages follow a similar layout & show the same fields. I'll make a note to circle back on this in a follow-up.

app/Filament/Resources/LeaderboardResource/Pages/Details.php
wescopeland: Thanks, fixed in latest.

app/Filament/Resources/LeaderboardResource/Pages/Edit.php
wescopeland: Thanks, fixed in latest.

Jamiras:
> This is a default Filament behavior for users who have the ability to edit the resource. If the user doesn't have the ability to edit the resource, they'll instead be linked to the resource's "Details" (view) page.

Odd. The achievement links weren't directed to the edit page for the achievement, and I had permission to edit those too.

wescopeland:
Ah, that is because in _AchievementsRelationManager.php_, for some reason I put in an override:
```php
->recordUrl(
    fn (Achievement $record): string => route('filament.admin.resources.achievements.view', ['record' => $record])
)
```

Now that you mention it, this should probably be removed as it creates an inconsistent user experience.

wescopeland:
Although, with that override removed, the table rows don't link anywhere at all. That must explain why I added it.

wescopeland:
I've made a change to those achievement rows in https://github.com/RetroAchievements/RAWeb/pull/2595/commits/2c80aa48c80364f50062ccc92ae34e62ff136daa so we can have a consistent behavior. The rows now do a policy check to determine what the href should be.

Jamiras (approved):

PR #2598, opened 2024-08-09
perf: squash all migrations

wescopeland:
A prep step for upgrading to Laravel 11. ([ref](https://laravel.com/docs/11.x/upgrade#modifying-columns))

> When modifying a column, you must now explicitly include all the modifiers you want to keep on the column definition after it is changed. Any missing attributes will be dropped ... In Laravel 10, this migration would retain the unsigned, default, and comment attributes on the column. However, in Laravel 11, the migration must now also include all of the attributes that were previously defined on the column. Otherwise, they will be dropped.

> If you do not want to update all of the existing "change" migrations in your application to retain the column's existing attributes, you may simply [squash your migrations](https://laravel.com/docs/11.x/migrations#squashing-migrations).

I've confirmed that migrating a fresh database still works. All migrations in _platform_, _community_ and the migrations root folders have been deleted. I left _upcoming_ alone. In the _platform_ and _community_ folders, I added _.gitkeep_ files so the empty folders wouldn't get blown away by git.

By squashing migrations, we also manage to bring down non-parallel test execution time by roughly 33%. On my machine it decreases from 180s to 120s. Parallel execution time decreases from 30s to 21s.

Additionally, [I've added Pest](https://discord.com/channels/476211979464343552/1026595325038833725/1153294768802107452) as a dependency. Pest seems to magically take over test execution when installed -- I found during local benchmarks that this also made a ~5% improvement in test execution time.

Performance improvements are noticeable in CI as well.

Once we are on Laravel 11, we should get [another significant improvement](https://laravel.com/docs/11.x/releases#database-performance) in sqlite testing performance.

**Before**
Screenshot 2024-08-09 at 10 24 57 AM

**After**
Screenshot 2024-08-09 at 10 25 27 AM


ioslife:
Let's gooo. That time improvement is huge

wescopeland:
@Jamiras One thing I'm curious about when you review this PR: are you able to run tests in parallel mode?

`composer test -- --parallel`

I believe Pest uses a different [parallel testing](https://pestphp.com/docs/optimizing-tests) engine than PHPUnit does. I am not sure how a Windows machine will react when parallel testing mode is used, but if things are fine, we may be able to switch the pre-push hook to use parallel testing.

Jamiras:
> I am not sure how a Windows machine will react when parallel testing mode is used...

I still run RAWeb in a Ubuntu VM within Windows, so I doubt I can provide the information you're looking for.

database/schema/sqlite-schema.sql
```diff
@@ -0,0 +1,301 @@
...
+INSERT INTO migrations VALUES(93,'2024_05_19_000000_update_game_hashes_table',1);
+INSERT INTO migrations VALUES(94,'2024_05_20_000000_update_ticket_table',1);
+INSERT INTO migrations VALUES(95,'2024_05_21_000000_update_forumtopiccomment_table',1);
+INSERT INTO migrations VALUES(96,'2024_05_21_000000_update_leaderboarddef_table',1);
+INSERT INTO migrations VALUES(97,'2024_05_21_000000_update_setclaim_table',1);
+INSERT INTO migrations VALUES(98,'2024_05_21_000000_update_siteawards_table',1);
+INSERT INTO migrations VALUES(99,'2024_05_25_000001_update_player_games_table',1);
+INSERT INTO migrations VALUES(100,'2024_06_08_000000_drop_websockets_statistics_entries_table',1);
+INSERT INTO migrations VALUES(101,'2024_06_18_000000_update_gamedata_table',1);
+INSERT INTO migrations VALUES(102,'2024_07_03_000000_update_codenotes_table',1);
```
Jamiras, line 301: Is there any benefit to inserting these migrations? The sqllite DB shouldn't ever be upgrade or downgraded.

I'm also curious about their necessity for mysql. If these files no longer exist, why track them?

From reading the documentation, it seems like the squash just fast-forwards over the migrations instead of reapplying them all. I don't believe it's intended that the migrations themselves be deleted from the repository. What if someone still has a 6-month old DB that needs migrated?

Jamiras (commented):
While I did see a significant performance increase using the squashed migrations, I didn't see much of a difference moving to pest.

`php artisan test` (master): 157.34s
`php artisan test` (after): 102.13s
`vendor/bin/pest`: 100.75s

`php artisan test --parallel` (master): 43.47s
`php artisan test --parallel` (after): 29.85s
`vendor/bin/pest --parallel`: 27.49s

I've also confirmed that `sail migrated:fresh --seed` generates a workable site.

database/schema/sqlite-schema.sql
wescopeland: Both of these SQL files are auto-generated by Laravel itself using `php artisan schema:dump`.

>I don't believe it's intended that the migrations themselves be deleted from the repository. What if someone still has a 6-month old DB that needs migrated?

The challenge with this is Laravel 11's breaking change to migrations  ([see here](https://laravel.com/docs/11.x/upgrade#modifying-columns)):
> If you do not want to update all of the existing "change" migrations in your application to retain the column's existing attributes, you may simply [squash your migrations](https://laravel.com/docs/11.x/migrations#squashing-migrations).

My interpretation of this is after we migrate to Laravel 11, the functionality of those migrations isn't going to be the same anyway, and we're better off squashing them all to avoid issues. If we don't do this, the logic of the migrations themselves could potentially change as a result of this breaking change.

If someone has an unmigrated database from 6 months ago, they have a lot more to worry about beyond just migrations. We've remediated tons of data integrity issues over the last 6 months - that person would be better off starting from a fresh DB, in my opinion.

Jamiras: > Now, when you attempt to migrate your database and no other migrations have been executed, **Laravel will first execute the SQL statements in the schema file of the database connection you are using**. After executing the schema file's SQL statements, **Laravel will execute any remaining migrations that were not part of the schema dump**.

The way I read it, the schema dump serves as a pre-migrated database to which new migrations will be applied. I don't see anything in the documentation that says to delete the old migrations.

I think I understand the problem though - the migrations may not behave correctly because of the breaking change, so applying them may not work as expected. In that case, I guess it's reasonable to destroy the old files. I'm still uncertain whether there's any benefit in keeping the old migration trail though. Having them there implies the user could still do a `sail artisan migrate:rollback --steps=102`. Removing them from my local file shaved another 0.75 seconds off the parallel run time (though for a sample size of one run that might just be noise).

After we're on Laravel 11, it would make sense to periodically update the schema dumps to avoid running any _new_ migrations added, but those new migrations should not be deleted.

wescopeland: >I think I understand the problem though - the migrations may not behave correctly because of the breaking change, so applying them may not work as expected.

Yes, this is the angle I'm seeing things from. I'm nervous keeping the old migrations around could be harmful because they no longer behave as they did when they were written, so even as a historical archive they are inaccurate.

Rather than destroying the migrations, perhaps as a happy medium they could be moved to a non-executed "old", "archive", or "legacy" folder?

Jamiras: Technically, people should be using the seeded database, rather than a dump, and I've confirmed that that is working with the raw schema files, so I think the impact of destroying the legacy files is minimal. Users could always go dig them up from github if they need them (even do so conveniently from the PR). They could also just checkout a commit prior to this PR and run the migration directly.

I think removing the files is the correct solution to ensure they don't cause unintended behavior going forward, but we should also remove the references to the files from the imports (both for performance and data integrity). Just delete the last 102 INSERTs from both files, and call it good.

Any migrations added after this PR (even if they're merged first) should remain to ensure the changes get applied to the captured schema files. And they should remain even if the schema files get updated in a future migration squash, assuming they're laravel 11 equivalent (which should be a requirement for future merges even if we haven't made the laravel 11 switch).

For production, I think it's reasonable to keep the legacy migration records even if they can't be rolled back. It's not like we'd do any rollbacks in production anyway.

wescopeland: In latest, the 102 INSERTs have been removed from both files. I've also stripped the Pest dependencies.

database/schema/sqlite-schema.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE IF NOT EXISTS "migrations" ("id" integer primary key autoincrement not null, "migration" varchar not null, "batch" integer not null);
 CREATE TABLE IF NOT EXISTS "Console" ("ID" integer primary key autoincrement not null, "Name" varchar not null, "Created" datetime, "Updated" datetime, "name_full" varchar, "name_short" varchar, "manufacturer" varchar, "active" tinyint(1), "order_column" integer, "deleted_at" datetime);
-CREATE TABLE IF NOT EXISTS "EmailConfirmations" ("User" varchar not null, "EmailCookie" varchar not null, "Expires" date not null);
+CREATE TABLE IF NOT EXISTS "EmailConfirmations" ("id" integer primary key autoincrement not null, "User" varchar not null, "EmailCookie" varchar not null, "Expires" date not null);
```
Jamiras, line 3: Was [this change](https://github.com/RetroAchievements/RAWeb/pull/2598/commits/aa4375b327085d0a78bf11b09cd1a47376e899c5#diff-3c4872576f46a8f07b1ada7f70f48973ed38eb1730c678c98ad16f5bf141288eR3) intentional (addition of ID column)? It doesn't appear in the mysql-schema file.

database/schema/mysql-schema.sql
```diff
@@ -0,0 +1,1588 @@
...
+/*!40101 SET character_set_client = @saved_cs_client */;
+DROP TABLE IF EXISTS `EmailConfirmations`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!50503 SET character_set_client = utf8mb4 */;
+CREATE TABLE `EmailConfirmations` (
+  `User` varchar(20) NOT NULL,
+  `user_id` bigint(20) unsigned DEFAULT NULL,
+  `EmailCookie` varchar(20) NOT NULL,
+  `Expires` date NOT NULL,
+  PRIMARY KEY (`id`),
```
Jamiras, line 94: You [explicitly removed](https://github.com/RetroAchievements/RAWeb/pull/2598/commits/fb630ee5ffa7d4aff07722d9968514594f09e7e1#diff-9f611797ecac538e43ae8149ee92ced6b58c8587375deb6f505d395312288643L90) the `id` column from this table, but there's still a key reference.

Which I assume is responsible for this error doing a `migrate:fresh`:
```
Error Output:
================
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1072 (42000) at line 89: Key column 'id' doesn't exist in table
```

wescopeland:
Sorry @Jamiras, I think I managed to confuse myself very much here.

I think the issues should be addressed now though. My intention was to also squash _2024_08_07_000000_update_emailconfirmations_table.php_. I've confirmed migrating a fresh database & tests are now all working again.

Jamiras (approved):

PR #2599, opened 2024-08-09
perf(game): resolve N+1 leaderboards query problem

wescopeland:
This PR significantly improves the performance of game pages with dozens / hundreds of leaderboards.

Currently, the leaderboard listing component is querying for the top entry on each individual leaderboard listed (an N+1 query problem). Now, the top scores are queried for at the root in a single pass.

Screenshots below are taken from game ID 28555.

**Before**
Screenshot 2024-08-09 at 4 09 54 PM

**After**
Screenshot 2024-08-09 at 4 09 41 PM


resources/views/components/game/leaderboards-listing/index.blade.php
```diff
@@ -8,8 +8,13 @@
...
-            :gameLeaderboards="$game->visibleLeaderboards->sortBy('DisplayOrder')"
-        />
+        @php
+            $gameLeaderboards = $game->visibleLeaderboards()
+                ->withTopEntry()
+                ->orderBy('DisplayOrder')
+                ->get();
+        @endphp
+
+        <x-game.leaderboards-listing.leaderboards-list :$gameLeaderboards />
```
Jamiras, line 18: This seems to work, but shouldn't it be `:gameLeaderboards="$gameLeaderboards"`?

resources/views/components/game/leaderboards-listing/leaderboards-list-item.blade.php
```diff
@@ -17,16 +17,13 @@ class="leading-3"
...
-        @php
-            $bestScore = $leaderboard->sortedEntries()->first();
-        @endphp
-        @if (!$bestScore)
+        @if (!$leaderboard->topEntry)
             <p>No entries.</p>
         @else
             <div class="flex justify-between">
-                {!! userAvatar($bestScore->user, iconSize: 16, iconClass: 'rounded-sm') !!}
+                {!! userAvatar($leaderboard->topEntry->user->username, iconSize: 16, iconClass: 'rounded-sm') !!}
```
Jamiras, line 24: You can pass the User model directly to userAvatar to avoid requerying it.

resources/views/components/game/leaderboards-listing/index.blade.php
wescopeland: `:$variableName` is Blade shorthand for `:variableName="$variableName"`. I only learned about this feature very recently.

resources/views/components/game/leaderboards-listing/leaderboards-list-item.blade.php
wescopeland: I keep forgetting that is a thing. Fixed in latest.

Jamiras (approved):

PR #2605, opened 2024-08-11
fix(game): add achievement filter query param to open tickets button

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1002693810037477377/1272131498719449159.


resources/views/components/game/link-buttons/index.blade.php
```diff
@@ -97,9 +97,16 @@
     @endif
 
     @if ($canSeeOpenTickets)
+        @php
+            $href = route('game.tickets', ['game' => $game, 'filter[achievement]' => 'core']);
+            if (!$isViewingOfficial) {
+                $href = route('game.tickets', ['game' => $game, 'filter[achievement]' => 'unofficial']);
+            }
```
Jamiras, line 104: Seems like this should be an if/else instead of calculating the core route, then possibly overwriting it.

However, the overhead is probably minimal, and the unofficial page isn't commonly used, so I'm not opposed to leaving it this way.

Jamiras (approved):

resources/views/components/game/link-buttons/index.blade.php
wescopeland: Went ahead and adjusted this in https://github.com/RetroAchievements/RAWeb/pull/2605/commits/0385486465e3c912c5f7597d6f15f76de00245ef.

Jamiras (approved):

PR #2607, opened 2024-08-11
feat(manage): make some slight ux adjustments to game achievements table

wescopeland:
Some minor UX adjustments here based on user feedback:

* Achievement descriptions now wrap the table row if they are very long.
* Rather than show both published and unpublished achievements simultaneously, by default only published are shown. A new filter has been added which lets the user select published or unpublished.

**Before**
Screenshot 2024-08-11 at 6 48 45 PM

**After**
Screenshot 2024-08-11 at 6 48 36 PM


app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -107,7 +98,13 @@ public function table(Table $table): Table
...
             ])
             ->filters([
-
+                Filters\SelectFilter::make('Flags')
+                    ->options([
+                        AchievementFlag::OfficialCore => 'Published',
+                        AchievementFlag::Unofficial => 'Unpublished',
+                    ])
+                    ->default(AchievementFlag::OfficialCore)
+                    ->selectablePlaceholder(false),
```
Jamiras, line 107: If you remove the Published filter, there's no way to reselect it as it's the default selection in the dropdown:
image

You can click Reset to reselect it, but that would also reset any other future filters we add.

You can also select Unpublished, then switch back to Published, but that feel awkward.

wescopeland: Should hopefully be a little better in latest.

I've added an "All" option that removes the `Flags` condition from the query that Filament builds up. When you manually removed the "Published" filter, the table falls back to this "All" option, thus making "Published" a selectable value.

Screenshot 2024-08-17 at 12 54 21 PM

Jamiras (approved):

Issue #127, opened 2018-05-23
All user profiles show current user's friends


Jamiras:
As a user, when I view someone's profile, I expect to see their friends, not mine.

The "Friends Leaderboard" always shows the current user's friends. It should either show the friends of the user being viewed, or not be present when viewing a profile other than the current user's.

Ben-Ghazi:
Same, what reason is there for needing to see other people's friend list?

Jamiras:
I'm just indicating that it's misleading as-is. It should either be corrected or removed.

PR #2620, opened 2024-08-17
feat(api): Get User Friend List

ioslife:
This endpoint is for fetching a user's friend list. Friends will only be returned if users are mutually following each other.

With respect to viewing other users' lists, this works the same way that we set up API_GetWantToPlayList as well. A target user's friend list can be viewed if the calling user and target user are friends.

```
/*
 *  API_GetUserFriendList - returns a list of Friends, with basic data, for a user
 *    u : username
 *  int         Count                       number of friend records returned in the response
 *  int         Total                       number of friend records the user actually has overall
 *  array       Results
 *   object      [value]
 *    string     Friend                     username of the friend
 *    int        Points                     number of hardcore points the friend has earned
 *    int        PointsSoftcore             number of softcore points the friend has earned
 *    string     LastSeen                   rich presence message for the friend
 *    int        ID                         unique id of the friend
 */

```

### DIFF
--- a/public/API/API_GetUserFriendList.php
+++ b/public/API/API_GetUserFriendList.php
@@ -2,7 +2,6 @@
 
 /*
  *  API_GetUserFriendList - returns a list of Friends, with basic data, for a user
- *    u : username
  *    o : offset - number of entries to skip (default: 0)
  *    c : count - number of entries to return (default: 100, max: 500)
  *  int         Count                       number of friend records returned in the response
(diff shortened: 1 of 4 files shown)


public/API/API_GetUserFriendList.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/*
+ *  API_GetUserFriendList - returns a list of Friends, with basic data, for a user
+ *    u : username
```
wescopeland, line 5: If I can only fetch my own friends list, why do I need to specify the `u` parameter?

wescopeland: Ah, I see:
> With respect to viewing other users' lists, this works the same way that we set up API_GetWantToPlayList as well. A target user's friend list can be viewed if the calling user and target user are friends.

Although, I'm thinking this should probably be more restricted. I want to say this endpoint should only allow for fetching the caller's friends list.

Jamiras: I agree. You should only be able to fetch your own friends list.

See also https://github.com/RetroAchievements/RAWeb/issues/127

ioslife:
I'm changing this endpoint to be usable to fetch following, follower, and friends lists. Converting to draft for the next few days until I'm happy with my implementation.

wescopeland:
Hi @ioslife! Is there anything we can do to support you here? Not sure if you're still interested in actively pursuing the implementation of this endpoint. Just let us know!

wescopeland:
Closing as stale. Please feel free to reopen this at any time!

PR #2596, opened 2024-08-07
refactor: migrate settings page back-end code to controllers

wescopeland:
This PR migrates the user settings page (controlpanel.php) back-end code to use Laravel controllers. As a result, almost all of the legacy _public/request_ files used by the page have been deleted.

There should be almost no functional changes as part of this PR. This is a precursor to migrating the UI of the page to React.

Where feasible, code coverage has been added. Some methods are a bit difficult to test, such as those for resetting progress, as these still use legacy SQL queries.

Additionally, a migration is included for `EmailConfirmations` to add an incrementing primary key and a `user_id` column.

For controller tests, I deliberately kept these as separate test cases. In my opinion, the trade-off of longer test execution times for easier-to-maintain tests is worth it. I think our focus should be on speeding up test execution time.

PR #2579, opened 2024-07-28
feat: finish migration of `/forums/recent-posts` to Inertia+React

wescopeland:
This PR is stacked on #2502 and has no functional changes. The purpose of this PR is to replace the Blade-powered recent posts page with the React-powered one.

**How to test this PR:**
* `sail artisan migrate` - the new recent posts Eloquent ORM query needs an index added.
* `sail npm run build` followed by `sail artisan inertia:start-ssr`.
* Navigate to `/forums/recent-posts`.

Summary of changes:
* Resource/Data classes are now used for automatic props mapping with `spatie/laravel-data`.
* TypeScript types are now auto-generated with `spatie/typescript-transformer`.
* The legacy `getRecentForumTopics()` helper has been removed.
* The controller code has moved to `ForumTopicController::recentlyActive()`, and an Eloquent ORM query is now utilized, along with `LengthAwarePaginator`.

PR #2624, opened 2024-08-18
feat(settings): migrate from legacy PHP to React

wescopeland:
I apologize for how large the diff is. While I reasonably tried to break out the changes into multiple PRs (back-end, dependencies, etc), this page just has many many editable forms which also required several added base components. It's a lot of concepts and patterns all wrapped up into one package, virtually none of which can be executed in isolation (a server-side React component cannot be rendered into a PHP file).

---

This PR rewrites the user settings page in React and is the culmination of work done in https://github.com/RetroAchievements/RAWeb/pull/2579, https://github.com/RetroAchievements/RAWeb/pull/2596, and https://github.com/RetroAchievements/RAWeb/pull/2621. 

## How to test this PR
If you just want to take a quick tour, check out https://preview.retroachievements.org/settings.

To run locally:
```
sail npm run build
```
Now you can navigate to http://localhost:64000/settings and start poking around.
The original route http://localhost:64000/controlpanel.php is still there to compare functionality. That file can be deleted in a quick follow-up.

There are virtually no back-end changes in this branch. It is 99.9% front-end, with the exception of adding `UserSettingsController::show()`.

### At a glance

The new page works great on both desktop and mobile. This is important, as 1 in 4 viewers on the page are [confirmed to be on mobile](https://plausible.retroachievements.org/retroachievements.org?auth=aMQAQjr37_a8AlTwzMXeN&filters=((is,page,(/controlpanel.php)))).

Screenshot 2024-08-18 at 10 12 41 AM

Screenshot 2024-08-18 at 10 12 07 AM

When the user submits a form, it hits one of the controller endpoints built out in https://github.com/RetroAchievements/RAWeb/pull/2596. The user sees a helpful toast notification, and the page does not refresh:

https://github.com/user-attachments/assets/ae8771a9-b0f8-47e7-965a-43d4988b45ef

### Notable Changes

* For accessibility reasons, most forms now _require_ the user to press some kind of submit button to update data.
* In the Profile section, there is a "Visible Role" field. Currently it's set as read-only text to the user's stringified legacy `Permissions` value. Eventually, this should map to the new site roles we've been building out, with the possibility of allowing the user to select which role they'd like to have prominently displayed on their profile, user card, and forum posts:

Screenshot 2024-08-18 at 10 20 45 AM

* In the Notifications section, the weekly RA newsletter option has been removed.
* Preferences are now driven by switches than checkboxes (although, they are still checkboxes in the DOM).
* In the Keys section, we no longer display the user's entire web API key. This is done so that if the user is screen recording, streaming, etc, they don't accidentally expose their web API key:

Screenshot 2024-08-18 at 10 23 59 AM

* In the Reset Game Progress section, the `<BaseSelect />` component is used rather than a native select. This prevents the select options from filling the entire height of the screen on some operating systems.

Screenshot 2024-08-18 at 10 25 20 AM

## How the Code Works

The root component of the page is _resources/js/pages/settings.tsx_.

Each section of the page is broken up logically into a separate & isolated component.

For the purpose of this walkthrough, we'll pick something simple: _ChangeEmailAddressSectionCard.tsx_. This is the component responsible for the "Change Email" section on the page:

Screenshot 2024-08-18 at 10 28 29 AM

Note that this component has nearly 100% code coverage via _ChangeEmailAddressSectionCard.test.tsx_ to verify its functionality.

The *.tsx file renders the template, of which we'll see several form fields, labels, etc.

The actual UI logic for submitting the data, handling validation, etc can be found within _useChangeEmailAddressForm.ts_:

```ts
const { form, mutation, onSubmit } = useChangeEmailAddressForm({ setCurrentEmailAddress });
```

In this React hook file, we first create the form schema using [zod](https://zod.dev/):
```ts
const changeEmailAddressFormSchema = z
  .object({
    newEmail: z.string().email(),
    confirmEmail: z.string().email(),
  })
  .refine((data) => data.newEmail === data.confirmEmail, {
    message: 'Email addresses must match.',
    path: ['confirmEmail'],
  });
```

The form schema is then input into our type-safe `useForm()` hook, powered by [react-hook-form](https://react-hook-form.com/):
```ts
const form = useForm<FormValues>({
  resolver: zodResolver(changeEmailAddressFormSchema),
  defaultValues: {
    newEmail: '',
    confirmEmail: '',
  },
});
```

All client-side AJAX handling is managed by [react-query](https://tanstack.com/query/latest). In this case, we're going to be sending some data to the server, so we use react-query's `useMutation()` hook:
```ts
const mutation = useMutation({
  mutationFn: (formValues: FormValues) => {
    return axios.put(route('settings.email.update'), formValues); // note our use of type-safe routes here
  },
  onSuccess: () => {
    props.setCurrentEmailAddress(form.getValues().newEmail);
  },
});
```

When the form indicates the user is submitting, we call `mutation.mutateAsync(formValues)` to send the data to the Laravel back-end:
```ts
const onSubmit = (formValues: FormValues) => {
  const confirmationMessage = auth?.user.roles.length
    ? 'Changing your email address will revoke your privileges and you will need to have them restored by staff. Are you sure you want to do this?'
    : 'Are you sure you want to change your email address?';

  if (!confirm(confirmationMessage)) {
    return;
  }

  toastMessage.promise(mutation.mutateAsync(formValues), {
    loading: 'Changing email address...',
    success: 'Changed email address!',
    error: 'Something went wrong.',
  });
};
```

This pattern is used in all of the forms on the page.

## Follow-up Stuff

In subsequent PRs, I will:
* Remove _controlpanel.blade.php_.
* Migrate all hardcoded strings to Laravel i18n resource JSON files, so we can support internationalization on this page and have an established pattern for using those translatable strings in React components.
* Increase the test coverage for the reset progress stuff. There are a few `it.todo()`s in _ResetGameProgressSectionCard.test.tsx_ that will require quite a bit of effort to fill in.


wescopeland:
I've redeployed this branch to the preview environment:
https://preview.retroachievements.org/settings

gboquizosanchez:
If I can give my feedback, I have seen something weird to me:

image
image

I think the color for the ring is huge. I'd put "focus:visible-ring-1" instead of "ring-2"

image
image

And in the light mode:

image

Those black buttons are really gaudy. So, I'd put a "light:bg-neutral-100" inside of the span that contains the switch buttons.

image

And something similar to the checkboxes.

wescopeland:
I am not sure what happened there with those Switch colors but that definitely was not the intent (see screenshots in OP).

Made some adjustments in latest for all those things: https://github.com/RetroAchievements/RAWeb/pull/2624/commits/27b72f1cf35b33ed6e493006951606220d1ca63d

Screenshot 2024-08-30 at 9 33 20 PM

I set focus-visible to ring-1 on `BaseInput`, but I'm going to keep it as ring-2 on the other fields as for those cases it's much more likely to be triggered via keyboard navigation.

resources/js/features/user/settings/components/+root/SettingsRoot.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const SettingsRoot: FC = () => {
+  return (
+    <div className="flex flex-col">
+      <h1>Settings</h1>
+
+      <div className="flex flex-col gap-4">
+        <ProfileSectionCard />
+        <NotificationsSectionCard />
+        <PreferencesSectionCard />
```
Jamiras, line 20: If changes are made to Notifications, and updated, then changes are made to Preferences and updated, the changes to Notifications are overwritten (only the changes to Preferences will be in the database).

This behavior does not occur on the controlpanel.php page.

resources/js/features/user/settings/components/KeysSectionCard/ManageWebApiKey.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+        <div className="col-span-3 flex w-full flex-col gap-2">
+          <SimpleTooltip
+            isOpen={isXs ? false : undefined}
+            isWrappingTapTarget={true}
+            tooltipContent="Copy to clipboard"
+          >
+            <BaseButton className="flex gap-2" onClick={handleCopyApiKeyClick}>
+              <LuCopy />
+              <span className="font-mono">{safeFormatApiKey(currentWebApiKey)}</span>
```
Jamiras, line 65: I'm guessing the truncation is to obscure sensitive information, but it looks weird to see the value collapsed when there's so much room.

image

resources/js/features/user/settings/components/+root/SettingsRoot.tsx
```diff
@@ -0,0 +1,29 @@
+import type { FC } from 'react';
```
Jamiras, line 1: generic errors appear in F12 console
image

resources/js/features/user/settings/components/ChangePasswordSectionCard/ChangePasswordSectionCard.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    await userEvent.click(screen.getByRole('button', { name: /update/i }));
+
+    // ASSERT
+    expect(putSpy).toHaveBeenCalledWith(route('settings.password.update'), {
+      currentPassword: '12345678',
+      newPassword: '87654321',
+      confirmPassword: '87654321',
+    });
+  });
```
Jamiras, line 59: I have concerns that a failed update attempt will not be noticed by the user. The new message is not plastered across the top of the screen like the old one was, and fades fairly quickly.
image

resources/js/features/user/settings/components/AvatarSection/AvatarSection.tsx
```diff
@@ -0,0 +1,119 @@
...
+              <div>
+                <BaseFormField
+                  control={form.control}
+                  name="imageData"
+                  render={({ field }) => (
+                    <BaseFormItem className="flex flex-col gap-2">
+                      <BaseFormLabel className="text-menu-link">New Image</BaseFormLabel>
+
+                      <BaseFormControl>
+                        <BaseInput
```
Jamiras, line 70: I feel like this should be a button. It's not obvious that you need to click Choose File. I immediately went to the Upload button expecting it to allow me to select a file.
image

Jamiras (commented):
Did not look into the code, but left some comments from interacting with the changes.

resources/js/features/user/settings/components/+root/SettingsRoot.tsx
wescopeland: Struggling to repo this locally:

Screenshot 2024-09-02 at 2 47 02 PM

It also isn't occurring on preview:

Screenshot 2024-09-02 at 2 47 32 PM

wescopeland: OK, it seems I can repro if a Vite dev/prod server instance isn't running.

If you do:
```
sail npm run build
```
Then visit the `/settings` route, do you still see errors?

Something must be wonky with your Vite setup locally, which may also be why things were odd in the i18n PR.


https://github.com/user-attachments/assets/24d568fc-f611-477a-9b17-22ab3d8f2924

Jamiras: `sail artisan inertia:start-ssr` does indeed eliminate those console errors. Is this something I'm going to have to remember to do every time I reboot my VM?

wescopeland: Shouldn't be - `sail artisan inertia:start-ssr` is meant for use in production environments.

`sail npm run dev` should be sufficient and is what I use for local dev (and also causes the console errors to not appear). I only see the console errors when Vite isn't running at all.

Screenshot 2024-09-02 at 3 15 13 PM

Note that I see this console output on _every_ page I navigate to, regardless of whether React is present:
```
[vite] connecting...
[vite] connected.
```
If you don't, something may be wrong with how the Vite dev server is spinning up locally in your environment.

resources/js/features/user/settings/components/+root/SettingsRoot.tsx
wescopeland: Good catch, I've fixed this in latest.

resources/js/features/user/settings/components/KeysSectionCard/ManageWebApiKey.tsx
wescopeland: In latest, this now only goes full-width on small device sizes.

Screenshot 2024-09-02 at 4 10 01 PM

resources/js/features/user/settings/components/ChangePasswordSectionCard/ChangePasswordSectionCard.test.tsx
wescopeland: I understand your concerns.

In latest, I've made two changes to try to help with this. First, I've increased the default duration of all toasts by 1 more second. I've also improved the contrast of all toasts so it's more obvious when one appears:

**Before**
Screenshot 2024-09-02 at 3 55 31 PM


**After**
Screenshot 2024-09-02 at 3 55 22 PM

resources/js/features/user/settings/components/AvatarSection/AvatarSection.tsx
wescopeland: As an HTML native file input, this took some doing to make it look like a button when it's actually not, but in latest it should be a fairly close approximation:

Screenshot 2024-09-02 at 4 12 19 PM

Additionally, it's appropriate for the Upload button on this particular form to be disabled if there's no file input, so I've made that change. This should help mitigate people reaching immediately for that button.

Jamiras (approved):

PR #2627, opened 2024-08-20
chore: upgrade to laravel 11

wescopeland:
Follow-up to #2598. This PR migrates the project to Laravel 11.

Fairly straightforward. This is another PR that is mostly just lockfile, vendor, and lint noise. I followed the official upgrade guide [here](https://laravel.com/docs/11.x/upgrade).

Note that Laravel 11 is compatible with both Carbon 2 and Carbon 3, however, Carbon 3 has a [major breaking change](https://laravel.com/docs/11.x/upgrade#carbon-3) which directly impacts a lot of our code. As a result, I've pinned the project to Carbon 2 (what we're already using in prod) until such time we're ready to take on the endeavor of walking through all our `diffIn*` function calls. For the time being, this seems like high effort with low value.

**How to test this PR:**
```
composer install
```
Click around and do some basic smoke tests. Log in, log out, visit a few game and achievement pages.

**Future work:**
* Install and set up [Laravel Pulse](https://pulse.laravel.com/) so we can rapidly identify performance bottlenecks in the site's back-end. Only a beta release of Pulse is compatible with Laravel 10-- seems like it'd make more sense just to use a stable release as supported in Laravel 11.


Jamiras (approved):
Brief smoketesting didn't encounter anything.

PR #2633, opened 2024-08-22
don't return users with 0 hardcore achievements unlocked

Jamiras:


wescopeland (approved):
Verified locally. Wondering if it may also be good to redirect back to the game page in case of empty states, but not necessary for this branch.

Jamiras:
I think a redirect would be more confusing to the user than an empty list.

PR #2638, opened 2024-08-23
allow muted users to message team accounts

Jamiras:
The "new message" button is currently hidden for muted users. This re-enables the "new message" button, but will generate an error if the user tries to send to a non-team account.

https://discord.com/channels/476211979464343552/1002689485005406249/1276533882983940219


wescopeland (approved):
Verified locally.

Issue #2640, opened 2024-08-24
Last played date is updated erroneously


suspect15:
**Describe the bug**
On some player profiles, the "last played" dates in the "Last 5 Games Played" section displays the current date despite the player not having been active.

**To Reproduce**
See [https://retroachievements.org/user/donutweegee](https://retroachievements.org/user/donutweegee) as an example

**Expected behavior**
The dates in the "Last 5 Games Played" to be accurate.

**Screenshots**
![image](https://github.com/user-attachments/assets/26eef415-0a13-4712-a3bf-a7c40610b493)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Firefox
 - Version 192.0.2

PR #2643, opened 2024-08-24
fix(user): don't show today's date on user profiles when given null dates

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/2640.

Though this fixes the UI issue, we may also want to decide if resetting a game should soft delete its associated `player_games` row. This is the ultimate source of the issue, though even if we decide to handle that we still need to make this UI update to handle rows that aren't already soft deleted.

**Before**
Screenshot 2024-08-24 at 7 39 55 AM

**After**
Screenshot 2024-08-24 at 7 39 46 AM


Jamiras (approved):

Jamiras:
> we may also want to decide if resetting a game should soft delete its associated player_games row.

I don't think so. The user still last played the game whenever they last played it. We already capture/store this information when playing games that don't have achievements, why wouldn't we retain it if the user decided they didn't want their achievements?

I believe the underlying issue isn't that the user reset their games. It's that they didn't have any unlocks when we started populating the player_games tables, so the last_played_at field was left as null.

PR #2644, opened 2024-08-24
perf(forum): significantly improve load performance of recent topics

wescopeland:
This PR rewrites and significantly optimizes the SQL queries used to load forum recent topics (`/forums/recent-posts`). On my local machine, I see a 95% reduction in time to load the page.

**Before**
Screenshot 2024-08-24 at 11 34 16 AM

**After**
Screenshot 2024-08-24 at 11 34 05 AM

There are a variety of tricks used to achieve this optimization, including forcing the query to use a new index and swapping several LEFT JOINs for INNER JOINs.


Jamiras (approved):

Jamiras:
My performance improvement wasn't as drastic. It was only about 2.5x faster.

wescopeland:
@Jamiras I've made one change here based on some feedback I received from @gboquizosanchez. `getTotalRecentForumTopics()` now uses Eloquent ORM rather than raw SQL for the query. Perf appears to be identical.

Jamiras (approved):

PR #2645, opened 2024-08-24
refactor: migrate the Terms page to React

wescopeland:
No functional changes. With the exception of fixing some invalid HTML and adding a meta description, should be 1:1 equivalent with the Blade page.

Accessible at `/terms`.


resources/js/pages/terms.tsx
```diff
@@ -0,0 +1,169 @@
...
+      </Head>
+
+      <AppLayout.Main>
+        <h1 className="mb-4">Legal & Terms</h1>
+
+        <div className="flex flex-col gap-y-4">
+          <div className="rounded bg-embed p-4">
+            <div className="flex flex-col gap-y-6">
+              <div>
+                <h2 className="text-h4" id="conditions">
```
Jamiras, line 24: I assume it's because of the sticky menu, but this link appears to go to the next paragraph.
image

Also verified this behavior is present outside of this PR.

wescopeland: Tricky - might as well go ahead and fix this while we're here.

In latest, I've moved the `id` to an invisible div that has some offset so the user lands here:
Screenshot 2024-08-28 at 6 54 21 PM

Jamiras (approved):

PR #2646, opened 2024-08-25
Add hash information to game history

Jamiras:
On tickets page and moderation page. Description of hash appears in tooltip.
image

If the description includes any of the keywords associated with multi-disc games, an asterisk is added with a tooltip indicating that only the first disc loaded is captured for a multi-disc game session.
image

Also attempts to pre-select a hash when creating a ticket based on session data, similar to how emulator/version information is pre-filled.
image

### DIFF
--- a/resources/views/components/user/game-activity.blade.php
+++ b/resources/views/components/user/game-activity.blade.php
@@ -32,6 +32,18 @@ use App\Platform\Enums\AchievementFlag;
                         @if ($session['type'] === PlayerGameActivitySessionType::Player)
                             <td>
                                 <span class='text-muted'>Started Playing</span>
+                                @if ($session['playerSession'] && $session['playerSession']['gameHash'])
+                                    @php $hash = $session['playerSession']['gameHash']; @endphp
+                                    <span class="smalltext" title="{{ $hash['md5'] }}">
+                                        {{ $hash['name'] }}
+                                        @if ($hash->isMultiDiscGameHash())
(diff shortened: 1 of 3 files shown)


resources/views/components/user/game-activity.blade.php
```diff
@@ -47,6 +47,18 @@
                                     @endif
                                     </span>
                                 @endif
+                                @if ($session['playerSession'] && $session['playerSession']['gameHash'])
+                                    @if ($session['userAgent'])
+                                        <span class="text-muted"> | </span>
+                                    @endif
+                                    @php $hash = $session['playerSession']['gameHash']; @endphp
+                                    <span class="smalltext" title="{{ $hash['name'] }}">
+                                        {{ $hash['md5'] }}
```
wescopeland, line 56: Looks good - only thing I'm wondering is if it may be worth flipping this (put the md5 in the tooltip and have the name visible). Users probably will just see a blur of characters for the md5, but I would imagine the hash name is much more recognizable.

wescopeland (approved):

resources/views/components/user/game-activity.blade.php
Jamiras: I considered that, but decided against it for reasons that I've already forgotten. I'll switch it around.

I've also changed the order so the hash description appears first:
image
image

wescopeland (approved):

PR #2651, opened 2024-08-27
chore: re-add migration value insertions into mysql schema dump

wescopeland:
Didn't reinclude them in the sqlite dump as that's probably not useful.


Jamiras:
For context: If the migrations table is empty, the next migration applied will recreate the database. While that's not the end of the world for seeded databases as they can just be re-seeded, it's unexpected behavior. Suddenly having a blank database is a surprise, and if users have been actively setting up the database for specific work, having to recreate the test data is undesirable.

There needs to be at least one record in the table to prevent `sail artisan migrate` from re-running the `mysql-schema.sql` file, which destroys and recreates the database. Putting all 107 rows in is probably overkill, but it's an infrequent operation, so it's okay.

NOT doing the inserts for the sqlite-schema.sql is desirable as the database already gets recreated for _every_ unit test, and we don't want _any_ unnecessary overhead.

Jamiras (approved):

PR #2652, opened 2024-08-28
ci(pre-push): run phpunit in parallel mode when detecting a macOS or Linux system

wescopeland:
Parallel mode is incompatible with Windows.

In _pre-push_, explicitly check if the developer is using macOS or Linux. If those operating systems are detected, use parallel mode. If they aren't detected, continue with today's behavior of using vanilla `phpunit`.


.hooks/pre-push
```diff
@@ -2,6 +2,9 @@
 # exit on error
 set -e
 
+# Properly handle SIGINT (Ctrl/Cmd+C) when running paratest.
+trap 'kill 0' SIGINT
```
Jamiras, line 6: I'm not sure that this is compatible with linux:
```
jamiras@ubuntu24 pr/2652:~/source/RAWeb$ .hooks/pre-push 
trap: SIGINT: bad trap
```

Jamiras: I'm suspicious about whether it would work in Windows either. `trap` is not a valid DOS command (but neither is `printf`).

.hooks/pre-push
```diff
@@ -14,7 +17,18 @@ npx eslint --quiet
...
-vendor/bin/phpunit
+# Check the OS. Windows does not support the --parallel flag.
+if [ "$(uname)" = "Linux" ] || [ "$(uname)" = "Darwin" ]; then
+  # Use --parallel if macOS or Linux are detected.
+  printf "\n⏳ composer test -- --parallel\n"
+  vendor/bin/paratest &
+else
+  # If neither of those are detected, don't use --parallel.
+  printf "\n⏳ composer test\n"
+  vendor/bin/phpunit &
```
Jamiras, line 28: could the trap/wait be moved into the paratest logic, leaving this to run in the foreground (no `&`)?

.hooks/pre-push
```diff
@@ -14,7 +17,18 @@ npx eslint --quiet
 printf "\n⏳ npm run test\n"
 npm run test
 
-printf "\n⏳ composer test\n"
-vendor/bin/phpunit
+# Check the OS. Windows does not support the --parallel flag.
+if [ "$(uname)" = "Linux" ] || [ "$(uname)" = "Darwin" ]; then
+  # Use --parallel if macOS or Linux are detected.
+  printf "\n⏳ composer test -- --parallel\n"
+  vendor/bin/paratest &
```
Jamiras, line 24: Why run this in the background (`&`)? It doesn't seem necessary, and you could avoid the trap/wait.
```
jamiras@ubuntu24 pr/2652:~/source/RAWeb$ vendor/bin/paratest
ParaTest v7.4.0 upon PHPUnit 10.5.10 by Sebastian Bergmann and contributors.

Processes:     4
Runtime:       PHP 8.2.22
Configuration: /home/jamiras/source/RAWeb/phpunit.xml

...............................................................  63 / 385 ( 16%)
............................................................... 126 / 385 ( 32%)
............................................................... 189 / 385 ( 49%)
............................................................... 252 / 385 ( 65%)
............................................................... 315 / 385 ( 81%)
............................................................... 378 / 385 ( 98%)
.......                                                         385 / 385 (100%)

Time: 00:27.435, Memory: 26.00 MB

OK (385 tests, 2470 assertions)
jamiras@ubuntu24 pr/2652:~/source/RAWeb$ vendor/bin/phpunit
PHPUnit 10.5.10 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.22
Configuration: /home/jamiras/source/RAWeb/phpunit.xml

...............................................................  63 / 385 ( 16%)
............................................................... 126 / 385 ( 32%)
............................................................... 189 / 385 ( 49%)
............................................................... 252 / 385 ( 65%)
............................................................... 315 / 385 ( 81%)
............................................................... 378 / 385 ( 98%)
.......                                                         385 / 385 (100%)

Time: 01:33.493, Memory: 123.00 MB

OK (385 tests, 2470 assertions)
```

.hooks/pre-push
wescopeland: Fired up a Linux VM to confirm the error - though is now moot with latest changes.

> I'm suspicious about whether it would work in Windows either
On Windows, I'm pretty sure this hook has always required Git Bash or some other kind of special shell.

.hooks/pre-push
wescopeland: When I was making this update, I encountered a bizarre issue where paratest ran in the background and could not be terminated, even after killing the foreground pre-push hook process.

I don't seem to be having this problem anymore though 🤔

I've removed the `trap`, `wait`, and `&` stuff in latest.

.hooks/pre-push
wescopeland: trap/wait all removed in latest and paratest now runs in the foreground.

Jamiras (approved):
Confirmed functional on linux

PR #2653, opened 2024-08-28
feat(analytics): track gamelist type and forum topic id

wescopeland:
Fixes a bug where the gamelist type is not being tracked properly. Also sends along what forum topic ID users are viewing.

**How to test this PR:**
Open your browser's network tab. View your Want to Play Games List. You should see the `type` prop being sent along in the pageview event:
Screenshot 2024-08-28 at 8 54 01 AM

Open any forum topic. You should see the `topicId` prop being sent along in the pageview event:
Screenshot 2024-08-28 at 8 54 33 AM

It's okay that IDs are passed along as strings.


resources/views/components/head-analytics.blade.php
```diff
@@ -53,6 +54,19 @@
...
+    if (str_ends_with($url, 'gameList.php') && isset($queryParams['t'])) {
+        $tParam = $queryParams['t'];
+        if ($tParam === 'play' || $tParam === 'develop') {
+            $props['type'] = $queryParams['t'];
+        }
+    }
+
+    // Track what topic ID users are viewing at viewtopic.php.
+    if (strpos($url, 'viewtopic') !== false && isset($queryParams['t'])) {
+        $props['topicId'] = $queryParams['t'];
```
Jamiras, line 68: I don't think it's worth capturing the current page for multi-page topics, but we need to be aware that "hot" topics might just be because users are scanning through pages of posts.

Jamiras (approved):

resources/views/components/head-analytics.blade.php
wescopeland: This is true - though likely still a valuable datapoint. We may find there are specific topics that people are likely to spend time scanning through (or not). Either way, a good thing to keep in mind as we parse this data.

PR #2661, opened 2024-08-31
chore: set up Laravel Pulse

wescopeland:
Adds [Laravel Pulse](https://pulse.laravel.com/) observability to the server.

After running the auto-generated migration, Pulse is accessible via the `/pulse` route. It looks something like this:

Screenshot 2024-08-31 at 2 28 35 PM

Some other notes:
* Accessible only to users with the admin or root role.
* In _config/pulse.php_, I've set up modest sampling rates so we're not writing an excessive amount of data to the DB.


Jamiras (approved):
Not really sure what to test here, but it loads up, and doesn't seem to affect anything else noticeably - although I'm hardly putting any load on the system.

wescopeland:
Yea, probably not much to test here other than it loads up fine. If we do find that it puts noticeable load on the system, we can aggressively scale back the sampling rate.

We'll need to remember to update the settings JSON file in app storage on deploy so the new menu option will appear.

PR #2662, opened 2024-08-31
feat(multiset): start double-writing to game_achievement_sets

wescopeland:
This PR introduces some foundational work required to support multiple achievement sets per game.

## Abstract
We aim to allow players to simultaneously pursue multiple achievement sets within a single game session. This initiative addresses the existing limitation of being confined to a single achievement set per ROM hash.

Multiset functionality is predicated on several key considerations that underpin the system's design and user interaction with achievement sets. From the user perspective, there will be four types of sets: Core, Bonus, Specialty, and Exclusive.

* **Core and Bonus Sets.** Bonus sets are designed to directly complement and supplement Core sets. When a user loads a ROM hash for a Core set, we aim to allow them to unlock achievements from Bonus sets too. This can be considered a "core to subset" relationship.
* **Specialty Sets.** Unlike Bonus sets, Specialty sets will continue to require loading a unique hash. However, players will be permitted to earn achievements from the Core set and any applicable Bonus sets simultaneously when the Specialty set hash is loaded. This can be considered a "subset to core" relationship.
* **Exclusive Sets.** Exclusive sets must be played in isolation. Like Specialty sets, Exclusive sets require loading a unique hash. Players are NOT permitted to earn achievements from the Core set or Bonus sets when an Exclusive set hash is loaded.

The large majority of sets will likely either be Core or Bonus.

---

This PR adds a sync script and double writes to `game_achievement_sets`, `achievement_sets`, and `achievement_set_achievements`.

Every set is currently identified as being a Core set, just like how production functionally works today.

To sync your database, run the following command:
```bash
sail artisan ra:sync:game-achievement-sets 1 # sync game ID 1
sail artisan ra:sync:game-achievement-sets # sync all games
```

After the sync command has executed, you should be able to observe that the values in `game_achievement_sets`, `achievement_sets`, and `achievement_set_achievements` are correct.

**Next steps:**
* In the Connect API, send achievement data to clients using `game_sets`.
* On game pages, use `game_sets` to render the achievement list.
* Attach leaderboards to sets rather than games.
* After #2624, add a new user setting for global subset opt-in/out.
* Add a new table to record local subset opt-in/out overrides.
* Build tooling allowing developers to specify set kinds for any game with the string "[Subset -" in the title.
* Build tooling allowing developers to specify hash compatibility for each set.


app/Platform/Enums/AchievementSetType.php
```diff
@@ -4,38 +4,47 @@
...
-        return match ($type) {
-            self::Core => 'Core',
-            self::Bonus => 'Bonus',
-            self::Specialty => 'Specialty',
-            self::WillBeBonus => 'Will Be Bonus',
-            self::WillBeSpecialty => 'Will Be Specialty',
-            default => 'Invalid state',
-        };
-    }
+    case CORE = "core";
```
Jamiras, line 9: All of our existing enums are PascalCase. Why ALLCAPS here?

database/factories/GameAchievementSetFactory.php
```diff
@@ -33,8 +33,8 @@ public function type(string $type): self
         return $this->state(function (array $attributes) use ($type) {
             return [
                 'type' => $type,
-                'title' => $type === AchievementSetType::Core ? null : $this->faker->words(3, true),
-                'order_column' => $type === AchievementSetType::Core ? 0 : $this->faker->numberBetween(1, 10),
+                'title' => $type === AchievementSetType::CORE->value ? null : $this->faker->words(3, true),
+                'order_column' => $type === AchievementSetType::CORE->value ? 0 : $this->faker->numberBetween(1, 10),
```
Jamiras, line 37: What does the "->value" do for us here?

app/Platform/Actions/UpsertGameCoreAchievementSetFromLegacyFlags.php
```diff
@@ -0,0 +1,103 @@
...
+                $game->achievementSets()->attach($newAchievementSet->id, [
+                    'type' => AchievementSetType::CORE->value,
+                    'order_column' => 0,
+
+                    // Preserve the existing timestamps. Note that updated_at reuses the created_at
+                    // value, as none of the fields stored in game_achievement_sets would've changed
+                    // from the time of the first achievement promotion.
+                    'created_at' => $newAchievementSet->created_at,
+                    'updated_at' => $newAchievementSet->created_at,
+                ]);
```
Jamiras, line 33: Should this also auto-generate a subset record for subsets?
i.e. game 25176 would have two entries in the game_achievement_sets table:
1) `game_id 25176 core` -> `achievement_set_id_N`
2) `game_id 9823 will_be_subset` -> `achievement_set_id_N`

Or is the intent to do that manually at some point, or to change the `25176 core` set to `will_be_subset`?

app/Platform/Actions/UpsertGameCoreAchievementSetFromLegacyFlags.php
```diff
@@ -0,0 +1,103 @@
...
+
+            // Preserve the existing timestamps as best as we can.
+            'created_at' => $allAchievements->min(fn ($achievement) => $achievement->DateCreated ?? now()),
+            'updated_at' => $allAchievements->max(fn ($achievement) => $achievement->DateModified ?? now()),
+        ]);
+
+        // Next, attach all the achievements to the set.
+        $allAchievements->each(function ($achievement) use ($achievementSet) {
+            $achievementSet->achievements()->attach($achievement->ID, [
+                // Preserve the existing timestamps as best as we can.
```
Jamiras, line 62: should also preserve the order_column.

app/Platform/Enums/AchievementSetType.php
wescopeland: This should indeed be using PascalCase. Fixed in latest.

database/factories/GameAchievementSetFactory.php
wescopeland: Most enums in RAWeb are not true enums, but rather abstract classes. PHP 8.1 introduced native enums, which have better type safety, are batteries-included with a set of sane instance methods, and are easily serializable.

When studying how to best utilize native enums, I came across this document which I learned a lot of best practices from: https://jeffochoa.me/php-enums-best-practices

`->value` is the built-in property for accessing the scalar value associated with an enum case. This is much safer than using an abstract class style enum, for example:
```php
// returns "Invalid flag", but nothing stops someone from passing in 7.
AchievementFlag::toString(7);
```

If `AchievementFlag` were a native enum, eg:
```php
enum AchievementFlag: int {
    case OfficialCore = 3;
    case Unofficial = 5;
}
```
This would throw an error:
```php
$AchievementFlag::from(7);
```

Another benefit we get is being able to use native enums as types in function args. Abstract class enums do not support this. eg:
```php
 // invalid!
function myFunction(AchievementFlag $flag) { ... }

 // valid.
function myFunction(AchievementSetType $type) { ... }
```

Ultimately, I believe native enums give us the same benefits that we get from abstract classes, but make the code much safer in the long run.

app/Platform/Actions/UpsertGameCoreAchievementSetFromLegacyFlags.php
wescopeland: The intent is to do it manually later with the `will_be_*` stuff and new tools provided for devs in the management app. They will explicitly set what kind of subset it is rather than us trying to intuit it.

app/Platform/Actions/UpsertGameCoreAchievementSetFromLegacyFlags.php
wescopeland: Good catch, fixed in https://github.com/RetroAchievements/RAWeb/pull/2662/commits/78adfa7f91345592f30b0ab5db2c1c39a2f43303.

Jamiras: Do I need to drop the table to see this change? I reran the artisan command, and I see updated timestamps on my records, but order_column is still null:
```
MariaDB [retroachievements-web]> select * from achievement_set_achievements where achievement_id=84700\G
                id: 281293
achievement_set_id: 5760
    achievement_id: 84700
      order_column: NULL
        created_at: 2019-10-04 01:07:15
        updated_at: 2024-09-08 13:59:53
1 row in set (0.002 sec)

MariaDB [retroachievements-web]> select DisplayOrder from Achievements where ID=84700;
+--------------+
| DisplayOrder |
+--------------+
|           16 |
+--------------+
1 row in set (0.001 sec)
```

database/factories/GameAchievementSetFactory.php
Jamiras: I didn't realize that PHP had strongly typed enums. I'm all for using them.

So the `->value` is necessary to downcast it back to an int? Makes sense.

wescopeland: Yep, just skimming over [this article](https://jeffochoa.me/php-enums-best-practices) is a great ~2 minute crash course on all the built-in functionality.

Jamiras (approved):

PR #2659, opened 2024-08-31
feat: migrate Supported Game Files to React

wescopeland:
I promise as more of these PRs come through they'll get smaller as the surface area of foundational work decreases 😓

This PR migrates the Supported Game Files page to React.js. There are a few changes to note:

* The page now has a meta description which provides the number of hashes available for the game.
* The copy in the embed under the page heading has been slimmed down to 3 short sentences. The first two sentences on the original page were mostly restating the same thing in different ways. I also swapped the order of the last two sentences, as we know now that people are likely to click on the 1st link they see - it would be reasonable that it's the game's forum topic.
* The mobile UX has been greatly improved. Breadcrumbs have also been removed on mobile. There is quite a bit of [UX research](https://www.nngroup.com/articles/breadcrumbs/) out there which suggests breadcrumbs can cause more harm than good on mobile, especially since mobile users can backswipe. Mobile users who land on the page sideways can still click the game avatar to land on the game page.
* **The page no longer requires authentication.** See [here](https://discord.com/channels/476211979464343552/1002696640001478769/1278845062499008512).

**Before**
Screenshot 2024-08-30 at 9 10 49 PM

Screenshot 2024-08-30 at 9 11 02 PM

**After**
Screenshot 2024-08-30 at 9 10 26 PM

Screenshot 2024-08-30 at 9 10 36 PM

PR #2663, opened 2024-09-01
refactor: improve type safety of forum Recent Posts page

wescopeland:
Brings over some developer experience improvements from https://github.com/RetroAchievements/RAWeb/pull/2659 into the Recent Posts (`/forums/recent-posts`) page:

* The page props now use autogenerated types from end-to-end. Previously, the page was using a handwritten `RecentPostsPageProps` type. Now it's using an autogenerated `App.Community.Data.RecentPostsPageProps` type.

> [!NOTE]
> To autogenerate types, `composer types` can be run in the terminal, which rebuilds _generated.d.ts_. `composer types` scans over every PHP class that has a `#[TypeScript]` annotation and outputs types for the React app to use, helping ensure end-to-end type safety.

* Breadcrumbs are removed from the page on the XS breakpoint. There is quite a bit of [UX research](https://www.nngroup.com/articles/breadcrumbs/) out there which suggests breadcrumbs can cause more harm than good on mobile, especially since mobile users can backswipe.

* `usePage()` has been replaced with `usePageProps()`, which requires much less boilerplate.
* `<RecentPostsBreadcrumbs />` has been lifted up to be `<ForumBreadcrumbs />`, though is still functionally the same.


Jamiras (approved):
Spot testing seems functional.

PR #2665, opened 2024-09-02
refactor: migrate the Random Set route to a controller method

wescopeland:
This route doesn't fit well into React. This PR drops the Blade template and moves the logic directly into a controller.

There is one functional change. The "Random Set" label has been changed to "Random Game". IMO, this better describes what the logic is actually doing - redirecting the user to a game page. "Set" may also not be a well understood term outside of power users and developers.


resources/views/components/menu/main.blade.php
```diff
@@ -121,7 +121,7 @@
             <x-dropdown-item :href="url('setRequestList.php')">Most Requested</x-dropdown-item>
             <x-dropdown-item :href="route('claims.completed')">New Sets & Revisions</x-dropdown-item>
             <x-dropdown-item :href="route('claims.active')">Sets in Progress</x-dropdown-item>
-            <x-dropdown-item :href="route('game.random')">Random Set</x-dropdown-item>
+            <x-dropdown-item :href="route('game.random')">Random Game</x-dropdown-item>
```
Jamiras, line 124: > There is one functional change. The "Random Set" label has been changed to "Random Game". IMO, this better describes what the logic is actually doing - redirecting the user to a game page. 

The idea behind using "Random Set" is that it only returns games _with_ achievements.

Jamiras (approved):

PR #2667, opened 2024-09-02
fix(deps): bump filament to latest to fix some Laravel 11 compatibility issues

wescopeland:
Filament on stage is exhibiting some strange behavior stemming from vendor code.

For example, if you navigate to https://stage.retroachievements.org/manage/games/1, then scroll down and click the Leaderboards tab, then try to do a row action, nothing will happen and you'll get a bunch of console errors:

Screenshot 2024-09-02 at 10 34 54 AM

This was resolved in a Filament release a few months ago (https://github.com/filamentphp/filament/pull/13178). Bumping to latest seems to fix the issue.


app/Support/Shortcode/Shortcode.php
wescopeland, line 1: These are all linting changes. Bumping Filament must have also bumped phpstan or something phpstan depends on. New errors started being thrown in this file from `$matches[1]` and `$matches[2]` always being truthy (the `isset()` and conditional are redundant).

Jamiras (approved):

PR #2568, opened 2024-07-21
feat(seeders): improve forum topic seeders

luchaos:
Forum topic & comments seeders required adjusting in order to test #2502 with synthetic data.

PR #2668, opened 2024-09-02
improve forum topic seeding

Jamiras:
Implements https://github.com/RetroAchievements/RAWeb/pull/2568#discussion_r1692013010, as well as improves the content of the posts to include several paragraphs and some embedded links and formatting.

image

image


wescopeland (approved):
Verified locally on a fresh DB. Both seeders run fast and produce data that's of much higher quality.

There's nothing in the implementation details that jumps out at me as worth changing or fixing.

PR #2670, opened 2024-09-02
feat(analytics): track user scheme and theme

wescopeland:
Tiny change to add tracking for the `scheme` and `theme` cookies. I would like to better understand the following things:

* What % of users are actually using light mode?
* What % of users are using a non-default color scheme, and which one is the most popular?
* Are there any non-default color schemes that get near-zero use?
* If we improve light mode, will its usage % increase?

**How to test this PR:**
You should see the two new props being sent in the pageview `event` payload in your network tab:

Screenshot 2024-09-02 at 7 06 32 PM


resources/views/components/head-analytics.blade.php
```diff
@@ -35,6 +35,8 @@
 
     $props = [
         'isAuthenticated' => auth()->check(),
+        'scheme' => request()->cookie('scheme') ?: 'dark',
+        'theme' => request()->cookie('theme') ?: 'default',
```
Jamiras, line 39: I don't think it hurts anything, but these values are still sent for the /manage/ pages, which have their own color theme settings.

Jamiras (approved):

resources/views/components/head-analytics.blade.php
wescopeland: Doesn't hurt anything, but a good callout nonetheless & I appreciate it.

Since I'm interested in the cookie value itself, I don't believe this will lead to any data poisoning.

PR #2671, opened 2024-09-03
fix: restore red background on laravel-debugbar timings

wescopeland:
There seems to be an issue with laravel-debugbar's red background that we commonly use to measure timings. Some vendor CSS styles are conflicting, causing the red background to not appear. I believe this could be related to the same order of operations issue we encountered in https://github.com/RetroAchievements/RAWeb/pull/2650, though I still do not understand the root cause.

This PR bumps laravel-debugbar to latest and adds `!important` to those styles so the red bar appears again.

**Before**
Screenshot 2024-09-03 at 7 56 58 AM

**After**
Screenshot 2024-09-03 at 7 56 46 AM


app/Support/Shortcode/Shortcode.php
wescopeland, line 1: These are the same linting changes as https://github.com/RetroAchievements/RAWeb/pull/2667/files#r1741016920.

Jamiras (approved):

PR #2674, opened 2024-09-03
feat(api): add API_GetGameHashes endpoint

wescopeland:
Accompanies https://github.com/RetroAchievements/RAWeb/pull/2659.

Now that the route is unauthenticated, it would be better to facilitate an endpoint for this information rather than invite scraping. I have received an explicit request that we add this endpoint from one of our larger API consumers.

This is a little bit different from API_GetGameList.php?h=1, as it's focused on single games and returns all the metadata about the game's hashes:

```
GET API_GetGameHashes.php?i=1
```
```json
{
	"Results": [
		{
			"MD5": "1b1d9ac862c387367e904036114c4825",
			"Name": "Sonic The Hedgehog (USA, Europe) (Ru) (NewGame).md",
			"Labels": [
				"nointro",
				"rapatches"
			],
			"PatchUrl": "https:\/\/github.com\/RetroAchievements\/RAPatches\/raw\/main\/MD\/Translation\/Russian\/1-Sonic1-Russian.zip"
		},
		{
			"MD5": "1bc674be034e43c96b86487ac69d9293",
			"Name": "Sonic The Hedgehog (USA, Europe).md",
			"Labels": [
				"nointro"
			],
			"PatchUrl": null
		}
	]
}
```


public/API/API_GetGameHashes.php
```diff
@@ -0,0 +1,44 @@
...
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'i' => ['required', 'min:1'],
+]);
+
+$game = Game::find($input['i']);
+
+if (!$game) {
+    return response()->json([], 404);
```
Jamiras, line 28: Should the 404 response be an object instead of an array, perhaps with an empty "Results" to match a game with no hashes?

Game with no hashes:
```
{"Results":[]}
```

Invalid game ID:
```
[]
```

wescopeland: No objection here - updated in latest.

Jamiras (approved):

PR #2682, opened 2024-09-07
feat(manage): allow users to upload game icon images

wescopeland:
This PR makes a change to allow developers to upload new game icon images via the management app's game edit page.

Other media types will be supported in the future, but this felt like a good chunk to knock out first just to establish a pattern that can be applied to those other fields.

https://github.com/user-attachments/assets/825fc594-a88a-4527-a55d-97cd77637dbc

### DIFF
--- a/app/Filament/Resources/GameResource.php
+++ b/app/Filament/Resources/GameResource.php
@@ -337,6 +337,9 @@ class GameResource extends Resource
                             ->acceptedFileTypes(['image/jpeg', 'image/png', 'image/gif'])
                             ->maxSize(1024)
                             ->rules('dimensions:width=96,height=96')
+                            ->validationMessages([
+                                'dimensions' => 'The game icon must be 96x96 pixels.'
+                            ])
                             ->preserveFilenames(),
                     ])
                     ->collapsible()
--- a/app/Filament/Resources/GameResource.php
+++ b/app/Filament/Resources/GameResource.php
@@ -322,6 +322,25 @@ class GameResource extends Resource
                             ->default('day')
                             ->reactive(),
                     ]),
+
+                Forms\Components\Section::make('Media')
+                    ->icon('heroicon-o-photo')
+                    ->description("Provide the game icon, screenshots, and box art.")
+                    ->columns(['md' => 2, 'xl' => 3, '2xl' => 4])
+                    ->schema([
+                        Forms\Components\FileUpload::make('ImageIcon')
+                            ->label('Game Icon')
+                            ->disk('local')
+                            ->directory('temp')
+                            ->visibility('private')
+                            ->image()
+                            ->acceptedFileTypes(['image/jpeg', 'image/png', 'image/gif'])
+                            ->maxSize(1024)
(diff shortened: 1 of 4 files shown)


app/Filament/Resources/GameResource.php
```diff
@@ -322,6 +322,26 @@ public static function form(Form $form): Form
...
+                    ->schema([
+                        Forms\Components\FileUpload::make('ImageIcon')
+                            ->label('Game Icon')
+                            ->disk('local')
+                            ->directory('temp')
+                            ->visibility('private')
+                            ->image()
+                            ->acceptedFileTypes(['image/jpeg', 'image/png', 'image/gif'])
+                            ->maxSize(1024)
+                            ->rules('dimensions:width=96,height=96')
```
Jamiras, line 339: Where are "uploaded" files stored? Do they automatically get cleaned up? I attached a wrongly sized file and it still appears to have been uploaded:
image

Also, the error message remains after uploading a different (correctly sized) image.

And uploading that generated an error:
image

What permission do I need?
If the user doesn't have permission, should the field even be there?

wescopeland: "Draft" files (for lack of a better term) are uploaded to _storage/app/temp_. Filament only tries to associate them with the game when the user clicks the form submit button.

They are indeed automatically cleaned up (see `ProcessUploadedImageAction::cleanUpTempFiles()`), but it's a side effect. Maybe it would make sense for the cleanup code to run on some kind of schedule?

The permission needed is `GamePolicy::update()`.

Visibility is indeed conditional based on this policy in _GameResource.php_ on line 347.

Authorization passes is the user has any role of GAME_HASH_MANAGER, DEVELOPER_STAFF, DEVELOPER, or DEVELOPER_JUNIOR.

The authorization call occurs in `GameResource/Pages/Edit.php` on line 20.

My user account has the `DEVELOPER` role. This is what I see when I use the control locally:

https://github.com/user-attachments/assets/0c13702e-729d-482c-a13c-e5e1c5c6c748

Jamiras: > "Draft" files (for lack of a better term) are uploaded to storage/app/temp.

It seems like it maintains the filename of the uploaded file. If two users both upload a "icon.png" at roughly the same time, will one overwrite the other?

Wrongly sized files don't appear in that folder.

Jamiras: I'm still getting the error. And I have at least two of the roles you've identified.

https://github.com/user-attachments/assets/e45b35bd-e143-4567-a1d2-fa7d42d20661

Jamiras: And the error still stays on screen when uploading a second (correctly sized) image.
image

You can see it in your video at the 25 second mark.

wescopeland:
I'm going to attack this in a separate branch using a different approach.

I think Filament's built-in file upload component enforces a few assumptions that may not make sense for how we want to handle uploading and storing these assets right now.

I'd like to try again with a simple native file input from within Filament. However, Filament doesn't support this out of the box, so I'll need to build the component by hand.

PR #2683, opened 2024-09-07
feat(achievement): credit users for logic, art, writing, and testing

wescopeland:
This PR makes it so that developers and team accounts can credit users for logic, art, writing, and testing on achievements. There are no player-facing UI elements added as part of this PR.

Achievement credit works by leveraging the `achievement_authors` table. Note that only achievement-related tasks are credited as part of this PR. For game-related tasks, a separate PR will be required which will need to leverage a different table.

A sync script, only for logic credit, has been provided and can be executed via:
```bash
sail artisan ra:sync:achievement-authors
```
Art credit is not synced as part of this script, as it's impossible to determine when a badge change occurs if the person making the change was actually the artist.

I think what may be preferable is if there's no art credit on an achievement, assume the achievement's `user_id` is the artist.

New functionality has been added to the management app for manually doing CRUD operations on achievement credit:
* On game view/edit pages (`/manage/games/1`), when rows are checked, a new bulk action appears for adding credit.
* On achievement view/edit pages (`/manage/achievements/9`), a credits table now appears, allowing authorized users to add, edit, and delete credit.

Screenshot 2024-09-07 at 2 58 52 PM

Screenshot 2024-09-07 at 2 59 07 PM

Screenshot 2024-09-07 at 2 59 12 PM

Screenshot 2024-09-07 at 2 59 24 PM


app/Models/AchievementAuthor.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        return LogOptions::defaults()
+            ->logOnly([
+                'achievement_id',
+                'user_id',
+                'task',
+            ])
+            ->logOnlyDirty()
+            ->dontSubmitEmptyLogs();
+    }
```
wescopeland, line 43: There is no way to access this currently from any UI. However, it makes sense to go ahead and start recording the data.

app/Helpers/database/achievement.php
```diff
@@ -272,6 +273,8 @@ function UploadNewAchievement(
         $achievement->save();
         $idInOut = $achievement->ID;
 
+        $achievement->upsertAuthorshipCredit($author, AchievementAuthorTask::LOGIC);
```
wescopeland, line 276: When an achievement is uploaded to the server, we immediately credit the uploader with logic credit.

Under the current design, everything else (writing, art, etc) is assumed unless someone or something adds credit for those tasks.

app/Helpers/database/achievement.php
```diff
@@ -359,6 +362,10 @@ function UploadNewAchievement(
             static_setlastupdatedgame($gameID);
             static_setlastupdatedachievement($idInOut);
 
+            if ($changingLogic) {
+                $achievement->upsertAuthorshipCredit($author, AchievementAuthorTask::LOGIC);
+            }
```
wescopeland, line 367: If a developer modifies the logic of an achievement, we also give the developer a logic task credit record.

It is easy enough to also give writing credit here if the title or description are being edited, but I decided to be a bit conservative and not do this, as it's possible someone is just fixing a typo and I wasn't sure if that constituted "writing credit". This isn't a strong opinion and I could just as easily go the other way, though.

app/Platform/Commands/SyncAchievementAuthors.php
```diff
@@ -0,0 +1,99 @@
...
+                    ->first();
+
+                if ($user) {
+                    $achievement->upsertAuthorshipCredit(
+                        $user,
+                        AchievementAuthorTask::LOGIC,
+                        backdate: $systemComment->Submitted,
+                    );
+
+                    $addedAuthors++;
```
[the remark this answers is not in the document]
Jamiras: The easiest solution to both of these problems would be to keep an array of authors who have been given credit, don't try to upsert authors that are already in the array, and count the number of entries in the array at the end of the function.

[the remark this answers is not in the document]
Jamiras: Might also just want to track the last credited author. If authorship changes, you might need the full trail:

* Date 1 - UserA created achievement - given logic credit
* Date 2 - UserA patched achievement - logic credit not updated
* Date 3 - UserB patched achievement - given logic credit
* Date 4 - UserA patched achievement - given new logic credit?

Or is that the reason that only the newest date is kept? "The logic was last updated at DateX by UserA". Is this meant to just capture the current state, and not the history? 

We currently report data for every user who has touched it (see http://localhost:64000/manage/achievements/44672). Is it more important to keep the most recent edit, or the oldest per user?

app/Platform/Commands/SyncAchievementAuthors.php
```diff
@@ -0,0 +1,99 @@
...
+
+    private function syncAchievementLogicAuthors(Achievement $achievement, bool $quiet = false): void
+    {
+        if (!$quiet) {
+            $this->info("\nUpserting logic authorship credit for achievement [{$achievement->id}:{$achievement->title}] ({$achievement->game->title}).");
+        }
+
+        $addedAuthors = 0;
+
+        $achievementAuthor = $achievement->developer()->withTrashed()->first();
```
Jamiras, line 58: I'm not certain the `withTrashed` is working properly: http://localhost:64000/manage/achievements/44672

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -215,10 +250,44 @@ public function table(Table $table): Table
...
+                                ->body('Successfully removed type from selected achievements.')
+                                ->success()
+                                ->send();
                         }),
                 ])
                     ->label('Bulk set type')
                     ->visible(fn (): bool => $user->can('updateField', [Achievement::class, null, 'type'])),
+
+                Tables\Actions\BulkAction::make('add-credit')
+                    ->label('Bulk add credit')
```
Jamiras, line 265: Is this supposed to have three vertical dots too?

image

app/Helpers/database/achievement.php
Jamiras: Same comment as below: Should the logic credit track the first edit by the user, or the last?

app/Models/Achievement.php
```diff
@@ -192,6 +194,22 @@ public function getActivitylogOptions(): LogOptions
...
+            ['created_at' => $backdate ?? now(), 'updated_at' => now()]
+        );
+    }
+
+    public function removeAuthorshipCredit(User $user, AchievementAuthorTask $task): int
+    {
+        return $this->authorshipCredits()
+            ->where('user_id', $user->id)
+            ->where('task', $task->value)
+            ->delete();
```
Jamiras, line 210: There's probably a valid reason to allow it, but should we be allowed to remove the credit entry for the user associated to the achievement? i.e. can I remove my logic credit from an achievement that I uploaded?

Jamiras: Kind of a forward-thinking question: Should dummy records be created for all tasks using the achievement author and created timestamp? It's implied that the original author is credited with the logic, artwork, writing, and testing until such time as someone else is given credit.

Instead of inserting dummy records for the all of those tasks, we could minimize the size of the table and just assume those values (this would make more sense for a user-facing UI, and not the management UI). An actual value would have to be inserted when another user is credited for historical purposes, but that could be deleted if the user didn't actually participate in that task.

app/Platform/Concerns/ActsAsDeveloper.php
```diff
@@ -73,5 +74,13 @@ public function resolvedTickets(): HasMany
         return $this->hasMany(Ticket::class, 'resolver_id', 'ID');
     }
 
+    /**
+     * @return HasMany<AchievementAuthor>
+     */
+    public function achievementAuthorshipTasks(): HasMany
```
Jamiras, line 80: I'm struggling with "authorship tasks". Tasks imply work to be done, and authorship implies writing. Two of the four tasks involve no writing.

Aren't these just "credits"?

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
wescopeland: This one shouldn't, as it doesn't open a dropdown menu.

app/Helpers/database/achievement.php
wescopeland: Personally, I think it should track the first. Updated in latest.

app/Platform/Commands/SyncAchievementAuthors.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        $systemComments = $achievement->legacyComments()->automated()->get();
+        foreach ($systemComments as $systemComment) {
+            $payload = $systemComment->Payload;
+
+            // Extract the username from the payload.
+            $username = strtok($payload, ' ');
+
+            if (str_contains($payload, 'logic')) {
```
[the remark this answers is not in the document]
wescopeland: Changed in latest to:
```php
$expectedPhrase = "edited this achievement's logic.";
if (str_ends_with($payload, $expectedPhrase)) {
```

app/Platform/Commands/SyncAchievementAuthors.php
```diff
@@ -0,0 +1,99 @@
...
+                    ->where('User', $username)
+                    ->orWhere('display_name', $username)
+                    ->first();
+
+                if ($user) {
+                    $achievement->upsertAuthorshipCredit(
+                        $user,
+                        AchievementAuthorTask::LOGIC,
+                        backdate: $systemComment->Submitted,
+                    );
```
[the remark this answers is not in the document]
wescopeland: Renamed to `->ensureAuthorshipCredit()` in latest and changed to use `->firstOrCreate()`, which seems to have resolved the issue:

Screenshot 2024-09-16 at 3 51 41 PM

app/Platform/Commands/SyncAchievementAuthors.php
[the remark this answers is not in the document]
Jamiras: Unfortunately, it's not going to be that easy:
```
[User] edited this achievement's points, logic, title, description.
```

app/Platform/Commands/SyncAchievementAuthors.php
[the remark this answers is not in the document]
wescopeland: I think it's more important to keep the oldest credit per user. For detailed trails/history, the (currently unused) `triggers` table seems like where that should go.

If the same user contributes two unique badges to the same achievement, I think it's likely sufficient that they're getting badge credit -- I don't think they'll need two credit rows, and this may lead to more complexity in building out the player-facing UX later on.

Per your suggestion, in latest I've changed `$addedAuthors` to be an array of credited authors so multiple edits by a single author aren't counted.

app/Platform/Commands/SyncAchievementAuthors.php
wescopeland: Good catch, this is now fixed in latest.

Screenshot 2024-09-16 at 3 55 24 PM

wescopeland: For non-moderators, navigating to that deleted user's page is going to lead to a forbidden access error. I'm not sure if this is worth worrying about, given the context is Filament.

app/Platform/Concerns/ActsAsDeveloper.php
wescopeland: No opposition from me in changing it. Renamed to `achievementCredits()` in latest.

app/Models/Achievement.php
wescopeland: I'm not sure why I even added this `removeAuthorshipCredit()` function, but it appears to be unused at the moment. I've deleted it in latest.

You've raised a good point; I think we need elevated privileges around soft deletion. In latest, only `DEVELOPER_STAFF`, `MODERATOR`, and `TEAM_ACCOUNT` can perform soft deletes.

wescopeland: > Should dummy records be created for all tasks using the achievement author and created timestamp? It's implied that the original author is credited with the logic, artwork, writing, and testing until such time as someone else is given credit.

Given the current implication, I lean towards no.

> Instead of inserting dummy records for the all of those tasks, we could minimize the size of the table and just assume those values (this would make more sense for a user-facing UI, and not the management UI).

Agreed - this is where I'm currently at. We can run with the assumption unless a value for some task with a different user ID is inserted.

> An actual value would have to be inserted when another user is credited for historical purposes, but that could be deleted if the user didn't actually participate in that task.

I've made an adjustment for this in latest. Now when adding non-logic credits to an achievement, the Filament relation manager will try to backfill data

app/Platform/Commands/SyncAchievementAuthors.php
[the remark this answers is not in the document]
Jamiras: It's still not perfect. All of these achievements were created in 2017 (http://localhost:64000/manage/games/1667), but most of them show a credit from 2023.
image
image

app/Models/Achievement.php
Jamiras: I'm still able to delete credits for achievements I've created.
image

app/Platform/Commands/SyncAchievementAuthors.php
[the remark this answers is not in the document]
Jamiras: This might be okay. I think I still had data in my DB from the last time I ran the command.

app/Models/Achievement.php
wescopeland: As someone with either a DEVELOPER_STAFF, MODERATOR, or TEAM_ACCOUNT role, this is expected (`AchievementAuthorPolicy::delete()`). Should we further refine these restrictions?

This is what it looks like for a developer without one of those three roles:
Screenshot 2024-09-22 at 10 59 46 AM

app/Platform/Commands/SyncAchievementAuthors.php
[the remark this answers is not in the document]
wescopeland: Re-checked and this does indeed seem okay on my local.

app/Platform/Commands/SyncAchievementAuthors.php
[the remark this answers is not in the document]
wescopeland: Updated in latest to:
```php
$expectedPhraseOne = "edited this achievement's";
$expectedPhraseTwo = "logic";
if (str_contains($payload, $expectedPhraseOne) && str_contains($payload, $expectedPhraseTwo)) {
```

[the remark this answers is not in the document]
wescopeland: Made another change in latest to resolve.

app/Models/Achievement.php
wescopeland: It's a valid concern.

In latest, the delete action is hidden for the earliest logic credit associated with the achievement.

app/Platform/Commands/SyncAchievementAuthors.php
```diff
@@ -73,11 +73,14 @@ private function syncAchievementLogicAuthors(Achievement $achievement, bool $qui
...
+            $words = explode(' ', $payload);
+            $username = array_shift($words); // Username is always the first word.
 
-            $expectedPhraseOne = "edited this achievement's";
+            $remainingPayload = implode(' ', $words);
+
+            $expectedPhraseOne = "edited";
             $expectedPhraseTwo = "logic";
-            if (str_contains($payload, $expectedPhraseOne) && str_contains($payload, $expectedPhraseTwo)) {
+            if (str_contains($remainingPayload, $expectedPhraseOne) && str_contains($remainingPayload, $expectedPhraseTwo)) {
```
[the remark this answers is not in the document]
wescopeland: Yep. Updated in latest.

suspect15:
Are achievement authors going to be given credit for each of the categories by default or is the intent only to give credit to people other than the author?

wescopeland:
> Are achievement authors going to be given credit for each of the categories by default or is the intent only to give credit to people other than the author?

When someone uploads an achievement to the server, they are by default given Logic credit. Credit for other tasks is implied.

Subsequent logic credit is automatically given if a different developer comes in and updates logic for the achievement.

Other forms of credit, such as art, must be manually granted. This is to err on the side of caution, as oftentimes the uploader for a new badge is not the artist of the badge, but doing it on a non-developer's behalf.

ladynadiad:
Would there be a way to ensure that things like art can be done in a whole batch?  It's rare for one artist to do just a few badges and would get annoying fast to have to do credits for a whole set when we have plenty of sets with 100+ achievements.  A way to edit multiple at once would be a godsend.  Maybe something like an option to apply the same art/writing/testing credit to all achievements or to select achievements it applies to?

wescopeland:
Yes, but not in this PR. The scope of this PR is already quite large; let's build that in a follow-up.

Hexadigital:
Would it be possible to include a category for Design credit? While I think this would normally line up with the current authorship/creator field, that's not always the case - I've seen some games where a user/player has put together an entire set plan, challenges and all, and posted it in the comments on the game page - and sometimes the developer who picks it up uses that as a guideline.

I don't remember the specific games at the moment, though it might be easy to find by looking for game wall comments containing Google Sheets links.

wescopeland:
Yes. That is a minor change. Added in latest:

Screenshot 2024-09-26 at 8 38 44 PM

Jamiras (approved):

PR #2684, opened 2024-09-08
Add emulators to filament console

Jamiras:
Requires the Release Manager role.

image

The seeder populates the database from the releases.php file. I've exported those tables for insertion into the production database.
[emulators.sql.txt](https://github.com/user-attachments/files/16922745/emulators.sql.txt)

The download page is now mostly built from DB records - the list of emulators, their supported consoles, the description blurb, and the documentation link come from the database.

The download and source code links still come from the releases.php file. I'll be tackling those separately in future PRs.

Also updated the create ticket page to pull the active supported emulators for the associated game's console from the DB.

This code is largely copy/pasted from other filament PRs. If I made a copy/paste error, or if I'm not doing something right, I apologize.

The ability to add/remove systems from an emulator is beyond my knowledge at this time. I'd appreciate it if someone else could add that functionality.


Jamiras:
I'm not sure if it makes sense to have emulator entries for discontinued emulators. In particular, I don't think RANester ever actually existed, and I don't think RAPCE an RAGens can log in due to outdated APIs. I could remove them from the import file, even though they are being correctly imported as inactive.

app/Filament/Resources/AchievementResource.php
wescopeland, line 1: Comment not related to this file. Related to _emulators.sql_.

Storing HTML in `emulators.description` could likely bite us later, as filtering this out using pattern matching could be [very very difficult](https://stackoverflow.com/a/1732454). Ideally, we only hold plain text in this column and leave our templates to apply document-related concerns.

wescopeland, line 1: Comment not related to this file.

As part of this work, would it also make sense for `RELEASE_MANAGER` to be able to manage systems? It seems in `SystemPolicy`, only `GAME_HASH_MANAGER` has this ability, but I'm just now noticing that doesn't quite seem right.

app/Filament/Resources/EmulatorResource.php
```diff
@@ -0,0 +1,234 @@
...
+        return $infolist
+            ->columns(1)
+            ->schema([
+                Infolists\Components\Split::make([
+                    Infolists\Components\Section::make('Details')
+                        ->columns(['xl' => 2, '2xl' => 2])
+                        ->schema([
+                            Infolists\Components\Group::make()
+                                ->schema([
+                                    Infolists\Components\TextEntry::make('handle')
```
wescopeland, line 69:
```suggestion
                    Infolists\Components\Section::make('Details')
                        ->schema([
                            Infolists\Components\Group::make()
                                ->columns(['md' => 2, 'xl' => 3, '2xl' => 4])
                                ->schema([
                                    Infolists\Components\TextEntry::make('handle')
```

I was able to get the fields to fit a little better into their container by moving some of these columns around:

Screenshot 2024-09-07 at 10 21 28 PM

app/Filament/Resources/EmulatorResource.php
```diff
@@ -0,0 +1,234 @@
...
+                                ->required()
+                                ->maxLength(255)
+                                ->helperText('Original name of emulator.'),
+                            Forms\Components\Textarea::make('description')
+                                ->label('Description')
+                                ->rows(5)
+                                ->helperText('Additional text to display on the download page.'),
+                            Forms\Components\TextInput::make('link')
+                                ->label('Documentation link')
+                                ->helperText('Link to emulator documentation.'),
```
wescopeland, line 126: Would a `->url()` validator be appropriate for this field?

app/Models/Emulator.php
```diff
@@ -106,4 +128,15 @@ public function scopeActive(Builder $query): Builder
     {
         return $query->where('active', true);
     }
+
+    /**
+     * @param Builder<Emulator> $query
+     * @return Builder<Emulator>
+     */
+    public function scopeForConsole(Builder $query, int $consoleId): Builder
```
wescopeland, line 136: nit: `scopeForSystem()` and `$systemId` may be more consistent with the `systems` relation name.

wescopeland (commented):
> The ability to add/remove systems from an emulator is beyond my knowledge at this time. I'd appreciate it if someone else could add that functionality.

No worries, I think I see how this would work. Let's get through this PR and then I'd be happy to help in a follow-up.

app/Filament/Resources/AchievementResource.php
Jamiras: That's coming directly from the releases.dist.php:
https://github.com/RetroAchievements/RAWeb/blob/753173d2fae382b4d971d719e747dbe8749ac715/storage/app/releases.dist.php#L22-L23

What's your suggestion? Use shortcode?

Jamiras: That makes sense. Adding a new emulator could also entail enabling a new system.

app/Filament/Resources/EmulatorResource.php
Jamiras: I don't like it.
image

But two columns seems okay (and matches the edit page)
image

app/Filament/Resources/EmulatorResource.php
Jamiras: yes.

app/Models/Emulator.php
Jamiras: yes.

app/Filament/Resources/EmulatorResource.php
wescopeland: Two columns looks good 👍

app/Filament/Resources/AchievementResource.php
wescopeland: A good compromise for now might be only allowing `<a>` tags and getting rid of things like `<strong>`. If we know up front that only anchor tags are allowed, we can pattern match against that limited restraint if/when we need to later.

Jamiras: I've removed all HTML from the descriptions and added a rule to prevent putting HTML into that field.

I've updated the emulators.sql attachment in the original post to reflect the changes.

wescopeland (approved):

Jamiras:
Rather than have further SQL statements to run for upgrading the seed data provided in emulators.sql, I've opted to just add the necessary migration in this PR and create an updated emulators.sql. The downside to this approach is that the migration has to be applied before the sql can be run, so the sql should be run ASAP after upgrading so the download page will be functional.

The emulators table now also includes download and source URLs, and releases are listed on the emulator details page. 

The following functionality is still missing:
* Cannot add/remove systems from emulator page
* Cannot view/edit/add releases from emulator page

Updated seed script:
[emulators.sql.txt](https://github.com/user-attachments/files/16924176/emulators.sql.txt)

wescopeland:
I seem to be having some trouble running the two SQL commands in the txt file. I have run `sail artisan migrate` and this was successful:
```
❯ sail artisan migrate

   INFO  Running migrations.

  2024_09_08_000000_update_emulators_table .................. 120.41ms DONE
```

Starting from an empty `emulators` and `system_emulators` tables, when I run the first query, I get:
```
Query 1 ERROR at Line 1: : Column count doesn't match value count at row 1
```

When I run the 2nd query, I get:
```
Query 1 ERROR at Line 3: : Cannot add or update a child row: a foreign key constraint fails (`retroachievements-web`.`system_emulators`, CONSTRAINT `system_emulators_emulator_id_foreign` FOREIGN KEY (`emulator_id`) REFERENCES `emulators` (`id`) ON DELETE CASCADE)
```

wescopeland:
Please disregard the above message: I accidentally reached for the txt file in the OP, not the one in your follow-up post.

resources/views/pages-legacy/download.blade.php
```diff
@@ -18,81 +16,88 @@
     <h2 class="mb-6">Download a supported emulator</h2>
 
     @foreach ($emulators as $emulator)
-        <h2 class="longheader" id="<?= mb_strtolower($emulator['handle'] ?? null) ?>">
-            <a href="#<?= mb_strtolower($emulator['handle'] ?? null) ?>"><?= $emulator['handle'] ?? null ?></a>
-            @if ($emulator['handle'] != $emulator['name'])
-                <small>(<?= $emulator['name'] ?? null ?>)</small>
+        @php
+            $systems = $emulator->systems()->active()->orderBy('Name')->pluck('Name')->toArray();
```
wescopeland, line 20: There seems to be an N+1 query problem here. It's not debilitating by any means, but could be worth adding an eager load somewhere while we're already in here.

Screenshot 2024-09-11 at 9 40 53 PM

Jamiras: Can you eagerly load a HasMany relationship? I thought that only works for a BelongsTo relationship.

wescopeland: Yep, you can eagerly load any relationship.

```php
$emulators = Emulator::active()
    ->orderBy('name')
    ->with(['latestRelease', 'systems' => function ($query) {
        $query->active()->orderBy('Name');
    }])
    ->get();
```

```php
@php
    $systems = $emulator->systems->pluck('Name')->toArray();
    if (empty($systems)) {
        continue;
    } else {
        // just some debugging output
        dump($systems);
    }
@endphp
```

Jamiras: I'm not seeing any difference.

Before:
image

With your changes:
image

It's just queries for the systems sooner.

It does seem to use one less query, but generates twice as many models.

Jamiras: I thought I had done that, but where you have

```
 $systems = $emulator->systems->pluck('Name')->toArray();
```

I had

```
 $systems = $emulator->systems()->pluck('Name')->toArray();
```

Removing those extra parenthesis did drop the number of queries, but the overall time was unchanged (still ~150ms)

wescopeland: Yep, not debilitating by any means as the queries themselves are very fast, there's just no reason in this case to run them in a waterfall.

The `->relations` vs `->relations()` thing trips me up often as well.

Jamiras: Oh, that's page load time. Query time dropped from 6.6ms to 5.3ms. 

Changes applied and committed.

wescopeland (approved):
Verified locally.

Jamiras:
I'm wondering if it would make sense to separate this into two PRs: One for the migration step, so that can be deployed and the SQL ran without affecting anything, and then the other for the management page and Downloads conversion.

wescopeland:
That is totally up to your judgment; I would be happy with either approach.

Jamiras:
I'm going to do it. This isn't blocking any feature work that can't wait an extra week or two, and the data isn't volatile enough to worry about keeping in sync.

Even if things are only broken for a couple minutes, I just don't feel right knowingly deploying something that will generate errors.

wescopeland:
Understood, and that is very fair. We could also do a staggered release if you'd like? Have changeset #1 sitting in `master` ready to go, then an hour after the release, merge & deploy changeset #2. No opposition whatsoever from me on taking that approach if it's palatable.

PR #2685, opened 2024-09-08
chore: move features/user/settings to features/settings

wescopeland:
No functional changes, just some housekeeping. This PR just moves that "settings" folder one layer up.

Now that the user settings & forum posts branches are both merged and I'm seeing both in the file tree, it's confusing to have _features/user_ and _features/users_ directories.


Jamiras (approved):
Settings page still seems functional.

PR #2686, opened 2024-09-12
feat: add BuildGameListAction

wescopeland:
This PR adds a new `BuildGameListAction` class in the `App\Platform\Actions` namespace.

This action is designed to be idempotent and provides context-agnostic game list records, ideal for use in Laravel controllers, such as those constructing Inertia.js reviews for a React.js datatable.

Key features of the action include:
* Support for multiple contexts (eg: Want to Play Games, Want to Develop Games, All Games, System Games, Hubs, and Developed Sets lists). Currently, only the Want to Play Games context is implemented.
* Sorting by any field in each record.
* Filtering by various fields, with support for applying multiple filters simultaneously.
* Reporting the user's progress and awards for each game in the list.
* Performance efficiency, with execution times ranging between 40-70ms under most circumstances, with potential for further optimization.
* Automatically reporting if a game is in the user's Want to Play Games list.
* Elimination of N+1 query issues.
* Nearly 100% statement test coverage.
* End-to-end type safety, with results fed into a `PaginatedData` DTO, similar to how the forum Recent Posts page functions.

**How to test this PR:**
In addition to comprehensive PHPUnit test coverage, a "playground" page has been set up for testing the action.

This page can be visited at http://localhost:64000/demo/playground.

In _playground.blade.php_, you'll find:
```php
render(function (View $view) {
    $user = User::firstWhere("User", "Jamiras");
    $action = new BuildGameListAction();

    $result = $action->execute(
        GameListType::UserPlay,
        $user,
        perPage: 50,
        sort: ["field" => "title", "direction" => "asc"], // see $validSortFields in BuildGameListAction for possible sorts
        filters: [], // eg: ['system' => [1]] or ['system' => [1, 5]] or ['award' => ['completed', 'mastered']]
    );

    return $view->with(['result' => $result]);
});
```

You can use this playground to experiment with various parameters of the action.

Currently, the playground page is the only place in RAWeb's UI where this action is exposed. This PR aims to get the back-end logic reviewed. A separate front-end PR will follow to implement the React.js datatable that will consume the action results. Splitting the PRs this way ensures logical separation between back-end and front-end components.


app/Platform/Actions/BuildGameListAction.php
```diff
@@ -0,0 +1,582 @@
...
+        // We support multiple filters, but only a single selected sort.
+        $this->applyFilters($query, $filters, $user);
+        $this->applySorting($query, $sort, $user);
+
+        // The base query may not respect ->distinct() operations done by some filters.
+        // We'll override its `total` value with the correct one.
+        $total = $query->count('GameData.ID');
+
+        /** @var LengthAwarePaginator<Game> $entries */
+        $entries = $query->paginate($perPage, ['*'], 'page', $page);
```
Jamiras, line 52: I'm not sure exactly where the difference is coming from, but this generates a massive query that takes 45.8ms to run (just the one query - I'm ignoring the additional 6.7ms to get the playerGames and backlogGames), whereas the equivalent queries on the Want to Play page take 12.98ms (2.58+4.63+0.31+3.33+2.13)

tests/Feature/Platform/Action/BuildGameListActionTest.php
```diff
@@ -0,0 +1,740 @@
...
+            'players_total' => 18,
+            'players_hardcore' => 17,
+            'achievements_published' => 38,
+            'points_total' => 240,
+            'TotalTruePoints' => 282,
+            'Updated' => Carbon::parse('2023-06-06'),
+        ]);
+    }
+
+    private function addAllGamesToUserPlayList(User $user, array $gameIds = []): void
```
Jamiras, line 733: This should just be `addGamesToUserPlayList`, `all` implies that all existing games would be added.

Also, why is the default an empty array? It doesn't branch into any sort of default behavior, so calling the method without providing a list of `gameIds` would do nothing, so there's no reason to call it without a list of `gameIds`.

app/Platform/Data/GameData.php
```diff
@@ -28,7 +37,15 @@ public static function fromGame(Game $game): self
...
-            system: Lazy::create(fn () => SystemData::fromSystem($game->system))
+            system: Lazy::create(fn () => SystemData::fromSystem($game->system)),
+            achievementsPublished: Lazy::create(fn () => $game->achievements_published),
+            pointsTotal: Lazy::create(fn () => $game->points_total),
+            pointsWeighted: Lazy::create(fn () => $game->TotalTruePoints),
+            releasedAt: Lazy::create(fn () => $game->released_at),
+            releasedAtGranularity: Lazy::create(fn () => $game->released_at_granularity),
+            lastUpdated: Lazy::create(fn () => $game->lastUpdated),
+            numVisibleLeaderboards: Lazy::create(fn () => $game->num_visible_leaderboards ?? 0),
+            numUnresolvedTickets: Lazy::create(fn () => $game->num_unresolved_tickets ?? 0),
```
Jamiras, line 48: This ties into another comment I had: Why are all of these lazy if the data is simple data types that are readily available? The only one that looks potentially complex is `system`, but I'm pretty sure that data is readily available at this point too.

app/Platform/Actions/BuildGameListAction.php
```diff
@@ -0,0 +1,582 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Platform\Actions;
```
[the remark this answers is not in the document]
wescopeland: Actions are stateless, self-contained, and have a single responsibility. The only way for the outside world to interact with an action is through its `->execute()` method.

Services operate at a higher level of abstraction than actions. They may have multiple ways for the outside world to interact with them and/or consume their data.

With as much complexity as there could be for building a datatable for a React front-end, I opted to use the action pattern to try to keep things as simple, predictable, and testable as possible.

The action is idempotent: calling `->execute()` with the same parameters should yield the same result every time, assuming the underlying data is unchanged.

app/Platform/Actions/BuildGameListAction.php
wescopeland: It's coming from `BuildGameListAction::applyGameTitleSorting()`.

`GameListService` does the sorting in memory with PHP, whereas `BuildGameListAction` is doing it with SQL. The bottleneck can be proven out by adding a `return;` to the top of `applyGameTitleSorting()`. Doing so drops the query time on my machine to 2ms.

I have explored doing the sort in-memory in this action similar to how `GameListService` does it, but doing so implies that the sort needs to happen before pagination. I'm very concerned about the long-term scalability implications of doing this.

I think what might be the best approach is to compute `SortPrefix` and SortTitle` in advance and store them directly on the `GameData` table as new columns. I may tackle that in a separate PR.

app/Platform/Actions/BuildGameListAction.php
```diff
@@ -0,0 +1,582 @@
...
+                    break;
+
+                /*
+                 * only show games matching a specific list of user earned awards
+                 */
+                case 'award':
+                    $this->applyAwardFilter($query, $filterValues, $user);
+                    break;
+
+                default:
```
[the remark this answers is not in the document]
wescopeland: New filter ("hasAchievementsPublished") + coverage added in latest. Here's an example test case:

```php
public function testItCanFilterByHavingSomeAchievementsPublished(): void
{
    // Arrange
    $user = User::factory()->create();

    $this->seedGamesForLists();
    $this->addGamesToUserPlayList($user, gameIds: [1000, 1001, 1002, 1003, 1004, 1005]);

    // Act
    $result = (new BuildGameListAction())->execute(
        GameListType::UserPlay,
        $user,
        filters: ['hasAchievementsPublished' => ['has']]
    );

    // Assert
    $this->assertEquals(5, $result->total);
    $this->assertEquals(5, count($result->items));
}
```

app/Platform/Actions/BuildGameListAction.php
Jamiras: Consider #1062 before you tackle that.

tests/Feature/Platform/Action/BuildGameListActionTest.php
wescopeland: Renamed to `addGameIdsToUserPlayList` in latest and removed the default empty array definition.

resources/views/pages/demo/playground.blade.php
```diff
@@ -0,0 +1,31 @@
...
+        filters: [], // eg: ['system' => [1]] or ['system' => [1, 5]] or ['award' => ['completed', 'mastered']]
+    );
+
+    return $view->with(['result' => $result]);
+});
+
+?>
+
+<x-app-layout>
+    @dump($result)
```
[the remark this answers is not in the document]
wescopeland: No. For all intents and purposes, `Lazy` is a view concern.

`Spatie\LaravelData\Lazy` gives a way for anything utilizing these `Data` DTOs to "opt out" of sending that information to a consumer if they want to. In other words, you can pick exactly what fields are needed for whatever is going to be using those fields further downstream, even if the back-end logic / SQL is fetching for or has access to more.

This is particularly useful when using Inertia.js, as during the server-side render, the page props are injected directly into the body even if they aren't actually used.

Screenshot 2024-09-12 at 6 40 09 PM

These particular lazy values you're pointing out are coming from _GameData.php_. It does stand to reason there are plenty of pages that won't actually need many of these fields but may instead need some other "more shallow" data from the game. With `Lazy`, we can accomplish this.

```php
// from BuildGameListAction.php
game: GameData::from($game)->include(
    'system.nameShort',
    'system.iconUrl',
    'achievementsPublished',
    'pointsTotal',
    'pointsWeighted',
    'releasedAt',
    'releasedAtGranularity',
    'lastUpdated',
    'numVisibleLeaderboards',
    'numUnresolvedTickets',
),
```

`->include()` is saying "the consumer needs these lazy fields, so include them in the data".

app/Platform/Data/GameData.php
wescopeland: Tried to do a bit of explanation in my comment here https://github.com/RetroAchievements/RAWeb/pull/2686#discussion_r1757679671 but want to add another follow-up on this comment.

A logical question might be: "why use these `Data` classes at all?" The primary benefit we get from them is end-to-end type safety, something we definitely do not have in our Blade UIs right now (see: mistyped routes causing pages to crash, us needing to constantly comment on prop types, etc)

You can prove this out very easily. Open _GameData.php_ and _generated.d.ts_. Remove any of those `Lazy|` from any of the fields and save the file. In the terminal, run `composer types`. _generated.d.ts_ will update with the new correct type for the UI to use, and if the UI is mistyped, `tsc` will fail.

app/Platform/Actions/BuildGameListAction.php
wescopeland: Thanks for sharing that.

Seems like it's almost the same thing. All games will have an autogenerated sort title, but it could manually be overwritten perhaps in the management app?

wescopeland: `sort_title` implementation is in https://github.com/RetroAchievements/RAWeb/pull/2690.

Jamiras (approved):
No further feedback on this. I'll probably have something once it's integrated into something.

PR #2687, opened 2024-09-12
fix(messages): replace [user=ID] in email/discord previews with @username

wescopeland:
Webhook and email previews for inbox messages are currently showing `[user=ID]` rather than the username (see [here](https://discord.com/channels/476211979464343552/1283654701379747904)).

This PR adds an action which finds these shortcodes in the message content and swaps in the usernames.


app/Community/Listeners/NotifyMessageThreadParticipants.php
```diff
@@ -73,6 +74,8 @@ private function forwardToDiscord(
         MessageThread $messageThread,
         Message $message
     ): void {
+        $message->body = (new ReplaceUserShortcodesWithUsernamesAction())->execute($message->body);
```
Jamiras, line 77: Shouldn't this just be reusing whatever thing is already doing this for recent forum posts? Then we'd get substitution for embedded achievements and games too. (Maybe it makes sense to ignore achievements for webhook so the user can get to the appropriate achievement. Are links a possibility?)

image

Also, seems like the edit post function isn't converting the user link back to a name.

image

wescopeland: Took care of both in latest.

The new action is still retained for the Edit post form, but it's isolated only to that component now. Elsewhere, we reuse the `Shortcode` helper method.

Jamiras (approved):
Verified private messages and editing. Assuming discord integration is fine based on changes matching private messaging.

PR #2688, opened 2024-09-12
fix(manage): improve management of invisible leaderboards

wescopeland:
Makes two changes to leaderboard management:

1. Display orders can now be manually entered to allow setting a leaderboard to be invisible (this can't be done from manual reordering).

Screenshot 2024-09-12 at 9 02 08 AM

2. When using drag+drop to reorder leaderboards, rows with a < 0 DisplayOrder will not automatically be adjusted on each drop action (so they stay invisible).

Screenshot 2024-09-12 at 9 02 45 AM


app/Filament/Resources/LeaderboardResource.php
```diff
@@ -124,6 +126,11 @@ public static function form(Form $form): Form
 
                         Forms\Components\TextInput::make('Description')
                             ->maxLength(255),
+
+                        Forms\Components\TextInput::make('DisplayOrder')
+                            ->numeric()
+                            ->helperText("If set to -1, the leaderboard will be invisible to regular players.")
```
Jamiras, line 132: I just wanted to point out that some users have ordered their hidden achievements:
image

It seems like this still works (they can't be reordered, and don't get updated when trying to reorder visible leaderboards. You can also still assign other negative values in the editor.

Maybe this comment needs updated? I think it's probably fine the way it is.

Jamiras (approved):

app/Filament/Resources/LeaderboardResource.php
wescopeland: I think it's a fair point to consider. I've updated the label to read "If set to less than 0,...".

Jamiras (approved):

PR #2691, opened 2024-09-14
automatically subscribe user to game wall and official forum topic when making primary claim

Jamiras:
https://discord.com/channels/310192285306454017/310195377993416714/1283896946402267148


app/Helpers/database/forum.php
```diff
@@ -438,10 +439,18 @@ function authorizeAllForumPostsForUser(User $user): bool
 
 function isUserSubscribedToForumTopic(int $topicID, int $userID): bool
 {
-    return isUserSubscribedTo(
-        SubscriptionSubjectType::ForumTopic,
-        $topicID,
-        $userID,
-        "SELECT 1 FROM ForumTopicComment WHERE ForumTopicID = $topicID AND author_id = $userID"
-    );
+    $explicitSubcription = Subscription::where('subject_type', SubscriptionSubjectType::ForumTopic)
```
wescopeland, line 442: nit: typo here, `$explicitSubscription`

app/Helpers/database/forum.php
```diff
@@ -438,10 +439,18 @@ function authorizeAllForumPostsForUser(User $user): bool
...
+        ->where('subject_id', $topicID)
+        ->where('user_id', $userID)
+        ->first();
+
+    if ($explicitSubcription) {
+        return $explicitSubcription->state == 1;
```
wescopeland, line 448: In the `Subscription` model, I see:
```php
protected $casts = [
    'state' => 'boolean',
];
```

And when I `dd($explicitSubscription->toArray())`, I see:
```
array:7 [▼ // app/Helpers/database/forum.php:448
  "id" => 65828
  "subject_type" => "ForumTopic"
  "subject_id" => 25399
  "user_id" => 117089
  "state" => false
  "created_at" => "2024-09-15T12:00:44.000000Z"
  "updated_at" => "2024-09-15T12:00:44.000000Z"
]
```

I think it's probably safer to change this line to:
```php
return $explicitSubscription->state === true;
```

app/Helpers/database/forum.php
```diff
@@ -438,10 +439,18 @@ function authorizeAllForumPostsForUser(User $user): bool
...
+    }
+
+    $implicitSubscription = ForumTopicComment::where('ForumTopicID', $topicID)
+        ->where('author_id', $userID)
+        ->exists();
+
+    return $implicitSubscription;
```
wescopeland, line 455: nit: Seems like this can just `return` rather than needing a variable (might also be more clear, as `$implicitSubscription` isn't a subscription - it's a boolean):

```php
return ForumTopicComment::where('ForumTopicID', $topicID)
    ->where('author_id', $userID)
    ->exists();
```

app/Helpers/database/subscription.php
```diff
@@ -25,74 +26,16 @@ function updateSubscription(string $subjectType, int $subjectId, int $userId, bo
...
-    $dbResult = s_mysql_query($query);
-    if (!$dbResult) {
-        return false;
-    }
-
-    $isSubscribed = mysqli_num_rows($dbResult) > 0;
-    mysqli_free_result($dbResult);
-
-    return $isSubscribed;
+    return $explicitSubcription && ($explicitSubcription->state == 1);
```
wescopeland, line 38: `->exists()` tends to be more performant than `->first()`, as it only checks existence and doesn't fetch a record/return a record.

```php
return Subscription::where('subject_type', $subjectType)
    ->where('subject_id', $topicID)
    ->where('state', true)
    ->exists();
```

app/Helpers/database/subscription.php
```diff
@@ -281,19 +224,19 @@ function isUserSubscribedToArticleComments(int $articleType, int $articleID, int
...
+    if ($explicitSubcription) {
+        return $explicitSubcription->state == 1;
+    }
+
+    $implicitSubscription = Comment::where('ArticleType', $articleType)
+        ->where('ArticleID', $articleID)
+        ->where('user_id', $userID)
+        ->exists();
+
+    return $implicitSubscription;
```
wescopeland, line 241: Similar minor refactors as above can make this more concise and not need to fetch a record:

```php
function isUserSubscribedToArticleComments(int $articleType, int $articleID, int $userID): bool
{
    $subjectType = SubscriptionSubjectType::fromArticleType($articleType);

    if ($subjectType === null) {
        return false;
    }

    // existence check only here
    $doesExplicitSubscriptionExist = Subscription::where('subject_type', $subjectType)
        ->where('subject_id', $articleID)
        ->where('user_id', $userID)
        ->where('state', true) // we can use `true` instead of `1` due to `$casts` in the model
        ->exists();

    if ($doesExplicitSubscriptionExist) {
        return true;
    }

    // immediate return without a variable
    return Comment::where('ArticleType', $articleType)
        ->where('ArticleID', $articleID)
        ->where('user_id', $userID)
        ->exists();
}
```

app/Helpers/database/forum.php
Jamiras: fixed

app/Helpers/database/forum.php
Jamiras: Didn't think to see if the model coerced the type. Was just looking at the database. Updated.

app/Helpers/database/forum.php
Jamiras: I wanted to describe why it was querying the ForumTopicComment table, and this seemed to pair better with the previous query. I guess a comment might be more appropriate.

app/Helpers/database/subscription.php
Jamiras: For this particular query it seems to be about 0.02ms on my machine, but it makes sense, so I've made the change anyway.

app/Helpers/database/subscription.php
Jamiras: Fetching the Subscription is required for the first query as the use can explicitly unsubscribe from something they're implicitly subscribed to. In that case, the record would exist with a state of 0.

I've changed the second query to use exists.

app/Helpers/database/forum.php
wescopeland: I like the comment more 👍

wescopeland (approved):

PR #2692, opened 2024-09-15
chore: make Inertia SSR port configurable via .env

wescopeland:
Still uses 13714 as the default, but can now be set via `VITE_INERTIA_SSR_PORT`.

This change should theoretically allow us to resolve an issue where the Inertia SSR daemon is using the same port in all 3 of our server's active environments.

**How to test this PR:**
In your _.env_, add the following line:
```
VITE_INERTIA_SSR_PORT=13715
```
Then:
```
sail npm run build
```

The Inertia instance should start in your terminal and report that it's running on port 13715. You should be able to navigate to `/game/1/hashes` and not see any console errors.

---

I have the Inertia daemons currently disabled in both preview and stage. After this PR is merged, we can probably try to re-enable them again and see how it goes. **We may need to be prepared to restart nginx again when this happens if something goes awry.**


config/inertia.php
```diff
@@ -23,7 +23,7 @@
 
         'enabled' => true,
 
-        'url' => 'http://127.0.0.1:13714',
+        'url' => 'http://127.0.0.1:' . env('VITE_INERTIA_SSR_PORT', 13714),
```
Jamiras, line 26: Suggest adding an entry in `.env.sample`.

wescopeland: Wasn't sure whether to do this or not -- updated in latest.

.env.example
```diff
@@ -47,6 +47,8 @@ LOG_LEVEL=debug
 OCTANE_SERVER=swoole
 OCTANE_HTTPS=false
 
+#VITE_INERTIA_SSR_PORT=13714
```
Jamiras, line 50: I expected this to be uncommented, but I guess it shouldn't really matter. It's hard to tell from looking at the rest of the file what should and shouldn't be commented out.

Jamiras (approved):

PR #2694, opened 2024-09-17
fix(forum): remediate recent posts hydration error

wescopeland:
This PR resolves a rare hydration error that is occurring on `/forums/recent-posts`.

The error can only be observed when both of these statements are true:
* A post was made within the last 60 seconds.
* The user is on a slow network connection (ie: mobile).

When the hydration error occurs, the page flashes and then reloads client-side. I have never been able to reproduce this locally, but I can reproduce it sometimes on prod.

**Root Cause**
When displaying a server-rendered React page, there are always two initial render cycles:

1. Server-side rendering (SSR): The initial HTML is generated and sent from the server to the browser. The user sees a fully-rendered page without needing to wait for JS to load. However, the page has no interactivity.
2. Client-side hydration: Once the JS bundle is loaded, React "hydrates" the static HTML by linking the pre-rendered DOM elements with the React components, making the page interactive. This process ensures the initial server-rendered markup matches what React expects.

When the server render hits the user's browser, the static HTML may contain something like "5 seconds ago".

On a slower connection, client-side hydration may take some time as JS assets are downloaded. If a full second passes and the client thinks "6 seconds ago" should be displayed, React panics and remounts the entire component tree.

We can't have these time measurements "drift" if we want to ensure hydration is stable.

Drifting _may_ still occur on larger time increments, however this should be much more rare.


resources/js/utils/diffForHumans.ts
```diff
@@ -5,10 +5,8 @@ export function diffForHumans(date: string, from?: string): string {
   const end = dayjs(date);
   const diff = Math.abs(start.diff(end, 'second'));
 
-  if (diff < 1) {
-    return 'now';
-  } else if (diff < 60) {
-    return `${Math.floor(diff)} second${diff === 1 ? '' : 's'} ${start.isAfter(end) ? 'ago' : 'from now'}`;
+  if (diff < 60) {
+    return 'A few seconds ago';
```
Jamiras, line 9: Doesn't this just shift the window such that something at diff==59 would send "A few seconds ago", but the hydration would see "1 minute ago"?

Also 59 is not "a few". Maybe we should have "less than 1 minute" for 6-59 seconds?

wescopeland: Yep, you're right. I've landed on a more robust solution in latest via a custom hook:
```ts
export function useServerRenderTime() {
  // `useState` runs on the server, so this value is populated during SSR.
  const [renderedAt, setRenderedAt] = useState(() => dayjs.utc().toISOString());

  // `useEffect` runs only on the client. The `renderedAt` state will try to
  // "drift" during hydration. Have the effect pin the value.
  useEffect(() => {
    setRenderedAt(renderedAt);
  }, [renderedAt]);

  return { renderedAt };
}
```

For all intents and purposes, this forces the value to never drift during hydration.

Additionally, I've adjusted the windows. <10s now reads "Just now", and 10-59s reads as "Less than a minute ago". Let me know if you think these windows or window labels should be further tweaked.

Jamiras (approved):

PR #2696, opened 2024-09-19
feat(manage): have table rows navigate to edit pages instead of view

wescopeland:
Reduces a click.

Now when clicking on a table row while viewing game, achievement, or leaderboard resources, if the user has permissions to edit the row, they'll be navigated to the edit page on click instead of view.

### DIFF
--- a/app/Filament/Resources/GameResource.php
+++ b/app/Filament/Resources/GameResource.php
@@ -333,6 +333,9 @@ class GameResource extends Resource
 
     public static function table(Table $table): Table
     {
+        /** @var User $user */
+        $user = Auth::user();
+
         return $table
             ->columns([
                 Tables\Columns\ImageColumn::make('badge_url')
@@ -445,6 +448,11 @@ class GameResource extends Resource
                         ->icon('fas-clock-rotate-left'),
                 ]),
             ])
+            ->recordUrl(
+                fn (Game $record) => $user->can('update', $record)
+                    ? GameResource::getUrl('edit', ['record' => $record])
+                    : GameResource::getUrl('view', ['record' => $record])
+            )
             ->bulkActions([
                 Tables\Actions\BulkActionGroup::make([
                     // Tables\Actions\DeleteBulkAction::make(),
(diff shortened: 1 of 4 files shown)


app/Filament/Resources/GameResource.php
```diff
@@ -445,6 +448,11 @@ public static function table(Table $table): Table
                         ->icon('fas-clock-rotate-left'),
                 ]),
             ])
+            ->recordUrl(
+                fn (Game $record) => $user->can('update', $record)
+                    ? GameResource::getUrl('edit', ['record' => $record])
+                    : GameResource::getUrl('view', ['record' => $record])
+            )
```
Jamiras, line 455: I'm not sure there's a correct solution for this. Some users might be going to the game specifically to edit it, others might be looking for summary information.

There's information on the non-edit page (like metrics) which may be what the user is looking for, so redirecting them to the edit page becomes extra work. Also, it's less clear that clicking "details" will leave the edit page. At least, it's clear that clicking "edit" will go to the edit page.

Similarly, Should the achievements/leaderboards/hashes/code notes be present on the edit game page since you're able to edit them without opening the edit page?

app/Filament/Resources/LeaderboardResource.php
```diff
@@ -257,6 +267,11 @@ public static function table(Table $table): Table
                         ->icon('fas-clock-rotate-left'),
                 ]),
             ])
+            ->recordUrl(
+                fn (Leaderboard $record) => $user->can('update', $record)
+                    ? LeaderboardResource::getUrl('edit', ['record' => $record])
+                    : LeaderboardResource::getUrl('view', ['record' => $record])
```
Jamiras, line 273: The most likely reason to go to the leaderboard management page is to edit the entries, which ties in to my other comment - since that's doable on the details page, should the user be immediately directed to the edit page?

Jamiras (commented):
I may be in the minority here, but I feel that automatically taking me to the edit page is not what I would expect to happen when I click a record from a row. Even if I do plan to edit the record, having the details pop up first serves as validation that I'm editing the correct record.

wescopeland:
> Even if I do plan to edit the record, having the details pop up first serves as validation that I'm editing the correct record.

I agree with this.

I think it may be best to close this for now. This PR was indeed motivated by user feedback, but unless multiple users are clamoring for this, it may be best to do nothing for the time being.

PR #2701, opened 2024-09-19
perf(getAchievementsList): optimize query

wescopeland:
This PR optimizes an incredibly slow query that can occur on `/achievementList.php?s=4&p=2`.

To repro the slow query:
* Check out the `master` branch.
* Open _app/Helpers/database/achievement.php_.
* On line 35, hardcode the user ID to 2452:
```php
$innerJoinPlayerAchievements = "LEFT JOIN player_achievements AS pa ON pa.achievement_id = ach.ID AND pa.user_id = " . 2452;
```
* Navigate to http://localhost:64000/achievementList.php?s=4&p=2

You should observe a query that takes approximately 14 to 20 seconds to resolve.

If you try again on this branch, the query should take 600ms - 1000ms.


app/Helpers/database/achievement.php
```diff
@@ -27,55 +27,77 @@ function getAchievementsList(
...
-    // 1 = my unlocked achievements
-    // 2 = achievements i haven't unlocked
-    // 3 = official
-    // 5 = unofficial
-    if ($params == 1) {
-        $query .= "AND pa.unlocked_at IS NOT NULL ";
-    } elseif ($params == 2) {
-        $query .= "AND pa.unlocked_at IS NULL ";
-    }
+                ach.ID,
```
Jamiras, line 63: Can remove the "slow query comment"?

app/Helpers/database/achievement.php
```diff
@@ -27,55 +27,77 @@ function getAchievementsList(
...
 
     // We can't run a sort on a user's achievements AwardedDate
     // if we don't have a user. Bail from the sort.
     if (($sortBy == 9 || $sortBy == 19) && !$user) {
         $sortBy = 0;
     }
 
+    if ($params === 1) {
+        // Achievements the user has unlocked.
+        $bindings['userId'] = $user->id;
```
Jamiras, line 46: Does this need a null check? `$user` is nullable, but this function only appears to be called from `achievementList.blade.php`, which forces `$params` to 3 if `$user` is null.

Jamiras (approved):

app/Helpers/database/achievement.php
wescopeland: I considered this but intentionally left it in. To be honest, I'd say at 600ms - 1s, it's still slow, just considerably less so.

app/Helpers/database/achievement.php
wescopeland: Doesn't hurt to add. The function probably shouldn't be at the mercy of its caller to pass args correctly. Added an `isset()` to the conditional in latest.

PR #2704, opened 2024-09-21
Convert claim actions to controller

Jamiras:
Additionally, makes the following improvements:
* If the primary claim for a collaboration is dropped, the oldest active collaboration claim will be changed to primary.
* If the primary claim for a collaboration is completed, all active collaboration claims will also be completed.


app/Community/Actions/UpdateGameClaimAction.php
wescopeland, line 1: nit: Where possible, all `!=` should probably be changed to `!==` to lean in to improved type safety.

`!=` may imply to a reader that those ints may be casted somewhere as a string.

app/Community/Actions/UpdateGameClaimAction.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        /** @var User $currentUser */
+        $currentUser = Auth::user();
+
+        $newValues = $request->validated();
+
+        $auditMessage = "{$currentUser->display_name} updated {$claim->user->display_name}'s claim.";
+
+        if (array_key_exists('type', $newValues)) {
+            $newType = (int) $newValues['type'];
```
wescopeland, line 32: It may be safer to lift up `int` casts to `UpdateGameClaimRequest`. This way, if the code ever gets refactored, whoever uses the request doesn't need to remember to manually perform the cast.

```php
// in UpdateGameClaimRequest.php

public function rules(): array
{
    // ... stuff here ...
}

protected function passedValidation(): void
{
    $this->merge([
        'status' => $this->status ? (int) $this->status : null,
        'special' => $this->special ? (int) $this->special : null,
        'type' => $this->type ? (int) $this->type : null,
        'set_type' => $this->set_type ? (int) $this->set_type : null,
    ]);
}
```

app/Community/Controllers/AchievementSetClaimController.php
wescopeland, line 1: Per https://laravel.com/docs/11.x/controllers#actions-handled-by-resource-controllers, I think we may want to use different names for the current `create()` and `store()` methods.

Conventionally, `create()` is a GET to return some UI responsible for POSTing a resource. I think in this case, `create()` should be renamed to `store()`.

To align with convention, the existing `store()` should be renamed to `update()`.

In other words:
`AchievementSetClaimController::create()` --> `AchievementSetClaimController::store()`
`AchievementSetClaimController::store()` --> `AchievementSetClaimController::update()`

app/Community/Controllers/AchievementSetClaimController.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->authorize('create', [AchievementSetClaim::class]);
+
+        $claim = $action->execute($game);
+
+        if (!$game->ForumTopicID && $this->authorize('update', $game)) {
+            generateGameForumTopic(Auth::user(), $game->ID);
+        }
+
+        return back()->with('success', $this->resourceActionSuccessMessage('claim',
+            $claim->Created == $claim->Updated ? 'create' : 'update'));
```
wescopeland, line 32: nit:
```suggestion
            $claim->Created === $claim->Updated ? 'create' : 'update'));
```

app/Community/Actions/UpdateGameClaimAction.php
Jamiras: done

app/Community/Actions/UpdateGameClaimAction.php
```diff
@@ -0,0 +1,131 @@
...
+
+    private function processCompletedClaim(AchievementSetClaim $claim, User $currentUser): void
+    {
+        addArticleComment("Server", ArticleType::SetClaim, $claim->game_id, "Claim completed by {$currentUser->display_name}");
+
+        // also complete any collaboration claims
+        $game = Game::find($claim->game_id);
+        $game->achievementSetClaims()->active()->update(['Status' => $claim->Status, 'Finished' => $claim->Finished]);
+
+        // TODO: these emails should be queued and sent asynchronously
```
wescopeland, line 106: I've tried various ways to optimize this locally, and I think I've landed at the same place as this comment.

It seems like we either need a bulk email send method, which I'm not even sure how to configure and/or implement, or we probably need a dedicated queue for these specific emails.

Either way, it's not a behavior change from what's happening in `master`.

wescopeland (approved):
Verified locally. Thanks for working through the feedback.

PR #2706, opened 2024-09-22
feat: implement React datatable for Want to Play Games List

wescopeland:
This PR implements the new React-driven datatable for the Want to Play Games List. The design and functionality are based on collecting numerous pieces of feedback from many users. The front-end and back-end both have nearly 100% code coverage.

Additionally, several primitive common components are implemented, such as game titles w/ tags. Like the back-end `BuildGameListAction`, the table is designed to be reusable in multiple contexts. However, all of the components currently live in a feature module and will be migrated out in a subsequent PR.

### How to test this PR
```
sail npm install
sail npm run build
```
Via the navbar, navigate to the Want to Play Games page.

## UX Overview

The datatable consists of three logical sections: the toolbar at the top, the table itself, and pagination at the bottom.

Screenshot 2024-09-22 at 8 56 50 AM

The default view consists of Title, System, Achievements, Points, Rarity, Release Date, and Progress columns.

Screenshot 2024-09-22 at 8 58 57 AM

All columns are sortable, and the sort labels are context-aware:

Screenshot 2024-09-22 at 9 00 35 AM

The table uses an aggressive prefetching and caching strategy. Combined with the innate speed of `BuildGameListAction`, this gives an illusion that, under stable network conditions, data fetches are instantaneous.

https://github.com/user-attachments/assets/13bb6e5a-e188-48a0-9c3f-e3aba6a31a84

Additionally, games can now be removed from the list via an in-line action. If the user makes a mistake, they can undo the removal.

https://github.com/user-attachments/assets/883a052d-5f0e-41f8-8d32-91b5d33f4d1e

The toolbar has several features that let the user customize what they're seeing. All columns can be toggled in and out of view. If the user can develop achievements, they also have access to the "Tickets" column.

Screenshot 2024-09-22 at 9 10 04 AM

Searching is also supported. For convenience, a desktop hotkey of "/" is included to quickly focus into the search field.

https://github.com/user-attachments/assets/2305a4c2-558d-41a6-8b74-422040ec25e6

There are two faceted filters on the table: "System" and "Has achievements". The list of systems available is sent to the page as a prop, and is based on what systems the user has in their Want to Play Games list:

Screenshot 2024-09-22 at 9 15 09 AM

When a filter is enabled, the games count in the toolbar changes to inform the user how narrowed their set of results is. A reset button also appears to let the user quickly clear all their filters.

Screenshot 2024-09-22 at 9 16 00 AM

The "Progress" column uses a much leaner component than what's present on other datagrids. Award symbols are much more subdued to hopefully portray less clutter on the screen. Useful metadata appears on either hover, mobile tap, or keyboard focus.

https://github.com/user-attachments/assets/79816b7a-044e-463d-a812-946713a3edce

---

**Next steps:**
* Reuse the table in additional contexts.
* Investigate virtual scrolling for pages that have long lists, ie: system game pages.
* Build a "custom views" feature, allowing users to save their column visibility setting(s).


.env.example
```diff
@@ -39,6 +39,7 @@ APP_URL=http://localhost:${APP_PORT}
 #INTERCEPTOR_CONNECT=
 
 #DEBUGBAR_ENABLED=false
+#VITE_REACT_QUERY_DEVTOOLS_ENABLED=false
```
wescopeland, line 42: The queries and mutations on the page are complex enough that I needed to reach for the query devtools to fully wrap my brain around what's going on.

That's what this thing is at the bottom right of the screen:
Screenshot 2024-09-22 at 9 35 35 AM

app/Community/RouteServiceProvider.php
```diff
@@ -39,6 +41,15 @@ protected function mapWebRoutes(): void
     {
         Route::middleware(['web', 'csp'])
             ->group(function () {
+                /*
+                 * client-side api calls
+                 */
+                Route::middleware(['auth'])->group(function () {
+                    Route::get('user-game-list', [UserGameListApiController::class, 'index'])->name('api.user-game-list.index');
+                    Route::post('user-game-list/{game}', [UserGameListApiController::class, 'store'])->name('api.user-game-list.store');
+                    Route::delete('user-game-list/{game}', [UserGameListApiController::class, 'destroy'])->name('api.user-game-list.destroy');
```
wescopeland, line 50: I put all these client-side calls under "api.*" for the name. Figured it probably makes sense to differentiate them from route names that the user can navigate to, etc. Maybe we should adopt this elsewhere in a subsequent change?

app/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -55,6 +55,7 @@ public function share(Request $request): array
             'ziggy' => fn () => [
                 ...(new Ziggy())->toArray(),
                 'location' => $request->url(),
+                'query' => $request->query(),
```
wescopeland, line 58: This lets us access the current query params on the server-side render in `usePageProps()`. If we don't do this, we only have access to them client-side using something like `window.location`, which could lead to hydration issues.

resources/js/common/utils/l10n/formatDate.ts
```diff
@@ -0,0 +1,47 @@
...
+  | 'LTS' // "8:02:18 PM"
+  | 'L' // "08/16/2018"
+  | 'LL' // "August 16, 2018"
+  | 'LLL' // "August 16, 2018 8:02 PM"
+  | 'LLLL' // "Thursday, August 16, 2018 8:02 PM"
+  | 'l' // "8/16/2018"
+  | 'll' // "Aug 16, 2018"
+  | 'lll' // "Aug 16, 2018 8:02 PM"
+  | 'llll' // "Thu, Aug 16, 2018 8:02 PM"
+  | 'MMM DD, YYYY, HH:mm'; // "Aug 16, 2018, 08:02"
```
wescopeland, line 30: `formatDate()` should hopefully allow us to start automatically formatting dates based on whether the user is in en-US, en-GB, pt-BR, etc.

Currently, the app defaults to en-US and there is no way for a user to change their locale. We should add that to the settings page at some point.

Related: https://github.com/RetroAchievements/RAWeb/discussions/2123

resources/js/common/utils/l10n/formatNumber.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Number.prototype.toLocaleString() can be quite slow if called frequently.
+ * User Intl.NumberFormat with a given locale is generally much faster.
+ *
+ * TODO pass in the user's locale from the `auth` object in page props
+ *
+ * @returns A localized number. eg: 12345 -> "12,345"
+ */
```
wescopeland, line 8: Similarly, this adds number separators to numbers in a localized fashion. Currently, this defaults to en-US, but should ultimately be driven by a user setting.

resources/js/features/game-list/components/WantToPlayGamesDataTable/WantToPlayGamesDataTable.tsx
```diff
@@ -0,0 +1,162 @@
...
+        table={table}
+        unfilteredTotal={gameListQuery.data?.unfilteredTotal ?? null}
+      />
+
+      <BaseTable
+        containerClassName={cn(
+          'overflow-auto rounded-md border border-neutral-700/80 bg-embed',
+          'light:border-neutral-300 lg:overflow-visible lg:rounded-sm',
+
+          // A sticky header cannot support this many columns. We have to drop stickiness.
```
wescopeland, line 99: It's worth nothing that sticky headers have to be disabled due to CSS restrictions once the user enables a certain column count on either LG or XL.

resources/js/features/game-list/components/WantToPlayGamesDataTable/buildColumnDefinitions.tsx
wescopeland, line 1: All column definitions live in this file.

resources/js/features/game-list/components/WantToPlayGamesMainRoot/WantToPlayGamesRoot.test.tsx
wescopeland, line 1: It isn't worth testing each table component individually without a whole table. This file serves as an integration test suite for the entire table.

resources/js/test/setup.tsx
```diff
@@ -60,3 +60,37 @@ export function render<TPageProps = Record<string, unknown>>(
...
+
+// -----------------
+
+type DefaultRenderHookParams = Parameters<typeof defaultRenderHook>;
+type RenderCallback = DefaultRenderHookParams[0];
+type RenderHookOptions<TPageProps = Record<string, unknown>> = DefaultRenderHookParams[1] & {
+  pageProps?: Partial<TPageProps & AppGlobalProps>;
+};
+
+export function renderHook<TPageProps = Record<string, unknown>>(
```
wescopeland, line 72: I really don't like ever using `renderHook()` to test React hooks without an accompanying component that they influence (ie: test a component that just happens to use the hook for behavior rather than explicitly testing the hook itself).

But some of the hooks with the table are complex enough that I felt it was warranted.

tests/Feature/Platform/Action/BuildGameListActionTest.php
```diff
@@ -307,58 +330,114 @@ public function testItCanSortByLastUpdated(): void
     }
 
     public function testItCanSortByReleasedAt(): void
+    {
+        $this->markTestSkipped('Has problems specifically with SQLite.');
```
wescopeland, line 334: For whatever reason, while this test runs fine locally, it seems to fail in CI. Something must be different about SQLite in GitHub Actions.

app/Community/RouteServiceProvider.php
wescopeland: I'm also wondering if it makes sense for us to start putting these under something like `/request`, eg `/request/user-game-list`.

That may better delineate the URLs from pages that can be navigated to.

tests/Feature/Platform/Action/BuildGameListActionTest.php
wescopeland: An alternative to skipping this altogether might be to detect if GitHub Actions is being used and skip it in that circumstance.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableFacetedFilter.tsx
```diff
@@ -0,0 +1,186 @@
...
+    column?.setFilterValue(filterValues.length ? filterValues : undefined);
+  };
+
+  return (
+    <BasePopover>
+      <BasePopoverTrigger asChild>
+        <BaseButton
+          size="sm"
+          className={cn(
+            'border-dashed',
```
wescopeland, line 66: Let me know if this `border-dashed` appears too distracting in the UI. I'm not 100% sure about changing it because the button label mutates as the user does things (the border is supposed to hint that the button can grow or shrink).

Table borders should be a bit more subdued than the demo. If they're still too loud, I can try to tweak them some more.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableRowActions.tsx
```diff
@@ -0,0 +1,85 @@
...
+      loading: 'Removing...',
+      success: () => {
+        // Trigger a refetch of the current table page data and bust the entire cache.
+        queryClient.invalidateQueries({ queryKey: ['data'] });
+
+        return `Restored ${gameTitle}!`;
+      },
+      error: 'Something went wrong.',
+    });
+  };
```
Jamiras, line 46: How can something be restored? The button immediately disappears on click.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableRowActions.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <BaseTooltip>
+      <BaseTooltipTrigger asChild>
+        <BaseButton
+          variant="ghost"
+          className="flex h-8 w-8 p-0 text-link"
+          onClick={handleRemoveFromBacklogClick}
+        >
+          <RxCross2 className="h-4 w-4" />
+          <span className="sr-only">Remove from backlog</span>
```
Jamiras, line 76: This should probably be right-aligned. As you remove columns, it creeps to the left:
image

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, [table.getState().columnFilters]);
+
+  /**
+   * Wait until the user is done typing before we fetch from the back-end.
+   */
+  useDebounce(
+    () => {
+      if (rawInputValue.length >= 3 || rawInputValue.length === 0) {
+        table.getColumn('title')?.setFilterValue(rawInputValue);
+      }
```
Jamiras, line 39: this behavior is a little weird. I typed in a three character string and got zero matches, then started backspacing to get some close matches, and the list didn't update until the search string was gone. then I started typing in the shorter matches without getting any matches, and at the third character, the list evaporated again.

I know it's not performant to do filters on small substrings - especially as you're typing, but the same could be said at 4 or five characters. If I'm typing in "dragon", it should either start filtering immediately at "D", or wait until I stop pushing keys for a second or so.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+    },
+    200,
+    [rawInputValue],
```
Jamiras, line 42: Should explicit wildcards be allowed?
image

I guess the big search bar does it too.

app/Platform/Actions/BuildGameListAction.php
```diff
@@ -57,35 +75,55 @@ public function execute(
...
+
+                return new GameListEntryData(
+                    game: GameData::from($game)->include(
+                        'system.nameShort',
+                        'system.iconUrl',
+                        'achievementsPublished',
+                        'badgeUrl',
+                        'pointsTotal',
+                        'pointsWeighted',
+                        'releasedAt',
```
Jamiras, line 105: sorting by releaseAt puts "unknown" at the end of the list for both ascending and descending. Not sure if it would be desirable to have them first ascending (as if year was 0)

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableRowActions.tsx
wescopeland: You should see an Undo button on the toast that appears:

https://github.com/user-attachments/assets/883a052d-5f0e-41f8-8d32-91b5d33f4d1e

Do you see something different?

resources/js/common/components/PlayerBadgeIndicator/PlayerBadgeIndicator.tsx
```diff
@@ -0,0 +1,30 @@
...
+      className={cn(
+        'h-2 w-2 rounded-full',
+
+        awardType === AwardType.Mastery && awardDataExtra ? 'bg-[gold] light:bg-yellow-600' : '', // Mastered
+        awardType === AwardType.Mastery && !awardDataExtra ? 'border border-yellow-600' : '', // Completed
+
+        awardType === AwardType.GameBeaten && awardDataExtra ? 'bg-zinc-300' : '', // Beaten
+        awardType === AwardType.GameBeaten && !awardDataExtra ? 'border border-zinc-400' : '', // Beaten (softcore)
+
+        className,
```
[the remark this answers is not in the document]
wescopeland: Yep.

There are a few problems with the current indicators that I'm trying to solve with this new design. The current indicators:
* Are quite loud when "stacked" in multiple successive rows.
* Can visually conceal progress near the end of the bar.
* Have labels which are impossible to see without a mouse to hover and read the `title` attribute.

The new indicators solve all of these problems. On hover, tap, or focus, the label appears. They're also slightly more visually subdued when "stacked":

Screenshot 2024-09-23 at 4 11 09 PM

tests/Feature/Platform/Action/BuildGameListActionTest.php
wescopeland: Note that I'm now having this test skipped explicitly in CI only, as of https://github.com/RetroAchievements/RAWeb/pull/2706/commits/449bfb8d8fdc5ffb92b2f44e77b2cd20c619f590.

app/Platform/Actions/BuildGameListAction.php
wescopeland: This differs from what I'm seeing on my local. Initiating any new sort should automatically pop the table back to page 1 without needing a page refresh:

Screenshot 2024-09-23 at 6 49 49 PM

With that in mind (and assuming you see the same behavior), how do you feel about "unknown" entries ordering at last ascending (as if year was 9999)?

I'm a bit torn on what the right course of action is. I've scoured the web trying to dig up as much UX research on this subject as I can, but [the consensus is mixed](https://ux.stackexchange.com/questions/70265/when-sorting-by-date-do-dateless-entries-come-first-or-last) and seems to always lean towards "it just depends on the context".

With that in mind, the only two contexts I can think of are [Chronological Guide to RetroAchievements](https://retroachievements.org/viewtopic.php?t=25851&o=0) and huge game databases like [IGDB](https://www.igdb.com/platforms/saturn/games). I would assume fans of the chronological guide are going to reach for sorting by earliest first, and seeing a bunch of unknown entries there would probably be perceived as clutter. IGDB also places unknown dates at the end of their ascending list.

I will defer to what you think we should do, but felt like this was a case probably worth making before we ultimately decide.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+            'text-neutral-400 peer-focus:opacity-0 light:bg-gray-200 light:text-gray-800',
+            'lg:block',
+          )}
+        >
+          /
+        </kbd>
+
+        <div aria-live="polite" className="sr-only">
+          Press / to focus the search field.
+        </div>
```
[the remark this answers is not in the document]
wescopeland: Changed to a tooltip in latest:
Screenshot 2024-09-23 at 7 13 57 PM

This hotkey does tend to be a common pattern among web search experiences, eg GitHub (theirs is much fancier and much more difficult to implement):
Screenshot 2024-09-23 at 7 11 18 PM

Apparently the JS code for our own docs, as well as Laravel's docs, automatically handles the `/` hotkey too. I just tested it out of curiosity and was surprised to find it worked.

app/Platform/Actions/BuildGameListAction.php
```diff
@@ -159,8 +212,8 @@ private function applyFilters(Builder $query, array $filters, ?User $user = null
                 /*
                  * only show games based on whether they have achievements published
                  */
-                case 'hasAchievementsPublished':
-                    $this->applyHasAchievementsPublishedFilter($query, $filterValues);
+                case 'achievementsPublished':
+                    $this->applyAchievementsPublishedFilter($query, $filterValues);
```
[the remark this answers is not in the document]
wescopeland: This took some doing, but "Has achievements" now defaults to "Yes".

Note that the Reset button in this screenshot is no longer visible:

Screenshot 2024-09-23 at 7 19 36 PM

Erring on the side of caution, I added several more test cases to account for the presence of default filters.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableRowActions.tsx
wescopeland: Good catch. Right-aligned in latest.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
wescopeland: Agreed, that can feel a bit strange.

I've tried to tweak this implementation in latest to (hopefully) improve the UX. Now, searches work regardless of input length. However, the shorter the input length, the longer the debounce.

```ts
useDebounce(
  () => {
    table.getColumn('title')?.setFilterValue(rawInputValue);
  },
  getDebounceDuration(rawInputValue),
  [rawInputValue],
);

function getDebounceDuration(input: string): number {
  // Don't debounce at all when the input is cleared.
  if (input.length === 0) {
    return 0;
  }

  // Use a longer debounce for shorter inputs.
  if (input.length < 3) {
    return 500;
  }

  // Use a shorter debounce for longer inputs.
  return 200;
}
```

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
wescopeland: No strong feeling from me either way on this. I haven't changed anything in latest, but happy to proceed with attacking this if you'd like to see it changed.

app/Community/RouteServiceProvider.php
wescopeland: Tried `api`, but it collides with our existing web API `/API/` prefix, resulting in numerous errors.
Went with `internal-api` in latest.

resources/js/common/components/PlayerBadgeIndicator/PlayerBadgeIndicator.tsx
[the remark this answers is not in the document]
Jamiras: What about something more similar to the Completion Progress page?
image

or the Completion Progress on the profile?
image

Both have text indicating how full the bar is.

The way you have it now, it's just too similar to the bar itself. It feels like there's a baby second progress bar and I just can reason why it would be there.

Maybe it would be less off-putting if the Mastered text was always visible instead of just appearing on mouseover.
image

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableRowActions.tsx
Jamiras: I wasn't looking for a button on the popup. I suppose it's better than nothing.

Is five seconds enough time for the user to register their mistake and get their cursor to that button before the toast goes away?

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
Jamiras: I'd probably go 1000 and 500. When typing in "dragon", i saw filters occurring for "dra", "drag", "drago", and "dragon", and I consider myself an above-average typist.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
Jamiras: It mimics the main search bar. As long as we can't do SQL injection, I'm fine with it

app/Platform/Actions/BuildGameListAction.php
Jamiras: >  Initiating any new sort should automatically pop the table back to page 1 without needing a page refresh:

That was working fine. I was just testing the various sorts with one page of results, and noticed that sorting by release date always left the null values at the bottom.

Games without achievements appear first when sorting ascending by achievements, points, and rarity, so I just assumed null dates would also sort first.

I hadn't considered the player experience for this. As a player, I'd most likely look for _either_ the oldest game on the list, or the newest, and wouldn't care at all about the ones without a date. As such, maybe it's better the way you had it.

app/Community/RouteServiceProvider.php
wescopeland: Apologies for the misunderstanding.

I think we should indeed keep "api.*" in the name. However, `prefix` is the route prefix, eg:

`/internal-api/foo`

When a group is given a prefix, all routes are given that prefix in the URL. Hopefully this makes sense.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableRowActions.tsx
wescopeland: In my opinion, it probably is. I can extend it by another second or two if you don't feel it's sufficient. I'm worried about a lot of long-running toasts stacking up if the user is removing a bunch of games.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
wescopeland: Bumped to 1000 and 500 in https://github.com/RetroAchievements/RAWeb/pull/2706/commits/363ff75d6913a037182fd3c5f432b0e8c504ef01.

resources/js/features/game-list/components/WantToPlayGamesDataTable/DataTableSearchInput/DataTableSearchInput.tsx
wescopeland: Should be safe from SQL prevention, as everything is passing through Eloquent ORM and `Request`.

app/Platform/Actions/BuildGameListAction.php
wescopeland: I see. To be clear, I'm only seeing null dates appear at the bottom of the list if the sort is by earliest (ascending).

If the sort is by latest, they appear at the top:
Screenshot 2024-09-25 at 5 25 37 PM

However, based on the findings reported in my previous comment, this is a behavior I'm in favor of.

resources/js/common/components/PlayerBadgeIndicator/PlayerBadgeIndicator.tsx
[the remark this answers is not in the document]
wescopeland: I'm doing my best to mitigate a disorienting sensation I experience when I see this:

Screenshot 2024-09-25 at 5 45 31 PM

My eyes feel like they're uncomfortably glued to these labels, drawing my attention away from nearly everything else on the page. The other pages' bars either do something similar when there's a lot of them, or would cause other undesirable effects on the rows of the grid, such as height or padding inflation.

The rows on this table are slim enough that I even feel this sensation when just percentages are shown under a lot of these bars within close proximity.

Here's where I've landed in latest which should hopefully be a healthy compromise:


https://github.com/user-attachments/assets/884c12f5-f616-45c2-9967-a5b3ee8d4986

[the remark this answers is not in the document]
Jamiras: What do you think about left-aligning the second line?
image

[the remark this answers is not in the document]
wescopeland: I like that idea. Updated in latest.

Jamiras (approved):

PR #2708, opened 2024-09-22
feat(manage): add CRUD operations for emulator systems and releases

wescopeland:
This PR:
* Adds the ability to add and remove systems from the emulator resource page.
* Adds the ability to add, edit, and delete releases from the emulator resource page.

Screenshot 2024-09-22 at 3 54 36 PM
Screenshot 2024-09-22 at 3 54 45 PM
Screenshot 2024-09-22 at 3 55 03 PM
Screenshot 2024-09-22 at 3 55 08 PM
Screenshot 2024-09-22 at 3 55 15 PM


app/Filament/Resources/EmulatorResource/RelationManagers/SystemsRelationManager.php
```diff
@@ -55,10 +55,11 @@ public function table(Table $table): Table
 
             ])
             ->headerActions([
-
+                Tables\Actions\AttachAction::make()
+                    ->preloadRecordSelect(),
```
Jamiras, line 59: Should this generate an audit log entry?

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
```diff
@@ -57,10 +64,11 @@ public function table(Table $table): Table
 
             ])
             ->headerActions([
-
+                Tables\Actions\CreateAction::make(),
```
Jamiras, line 67: Should this create an audit log entry?

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
```diff
@@ -57,10 +64,11 @@ public function table(Table $table): Table
...
             ])
             ->actions([
-
+                Tables\Actions\EditAction::make(),
```
Jamiras, line 70: There should only be one version set as Minimum at any give time. I'm not sure if that would be easy to do automatically.

Since there isn't a view for releases (just an edit page), should edits be shown in the emulator log?

Jamiras (commented):
Thanks for doing this. It looks simple enough, but I had no idea where to begin.

app/Filament/Resources/EmulatorResource/RelationManagers/SystemsRelationManager.php
wescopeland: It should indeed, updated in latest.

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
wescopeland: Updated in latest to begin recording these logs, however there isn't a dedicated page for them at the moment (see other comment below).

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
wescopeland: Fixed in latest so only one value can be set to minimum at a time. If any row is set to minimum, any row which has a truthy minimum will flip to false.

> Since there isn't a view for releases (just an edit page), should edits be shown in the emulator log?

I don't think so. I believe this breaks several assumptions made by the audit log page itself.

I think what the best course of action is would be to create an audit log page for emulator releases resources which is accessible via the actions column in the rows, kind of like some of the other resources such as achievements:

Screenshot 2024-09-23 at 9 01 29 PM

This is not currently present in latest though, as I haven't explored what would be involved in building a "detached" audit log page such as this.

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
Jamiras, line 67: Edge case: Deleted a row and tried to recreate it and got a duplicate key error.

app/Filament/Resources/EmulatorResource/RelationManagers/SystemsRelationManager.php
Jamiras, line 1: Unrelated to this PR, but if you have a quick solution, it would be nice to include it here.

Attempting to search the systems on the emulator page generates an error:

```
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'ID' in where clause is ambiguous
select
  count(*) as aggregate
from
  `Console`
  inner join `system_emulators` on `Console`.`ID` = `system_emulators`.`system_ID`
where
  `system_emulators`.`emulator_id` = 7
  and (
    `ID` like % nintendo %
    or `name_full` like % nintendo %
  )
  and `Console`.`deleted_at` is null
  ```

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
wescopeland: This is a tough one - it's happening due to the records being soft deleted.

The only resolution I know if is to allow privileged users to restore these records. Added in latest:

Screenshot 2024-09-25 at 10 00 14 PM

Jamiras: It's not something I would expect to occur in normal use. If it's easier to ignore the error, I'm fine with that.

app/Filament/Resources/EmulatorResource/RelationManagers/SystemsRelationManager.php
wescopeland: No problem. Fixed in latest.

Jamiras (approved):

PR #2709, opened 2024-09-22
feat(manage): improve management of untitled leaderboards

wescopeland:
This PR improves the management/visibility of untitled leaderboards within the management app.

At the moment, it is very ambiguous what row you're working with if a leaderboard does not have a title.

**Follow up question**
Should titles for newly-created leaderboards be required by _dorequest.php_?

**Before**
Screenshot 2024-09-22 at 4 12 37 PM

**After**
After


Jamiras:
Apparently, I brought it up for discussion over a year ago (and it does look like crap in the emulators): https://discord.com/channels/476211979464343552/1002688256309211257/1133041963009048627

Looks like the conversation died pretty quickly. Might be time to revisit it.

wescopeland:
> Looks like the conversation died pretty quickly. Might be time to revisit it.

Yes. This is bad.

I am only now fully processing the totality of the problem.

As a web "feature", it makes assumptions about the site UI that we can't guarantee. In emulators, it looks terrible and users may assume it's a bug. It sounds like we're already on the same page.

I'm going to say we should just call "title-less leaderboards" an anti-pattern and put an end to this practice by adding the validation in _dorequest.php_.

We'll still need the changes in this PR though, as there's something like 1000+ of these leaderboards floating around and they'll need to be managed.

Jamiras (approved):

PR #2710, opened 2024-09-22
feat(manage): add in-line display order editing for leaderboards and achievements

wescopeland:
Users now have the option of using drag and drop reordering or inline editing.

Screenshot 2024-09-22 at 4 49 21 PM

Screenshot 2024-09-22 at 4 49 28 PM

### DIFF
--- a/app/Filament/Resources/AchievementResource.php
+++ b/app/Filament/Resources/AchievementResource.php
@@ -310,10 +310,10 @@ class AchievementResource extends Resource
                     ->searchable()
                     ->toggleable(isToggledHiddenByDefault: true),
 
-                Tables\Columns\TextColumn::make('DisplayOrder')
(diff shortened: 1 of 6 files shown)


Jamiras (approved):
This seems functional, but I find it hard to use. As soon as I change one value, the list automatically rearranges and I have to click into another field to move the next item. 

When multi-reordering, it's even worse. Attempting to drag-select the text to replace it actually starts to drag the row.

Previously, when multi-reordering achievements, I would tab through the fields: Highlight first value, type in new value, tab, tab, type in next value, tab, tab, type in next value, etc. Save.

wescopeland:
After reordering site awards is migrated to React, I plan on revisiting this with an alternative approach that should hopefully be more user-friendly.

PR #2713, opened 2024-09-25
feat(multiset): introduce the set management tool for developers

wescopeland:
This PR implements a tool which enables developers to associate subsets to a game. With this tool, developers can go ahead and hook up game subset associations. There is no player-facing UX involved here, and there is still a lot left to do (see "Next steps" in the #2662 description).

### Video

https://github.com/user-attachments/assets/4fecc842-1ff5-4e64-87eb-bb60a8f605e6

## UX Walkthrough

On game pages in the management app, there is a new "Sets" tab on the bottom panel:
Screenshot 2024-09-24 at 8 34 10 PM

When the "Attach Subset" button is pressed, a wizard dialog appears:
Screenshot 2024-09-24 at 8 35 12 PM

The "Subset Game" dropdown is filtered by games of the currently-selected game's system. In this case, I'm viewing Dragon Quest III for GBC, so I'm only going to see GBC subset "games" in the list:
Screenshot 2024-09-24 at 8 36 24 PM

I want the "Gold Medals" subset to be attached to the game. I select it and press the "Next" button to see the second screen:
Screenshot 2024-09-24 at 8 37 17 PM

I can give the subset a title (which is conveniently pre-filled for me), and I can select the subset's type. After clicking "Submit", the subset is associated with the game:
Screenshot 2024-09-24 at 8 37 54 PM

Audit log entries are created on both "attach" and "detach" actions.

Display orders can be changed for any type of set other than "Core".

Core sets cannot be detached.


app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
wescopeland, line 1: Open question: should there be any validation against linking the same subset to more than one game?

drisc: I'd say yes, there is no guarentee that a subset will work for more than one game.

Jamiras: The only place this I'd even consider linking a subset to multiple games is for a hack. It's possible that a hack could share the same set of collectables, or glitch showcase, or whatever with it's base game. But there's probably logistical reasons that it shouldn't be allowed even if they were compatible (maybe it's easier to get some rare collectable in the hack).

app/Models/Game.php
```diff
@@ -120,6 +124,46 @@ protected static function newFactory(): GameFactory
...
+                    ->withProperty('attributes', [$relationName => (new Collection($pivotIds))
+                        ->map(fn ($pivotId) => [
+                            'id' => $pivotId,
+                            'attributes' => $pivotIdsAttributes[$pivotId],
+                        ]),
+                    ])
+                    ->event('pivotAttached')
+                    ->log('pivotAttached');
+            }
+        });
```
Jamiras, line 147: Are the audit logs supposed to be raw?
image

app/Platform/Enums/AchievementSetType.php
```diff
@@ -47,4 +47,17 @@ enum AchievementSetType: string
...
+    public function label(): string
+    {
+        return match ($this) {
+            self::Core => 'Core',
+            self::Bonus => 'Bonus',
+            self::Specialty => 'Specialty',
+            self::Exclusive => 'Exclusive',
+            self::WillBeBonus => 'Bonus*',
+            self::WillBeSpecialty => 'Specialty*',
+            self::WillBeExclusive => 'Exclusive*',
```
Jamiras, line 60: It makes sense now that I see this, but I was confused what the asterisk meant. Would it be worth adding a tooltip, or since this is a "temporary" thing, just disseminate the knowledge as we go?

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
```diff
@@ -0,0 +1,201 @@
...
+                                ->send();
+
+                            return;
+                        }
+
+                        $game->achievementSets()->attach($legacySubsetAchievementSet->id, [
+                            'type' => $data['type'],
+                            'order_column' => 1,
+                            'title' => $data['title'],
+                        ]);
```
Jamiras, line 176: These rows are being created without created_at or updated_at timestamps.

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
RALordAndrew: I know of one instance where a multi-game subset could be useful. There is a Mario Bros. minigame included in five different GBA games, and currently only one of the sets covers this content. Of course, such a subset would have to be made compatible with all supported hashes of all linked games.

wescopeland: Another case I can think of is: https://retroachievements.org/game/12399 which is actually already linked to two game hashes right now (TLoZ: Oracle of Ages and TLoZ: Oracle of Seasons).

I think it may indeed be desirable for subsets to be linkable to multiple games.

I think we should at the very least show a warning about it in the UI though.

app/Models/Game.php
wescopeland: Yes. This caught me by surprise too. At first, these elements crashed the page altogether because they don't have a `->name` field. After I fixed that, they fall into the next branching case, which is to just dump the JSON:
```php
{{-- audit-log.blade.php --}}

@elseif (is_array($newValue))
    <pre class="text-xs text-gray-500">{{ json_encode($newValue, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES) }}</pre>
```
I started working to enhance this to look a bit better, but I quickly realized that the complexity of doing so would be kinda high. May be worth revisiting later.

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
wescopeland: Filament doesn't leave a lot of wiggle room for rendering a warning, but here's where I've ended up in latest:

Screenshot 2024-09-26 at 6 00 54 PM

I only reached for an emoji in this instance because Tailwind classnames don't seem to work properly for that warning, so I can't change the text color based on the user's light/dark mode preference for the management app.

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
wescopeland: Good catch, fixed in latest.

app/Platform/Enums/AchievementSetType.php
wescopeland: No harm in adding a tooltip for now. Added in latest:

Screenshot 2024-09-26 at 6 09 37 PM

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
Jamiras: Minor: selecting a subset already attached to the current game gives the same warning (attached to another game...). Could the message be updated, or an error be generated at this point? 

An error does get reported when submitting the duplicated subset.

Jamiras (approved):

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
wescopeland: It might be better from a UX standpoint to not show the subset in the dropdown if it's already attached to the record. I've made this change in latest: https://github.com/RetroAchievements/RAWeb/pull/2713/commits/7f8663d544f832a3e9bac21a58a647bc4a2350e7

PR #2719, opened 2024-09-27
feat(multiset): send core achievements to emulators from game_achievement_sets

wescopeland:
This PR makes a change to the Connect API's `patch` routine so core achievements are now being sent by doing a lookup on `game_achievement_sets` for the game.

Additionally, an `AchievementSets` field is added to the response, but not yet populated. Ideally, this would contain achievements and set metadata for all the subsets the user has opted in to for the game (either on a global or local basis).

I wanted to get this PR open first to ensure I'm on the right track and we're happy with this direction.


app/Helpers/database/game.php
```diff
@@ -993,17 +994,10 @@ function GetPatchData(int $gameID, ?User $user, int $flag): array
         'ConsoleID' => $game->ConsoleID,
         'ImageIconURL' => media_asset($game->ImageIcon),
         'Achievements' => [],
+        'AchievementSets' => [],
```
Jamiras, line 997: I'd rather leave this out until we actually populate it. The name might change depending on how the data is returned. For backwards compatibility we want to keep the core set in the Achievements list, and the additional sets would be in here. This name has some implications of holding both.

Additionally, there could be leaderboards for the subset, so it won't just be achievements in here.

Jamiras (commented):
Seems to be returning the same data for the couple of sets I checked.

app/Helpers/database/game.php
wescopeland: That makes sense. Removed `AchievementSets` from the response in latest.

Jamiras (approved):

wescopeland:
Before merging this, I want to be absolutely 100% sure of data integrity in prod.

I've been doing some local testing and I've found some minor integrity issues that I'll be opening a PR to resolve later today.

PR #2721, opened 2024-09-27
use news link in RSS post when it is internal link

Krylan:
This is a small change, which changes links in RSS News feed.
For now, every news points at the homepage (where the news would be available in the slider). 
After my changes, if news link is internal link (for example pointing to a forum thread of ongoing event), then RSS feed will put it into the post link.


wescopeland:
Hi @Krylan!

It looks like a check is failing. You may need to run `composer fix`.

Krylan:
@wescopeland, fixed!

By the way, this time, like many times before, using linter modified much more files, than only those I edited – in this case it was 90 more files, mostly adding a comma at the end of arguments list of function.

public/rss-news.xml.php
```diff
@@ -34,6 +34,9 @@
     $newsDate = date("D, d M Y H:i:s O", strtotime($news['Timestamp']));
     $newsImage = $news['Image'];
     $newsLink = config('app.url');
+    if (str_starts_with($news['Link'], config('app.url'))) {
+        $newsLink = $news['Link'];
+    }
```
wescopeland, line 39: In _.env_, when I set `APP_URL` to "https://retroachievements.org" and then visit http://localhost:64000/rss-news, for some reason an exception is thrown:

> DOMDocument::createElement(): unterminated entity reference c=228801#228801

```php
    $article->appendChild($dom->createElement('link', $newsLink));
```

Krylan: I didn't have this issue, but I guess that it could come from some special characters in links.
I think that remediation for this issue would be treating them the same way as other text content appended in XML (which was lacking before).
Please, update the branch and check again, if issue still persists.

wescopeland: That fixed the error 👍

wescopeland (approved):
During my local testing, I found an unrelated encoding issue on the page that may also be worth fixing up in this branch:

```xml
<description>
<![CDATA[ The Unwanted&#039;s 50th Month is underway (Aug 15 – Sep 15) Enjoy some underplayed games or help contribute to the website to earn tokens towards an event badge! Click here to see this month&#039;s list of games. ]]>
</description>
```

To repro, hardcode `$payload` before `$newsPayload` gets initialized on line 47:
```php
$payload = "The Unwanted's 50th"; // add this
$newsPayload = "<![CDATA[" . htmlspecialchars(strip_tags($payload)) . "]]>";
```

Krylan:
Now I removed excessive `htmlspecialchars()` calling both for payload and the title. Both gets `htmlentity()` treating on later stage, so this one shouldn't be needed back there.
For me it now looks good (my RSS reader was converting those codes to original characters anyway, but now it looks better directly in XML source).

PR #2722, opened 2024-09-27
update game metrics when player is tracked/untracked

Jamiras:
Causes jobs to be dispatched to recalculate the player count and latest masteries for every game a player being tracked/untracked has ever played.

https://discord.com/channels/310192285306454017/1289178214492536945/1289178214492536945


app/Platform/Actions/UpdatePlayerPlayedGames.php
```diff
@@ -0,0 +1,50 @@
...
+use App\Models\PlayerGame;
+use App\Models\User;
+use App\Platform\Jobs\UpdateGameMetricsJob;
+use App\Platform\Services\GameTopAchieversService;
+use Illuminate\Bus\Batch;
+use Illuminate\Bus\BatchRepository;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Bus;
+
+class UpdatePlayerPlayedGames
```
Jamiras, line 16: This could use a better name. It's purpose is to update data depending on the player's visibility for all games they've ever played.

app/Platform/Actions/UpdatePlayerPlayedGames.php
```diff
@@ -0,0 +1,50 @@
...
+                    ->onQueue('game-metrics')
+                    ->name('player-played-games ' . $user->id . ' ' . $page)
+                    ->allowFailures()
+                    ->finally(function (Batch $batch) {
+                        // mark batch as finished even if jobs failed
+                        if (!$batch->finished()) {
+                            resolve(BatchRepository::class)->markAsFinished($batch->id);
+                        }
+                    })
+                    ->dispatch();
```
Jamiras, line 42: Most of the code in this PR is copied/modified from other classes. Let me know if I something got copied and is inappropriate.

This comes from UpdateGamePlayerGames.

app/Platform/Jobs/UpdatePlayerPlayedGamesJob.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $user = User::find($this->userId);
+
+        if (!$user) {
+            // might've been deleted
+            return;
+        }
+
+        app()->make(UpdatePlayerPlayedGames::class)
+            ->execute(User::findOrFail($this->userId));
```
wescopeland, line 59: nit: I can't help but notice we're fetching the user twice here. Not sure why I didn't notice before. Seems like this is in happening both in this class and in _UpdatePlayerMetricsJob.php_.

app/Platform/Jobs/UpdateGameMetricsJob.php
```diff
@@ -13,6 +14,7 @@
 
 class UpdateGameMetricsJob implements ShouldQueue, ShouldBeUniqueUntilProcessing
 {
+    use Batchable;
```
wescopeland, line 17: nit: I think we may want to also include:
```php
if ($this->batch()?->cancelled()) {
    return;
}
```
in `handle()`. Without it, I believe batch cancellation is ignored.

app/Platform/Actions/UpdatePlayerPlayedGames.php
wescopeland: My first idea was `UpdateGamesAffectedByPlayerVisibilityChange`, but I'm afraid this could be too prescriptive.

`UpdateGameMetricsTouchedByUser`?

wescopeland (approved):

app/Platform/Actions/UpdatePlayerPlayedGames.php
Jamiras: I went with `UpdateGameMetricsForGamesPlayedByUser`

app/Platform/Jobs/UpdatePlayerPlayedGamesJob.php
Jamiras: Copied from `UpdatePlayerMetricsJob`. Updated both.

app/Platform/Jobs/UpdateGameMetricsJob.php
Jamiras: done

wescopeland (approved):

PR #2649, opened 2024-08-26
feat: set up basic Inertia l10n/i18n, use lang strings on `/forums/recent-posts`

wescopeland:
This PR adds basic internationalization support to the Inertia.js+React code, all while having the capability to leverage Laravel's localization (l10n) and internationalization (i18n) JSON files in the _lang_ directory.

This is by no means a comprehensive i18n/l10n solution. The objective is to play nicely with Laravel's built-in localization system and have the ability to move forward without hardcoding strings in the React UI.

## How to test this PR

Open _lang/en_US.json_. Modify some of the values, for example:
```
    "Recent Posts": "Recent Posts123"
```
Navigate to `/forums/recent-posts`, or if you're already on it, refresh the page (Vite will not hot reload this change). You should see the text on the page has changed.

You can also try changing your locale. If you open the already-existing _pt_BR.json_ and add the following content:
```
    ":count posts in the last 24 hours": ":count posts nas últimas 24 horas",
    ":count posts in the last 7 days": ":count posts nos últimos 7 dias",
    "Additional Posts": "Posts adicionais",
    "Forum Index": "Índice do Fórum",
    "in": "em",
    "Last Post By": "Última postagem por",
    "Message": "Mensagem",
    "Next :count": "Próximo :count",
    "Previous :count": "Anterior :count",
    "Recent Posts": "Posts Recentes"
```
You can then open _resources/js/common/components/AppProviders/AppProviders.tsx_ and change the `locale` prop on line 11 to "pt_BR". You will then see:
Screenshot 2024-08-26 at 6 21 12 PM

## How it works
The entire app is wrapped in a `LaravelReactI18nProvider` component like:
```tsx
<LaravelReactI18nProvider>
  <div>{/* THE WHOLE APP IS HERE */}</div>
</LaravelReactI18nProvider>
```

This component injects strings from the user's current locale into the component tree and works with server-side rendering. These strings can be placed in the DOM using a single React hook: `useLaravelReactI18n()`.

```tsx
const { t, tChoice } = useLaravelReactI18n();

t("key"); // outputs the value of "key" from en_US.json

t('Welcome, :name!', { name: 'Francisco' }); // "Welcome, Francisco!"
t(':count apples', { count: 2 }); // "2 apples"

tChoice('There is one apple|There are many apples', 1); // "There is one apple"
tChoice('There is one apple|There are many apples', 4); // "There are many apples"

tChoice('{0} There are none|[1,19] There are some|[20,*] There are many', 0); // "There are none"
[...]

PR #2723, opened 2024-09-27
feat: set up Inertia i18n, use lang strings, attempt #2

wescopeland:
This PR is a repeat of #2649, which was approved, but reverted in #2673 after it was found to being causing hydration issues related to server-side rendering.

I really like how easy the `laravel-react-i18n` library is, and I think our i18n/l10n solution needs to be so easy that we barely ever have to think about it.

The bug responsible for our hydration issue was reported in Dec 2023 here: https://github.com/EugeneMeles/laravel-react-i18n/issues/16. The bug is directly related to how the current major release of React (shipped in 2022) changed React's internal hydration algorithm to be more efficient.

I found the root cause of the hydration bug in the library itself and fixed it locally.

Upstream doesn't seem to be accepting PRs at the moment to fix the issue.

My original plan was to maintain a fork of the lib and add it to our _package.json_ file. However, that seems to cause other spurious errors in the RAWeb project due to peer dependency conflicts.

This PR implements my next-best idea, which is to maintain an internal fork of the lib with the patched changes. This is stored under _@/lib/laravel-react-i18n_. Most of this PR's diff is just pulling in the source code for the lib. Here's the summary of changes in this PR:

* Pull in a modified version of `laravel-react-i18n` into _@/lib/laravel-react-i18n_.
* Restore the changes from #2649.
* Pass down the user's locale setting in the server-side and client-side renders in _app.tsx_ and _ssr.tsx_.

Implementation details under _@/lib/laravel-react-i18n_ are safe to ignore.

---

**How to test this PR:**
* `sail npm install`
* `sail npm run build`
* `sail artisan inertia:start-ssr`
* Navigate to `/forums/recent-posts`.
* Observe that there is no hydration error or devtools console errors. Also navigate to `/game/1/hashes` and observe the same.
* In the database on your UserAccounts row, set your "locale" value to "pt_BR". Check `/forums/recent-posts` again.

en_US:
Screenshot 2024-09-27 at 6 29 10 PM

pt_BR:
Screenshot 2024-09-27 at 6 28 45 PM

---

**Next steps:**
* Add a way for users to change their `locale` value via the settings page.
* Use the `t()` function on all React pages.
* After #2706 is merged, make sure the locale value is also flowing into our number and date formatting utils.
* Write some kind of documentation so native speakers can understand how to contribute translations.


lang/pt_BR.json
wescopeland, line 1: None of the deleted strings in this file were used, so I went ahead and wiped them out.

resources/js/app.tsx
wescopeland, line 1: `LaravelReactI18nProvider` causes hydration issues if it lives inside of `<AppProviders />` due to needing different config based on server/client context. Due to a quirk in Vite, that config must be completely static and cannot be driven even by a variable or a prop.

This will probably be the only exception where we have a provider that lives outside of `<AppProviders />`.

resources/js/lib/laravel-react-i18n/README.md
wescopeland, line 1: Everything under this directory is vendor code (and some of these implementation details are kinda wild). Hopefully, we don't need to do something like this very often.

resources/js/lib/laravel-react-i18n/provider.ts
wescopeland, line 1: For any reader from https://github.com/EugeneMeles/laravel-react-i18n/issues/16, if you want to patch your own project, copy & paste the changes in this file.

wescopeland:
In the event that upstream is open to changes, I've opened a PR for them here https://github.com/EugeneMeles/laravel-react-i18n/pull/22.

If it ever gets merged, we can drop this internally-maintained lib.

resources/js/ziggy.d.ts
```diff
@@ -309,6 +309,27 @@ declare module 'ziggy-js' {
...
+            "binding": "ID"
+        }
+    ],
+    "achievement-set-claim.update": [
+        {
+            "name": "claim",
+            "required": true,
+            "binding": "ID"
+        }
+    ],
```
Jamiras, line 332:
```
$ sail npm run build

> build
> vite build && vite build --ssr

failed to load config from /var/www/html/vite.config.ts
error during build:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'php-array-reader' imported from /var/www/html/vite.config.ts.timestamp-1727622897317-b2c9612ea5222.mjs
    at packageResolve (node:internal/modules/esm/resolve:854:9)
    at moduleResolve (node:internal/modules/esm/resolve:927:18)
    at defaultResolve (node:internal/modules/esm/resolve:1157:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:383:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:352:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:227:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:87:39)
    at link (node:internal/modules/esm/module_job:86:36)
```

resources/js/ziggy.d.ts
```diff
@@ -309,6 +309,27 @@ declare module 'ziggy-js' {
             "required": true
         }
     ],
+    "achievement-set-claim.create": [
```
Jamiras, line 312: was this required in my PR? or is it bleeding in from the management console PR?

resources/js/ziggy.d.ts
wescopeland: If you do `sail npm install`, does that fix `sail npm run build`?

resources/js/ziggy.d.ts
wescopeland: It's just picking up the new routes from me needing to regenerate types with `composer types`. I don't think it's a big deal to re-run `composer types` unless something in them changes. Maybe we should explore automating this somehow so it's always current?

resources/js/ziggy.d.ts
Jamiras: it's automatically run by `.hooks/post-merge`. apparently it was failing with a different error. I've nuked my `node_modules` directory and everything seems happier now.

wescopeland:
Good news: upstream accepted, merged, and released my PR. I've updated this branch accordingly, significantly reducing the diff from +1,229 to +220.

PR #2725, opened 2024-09-28
include post content in forum post/comment email notifications

Jamiras:
Forum post email:
image

Game wall comment email:
image

### DIFF
--- a/app/Helpers/database/forum.php
+++ b/app/Helpers/database/forum.php
@@ -201,16 +201,14 @@ function submitTopicComment(
     }
 
     if ($user->ManuallyVerified ?? false) {
-        notifyUsersAboutForumActivity($topicId, $topicTitle, $user->User, $newComment->id);
+        notifyUsersAboutForumActivity($topicId, $topicTitle, $user, $newComment->id);
     }
 
     return $newComment;
 }
 
-function notifyUsersAboutForumActivity(int $topicID, string $topicTitle, string $author, int $commentID): void
+function notifyUsersAboutForumActivity(int $topicID, string $topicTitle, User $author, int $commentID): void
 {
-    sanitize_sql_inputs($author);
-
     // $author has made a post in the topic $topicID
     // Find all people involved in this forum topic, and if they are not the author and prefer to
     // hear about comments, let them know! Also notify users that have explicitly subscribed to
@@ -238,7 +236,7 @@ function notifyUsersAboutForumActivity(int $topicID, string $topicTitle, string 
 
     $urlTarget = "viewtopic.php?t=$topicID&c=$commentID#$commentID";
     foreach ($subscribers as $sub) {
-        sendActivityEmail($sub['User'], $sub['EmailAddress'], $topicID, $author, ArticleType::Forum, $topicTitle, $urlTarget, payload: $payload);
+        sendActivityEmail($sub['User'], $sub['EmailAddress'], $topicID, $author->User, ArticleType::Forum, $topicTitle, $urlTarget, payload: $payload);
     }
 }
 
@@ -426,7 +424,7 @@ function authorizeAllForumPostsForUser(User $user): bool
             notifyUsersAboutForumActivity(
                 $unauthorizedPost->forumTopic->id,
                 $unauthorizedPost->forumTopic->title,
-                $user->User,
+                $user,
                 $unauthorizedPost->id,
             );
         }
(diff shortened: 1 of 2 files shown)


app/Helpers/database/forum.php
```diff
@@ -230,9 +230,15 @@ function notifyUsersAboutForumActivity(int $topicID, string $topicTitle, string
         "
     );
 
+    $payload = null;
+    $comment = ForumTopicComment::find($commentID);
+    if ($comment) {
+        $payload = nl2br(Shortcode::stripAndClamp($comment->Payload, previewLength: 1000, preserveWhitespace: true));
```
Jamiras, line 236: `Shortcode::render` mostly works here. It does the formatting and creates clickable links, but there's a lot of extra decorations for the tooltips that can't function in the email, so I opted to unformat everything.

wescopeland (approved):
I approve of the implementation, but I'd be lying if I said I didn't have any reservations about us doing this.

It makes sense to show previews for private messages and user wall comments because users have direct control over who can send them those things. 
Forum topic comments also make sense because we have an extra layer of verification required before a user's post is public (and we don't send emails for unverified posts).

Where I have great concern are game and achievement walls. On what seems like a weekly basis, people post wild, crazy, toxic, and deranged things to these, oftentimes under alt accounts with fake verified email addresses. I have seen death threats, porn ads, and all kinds of other stuff that is wildly inappropriate.

The emails from these right now probably just land in peoples inboxes as noise, but I have great concern about someone weaponizing our email capabilities to mass email people with something like this:

Screenshot 2024-09-28 at 7 05 33 AM

This concern is mostly rooted from two things:
* We are too lax in who we allow to write comments on game and achievement pages.
* We don't have a dedicated unsubscribe button in our email bodies.

I have a few options I'd like to propose to tackle bullet point one:
* If a user has no attached role and zero achievement unlocks, don't allow them to post comments on game and achievement walls. This is already [something I've proposed](https://discord.com/channels/476211979464343552/1287828019104321684/1287915413446983802), but I haven't implemented it yet. It should be straightforward enough though. This proposal was in response to some clown who keeps creating alt accounts to post porn ads all over the site.
* If a user's account is less than N days old (or some other heuristic?), don't show the preview for emails they are responsible for triggering.

Jamiras:
Forum posts have to be manually verified before the emails are sent, so I don't think there's anything to be concerned about there.

Regarding posting spam on game walls. Is this really any worse? The user wouldn't know about the offensive content in the email, but if they're getting the email, they'll probably will want to see what was posted. This just eliminates the click.

If I remove the preview for accounts less than 24 hours old, or with less than 5 minutes of game play time, does it really solve the issue?

wescopeland:
> Forum posts have to be manually verified before the emails are sent, so I don't think there's anything to be concerned about there.

Agreed. My only concern is with game & achievement walls.

> Regarding posting spam on game walls. Is this really any worse? The user wouldn't know about the offensive content in the email, but if they're getting the email, they'll probably will want to see what was posted. This just eliminates the click.
If I remove the preview for accounts less than 24 hours old, or with less than 5 minutes of game play time, does it really solve the issue?

It sounds like the best mitigation may be the 1st bullet point I proposed:

> If a user has no attached role and zero achievement unlocks, don't allow them to post comments on game and achievement walls. This is already [something I've proposed](https://discord.com/channels/476211979464343552/1287828019104321684/1287915413446983802), but I haven't implemented it yet. It should be straightforward enough though. This proposal was in response to some clown who keeps creating alt accounts to post porn ads all over the site.

Some of my concern is also about what email providers (Gmail, etc) may see. As an example, let's circle back to the person posting ads under brand new alt accounts. A provider receiving hundreds of emails containing that ad content from us in the body may just determine that all our emails are actually spam/ads, which could adversely affect other functionality like email verification, etc.

Jamiras:
I see how mail platforms could flag spam coming from our users as coming from us. I've added a the 24-hours/5 minutes of playtime restriction (which matches our requirement to create tickets).

app/Helpers/util/mail.php
```diff
@@ -271,7 +271,13 @@ function informAllSubscribersAboutActivity(
 
         $comment = Comment::find($commentID);
         if ($comment) {
-            $payload = nl2br($comment->Payload);
+            // this is similar to the logic for being able to create a ticket. the user account
+            // must be at least 24 hours old and have at least five minutes on playtime. if not,
+            // w
```
wescopeland, line 276: nit: Looks like maybe a small typo here.

wescopeland (approved):

app/Helpers/util/mail.php
Jamiras: oops. incomplete thought.

PR #2726, opened 2024-09-28
feat(manage): add Achievement Set Claims list

wescopeland:
This PR adds an `AchievementSetClaim` resource to Filament with a basic `table()` view. Creating / editing / auditing claims is not yet supported.

I did my best to recreate http://localhost:64000/claims, including the available filter and sort options.

Active and In Review statuses are selected by default. I'm not 100% sure if this is a reasonable default - happy to change it.

Screenshot 2024-09-28 at 12 51 34 PM

Screenshot 2024-09-28 at 12 55 06 PM


app/Filament/Resources/AchievementSetClaimResource.php
```diff
@@ -0,0 +1,200 @@
...
+                    ->label('Set Type')
+                    ->formatStateUsing(fn ($record) => ClaimSetType::toString($record->SetType)),
+
+                Tables\Columns\TextColumn::make('Special')
+                    ->formatStateUsing(fn ($record) => ClaimSpecial::toString($record->Special)),
+
+                Tables\Columns\TextColumn::make('Created')
+                    ->label('Claimed At')
+                    ->dateTime('d M Y')
+                    ->sortable(),
```
Jamiras, line 95: The default sort on the old page was "Newest Claim first". I don't think the new page has any default sort.

wescopeland: Updated in latest.

Jamiras (approved):
Should the old page be removed now?

wescopeland:
Probably so.

Here's a very dumb question: how does someone navigate to the existing page? https://retroachievements.org/claims

I only know how to get there manually via the URL. Want to make sure I'm not missing a redirect if one needs to be added.

Jamiras:
Per #2481:
> Additionally, as each of the above pages presents its own view and filters, a master view is now available at /claims, with access to all of the data. There are no direct links to that page.

wescopeland:
According to Plausible, `/claims` has had about 30 unique visitors in the last month.

In latest, I've removed the existing page and added a redirect to `/manage/achievement-set-claims` in the event that any of those visitors happen to be developers.

PR #2727, opened 2024-09-30
feat: migrate All Games to React datatable

wescopeland:
This PR migrates the "All Games" page to the new datatable introduced in https://github.com/RetroAchievements/RAWeb/pull/2706.

Most of this PR's diff is code coverage and refactoring. Getting the table to work in this context actually took very little work.

The page is accessible at `/games`.

Things to be aware of:
* This table lets users both remove & add games to their backlog.
* The table excludes hub, event, subset, and inactive system games. I personally feel this is better for the UX, as it removes a lot of "surprises" most users would not expect to see.
* All breadcrumbs have been adjusted, including ones in the React app. This is important: **this is first time we're doing client-side routing via the Inertia.js `<Link />` component.** To see this in action, navigate to `/game/1/hashes` and then click the "All Games" breadcrumb link.

Screenshot 2024-09-29 at 9 17 23 PM


resources/js/features/achievements/components/AchievementBreadcrumbs/AchievementBreadcrumbs.tsx
```diff
@@ -34,7 +35,9 @@ export const AchievementBreadcrumbs: FC<AchievementBreadcrumbsProps> = ({
       <BaseBreadcrumb>
         <BaseBreadcrumbList>
           <BaseBreadcrumbItem>
-            <BaseBreadcrumbLink href="/gameList.php">All Games</BaseBreadcrumbLink>
+            <BaseBreadcrumbLink asChild>
+              <Link href={route('game.index')}>All Games</Link>
+            </BaseBreadcrumbLink>
```
wescopeland, line 40: `<Link />` does client-side routing to the next page, and works as long as that page is also an Inertia.js page. This results in faster page transitions.

resources/js/features/game-list/components/AllGamesDataTable/AllGamesDataTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+  setColumnVisibility,
+  setPagination,
+  setSorting,
+  sorting,
+}) => {
+  const { can } = usePageProps<App.Community.Data.UserGameListPageProps>();
+
+  const gameListQuery = useGameListQuery({ columnFilters, pagination, sorting });
+
+  const table = useReactTable({
```
wescopeland, line 45: This hook should likely be made reusable, but I want to have one or two more datatables so I can be more sure of what the interface for the reusable hook should be.

resources/js/features/game-list/components/AllGamesDataTable/buildColumnDefinitions.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const columnDefinitions: ColumnDef<App.Platform.Data.GameListEntry>[] = [
+    buildTitleColumnDef({ forUsername: options.forUsername }),
+    buildSystemColumnDef({}),
+    buildAchievementsPublishedColumnDef({}),
+    buildPointsTotalColumnDef({}),
+    buildRetroRatioColumnDef({}),
+    buildLastUpdatedColumnDef({}),
+    buildReleasedAtColumnDef({}),
+    buildPlayersTotalColumnDef({}),
+    buildNumVisibleLeaderboardsColumnDef({}),
```
wescopeland, line 29: Column definitions are now reusable functions.

It's plausible some datatables may have columns that aren't used by other tables. By keeping these as individual functions, Vite can perform dead code elimination for the user's page, meaning they only receive the JS that's actually needed for where they're at.

resources/js/features/game-list/components/DataTableRowActions/DataTableRowActions.tsx
```diff
@@ -0,0 +1,80 @@
...
+    // This should never happen.
+    if (!gameId) {
+      throw new Error('No game ID.');
+    }
+
+    if (!auth?.user && typeof window !== 'undefined') {
+      window.location.href = route('login');
+
+      return;
+    }
```
wescopeland, line 43: Important implementation detail: if users aren't authenticated, they'll still see the add button. Clicking it will redirect them to login.

I'm not sure if we have any means of setting a callback URL so they return to this page after logging in, but I'm not sure that's too important. New users will need to register, and existing users probably already know where they are.

resources/js/features/game-list/components/AllGamesMainRoot/AllGamesMainRoot.test.tsx
wescopeland, line 1: Almost all of this coverage is duplicated from the Want to Play Games table. One or two of these test cases is unique. The tests run really fast, so I have no hesitation with all the duplication here. There's also no telling how the code might evolve or be refactored later, so this seems like a safe choice.

resources/js/features/game-list/utils/column-definitions/buildPlayerGameProgressColumnDef.tsx
```diff
@@ -0,0 +1,33 @@
...
+        table={table}
+        sortType="quantity"
+        tableApiRouteName={tableApiRouteName}
+      />
+    ),
+    cell: ({ row }) => {
+      const { game, playerGame } = row.original;
+
+      return <PlayerGameProgressBar game={game} playerGame={playerGame} />;
+    },
```
Jamiras, line 31: Noticed on this grid, but it happens on the Want to Play grid as well: The progress widget changes the cursor to a hand, making it seem like it should be clickable, but clicking does nothing.

app/Platform/Actions/BuildGameListAction.php
```diff
@@ -156,6 +157,16 @@ private function buildBaseQuery(GameListType $listType, ?User $user = null): Bui
...
 
         switch ($listType) {
+            case GameListType::AllGames:
+                // Exclude events, hubs, subsets, and inactive systems.
+                $query->where('GameData.ConsoleID', '!=', System::Events)
+                    ->where('GameData.ConsoleID', '!=', System::Hubs)
+                    ->where('GameData.Title', 'not like', "%[Subset -%")
+                    ->whereHas('system', function ($q) {
+                        return $q->active();
+                    });
```
Jamiras, line 167: Recommend replacing the ConsoleID checks with `System::scopeGameSystems`.

(untested):
```
->whereHas('system', function ($q) {
    return $q->gameSystems()->active();
});
```

resources/js/features/game-list/components/DataTableRowActions/DataTableRowActions.tsx
```diff
@@ -0,0 +1,83 @@
...
+            variant="ghost"
+            className="group flex h-8 w-8 p-0 text-link"
+            onClick={handleToggleFromBacklogClick}
+            disabled={isPending}
+          >
+            <MdClose
+              className={cn(
+                'h-4 w-4 transition',
+                'hover:text-neutral-50 disabled:text-neutral-50 light:hover:text-neutral-900 light:disabled:text-neutral-900',
+                isInBacklog ? 'text-red-600 light:text-red-500' : 'rotate-45',
```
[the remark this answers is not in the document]
wescopeland: Sure, applied both suggestions in latest:

Screenshot 2024-10-01 at 5 10 50 PM

resources/js/features/game-list/utils/column-definitions/buildPlayerGameProgressColumnDef.tsx
wescopeland: Agreed. In latest, if the tooltip can appear, the progress bar now links to the game page.

app/Platform/Actions/BuildGameListAction.php
wescopeland: That seems to work great. Applied in latest.

resources/js/features/game-list/components/DataTableRowActions/DataTableRowActions.tsx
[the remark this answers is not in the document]
Jamiras: Actually playing with the minuses, I think I like the yellow X (like current production) better.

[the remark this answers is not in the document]
wescopeland: Updated in latest. I've also made some slight tweaks to the animation and how the client makes the API call. The UI update is now optimistic, so the perceived performance is better.


https://github.com/user-attachments/assets/0797fcb2-a7da-4910-ae14-764590128073

Jamiras (approved):
Still have a couple questions about the sort title generation code, but I'll bring that up elsewhere.

PR #2728, opened 2024-09-30
fix: move achievement migration tool to Filament, update achievement_sets

wescopeland:
This PR moves the admin achievement migration tool to the "Admin Tools" page in Filament. Additionally, migrating an achievement now also updates `achievement_sets`.

Screenshot 2024-09-30 at 4 39 54 PM


resources/views/livewire/administrative-tools/migrate-achievements.blade.php
```diff
@@ -0,0 +1,116 @@
...
+                        return Achievement::with('game')
+                            ->where('Title', 'like', "%{$search}%")
+                            ->orWhere('ID', 'like', "%{$search}%")
+                            ->limit(50)
+                            ->get()
+                            ->mapWithKeys(function ($achievement) {
+                                return [$achievement->id => "ID: {$achievement->id} - Title: {$achievement->title} - Game: {$achievement->game->title}"];
+                            })
+                            ->toArray();
+                    })
```
Jamiras, line 65: I don't know that this feature gets used all that often, but I imagine that when it is used, people just paste in a CSV of achievements to move. CSVs don't seem to work with the new component. You have to enter each ID individually, and select the matching result.

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
```diff
@@ -0,0 +1,59 @@
...
+
+        // Associate the achievements to the new game.
+        Achievement::whereIn('ID', $achievementIds)->update(['GameID' => $gameId]);
+
+        // Add an audit comment to the new game.
+        addArticleComment(
+            'Server',
+            ArticleType::GameModification,
+            $gameId,
+            "$user migrated " . Str::plural('achievement', count($achievementIds)) . ' ' .
```
Jamiras, line 32: Should be `{$user->display_name}`:
```
{"ID":...} migrated achievements 458393,458394,458392 from games 29978,4068.
```

Jamiras: But wouldn't it be better to put the GameID modifications in the audit logs for the individual achievements?

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
```diff
@@ -0,0 +1,59 @@
...
+                $playerGame->save();
+                dispatch(new UpdatePlayerGameMetricsJob($userId, $gameId));
+            }
+        }
+
+        // Update the metrics on the new game and the old game(s).
+        dispatch(new UpdateGameMetricsJob($gameId))->onQueue('game-metrics');
+        foreach ($oldGameIds as $oldGameId) {
+            dispatch(new UpdateGameMetricsJob($oldGameId))->onQueue('game-metrics');
+        }
```
Jamiras, line 50: There's an `AchievementMoved` event that notifies the `DispatchUpdateGameMetricsJob` that seems like it should be used here. The trade-off is that it has to be raised for each achievement, and relies on the queue to de-dup the old/new game updates.

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
```diff
@@ -0,0 +1,59 @@
...
+                implode(',', $achievementIds) . ' from ' .
+                Str::plural('game', count($oldGameIds)) . ' ' . $oldGameIds->implode(',') . '.'
+        );
+
+        // Ensure player_game entries exist for the new game for all affected users.
+        foreach (PlayerAchievement::whereIn('achievement_id', $achievementIds)->select(['user_id'])->distinct()->pluck('user_id') as $userId) {
+            if (!PlayerGame::where('game_id', $gameId)->where('user_id', $userId)->exists()) {
+                $playerGame = new PlayerGame(['user_id' => $userId, 'game_id' => $gameId]);
+                $playerGame->save();
+                dispatch(new UpdatePlayerGameMetricsJob($userId, $gameId));
```
Jamiras, line 42: The `PlayerGameMetrics` should be updated for the user whether or not they've previously played the game. The player metrics for the old game(s) should also be updated.

resources/views/livewire/administrative-tools/migrate-achievements.blade.php
wescopeland: Updated in latest to use a CSV input field.

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
wescopeland: Applied both in latest. GameID changes are now correctly recorded in the audit log.

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
wescopeland: I didn't know about this event. I believe with my fix for the audit log, `AchievementMoved` is now being correctly automatically dispatched.

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
wescopeland: Updated in latest:
```php
// Update all affected player game metrics.
$affectedUserIds = PlayerAchievement::whereIn('achievement_id', $achievementIds)
    ->select(['user_id'])
    ->distinct()
    ->pluck('user_id');
foreach ($affectedUserIds as $userId) {
    // Ensure player_game entries exist for the new game for all affected users.
    if (!PlayerGame::where('game_id', $gameId)->where('user_id', $userId)->exists()) {
        $playerGame = new PlayerGame(['user_id' => $userId, 'game_id' => $gameId]);
        $playerGame->save();
    }

    dispatch(new UpdatePlayerGameMetricsJob($userId, $gameId));
    foreach ($oldGameIds as $oldGameId) {
        dispatch(new UpdatePlayerGameMetricsJob($userId, $oldGameId));
    }
}
```

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
```diff
@@ -0,0 +1,63 @@
...
+
+            dispatch(new UpdatePlayerGameMetricsJob($userId, $gameId));
+            foreach ($oldGameIds as $oldGameId) {
+                dispatch(new UpdatePlayerGameMetricsJob($userId, $oldGameId));
+            }
+        }
+
+        // Update achievement_sets associated with the given achievement IDs.
+        $upsertAction = new UpsertGameCoreAchievementSetFromLegacyFlags();
+        $upsertAction->execute(Game::find($gameId));
```
Jamiras, line 58: I assume it's fine, but I noticed that moving achievements to a set with no achievements generates an audit log entry for the new game, which is displayed awkwardly:
image

Jamiras (approved):

app/Platform/Actions/MigrateAchievementIdsToDifferentGameId.php
wescopeland: I didn't see that earlier -- good spot. 

It's probably safe to ignore for now. This appears to be a side effect of how we're handling double writes. We'll likely reach a state in the future where this upsert action is the default behavior, and `Flags` is dropped from achievements altogether (tracked instead via achievement sets).

@luchaos put this TODO in the `Achievement` model:
```
// TODO drop Flags, derived from being included in an achievement set
```
It's worth noting that `achievement_set_achievements` has no notion of whether the achievement is published or not. We'll likely need to store that info in `achievement_set_achievements` rather than `Achievements`, as an achievement should probably never be orphaned from a set.

PR #2736, opened 2024-10-02
fix: password reset functionality

wescopeland:
https://discord.com/channels/476211979464343552/1002689485005406249/1291000365705199689

### DIFF
--- a/public/request/auth/send-password-reset-email.php
+++ b/public/request/auth/send-password-reset-email.php
@@ -1,5 +1,6 @@
 <?php
 
+use App\Models\User;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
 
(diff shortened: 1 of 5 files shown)


public/request/auth/send-password-reset-email.php
```diff
@@ -1,12 +1,17 @@
...
 use Illuminate\Support\Facades\Validator;
 
 $input = Validator::validate(Arr::wrap(request()->post()), [
     'username' => 'required',
 ]);
 
-RequestPasswordReset($input['username']);
+$targetUser = User::firstWhere('User', $input['username']);
+
+if (!$targetUser->isBanned()) {
```
Jamiras, line 13: Need a null check here. If you try to reset the password for a non-existant username, you get an error.

wescopeland: Good point - added in latest.

Jamiras (approved):

PR #2739, opened 2024-10-05
fix(game): force sort tagged games at the end of lists

wescopeland:
Fixes an issue where, when `sort_title` ordering is applied (ie: in latest `master`), tagged games such as "~Homebrew~ My Homebrew Game" are sorted at the top of the Title (Asc) list rather than the bottom.

The test which checks for this is passing due to collation differences between SQLite and MariaDB.

To fix the issue, tagged games now have their sort titles prefixed with "zzzz{", eg:
`'~Hack~ Dragoon X Omega' -> 'zzzz{hack dragoon x omega'`


Jamiras (approved):

wescopeland:
Going to go ahead and merge this. I will address the follow-up feedback we discussed elsewhere in a new PR.

PR #2740, opened 2024-10-06
feat: add 'Claimed' column to React datatable

wescopeland:
Adds a new (hidden by default) "Claimed" column:

Screenshot 2024-10-06 at 1 50 39 PM

I didn't want to conflate this with the Progress column because I believe the UI/UX complexity of doing so is quite high. Revisions still count as claims, which could bloat the Progress column's design. Sorting UX also becomes a challenge in this scenario.

Additionally, to keep things simple with UI/UX and back-end sorting, **this is a boolean column**. The actual users behind the claim can be seen in many places elsewhere. On the table, I am just assuming that users care if the game has any active claims at all. This is what the column looks like:

Screenshot 2024-10-06 at 1 52 59 PM

The column is sortable:
Screenshot 2024-10-06 at 1 54 59 PM

Currently, the column does not toggle by default when changing the value of the "Has achievements" filter. In a follow-up PR, I am planning on building another feature that lets users persist their column preferences. Auto-toggling columns while supporting such a feature may be cumbersome, and from the user perspective may not make much sense.


resources/js/features/game-list/utils/column-definitions/buildHasActiveOrInReviewClaimsColumnDef.tsx
```diff
@@ -0,0 +1,55 @@
...
+  tableApiRouteName?: RouteName;
+}
+
+export function buildHasActiveOrInReviewClaimsColumnDef({
+  tableApiRouteName = 'api.game.index',
+}: BuildHasActiveOrInReviewClaimsColumnDefProps): ColumnDef<App.Platform.Data.GameListEntry> {
+  return {
+    id: 'hasActiveOrInReviewClaims',
+    accessorKey: 'game',
+    meta: { label: 'Claimed', align: 'right' },
```
wescopeland, line 22: `align: 'right'` may not be right. Maybe it'd be best if this is left-aligned?

Jamiras: I agree. left probably makes more sense here

wescopeland: 👍  Left-aligned in latest.

Jamiras (approved):

wescopeland:
Note that in latest, I have also added iconography for the Claimed sort order in the mobile view:

Screenshot 2024-10-27 at 7 06 14 PM

Screenshot 2024-10-27 at 7 06 23 PM

PR #2741, opened 2024-10-06
fix(user): allow writing on own wall when only followers setting is enabled

wescopeland:
This PR fixes a minor bug on the user profile.

Currently, if the "Only people I follow can message me or post on my wall" setting is enabled, users cannot post on their own walls because they are not following themselves. This PR slightly adjusts the policy to fix this issue.

### DIFF
--- a/app/Policies/UserCommentPolicy.php
+++ b/app/Policies/UserCommentPolicy.php
@@ -59,7 +59,11 @@ class UserCommentPolicy
             || !$user->hasVerifiedEmail()
             || $commentable->isBlocking($user)
             || !$commentable->UserWallActive
-            || ($commentable->only_allows_contact_from_followers && !$commentable->isFollowing($user))
+            || (
+                $commentable->only_allows_contact_from_followers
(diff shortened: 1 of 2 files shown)


app/Policies/UserCommentPolicy.php
```diff
@@ -59,7 +59,11 @@ public function create(User $user, ?User $commentable): bool
             || !$user->hasVerifiedEmail()
             || $commentable->isBlocking($user)
             || !$commentable->UserWallActive
-            || ($commentable->only_allows_contact_from_followers && !$commentable->isFollowing($user))
+            || (
+                $commentable->only_allows_contact_from_followers
+                && !$commentable->isFollowing($user)
+                && !$user->is($commentable)
```
Jamiras, line 65: Consider moving the `user->is` check up to avoid the query embedded in the `isFollowing` call

wescopeland: Good point - updated in latest.

Jamiras (approved):

PR #2745, opened 2024-10-07
feat(multiset): add Sets resource and hash compatibility management tool

wescopeland:
This PR adds a Sets resource and multiset hash compatibility management tool to the management app.

With this advanced tool, developers can explicitly set some game hashes to not be compatible with some sets.

Note: The badge count for the "Incompatible Hashes" tab does not automatically update when adding or removing a hash. This is a [Filament limitation](https://github.com/filamentphp/filament/discussions/7221#discussioncomment-6551399).

Screenshot 2024-10-07 at 4 59 21 PM

Screenshot 2024-10-07 at 4 59 59 PM

Screenshot 2024-10-07 at 5 04 47 PM

Screenshot 2024-10-07 at 5 04 54 PM

Screenshot 2024-10-07 at 5 05 01 PM

Screenshot 2024-10-07 at 5 06 04 PM


app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
```diff
@@ -0,0 +1,112 @@
...
+                                    ->mapWithKeys(function (GameHash $hash) {
+                                        return [$hash->id => "{$hash->name} {$hash->md5}"];
+                                    })
+                                    ->toArray()
+                            ),
+                    ])
+                    ->action(function (array $data) use ($achievementSet) {
+                        $gameHashId = (int) $data['hash'];
+
+                        $achievementSet->gameHashes()->attach($gameHashId, ['compatible' => false]);
```
Jamiras, line 97: Is there any way to report the `compatible` value in the audit message? When making a hash explicitly incompatible, the audit message just says the hash was attached, which implies it should be valid.
image

app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
```diff
@@ -0,0 +1,112 @@
...
+use Filament\Tables;
+use Filament\Tables\Table;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\HtmlString;
+
+class GameHashesRelationManager extends RelationManager
+{
+    protected static string $relationship = 'gameHashes';
+
+    protected static ?string $title = 'Incompatible Hashes';
```
Jamiras, line 24: As best as I can tell, this implementation assumes all entries in the `gameHashes` list are marked as incompatible. Should the property be renamed to reflect that? `incompatibleGameHashes`?

Are there plans to use the `compatible` field in the table for anything else? Maybe it can be removed and the table renamed to `achievement_set_incompatible_game_hashes`?

app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
wescopeland: With the migration added in latest which repurposes the table, this should now hopefully be more clear:

Screenshot 2024-10-12 at 11 01 04 AM

app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
wescopeland: To the best of my knowledge, there are no plans to use the `compatible` field for anything else. When the table was created, I was a bit less sure, but now that the system design for multiset has started to take a more solid shape, it's easier to see that field is very likely redundant.

I've added a migration to latest which renames the table and drops that field.

app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
```diff
@@ -0,0 +1,112 @@
...
+                                    ->mapWithKeys(function (GameHash $hash) {
+                                        return [$hash->id => "{$hash->name} {$hash->md5}"];
+                                    })
+                                    ->toArray()
+                            ),
+                    ])
+                    ->action(function (array $data) use ($achievementSet) {
+                        $gameHashId = (int) $data['hash'];
+
+                        $achievementSet->incompatibleGameHashes()->attach($gameHashId);
```
Jamiras, line 97: Error if hash is already marked as incompatible:
> SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1882-33972' for key 'set_hash_unique'

wescopeland: Addressed in latest. Now after marking a hash as incompatible, it no longer appears as a select option. Additionally, I've added a few more guardrails in latest:

* The select field has been set to required.
* The action is disabled if there are no hashes to choose from.
* The "Detach" label and associated modal copy has been changed to be more clear.

Jamiras:
I don't think it's caused by this PR, but searching using the topmost search field (instead of the lower one next to the game list) generates an error:
> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'title' in 'where clause'

app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
Jamiras: Cannot mark hash as incompatible if it's selected by default:
image

Jamiras:
Could be part of this PR, or a separate one, but the Sets list on the game page doesn't link to the set page.

app/Filament/Resources/AchievementSetResource/RelationManagers/GameHashesRelationManager.php
wescopeland: Fixed in latest.

wescopeland:
That global search issue was indeed caused by this PR. Fixed in latest. I also added that link on the game page sets table.

Jamiras (approved):

PR #2747, opened 2024-10-10
feat: propagate internationalization and localization to the React app

wescopeland:
This PR is a follow-up to #2723. It propagates our internationalization (i18n) and localization (l10n) systems throughout the React app.

Screenshot 2024-10-09 at 8 38 39 PM

**How to test this PR:**
```
sail npm run build
```
Now navigate to any of the React pages. You shouldn't see any hydration errors.

Without rebuilding or restarting your server, set your `UserAccounts.locale` value to "pt_BR". After refreshing the page, you should see numbers and dates being formatted differently, as well as Portuguese strings appear in the UI.

**Implementation details:**
1. A new linting rule has been added which blocks non-interpolated strings from being used in templates, eg:
```tsx
<p>Hello!</p> FAIL
<p>{"Hello!"}</p> PASS
<p>{t('Hello!')}</p> PASS
```
This should help us make significantly fewer mistakes.

2. en_US has been updated with all the correct key-value pairs from the React UI. pt_BR has also been updated with (mostly) machine translations. Some helpful community assistance was utilized to improve it further. **There is no expectation that the pt_BR file should be actively maintained.** It's mostly just a demo for this branch. Only en_US should be maintained; the community should maintain the other locales.

---

Next steps:
* Create a linting rule for props/args prefixed with "t_", eg: `t_label`. These should also block string literals.
* Create a `<Trans />` component for rendering complex formatted/interpolated strings, ie: https://react.i18next.com/latest/trans-component
* Allow users to change their locale via the Settings page.
* Write documentation on how users can contribute translations.


Jamiras (approved):
Only did minimal testing

PR #2748, opened 2024-10-10
fix(game): more sort title algorithm adjustments

wescopeland:
More `sort_title` algorithm changes, specifically with padded numbers.

**Before**
"007: Agent Under Fire" -> "00007 agent under fire"
"20Q" -> "000020q"
"64th. Street" -> "000064th street"

**After**
"007: Agent Under Fire" -> "0000007 agent under fire"
"20Q" -> "20q"
"64th. Street" -> "64th street"

If I'm missing or overlooking anything else, I'm happy to massage the algorithm even further. Just let me know!


app/Platform/Actions/ComputeGameSortTitleAction.php
```diff
@@ -26,15 +26,45 @@ public function execute(string $gameTitle): string
...
      * set to "mega man 00010" and "mega man 00002" respectively, we can mitigate this.
+     *
+     * Standalone numbers with leading zeroes are padded to 7 digits to preserve significance.
+     * eg: "007" -> "0000007"
+     *
+     * Numbers already padded by replaceRomanNumerals (5 digits) are not given more padding.
+     * eg: "00002" remains "00002"
+     *
+     * Numbers attached to letters are not padded.
+     * e.g: "20Q" -> "20Q", "64th" -> "64th"
```
Jamiras, line 37: Are you sure this is what we want? I'd expect to find "20Q" between "20 Bottles of Beer" and "21 or Over".

Or using actual titles from the DB:
* 7 Wonders of the Ancient World
* 7th Saga, The
* 8 Eyes

app/Platform/Actions/ComputeGameSortTitleAction.php
```diff
@@ -26,15 +26,45 @@ public function execute(string $gameTitle): string
...
+                if (strlen($number) > 1 && strpos($number, '0') === 0) {
+                    // If the number is already 5 digits (from replaceRomanNumerals), don't pad.
+                    if (strlen($number) === 5) {
+                        return $number;
+                    }
+
+                    // Else, pad to 7 digits to preserve leading zeroes.
+                    if (strlen($number) < 7) {
+                        return str_pad($number, 7, '0', STR_PAD_LEFT);
+                    }
```
Jamiras, line 58: As the only significant 0s in the database seem to all belong to `007`, this is probably fine. But for a generic solution, I'd recommend padding the number to 5 digits, then inserting the significant zeroes in front of that.

`007` (pad to 5) => `00007` (insert significant zeroes) => `00` + `00007` => `0000007` (result)

app/Platform/Actions/ComputeGameSortTitleAction.php
Jamiras: There's also an abundance of "3D" games that should be sorted near the "3-D" games, though the latter should probably be normalized to "3D" (not suggesting you do that algorithmically).

wescopeland: I think I may have finally gotten it right in latest:

Screenshot 2024-10-12 at 12 49 27 PM

app/Platform/Actions/ComputeGameSortTitleAction.php
wescopeland: Adjusted in latest - the test case for 007 is still passing. I believe my change should now properly account for some level of dynamism with significant zeroes.

Jamiras (approved):

PR #2749, opened 2024-10-10
fix(DataTableRowActions): disable optimistic ui on user backlog table

wescopeland:
This PR fixes a regression introduced in #2727 on the user Want to Play Games datatable.

If you go to your Want to Play Games table and click the button to remove a game, the game will indeed be removed, but a "+" sign now appears on one of the rows:

Screenshot 2024-10-10 at 5 49 55 PM

**Root Cause**
React is getting confused about which row was removed from the table. To improve this UX, I've made two changes:

* A `key` prop has been given to the `<DataTableRowActions />` cell. This is seldom needed, but helps React recognize what is changing in the DOM.
* The optimistic state and animation on `<DataTableRowActions />` can now be conditionally disabled via a new prop.


Jamiras (approved):
I don't see the behavior on main or this branch, so I'll assume it does what it claims to be doing.

PR #2750, opened 2024-10-10
fix: use game title component in breadcrumbs

wescopeland:
This PR brings over the `<GameTitle />` component introduced in the first datatable PR to ensure it's also being used where appropriate in page breadcrumbs.

**Before**
Screenshot 2024-10-10 at 6 03 21 PM

**After**
Screenshot 2024-10-10 at 6 03 33 PM


Jamiras (approved):
Took me a moment to find the pages this addresses.

I did notice that the same "issue" is present all over the management pages. I presume we're going to just ignore it there.

wescopeland:
Yep, probably just best to ignore it in the management app.

The long-term goal should be to remove those `~Tag~` magic strings and switch to using the `tags` and `taggables` tables powered by `spatie/laravel-tags`.

PR #2751, opened 2024-10-11
fix(multiset): attach the correct id in set management tool

wescopeland:
This PR fixes a bug where the incorrect set can sometimes be attached to a game via the multiset management tool.

To the best of my knowledge, this has occurred twice on two different games:
* Final Fantasy (NES)
* Metroid Prime (GameCube)

In these cases, the developer selected a valid subset, attached it, and the tool actually attached core sets from entirely different games.

**Root Cause**
The tool is (incorrectly) attaching game_achievement_set IDs instead of achievement_set IDs on submit. These just so happen to line up correctly 99% of the time from the original double-write process. When they don't line up correctly (such as from newly-created subsets), the bug surfaces.

If there are any other cases of these floating around, the changes in https://github.com/RetroAchievements/RAWeb/pull/2745 will let us spot them quite easily.


app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
```diff
@@ -194,7 +194,9 @@ public function table(Table $table): Table
                         // It'll use the same achievement_set_id.
 
                         $legacySubsetGame = Game::find((int) $data['game']);
-                        $legacySubsetAchievementSet = $legacySubsetGame->gameAchievementSets()->core()->first();
+                        $legacySubsetAchievementSet = $legacySubsetGame->achievementSets()
+                            ->wherePivot('type', AchievementSetType::Core->value)
+                            ->first();
```
Jamiras, line 199: Could a similar problem be happening in the Search widget? I went to Metroid Prime > Sets > Attach, typed in Metroid, and got this:
image

wescopeland: I think this is a separate ordeal.

For the specific bug addressed by the change in this branch, the user actually selected the correct game in the "Subset game" field, but the tool attached an entirely different core set.

For example, when the user tried to attach a Final Fantasy subset, a Mortal Kombat set was attached by the tool instead.

These options values are coming from:
```php
->searchable()
->options(
    Game::where('Title', 'like', "%[Subset -%")
        ->where('ConsoleID', $game->ConsoleID)
        ->where('achievements_published', '>', 0)
        ->whereDoesntHave('gameAchievementSets', function ($query) use ($attachedAchievementSetIds) {
            $query->core()->whereIn('achievement_set_id', $attachedAchievementSetIds);
        })
        ->orderBy('Title')
        ->get()
        ->mapWithKeys(function ($game) {
            return [$game->id => $game->title];
        })
        ->toArray()
```

I figured fixing this select field bug would be trivial, but it seems I don't have access to the user's input search value in either `->searchable()` or `->options()`.

Jamiras (approved):

PR #2757, opened 2024-10-14
Update pt_BR.json

vagnercruz:
Finalizing language changes!


lang/pt_BR.json
```diff
@@ -13,21 +13,21 @@
...
-    "There are no copyright-protected ROMs available for download on RetroAchievements.org.": "Não há ROMs protegidas por direitos autorais disponíveis para download no RetroAchievements.org.",
+    "There are no copyright-protected ROMs available for download on RetroAchievements.org.": "Não há ROMs protegidas por direitos autorais disponíveis para baixar no RetroAchievements.org.",
     ":earned of :total achievements unlocked": ":earned de :total conquistas desbloqueadas",
     ":earned of :total softcore achievements unlocked": ":earned de :total conquistas softcore desbloqueadas",
-    ":earned of :total points earned": ":earned de :total pontos ganhos",
-    ":earned of :total softcore points earned": ":earned de :total pontos softcore ganhos",
-    "Finished": "Finalizado",
+    ":earned of :total points earned": ":earned de :total de pontos ganhos",
+    ":earned of :total softcore points earned": ":earned de :total de pontos softcore ganhos",
+    "Finished": "Zerado",
```
wescopeland, line 23: This may not be right. This label may appear on "games" that aren't really games, such as events. In other words, it can't really be beaten or mastered. Given that context, should "Zerado" still be used?

lang/pt_BR.json
```diff
@@ -42,7 +42,7 @@
     "Yes": "Sim",
     "No": "Não",
     "Either": "Ambos",
-    "System": "Sistema",
+    "System": "Consoles",
```
wescopeland, line 45: Also wondering if this may be best left as-is. On the back-end, we refer to them as systems because they may not be consoles at all, handheld or otherwise. System is more generic to fit this purpose.

lang/pt_BR.json
```diff
@@ -266,7 +266,7 @@
     "in softcore": "em softcore",
     "Additional information for these hashes may be listed on": "Informações adicionais para esses hashes podem estar listadas em",
     "the game's official forum topic": "o tópico oficial do fórum do jogo",
-    "Details on how the hash is generated for each system can be found": "Detalhes sobre como o hash é gerado para cada sistema podem ser encontrados",
+    "Details on how the hash is generated for each system can be found": "Detalhes sobre como o hash é gerado para cada console podem ser encontrados",
```
wescopeland, line 269: Similar feedback as the other comment about system vs console.

lang/pt_BR.json
vagnercruz: I conducted research with the platform users overall, and everyone requested the change to consoles. I know it might not make total sense, but it's because of how gamers generally refer to video games. The term 'console' becomes a general term we use (in fact, 90% of Brazilians end up including PCs along with consoles).

lang/pt_BR.json
vagnercruz: It's because of the same reason, if we translate the word system it's not wrong, but if we localize it to Brazil it ends up being wrong, system is used for other purposes, here in Brazil console is used by practically 99% of players, it makes more sense one than the other

wescopeland (approved):

PR #2758, opened 2024-10-14
feat(game-list): redesign the ui for mobile devices

wescopeland:
This PR redesigns the React game list datatable UI for mobile devices. The ultimate goals are to:

* Eliminate all horizontal scrolling.
* Simplify searching, filtering, sorting, and pagination.
* Not impact the desktop UX in any way.

## How to test this PR
```bash
sail npm install
sail npm run build
```
Navigate to `/games` and/or `/game-list/play`.

Shrinking your browser window to XS won't be sufficient. The page is explicitly checking for a mobile user agent to cut down on the JS sent to the browser and improve performance.

In your browser devtools, you'll need to explicitly simulate a mobile device. Click the highlighted button, and then ensure you're set to iPhone SE:

Screenshot 2024-10-14 at 3 40 00 PM

Screenshot 2024-10-14 at 3 41 18 PM

Be sure to refresh the page after you're set to iPhone SE.

### Before and After

Screenshot 2024-10-14 at 3 50 21 PM

* There is no table in the DOM. Things now render in a touch-oriented list.
* The mobile UX is modeled closely after music apps such as Apple Music and Spotify. Above the list, we have a count of games, a "super filter button", and a searchbar.
* Game images are bigger, and titles are truncated to a single line of text. The system is always shown under the title. Other information can appear next to the system chip based on the user's current context.
* Each row has two actions: backlog toggle and open more details.

### The Super Filter Button

All filters and sort orders are wrapped into a single button which pops a drawer onto the screen:

https://github.com/user-attachments/assets/459354db-7636-46ec-8eda-62c4742c4e40

Tracking has been added to sort orders on both desktop and mobile. I highly suspect there are a few of these sort orders which never actually get used. After we have some data, we can organize these a little more intelligently in the list and possibly prune sort orders that are bloating the UX.

When the user changes their sort order, their current sort context always appears on the list items:

https://github.com/user-attachments/assets/cf26de17-c83f-49d8-8e4c-0b2ca3439494

### Backlog Toggle

The backlog toggle control works just like it does on desktop. It's worth noting that the word "playlist" is used instead of "Want to Play Games List", due to space constraints. "Playlist" and "devlist" are terms that may be worth considering rebranding both of those features to in the future.

https://github.com/user-attachments/assets/4732ce76-b7b3-4a8b-96ff-e7d24cb4f847

### Game Details Button

Tapping the three vertical dots button on each element opens a drawer from the bottom of the page. This drawer is something of a mobile/touch alternative to the game tooltip cards that users with a mouse currently enjoy:

https://github.com/user-attachments/assets/8d72be97-5b8e-4b23-adc0-1a8cdc3cd5ef

Screenshot 2024-10-14 at 4 24 53 PM

There are a few nuances here that were worth pointing out:
* The progress list item is not displayed if the user is not authenticated.
* While the drawer is open, toggling the item on the backlog will _not_ pop a toast. Popping a toast while a drawer is open causes some bizarre UX consequences. Visual feedback is still provided: the + icon rotates like usual, and the button changes variant.
* There are three links to the game page in the drawer. The "Open Game" button is the obvious one. The other two are subtle: the user can tap the game avatar and the game title. These two subtle links don't really announce themselves with `text-link` and are just there to capture panic taps.
* If the user taps the backlog toggle while not authenticated, they're redirected to login.

### Simplified Pagination

I wanted to eliminate the mobile user from having to think about pagination as much as possible. I tried several different approaches, including infinite scroll, but I scrapped this after I wasn't too happy with how it tested. I landed somewhere in the middle between infinite scrolling and manual pagination:

https://github.com/user-attachments/assets/25911ee2-670e-4309-82da-aa937c1e5efa

There's some trickery happening here. When the user scrolls to within 300px of the "Load more" button, we kick off the fetch for the next page. When/if they tap the button, assuming they have a stable network connection, the results instantly display. This creates an illusion that the back-end is responding instantaneously, even though it most assuredly is not.

If the user has a slow network connection, the UI is smart enough to detect that the fetch is still in flight and shows a temporary loading state until the fetch has completed.

## Other things worth noting
* Mobile's default page size is 100, not 25.
* Landscape (widescreen / side-to-side orientation) mode is basically ignored just due to how little use it gets.
* A lot of test coverage has been added for all this functionality.


wescopeland:
To make testing on a real device easier, I've deployed this branch to preview: https://preview.retroachievements.org/games

Note that this environment is quite a bit slower than prod and has lots of missing assets.

IMG_2161

gboquizosanchez:
If I can express my opinion, mobile is hard to handle but I'd prefer pagination over load more or infinite scroll. The view looks amazing, and I'm fan of Game Details. 

Excellent work!

wescopeland:
Hi @gboquizosanchez !

Thanks for your feedback! I spent a lot of time thinking about the game details, and I'm glad you like this UX.

There are three key problems with pagination on mobile, specifically with this data, that led to the UX I developed in this branch:

* We're currently able to prefetch data with a Load More button, which improves performance on mobile. If we have full pagination controls, we can no longer prefetch the data, as we can't predict anymore which button the user will press.
* I'm afraid pagination on small screens with this kind of list-driven data can get disorienting. On each page change, the user's scroll position is going to dramatically shift, and they'll need to re-orient themselves. This is already happening in the list on desktop, and multiple users have requested a better pagination UX. I think the issue would be more exacerbated here. 
* Pagination on mobile leads to a higher cognitive load in the UX. Ideally, on mobile, I don't want users to think about what page they're viewing.

resources/js/features/game-list/components/GameListItems/GameListItemElement.tsx
```diff
@@ -0,0 +1,176 @@
...
+                'disabled:!text-neutral-100 light:disabled:text-neutral-950',
+                isInBacklogOptimistic ? '' : 'rotate-45',
+              )}
+            />
+
+            <span className="sr-only">
+              {isInBacklogOptimistic
+                ? t('Remove from Want To Play Games')
+                : t('Add to Want to Play Games')}
+            </span>
```
Jamiras, line 131: removing items doesn't dynamically remove the row in mobile mode.
undoing a remove in non-mobile mode no longer restores the row.

Jamiras (commented):
I only did very sparse testing, but it seems to work as designed. I have not looked through the code changes.

My two pieces of feedback are:
* I think progress might be desirable to be shown regardless of the sort. 
* It might be useful to have some way to jump forward through the list (though this seems like an issue on non-mobile too). There's no easy way to get to - say, the Ns - in either UI.

resources/js/features/game-list/components/GameListItems/GameListItemElement.tsx
wescopeland: Good catch, this should now be fixed in latest.

There is one quirk around this that feels worth noting:
* As a user, I open my Want to Play Games List on mobile.
* I click the three dots icon to pull up the game details sheet.
* I click the button to remove the game from my list.
When the user does these steps, the sheet instantly vanishes rather than smoothly closing by scrolling to the bottom of the screen. This is occurring because the sheet is a child element of the row, and by removing the game, the row is no longer mounted in the DOM.

This quirk is probably fine for now. The UX can be further smoothed out later.

wescopeland:
> I think progress might be desirable to be shown regardless of the sort.

Agreed. I've made this change in latest.

> It might be useful to have some way to jump forward through the list (though this seems like an issue on non-mobile too). There's no easy way to get to - say, the Ns - in either UI.

I agree with this as well. I'd like to push tackling this to a subsequent PR which solely focuses on the pagination UX for desktop and mobile. In the short term, I have 3 goals which will immediately follow this PR:

* Increase code coverage and refactor/simplify some of the more complex components.
* Improve the pagination UX.
* Add a "Surprise me" button which takes the user to a random game on the list.

Jamiras (approved):

PR #2764, opened 2024-10-19
feat(game): link Manage button to the edit view if user is allowed

wescopeland:
Makes a tiny update to the Manage link button on game pages.

Now, if a user has permission to update games, they land on the edit page for the game rather than the view page.

Screenshot 2024-10-18 at 10 00 10 PM


Jamiras:
I'm not sure this is an improvement. Sure, it saves a click if you're editing the game title, or genre, or other metadata for the game itself, but that data is fairly static. And once correct, there shouldn't be any reason to make further changes to it. At that point, users going to the management page are more likely to be editing the related records. If they do want to explicitly edit the game metadata, it's just one more click.

Jamiras (approved):

Soulsderfuchs:
When managing multiple game pages at once it gets really cumbersome, I personally wouldn't really be using the management page at all if the date field wasn't moved to there, if the rest of the fields get removed from the game page in the future to fully support the management system then I would use it fully and then this would be even more of a cumbersome process.

wescopeland:
Worth mentioning that related records appear on both view and edit.

**View**
Screenshot 2024-10-24 at 5 19 37 PM

**Edit**
Screenshot 2024-10-24 at 5 19 30 PM

This is just to save 1 click from the user for easier editing.

PR #2767, opened 2024-10-19
feat(settings): allow users to set their locale

wescopeland:
This PR adds a new form to the `/settings` page which allows users to change their current locale option. Accompanies https://github.com/RetroAchievements/RAWeb/pull/2765.

https://github.com/user-attachments/assets/41efe5b6-051d-4258-a2eb-c93582a9f30f

PR #2765, opened 2024-10-19
docs: add TRANSLATIONS.md

wescopeland:
Adds _docs/TRANSLATIONS.md_, which outlines how to contribute translations to the project.

PR #2770, opened 2024-10-20
chore: added translations for es_ES

gboquizosanchez:
To continue with #2767. I have added Spanish (Spain) localization. I have followed docs contained in #2765.


wescopeland:
Hi @gboquizosanchez! Thanks for contributing. Unless anyone voices any opinions about the strings in the file over the next few days, I plan on merging this.

Important to note: es_ES will not initially be available as a selectable locale. Before it can be made available, it also needs to be hooked up into the number and date l10n utils so those values display properly on the initial React.js server render. It's still early days with our i18n/l10n stuff :)

I have a few tasks I'm in the middle of finishing up. Afterwards, I'll be happy to hook es_ES up into our formatter utils. I'll also add a reference to that PR into the translations doc so the process is smoother for new locales going forward.

lang/es_ES.json
```diff
@@ -0,0 +1,313 @@
...
+    ":earned of :total points earned": ":earned de :total de puntos ganados",
+    ":earned of :total softcore points earned": ":earned de :total de puntos softcore ganados",
+    "Finished": "Cerrado",
+    "Mastered": "Masterizado",
+    "Completed": "Completado",
+    "Beaten": "Superado",
+    "Beaten (softcore)": "Superado (softcore)",
+    ":awardLabel on :awardDate": ":awardLabel en :awardDate",
+    "RetroPoints: A measurement of rarity and estimated difficulty.": "RetroPoints: Una medida de rareza y dificultad estimada.",
+    "Derived from points, number of achievers, and number of players.": "Derivado de puntos, número de logros y número de jugadores.",
```
wescopeland, line 30: I am wondering if "logros" is a typo. "achievers" in this context represents the number of players who have unlocked the achievement, not the count of achievements.

gboquizosanchez: Probably is a typo, I'll check it asap. I think "desbloqueadores" or similar.

gboquizosanchez: Humm, the translation of this is "complicated" beacuse "desbloqueadores de logros" is redundant. Achiever is a person that has success, and in this case has unlocked an achievement. But we don't have an appropiate translation for this term...

If I put this into the translator DeepL. It suggests me those translations or alternatives.

image

I suggest "número de usuarios que han desbloqueado logros" but it's excessive long. I'm really confused about the correct term for this.

gboquizosanchez: I have pushed a new translation for this. @wescopeland.

wescopeland (approved):

wescopeland:
Hi @gboquizosanchez! Happy to take what you already have. Not sure if you'd like to go over the latest en_US file before or after merging this one. With the introduction of the <Trans /> component, several keys have changed.

gboquizosanchez:
I think I prefer to change those new keys before you merge this PR. So, I'll start asap. Thanks for the notice!

lang/es_ES.json
```diff
@@ -0,0 +1,313 @@
...
+    "Report Issue": "Reportar problema",
+    "Supported Game Files": "Archivos de juegos compatibles",
+    "Supported Game Files - :gameTitle": "Archivos de juegos compatibles - :gameTitle",
+    "Manage Hashes": "Administrar hashes",
+    "This page shows you what ROM hashes are compatible with this game's achievements.": "Esta página te muestra qué hashes de ROM son compatibles con los logros de este juego.",
+    "Hide": "Ocultar",
+    "Want to Play Games": "Juegos que quiero jugar",
+    "Download Patch File": "Descargar archivo de parche",
+    "Settings": "Configuración",
+    "Are you sure you want to reset your avatar to the default? This cannot be reversed.": "¿Estás seguro de que quieres restablecer tu avatar al predeterminado? Esto no se puede deshacer.",
```
WedgeSparda, line 73: Hi, I was working on another spanish translation before discovering about you work. So I'll drop my branch since you got here first.

Just a point, what do you think about using `¿Realmente quieres` instead of `¿Estás seguro de que quieres` in all confirmation prompts. That's how I was setting those translations in my branch. The reason behind that choose of words was to prompt a message without distinction of genre and also a bit shorter.

What do you think?

gboquizosanchez: Hi @WedgeSparda, I've added the remaining keys after @wescopeland merged the new <Trans> component that includes advanced translation support.

I've incorporated some of the changes you introduced in your PR #2791 because they improved upon my original translation.

Regarding the phrase _Realmente quieres_, I acknowledge it's more concise, but _Estás seguro de que_ might be a better fit. If it appears odd on the website, we can consider altering it. However, I believe this version is more effectively localized.

What are your thoughts? Should we conduct a direct test to see which option performs better on the website?

lang/es_ES.json
```diff
@@ -184,47 +184,47 @@
...
+    "Our web pages and their contents are subject to copyright law. Unless expressly permitted by law, every form of utilizing, reproducing or processing works subject to copyright protection on our web pages requires the prior consent of the respective owner of the rights. Individual reproductions of a work are allowed only for private use. The company names, product names, service names and logotypes on this web site are for identification purposes only. All trademarks and registered trademarks are the property of their respective owners.": "Nuestras páginas web y sus contenidos están sujetos a la ley de derechos de autor. A menos que la ley lo permita expresamente, cualquier tipo de uso, reproducción o procesamiento sujeto a protección por derechos de autor en nuestras páginas web requiere el consentimiento previo del respectivo propietario de los derechos. Las reproducciones individuales de una obra solo están permitidas para uso privado. Los nombres de empresas, nombres de productos, nombres de servicios y logotipos en este sitio web son solo para fines de identificación. Todas las marcas comerciales y marcas registradas son propiedad de sus respectivos propietarios.",
```
WedgeSparda, line 189: English text remains

lang/es_ES.json
WedgeSparda: If you see is a better fit, we can go with `Estás seguro de que` my only concern was the genre, but I think we're all used to that message.

lang/es_ES.json
```diff
@@ -303,11 +286,63 @@
...
+    "Remove from Want to Play Games": "Quitar de juegos que quiero jugar",
     "Ascending (A - Z)": "Ascendente (A - Z)",
     "Descending (Z - A)": "Descendente (Z - A)",
-    "Earliest": "Más antigua",
+    "Earliest": "Más antiguo",
     "Latest": "Más reciente",
     "Less": "Menos",
-    "Both": "Ambos"
+    "Both": "Ambos",
+
```
WedgeSparda, line 296: Jump line

lang/es_ES.json
gboquizosanchez: image

I don't got it.

lang/es_ES.json
wescopeland: FWIW I see this as well. The line count in this file should match en_US in `master`.

Screenshot 2024-10-29 at 12 03 03 PM

gboquizosanchez: I'm checking my last commits because I see it's not sync. :\

gboquizosanchez: image

It's seems a GitHub issue or something... Now it's showing the good one.

lang/es_ES.json
gboquizosanchez: We can make RA more inclusive; however, this requires consideration in all the texts we translate.

What do you think? Should we change?

WedgeSparda: I would say yes. We should review future translations to match the current ones, not just in this specific topic.

gboquizosanchez: Alright, I will make the changes then.

WedgeSparda: Thank you

lang/es_ES.json
```diff
@@ -0,0 +1,347 @@
...
+    "To upload an avatar, earn 250 points or wait until your account is at least 14 days old.": "Para subir un avatar, gana 250 puntos o espera hasta que tu cuenta tenga al menos 14 días.",
+    "Uploading new avatar...": "Subiendo nuevo avatar...",
+    "Uploaded!": "¡Subido!",
+    "Changing your email address will revoke your privileges and you will need to have them restored by staff. Are you sure you want to do this?": "Cambiar tu dirección de correo electrónico revocará tus privilegios y necesitarás que nuestro equipo los restaure. ¿Realmente quieres hacer esto?",
+    "Are you sure you want to change your email address?": "¿Realmente quieres cambiar tu dirección de correo electrónico?",
+    "Changing email address...": "Cambiando dirección de correo electrónico...",
+    "Changed email address!": "¡Dirección de correo electrónico cambiada!",
+    "Changing password...": "Cambiando contraseña...",
+    "Required": "Obligatorio",
+    "Must be at least 8 characters": "Debe tener al menos 8 caracteres",
```
wescopeland, line 99:
```suggestion
    "Must be at least 8 characters.": "Debe tener al menos 8 caracteres",
```

lang/es_ES.json
```diff
@@ -0,0 +1,347 @@
...
+    "Without your explicit consent or a legal basis, your personal data is not passed on to third parties.": "Sin tu consentimiento explícito o una base legal, tus datos personales no se transmiten a terceros.",
+    "Before or at the time of collecting personal information, we will identify the purposes for which information is being collected.": "Antes o en el momento de recopilar información personal, identificaremos los propósitos para los cuales se está recopilando dicha información.",
+    "We will only retain personal information as long as necessary for the fulfillment of those purposes.": " Solo mantendremos información personal mientras sea necesario para el cumplimiento de dichos propósitos.",
```
wescopeland, line 194:
```suggestion
    "We will only retain personal information as long as necessary for the fulfillment of those purposes.": "Solo mantendremos información personal mientras sea necesario para el cumplimiento de dichos propósitos.",
```

lang/es_ES.json
```diff
@@ -0,0 +1,347 @@
...
+    "enter your new email address here...": "introduce aquí tu nueva dirección de correo electrónico...",
+    "Confirm New Email Address": "Confirmar nueva dirección de correo electrónico",
+    "confirm your new email address here...": "confirma aquí tu nueva dirección de correo electrónico...",
+    "Delete Account": "Borrar cuenta",
+    "Keys": "Claves",
+    "Notification type": "Tipo de notificación",
+    "Email notifications": "Notificaciones por correo electrónico",
+    "Site notifications": "Notificaciones en el sitio",
+    "Comments on my activity": "Comentarios en mi actividad",
+    "Comments on an achievement I created": "Comentarios en un logro ",
```
wescopeland, line 227: Possible typo here?

wescopeland (commented):
Just doing a final review! Spotted a few things worth pointing out.

lang/es_ES.json
WedgeSparda, line 227: I think it should be:

```suggestion
    "Comments on an achievement I created": "Comentarios en un logro creado por mí",
```

wescopeland:
I will open a follow-up PR to make es_ES selectable and compatible with the localization utils for numbers and dates.

PR #2771, opened 2024-10-21
chore: add translations for pl_PL

Krylan:
This Pull Request introduces translation file for Polish (pl_PL) language.
It's not fully finished yet, but I'm publishing it, so others could look and suggest some changes or report some mistakes to correct them.
There are still some strings to translate (mostly the big text paragraphs) and probably some of others will change to be more consistent and better sounding.


wescopeland:
Hi @Krylan! Thanks for opening this PR.

Similar to #2770, unless anyone voices any opinions about the strings in the file over the next few days, I plan on merging this.

Important to note similar to es_ES: pl_PL will not initially be available as a selectable locale when this PR gets merged. Before it can be made available, it also needs to be hooked up into the number and date l10n utils so those values display properly on the initial React.js server render. It's still early days with our i18n/l10n stuff :)

I have a few tasks I'm in the middle of finishing up. Afterwards, I'll be happy to hook pl_PL up into our formatter utils. I'll also add a reference to that PR into the translations doc so the process is smoother for new locales going forward.

lang/pl_PL.json
```diff
@@ -0,0 +1,313 @@
...
+    "Yes": "Tak",
+    "No": "Nie",
+    "Either": "Oba",
+    "System": "System",
+    "Has achievements": "Ma osiągnięcia",
+    "Reset": "Zresetuj",
+    ":count of :total game|:count of :total games": ":count z :total gier|:count z :total gier",
+    ":count game|:count games": ":count gra|:count gier",
+    ":count selected": ":count wybranych",
+    "No options found.": "No options found.",
```
wescopeland, line 51: Would "Nie znaleziono opcji." fit here? This is for when the user searches for a filter option, and no filter option is found for their search.

wescopeland (approved):

lang/pl_PL.json
Krylan: This was one of the things I wanted to ask about, to learn more context. Now I see it's in "Watn To Play" when using "System" dropdown search.
I think this translation would fit in there.

Krylan:
Updated file with all the rest remaining translations.
After deploying it and using live, I guess that some corrections could be found to apply.

wescopeland:
Hi @Krylan! Happy to take what you already have. Not sure if you'd like to go over the latest en_US file before or after merging this one. With the introduction of the `<Trans />` component, several keys have changed.

Krylan:
Updated translation with new keys.
I hope I found them all :D

lang/pl_PL.json
```diff
@@ -0,0 +1,347 @@
...
+    "Descending (Z - A)": "Malejąco (Z - A)",
+    "Earliest": "Najwcześniejsze",
+    "Latest": "Najpóźniejsze",
+    "Less": "Mniej",
+    "Both": "Oba",
+    "Customize View": "Dopasuj widok",
+    "Playable": "Grywalne",
+    "Not Playable": "Niegrywalne",
+    ":count System|:count Systems": ":count system|:count systemy",
+    "All Systems": "All Systems",
```
wescopeland, line 300: Possible missing translation?

lang/pl_PL.json
```diff
@@ -0,0 +1,347 @@
...
+    "Ascending (A - Z)": "Rosnąco (A - Z)",
+    "Descending (Z - A)": "Malejąco (Z - A)",
+    "Earliest": "Najwcześniejsze",
+    "Latest": "Najpóźniejsze",
+    "Less": "Mniej",
+    "Both": "Oba",
+    "Customize View": "Dopasuj widok",
+    "Playable": "Grywalne",
+    "Not Playable": "Niegrywalne",
+    ":count System|:count Systems": ":count system|:count systemy",
```
wescopeland, line 299: Definitely outside of my knowledge, but just curious: should `system` and/or `systemy` be capitalized?

wescopeland (approved):

lang/pl_PL.json
Krylan: Good catch. Fixed in next commit.

lang/pl_PL.json
Krylan: That's one of the things I was contemplating about. In Polish, we capitalize only the first word of the sentence, or if the word is a "proper name" (firstname, surname, brand, city name etc.). It's different than for example German, where every noun is capitalized.
In Polish journalism, we also don't use popular on the west "heading rules", in which most of the words starts with capitalized letter.

By the way, I was also thinking if I should use word "System" or "Platform", as the latter one is in Poland more often used in regards to consoles (like PS or Xbox). I think that the meaning of these two is similar as in English. Though, if RA uses word "System" for naming it, I followed this manner.

wescopeland: That is very interesting. Thank you for the explanation!

> By the way, I was also thinking if I should use word "System" or "Platform", as the latter one is in Poland more often used in regards to consoles (like PS or Xbox). I think that the meaning of these two is similar as in English.

If the Polish word for "Platform" has more cultural significance, that definitely seems like it would be more appropriate to use. pt_BR had a similar situation: they chose to use "Platinado" as the word for mastery as it had more cultural significance than the direct translation of "Mastered".

Happy to hold off on merging if you'd like to make this change in this branch.

Krylan: I'm not sure if I should make that change. I think that "System" could still fit in there, and that's what RA uses originally (and in URL schema as well). Optionally, we could left it as is, and if Polish community would prefer changing it later to "Platform", the change could be done.

wescopeland: It is totally your choice! 🙂
For now, shall I take the branch as-is?

Krylan: Yeah, I think that at this state it is ready to take off.

wescopeland (approved):

PR #2775, opened 2024-10-24
perf(UpdateGameAchievementsMetrics): reduce query count

wescopeland:
`UpdateGameMetrics` tends to be a long-running job that can cause high server load: https://discord.com/channels/476211979464343552/1026595325038833725/1298616276893040722.

Most of the processing time actually takes place in `UpdateGameAchievementsMetrics`, which it calls synchronously:
```php
$game->save();

app()->make(UpdateGameAchievementsMetrics::class)
    ->execute($game);
```

The main bottleneck in `UpdateGameAchievementsMetrics` is `$unlocksCount` and `$unlocksHardcoreCount` are both fetched in a loop for every achievement in the set.

This PR optimizes the action by lifting this up into a single query.

**How to test this PR:**
In _app/Platform/Commands/UpdateGameMetrics.php_, put this at the top of `handle()`:
```php
$startTime = microtime(true);
```
And put this at the bottom of `handle()`:
```php
$endTime = microtime(true);
$executionTime = $endTime - $startTime;
$this->line("Command executed in " . round($executionTime, 2) . " seconds.");
```
Then run:
```bash
sail artisan ra:platform:game:update-metrics 228
```

**UpdateGameMetrics Execution Time Before**
1.65s
1.61s
1.62s
1.66s
1.74s
1.62s
Average: 1.65s

**UpdateGameMetrics Execution Time After**
1.12s
1.16s
1.21s
1.25s
1.14s
1.18s
Average: 1.18s (~28% improvement)


app/Platform/Actions/UpdateGameAchievementsMetrics.php
```diff
@@ -20,16 +21,38 @@ public function execute(Game $game): void
...
+            ->leftJoin('UserAccounts as user', 'user.ID', '=', 'pa.user_id')
+            ->whereIn('pa.achievement_id', $achievementIds)
+            ->where('user.Untracked', false)
+            ->groupBy('pa.achievement_id')
+            ->select([
+                'pa.achievement_id',
+                DB::raw('COUNT(*) as total_unlocks'),
+                DB::raw('SUM(CASE WHEN unlocked_hardcore_at IS NOT NULL THEN 1 ELSE 0 END) as hardcore_unlocks'),
+            ])
+            ->get();
```
Jamiras, line 41: Seems like this could be more eloquent-ized. I'm not sure if that would affect performance.
```
$unlockStats = PlayerAchievement::whereIn('achievement_id' $achievementIds)
     ->where('user', function ($query) { $query->where('Untracked', 0); })
     ->groupBy('achievement_id')
     ->select(...)
     ->get()
 ```

Jamiras (approved):

app/Platform/Actions/UpdateGameAchievementsMetrics.php
wescopeland: On my local machine, this unfortunately did result in degraded performance. I believe the reason why Eloquent creates a subquery on the user stuff rather than using a left join.

However, I made some slight changes in latest to at least remove `DB::`.

PR #2778, opened 2024-10-24
support for automating achievement of the week

Jamiras:
Provides a command to generate an Achievement of the Week event game with 52 placeholders for the weekly events.
```
sail artisan ra:platform:event:create-aotw "2025-01-01" 
```

Someone with the Event Manager role can edit these event achievements to associate an achievement to track and define the period to track the achievement (the command above pre-populates the period based on the start date provided).
image

Immediately upon updating the linked achievement or active period, any unlocks of the linked achievement during the active period will be copied to the event achievement. While the period is active, any unlocks (including manual unlocks) of the linked achievement will also be copied to the event achievement.

The date fields are inclusive (saying 1/1-1/7 would indicate any unlocks on 1/1, 1/2, 1/3, 1/4, 1/5, 1/6, or 1/7 would get copied). The database fields are bookends to simplify queries (saying 1/1-1/7 would store 1/1 and 1/8).

Changing the linked achievement will update the event achievement's title, description and badge:
image

Existing unlocks are not destroyed if the achievement is changed after linking. This allows for substitutions if an achievement is found to be buggy or cheatable after becoming active. However, the existing unlocks will still be associated to the same event achievement, which will have had it's title/description/badge updated, so it will appear as if the user unlocked the replacement.

Active ranges are also shown on the event page:
image

And since these achievements are unique to the event, event participation can be easily tracked. Here's an example reconstructed from unlocks for the [2023 AotW](https://retroachievements.org/viewtopic.php?t=19685).
image

image

Users can also quickly see which achievements they've personally earned towards the event:
(Unlocked achievements are not moved to the top of the list - they're still sorted such that they appear in the order they were announced)
image
image

This will automatically award the badge if a player "masters" the event by unlocking all 52 achievements. However, the event is structured such that tiered badges can be earned for as few as 12 unlocked achievements. Hopefully, we'll have some sort of support to award tiered badges before the end of the 2025 AotW season.

Additionally, the badges are currently awarded based on the number of points earned between the AotW and AotM events. I'm not sure it would be reasonable to automate that behavior.

Final thought: This effectively allows constructing quest events (as long as each task can be associated to a single achievement). As such, should the language be changed from "event_achievements" to "quest_achievements", or is there maybe a better name?

-----

Future PRs to compete this feature:
* Modify Connect `unlocks` and `startsession` APIs to not indicate the achievement is unlocked if there's an active event achievement that the user hasn't unlocked. This will allow users to earn the event achievement without having to reset the achievement or use an alt.
* Modify `API_GetAchievementOfTheWeek` to query this data
* Modify home page widget to query this data
* Allow Most Points Earned widget to display "more..."
* Allow achievement Recent Unlocks to display "more..."
* Move logic to set associated achievement/period into filament
* Add command to generate achievement of the month event game


Jamiras:
Commands to backfill the last few years (I'm not sure if we should actually do this as there are already event pages for the previous events):
```
sail artisan ra:platform:event:create-aotw "2020-01-05" 30631,49307,92276,34853,71436,81977,84142,98264,69285,21833,12613,67132,22128,77147,23441,5854,71060,70473,106331,3181,34529,38502,108013,80055,12950,33225,7569,93487,38438,75779,66435,48348,2040,45399,70566,80104,62503,6050,105837,39623,14643,128968,100650,7287,12364,46479,67139,50940,114492,7311,78373,85393
sail artisan ra:platform:event:create-aotw "2021-01-03" 58360,16264,137451,115451,4738,117123,4673,114219,78685,136014,13126,133517,50441,31586,48543,73980,16869,47172,13566,70089,129974,125937,156172,26519,155612,156253,8306,131820,102129,109418,100414,140885,124425,15569,34402,150781,80943,101418,168388,48648,2811,80125,71178,98638,13072,140047,19595,97397,157177,1344,82308,176857
sail artisan ra:platform:event:create-aotw "2022-01-03" 82634,112016,26996,103469,47314,3524,104634,33246,3369,1904,70239,66977,27089,7071,179974,169496,177842,83357,18523,93571,56752,43874,189217,25036,217350,48117,187270,126929,51865,3078,48615,24684,142853,92424,229835,52861,150935,140379,51502,235444,165062,191610,240391,1,1801,113871,49123,146664,234608,261040,225742,28312
sail artisan ra:platform:event:create-aotw "2023-01-02" 260166,212964,230525,73042,274930,185099,24646,4838,83259,78981,4325,109069,22689,18818,58325,1416,54780,95373,173593,51710,23368,96778,295871,301448,5023,104869,157193,211745,280643,252126,188882,327046,262494,7775,36328,184233,316765,59368,110384,8290,179859,143077,178634,25966,181455,62895,363780,16928,78614,303560,22380,80855
sail artisan ra:platform:event:create-aotw "2024-01-01" 188961,341928,119310,85316,296849,183495,275711,342048,22457,259367,210457,7391,10100,231269,186149,52973,2619,165924,11404,192727,309235,234608,183470,307884,262265,95458,175853,1960,87552,412856,320990,370124,187631,293328,442338,129817,126400,260274,396136,4606,133279,398296,363705
```

resources/views/pages-legacy/achievementInfo.blade.php
wescopeland, line 1: Would you have any objection to moving the event achievement management functionality into Filament? It seems like `EventAchievement` would be a new resource, and it would be possible to attach event achievements to real achievements.

Jamiras: Nope. I just couldn't figure out how to do it cleanly. It's particularly annoying as all three fields (associated achievement, from, and until) are optional. I've added that to the "future PRs" list.

wescopeland: Based on my current understanding of the functionality of the PR, I believe it would be most appropriate to have it on the `EventAchievement` model (ie: create an `EventAchievementResource`). I'm assuming this for a few reasons:

1. We're going to be dealing with `EventAchievement` records that might not yet be associated with a source achievement, so we'll want to manage those records directly. If the related resource is `EventAchievement`, we can handle both the associated and unassociated entities in the same UI.

2. We want to be able to edit `source_achievement_id`, `active_from`, and `active_until`, regardless of whether they are currently set. If the related resource is `EventAchievement`, we should be able to more easily accommodate this.

3. While a single source achievement can be linked to multiple event achievements, each `EventAchievement` represents a distinct event achievement instance. Managing these things individually aligns well with the one-to-many relationship from the source achievement to event achievements.

Our resource might look something like this:
```php
// !!!!!!!!!!!!!!!!!!!!!!!!!!!
// !!! untested pseudocode !!!
// !!!!!!!!!!!!!!!!!!!!!!!!!!!

<?php

class EventAchievementResource extends Resource
{
    protected static ?string $model = EventAchievement::class;

    public static function form(Form $form): Form
    {
        return $form
            ->schema([
                Select::make('achievement_id')
                    ->label('Event Achievement')
                    ->relationship('achievement', 'Title')
                    ->searchable()
                    ->required(),

                Select::make('source_achievement_id')
                    ->label('Source Achievement')
                    ->relationship('sourceAchievement', 'Title')
                    ->searchable()
                    ->nullable(),

                DatePicker::make('active_from')
                    ->label('Active From')
                    ->nullable(),

                DatePicker::make('active_until')
                    ->label('Active Until')
                    ->nullable(),
            ]);
    }

    public static function table(Table $table): Table
    {
        return $table
            ->columns([
                Tables\Columns\TextColumn::make('id')
                    ->label('ID')
                    ->sortable()
                    ->toggleable(isToggledHiddenByDefault: true),

                Tables\Columns\TextColumn::make('achievement.Title')
                    ->label('Event Achievement')
                    ->sortable()
                    ->searchable(),

                Tables\Columns\TextColumn::make('sourceAchievement.Title')
                    ->label('Source Achievement')
                    ->sortable()
                    ->searchable()
                    ->toggleable(),

                Tables\Columns\DateColumn::make('active_from')
                    ->label('Active From')
                    ->sortable(),

                Tables\Columns\DateColumn::make('active_until')
                    ->label('Active Until')
                    ->sortable(),
            ])
            ->filters([
                // do we want to filter by anything?
            ])
            ->actions([
                Tables\Actions\EditAction::make(),
                Tables\Actions\DeleteAction::make(),
            ])
            ->bulkActions([
                Tables\Actions\DeleteBulkAction::make(),
            ]);
    }

    public static function getRelations(): array
    {
        return [
            //
        ];
    }

    public static function getPages(): array
    {
        // ... put the usual stuff in here i think ...
    }
}
```

Jamiras: I've moved the event achievement management logic to Filament.

Things I couldn't figure out (could use help with):
* adding audit log entry to parent achievement when event achievement added/updated
* refreshing parent achievement page when event achievement added/updated

I might be missing other things too, as my Filament expertise is lacking.

wescopeland:
```
❯ sail artisan ra:platform:event:create-aotw "2025-01-01"
Initialized game 31015
```

But, when I navigate to game 31015, I get an error. Is it possible I missed a step?

> Cannot assign null to property App\Platform\Services\GameTopAchieversService::$masteryAchievements of type int
Screenshot 2024-10-27 at 12 16 18 PM

Jamiras:
> But, when I navigate to game 31015, I get an error. Is it possible I missed a step?

It's an in-progress error. The mastery stat hasn't been calculated yet. It should be fine once the horizon queue completes.

wescopeland:
Confirmed that running all the queued jobs fixed the issue for me 👍

app/Models/Achievement.php
```diff
@@ -351,6 +353,28 @@ public function playerAchievements(): HasMany
...
+    }
+
+    /**
+     * An achievement should have at most one "event data", but this is needed for the management panel
+     * @return HasMany<EventAchievement>
+     */
+    public function eventDatas(): HasMany
+    {
+        return $this->hasMany(EventAchievement::class, 'achievement_id', 'ID');
+    }
```
wescopeland, line 376: `eventData()` is a query, not a relationship. Judging by the comment for `eventDatas()`, the definition of `eventData()` should probably look more like this:
```php
/**
 * @return HasOne<EventAchievement>
 */
public function eventData(): HasOne
{
    return $this->hasOne(EventAchievement::class, 'achievement_id');
}
```

`eventDatas()` can likely be removed altogether, due to a comment I will make elsewhere.

app/Filament/Resources/AchievementResource/RelationManagers/EventAchievementsRelationManager.php
wescopeland, line 1: I spent some time digging into `eventDatas` and Filament's answer to `HasOne` relationships like this. During my digging, I came across this documentation: https://filamentphp.com/docs/3.x/panels/resources/relation-managers#layout-form-components---saving-form-fields-to-a-single-relationship

This suggests that for relations such as these, it's preferable to treat the `HasOne` as just another set of related fields on the parent resource. This allows us to lift the relation up into the main form, remove the relation manager altogether, and get away with having less code.

I didn't want to send you on a wild goose chase, so I implemented this locally to be sure it was a viable approach:

Screenshot 2024-10-27 at 2 03 46 PM

This requires `eventData` to be defined as a true `HasOne` relationship (see my comment elsewhere). This also doesn't take care of executing `CopyAchievementUnlocksToEventAchievement()`. I'll have another comment in this PR for the approach there.

In `AchievementResource::form()`:
```php
Forms\Components\Fieldset::make('Event')
    ->relationship('eventData')
    ->schema([
        Forms\Components\Select::make('source_achievement_id')
            ->label('Source Achievement')
            ->columnSpan(2)
            ->searchable()
            ->getSearchResultsUsing(function (string $search): array {
                return Achievement::where('Title', 'like', "%{$search}%")
                    ->orWhere('ID', 'like', "%{$search}%")
                    ->limit(50)
                    ->get()
                    ->mapWithKeys(function ($achievement) {
                        return [$achievement->id => "[{$achievement->id}] {$achievement->title}"];
                    })
                    ->toArray();
            })
            ->getOptionLabelUsing(function (int $value): string {
                $achievement = Achievement::find($value);

                return "[{$achievement->id}] {$achievement->title}";
            }),

        Forms\Components\DatePicker::make('active_from')
            ->label('Active From')
            ->native(false)
            ->date(),

        Forms\Components\DatePicker::make('active_through')
            ->label('Active Through')
            ->native(false)
            ->date(),
    ])
    ->hidden(fn ($record) => $record->game->system->id !== System::Events)
```

app/Filament/Resources/AchievementResource/RelationManagers/EventAchievementsRelationManager.php
```diff
@@ -0,0 +1,135 @@
...
+            ])
+            ->actions([
+                Tables\Actions\ActionGroup::make([
+                    Tables\Actions\EditAction::make()
+                        ->visible(function (EventAchievement $eventAchievement) use ($user) {
+                            return $user->can('update', $eventAchievement);
+                        })
+                        ->after(function (EventAchievement $eventAchievement, Component $livewire) {
+                            $eventAchievement->refresh();
+                            (new CopyAchievementUnlocksToEventAchievement())->execute($eventAchievement);
```
wescopeland, line 116: It is my understanding we'll want this action to run any time `source_achievement_id`, `active_from`, and/or `active_until` on the event achievement are modified.

If that assumption is true, I think it will be much safer to use a Laravel model observer, which I've verified is working on my local:

```php
// app/Observers/EventAchievementObserver.php

<?php

declare(strict_types=1);

namespace App\Observers;

use App\Models\EventAchievement;
use App\Platform\Actions\CopyAchievementUnlocksToEventAchievement;

class EventAchievementObserver
{
    public function saved(EventAchievement $eventAchievement): void
    {
        $isFieldDirty = $eventAchievement->isDirty(['source_achievement_id', 'active_from', 'active_until']);
        $isConnectedToGameAchievement = $eventAchievement->achievement_id && $eventAchievement->source_achievement_id;

        if ($isFieldDirty && $isConnectedToGameAchievement) {
            (new CopyAchievementUnlocksToEventAchievement())->execute($eventAchievement);
        }
    }
}
```

```php
// app/Providers/EventServiceProvider.php

public function boot(): void
{
    // User::observe(UserObserver::class);
    EventAchievement::observe(EventAchievementObserver::class);
}
```

Now, Laravel will automatically execute the action any time one of those three fields changes. This is beneficial for two reasons:
1. Filament's sometimes-wacky API has areas where after-save hooks are not provided.
2. Regardless of how code gets changed or refactored in the future, we can trust that the observer will keep running.

app/Platform/Actions/CopyAchievementUnlocksToEventAchievement.php
```diff
@@ -0,0 +1,43 @@
...
+        $eventAchievement->loadMissing(['achievement', 'sourceAchievement']);
+
+        $achievement = $eventAchievement->achievement;
+        $sourceAchievement = $eventAchievement->sourceAchievement;
+
+        if ($achievement && $sourceAchievement) {
+            $achievement->title = $sourceAchievement->title;
+            $achievement->description = $sourceAchievement->description;
+            $achievement->BadgeName = $sourceAchievement->BadgeName;
+            $achievement->save();
```
wescopeland, line 25: During my local testing and trying to break things, this caught me by surprise. I didn't expect an action named `CopyAchievementUnlocksToEventAchievement` to be mutating these primary data fields on the event achievement.

Maybe it's worth considering a different name for this action.

wescopeland:
> As such, should the language be changed from "event_achievements" to "quest_achievements"

I think `event_achievements` is fine for now. It aligns well with the "system" name at the moment. Before naming it something like `quest_achievements`, we should probably think more about the design of quests themselves. Ideally, a quest could have more requirements than just unlocking achievements, such as getting an award associated with a game/set.

app/Platform/Actions/CopyAchievementUnlocksToEventAchievement.php
Jamiras: It used to be called `UpdateEventAchievement` and thought that was too generic. I'm open to suggestions.

Jamiras: Actually, I think the action can be eliminated and this code just moved into the Observer

Jamiras:
Made suggested changes: relationship -> embedded; action -> observer

wescopeland:
When navigating to an event game while unauthenticated, I get an error:
Screenshot 2024-10-30 at 8 02 23 PM

wescopeland:
The above could actually be an issue in `master` and not your branch. Taking a deeper look...

wescopeland:
Confirmed, the issue above is my fault. Please disregard. I am going to open a tiny PR to resolve.

app/Filament/Resources/AchievementResource.php
```diff
@@ -208,6 +228,41 @@ public static function form(Form $form): Form
                                 ->disabled(!$user->can('updateField', [$form->model, 'DisplayOrder'])),
                         ]),
                 ])->from('md'),
+
+                Forms\Components\Fieldset::make('Event Association')
+                    ->relationship('eventData')
+                    ->schema([
+                        Forms\Components\Select::make('source_achievement_id')
```
wescopeland, line 235: This is looking a lot better. One minor suggestion: I think we may want to change this from a `Fieldset` to a `Section`, and give it a `->columns()` value, like:

```php
Forms\Components\Section::make('Event Association')
    ->relationship('eventData')
    ->columns(['xl' => 2, '2xl' => 4])
    ->schema([
```

After this change, the new section blends in with the others on the page as well as View mode:


Screenshot 2024-10-30 at 8 19 34 PM

app/Models/EventAchievement.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public function setActiveThroughAttribute(Carbon|string|null $value): void
+    {
+        if (is_string($value)) {
+            $value = Carbon::parse($value);
+        }
+
+        $this->active_until = $value ? $value->clone()->addDays(1) : null;
+    }
```
wescopeland, line 47: nit: This is a mutator. Recommend moving this down to the mutators section.

app/Observers/EventAchievementObserver.php
```diff
@@ -0,0 +1,54 @@
...
+            $eventAchievement->loadMissing(['achievement', 'sourceAchievement']);
+
+            $achievement = $eventAchievement->achievement;
+            $sourceAchievement = $eventAchievement->sourceAchievement;
+
+            if ($achievement && $sourceAchievement) {
+                // make the event achievement look like the source achievement
+                $achievement->title = $sourceAchievement->title;
+                $achievement->description = $sourceAchievement->description;
+                $achievement->BadgeName = $sourceAchievement->BadgeName;
```
wescopeland, line 33: What happens if any of these values change in the source achievement after the event achievement has been set? Should there be something to resync the event achievement? Or would it be better to almost treat the event achievement like a "pointer" of sorts to the source?

app/Models/EventAchievement.php
wescopeland, line 1: Comment unrelated to this file. We probably want to restrict setting the `type` on event achievements.

Screenshot 2024-10-30 at 8 31 12 PM

wescopeland: The changes in _achievementInfo.blade.php_ look good. There's one place left in Filament the type is still settable:

Screenshot 2024-11-03 at 7 13 14 AM

This is on the actual "game" page for the event in the achievements relation manager.

Jamiras: "bulk set type" is now hidden for event games.

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -219,7 +220,8 @@ public function table(Table $table): Table
                         }),
                 ])
                     ->label('Bulk set type')
-                    ->visible(fn (): bool => $user->can('updateField', [Achievement::class, null, 'type'])),
+                    ->hidden($this->getOwnerRecord()->system->id === System::Events)
+                    ->visible(fn ($record): bool => $user->can('updateField', [Achievement::class, null, 'type'])),
```
wescopeland, line 224: There's some risk here. Filament's API seems unclear about whether `hidden()` or `visible()` takes precedence, and the current precedence rules could change in a Filament update.

We may want to consider combining this into a single `visible()` and removing `hidden()`:

```php
// untested pseudocode

->visible(function ($record) use ($user) {
    // $record might work instead of $this
    if ($this->getOwnerRecord()->system->id === System::Events) {
        return false;
    }

    return $user->can('updateField', [Achievement::class, null, 'type']);
});
```

wescopeland (approved):
Thanks for working through all my feedback.

wescopeland (approved):

PR #2781, opened 2024-10-26
Include link to Supported Game Files when creating new game thread

Hexadigital:
Following from https://discord.com/channels/310192285306454017/1299607482842349608

Is it fine to use the raw emoji here or would something else be preferred? I noticed it was being used like this in resources/views/components/game/link-buttons/index.blade.php, but I wasn't sure if it'd be better to have \u{1F4BE} or similar.


app/Helpers/database/forum.php
```diff
@@ -317,6 +317,7 @@ function generateGameForumTopic(User $user, int $gameId): ?ForumTopicComment
 
     $topicPayload = "Official Topic Post for discussion about [game=$gameId]\n" .
         "Created " . date("j M, Y H:i") . " by [user={$user->User}]\n\n" .
+        "[b][url=https://retroachievements.org/linkedhashes.php?g=$gameId]💾 Supported Game Files[/url][/b]\n\n" .
```
wescopeland, line 320: This is a redirect URL. The current URL is `route('game.hashes.index', ['game' => $gameId])`.

Thanks for asking about the emoji. For the time being I think my preference would be no emoji or iconography associated with the link.

wescopeland (commented):
Hi @Hexadigital! Thanks for helping us knock this out.

app/Helpers/database/forum.php
Hexadigital: Oh whoops, I didn't even think about that. Let me adjust it real quick.

Hexadigital:
Thanks for the catch! I've updated my change to use the URL builder and removed the emoji.

wescopeland (approved):
Verified locally.

Note, in local environments `$hashesURL` will use an absolute path at `localhost`, which causes the redirect gate to be raised.

When manually changing the base URL to be `retroachievements.org`, this does not occur.

PR #2783, opened 2024-10-27
feat: migrate Create Ticket page to React

wescopeland:
This PR migrates the Create Ticket page's back-end to Laravel controllers and the page itself to React. Roughly 45% of this PR's diff is tests.

Screenshot 2024-10-27 at 5 23 11 PM

Behaviors I've verified:
* The `?extra`  param (and it's now covered by tests).
* Automatic prepopulation of all applicable fields.
* Duplicate ticket prevention & redirect to existing ticket.
* Unhelpful ticket description validation / warnings.
* Mobile support.

There are also a few behavior changes:

**All modal dialogs have been removed**
The text is now in-line in the form. This also has the side-effect of improving accessibility of the form.

Screenshot 2024-10-27 at 5 27 28 PM

**The Issue dropdown now has 3 choices**
There is a new strategically-placed third choice: "Triggered, but now showing as earned on the site". When selected, the submit button is disabled (the form is invalid) and the page yells at the user.

Screenshot 2024-10-27 at 5 28 15 PM

Screenshot 2024-10-27 at 5 28 51 PM

**The description field now has a template**
Similar to other structured messages, the description field now has a template which hopefully can serve as a mad-lib for helping users write better tickets. The user cannot submit the form if the field remains pristine.

Screenshot 2024-10-27 at 5 31 06 PM


resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/GameHashSelectField.tsx
```diff
@@ -0,0 +1,81 @@
...
+          >
+            {t('Supported Game File')}
+          </BaseFormLabel>
+
+          <div className="flex w-full flex-col gap-1">
+            <BaseFormControl>
+              <BaseSelect
+                onValueChange={field.onChange}
+                defaultValue={field.value ? String(field.value) : undefined}
+              >
```
Jamiras, line 56: Is there some way to make this stretch full? It's weird to see it wrap when there's clearly more room on the right:
image

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
```diff
@@ -0,0 +1,96 @@
...
+  hash: z.string().min(1),
+  description: z
+    .string()
+    .min(25, { message: 'Please be more detailed in your description.' })
+    .refine((desc) => !/(n'?t|not?).*(work|trigger)/i.test(desc), {
+      message: 'Please be more detailed in your description.',
+    })
+    .refine((desc) => !/(internet|wifi|connection|network)/i.test(desc), {
+      message: 'Do not open tickets for network issues.',
+    }),
```
Jamiras, line 28: This appears to completely prevent submission. There are times where "connection" or "network" might be appropriate to include in the description.

> I can't seem to unlock the connection between John and Jane in my social network.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
```diff
@@ -0,0 +1,96 @@
...
+  emulatorVersion: z.string().optional(),
+  core: z.string().optional(),
+  mode: z.enum(['hardcore', 'softcore']),
+  hash: z.string().min(1),
+  description: z
+    .string()
+    .min(25, { message: 'Please be more detailed in your description.' })
+    .refine((desc) => !/(n'?t|not?).*(work|trigger)/i.test(desc), {
+      message: 'Please be more detailed in your description.',
+    })
```
Jamiras, line 25: "doesn't work" may be valid. I think the old logic had a minimum length check too.

> This doesn't appear to work if you haven't been to see the king.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
```diff
@@ -0,0 +1,96 @@
...
+  /** @see TicketType.php */
+  issue: z.enum(['DidNotTrigger', 'TriggeredAtWrongTime', 'NetworkIssue']),
+  emulator: z.string().min(1),
+  emulatorVersion: z.string().optional(),
+  core: z.string().optional(),
+  mode: z.enum(['hardcore', 'softcore']),
+  hash: z.string().min(1),
+  description: z
+    .string()
+    .min(25, { message: 'Please be more detailed in your description.' })
```
Jamiras, line 22: This is almost pointless with the template. You can add a space anywhere in the text and hit submit without providing any information.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
```diff
@@ -0,0 +1,96 @@
...
+  const onSubmit = async (formValues: CreateAchievementTicketFormValues) => {
+    toastMessage.promise(mutation.mutateAsync(formValues), {
+      loading: t('Submitting...'),
+      success: (submitResponse) => {
+        setTimeout(() => {
+          const { ticketId } = submitResponse.data;
+
+          // TODO use router.visit after migrating this page to React
+          window.location.href = route('ticket.show', { ticket: ticketId });
+        }, 1000);
```
Jamiras, line 74: Can this button be disabled while submitting? The delay here offers plenty of opportunity to click the button multiple times before the page turns.
image

This first time I did it because I wasn't sure I had actually clicked on it because the page turn wasn't immediate.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/IssueSelectField.tsx
```diff
@@ -0,0 +1,114 @@
...
+                  </BaseSelectItem>
+                  <BaseSelectItem value="NetworkIssue">
+                    {t('Triggered, but not showing as earned on site')}
+                  </BaseSelectItem>
+                </BaseSelectContent>
+              </BaseSelect>
+            </BaseFormControl>
+
+            {field.value ? (
+              <BaseFormDescription className="text-menu-link light:text-neutral-800">
```
Jamiras, line 61: Should these example texts be muted like the helper text for the other fields?

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/GameHashSelectField.tsx
wescopeland: Not a problem - made this change in latest.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
wescopeland: In latest, I believe I've updated the form validation / info message logic to mirror what Alpine.js is doing.

The form should only be invalidated by the description now if the user enters unhelpful text that's less than 25 characters long.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
wescopeland: True - similar to the existing page, a warning will now display, but the form will not be invalidated.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
wescopeland: I went ahead and decided to remove the template in latest.

I think if we're going to do a template, it'd probably make sense to build it into the form itself as actual templated fields, rather than a placeholder value the user has to go out of their way to replace.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/useCreateAchievementTicketForm.ts
wescopeland: Not a problem - updated in latest.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/IssueSelectField.tsx
wescopeland: Yes, I think so. The most important helper text is the highlighted text near the description field. Otherwise, they should probably all be muted. Updated in latest.

Note that I did not use the actual muted text color for this -- the contrast is not high enough against the page's background.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/DescriptionField.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const DescriptionField: FC = () => {
+  const { t } = useLaravelReactI18n();
+
+  const form = useFormContext<CreateAchievementTicketFormValues>();
+
+  const [description] = form.watch(['description']);
+
+  const showTriggerWarning =
+    description.length < 25 && /(n'?t|not?).*(work|trigger)/gi.test(description);
```
Jamiras, line 28: There seems to be something else checking for a minimum length without providing any feedback. If I just type in something like "Please fix this", there's no feedback at all, and the Submit button does nothing.

wescopeland: Looks like validation is kicking in correctly but not surfacing a visual label. I've adjusted this in latest. On submit, the validation error should now correctly be displayed:

Screenshot 2024-11-06 at 5 30 13 PM

Jamiras: "this achievement doesn't work." doesn't show any error message and doesn't submit.

Jamiras: Looks like I can't submit long strings of keyboard vomit either.

wescopeland: I believe both issues are now addressed in latest:

* "this achievement doesn't work" should now display the proper validation error.
* Required fields were not displaying validation errors, which I why I suspect the submit on keyboard vomit appeared to do nothing. These fields now display validation errors if values are missing.

Jamiras (approved):

PR #2785, opened 2024-10-28
chore(forwardToDiscord): add logging

wescopeland:
Adds a bunch of logging to `forwardToDiscord()` to try to debug some of the spurious issues we've been seeing.


app/Community/Listeners/NotifyMessageThreadParticipants.php
```diff
@@ -74,7 +76,26 @@ private function forwardToDiscord(
...
+
         $message->body = Shortcode::stripAndClamp($message->body, 1850, preserveWhitespace: true);
+        Log::info("[{$messageThread->title}] Discord forward: Message body processed", [
+            'message_id' => $message->id,
+            'thread_id' => $messageThread->id,
+            'from_user' => $userFrom->username,
+            'to_user' => $userTo->username,
+            'original_title' => $messageThread->title,
+            'processed_length' => mb_strlen($message->body),
+            'truncated' => mb_strlen($message->body) === 1850,
```
Jamiras, line 96: This should be 1853. "..." is appended after clamping.

Jamiras: Or perhaps just ">= 1850"?

Jamiras: Also applies to copies below

Jamiras (commented):
Why not just log the payload and the response?

wescopeland:
> Why not just log the payload and the response?

I was particularly curious if anything in our code was perhaps causing the outgoing data to Discord to be corrupted somehow, and if so, where.

I've cut a lot down in latest.

app/Community/Listeners/NotifyMessageThreadParticipants.php
```diff
@@ -141,6 +143,7 @@ private function forwardToDiscord(
                     //  - Original:  "Unwelcome Concept: Lots of Rings [12345] (Sonic the Hedgehog)"
                     //  - Formatted: "12345: Lots of Rings (Sonic the Hedgehog)"
                     if (preg_match('/^(Incorrect type:|Issue:|Unwelcome Concept:|Writing:)\s*(.*)\s*\[([0-9]+)\]\s*(\(.*\))$/', $messageThread->title, $titleMatches)) {
+                        $originalTitle = $messageThread->title;
```
Jamiras, line 146: This doesn't look to be used.

wescopeland: Looks like I missed that line. Removed in latest.

Jamiras (approved):

PR #2786, opened 2024-10-28
refactor(game-list): consolidate repetitive backlog toggle logic into a hook

wescopeland:
No functional changes. Cuts quite a bit of duplication down between the three backlog toggle buttons in the game list:
* The desktop row action that appears on the right-hand side.
* The mobile list item action that appears on the right-hand side.
* The mobile bottom drawer button.

Now all this logic is centralized into a single reusable hook.

Testing this PR is a matter of ensuring all three of those backlog toggles should continue functioning as normal. I've done manual testing on this and there are some automated tests also covering this.


resources/js/features/game-list/components/GameListItems/GameListItemDrawerBacklogToggleButton.tsx
```diff
@@ -1,72 +1,43 @@
...
-
   return (
     <BaseButton
       variant="secondary"
       className={cn(
         'border disabled:!pointer-events-auto disabled:!opacity-100',
         isInBacklogOptimistic ? '!border-red-500 !bg-embed !text-red-500' : 'border-transparent',
       )}
-      onClick={handleToggleFromBacklogClick}
+      onClick={toggleBacklog}
```
Jamiras, line 40: Clicking remove from the popup all the Want to Play list (in mobile) does not remove the item from the list, nor does it change the X to a plus.

Similarly, adding a game from the popup on the All Games list (in mobile) does not change the plus to an X. Furthermore, clicking on the plus generates a Something Went Wrong error.

wescopeland: Thanks for catching that.

I've made some changes in latest to reduce duplicate backlog states, as well as improve this mobile UX.

Now, clicking remove from the drawer in the Want to Play list cleanly closes the drawer, then removes the list item and pops a toast.

I've double-checked table rows, list item rows, and the drawer toggle on all games & the want to play list.

Jamiras (approved):

PR #2790, opened 2024-10-29
perf(BuildGameListAction): make All Games substantially faster

wescopeland:
This PR dramatically improves the performance of the All Games list.

Currently the query takes approximately 350-450ms on my local machine. This is due to the join from `GameData` to `Console`. This PR removes the join and filters `GameData` directly.

**Before**
364ms
369ms
358ms
363ms
359ms
Average = 363ms

**After**
9.83ms
9.86ms
10.14ms
9.91ms
10.42ms
Average = 10.03ms

Note that these benchmarks were taken on my personal account, which has developer privileges. This means the aggregate query for open ticket counts is always running.

For unauthenticated users, with this optimization I'm finding the query takes less than 1ms.


app/Platform/Actions/BuildGameListAction.php
```diff
@@ -158,10 +159,14 @@ private function buildBaseQuery(GameListType $listType, ?User $user = null): Bui
         switch ($listType) {
             case GameListType::AllGames:
                 // Exclude non game systems, inactive systems, and subsets.
+                $validSystemIds = System::query()
+                    ->active()
+                    ->gameSystems()
+                    ->pluck('ID')
+                    ->all();
```
Jamiras, line 166: `query()` is unnecessary:
```
                $validSystemIds = System::gameSystems()
                    ->active()
                    ->pluck('ID')
                    ->all();
```

Jamiras (approved):

app/Platform/Actions/BuildGameListAction.php
wescopeland: Removed in latest.

PR #2772, opened 2024-10-22
feat(home): build initial React components

wescopeland:
This PR is the first of **four** PRs that will migrate the home page to Inertia.js / React:

PR 1 of 4: (This PR) Build some basic components with hardcoded data. Lots of functionality missing, no back-end.
PR 2 of 4: Use real data from the back-end / Laravel controllers rather than hardcoded data.
PR 3 of 4: Rebuild the Active Players component.
PR 4 of 4: Finalize the swap of the PHP home page for the React home page.

---

The intent of this PR is to build basic components for the home page that contain hardcoded data and extremely simple code coverage. This serves as a foundation to actually rebuild the page.

**Lots of pieces of functionality are still missing which will arrive in subsequent PRs.**

The React page is available at `/demo/home`:

Screenshot 2024-10-21 at 8 51 31 PM

I will describe each component in brief.

### News
* The carousel is gone.
* Three news posts and a "See more" link (to a currently non-existent page) are visible.

Screenshot 2024-10-21 at 8 52 41 PM

Screenshot 2024-10-21 at 8 53 43 PM

### New Sets/Revisions
* Renamed to "Just Released".
* Virtually identical with the existing Blade UI.

Screenshot 2024-10-21 at 8 54 37 PM

Screenshot 2024-10-21 at 8 54 47 PM

### Active Players
* Shows a static empty state at the moment. This component will be rebuilt in PR 3 of 4. **This will have an entire PR dedicated towards it**, as it is by far the most complex component on the page.

### Currently Online
* No longer uses Google Charts. Now uses Recharts/D3.js.
* Shows the all-time recorded high (currently hardcoded, but it's available in the logs in server storage).

https://github.com/user-attachments/assets/93fa3f60-16d3-45fb-9de6-632991b11d73

### Sets in Progress
* Renamed to "Latest Sets in Progress".
* Virtually identical with the existing Blade UI.

Screenshot 2024-10-21 at 8 58 23 PM

Screenshot 2024-10-21 at 8 58 32 PM

### Forum Activity
* Renamed to "Forum Recent Posts".
* Virtually identical with the existing Blade UI.

Screenshot 2024-10-21 at 8 59 55 PM

Screenshot 2024-10-21 at 9 00 08 PM

### Top Links
* Uses similar iconography for points as introduced in #2758.
* Is much more responsive at tablet breakpoints.

Screenshot 2024-10-21 at 9 01 01 PM

Screenshot 2024-10-21 at 9 01 20 PM

### Achievement of the Week
[...]

PR #2794, opened 2024-10-30
chore: switch from npm to pnpm

wescopeland:
In the Node.js ecosystem, there are three commonly-used package managers: npm, yarn, and pnpm.

There is a bug in npm that keeps requiring me to reset our entire lockfile just to be able to merge changes. This has happened at least three times now: https://github.com/RetroAchievements/RAWeb/pull/2772#issuecomment-2448586914

Most Node.js projects that reach our scale have to decide if continuing to use vanilla npm is right for them, often due to issues like what I've been facing recently. I am strongly of the opinion we will be better-positioned for the future if we switch to pnpm now.

**pnpm is natively supported by Sail and already installed on the build server.**

There are [many](https://pnpm.io/feature-comparison) benefits to using pnpm. For us, the main one is it can automatically resolve lockfile merge conflicts. More information about the motivation behind pnpm's creation can be found [here](https://pnpm.io/motivation).

`api-docs` and `api-js` have already been using pnpm for months. `docs` is still using npm only because I didn't want non-technical contributors to have to worry about one more thing.

**How to test this PR**
Delete your _node_modules_ folder. Then run `sail pnpm install`. pnpm will reinstall _node_modules_ in a more optimized way described in the motivation link above.

All commands should continue working normally. Instead of `npm run x`, do `pnpm x`, eg:
```
pnpm dev
pnpm lint
pnpm tsc
pnpm test
pnpm test:coverage
pnpm build
```


Jamiras (approved):

wescopeland:
I've updated the deployment scripts in all envs to use pnpm instead of npm. Stage was successfully deployed with the pnpm changes active.

PR #2780, opened 2024-10-25
feat: migrate `/user/{user}/comments` to React

wescopeland:
This PR builds on https://github.com/RetroAchievements/RAWeb/pull/2762 by migrating the user comments page to React. There is actually very little new UI code here. A large chunk of this PR is just code coverage, autogenerated code, and refactoring to make stuff more reusable.

Screenshot 2024-10-24 at 10 02 38 PM

Behaviors I've verified:
* If the user's wall is disabled, other users cannot visit it.
* Policies are still respected.

PR #2762, opened 2024-10-18
feat: migrate `/game/{game}/comments` to React

wescopeland:
This PR migrates the `/game/{game}/comments` route from Blade to React.

The UI has very nearly 100% code coverage. There should be only a few functional changes.

https://github.com/user-attachments/assets/b34d6ba6-1358-49f3-be4e-767732b55eaa

Behaviors that I've verified:
* Unauthenticated users are prompted to sign in and not shown a textarea.
* The textarea still auto-expands and validates against input size.
* Submitting a comment sends the user to the correct page.
* Deleting a comment sends the user to the correct page.
* Submitting a comment for the first time to a game wall subscribes the user to that game wall.
* The user can still toggle their subscription.
* The user can paginate, and paginations are client-side navigations.
* Muted users are not shown a textarea and are informed of their mute status.

Some new behaviors worth noting:
* The textarea's placeholder text has been slightly modified.
* Previously, banned user comments were removed by PHP template code. Now they're filtered out at the controller level, based on user privileges to actually see them.
* Submitting a comment will always take the user to the correct page on submit.
* Hovering over timestamps now uses a proper `<BaseTooltip />` component (forum recent posts is also affected by this change).
* User absolute timestamp preferences are now respected.

**Next steps:**
* Migrate the other comment pages.

PR #2797, opened 2024-10-31
feat: migrate `/achievement/{achievement}/comments` to React

wescopeland:
This is another PR that builds on #2762 by migrating the achievement comments page to React. Similar to #2780, there is very little new UI code here. The only real behavior change is the UI now renders Server comments differently, similar to the Blade UI.

Screenshot 2024-10-31 at 6 44 53 PM

Behaviors I've verified:
* Server comments are not deletable.
* The page loads for unauthenticated users.
* Policies are still respected.
* Automatic subscriptions work correctly.


app/Policies/AchievementCommentPolicy.php
```diff
@@ -68,9 +72,12 @@ public function update(User $user, AchievementComment $comment): bool
 
     public function delete(User $user, AchievementComment $comment): bool
     {
+        if ($comment->is_automated) {
+            return false;
+        }
+
```
Jamiras, line 78: I'm unable to delete my own comments without the MODERATOR role.

wescopeland: Great catch. Looks like there was a bug in the policy. Fixed in latest.

Jamiras (approved):

PR #2801, opened 2024-11-02
fix: save user's LastLogin value on page navigation

wescopeland:
Currently, `$user->LastLogin` is being set when `authenticateFromCookie()` is called and when a player activity record is stored.

In other words, for any page that doesn't use `authenticateFromCookie()`, the `LastLogin` value is not updated when navigating to that page (ie: all Inertia pages, maybe Folio pages too?).

Our active player count on the home page is being underreported as a result of this issue.

This PR:
* Adds a new middleware which runs on page navigation. The middleware updates the user's `LastLogin` value if they are authenticated.

### DIFF
--- a/app/Http/Middleware/UpdateLastLogin.php
+++ b/app/Http/Middleware/UpdateLastLogin.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Middleware;
+
+use App\Models\User;
+use Carbon\Carbon;
+use Closure;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+
+class UpdateLastLogin
+{
+    public function handle(Request $request, Closure $next): mixed
+    {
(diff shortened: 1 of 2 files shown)


app/Http/Middleware/UpdateLastLogin.php
```diff
@@ -0,0 +1,27 @@
...
+class UpdateLastLogin
+{
+    public function handle(Request $request, Closure $next): mixed
+    {
+        /** @var ?User $user */
+        $user = Auth::user();
+
+        if ($user) {
+            $user->LastLogin = Carbon::now();
+            $user->save();
```
Jamiras, line 22: Since this runs on all pages, should the equivalent code be removed from `authenticateFromCookie`?

wescopeland: Good call. Removed from `authenticateFromCookie` in latest.

Jamiras (approved):

PR #2802, opened 2024-11-02
feat(home): implement `HomeController::index()`

wescopeland:
This is PR 2 of the React home page migration, and a follow-up to https://github.com/RetroAchievements/RAWeb/pull/2772. These are the planned steps of the migration:

PR 1 of 4: Build some basic components with hardcoded data. Lots of functionality missing, no back-end.
**PR 2 of 4: (we are here) Use real data from the back-end / Laravel controllers rather than hardcoded data.**
PR 3 of 4: Rebuild the Active Players component.
PR 4 of 4: Finalize the swap of the PHP home page for the React home page.

Note related to news: there are new changes in https://github.com/RetroAchievements/RAWeb/pull/2572, a previously-approved PR.

Rather than a before and after, I will include screenshots of what the UI looks like on my local machine now that the components are receiving real data:

Screenshot 2024-11-02 at 5 38 29 PM

Screenshot 2024-11-02 at 5 39 18 PM

Screenshot 2024-11-02 at 5 39 25 PM

The UI has 100% code coverage, and all the back-end code is fairly close with 90%+. I will add comments in the code itself below for things that I feel should be highlighted.


resources/views/layouts/app.blade.php
wescopeland, line 1: There is a bug here that will need to be mitigated in a subsequent PR.

Currently, when you navigate to another Inertia page from `demo.home`, the giant RetroAchievements logo at the top of the page doesn't disappear. This is because it's a client-side navigation rather than a full page reload.

tests/Feature/Http/Controllers/HomeControllerTest.php
```diff
@@ -0,0 +1,665 @@
...
+        $game = Game::factory()->create([
+            'ConsoleID' => $system->id,
+        ]);
+
+        // TODO use event achievements
+        $achievement = Achievement::factory()->create([
+            'ID' => 9,
+            'Title' => 'That Was Easy',
+            'GameID' => $game->id,
+        ]);
```
wescopeland, line 158: Just noting more things to update here when we cut over to using event achievements.

app/Community/Actions/BuildThinRecentForumPostsDataAction.php
```diff
@@ -0,0 +1,89 @@
...
+         * syntax highlighting.
+         */
+        $results = DB::select(<<<SQL
+            SELECT LatestComments.DateCreated AS PostedAt,
+                LatestComments.Payload,
+                ua.User as Author,
+                ua.display_name as AuthorDisplayName,
+                ua.RAPoints,
+                ua.Motto,
+                ft.ID AS ForumTopicID,
```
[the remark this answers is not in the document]
gboquizosanchez: And something like this or similar? 

```
$subQuery = DB::table('ForumTopicComment as ftc')
    ->select('*')
    ->whereRaw($userClause)
    ->orderBy('ftc.DateCreated', 'desc')
    ->limit($limit + 20); // cater for spam messages

$latestComments = DB::table(DB::raw("({$subQuery->toSql()}) as LatestComments"))
    ->mergeBindings($subQuery)
    ->join('ForumTopic as ft', 'ft.ID', '=', 'LatestComments.ForumTopicID')
    ->leftJoin('Forum as f', 'f.ID', '=', 'ft.ForumID')
    ->leftJoin('UserAccounts as ua', 'ua.ID', '=', 'LatestComments.author_id')
    ->select([
        'LatestComments.DateCreated as PostedAt',
        'LatestComments.Payload',
        'ua.User as Author',
        'ua.display_name as AuthorDisplayName',
        'ua.RAPoints',
        'ua.Motto',
        'ft.ID as ForumTopicID',
        'ft.Title as ForumTopicTitle',
        'LatestComments.author_id as author_id',
        'LatestComments.ID as CommentID'
    ])
    ->where('ft.RequiredPermissions', '<=', $permissions ?? Permissions::Unregistered)
    ->whereNull('ft.deleted_at')
    ->orderBy('LatestComments.DateCreated', 'desc')
    ->limit($limit)
    ->get();
```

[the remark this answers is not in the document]
wescopeland: That works great, thanks! Updated in latest.

app/Helpers/database/release.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 use App\Models\System;
+use Illuminate\Support\Facades\Cache;
```
Jamiras, line 4: It's not this line, but something in this PR is causing an error when loading a game from RAlibretro:
```
111855.649|INFO| << startsession response (500): {"Success":false,"Error":"Declaration of Laravel\\Horizon\\RedisQueue::pop($queue = null) must be compatible with Illuminate\\Queue\\RedisQueue::pop($queue = null, $index = 0)","Status":500}
```
I don't know why it isn't happening in the unit test.

resources/js/features/home/components/+sidebar/TopLinks/TopLinks.tsx
```diff
@@ -11,54 +11,100 @@ import {
...
           {t('Global Points Ranking')}
         </a>
 
-        <a href={route('ranking.beaten-games')} className={baseButtonVariants({ size: 'sm' })}>
+        <a
+          href={route('ranking.beaten-games')}
+          className={cn(
+            baseButtonVariants({ size: 'sm' }),
+            'Click Top Link Global Beaten Games Ranking',
+          )}
```
Jamiras, line 51: I'm getting a 500 error trying to access this page (happens on master too), but not on stage, so must be something in my environment. It's not writing an error log, so I don't know how to diagnose.

wescopeland: On master, does it also happen on the normal home page, or just /demo/home ?

wescopeland: If you try manually running `composer install` does that change anything?

Jamiras: It's not the link, it's the page itself.
image

`composer install` did not change the behavior.

wescopeland: Ah, I misinterpreted entirely earlier and understand the issue very clearly now.

It's an out-of-memory error and unrelated to the changes in this PR. I bet if you check your terminal, you'll see it yelling about running out of memory / memory allocation / etc when you visit the unfiltered beaten games leaderboard. If you add a filter manually in the address bar, the page will magically start working again.

When we built the beaten games leaderboard, the memory allocation of the page was fine. Now that we've grown, the total unfiltered list uses more memory. I'll need to hunt that down.

Other pages exhibit this problem too, for example: http://localhost:64000/codenotes.php?g=2721
This is the code notes page for Dragon Quest VIII. It has an ungodly high amount of code notes - enough to crash our local envs without bumping the memory limit.

You can get around this by bumping the PHP memory limit. To do so, open _docker/8.2/php.ini_ and change `memory_limit` to something like 256M. For the memory limit change to take effect, you have to rebuild your Sail images like so:
```bash
sail down
sail build --no-cache
sail up
```

app/Helpers/database/release.php
wescopeland: I'm having trouble reproducing this issue:

Screenshot 2024-11-04 at 8 29 25 PM

Just for the sake of ruling things out, in latest I've reverted the _composer.json_ and _composer.lock_ changes. Does that fix things for you on latest?

Jamiras: I can confirm that the problem does not occur on 9f375249f8872adc641e369ff0941091387edc5c but does on de4bc5f8946807a0583624ddf3c93de2724b1fff.

And I can reproduce from the command line:
```
$ curl "localhost:64000/dorequest.php?r=startsession&u=Jamiras&t=redacted&g=1&m=1bc674be034e43c96b86487ac69d9293"
{"Success":false,"Error":"Declaration of Laravel\\Horizon\\RedisQueue::pop($queue = null) must be compatible with Illuminate\\Queue\\RedisQueue::pop($queue = null, $index = 0)","Status":500}
```

The problem also doesn't occur if I comment out this line: https://github.com/RetroAchievements/RAWeb/blob/e735815e6e112bf9426b5ad85c59b16204ce0391/public/dorequest.php#L541

Jamiras (approved):

PR #2803, opened 2024-11-02
refactor(ForumTopicController): migrate recentPosts logic to an action

wescopeland:
This PR thins out `ForumTopicController::recentPosts()` so that most of that method's logic lives in an action.

The action is still untestable due to the query using code that is unsupported by SQLite.

There is a single functional change, which I will document in the code below.


app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
```diff
@@ -0,0 +1,137 @@
...
+        $topics = [];
+        if ($page === null || $page <= 3) {
+            // The query to get topics is kinda expensive, and it's called
+            // frequently. Therefore, let's use a stale-while-revalidate
+            // caching strategy so this is fast for most users.
+            $topics = Cache::flexible(
+                "recent-topics:{$permissions}:{$limit}:{$page}",
+                [10, 20],
+                fn () => $this->getRecentForumTopics($page ?? 1, $permissions, $limit)
+            );
```
wescopeland, line 37: [Cache::flexible()](https://laravel.com/docs/11.x/cache#swr) was added in Laravel 11.3, and is particularly useful here.

The first time the data is fetched, it's cached in Redis.

For 10 seconds, the data is considered "fresh". Any request within this 10-second window retrieves the cached data directly without refetching from the database.

The data can be served as "stale" for an additional 10 seconds (from the 10th to 20th second). During this stale period, if a request is made, the cached (stale) data will be served to the user, and the cache will be magically refreshed by the server in the background. This prevents users from experiencing a delay while waiting for a recalculated value.

If the page goes 20 seconds without receiving a request, the cache is purged/expired, and a fresh value will be calculated on the next page load.

wescopeland: I actually just checked our analytics, and the page gets so little traffic, I don't think this is worth it.

Jamiras (approved):

PR #2807, opened 2024-11-06
perf: migrate from laravel-react-i18n to react-i18next

wescopeland:
This PR switches our React i18n engine from `laravel-react-i18n` to `react-i18next`.

**There should be no functional changes in this branch.**

## Why?
`laravel-react-i18n` is not code splitting locale strings correctly. As a result, users are receiving code for locales they are not even using. This is directly observable by running `pnpm build` in the latest `master`. You will see this:

Screenshot 2024-11-06 at 5 06 35 PM

As we add more strings and locales, this will continue to grow exponentially, ultimately causing hydration to become slower and slower as the core JS bundle gets larger and larger.

I originally reached for `laravel-react-i18n` because of how nicely it integrates with Laravel itself. However, this issue is untenable, and fixing it is very nontrivial. I don't want to spend more time addressing issues with this library; I want to spend time solving problems for RetroAchievements.

After migrating to i18next, nearly 100KB is shaved from the uncompressed bundle:
Screenshot 2024-11-06 at 5 09 50 PM

## Engine Differences

* i18next was first released in 2011, and is battle-tested & widely-adopted with thousands of stars on GitHub.
* Interpolated values now have delimiters. `:foo` is now `{{foo}}`.
* We no longer need to roll our own `<Trans />` component - we can just use i18next's. All of our `<Trans />` usages have been converted over.
* i18next is smart enough to format interpolated number values automatically without the need to explicitly call a `formatNumber()` function.
* We can no longer do Laravel-style choice `:count game|:count games` for pluralization. These have to be two separate keys with a suffix like `gameCount_one` and `gameCount_other`.

## How to test this PR

```bash
sail pnpm install
sail pnpm build
```

It's important to test in SSR mode to ensure there are no hydration errors.

You should be able to visit various React pages and see that en_US strings are correct and working.

Thankfully, large refactors such as this are fairly safe given the UI's ~95% code coverage. Another benefit of switching to i18next is we no longer need to mock the hook responsible for translating things. The real hook is now used by components in tests, giving us even more confidence.

## Final Note

A new command has been added to _package.json_: `i18n:sync`. It can be executed by running `pnpm i18n:sync`.

This takes all the keys from `en_US.json` and propagates them to the other locale files. It also sorts all keys.


Jamiras (approved):
performed minimal spot checking

PR #2809, opened 2024-11-06
chore: always hide phpdebugbar at XS breakpoint

wescopeland:
This is a tiny QoL PR that causes the phpdebugbar button to always be hidden at the XS breakpoint, even if the debug bar is enabled. It is unusable at this breakpoint and is only getting in the way.

**Before (check bottom left)**
Screenshot 2024-11-06 at 6 15 24 PM

**After**
Screenshot 2024-11-06 at 6 15 49 PM


resources/css/utilities.css
```diff
@@ -54,6 +54,10 @@
   }
 }
 
+div.phpdebugbar {
+  @apply hidden sm:block;
+}
```
Jamiras, line 59: Would it be possible (would it make sense) to still show the size/branch/language?
image

Jamiras (approved):

resources/css/utilities.css
Jamiras: Also, `master` is not correct :(
How is that determined?

wescopeland: Thanks for pointing this out. I didn't even notice that all this stuff was incorrect. The locale is incorrect, too.

Unfortunately, it looks like the current branch is determined from an environment variable. After thinking things over, I've made the following changes in an attempt to make this debug element a bit more useful:

* If the environment is `local`, then the branch is no longer shown.
* Right now, language is tightly coupled to locale. I've removed display of the language.
* Locale is now correctly shown (try changing to pl_PL or pt_BR).
* On mobile, the debug content displays as a row.

Screenshot 2024-11-09 at 3 00 50 PM

Screenshot 2024-11-09 at 3 01 12 PM

resources/views/layouts/components/body-end.blade.php
```diff
@@ -10,43 +10,46 @@ function handleClick(el) {
...
     <div
         id="debug"
         role="button"
         x-init="{}"
         @click="handleClick($el)"
-        class="hidden sm:flex text-2xs flex-col rounded-lg bg-black/20 p-1"
+        @class([
+            'fixed left-0 -bottom-0.5 z-50 sm:left-1',
+            'flex gap-1 sm:flex-col rounded bg-black p-1 text-neutral-200 text-2xs',
+            app()->environment('local') ? 'sm:bottom-10' : 'sm:bottom-2.5'
```
gboquizosanchez, line 23: I suggest you `app()->isLocal()` here

resources/views/layouts/components/body-end.blade.php
```diff
@@ -10,43 +10,46 @@ function handleClick(el) {
...
+        ])
     >
-        <b class="text-danger text-capitalize">
+        <span class="font-bold text-danger text-capitalize">
             {{ app()->environment() }}
             {{ $_SERVER['LARAVEL_OCTANE'] ?? false ? '[Octane]' : '' }}
-            ({{ config('app.branch') }})
-        </b>
+            @if (app()->environment() !== 'local')
```
gboquizosanchez, line 29: And here as well

wescopeland: Resolved both in latest

PR #2810, opened 2024-11-06
ci: cache and fan out from install steps

wescopeland:
We currently have an issue where, sometimes, `composer install` hangs during CI. This happens probably every 1 in 15 commits for me: https://github.com/RetroAchievements/RAWeb/actions/runs/11713046341

When this happens, we unnecessarily blow through a lot of CI minutes. I started thinking about ways to fix this, and it occurred to me we should rarely even need to install at all.

With the changes in this PR:
* `composer install` will now only happen when _composer.lock_ has changes.
* `pnpm install` will now only happen when _pnpm-lock.yml_ has changes.
* When those files have changes, installs occur and the dependencies are cached.
* If those files do not have changes, the caches are pulled and loaded instead.
* The three PHP tasks fan out from the PHP Setup task which sets up composer dependencies.
* The three Node.js tasks fan out from the Node.js Setup task which sets up pnpm dependencies.


Jamiras (approved):
I don't really understand the problem, but the renamed CI tasks seem to be passing. We can proceed with these changes and see if the issue crops up again.

PR #2811, opened 2024-11-07
fix(BaseDrawerFooter): adjust footer padding

wescopeland:
This PR makes a very minor adjustment to the `BaseDrawerFooter` component to make it play more nicely with iOS Safari. Bottom padding has ever-so-slightly been increased so the buttons are no longer flush with the home bar on the bottom of the screen.

Screenshot 2024-11-06 at 7 42 26 PM


resources/js/common/components/+vendor/BaseDrawer.tsx
```diff
@@ -74,7 +74,7 @@ const BaseDrawerHeader = ({ className, ...props }: React.HTMLAttributes<HTMLDivE
 BaseDrawerHeader.displayName = 'BaseDrawerHeader';
 
 const BaseDrawerFooter = ({ className, ...props }: React.HTMLAttributes<HTMLDivElement>) => (
-  <div className={cn('mt-auto flex flex-col gap-2 p-4', className)} {...props} />
+  <div className={cn('mt-auto flex flex-col gap-2 px-4 pb-6 pt-4', className)} {...props} />
```
Jamiras, line 77: Should the drawer height be adjusted to account for the change ? It's slightly obscuring the top bar now:
image

before:
image

Or is that just coincidental?

Jamiras (approved):

resources/js/common/components/+vendor/BaseDrawer.tsx
wescopeland: Good question to ask. Coincidental and likely should not be changed - this will unfortunately be something that varies on a device-by-device basis.

PR #2812, opened 2024-11-07
fix(game-list): show Players column instead of Progress column for guests

wescopeland:
This PR makes a slight change to the React datatable default state.

Now, if the user visits `/games` and is unauthenticated, they will see the "Players" column instead of the "Progress" column.


resources/js/features/game-list/components/AllGamesMainRoot/AllGamesMainRoot.tsx
```diff
@@ -25,7 +25,10 @@ export const AllGamesMainRoot: FC = () => {
     setPagination,
     setSorting,
     sorting,
-  } = useGameListState(paginatedGameListEntries, { defaultColumnFilters: allGamesDefaultFilters });
+  } = useGameListState(paginatedGameListEntries, {
+    canShowProgressColumn: !!auth?.user,
+    defaultColumnFilters: allGamesDefaultFilters,
```
Jamiras, line 30: Should the "Add to Want to Play Games" button also be hidden? It does redirect to login, but doesn't actually add the game after logging in.

Jamiras (approved):

resources/js/features/game-list/components/AllGamesMainRoot/AllGamesMainRoot.tsx
wescopeland: I lean towards no but would be interested in a 2nd opinion if you feel differently.

The table below is traffic from `/login` that lands there from somewhere on our site. Most of it is from the home page, but we get referrers from other places too.

I'm able to see that guest users do indeed click that button while unauthenticated and do not bounce from the site, which means they are logging in. That is true, but the volume is also relatively low.

Screenshot 2024-11-09 at 11 15 23 AM

Jamiras: I have no strong opinion. It's just broken currently, so it should either be fixed, or removed.

PR #2818, opened 2024-11-08
chore: name actions consistently

wescopeland:
No functional changes.

Right now there is an inconsistency in the codebase regarding actions.

Some actions are named like `FooAction`, whereas others are just named like `Foo`. This PR ensures all actions are named like `FooAction`.


app/Actions/CreateNewUser.php
Jamiras, line 1: Should these actions also be renamed?

Jamiras (approved):

app/Actions/CreateNewUser.php
wescopeland: I originally considered this, but am not entirely sure. These are implementing Laravel Fortify methods which don't use `execute()` methods, so they're definitely not traditional actions even though they do only expose a single public method.

PR #2822, opened 2024-11-09
add versionCompare function to UserAgentService

Jamiras:
Allows determining which version is newer between two versions given the version formats currently in the database.


tests/Feature/Platform/Services/UserAgentServiceTest.php
```diff
@@ -318,4 +356,64 @@ public function testDolphinAlphaUserAgent(): void
...
+            'client' => 'Flycast',
+            'clientVersion' => '45bf218df',
+            'extra' => [
+                'CFNetwork' => '1498.700.2',
+                'Darwin' => '23.6.0',
+            ],
+        ], $this->parseUserAgent($userAgent));
+
+         // we can't enforce minimum versions on git hashes
+         $this->assertEquals(0, UserAgentService::versionCompare('45bf218df', '2d4684a46'));
```
wescopeland, line 403: Is the intention to treat all Git hashes as equal versions?

For example:
```php
$hashOne = "abcdef1";
$hashTwo = "1234567";

$result = UserAgentService::versionCompare($hashOne, $hashTwo);

$this->assertEquals(0, $result); // FAILS
```
I could be misinterpreting the code though.

app/Platform/Services/UserAgentService.php
```diff
@@ -163,4 +164,101 @@ private function trimOperatingSystem(string $os): string
...
+
+                // neither part is fully numeric. compare them as strings
+                return ($diff < 0) ? -1 : 1;
+            }
+
+            $index++;
+        }
+    }
+
+    private static function splitVersion(string $version): array
```
wescopeland, line 220: What happens if I release version 1.0a and it's followed by 1.0B?

wescopeland, line 220: Possible edge case: multiple trailing characters may not be handled, ie- 1.0ab. Not sure if this is even a relevant issue.

wescopeland, line 220: What if there's a "v1.2.3" and just a "1.2.3"?

wescopeland, line 220: What about a case like "1.2.3-beta.1" and "1.2.3-beta.2"? Not sure how this is handled currently - this may already be covered.

tests/Feature/Platform/Services/UserAgentServiceTest.php
Jamiras: hashTwo is completely numeric, so it's considered version "1234567". hashOne is alphanumeric, so it's ignored. As a result, the "unknown" version is considered older than the "1234567" version.

wescopeland: Ah, that makes sense.

app/Platform/Services/UserAgentService.php
Jamiras: `1.0a` is converted to `1.0.1` and `1.0b` is converted to `1.0.2` by this code:
https://github.com/RetroAchievements/RAWeb/pull/2822/files#diff-0e7cd5df70a77fe7e572b4a371f1cf4390006cf2cac2c57378ab285fbbe4c679R248

wescopeland: `1.0b` is handled correctly, but the algorithm is case sensitive, eg:
```php
public function testSampleAgent(): void
{
    $this->assertEquals(1, UserAgentService::versionCompare('1.0B', '1.0a')); // FAILS
}
```

tests/Feature/Platform/Services/UserAgentServiceTest.php
wescopeland, line 403: nit: Minor typo here, these are spaced over 1 character too far.

app/Platform/Services/UserAgentService.php
Jamiras: This is not handled. Both versions will be normalized to "1.2.3.[extra]" and sort after "1.2.3", but be equivalent to each other. It's a similar problem to "1.2.3-branch1" and "1.2.3-branch2", where the order can't really be determined, we just know they're newer than "1.2.3".

tests/Feature/Platform/Services/UserAgentServiceTest.php
Jamiras: fixed

app/Platform/Services/UserAgentService.php
Jamiras: Alpha/beta versions may generically problematic in any scenario, as "1.2-beta" should presumably be _before_ "1.2".

Jamiras: fixed

wescopeland (approved):

PR #2824, opened 2024-11-09
fix(DataTableSearchInput): prevent auto-zoom on iOS

wescopeland:
This PR fixes an annoying issue on the mobile games list that only happens on iOS.

When focusing on an input field, iOS auto-zooms the window if the input field has a font size smaller than 16px. Unfortunately, it isn't possible to repro this issue without a real device or the iPhone simulator built into macOS. I've recorded the videos below using the simulator.

**Before**

https://github.com/user-attachments/assets/f80d96dd-7492-4a49-a72b-5ab198769fa6

**After**

https://github.com/user-attachments/assets/757455cf-ddbc-4c3f-a5e1-ec2b9d77ca79


Jamiras (approved):
first video doesn't play for me, so I'm not entirely sure what the problem is, but I can see the text is slightly bigger when viewing the page in mobile mode.

PR #2827, opened 2024-11-10
use event achievement data for API_GetAchievementOfTheWeek

Jamiras:
followup to #2778

  [X] Modify API_GetAchievementOfTheWeek to query this data (fallback to static data)


public/API/API_GetAchievementOfTheWeek.php
```diff
@@ -82,26 +103,49 @@
...
-];
+if ($achievementOfTheWeek) {
+    $unlocks = collect();
 
-$parentGame = getParentGameFromGameTitle($game['Title'], $achievementData['ConsoleID']);
-$unlocks = getAchievementUnlocksData((int) $achievementID, null, $numWinners, $numWinnersHardcore, $numPossibleWinners, $parentGame?->id, 0, 500);
+    $playerAchievements = $achievementOfTheWeek->achievement->playerAchievements()
+        ->with('user')
+        ->orderBy(DB::raw('IFNULL(unlocked_hardcore_at, unlocked_at)'))
+        ->get();
```
wescopeland, line 112: If `$achievementOfTheWeek` is achievement ID 9, I get an out of memory error when running this query:

```php
$achievement = Achievement::find(9);

$playerAchievements = $achievement
  ->playerAchievements()
  ->with("user")
  ->orderBy(DB::raw("IFNULL(unlocked_hardcore_at, unlocked_at)"))
  ->get();
```

This is pulling too much data. I think we may need to implement pagination at the endpoint level via `o` (offset) and `c` (count) params.

wescopeland: Sorry. I actually think it may be my interpretation of this code that's incorrect.

I think I'm getting confused by the `achievement` and `sourceAchievement` relations. In this case, `sourceAchievement` would be ID 9. `achievement` is the literal event achievement (visible on the game page for the event).

Please disregard this feedback assuming I made a mistake here.

wescopeland (approved):

public/API/API_GetAchievementOfTheWeek.php
Jamiras: It might be worth putting an upper bound on that anyway. The old code had a limit of 500, though the imported data for 2024 shows the most unlocked AotW has 292 entries.

wescopeland (approved):

PR #2829, opened 2024-11-10
use event achievement data for AotW widget on demo home page

Jamiras:
followup to #2778

[X] Modify home page widget to query this data

only updated the /demo/home code.

will fallback to static AotW data so we don't have to create AotW event achievements until 2025.


resources/js/features/home/components/+sidebar/AchievementOfTheWeek/AchievementOfTheWeek.tsx
```diff
@@ -54,15 +63,30 @@ export const AchievementOfTheWeek: FC = () => {
...
+                        <span>{t('Ends')} </span>
+                        <DiffTimestamp
+                          at={achievementOfTheWeek.activeUntil}
+                          asAbsoluteDate={auth?.user.preferences.prefersAbsoluteDates}
+                        />
+                      </span>
+                    </div>
+                  ) : null}
+                </div>
               </div>
```
wescopeland, line 82: This is pretty cool. I'm looking forward to seeing this go live. Happy to help button this up.

First, let's fix the CSS thing. This was kinda tricky to solve. Here's the solution:
```tsx
<div className="flex w-full flex-col gap-0.5">
  <GameAvatar {...game} showImage={false} />

  <div className="flex w-full items-center justify-between">
    <SystemChip {...system} className="bg-zinc-800" />

    {achievementOfTheWeek.activeUntil ? (
      <span className="smalldate">
        {/* TODO see below */}
      </span>
    ) : null}
  </div>
</div>
```

Next, for the `smalldate` content, this is a circumstance we'll need to reach for `<Trans />` to interpolate the value rather than concatenate it. In SOV languages and some of our top 10 locales by traffic, the timestamp will come before the label, so we'll need to facilitate that working correctly.

First, in _en_US.json_, adjust to:
```json
"Ends <1>{{when}}</1>": "Ends <1>{{when}}</1>"
```

The `<Trans />` syntax is kinda weird. Here's the solution:

```tsx
<span className="smalldate">
  <Trans
    i18nKey="Ends <1>{{when}}</1>"
    values={{ when: achievementOfTheWeek.activeUntil }}
    components={{
      1: (
        <DiffTimestamp
          at={achievementOfTheWeek.activeUntil}
          asAbsoluteDate={auth?.user.preferences.prefersAbsoluteDates}
        />
      ),
    }}
  />
</span>
```

In the UI, you should now see "Ends [some diff]". To prove that `<Trans />` is working correctly, go back to _en_US.json_ and adjust to:
```json
"Ends <1>{{when}}</1>": "<1>{{when}}</1> Ends"
```
Now when you refresh the UI, you'll see "[some diff] Ends".

Working with these interpolations takes some getting used to. I'm always happy to help with these as they come up.

resources/js/features/home/components/+sidebar/AchievementOfTheWeek/AchievementOfTheWeek.test.tsx
wescopeland, line 1: The new test cases look great. Thank you for adding them 👍

There is an uncovered branch I think we'll want to add another test case for: showing the right timestamp when the user has their absolute date preference set.

Here's how I might go about writing such a test case:
```tsx
it('given the user prefers absolute dates and the achievement has an end date, displays the remaining time and respects the user preference', () => {
  // ARRANGE
  vi.setSystemTime(dayjs.utc('2024-11-15').toDate()); // add `import dayjs from 'dayjs'` to the top of the file

  const now = dayjs.utc();
  const tomorrow = now.add(64, 'hours');
  const sourceAchievement = createAchievement({ id: 9, title: 'That Was Easy' });
  const achievementOfTheWeek = createEventAchievement({
    achievement: sourceAchievement,
    sourceAchievement: sourceAchievement,
    activeUntil: tomorrow.toISOString(),
  });

  render<App.Http.Data.HomePageProps>(<AchievementOfTheWeek />, {
    pageProps: {
      auth: { user: createAuthenticatedUser({ preferences: { prefersAbsoluteDates: true } }) },
      ...createHomePageProps({ achievementOfTheWeek }),
    },
  });

  // ASSERT
  expect(screen.getByText(/ends/i)).toBeVisible();
  expect(screen.getByText(/nov 17, 2024, 16:00/i)).toBeVisible();
});
```

`dayjs` is kinda like the front-end's equivalent to Carbon.

The spread operator (`...createHomePageProps()`) makes building the page props easier. If you've never seen it before, here's how it works:
```ts
const list1 = [1, 2, 3];
const list2 = [4, 5, 6];
const combinedList = [...list1, ...list2]; // [1, 2, 3, 4, 5, 6]
```

Jamiras: I assumed the date vs offset would be tested by the control's test file. I guess it makes sense to ensure we're passing the correct parameters to the control.

Jamiras: Does the setting description need to be changed? It currently says "on forum posts".
image

Maybe this should always display remaining time?

wescopeland: > Maybe this should always display remaining time?
Good question ... I lean towards this behind the right approach. If we go that route, we obviously wouldn't have a missing branch to cover, as the `asAbsoluteDate` prop would be hardcoded to `false` I think.

> Does the setting description need to be changed? It currently says "on forum posts".
Yeah, we should probably change this setting label no matter what we decide here. Maybe just "Prefer absolute dates"? That way it doesn't imply it'll always be available - just available where we think it makes sense.

Jamiras: I've changed the setting to "Prefer absolute dates" as suggested. As a result, I feel that having the control toggle its behavior is still appropriate.

wescopeland (approved):

PR #2830, opened 2024-11-10
feat(game-list): colorize the backlog toggle button

wescopeland:
Right now it's pretty tough to visually parse at a glance if a game is on the user's backlog or not. This PR colorizes the button in "Remove" mode so it's easier to see at a glance what the current state is.

**Before**
Screenshot 2024-11-10 at 10 01 16 AM

Screenshot 2024-11-10 at 10 01 32 AM

**After**
Screenshot 2024-11-10 at 10 00 39 AM

Screenshot 2024-11-10 at 10 00 51 AM


Jamiras (approved):
https://github.com/RetroAchievements/RAWeb/pull/2727#discussion_r1782020943
> I see a red X and I think delete.

But it seems like the red isn't as pronounced now, so it's probably okay.

PR #2833, opened 2024-11-10
refactor(loadDayjsLocale): simplify, add coverage

wescopeland:
This PR simplifies the `loadDayjsLocale()` util and bumps coverage to 100%.


resources/js/common/utils/l10n/loadDayjsLocale.ts
```diff
@@ -6,30 +6,27 @@ import dayjs from 'dayjs';
...
+        break;
 
-  if (userLocale === 'es_ES') {
-    try {
-      await import('dayjs/locale/es.js');
-      dayjs.locale('es');
-    } catch (err) {
-      console.warn('Unable to load Day.js locale for es_ES.', err);
+      default:
+        console.warn(`Locale ${userLocale} is not supported.`);
```
Jamiras, line 27: What happens when a locale is not supported besides logging this message? Does it default to `en_US`? Also, shouldn't `en_US` have a case to prevent logging this message?

wescopeland: Yes - the UI always falls back to en_US in that circumstance:
Screenshot 2024-11-13 at 6 30 01 PM

However, there is indeed a bug where if the locale is en_US, the UI logs this locale is unsupported. In latest, I've fixed this bug and added the test case you've recommended to verify the fix.

Jamiras (approved):

PR #2834, opened 2024-11-10
fix(l10n): remove hardcoded string, remove capitalize classname

wescopeland:
Addresses feedback from https://discord.com/channels/310192285306454017/1299083037908406402/1305276501729742859.


resources/js/features/game-list/components/DataTableFacetedFilter/DataTableFacetedFilter.tsx
```diff
@@ -82,14 +82,14 @@ export function DataTableFacetedFilter<TData, TValue>({
           size="sm"
           className={cn(
             'border-dashed',
-            buildTrackingClassNames(`Click ${title} Filter`),
+            buildTrackingClassNames(`Click ${t_title} Filter`),
```
Jamiras, line 85: For metrics tracking, shouldn't this use an untranslated string?

resources/js/features/game-list/components/DataTableFacetedFilter/DataTableFacetedFilter.tsx
```diff
@@ -82,14 +82,14 @@ export function DataTableFacetedFilter<TData, TValue>({
...
             className,
           )}
-          data-testid={`filter-${title}`}
+          data-testid={`filter-${t_title}`}
```
Jamiras, line 88: This also seems like something that might be better off using a non-translated string. Perhaps even a string without spaces.

resources/js/features/game-list/components/DataTableFacetedFilter/DataTableFacetedFilter.tsx
wescopeland: Good call - in latest I've changed this to use the column ID instead.

resources/js/features/game-list/components/DataTableFacetedFilter/DataTableFacetedFilter.tsx
wescopeland: Agreed, I've also changed this in latest to now use the column ID.

Jamiras (approved):

PR #2842, opened 2024-11-13
feat(multiset): add action to resolve sets for emulators

wescopeland:
Creates an action where, given a `GameHash` and a `User`, returns the correct list of achievement sets to be loaded by an emulator. The action is called `ResolveAchievementSetsAction`.

Additionally, migrates the attach code currently used in Filament to a reusable action (so it can be used to test the new action).

The resolve action is not currently surfaced anywhere. Tests are present to verify it works.

Next steps:
* Use the new action in `patch` if a game hash is present in the request.


app/Platform/Actions/AssociateAchievementSetToGameAction.php
```diff
@@ -0,0 +1,58 @@
...
+
+    private function associateAchievementSet(
+        Game $game,
+        AchievementSet $achievementSet,
+        AchievementSetType $type,
+        string $title
+    ): void {
+        $game->achievementSets()->attach($achievementSet->id, [
+            'type' => $type->value,
+            'order_column' => 1,
```
Jamiras, line 52: Should this be `$game->achievementSets()->count() + 1`?

app/Platform/Actions/AssociateAchievementSetToGameAction.php
```diff
@@ -0,0 +1,58 @@
...
+            'title' => $title,
+            'created_at' => now(),
+            'updated_at' => now(),
```
Jamiras, line 55: Aren't these automatic?

app/Platform/Actions/ResolveAchievementSetsAction.php
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @return array [int $gameIdToUse, array|null $typesToLoad]
+     */
+    private function buildResolutionContext(GameAchievementSet $initialSet, int $defaultGameId): array
+    {
+        $links = GameAchievementSet::where('achievement_set_id', $initialSet->achievement_set_id)->get();
+
+        $exclusiveLink = $links->firstWhere('type', AchievementSetType::Exclusive);
+        $specialtyLink = $links->firstWhere('type', AchievementSetType::Specialty);
+        $bonusLink = $links->firstWhere('type', AchievementSetType::Bonus);
```
Jamiras, line 63: Suggest moving these down next to their `if`s. 

I realize the query only runs once and these are filters, but each of the `if`s can short-circuit out, and I prefer to keep the logic next to the usage.

app/Platform/Actions/ResolveAchievementSetsAction.php
```diff
@@ -0,0 +1,201 @@
...
+     * @return Collection<int, GameAchievementSet>
+     */
+    private function sortSets(Collection $sets): Collection
+    {
+        return $sets->sortBy(function (GameAchievementSet $set) {
+            return match ($set->type) {
+                AchievementSetType::Exclusive => 0,
+                AchievementSetType::Core => 1,
+                AchievementSetType::Bonus => 2,
+                AchievementSetType::Specialty => 3,
```
Jamiras, line 173: I wonder if specialty should be above bonus. If I'm playing a low-level specialty set, I might care more about it that the bonus sets.

tests/Feature/Platform/Action/ResolveAchievementSetsActionTest.php
```diff
@@ -0,0 +1,506 @@
...
+        $resolved = (new ResolveAchievementSetsAction())->execute($baseGameHash, $user);
+
+        // Assert
+        $this->assertCount(1, $resolved);
+
+        $coreSet = $resolved->first();
+        $this->assertAchievementSet($coreSet, AchievementSetType::Core, $baseGame->id, 6, 1);
+    }
+
+    public function testItIncludesSubsetIfUserIsGloballyOptedOutButLocallyOptedIn(): void
```
Jamiras, line 346: I may not have seen them, but what about:
* Globally opted out of subsets, load subset hash -> expect subset and core (or just subset?)
* Explicitly opted out of subset, load subset hash -> expect subset and core (or just core?)

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
```diff
@@ -197,38 +199,26 @@ public function table(Table $table): Table
                             ]),
                     ])
                     ->action(function (array $data) use ($game) {
-                        // Add a new game_achievement_set row. The existing row stays
-                        // New row is basically the same as the old row, but the game_id and type are different values.
-                        // It'll use the same achievement_set_id.
+                        try {
```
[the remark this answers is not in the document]
wescopeland: There's some protection built into this search to prevent users from attaching the wrong things as subsets.

First and foremost, only subsets with the same system ID as the current game you're viewing are shown:
```php
Game::where('Title', 'like', "%[Subset -%")
    ->where('ConsoleID', $game->ConsoleID)
    ->where('achievements_published', '>', 0)
```

For game 28275, the list is only populated with a single item because that's the only standalone subset that has any published achievements.

For game 764, searching for "final" only shows two results because those are the only GBA subsets with the word "final" in their name.

For the case where you typed "happ" and got mixed results, I assume that's because Filament's `->searchable()` attribute is actually doing a client-side fuzzy search from the server-side fetched results. Pokemon Sapphire Version ranks 1st because it scores highest in the fuzzy search criteria.

app/Platform/Actions/AssociateAchievementSetToGameAction.php
wescopeland: It definitely shouldn't be what it is right now. I opted for:
```php
'order_column' => ($game->achievementSets()->max('order_column') ?? 0) + 1,
```
`max()` looks for the highest value rather than counting all rows, so slightly more efficient. Also, if some rows get deleted, we'll still get the next number in the order sequence, whereas with `count()` we may wind up with unexpected duplicate order values.

app/Platform/Actions/AssociateAchievementSetToGameAction.php
wescopeland: They weren't, but are now in latest. I needed to update the relationship in the game model:
```php
public function achievementSets(): BelongsToMany
{
    return $this->belongsToMany(AchievementSet::class, 'game_achievement_sets', 'game_id', 'achievement_set_id', 'ID', 'id')
        ->withPivot(['type', 'title', 'order_column'])
        ->withTimestamps('created_at', 'updated_at');
}
```

app/Platform/Actions/ResolveAchievementSetsAction.php
wescopeland: Updated in latest.

app/Platform/Actions/ResolveAchievementSetsAction.php
wescopeland: Happy to make this change, updated in latest.

tests/Feature/Platform/Action/ResolveAchievementSetsActionTest.php
wescopeland: These cases were not included. This was really tough to solve, but I think I've landed somewhere I'm happy with.

I've added two new test cases to cover these bullet points. I've also updated some of the behavior of the action. All the existing tests remain untouched & passing.

For your first bullet point:
* The user is **globally opted out** of subsets.
* They load a bonus subset's hash in their emulator.

I think for this case, they should be treated as still living in a pre-multiset environment. The only set that should be resolved is the one for the bonus set, just like what happens today.

For your second bullet point:
* The user is **globally opted in** to subsets.
* The user is opted out of Bonus Set X.
* They load a hash associated with Bonus Set X in their emulator.

For this case, I think the action should behave just like they loaded a hash associated with the base game / core set. They get the core set, any other bonus sets they're not opted out of, and they _don't_ get the subset they are opted out of.

In other words, let's say we have these three sets:
* Dragon Quest III
* Dragon Quest III [Subset - Bonus]
* Dragon Quest III [Subset - Bonus 2]

The user has explicitly opted out of [Subset - Bonus]. They then load a hash associated with [Subset - Bonus]. The core set and [Subset - Bonus 2] are going to be resolved. The action will behave just like they loaded a hash associated with the base game.

This is the only behavior I can think of that keeps things simple and will be easy to document/explain to people. If you're opted out, then the system will respect that no matter what.

A test has been added which verifies the above scenario.

app/Platform/Actions/AssociateAchievementSetToGameAction.php
Jamiras: So you don't need to set them?

Jamiras (approved):

app/Platform/Actions/AssociateAchievementSetToGameAction.php
wescopeland: Looks like I forgot to kill those two lines. Removed & verified behavior in latest.

PR #2777, opened 2024-10-24
chore: add translations for fr_FR

amine4567:
This add french (France) localization

PR #2846, opened 2024-11-14
fix(l10n): remediate <Trans /> component bugs, enable fr_FR

wescopeland:
A bunch of `<Trans />` components are not rendering for non-English locales correctly in `master`. This branch remediates that issue. Additionally, enables fr_FR from #2777.


Jamiras (approved):
Only tested a couple of the changes. Assuming the rest are okay.

PR #2847, opened 2024-11-15
fix: sync i18n plurals correctly

wescopeland:
i18next plural forms are not syncing correctly across locales with the `pnpm i18n:sync` command.

For example, while this is valid in en_US:
```json
{
    "supportedGameFilesCountLabel_one": "There is currently <1>{{count, number}}</1> supported game file hash registered for this game.",
    "supportedGameFilesCountLabel_other": "There are currently <1>{{count, number}}</1> supported game file hashes registered for this game.",
}
```

It is actually _invalid_ in pl_PL, as pl_PL has multiple plural forms:
```json
{
    "supportedGameFilesCountLabel_one": "Obecnie jest <1>{{count, number}}</1> zarejestrowany hash pliku gry",
    "supportedGameFilesCountLabel_few": "Obecnie są <1>{{count, number}}</1> zarejestrowane hashe pliku gry",
    "supportedGameFilesCountLabel_many": "Obecnie jest <1>{{count, number}}</1> zarejestrowanych hashy plików gry",
    "supportedGameFilesCountLabel_other": "Obecnie jest <1>{{count, number}}</1> zarejestrowanych hashy plików gry"
}
```

When there are 2 supported game files and the user is using pl_PL, i18next is naturally reaching for the "_few" key, but it currently isn't present, causing that string to fall back to English.

After the changes in this file, `pnpm i18n:sync` syncs plural keys correctly. The new sync script is intelligent enough to recognize what plural forms the target locale file requires, and it writes the keys accordingly.

---

**How to test this PR**
```
pnpm install
```

Add the following to en_US.json:
```
    "people_one": "Person",
    "people_other": "People"
```
Run the command:
```
pnpm i18n:sync
```
Proper plural forms should be written to each JSON file, ie: pl_PL:
```json
    "people_one": "Person",
    "people_few": "People",
    "people_many": "People",
    "people_other": "People"
```


package.json
```diff
@@ -15,7 +15,9 @@
         "test:watch": "TZ=UTC vitest",
         "tsc": "tsc --noEmit --skipLibCheck",
         "analyze": "cross-env ANALYZE=true pnpm prod",
-        "i18n:sync": "sync-i18n --files 'lang/*.json' --primary en_US --languages en_GB es_ES pl_PL pt_BR --space 4"
+        "i18n:rename-to-dashes": "renamer --find \"_\" --replace \"-\" \"lang/*.json\"",
+        "i18n:rename-to-underscores": "renamer --find \"-\" --replace \"_\" \"lang/*.json\"",
```
wescopeland, line 19: The sync script uses filenames to auto-recognize the source/target locales. This presents a challenge: in the Laravel ecosystem, locales are conventionally written as "en_US". But in the Node.js ecosystem, they are conventionally written as "en-US".

Before the sync command runs, quickly rename all the files to the Node.js standard, and then after the sync command is done, name them back.

The files should remain using the Laravel convention, as it is plausible there will be strings returned from the back-end that need to be localized.

lang/es_ES.json
```diff
@@ -384,11 +384,13 @@
     "players are currently online.": "jugadores actualmente en línea.",
     "players in-game.": "jugadores echando una partida.",
     "supportedGameFilesCountLabel_one": "Actualmente <1>{{count, number}}</1> registro de hash de archivo soportado registrado para este juego.",
+    "supportedGameFilesCountLabel_many": "Actualmente hay registrados <1>{{count, number}}</1> hashes de archivos soportados registrados para este juego.",
```
Jamiras, line 387: Your description specifically mentions `_few` keys missing, but you didn't add any. Was that intentional?
> When there are 2 supported game files and the user is using pl_PL, i18next is naturally reaching for the "_few" key, but it currently isn't present, causing that string to fall back to English.

wescopeland: The script only adds them in the locales where they're needed, such as the pl_PL file:

Screenshot 2024-11-15 at 7 21 57 AM

i18next is smart enough to know based on the locale code what plural forms are expected of the locale. This is why the file rename before & after the sync script is necessary, so the locale detection can work correctly.

PR #2848, opened 2024-11-15
fix(home): make some minor css adjustments

wescopeland:
This PR makes some minor adjustments to `/demo/home` to resolve some issues I'm seeing in prod.

Side note: this strategy of running a demo page in parallel with the legacy PHP page is actually working quite well. For the other "big traffic" pages (game info, user profile, achievement info), we should probably follow the same routine so issues can be caught before shifting everyone over.

**Fix 1**
There are multiple issues with the Currently Online chart in XS. Note that the legend fix only works if an actual mobile user agent is detected.

Screenshot 2024-11-15 at 5 51 20 PM

**Fix 2**
Game awards need to clamp the game title if it's excessively long, otherwise the container overflows.

Screenshot 2024-11-15 at 5 52 34 PM

**Fix 3**
I've made a very subtle adjustment to user avatars on the "Just Released" and "Latest Sets in Progress" table rows. Their widths and heights have been reduced by 4px.


Jamiras:
> I've made a very subtle adjustment to user avatars on the "Just Released" and "Latest Sets in Progress" table rows. Their widths and heights have been reduced by 4px.

Noticed while verifying this: The timestamp on New Claims is wrong in mobile mode.
image
image

Jamiras (approved):

wescopeland:
> Noticed while verifying this: The timestamp on New Claims is wrong in mobile mode.

Good catch. I've fixed this in latest and added a few more test cases to validate the behavior.

PR #2856, opened 2024-11-16
feat: add Active Players back-end actions for `/demo/home`

wescopeland:
This PR adds three new actions:
* `LoadThinActivePlayersListAction`
* `BuildTrendingGamesAction`
* `BuildActivePlayersAction`

These actions are meant to facilitate the back-end logic for the Active Players widget on `/demo/home`. The actions are currently not exposed anywhere. Tests are present to validate their behavior is correct.


app/Community/Actions/BuildActivePlayersAction.php
```diff
@@ -0,0 +1,175 @@
...
+use App\Models\Game;
+use App\Models\User;
+use Illuminate\Pagination\LengthAwarePaginator;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
+
+class BuildActivePlayersAction
+{
+    public function execute(
+        int $perPage = 300,
```
wescopeland, line 18: The current component in prod initially shows 20 results, then loads all N after the user scrolls.

That behavior isn't really great from a UI performance perspective. The browser freezes on low-power devices because it's a ton of nodes being inserted simultaneously into the DOM.

The new component will either infinite scroll or virtual scroll - I haven't decided yet. This action will support either approach.

app/Community/Actions/LoadThinActivePlayersListAction.php
```diff
@@ -0,0 +1,61 @@
...
+     *   game_id: int,
+     *   username: string,
+     *   display_name: string,
+     *   rich_presence: string,
+     *   game_title: string,
+     * }>
+     */
+    public function execute(): array
+    {
+        return Cache::flexible('all-active-players', [20, 60], function () {
```
wescopeland, line 26: Data is considered fresh 20 seconds after it's fetched.

After 20 seconds of the data being fresh, but sooner than 60 seconds, the server will retry to update the data in the background.

If no requests come in after 60 seconds, the data is considered stale and the cache will be purged.

With this approach, realistically on the home page there will never be an unlucky user who gets a stale cache hit.

app/Data/UserData.php
```diff
@@ -19,62 +19,51 @@ class UserData extends Data
     public function __construct(
         public string $displayName,
         public string $avatarUrl,
-        public bool $isMuted,
```
wescopeland, line 22: Set `$isMuted` to lazy and gave defaults to all of these values.

app/Community/Actions/BuildTrendingGamesAction.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * come up with some kind of "trending score" that would combine various facets, such as:
+ * - boost for recent player sessions
+ * - boost for recent achievement unlocks
+ * - boost if game is gaining new unique players
+ * - boost if people are getting beats/masteries
+ * - decay the score every N days the game has been trending. after it isn't trending anymore, stop the decay.
+ */
+
+// TODO allow filtering by system
```
wescopeland, line 22: I'd like to make this algorithm much more intelligent in the future. Right now isn't a good time for it though - I just want to get this new home page over the finish line.

For now, I'm just throwing my ideas/brainstorming for an improved algorithm in here.

app/Community/Actions/BuildTrendingGamesAction.php
```diff
@@ -0,0 +1,64 @@
...
+        $gameCounts = [];
+        foreach ($players as $player) {
+            $gameCounts[$player['game_id']] = ($gameCounts[$player['game_id']] ?? 0) + 1;
+        }
+
+        // Sort by count in descending order.
+        arsort($gameCounts);
+
+        // Take the top 4 game IDs.
+        $trendingGameIds = array_slice(array_keys($gameCounts), 0, 4);
```
wescopeland, line 44: Note that this is no longer filtering out the most popular games by new players. I think this filtering is redundant given the staying power of the current games that frequently appear in the list.

app/Community/Actions/LoadThinActivePlayersListAction.php
```diff
@@ -0,0 +1,64 @@
...
+        int $minimumPermissions = Permissions::Registered
+    ): array {
+        return Cache::flexible(
+            "all-active-players:{$lookbackMinutes}:{$minimumPermissions}",
+            [20, 60],
+            function () use ($lookbackMinutes, $minimumPermissions) {
+                $timestampCutoff = Carbon::now()->subMinutes($lookbackMinutes);
+
+                $activePlayers = User::select([
+                    'ID as user_id',
```
Jamiras, line 35: I imagine this is copied from somewhere else, but if you're mapping the values already, is the alias still desirable?

app/Community/Actions/LoadThinActivePlayersListAction.php
```diff
@@ -0,0 +1,64 @@
...
+     *   game_title: string,
+     * }>
+     */
+    public function execute(
+        int $lookbackMinutes = 10,
+        int $minimumPermissions = Permissions::Registered
+    ): array {
+        return Cache::flexible(
+            "all-active-players:{$lookbackMinutes}:{$minimumPermissions}",
+            [20, 60],
```
Jamiras, line 30: what does this do?

app/Community/Actions/BuildActivePlayersAction.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    private function filterBySearch(array $players, ?string $search): array
+    {
+        if (!$search) {
+            return $players;
+        }
+
+        $search = Str::lower(trim($search));
+
+        // We should also support logical OR searches, for example:
```
Jamiras, line 104: "should" sounds like a missing feature, but it looks like you're at least trying to support it.

app/Community/Actions/BuildActivePlayersAction.php
```diff
@@ -0,0 +1,175 @@
...
+        return array_filter(
+            $players,
+            function ($player) use ($searchTerms) {
+                // For each search term, check if it exists in any of the fields.
+                foreach ($searchTerms as $term) {
+                    if (
+                        str_contains(Str::lower($player['username']), $term)
+                        || str_contains(Str::lower($player['display_name']), $term)
+                        || str_contains(Str::lower($player['rich_presence']), $term)
+                        || str_contains(Str::lower($player['game_title']), $term)
```
Jamiras, line 126: I can't help but wonder if there's a way to keep the lowercased values between iterations of the `$searchTerms` loop.

Google suggests using `mb_stripos` for case-insensitive matching, but it's probably a lot more expensive than normalizing to lowercase and doing a string compare.

tests/Feature/Community/Actions/BuildTrendingGamesActionTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function testItReturnsTrendingGames(): void
+    {
+        // Arrange
+        $system = System::factory()->create();
+        $game1 = Game::factory()->create(['ConsoleID' => $system->id, 'Title' => 'most_popular']);
+        $game2 = Game::factory()->create(['ConsoleID' => $system->id, 'Title' => 'second_most_popular']);
+        $game3 = Game::factory()->create(['ConsoleID' => $system->id, 'Title' => 'third_most_popular']);
+        $game4 = Game::factory()->create(['ConsoleID' => $system->id, 'Title' => 'fourth_most_popular']);
+        $game5 = Game::factory()->create(['ConsoleID' => $system->id, 'Title' => 'fifth_most_popular']);
```
Jamiras, line 37: Recommend inserting these out-of-order to also verify the sorting.

app/Community/Actions/LoadThinActivePlayersListAction.php
wescopeland: https://github.com/RetroAchievements/RAWeb/pull/2856#discussion_r1845242069

> Data is considered fresh 20 seconds after it's fetched.
>
> After 20 seconds of the data being fresh, but sooner than 60 seconds, the server will retry to update the data in the background.
>
> If no requests come in after 60 seconds, the data is considered stale and the cache will be purged.
>
> With this approach, realistically on the home page there will never be an unlucky user who gets a stale cache hit.

wescopeland: Documentation reference: https://laravel.com/docs/11.x/cache#swr

app/Community/Actions/LoadThinActivePlayersListAction.php
wescopeland: Good point - removed `as user_id` in latest.

app/Community/Actions/BuildActivePlayersAction.php
wescopeland: Yep, there was a TODO in front of this comment at one point. Updated in latest to remove the word "should".

app/Community/Actions/BuildActivePlayersAction.php
wescopeland: This was mostly copied from `ActivePlayersService` (which is currently using `stripos()`), but I agree it can be heavily optimized.

In latest, I concatenate all the searchable fields into a single string and then lowercase it one time. Then I do `str_contains($searchableText, $term)`.

app/Data/UserData.php
```diff
@@ -19,62 +19,51 @@ class UserData extends Data
...
-        public Lazy|int|null $websitePrefs,
+        public Lazy|array|null $roles = [],
     ) {
     }
 
-    // TODO remove this
+    /**
+     * @see ActivePlayersService.php
+     */
+    public static function fromActivePlayers(array $activePlayerRecord): self
```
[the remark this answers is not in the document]
wescopeland: Should be deleted altogether, this is leftover dead code I ended up not needing. Removed in latest.

tests/Feature/Community/Actions/BuildTrendingGamesActionTest.php
wescopeland: Updated in latest.

app/Community/Actions/BuildActivePlayersAction.php
```diff
@@ -101,37 +109,44 @@ private function filterBySearch(array $players, ?string $search): array
...
-                    }
+        // This lookup function checks all fields at once.
+        $checkPlayer = function (array $player) use ($searchTerms): bool {
+            // Combine all searchable fields into one string and lowercase it.
+            $searchableText = Str::lower(
+                $player['username'] . ' ' .
+                $player['display_name'] . ' ' .
+                $player['rich_presence'] . ' ' .
+                $player['game_title'] . ' '
+            );
```
Jamiras, line 138: Recommend separating fields with `|`, so you don't accidentally match "display_name rich_presence"

wescopeland: Updated in latest.

PR #2857, opened 2024-11-17
feat(multiset): send sets to emulators

wescopeland:
Relies on https://github.com/RetroAchievements/rcheevos/pull/384.

---

This PR modifies `dorequest.php?r=patch` to send resolved subsets to emulators to enable subset+core functionality.

How it works:
* `GetPatchData()` has been removed. The tests are still intact and unchanged.
* The "patch" case in _dorequest.php_ has been modified to execute a new action: `BuildClientPatchDataAction`.
* If this new action receives only a game ID (legacy client), the existing legacy response is returned.
* If this new action receives a game hash (updated client), then `ResolveAchievementSetsAction` is executed.
* A new field: "Sets" is added to the response. It is not added for legacy clients.

The shape of `Sets` is:
```json
{
    "GameAchievementSetID": 1234,
    "CoreGameID": 1234,
    "Title": "Gold Medals",
    "Type": "bonus",
    "ImageIcon": "12345",
    "ImageIconURL": "foo",
    "Achievements": [ ... ],
    "Leaderboards": [ ... ]
}
```

In other words, a multiset patch response will look like:
```json
{
    "Success": true,
    "PatchData": {
        "ID": 1,
        "Title": "Sonic the Hedgehog",
        "ImageIcon": "12345",
        "RichPresencePatch": "foo",
        "ConsoleID": 1,
        "ImageIconURL": "foo",
        "Achievements": [ ... ],
        "Leaderboards": [ ... ],
        "Sets": [ ... ]
    }
}
```

Tests for `BuildClientPatchDataAction` to verify its behavior have been included in this changeset.


wescopeland:
This is now ready for review. I've merged in latest and resolved all conflicts from #2862.

app/Connect/Actions/BuildClientPatchDataAction.php
```diff
@@ -0,0 +1,247 @@
...
+use App\Models\GameHash;
+use App\Models\PlayerGame;
+use App\Models\User;
+use App\Platform\Enums\AchievementFlag;
+use Illuminate\Database\Eloquent\Collection;
+use InvalidArgumentException;
+
+// TODO more tests
+
+// OPEN QUESTION when given a GameHash, should the root Achievements & Leaderboards be returned as `null`?
```
wescopeland, line 17: This is a discussion point I'm not sure how to best handle. I also referenced this in a test:

https://github.com/RetroAchievements/RAWeb/blob/75f2b68d338d3cd43d56599719e4a04e706280bb/tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php#L402-L407

Given that the user is using a modern client and we send a list of sets to that client from `r=patch`, what should the root-level data be?
* Do root-level `Achievements` and `Leaderboards` even matter?
* Should root-level `Title`, `ImageIcon`, and `ImageIconURL` always point to the core game?
* Should `RichPresencePatch` take on the RP script of the "subset game"? 

For reference, here is a full data structure of the response in the test case above:
```php
array:2 [
  "Success" => true
  "PatchData" => array:9 [
    "ID" => 2
    "Title" => "Dragon Quest III [Subset - Bonus]"
    "ImageIcon" => "/Images/000011.png"
    "RichPresencePatch" => """
      Display:\n
      Test
      """
    "ConsoleID" => 1
    "ImageIconURL" => "/media/Images/000011.png"
    "Achievements" => array:2 [ ... two achievements ... ]
    "Leaderboards" => []
    "Sets" => array:2 [
      0 => array:8 [
        "GameAchievementSetID" => 1
        "CoreGameID" => 1
        "Title" => null
        "Type" => "core"
        "ImageIcon" => "/Images/000011.png"
        "ImageIconURL" => "/media/Images/000011.png"
        "Achievements" => array:1 [ ... one achievement ... ]
        "Leaderboards" => []
      ]
      1 => array:8 [
        "GameAchievementSetID" => 5
        "CoreGameID" => 3
        "Title" => "Bonus 2"
        "Type" => "bonus"
        "ImageIcon" => "/Images/000011.png"
        "ImageIconURL" => "/media/Images/000011.png"
        "Achievements" => array:3 [ ... three achievements ... ]
        "Leaderboards" => []
      ]
    ]
  ]
] // tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php:410
```

public/dorequest.php
```diff
@@ -575,7 +578,38 @@ function LogDeprecatedUserAgent(string $requestType, ClientSupportLevel $clientS
...
+
+            // Based on the user's current client support level, we may want to attach
+            // some metadata into the patch response. We'll do that as part of a separate
+            // action to keep the original data construction pure.
+            $response = (new InjectPatchClientSupportLevelDataAction())->execute(
+                $response,
+                $clientSupportLevel,
+                $gameHash,
+                $game,
+            );
```
wescopeland, line 609: This is how I integrated the new stuff from #2862.

First, execute `BuildClientPatchDataAction`, which always should build the correct patch package for the game regardless of the user's client.

Next, call `InjectPatchClientSupportLevelDataAction` on the returned patch data. This is almost like a side-effect, injecting whatever warning achievements and warning metadata we may want to send to a client.

tests/Feature/Connect/Actions/ResolveAchievementSetsActionTest.php
wescopeland, line 1: No functional changes in here, just some cleanup for readability.

app/Connect/Actions/BuildClientPatchDataAction.php
Jamiras: > Should root-level Title, ImageIcon, and ImageIconURL always point to the core game?

I think that if the client is requesting patch data by a subset hash, the top-level data should be from the core set, and the subset as a child record.

> Do root-level Achievements and Leaderboards even matter?

The [response structure](https://github.com/RetroAchievements/rcheevos/wiki/rc_api_process_fetch_game_data_response#parameters) still has root-level fields, it would make sense to return the core data at the root level.

For legacy compatibility, when requesting a subset by game_id, the subset information should be returned at the top level, and no child records should be returned.

If we're going to eliminate the top-level records, maybe we should have a new API (i.e. `patch2`) rather than trying to overload the existing API.

> Should RichPresencePatch take on the RP script of the "subset game"?

I think subsets shouldn't have their own RP scripts. This may need further discussion.

Jamiras: > In other words, let's say I load "Dragon Quest III [Subset - Bonus]". Is it correct that the top-level Title, ImageIcon, ImageIconURL, Achievements, and Leaderboards should be from "Dragon Quest III", not "Dragon Quest III [Subset - Bonus]"?

Yes.

> I can imagine some developers will get frustrated if their RP scripts attached to subsets are suddenly considered invalid.

Are there non-exclusive subsets with unique RP?

Subset:
image

Core:
image


Subset:
image

Core:
image


Subset:
image

Core:
image


Found one. Though I could argue that the core RP is better. Maybe the prefix could be merged into that?

Subset:
image

Core:
image

Jamiras: Perhaps since "specialty" and "exclusive" subsets require loading unique hashes, they could also have unique RP, but generic "bonus" sets would not.

wescopeland: All this logic should now be implemented in latest, with additional tests added to verify it.

* RP patch data from specialty and exclusive subset games is now prioritized when a user loads a hash associated with one of those games.
* If the user is not globally opted out of multiset, root-level data reflects the values of the core/base game.

Jamiras: Thought related to RP: how are we going to maintain/create sessions for the subsets? Will `ping` be overloaded to send all game_ids that are active? Will we just run the active subset logic on the server like we do for building the patch data? 

I thin this comes down to "specialty" and "exclusive" subsets again. Do we still need "active players" for generic bonus subsets?

Do we still need subset sessions in the database for associating unlocks, or can they be tied to the core subset?

tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php
```diff
@@ -421,7 +424,267 @@ public function testLocallyOptedOutOfSubsetsAndLoadedOptedOutSubsetHash(): void
...
+        $result = (new BuildClientPatchDataAction())->execute(gameHash: $specialtyGameHash, user: $user);
+
+        // Assert
+        $this->assertTrue($result['Success']);
+
+        // ... root data should be from the base game ...
+        $this->assertEquals($baseGame->id, $result['PatchData']['ID']);
+        $this->assertEquals($baseGame->title, $result['PatchData']['Title']);
+        $this->assertEquals($baseGame->ImageIcon, $result['PatchData']['ImageIcon']);
+        $this->assertCount(3, $result['PatchData']['Achievements']); // the base game's achievements
```
Jamiras, line 587: I'm not convinced this is right. If I'm playing a specialty set, that's what I'm focused on and the core achievements are "bonus".

tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php
```diff
@@ -421,7 +424,267 @@ public function testLocallyOptedOutOfSubsetsAndLoadedOptedOutSubsetHash(): void
...
+        $result = (new BuildClientPatchDataAction())->execute(gameHash: $exclusiveGameHash, user: $user);
+
+        // Assert
+        $this->assertTrue($result['Success']);
+
+        // ... root data should be from the base game ...
+        $this->assertEquals($baseGame->id, $result['PatchData']['ID']);
+        $this->assertEquals($baseGame->title, $result['PatchData']['Title']);
+        $this->assertEquals($baseGame->ImageIcon, $result['PatchData']['ImageIcon']);
+        $this->assertCount(3, $result['PatchData']['Achievements']); // ... base game's achievements ...
```
Jamiras, line 624: Exclusive sets load without the core achievements, so there's even less reason for the root data to be from the core game.

Jamiras (commented):
I'm still waffling on how to handle "root data" for specialty/exclusive sets. In those cases, the core set is secondary or not even present, so is the player still playing the base game? My gut says the root id and achievements shouldn't be from the base game, but maybe the title and image should?

public/dorequest.php
Jamiras: I wonder if it would make sense to send the warning back as a new field if the input is a hash. There aren't any clients sending hashes yet, so we could structure the warning more appropriately instead of shoehorning in into something that existing clients would understand.

wescopeland: Definitely not opposed to this. I don't want to be presumptuous - what would you prefer in the response in this circumstance?

app/Connect/Actions/BuildClientPatchDataAction.php
wescopeland: My gut feelings are:
* `ping` should never consider more than 1 game ID as being active.
* The active game ID should probably be the same `ID` that's sent back from `patch` at the root level.

We should probably tackle this piece in a subsequent PR.

tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php
wescopeland: In latest, `Title` and `ImageIcon` come from the base game, whereas `ID` and `Achievements` come from the subset.

wescopeland:
I've added more edge case test coverage in latest.

public/dorequest.php
Jamiras: `rc_client` doesn't currently have any way to plumb the warning through, so we could design whatever is appropriate. 

The closest mechanism currently available would be to raise a [SERVER_ERROR](https://github.com/RetroAchievements/rcheevos/wiki/rc_client_set_event_handler#rc_client_event_server_error) event, which would end up displaying something like "patch failed: Outdated Emulator (please update)" in RetroArch. I'm not sure how other clients handle the event.

The more I think about it, the more I like using an achievement to report the warning. In addition to ensuring a high level of compatibility across clients (they'd be much more likely to handle the `ACHIEVEMENT_TRIGGERED` event than the `SERVER_ERROR` event), it keeps the warning around in the achievement list so users can be reminded while they're in the game.

wescopeland: > it keeps the warning around in the achievement list so users can be reminded while they're in the game.

This is a good point. It sounds like it may be best to keep it as-is for now then.

app/Connect/Actions/BuildClientPatchDataAction.php
```diff
@@ -0,0 +1,288 @@
...
+            $coreGameIds = $resolvedSets->pluck('core_game_id')->unique();
+            $games = Game::whereIn('ID', $coreGameIds)->get()->keyBy('ID');
+
+            foreach ($resolvedSets as $resolvedSet) {
+                $setGame = $games[$resolvedSet->core_game_id];
+                $sets[] = [
+                    'GameAchievementSetID' => $resolvedSet->id,
+                    'CoreGameID' => $resolvedSet->core_game_id,
+                    'CoreGameTitle' => $setGame->title,
+                    'SetTitle' => $resolvedSet->title,
```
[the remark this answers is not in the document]
wescopeland: For core sets, the set title will typically be `null`. Check the value for a non-core set.

app/Connect/Actions/BuildClientPatchDataAction.php
```diff
@@ -0,0 +1,288 @@
...
+        if ($resolvedSets?->isNotEmpty()) {
+            $coreGameIds = $resolvedSets->pluck('core_game_id')->unique();
+            $games = Game::whereIn('ID', $coreGameIds)->get()->keyBy('ID');
+
+            foreach ($resolvedSets as $resolvedSet) {
+                $setGame = $games[$resolvedSet->core_game_id];
+                $sets[] = [
+                    'GameAchievementSetID' => $resolvedSet->id,
+                    'CoreGameID' => $resolvedSet->core_game_id,
+                    'CoreGameTitle' => $setGame->title,
```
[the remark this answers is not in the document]
wescopeland: I think with the most recent round of changes, these values are probably redundant at this point. Removed in latest.

app/Connect/Actions/BuildClientPatchDataAction.php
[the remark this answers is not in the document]
Jamiras: Then this isn't working right, or at least not how I expect. When I pass the hash for game 14728, I expect its information to be at the root level (working correctly) and the subset (game 25998) to be in the Sets collection. Instead, I get a duplicate of the core achievements there.
```
curl --header "User-Agent: RAlibretro/1.0" "localhost:64000/dorequest.php?u=Jamiras&t=[redacted]&r=patch&m=108ffe4ba5c2c8c5e0459cb7f0cfac0e"
```
image

There's only one set provided:
image

Which is redundant with the top-level data.

The relationship appears to exist:

image

I would expect the Sets to contain a single item for "Overtime Overload" and its 6 achievements and 4 leaderboards.

app/Connect/Actions/BuildClientPatchDataAction.php
Jamiras: Something else we'll need to address in a later PR: `r=startsession` needs to return unlocks for the subsets as well so the client can disable previously earned subset achievements.

app/Connect/Actions/BuildClientPatchDataAction.php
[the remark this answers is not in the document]
wescopeland: That subset is currently set to `will_be_bonus` in your local DB.

```sql
SELECT * FROM game_achievement_sets WHERE game_id = 14728;
```
If you change it to `bonus`, its assets will appear in the Sets array.

However, that doesn't change that the duplicate root-level data is still within sets. Would you like to see that removed and only exist at the top-level?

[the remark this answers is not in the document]
Jamiras: > That subset is currently set to `will_be_bonus` in your local DB.

Ok, that makes sense.

> Would you like to see that removed and only exist at the top-level?

Yes. It's redundant to have in both places, and I think it makes more sense to keep the primary set at the top level.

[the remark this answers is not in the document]
wescopeland: Sets should now be de-duped with the root-level data in latest.

I am noticing that, while the performance of the action is not dreadful, there appears to be plenty of room for optimization and reduction of SQL queries. I'd also like to kick that further down the road & prioritize this PR towards getting things functionally correct.

Jamiras (approved):

PR #2860, opened 2024-11-19
feat(home): migrate Active Players to React

wescopeland:
This is PR 3 of the React home page migration, and a follow-up to https://github.com/RetroAchievements/RAWeb/pull/2802. These are the planned steps of the migration:

~PR 1 of 4: Build some basic components with hardcoded data. Lots of functionality missing, no back-end.~
~PR 2 of 4: Use real data from the back-end / Laravel controllers rather than hardcoded data.~
**PR 3 of 4: Rebuild the Active Players component.**
PR 4 of 4: Finalize the swap of the PHP home page for the React home page.

This PR sets up the Active Players component on `/demo/home` with minor behavior changes.

### How to test this PR
1. Open _LoadThinActivePlayersListAction.php_.
2. Adjust `$lookbackMinutes` on line 25 from 10 to something like 28000.
3. Visit `/demo/home` and observe that active players and trending games are populated. If you don't see any, keep increasing the `$lookbackMinutes` number.

**Before**
Screenshot 2024-11-19 at 4 29 03 PM

**After**
Screenshot 2024-11-19 at 4 28 33 PM

Note: Player counts in the screenshots above are artificially inflated due to my high lookback time for testing purposes.

There are a few minor behavior changes to note:
* Game titles are now the focal point of each row, and RP is slightly suppressed.
* The "Menu" button now uses a search icon.
* The "last updated" label has been removed. This updates so infrequently I doubt most users even realize it happens automatically if the page is idle.
* The client-side fetch no longer fetches all records. In prod, this slows down the page considerably as this action is immediately followed by inserting thousands of DOM nodes onto the page. The UI now fetches in increments of 300.
* Searching by system is no longer possible.

The UI has 100% code coverage.


resources/js/common/components/ActivePlayerFeed/ActivePlayerFeedList.tsx
```diff
@@ -0,0 +1,49 @@
...
+          </div>
+
+          <div className="flex flex-col">
+            <div className="max-w-fit">
+              <GameAvatar {...player.game} showImage={false} />
+            </div>
+
+            <p
+              className="line-clamp-1 text-2xs text-neutral-600 group-hover:text-neutral-300"
+              style={{ wordBreak: 'break-word' }}
```
Jamiras, line 33: This is too subtle. The focus should be on the rich presence text, not the game name.
image

Just scanning through the list, I pick up on the game titles and completely miss the rich presence.

Jamiras: Blue test stands out a bit more, but the game title is still large and looming
image

I was able to change the color of the game title, but not the size. This is kind of what I would like to see:
image

Maybe you have some ideas on how to make the rich presence more prominent?

wescopeland: > The focus should be on the rich presence text, not the game name.

That is indeed the current design for the Blade component, but I question this as a design choice. There are two reasons why.

1. In the current production component, unless I hover over the game avatar, I can't tell what someone is playing unless I leave the page.
2. A [significant](https://discord.com/channels/310192285306454017/645777658319208448/1253786608461676607) [chunk](https://discord.com/channels/310192285306454017/645777658319208448/1253786704263905280) [of](https://discord.com/channels/310192285306454017/310195377993416714/1128859527555596388) our rich presence is incomprehensible to people who have never played the game before. Unfortunately, I don't see this changing any time soon.

I wouldn't want anyone to feel attacked, but there are some pretty egregious examples I can find by just casually scrolling through the component on prod at any moment of the day.

As a player, I am less interested in the rich presence text and more interested in what people are actually playing. Additionally, at some point in the future I could see us adding more context-sensitive UX to this component, such as the ability to filter down to just people I'm following. In this circumstance, I would also be more interested in the games people are playing - the RP seems like secondary information that is just a nice touch.

With all this in mind:
* I can't help but feel the game title should be the primary focal point.
* I agree that the RP text may be too subdued at the moment - perhaps it should be made a bit louder.

How do you feel about what I've written above? Maybe there is something I am not yet considering.

Jamiras: I completely agree that much of the rich presence is incomprehensible to people not familiar with the games, and having the game title makes it slightly easier to decode the rich presence text.

I could argue that the point of having rich presence in the first place is for the extra detail it provides, but maybe that's best left on the individual game pages where people looking at it will presumably have more context about what it might mean.

Many years ago when there was only a few hundred Active Players, it probably made sense to dump all the information on the main page, but now that there's almost 3000 Active Players in that list, I can see it being too much. Having the game names being the primary focus probably makes sense, but I think letting the rich presence text be the standard color (blue for dark theme) would still be better than what it is now.

Maybe we should ask users who they use the component?

Jamiras: > unless I hover over the game avatar, I can't tell what someone is playing

The same could be said about the user avatar. But visitors are probably more interested in what people are playing moreso than who. We already sort the list so more prominent players are at the top, but just looking at the live list now, I only recognize two of the first seven user avatars.

wescopeland: > Having the game names being the primary focus probably makes sense, but I think letting the rich presence text be the standard color (blue for dark theme) would still be better than what it is now.

Not a problem, I've made this change in latest:

Screenshot 2024-11-30 at 7 44 45 PM

> The same could be said about the user avatar.

I agree. At some point, maybe we should test a different design for these rows altogether and see how people respond.

Jamiras (approved):

PR #2864, opened 2024-11-20
add create event admin tool

Jamiras:
Simple tool to generate an event game with _n_ achievements owned by RAEvents. The existing game and achievement edit pages can be used to flesh out the event after its been created.

### DIFF
--- a/resources/views/livewire/administrative-tools/create-event.blade.php
+++ b/resources/views/livewire/administrative-tools/create-event.blade.php
@@ -1,0 +1,94 @@
+<?php
+
+use App\Models\Achievement;
+use App\Models\EventAchievement;
+use App\Models\Game;
+use App\Models\System;
+use App\Models\User;
+use App\Platform\Actions\ComputeGameSortTitleAction;
+use App\Platform\Enums\AchievementFlag;
+use Filament\Forms;
+use Filament\Forms\Concerns\InteractsWithForms;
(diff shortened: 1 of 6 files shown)


resources/views/livewire/administrative-tools/create-event.blade.php
wescopeland, line 1: This should probably also call `UpsertGameCoreAchievementSetFromLegacyFlagsAction()`, otherwise a `game_achievement_set` for the event game is not created.

resources/views/livewire/administrative-tools/create-event.blade.php
```diff
@@ -0,0 +1,94 @@
...
+        $eventTitle = $this->title ?? 'New Event';
+        $event = Game::Create([
+            'Title' => $eventTitle,
+            'sort_title' => (new ComputeGameSortTitleAction())->execute($eventTitle),
+            'Publisher' => 'RetroAchievements',
+            'ConsoleID' => System::Events,
+        ]);
+
+        for ($i = 0; $i < $this->numberOfAchievements; $i++) {
+            $achievement = Achievement::Create([
```
wescopeland, line 38: I didn't know PHP function names are case-insensitive. That's probably why this still works. I think we should change `Create()` to `create()` though to stick with convention.

wescopeland (approved):

PR #2867, opened 2024-11-21
fix: discard subscribers with empty email addresses

wescopeland:


resources/js/features/home/components/+sidebar/AchievementOfTheWeek/AchievementOfTheWeek.test.tsx
wescopeland, line 1: First CI run failed because "Ends" was randomly included as a word from the auto-generated achievement description. I've updated the affected tests so description is now a static value.

PR #2869, opened 2024-11-22
fix: remediate some minor issues on the Create Ticket page

wescopeland:
This PR remediates two minor issues present in the new Create Ticket page.

1: Unhelpful description validation is a little too aggressive.
https://discord.com/channels/310192285306454017/453242743292952578/1309231489845559296

2: The toggle button text is hard to read on light mode when hovered.
Screenshot 2024-11-21 at 8 00 46 PM

### DIFF
--- a/resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketMainRoot.test.tsx
+++ b/resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketMainRoot.test.tsx
@@ -1026,7 +1026,7 @@ describe('Component: CreateAchievementTicketMainRoot', () => {
   });
 
   it(
-    'given the user writes a perfectly valid description with the word "trigger", does not pop validation on submit',
+    'given the user writes a perfectly valid description with the phrase "not trigger", does not pop validation on submit',
     { timeout: 20_000 },
     async () => {
       // ARRANGE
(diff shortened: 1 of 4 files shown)


resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketMainRoot.test.tsx
```diff
@@ -1028,6 +1025,61 @@ describe('Component: CreateAchievementTicketMainRoot', () => {
     });
   });
 
+  it(
+    'given the user writes a perfectly valid description with the word "trigger", does not pop validation on submit',
```
Jamiras, line 1029: "trigger" was okay. It was "not trigger" that was being detected, and the "not" didn't have to be anywhere near the "trigger" because of the `.*`.

Jamiras (approved):
I'm pretty sure that the old code just showed the warning without preventing the submission to allow for things like what was reported. The minimum length restriction is probably better suited to the job.

resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketMainRoot.test.tsx
wescopeland: Corrected this description in latest.

PR #2870, opened 2024-11-22
fix: update RAPatches regex validation pattern

wescopeland:
A request from @televandalist:
> could you update the RAPatches URL field to allow the /refs/heads/ that GitHub added to file links?

Links such as `https://github.com/RetroAchievements/RAPatches/raw/refs/heads/main/SNES/Hacks/Super%20Mario%20World/31924-SMW-NewNightmareCafe.zip` are now valid.


Jamiras (approved):
I'm not sure how to generate a raw link. When I click on any of the related elements, it just downloads the file, and the "Copy Permalink" item in the menu generates something with `/blob/[commit]/`.

I wonder if it should just be `.../RAPatches/.*/.*\.(zip|7z)$`.

televandalist:
> I'm not sure how to generate a raw link. When I click on any of the related elements, it just downloads the file, and the "Copy Permalink" item in the menu generates something with `/blob/[commit]/`.
> 
> I wonder if it should just be `.../RAPatches/.*/.*\.(zip|7z)$`.

I always go to the patch page and right click on "View raw" to copy the link.

This looks like a link to a patch, but it isn't:  https://github.com/RetroAchievements/RAPatches/blob/main/PlayStation%20Portable/Subset/11115-TrailsintheSky-Subset-NGNightmare.7z 

The URL copied from "View raw" on that page is https://github.com/RetroAchievements/RAPatches/raw/refs/heads/main/PlayStation%20Portable/Subset/11115-TrailsintheSky-Subset-NGNightmare.7z

wescopeland:
> I wonder if it should just be .../RAPatches/.*/.*\.(zip|7z)$.

> This looks like a link to a patch, but it isn't: https://github.com/RetroAchievements/RAPatches/blob/main/PlayStation%20Portable/Subset/11115-TrailsintheSky-Subset-NGNightmare.7z

Given this, it seems like we should still keep the regex pattern more restrictive/as-is.

PR #2878, opened 2024-11-23
fix(manage): sync dev role attach/detach with legacy permissions value

wescopeland:
Resolves an issue where attaching/detaching developer roles from the management app does not synchronize that user's legacy `Permissions` value.

In other words, you can attach a Junior Developer role, but their permissions value won't be automatically updated to 2.

---

After deploy, run:
```sql
UPDATE `auth_roles` SET `name` = 'developer-retired' WHERE `name` = 'developer-veteran';
```


app/Filament/Resources/UserResource/Pages/Roles.php
```diff
@@ -37,16 +43,60 @@ public function table(Table $table): Table
...
+                            $targetUser->removeRole(Role::DEVELOPER_VETERAN);
+
+                            $targetUser->setAttribute('Permissions', Permissions::Developer);
+                            $targetUser->save();
+                        } elseif ($attachedRole->name === Role::DEVELOPER_VETERAN) {
+                            $targetUser->removeRole(Role::DEVELOPER_JUNIOR);
+                            $targetUser->removeRole(Role::DEVELOPER);
+                            $targetUser->removeRole(Role::DEVELOPER_STAFF);
+
+                            $targetUser->setAttribute('Permissions', Permissions::Registered);
```
Jamiras, line 82: What is "Developer Veteran"? Based on the implementation, I recommend renaming it to "Retired Developer" or something similar.

Jamiras (commented):
I have many reservations about this implementation. If a Moderator applies for Jr. Dev, they'll automatically have their Permissions lowered. Similarly, if they give up being a developer, they'll get reset to Registered.

It probably makes sense to keep these changes, but only apply them for users under Moderator permissions level.

wescopeland:
> It probably makes sense to keep these changes, but only apply them for users under Moderator permissions level.

I believe this should now be done in latest. The role attach/detach is still in place for all users as before, but permissions are not reassigned unless `$currentPermissions < Permissions::Moderator`.

app/Filament/Resources/UserResource/Pages/Roles.php
```diff
@@ -37,16 +43,62 @@ public function table(Table $table): Table
...
                     ->preloadRecordSelect()
-                    ->recordSelectOptionsQuery(fn (Builder $query) => $query->whereIn('name', auth()->user()->assignableRoles)),
+                    ->recordSelectOptionsQuery(fn (Builder $query) => $query->whereIn('name', $user->assignableRoles))
+                    ->after(function ($data) {
+                        /** @var User $targetUser */
+                        $targetUser = $this->getRecord();
+
+                        $attachedRole = Role::findById((int) $data['recordId']);
+
+                        $newPermissions = Permissions::Registered;
```
Jamiras, line 56: I believe having this default will strip away Developer permissions if a user is being assigned a secondary role (like Writer)

wescopeland: Should be addressed both here and in the detachment action in latest.

app/Filament/Resources/UserResource/Pages/Roles.php
```diff
@@ -53,6 +53,15 @@ public function table(Table $table): Table
...
                         $attachedRole = Role::findById((int) $data['recordId']);
 
+                        if (!in_array($attachedRole->name, [
+                            Role::DEVELOPER_JUNIOR,
+                            Role::DEVELOPER,
+                            Role::DEVELOPER_STAFF,
+                            Role::DEVELOPER_RETIRED,
+                        ])) {
+                            return;
+                        }
```
Jamiras, line 63: Having this as a secondary gate duplicates the lower logic and creates an opportunity for them to become desync'd. Suggest adding an `else return` to the logic below:
```
else
{
    // nothing to synchronize for non-developer roles. we're done
    return;
}
```

wescopeland: Addressed in latest.

Jamiras: I think you missed my point. You're still checking the conditions twice.

Instead of:
```
if (in_array(A, B)) {
    if (A) {
    } elseif (B) {
    }
    do_stuff();
}
```
just do:
```
if (A) {
} elseif (B) {
} else {
    return;
}
do_stuff();
```

wescopeland: Updated in latest.

Jamiras (approved):

PR #2884, opened 2024-11-25
fix(seeders): add key existence check in ReleaseTablesSeeder

gastongaiduk:
Hello team,

I've been working recently on building the project locally to see if I can collaborate with some features or bug fixes. This small issue was found when following the README steps. Adding an extra check before trying to access the array key fixed the problem. Tested it with sail.

Image

After

image

Best of the lucks!


database/seeders/ReleaseTablesSeeder.php
```diff
@@ -15,11 +15,11 @@ public function run(): void
     {
         $releasesData = getReleasesFromFile() ?? [];
 
-        if (!empty($releasesData['integration']) !== null && IntegrationRelease::count() === 0) {
+        if (array_key_exists('integration', $releasesData) && !empty($releasesData['integration']) !== null && IntegrationRelease::count() === 0) {
```
Jamiras, line 18: The same change needs to be applied to the right. It seems illogical to check both `!empty` and `!== null`, as you've corrected for the front half of the condition.

gastongaiduk: Good catch, I've updated the second part to remove the `!== null` conditional. Thanks for reviewing!

Jamiras (approved):

PR #2885, opened 2024-11-26
feat: add a 'Rows per page' control to the React datatable

wescopeland:
This PR gives users the ability to customize how many rows appear per page on the React datatable.

The default page size remains set to 25. Mobile is still locked to 100 records per page.

This change is a precursor to migrating the system games list to the React datatable.

The new UI code has 100% coverage.

https://github.com/user-attachments/assets/b257e174-8fb4-4c58-9440-19ea7817428e


resources/js/features/game-list/components/DataTablePagination/DataTablePagination.tsx
```diff
@@ -68,28 +74,82 @@ export default function DataTablePagination<TData>({
...
+            <BaseSelectTrigger
+              id="rows-per-page-select"
+              aria-labelledby="rows-per-page-label"
+              className="h-8 w-[70px]"
+            >
+              <BaseSelectValue placeholder={table.getState().pagination.pageSize} />
+            </BaseSelectTrigger>
+
+            <BaseSelectContent side="top">
+              {[25, 50, 150].map((pageSize) => (
```
Jamiras, line 128: Would it be reasonable to have a smaller value (like 10) to eliminate scrolling entirely?

wescopeland: Having a smaller value like 10 sounds like a good idea.

150 and not 100 is a fair question. In latest, I've updated the possible page sizes to:
```
private const VALID_PAGE_SIZES = [10, 25, 50, 100, 200];
```

app/Community/Controllers/Api/UserGameListApiController.php
```diff
@@ -23,7 +23,7 @@ public function index(UserGameListRequest $request): JsonResponse
             page: $request->getPage(),
             filters: $request->getFilters(),
             sort: $request->getSort(),
-            perPage: (new Agent())->isMobile() ? 100 : 25,
+            perPage: (new Agent())->isMobile() ? 100 : $request->getPageSize(),
```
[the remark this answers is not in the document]
wescopeland: We probably shouldn't allow that many. I've adjusted the limits in latest and updated the request to validate against the possible page sizes.

Jamiras (approved):

PR #2887, opened 2024-11-27
feat(api): add API_GetUsersIFollow and API_GetUsersFollowingMe

gastongaiduk:
Two API endpoints to list authenticated user followed and follower users.

```
/*
 *  API_GetUsersFollowingMe - returns list of the caller's follower users
 *  int         Count                       number of user records returned in the response
 *  int         Total                       number of user records the user actually has overall
 *  array       Results
 *   object      [value]
 *    string     User                       username
 *    int        Points                     number of hardcore points the user has earned
 *    int        PointsSoftcore             number of softcore points the user has earned
 *    boolean    AmIFollowing               whether the caller user follows the follower user back
 */
```

```
/*
 *  API_GetUsersIFollow - returns list of the caller's followed users
 *  int         Count                       number of user records returned in the response
 *  int         Total                       number of user records the user actually has overall
 *  array       Results
 *   object      [value]
 *    string     User                       username
 *    int        Points                     number of hardcore points the user has earned
 *    int        PointsSoftcore             number of softcore points the user has earned
 *    boolean    IsFollowingMe              whether the followed user follows the caller user back
 */
```


public/API/API_GetFollowedUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+    ->take($count)
+    ->get()
+    ->map(function ($followedUser) use ($user) {
+        return [
+            'Friend' => $followedUser->User,
+            'Points' => $followedUser->points,
+            'PointsSoftcore' => $followedUser->points_softcore,
+            'LastSeen' => empty($followedUser->RichPresenceMsg) ? 'Unknown' : strip_tags($followedUser->RichPresenceMsg),
+            'FollowsBack' => $followedUser->isFollowing($user),
+            'ID' => $followedUser->id,
```
wescopeland, line 52: Let's not expose user ID values.

public/API/API_GetFollowedUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+
+$usersList = $user->followedUsers()
+    ->whereNull('Deleted')
+    ->orderBy('LastActivityID', 'DESC')
+    ->skip($offset)
+    ->take($count)
+    ->get()
+    ->map(function ($followedUser) use ($user) {
+        return [
+            'Friend' => $followedUser->User,
```
wescopeland, line 47: Let's use `display_name` here.

public/API/API_GetFollowedUsersList.php
wescopeland, line 1: The documentation of this endpoint needs to be more clear that it's the caller's followed users.

public/API/API_GetFollowedUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+            'Points' => $followedUser->points,
+            'PointsSoftcore' => $followedUser->points_softcore,
+            'LastSeen' => empty($followedUser->RichPresenceMsg) ? 'Unknown' : strip_tags($followedUser->RichPresenceMsg),
```
wescopeland, line 50: I think this should either include metadata about the game the RP message is about or be removed.

public/API/API_GetFollowerUsersList.php
wescopeland, line 1: The documentation of this endpoint needs to be more clear that it's the caller's follower users.

public/API/API_GetFollowerUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+    ->take($count)
+    ->get()
+    ->map(function ($followerUser) use ($user) {
+        return [
+            'Friend' => $followerUser->User,
+            'Points' => $followerUser->points,
+            'PointsSoftcore' => $followerUser->points_softcore,
+            'LastSeen' => empty($followerUser->RichPresenceMsg) ? 'Unknown' : strip_tags($followerUser->RichPresenceMsg),
+            'FollowingBack' => $user->isFollowing($followerUser),
+            'ID' => $followerUser->id,
```
wescopeland, line 52: This should be removed.

public/API/API_GetFollowerUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+    ->orderBy('LastActivityID', 'DESC')
+    ->skip($offset)
+    ->take($count)
+    ->get()
+    ->map(function ($followerUser) use ($user) {
+        return [
+            'Friend' => $followerUser->User,
+            'Points' => $followerUser->points,
+            'PointsSoftcore' => $followerUser->points_softcore,
+            'LastSeen' => empty($followerUser->RichPresenceMsg) ? 'Unknown' : strip_tags($followerUser->RichPresenceMsg),
```
wescopeland, line 50: Similarly to the other endpoint, metadata should be included in the response about the game or this should be removed.

wescopeland (commented):
Per our contributing guide, we cannot accept API changes without an accompanying docs PR.

https://github.com/RetroAchievements/RAWeb/blob/master/docs/CONTRIBUTING.md#requirements

public/API/API_GetFollowedUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+use App\Models\User;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'c' => 'nullable|integer|min:0',
+    'o' => 'nullable|integer|min:1|max:500',
```
wescopeland, line 26: `c` doesn't have a max, but `o` does.

public/API/API_GetFollowerUsersList.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+use App\Models\User;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Validator;
+
+$input = Validator::validate(Arr::wrap(request()->query()), [
+    'c' => 'nullable|integer|min:0',
+    'o' => 'nullable|integer|min:1|max:500',
```
wescopeland, line 26: `c` doesn't have a max, but `o` does.

tests/Feature/Api/V1/GetFollowerUsersListTest.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    protected function apiUrl(string $method, array $params = []): string
+    {
+        $params = array_merge(['y' => $this->user->APIKey], $params);
+
+        return sprintf('API/API_%s.php?%s', $method, http_build_query($params));
+    }
+
+    public function testGetFollowedUsersList(): void
```
wescopeland, line 36: This is incorrectly named.

public/API/API_GetFollowedUsersList.php
gastongaiduk: Definitively something that can be fetched from other places when needed, I'll remove it.

public/API/API_GetFollowedUsersList.php
gastongaiduk: Added!

public/API/API_GetFollowerUsersList.php
gastongaiduk: Added!

tests/Feature/Api/V1/GetFollowerUsersListTest.php
gastongaiduk: Good catch!

wescopeland:
Just curious, building anything cool?

gastongaiduk:
> Just curious, building anything cool?

Not yet haha. Aiming mainly to practice Kotlin Multiplatform and thought RA would be a nice way to start putting my hands on to something real. I was looking to create some sort of leaderboard comparison APP to be used between friends, nothing too crazy :D

wescopeland:
Ah, I see! I'm not sure if you know, but we have an official Kotlin client library: https://github.com/RetroAchievements/api-kotlin

I'm always happy to see what people build with our APIs. Given that, I understand the desire to build these endpoints.

wescopeland:
This may be kinda different, but for the sake of my own efficiency, I do like to be the one to mark my feedback as resolved. This helps me know what's outstanding or what still needs further discussion :-)

public/API/API_GetFollowedUsersList.php
wescopeland, line 1: I think there's still a problem with the query parameters on both endpoints.

This may seem hard to believe, but in the production DB, there's a user who follows over 6000 people 😅 

`c` currently has no max on either endpoint, so in theory that user could receive thousands of results per request.

wescopeland, line 1: Now that the implementation details of these endpoints are beginning to solidify, I'll start focusing more on their developer experience.

The main thing I'm having trouble getting past right now is the names of the endpoints. Because our other endpoints do allow consumers to specifically target individual users, I'm still anxious that we may have an ambiguous design by these not allowing that - though I do certainly agree that it should only return data for the caller's context (like they're currently doing).

I think maybe we should rename them. I have a few ideas, but I'm not really sure what's best:

**Idea 1**
API_GetMyFollowing
API_GetMyFollowers

**Idea 2**
API_GetUsersIFollow
API_GetUsersFollowingMe

**Idea 3**
API_GetMyOutgoingFollows
API_GetMyIncomingFollows

I'm definitely open to other ideas as well. I think these names more closely align with the design of the endpoints - they directly tell a consumer before that consumer even reads documentation that it's based on the consumers account context and nothing else.

public/API/API_GetFollowedUsersList.php
```diff
@@ -0,0 +1,56 @@
...
+    ->orderBy('LastActivityID', 'DESC')
+    ->skip($offset)
+    ->take($count)
+    ->get()
+    ->map(function ($followedUser) use ($user) {
+        return [
+            'User' => $followedUser->display_name,
+            'Points' => $followedUser->points,
+            'PointsSoftcore' => $followedUser->points_softcore,
+            'FollowsBack' => $followedUser->isFollowing($user),
```
wescopeland, line 48: There's an N+1 query problem here. If the endpoint returns 100 results, it's going to initiate 100+ database calls.

public/API/API_GetFollowerUsersList.php
```diff
@@ -0,0 +1,56 @@
...
+    ->orderBy('LastActivityID', 'DESC')
+    ->skip($offset)
+    ->take($count)
+    ->get()
+    ->map(function ($followerUser) use ($user) {
+        return [
+            'User' => $followerUser->display_name,
+            'Points' => $followerUser->points,
+            'PointsSoftcore' => $followerUser->points_softcore,
+            'FollowingBack' => $user->isFollowing($followerUser),
```
wescopeland, line 48: There's an N+1 query problem here. If the endpoint returns 100 results, it's going to initiate 100+ database calls.

gastongaiduk:
> This may be kinda different, but for the sake of my own efficiency, I do like to be the one to mark my feedback as resolved. This helps me know what's outstanding or what still needs further discussion :-)

Of course, sorry about that, old habits. No worries I'll leave them unresolved from now on :)

public/API/API_GetFollowedUsersList.php
gastongaiduk: Came up with a solution, might not be the most elegant but it does do the job of avoiding the N+1 queries to fetch this data. With the new implementation it's just two queries, one for the total and another one for the filtered list.

Looking forward to your thoughts :)

public/API/API_GetFollowerUsersList.php
gastongaiduk: Came up with a solution, might not be the most elegant but it does do the job of avoiding the N+1 queries to fetch this data. With the new implementation it's just two queries, one for the total and another one for the filtered list.

Looking forward to your thoughts :)

public/API/API_GetFollowedUsersList.php
gastongaiduk: 6000 😮

Indeed, did the other way around and added _max_ to the wrong parameter. Fixed this and added a couple of extra tests 🙌

public/API/API_GetFollowedUsersList.php
gastongaiduk: Hey, thanks for the suggestion, did tried several ways but unfortunately none of them kept the logic working without actually doing a raw SQL. Feel free to guide me a bit more if you have any idea.

I did removed the shareable code from `ActsAsCommunityMember` tho.

wescopeland: No worries. How about something like this?:
```php
$usersList = $user
  ->followedUsers()
  ->whereNull("Deleted")
  ->with([
    "relatedUsers" => fn($q) => $q
      ->select("UserAccounts.ID", "related_user_id")
      ->where("related_user_id", $user->id)
      ->withPivot("Friendship")
  ])
  ->orderBy("LastActivityID", "DESC")
  ->skip($offset)
  ->take($count)
  ->get()
  ->map(
    fn($followedUser) => [
      "User" => $followedUser->display_name,
      "Points" => $followedUser->points,
      "PointsSoftcore" => $followedUser->points_softcore,
      "FollowsBack" =>
        $followedUser->relatedUsers->first()?->pivot?->Friendship ===
        UserRelationship::Following
    ]
  );
```

This cuts the query count down to 3, which is much more palatable, though I have not checked the data for accuracy :)

gastongaiduk: This last suggestion looks interesting, I can give it a try if you'd like, but perhaps is not a bad idea to use the DB::raw solution which keeps the total query count on 2.

A bit more lines, but fetches the data in a single query.
```
$userTable = $user->getTable();
$friendTable = (new UserRelation())->getTable();
$friendshipStatus = UserRelationship::Following;

$usersList = $user->followerUsers()
    ->select([
        "$userTable.*",
        DB::raw("EXISTS (
            SELECT 1
            FROM $friendTable AS ur
            WHERE ur.user_id = $friendTable.related_user_id
            AND ur.related_user_id = $friendTable.user_id
            AND ur.Friendship = $friendshipStatus
        ) AS am_i_following_back"),
        "$friendTable.user_id as pivot_user_id",
        "$friendTable.related_user_id as pivot_related_user_id",
        "$friendTable.Friendship as pivot_Friendship",
    ])
    ->whereNull('Deleted')
    ->orderBy('LastActivityID', 'DESC')
    ->skip($offset)
    ->take($count)
    ->get()
    ->map(function ($followerUser) {
        return [
            'User' => $followerUser->display_name,
            'Points' => $followerUser->points,
            'PointsSoftcore' => $followerUser->points_softcore,
            'AmIFollowing' => filter_var($followerUser->am_i_following_back, FILTER_VALIDATE_BOOLEAN),
        ];
    });
```

What do you think?

wescopeland: I lean towards the query proposed in https://github.com/RetroAchievements/RAWeb/pull/2887#discussion_r1864254157 :-)

The reasons why are:
* Using Eloquent relationships is a bit more idiomatic with respect to Laravel than reaching for a `DB::raw()`.
* Our database is still frequently changing. Using `DB::raw()` is great for optimizing performance in a maximal way (which you've done!), but in this case it may be a premature optimization. When underlying table structures change, ideally these queries will continue working because they're using Eloquent's fluent syntax.

gastongaiduk: That really does make sense, thanks for taking the time to share your point. I have tested your code snippet and worked perfectly, now code-wisely looks much prettier :)

public/API/API_GetUsersFollowingMe.php
```diff
@@ -0,0 +1,65 @@
...
+$usersList = $user
+  ->followerUsers()
+  ->whereNull("Deleted")
+  ->with([
+      "inverseRelatedUsers" => fn ($q) => $q
+        ->select(sprintf("%s.ID", $user->getTable()), "related_user_id")
+        ->where("user_id", $user->id)
+        ->withPivot("Friendship"),
+  ])
+  ->orderBy("LastActivityID", "DESC")
```
wescopeland, line 47:
```suggestion
  ->orderByDesc("LastActivityID")
```

nit: `->orderByDesc()` is even more idiomatic. These two endpoints would be the only place we have `->orderBy("Foo", "DESC")` (desc in all-caps).

public/API/API_GetUsersIFollow.php
```diff
@@ -0,0 +1,65 @@
...
+$usersList = $user
+  ->followedUsers()
+  ->whereNull("Deleted")
+  ->with([
+      "relatedUsers" => fn ($q) => $q
+        ->select(sprintf("%s.ID", $user->getTable()), "related_user_id")
+        ->where("related_user_id", $user->id)
+        ->withPivot("Friendship"),
+  ])
+  ->orderBy("LastActivityID", "DESC")
```
wescopeland, line 47:
```suggestion
  ->orderByDesc("LastActivityID")
```

nit: `->orderByDesc()` is even more idiomatic. These two endpoints would be the only place we have `->orderBy("Foo", "DESC")` (desc in all-caps).

wescopeland (approved):

public/API/API_GetUsersIFollow.php
gastongaiduk: Thank you for the suggestion!

public/API/API_GetUsersFollowingMe.php
gastongaiduk: Thank you for the suggestion!

PR #2889, opened 2024-11-29
ci: add crowdin-download.yml

wescopeland:
This PR adds a CI workflow which utilizes the new `pnpm crowdin:sync` script.

If there are translation changes, a PR will automatically be opened every Wednesday to merge them into the project. The workflow can also be manually initiated from GitHub's UI (after it is in the `master` branch).

I've already set up `CROWDIN_PERSONAL_TOKEN` as a secret which can be accessed by GitHub Actions. Additionally, GitHub Actions in this repo are already configured to have the ability to open pull requests.

Unfortunately, there probably isn't a great way to test this unless it's merged and initiated from this repo.

The action is named _crowdin-download.yml_, as it's likely we'll also need a _crowdin-upload.yml_ which pushes _en_US.json_ changes from the repo to Crowdin.


wescopeland:
To prevent confusion with https://github.com/RetroAchievements/RAWeb/pull/2893, I've renamed `crowdin-sync` to `crowdin-download`.

.github/workflows/crowdin-download.yml
```diff
@@ -0,0 +1,56 @@
+name: Create Automated Crowdin Translations PR
+
+on:
+  schedule:
+    - cron: '0 0 * * 3' # 00:00 UTC on Wednesday
+  workflow_dispatch: # Allow the workflow to be manually dispatched.
```
Jamiras, line 6: This hasn't run against the PR yet. Should we wait until Wednesday to validate it before merging?

Jamiras: I usually test github actions by creating a PR. Will it not run because it's scheduled?

Jamiras (approved):

.github/workflows/crowdin-download.yml
wescopeland: Yep, this is a scheduled/manual pipeline, so it won't execute when this PR changes.

PR #2891, opened 2024-11-30
allow inactive emulators to be selected when creating tickets

Jamiras:
Adds additional items to the emulators drop down based on user agents seen in the player's session history - including an "Other" option if a user agent is not recognized. Inactive emulators are not added to the list unless one is seen in the player history.

image

Side note: Emulator Core should be hidden if the selected emulator is not a multi-emulator. Currently RetroArch, RAlibretro, and Bizhawk, but it would also apply for Firelight and potentially more in the future. It should be exposed through the emulator list as a property on EmulatorData, but I don't know how to toggle the UI visibility based on the selected emulator.


resources/js/features/achievements/components/CreateAchievementTicketMainRoot/CreateAchievementTicketForm/EmulatorSelectField.tsx
```diff
@@ -27,7 +27,9 @@ export const EmulatorSelectField: FC = () => {
 
   const form = useFormContext<CreateAchievementTicketFormValues>();
 
-  const sortedEmulators = emulators.sort((a, b) => a.name.localeCompare(b.name));
+  const sortedEmulators = emulators.sort((a, b) =>
+    a.name.startsWith('Other') ? 1 : b.name.startsWith('Other') ? -1 : a.name.localeCompare(b.name),
+  );
```
wescopeland, line 32: I think we should cover this with a test. Non-trivial `.sort()`s tend to be the kind of thing future maintainers inevitably break.

Testing this is certainly nontrivial. If you create an _EmulatorSelectField.test.tsx_, this should work:
```tsx
import userEvent from '@testing-library/user-event';
import type { FC, ReactNode } from 'react';
import { FormProvider, useForm } from 'react-hook-form';

import { render, screen } from '@/test';
import { createEmulator } from '@/test/factories';

import { EmulatorSelectField } from './EmulatorSelectField';

const Wrapper: FC<{ children: ReactNode }> = ({ children }) => {
  const form = useForm();

  return <FormProvider {...form}>{children}</FormProvider>;
};

describe('Component: EmulatorSelectField', () => {
  beforeEach(() => {
    window.HTMLElement.prototype.hasPointerCapture = vi.fn();
    window.HTMLElement.prototype.scrollIntoView = vi.fn();
  });

  it('renders without crashing', () => {
    // ARRANGE
    const { container } = render(
      <Wrapper>
        <EmulatorSelectField />
      </Wrapper>,
      { pageProps: { emulators: [] } },
    );

    // ASSERT
    expect(container).toBeTruthy();
  });

  it('sorts emulators alphabetically with "Other" emulators at the end', async () => {
    // ARRANGE
    const emulators = [
      createEmulator({ name: 'ZSNES' }),
      createEmulator({ name: 'Other - Custom' }),
      createEmulator({ name: 'Dolphin' }),
      createEmulator({ name: 'Other - Test' }),
    ];

    render(
      <Wrapper>
        <EmulatorSelectField />
      </Wrapper>,
      { pageProps: { emulators } },
    );

    // ACT
    await userEvent.click(screen.getByRole('combobox'));

    // ASSERT
    const options = screen.getAllByRole('option');
    const optionTexts = options.map((option) => option.textContent);

    expect(optionTexts).toEqual(['Dolphin', 'ZSNES', 'Other - Custom', 'Other - Test']);
  });
});
```

wescopeland (approved):

PR #2894, opened 2024-12-01
build(i18n): use type-safe `t()` function

wescopeland:
This PR enforces type safety on the `t()` function.

When calling `t()`, you must use a key that exists in _en_US.json_, otherwise a type error will be thrown. This ensures we don't have any orphaned untranslatable strings in the front-end codebase.

An added benefit is IntelliSense now also works for `t()`. Your IDE should show you a list of all keys available.

Additionally, props prefixed with `t_` are now required to return values from `t()`.


resources/js/types/i18next.d.ts
wescopeland, line 1: This is where the type rules are defined.

lang/en_US.json
```diff
@@ -458,5 +458,8 @@
     "Hash Comments": "Hash Comments",
     "Surprise me": "Surprise me",
     "Claim Comments - {{gameTitle}}": "Claim Comments - {{gameTitle}}",
-    "Claim Comments": "Claim Comments"
+    "Claim Comments": "Claim Comments",
+    "Rank": "Rank",
+    "User": "User",
+    "Top Achievers - {{gameTitle}}": "Top Achievers - {{gameTitle}}"
```
Jamiras, line 464: Have we given up trying to keep this alphabetized?

Jamiras (approved):

lang/en_US.json
wescopeland: I think it's ok for us to do it, but if we do, I'd like to see it done automatically by ESLint so it doesn't become one more thing to remember.

I know of at least one ESLint rule that can probably do this for us. Let's push to a subsequent PR as this will likely contain a lot of noise.

PR #2898, opened 2024-12-02
feat: add Progress filter to React datatable

wescopeland:
This PR adds a Progress filter to the React datatable (`/games`, `/game-list/play`). This is another precursor to introducing the datatable to additional contexts, such as systems and hubs.

**Before**
Screenshot 2024-12-02 at 3 33 39 PM

**After**
Screenshot 2024-12-02 at 3 33 13 PM

The table filters now live in a meta space similar to our Blade tables. Now that we have 3 filters, they no longer fit on 1 row in a worst-case scenario. This also gives us more room to expand later.

On mobile, the filter lives in the super filter drawer:
Screenshot 2024-12-02 at 3 35 55 PM

The filter options are based on the user's current context.
* Given the user has a hardcore rank and no softcore rank, they see only hardcore options.
* Given the user has a softcore rank and no hardcore rank, they see only softcore options.
* Given the user has ranks on both modes, they see options for both modes.
* Given the user has no rank, they see only hardcore options.

**Conditional UIs**
Screenshot 2024-12-02 at 3 41 17 PM

The available filter options are based on user feedback from different communities.
* Power users tended to want multiselects and the ability to exclude things.
* Casual users tended to want single selects and easy-to-read labels.
* Both sets of users wanted to not see options unrelated to their playing preference (hardcore / softcore).

After several rounds of iteration, the options in the screenshot above are where I landed as the "least bad" choice that satisfied all the use cases that were desirable. The only existing table progress filter that is not satisfied by these options is the "Has any softcore progress" filter - but this filter is available elsewhere where it's arguably more useful, ie: the Completion Progress page.

If the user is not logged in, the filter is hidden altogether on mobile. On desktop, it is visible albeit disabled:

Screenshot 2024-12-02 at 3 49 49 PM

The back-end and front-end changes both have 100% coverage.


Jamiras (approved):
verified functionality. did not look at code.

PR #2900, opened 2024-12-03
ci(crowdin-upload): fix error

wescopeland:


wescopeland:
I've confirmed after the tweak made in this commit, the action is working and uploading stuff correctly: https://github.com/RetroAchievements/RAWeb/actions/runs/12130923670/job/33822155769

PR #2905, opened 2024-12-04
add ra:community:generate-annual-recap command

Jamiras:
The command can be run against a specific user, or it will queue up the command to be run against all users who have accounts older than Sep 1 and have logged in after Dec 1. The command processes about 250 users per minute on my machine, so it shouldn't be too taxing to run it on production. However, if we adjust the date windows to 30 days prior to today, this would queue up 85k commands, which would take around 6 hours. I'm not sure if batching the requests would help. The queue is constructed so the least recently active users are deprioritized (most recently active users will be processed first). There's still 25k users who have been active in the last 24 hours.

To test:
```
sail artisan ra:community:generate-annual-recap Jamiras
```

Then go look at the results in Mailpit. You can also test single users from the command line by uncommenting the ConsoleOutput line in the Action file.

For purposes of determining the data to examine, it takes the current date and subtracts six months to determine the target year. That means this can be run up to 6 months into a year to run against the previous year. It does have an upper bound of the 1st of the next year, so could be run any time after the start of the new year.


wescopeland (approved):
Played with this a little bit last night. I had a few recommended optimizations, but looks like you've already made several of them :)

PR #2908, opened 2024-12-05
feat: migrate the system game list pages to React

wescopeland:
This PR migrates the system game pages to React.

Screenshot 2024-12-04 at 9 49 26 PM

There are several behavior changes on this particular game list to be aware of:
* The system filter and column are not available.
* By default, the Players column is always shown.
* For the time being, subsets are included in the game list results. This is temporary to achieve functional parity, but I plan changing how this behavior is surfaced in a subsequent PR.
* The default page size is 100.
* URLs are self-healing and SEO-friendly. For example: `/system/4/games` will self-heal to `/system/game-boy-4/games`.


app/Platform/Controllers/SystemController.php
```diff
@@ -34,6 +44,42 @@ public function store(Request $request): void
...
+        /** @var ?User $user */
+        $user = $request->user();
+
+        $isMobile = (new Agent())->isMobile();
+
+        $paginatedData = (new BuildGameListAction())->execute(
+            GameListType::System,
+            targetSystemId: $system->id,
+            user: $user,
+            filters: $request->getFilters(targetSystemId: $system->id),
```
Jamiras, line 60: These filters don't seem to work. If I change them, there's like a 7 second delay, then the grid clears out:
image

It still works on the want to play page. I have tried rebooting my VM.

wescopeland: Definitely a bug here, and the test was validating that the bug was present. Fixed the code & the test in latest.

app/Models/Achievement.php
```diff
@@ -45,18 +45,22 @@ class Achievement extends BaseModel
...
-    use SoftDeletes;
 
+    /*
+     * Behavioral Traits
+     */
     use CausesActivity;
-    use LogsActivity {
-        LogsActivity::activities as auditLog;
-    }
+    use LogsActivity { LogsActivity::activities as auditLog; }
```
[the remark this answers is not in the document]
wescopeland: Reverted in latest.

Jamiras (approved):
Verified functionality. Minimally looked at code.

PR #2911, opened 2024-12-05
fix(manage): dont error when restoring deleted achievement credit

wescopeland:
Resolves https://discord.com/channels/310192285306454017/310195377993416714/1314263872303005750.

To reproduce the bug:
* Add any form of achievement credit for a user.
* Delete the credit.
* Try to re-add the same credit for the user.

This surfaces a SQL error because it's a soft-deleted record that actually needs to be restored.

### DIFF
--- a/app/Filament/Resources/AchievementResource/RelationManagers/AuthorshipCreditsRelationManager.php
+++ b/app/Filament/Resources/AchievementResource/RelationManagers/AuthorshipCreditsRelationManager.php
@@ -58,6 +58,24 @@ class AuthorshipCreditsRelationManager extends RelationManager
                         $achievement = $this->ownerRecord;
 
                         $task = $data['task'];
+                        $userId = (int) $data['user_id'];
+
+                        // First, check for and restore any soft-deleted record.
+                        $existingRecord = AchievementAuthor::withTrashed()
+                            ->whereAchievementId($achievement->id)
+                            ->whereUserId($userId)
+                            ->whereTask($task)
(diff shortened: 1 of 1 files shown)


app/Filament/Resources/AchievementResource/RelationManagers/AuthorshipCreditsRelationManager.php
```diff
@@ -58,6 +58,24 @@ public function table(Table $table): Table
...
+                            ->whereAchievementId($achievement->id)
+                            ->whereUserId($userId)
+                            ->whereTask($task)
+                            ->first();
+
+                        if ($existingRecord) {
+                            if ($existingRecord->trashed()) {
+                                $existingRecord->restore();
+
+                                return $existingRecord;
```
Jamiras, line 74: this return is redundant with the return on line 77.

Jamiras (commented):
Noticed while testing this: Creating a new credit with a specific date creates a record with today's date. It is possible to then edit the record to select the specific date.

app/Filament/Resources/AchievementResource/RelationManagers/AuthorshipCreditsRelationManager.php
wescopeland: Removed in latest.

wescopeland:
> Noticed while testing this: Creating a new credit with a specific date creates a record with today's date.

Good catch. The input date should now be respected in latest.

PR #2920, opened 2024-12-07
fix(game_sets): refresh updated_at when adding/removing games and links

wescopeland:
`game_sets` (hubs) are not currently having their `updated_at` values updated when a game or link is added/removed from the set.

This PR corrects the issue. We'll need to resync `game_sets` using `php artisan ra:sync:game-sets` on deploy.

### DIFF
--- a/app/Models/GameSet.php
+++ b/app/Models/GameSet.php
@@ -26,7 +26,31 @@ class GameSet extends BaseModel
         'user_id',
     ];
 
+    // == constants
+
+    public const CentralHubId = 6591;
+
     // == accessors
+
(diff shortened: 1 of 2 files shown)


app/Models/GameSet.php
wescopeland, line 1: These changes are picked from a separate branch that will ultimately be stacked onto this one (migration from `GameAlternatives` to `game_sets` w/ React datatable). They're in here so as to avoid any potential merge conflicts.

wescopeland:
Rolled into https://github.com/RetroAchievements/RAWeb/pull/2923.

PR #2923, opened 2024-12-08
feat(hub): build dedicated pages for hubs

wescopeland:
This is the first of several PRs to completely extract hubs from games.

**PR 1: Build dedicated and non-exposed `/hub/{gameSet}` pages.** 👈
PR 2: Create management tools for hubs, make the new hub pages generally available.
PR 3: Remove all hubs code from games, stop double writing to hubs, drop `GameAlternatives`.

## How to test this PR
In _.env_:
```
FEATURE_ENABLE_MODERN_HUBS=true
```
Then, navigate to any hub, or just `/hubs`.

## UX Walkthrough

This PR adds dedicated pages for hubs, which have been double writing to `game_sets` for a few months. `GameData` and `GameAlternatives` are both completely unused as part of this migration.

Screenshot 2024-12-08 at 3 26 42 PM

### Intelligent Breadcrumbs
Screenshot 2024-12-08 at 3 28 27 PM

Screenshot 2024-12-08 at 3 28 46 PM

Screenshot 2024-12-08 at 3 28 56 PM

Screenshot 2024-12-08 at 3 29 12 PM

A new action, `BuildHubBreadcrumbsAction`, attempts to infer the breadcrumbs for the currently-viewed hub by walking up the relationships of the hub. This helps users orient themselves in our labyrinth of hubs.

### React Datatable
Screenshot 2024-12-08 at 3 31 26 PM

The familiar and battle-tested React datatable is used to display the list of games. Note that the default page size is 25 and "Has achievements" defaults to "Both" rather than "Yes".

If a hub has no games and only links attached, the table does not render on the page.

### Related Hubs
Underneath the React datatable is a list of all the related hubs.

Screenshot 2024-12-08 at 3 33 13 PM

### Other Implementation Notes
* In the UI, hub titles are cleaned of their square brackets. "[Central - Meta]" becomes "Central - Meta". We should eventually enforce this at the DB level.
* Breadcrumb creation uses a stale-while-revalidate (`Cache::flexible()`) cache. It is inherently imperfect, as hubs are totally unstructured, but seems to get things "close enough" most of the time. Technical accuracy is not the ultimate goal with the breadcrumbs - the goal is just to help users orient themselves.
* With the feature flag enabled, all "game hub" links will redirect to the corresponding "hub" route. When we make this feature generally available, this logic should also go live for all users/guests/crawlers.


app/Platform/Actions/BuildHubBreadcrumbsAction.php
wescopeland, line 1: This is the action responsible for building the inferred breadcrumbs on the page.

app/Platform/Actions/UpdateGameSetFromGameAlternativesModificationAction.php
wescopeland, line 1: Currently, `game_sets.updated_at` values are not being updated in prod when games/links are attached or detached from a hub. This fixes the issue.

We'll need to resync hubs with `php artisan ra:sync:game-sets` after deploy.

app/Platform/RouteServiceProvider.php
```diff
@@ -50,7 +54,14 @@ public function map(): void
     protected function mapWebRoutes(): void
     {
         Route::middleware(['web', 'csp'])->group(function () {
+            Route::get('hub/' . GameSet::CentralHubId, function () {
+                return Redirect::to('/hubs');
+            });
```
wescopeland, line 59: `/hub/[central-hub-id]` redirects to `/hubs`.

app/Platform/RouteServiceProvider.php
```diff
@@ -64,6 +75,12 @@ protected function mapWebRoutes(): void
 
                 Route::get('games', [GameController::class, 'index'])->name('game.index');
 
+                Route::get('hubs', [HubController::class, 'show'])
+                    ->defaults('gameSet', GameSet::CentralHubId)
+                    ->name('hubs.index');
```
wescopeland, line 80: `/hubs` is just a rewrite/fallthrough to `/hub/{gameSet}`, with the central hub ID hardcoded for its `GameSet`.

resources/js/features/game-list/components/HubMainRoot/HubBreadcrumbs/HubBreadcrumbs.tsx
```diff
@@ -0,0 +1,121 @@
...
+      'Series - ',
+      'Series Hacks - ',
+      'Subgenre - ',
+      'Subseries -',
+    ];
+
+    for (const prefix of alwaysStripPrefixes) {
+      if (cleaned.startsWith(prefix)) {
+        const stripped = cleaned.slice(prefix.length);
+        const firstWord = stripped.split(' - ')[0];
```
wescopeland, line 90: Lots of UI handwaving occurs in this hook/function to get the hub titles into a state that makes sense for breadcrumbs.

resources/js/features/game-list/components/HubMainRoot/HubHeading/HubHeading.tsx
wescopeland, line 1: Lots of TODO / comments in here. Will be followed through in the subsequent PR.

Soulsderfuchs:
Please don't add pagination to hubs, please.

wescopeland:
... You know you can change the default page size to 200, right?

Soulsderfuchs:
That would still cut out games in many hubs.

wescopeland:
This is a working space. I am happy to discuss implementation details about the code.

There are technical reasons why we shouldn't send data to every browser completely unpaginated.

wescopeland:
There is a hub with over 4,700 games. The expectation that we should send all that data unpaginated is unreasonable and unrealistic.

wescopeland:
After chatting a bit privately with @Soulsderfuchs to better understand the root of their concerns, I will:

* Accelerate the development of the custom/persisted views feature.
* Make some minor UX enhancements to sorting by system.

These are context-agnostic, so I will handle these in a separate branch.

resources/js/features/game-list/components/HubMainRoot/HubMainRoot.tsx
```diff
@@ -0,0 +1,78 @@
...
+    columnFilters,
+    columnVisibility,
+    pagination,
+    setColumnFilters,
+    setColumnVisibility,
+    setPagination,
+    setSorting,
+    sorting,
+  } = useGameListState(paginatedGameListEntries, {
+    alwaysShowPlayersTotal: true,
```
Jamiras, line 30: Having this field visible adds an unnecessary horizontal scrollbar to my view:
image

resources/views/components/menu/main.blade.php
```diff
@@ -83,7 +83,6 @@
             <x-dropdown-item :href="route('claims.active')">Sets in Progress</x-dropdown-item>
             <x-dropdown-item :href="route('game.random')">Random Set</x-dropdown-item>
             <x-dropdown-header>Hubs</x-dropdown-header>
-            <x-dropdown-item :href="url('gameList.php?s=6&c=100&f=1')">Hub List</x-dropdown-item>
             <x-dropdown-item :href="url('game/6914')">Central Hub</x-dropdown-item>
             <x-dropdown-item :href="url('game/9553')">Genre & Subgenre Hub</x-dropdown-item>
             <x-dropdown-item :href="url('game/5771')">Series Hub</x-dropdown-item>
```
Jamiras, line 88: Should these links be updated to point directly to the hubs?

app/Platform/Actions/BuildHubBreadcrumbsAction.php
```diff
@@ -0,0 +1,393 @@
...
+ * This action  generates breadcrumb navigation paths for hubs, handling various
+ * special cases and hierarchical relationships. It supports different types of
+ * game sets including themes, difficulty levels, and various hub types.
+ *
+ * - Caches breadcrumb paths with fresh/stale timing.
+ * - Handles special cases for Theme and Difficulty hubs.
+ * - Supports derived hierarchical navigation through hub types.
+ * - Manages hub relationships via predefined mappings.
+ */
+class BuildHubBreadcrumbsAction
```
Jamiras, line 24: Overall, this seems to work really well, but I did find an oddity:
image

I think it should be "All Hubs > Miscellaneous > Virtual Console > New Nintendo 3DS"

Jamiras: And `Publisher` isn't being extracted from:
image

Jamiras: And a couple more oddities:
image
image
image
image

resources/js/features/game-list/components/HubMainRoot/HubMainRoot.tsx
wescopeland: In latest, this field is no longer visible by default.

resources/views/components/menu/main.blade.php
wescopeland: I think probably not in this PR, but yes in the subsequent one.

app/Platform/Actions/BuildHubBreadcrumbsAction.php
wescopeland: I've updated the action's logic in latest to hopefully handle these cases.

I was unable to repro the "Publisher - Absolute Entertainment" case - that seemed to have been working correctly on my local.

resources/js/features/game-list/components/HubMainRoot/HubBreadcrumbs/HubBreadcrumbs.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    // Always strip these organizational prefixes first.
+    const alwaysStripPrefixes = [
+      'ASB -',
+      'Central - ',
+      'Events -',
+      'Genre - ',
+      'Meta - ',
+      'Meta|DevComp - ',
+      'Meta|QA - ',
```
Jamiras, line 76: There's a `Meta|Art - ` prefix too:
image

app/Platform/RouteServiceProvider.php
```diff
@@ -64,6 +75,12 @@ protected function mapWebRoutes(): void
...
+
+                Route::get('hub/{gameSet}', [HubController::class, 'show'])->name('hub.show');
```
Jamiras, line 82: Not sure where the game redirect is happening, so this comment it's well placed.

There seems to be some bad data in the hub mappings.
http://localhost:64000/game/19594
Should be 
image

per the link on http://localhost:64000/game/2809, but it looks like the hub maybe got repurposed?
image

app/Platform/Actions/BuildHubBreadcrumbsAction.php
Jamiras: "Publisher - Absolute Entertainment" is failing for me because the hub isn't linked to "Central - Publisher". That appears to have been done more recently than my dump
image

Which brings up the fact that there's no way to add that association in this state. I'm not sure it's reasonable to merge this without
 >PR 2: Create management tools for hubs
 
 I guess that's what the feature flag is for?

wescopeland: > I guess that's what the feature flag is for?

Yes, while the feature flag is disabled (ie: prod), users should only be able to navigate to these pages by landing on them directly. I agree we can't make this generally available without management tools. I didn't want to squeeze that functionality into this PR because there's already so much stuff in this branch.

wescopeland: > Unwanted - Current Games and Unwanted - Past Games are still parenting to each other. They should parent to Events - Unwanted.

This should now be fixed in latest, though it should be noted I've removed the ⭐️ characters from the hub titles in both my local environment and prod:

Screenshot 2024-12-12 at 3 09 24 PM

resources/js/features/game-list/components/HubMainRoot/HubBreadcrumbs/HubBreadcrumbs.tsx
wescopeland: Fixed in latest. I didn't notice this one because there's no parent for Meta|Art to roll up into (other than Central - Misc).

app/Platform/RouteServiceProvider.php
wescopeland: Bad data indeed. I've made some adjustments to the game model in latest to ensure these hub titles are being written correctly. When we run `php artisan ra:sync:game-sets` on deploy, all the bad data should be fixed.

Jamiras:
All Hubs seems to be missing from all breadcrumbs now.
image
image

wescopeland:
I'm seeing something different on my local 🤔

Screenshot 2024-12-14 at 10 20 03 AM

Screenshot 2024-12-14 at 10 20 38 AM

Do you see any change if you do `sail artisan cache:clear`?

Jamiras:
When I re-synced my hubs, my `GameSet::CentralHubId` changed:
```
mysql> select * from game_sets where id=6591;
Empty set (0.00 sec)

mysql> select * from game_sets where title='[Central]';
+-------+------+-----------+--------------------+---------+------------+---------+---------------------+---------------------+------------+
| id    | type | title     | image_asset_path   | user_id | definition | game_id | created_at          | updated_at          | deleted_at |
+-------+------+-----------+--------------------+---------+------------+---------+---------------------+---------------------+------------+
| 34268 | hub  | [Central] | /Images/052108.png |    NULL | NULL       |    6914 | 2024-12-14 14:59:54 | 2024-12-14 15:06:30 | NULL       |
+-------+------+-----------+--------------------+---------+------------+---------+---------------------+---------------------+------------+
1 row in set (0.01 sec)

```

wescopeland:
> When I re-synced my hubs, my GameSet::CentralHubId changed

I hadn't considered that value needs to probably be dynamic.

In latest, I've removed the `::CentralHubId` constant and replaced it with `scopeCentralHub()`. The scope finds the central hub and caches its ID forever:
```php
public function scopeCentralHub(Builder $query): Builder
{
    $centralHubId = Cache::rememberForever('central_hub_id', function () {
        return $this->whereType(GameSetType::Hub)
            ->where(function ($query) {
                $query->where('title', '[Central]')
                    ->orWhere('title', 'Central');
            })
            ->value('id');
    });

    return $query->whereId($centralHubId);
}
```

`SyncGameSets` has been modified to also include:
```php
Cache::forget('central_hub_id');
```

I've made some additional changes to the back-end and front-end of breadcrumbs to also make their results a bit more predictable and scalable. The front-end now checks for repeated keywords in the breadcrumbs and strips them automatically. The back-end handles events breadcrumbs a little better.

All tests continue to pass, and a few new tests have been added.

app/Platform/Commands/SyncGameSets.php
```diff
@@ -28,11 +28,16 @@ public function handle(): void
...
         // This will be a full reset. Delete any existing game_sets data.
+        // We'll use TRUNCATE to reset the auto-incrementing ID counter back to 1.
         $this->info("\nDeleting any existing game_sets data...");
-        GameSetLink::query()->forceDelete();
-        GameSetGame::query()->forceDelete();
-        GameSet::query()->forceDelete();
+        DB::statement('SET FOREIGN_KEY_CHECKS=0;');
+        DB::statement('TRUNCATE TABLE game_set_links;');
+        DB::statement('TRUNCATE TABLE game_set_games;');
+        DB::statement('TRUNCATE TABLE game_sets;');
```
Jamiras, line 39: You can use `GameSet::truncate()` for this.

app/Models/GameSet.php
```diff
@@ -53,12 +75,40 @@ public function games(): BelongsToMany
...
 
     // == scopes
+
+    /**
+     * @param Builder<GameSet> $query
+     * @return Builder<GameSet>
+     */
+    public function scopeCentralHub(Builder $query): Builder
+    {
+        $centralHubId = Cache::rememberForever('central_hub_id', function () {
```
Jamiras, line 103: I'm not fond of this solution. Can't we just force insert a `Central` hub after truncating the table, then go back to hardcoding the value (to 1)?

Jamiras: Could even consider force inserting the hubs that appear in the menu so they'll work for the seeded DB too (how does seeded data interact with the sync command)?

app/Platform/Commands/SyncGameSets.php
wescopeland: Done in latest.

app/Models/GameSet.php
wescopeland: I like that idea.

In latest, I've added this logic to the sync command which inserts the 5 "standard" hubs with hardcoded IDs of 1-5.

If the target games "backing" the hubs don't exist (ie: not a prod snapshot), the hubs are created anyway.

Realistically, someone on a seeded database will likely never execute this command. Therefore, given that these standard hubs are nav items, it probably also makes sense to seed these standard hubs.

A tiny new seeder has been added in latest:
```php
public function run(): void
{
    foreach (self::STANDARD_HUBS as $title => $id) {
        GameSet::unguard(); // temporarily allow filling the "id" field
        GameSet::create([
            'id' => $id,
            'title' => $title,
            'type' => GameSetType::Hub,
        ]);
        GameSet::reguard();
    }
}
```

app/Platform/Actions/BuildHubBreadcrumbsAction.php
```diff
@@ -0,0 +1,516 @@
...
+        'Custom Awards' => 'Central',
+        'Feature' => 'Central',
+        'Game Mechanic' => 'Central',
+        'Meta' => 'Central',
+        'Meta|QA' => 'Meta',
+        'Meta|DevComp' => 'Meta',
+        'Regional' => 'Central',
+        'Series' => 'Central',
+        'Subseries' => 'Series',
+        'Technical' => 'Central',
```
Jamiras, line 53: Here's another one-off:
image

The DevQuest sets hubs aren't parented to the the DevQuest event

Jamiras (approved):

PR #2926, opened 2024-12-09
feat: make React home page generally available

wescopeland:
This PR migrates `/demo/home` to `/`, making the new React/Inertia.js homepage generally available to all users.

Most of this PR's diff is removal of Blade components from the current PHP home page.

One additional feature was added that was missed in the current React implementation: new users see a CTA encouraging them to check out our FAQ. This has been ported to the React implementation.


resources/js/features/home/components/+root/HomeRoot.tsx
```diff
@@ -16,7 +17,9 @@ export const HomeRoot: FC = () => {
 
   return (
     <div className="flex flex-col gap-6">
-      {auth?.user ? null : <GuestWelcomeCta />}
+      {!auth?.user ? <GuestWelcomeCta /> : null}
+
+      {auth?.user?.isNew ? <NewUserCta /> : null}
```
Jamiras, line 22: I misunderstood this. It looks like we're missing the "Newest Registered User" component too. Was that intentionally removed?

wescopeland: It was intentionally removed because we gain a new user approximately every ~65s. I can add it back if you'd like; I suspect most users won't even notice it's missing.

Jamiras: I didn't notice until trying to test this change. I can't imagine anyone would complain.

Jamiras (approved):

PR #2927, opened 2024-12-09
feat(GameListDataTable): add automatic 'Group by console' functionality

wescopeland:
This PR enables the React datatable to simulate the "Group by console" functionality seen in our Blade game lists.

For the grouping functionality to kick in, all of the following must be true:
* The user must be sorting by system.
* There must be more than 1 system on the user's current page.
* The table must not be currently loading data.

Additionally, a loading state has been added to the table. While loading, the table's opacity slightly changes.

https://github.com/user-attachments/assets/756b0e3b-c560-46fe-a943-0b3dc81e6773


lang/en_US.json
gboquizosanchez, line 1: Minor issue in "rowGroupAriaLabel_one": "{{systemName}} group wtih {{val, number}} game".

With instead of wtih

wescopeland: Fixed, thanks!

resources/js/features/game-list/components/GameListDataTable/GameListDataTable.tsx
```diff
@@ -1,87 +1,172 @@
...
+  const getShouldGroupBySystem = (): boolean => {
+    if (isLoading && wasShowingGroups.current) {
+      return true;
+    }
+
+    if (isLoading && isSortedBySystem && !lastSortState.current) {
+      return false;
+    }
+
+    const hasMultipleSystems = rows?.length > 0 && new Set(rows.map(getSystemName)).size > 1;
```
Jamiras, line 55: This logic only applies to the data on the current page. If I set my "per page" to 10 and start paging through it, the system header disappears on pages that only match a single system.

Additionally, if the data rolls over to the next page, the counter only shows the number for the current page (for example, there's 3 more on the next page):
image

I imagine that using true counts regardless of the pagination wouldn't be worth the effort, but I don't imagine this sort will get used much outside of the Developed Sets page (which hasn't been converted yet).

wescopeland: Unfortunately, trying to make this work across pages slows the DB query down by 20x regardless of what SQL incantations I try.

wescopeland: In latest, I've removed the game count from the header rows.

Jamiras: It's still weird that it disappears on pages that only display games for a single system, but I'm not sure of a good solution for paging with headers.

Jamiras (approved):

PR #2928, opened 2024-12-10
fix(manage): allow Artists to give achievement badge artwork credit

wescopeland:
This PR fixes an issue where users solely with the Artist role cannot award credit for achievement badge artwork.

Artist users should be able to award badge artwork credit, but no other kind of credit, on achievements.

Screenshot 2024-12-10 at 5 48 22 PM


app/Policies/AchievementAuthorPolicy.php
```diff
@@ -65,4 +69,15 @@ public function forceDelete(User $user, AchievementAuthor $achievementAuthor): b
     {
         return false;
     }
+
+    public function canUpsertTask(User $user, AchievementAuthorTask $task): bool
+    {
+        // Artists can only give artwork credit.
+        if ($user->roles->count() === 1 && $user->hasRole(Role::ARTIST)) {
+            return $task === AchievementAuthorTask::Artwork;
+        }
```
Jamiras, line 78: I don't think you want to limit this to users with a single role. What if someone is an Artist and a Play Tester?

wescopeland: Yep, this is a logic error. Resolved in latest.

app/Policies/AchievementAuthorPolicy.php
```diff
@@ -40,7 +42,9 @@ public function create(User $user): bool
 
     public function update(User $user, AchievementAuthor $achievementAuthor): bool
     {
-        return $this->manage($user);
+        return
+            $this->manage($user)
```
Jamiras, line 46: Is this `manage` check still desirable?

Jamiras (approved):

app/Policies/AchievementAuthorPolicy.php
wescopeland: Good catch - it is now redundant. Removed in latest.

Jamiras (approved):

PR #2935, opened 2024-12-13
feat(api): add API_GetUserGameLeaderboards

gastongaiduk:
Hello guys, kind of had the need to fetch a specific user's leaderboards in a given game and found out quite time and resources consuming since the only way to do that is by calling each leaderboard entries API. This solution will use 6 requests to fetch all the content (which 3 are for validating request's parameters, 1 for fetching the total and the other 2 to fetch the needed information).

The endpoint can return the following errors
- 404 User not found
- 404 Game not found
- 422 Game has no leaderboards
- 422 User has no leaderboards on this game

```
 /*
 *  API_GetUserGameLeaderboards - returns a list of Leaderboards for the given User and GameID
 *    i : gameID
 *    u : username
 *    c : count - number of entries to return (default: 200, max: 500)
 *
 *  int         Count                       number of leaderboard records returned in the response
 *  int         Total                       number of leaderboard records the user has actually on the game
 *  array       Results
 *   object      [value]
 *    int        ID                         unique identifier of the leaderboard
 *    boolean    RankAsc                    true if the leaderboard views a lower score as better, false otherwise
 *    string     Title                      the title of the leaderboard
 *    string     Description                the description of the leaderboard
 *    string     Format                     the format of the leaderboard (see: ValueFormat enum)
 *    object     UserEntry                  details of the requested user's leaderboard entry
 *     object      [value]
 *      string     User                     username
 *      int        Score                    raw value score
 *      string     FormattedScore           formatted string value of score
 *      int        Rank                     user's leaderboard rank
 *      string     DateUpdated              an ISO8601 timestamp string for when the entry was updated
 */
```

Looking forward to know your thoughts


wescopeland:
Hi @gastongaiduk!

I definitely understand the desirability of this endpoint. Feel free to proceed.

I haven't looked over the code yet; I'm mostly just thinking out loud for the sake of design.

There are several games with hundreds or even thousands of leaderboards:
Screenshot 2024-12-13 at 6 57 51 PM

I'm thinking we may want this response to be paginated 😅 After doing a check in the prod DB, it seems there are only ~200 games with >=200 leaderboards. A default page size of 200 is probably fine along with a max page size of 500.

Jamiras:
Recommend using a return structure similar to `API_GetGameLeaderboards`:
https://github.com/RetroAchievements/RAWeb/blob/1ce55bdb7bf0a5f1c4c8d18948881172e8f4fa81/public/API/API_GetGameLeaderboards.php#L3-L22

Replacing `TopEntry` with `UserEntry`.

gastongaiduk:
Thanks for your messages!

@wescopeland I added pagination to the results, now I need one extra db query in order to know the total number of entries a user has for a given game (for the Total property), but the result is that we can also avoid doing the query if the count is 0, a pre check :)

@Jamiras turns out to be more convenient, thanks for the suggestion! I added also a couple of extra properties to enhance the result (rank and date updated).

Sabrina902-prog:
😺

tests/Feature/Api/V1/GetUserGameLeaderboardsTest.php
```diff
@@ -0,0 +1,406 @@
...
+        $game = Game::factory()->create(['ConsoleID' => $system->ID]);
+
+        Leaderboard::factory()->create(['GameID' => $game->ID]);
+
+        $this->get($this->apiUrl('GetUserGameLeaderboards', ['i' => $game->ID, 'u' => $user->User]))
+            ->assertUnprocessable()
+            ->assertJson(['User has no leaderboards on this game']);
+    }
+
+    public function testGetUserGameLeaderboards(): void
```
wescopeland, line 82: IMO, there's too much happening in this test case. Ideally, each individual test case should test a single thing and be easy for future maintainers to reason about.

I propose splitting this into separate test cases that each verify a specific behavior:
1. A basic test for a user with a single leaderboard entry.
2. A test for a user with multiple leaderboard entries.
3. Separate pagination tests (first page, last page, partial results, etc.)
4. A test specifically for rank calculation.

These focused test cases would make the test suite more maintainable, improve DX, and make it immediately clear which functionality is broken when a test fails. It would also make it easier to add new test cases in the future should any logic or behavior change.

Historically, in the past we crammed a lot of logic into single test cases because we were fearful of long test execution times. More recently, we've been trending towards caring much less about execution times and instead favoring more bite-sized test cases.

public/API/API_GetUserGameLeaderboards.php
```diff
@@ -0,0 +1,124 @@
...
+    ->whereNull('deleted_at')
+    ->count();
+
+if ($userLeaderboardEntriesCount === 0) {
+    return response()->json(['User has no leaderboards on this game'], 422);
+}
+
+$leaderboardEntries = LeaderboardEntry::select('leaderboard_entries.*')
+    ->addSelect([
+        'calculated_rank' => LeaderboardEntry::from('leaderboard_entries as entries_bis')
```
wescopeland, line 75: nit: I know what "bis" means, but I'm afraid it may confuse a future maintainer. Let's use a more descriptive alias.

public/API/API_GetUserGameLeaderboards.php
```diff
@@ -0,0 +1,124 @@
...
+    ->orderBy('LeaderboardDef.ID', 'asc')
+    ->skip($offset)
+    ->take($count)
+    ->get();
+
+$results = [];
+foreach ($leaderboardEntries as $leaderboardEntry) {
+    $results[] = [
+        'ID' => $leaderboardEntry->leaderboard->ID,
+        'RankAsc' => $leaderboardEntry->leaderboard->LowerIsBetter ? 'false' : 'true',
```
wescopeland, line 106: It seems like it may be better to just return `true` or `false` here rather than "true" or "false".

public/API/API_GetUserGameLeaderboards.php
Jamiras: I'm among the group of people not familiar with the term.

public/API/API_GetUserGameLeaderboards.php
```diff
@@ -0,0 +1,124 @@
...
+            ->where(function ($query) {
+                $query->where(function ($q) {
+                    $q->where('leaderboardDefBis.LowerIsBetter', 1)
+                      ->whereColumn('entries_bis.score', '<', 'leaderboard_entries.score');
+                })->orWhere(function ($q) {
+                    $q->where('leaderboardDefBis.LowerIsBetter', 0)
+                      ->whereColumn('entries_bis.score', '>', 'leaderboard_entries.score');
+                });
+            })
+            ->selectRaw('COUNT(*) + 1'),
```
Jamiras, line 89: I have a feeling that this isn't excluding unranked users.

public/API/API_GetUserGameLeaderboards.php
Jamiras: I think this is the result of a documentation error. I suggested making this look like the `API_GetGameLeaderboards` response, which says `RankAsc` is a string: https://github.com/RetroAchievements/RAWeb/blob/e402065f76947ac4ca356339c8eb1cc86b43b025/public/API/API_GetGameLeaderboards.php#L13

But it looks like it was changed to a true boolean at some point (or was just misdocumented):
https://github.com/RetroAchievements/RAWeb/blob/e402065f76947ac4ca356339c8eb1cc86b43b025/public/API/API_GetGameLeaderboards.php#L73

public/API/API_GetUserGameLeaderboards.php
```diff
@@ -0,0 +1,124 @@
...
+    return response()->json(['Game has no leaderboards'], 422);
+}
+
+$userLeaderboardEntriesCount = LeaderboardEntry::where('user_id', $user->id)
+    ->whereIn('leaderboard_id', function ($query) use ($game) {
+        $query->select('ID')
+              ->from('LeaderboardDef')
+              ->where('GameID', $game->ID);
+    })
+    ->whereNull('deleted_at')
```
Jamiras, line 66: Isn't this implied without the `withTrashed()` clause?

public/API/API_GetUserGameLeaderboards.php
wescopeland: In case it ever comes up again in code you read or if you're curious: "bis" is Latin for "again", "to repeat", or "to do a second time". In the context of a SQL query, it means to reuse a table for a subquery or some other operation.

public/API/API_GetUserGameLeaderboards.php
gastongaiduk: Then is a good opportunity to update the documentation on the GetGameLeaderboards API, I'll do that and apply a _boolval_ to this one too

public/API/API_GetUserGameLeaderboards.php
gastongaiduk: I have it a bit internalized haha, I'll use an alias a bit easier to understand :)

public/API/API_GetUserGameLeaderboards.php
gastongaiduk: Indeed it is implied, thanks for pointing out

public/API/API_GetUserGameLeaderboards.php
gastongaiduk: Most likely. Although, I'll need some details in order to confirm my changes are good :)

How does the unranked works? Not null is because they don't want to appear on leaderboards? What's the difference between unranked and untracked?

Thanks in advance

tests/Feature/Api/V1/GetUserGameLeaderboardsTest.php
gastongaiduk: I agree, was turning to be complex even to write them. I have to say I also prefer segregation, I'll adjust these. Thanks for the suggestions, clearly helpful :)

public/API/API_GetUserGameLeaderboards.php
wescopeland: Unranked and untracked are synonymous - we plan on ultimately dropping `Untracked` and fully moving over to simply using `unranked_at`, but this migration hasn't happened yet.

public/API/API_GetUserGameLeaderboards.php
```diff
@@ -0,0 +1,128 @@
...
+    })
+    ->whereNull('UserAccounts.unranked_at')
+    ->where('UserAccounts.Untracked', 0)
+    ->count();
+
+if ($userLeaderboardEntriesCount === 0) {
+    return response()->json(['User has no leaderboards on this game'], 422);
+}
+
+$leaderboardEntries = LeaderboardEntry::select('leaderboard_entries.*')
```
wescopeland, line 75: What happens if the user's leaderboard result is a tie with another player?

public/API/API_GetUserGameLeaderboards.php
Jamiras: It looks like the rank calculation needs to filter out deleted entries:
```
{"ID":85289,"RankAsc":true,"Title":"Tutorial Speedster","Description":"Complete the tutorial in the quickest time","Format":"TIMESECS","UserEntry":{"User":"Jamiras","Score":1004,"FormattedScore":"16:44","Rank":16,"DateUpdated":"2024-09-16T00:35:37+00:00"}}]}
```
image

image

```
| 123263 | Nightfire391  |   534 | 2024-07-28 11:37:13 |
| 316263 | Roninx2       |   544 | NULL                |
``

public/API/API_GetUserGameLeaderboards.php
gastongaiduk: It will work as it's currently done on the existing entries endpoint. If they are tied they will both get the same rank. Added a test to cover this scenario.

public/API/API_GetUserGameLeaderboards.php
gastongaiduk: Looks like the addSelect its not doing it by default while the regular select does. I added the conditionals to do this again and also covered the results with a test. Feel free to pull changes and give it another try :)

wescopeland (approved):
Seems good on my end - would also like @Jamiras's approval

Jamiras (approved):

PR #2936, opened 2024-12-13
feat(GameListDataTable): add 'Remember my view' checkbox

wescopeland:
This PR adds a cookie-driven "Remember my view" checkbox to the React datatables. Each table gets its own cookie.

https://github.com/user-attachments/assets/677211a9-f8b1-4669-948a-341d5ab939f9

**How it works:**
* Each table has its own persist cookie name, which is passed down from the controller as a prop.
* Since cookies can obviously be read server-side, `GameListRequest` has been modified to check for persisted cookie values and use those if URL query params for state are not present. URL query params always take priority.
* If the user unchecks the "Remember my view" checkbox, their local cookie is deleted.

**What gets persisted:**
* Column visibility.
* Sort order.
* Page size.
* Filter values that aren't the search input.

**What gets ignored/not persisted:**
* Page index.
* Search input (title filter).

**Tables affected:**
All React datatables.
* All Games.
* Want to Play Games.
* System Games.
* Hub Games.

I evaluated several different storage approaches while building this solution:
* I decided not to use a DB table because it required much more code and was much more difficult to test, especially with unstructured JSON payloads.
* I decided to use multiple cookies so there is no risk of hitting the 4KB max cookie size restriction. Overall, browsers can generally hold 6MB of distributed cookies per domain, and with this solution we stay well within that limit even if we somehow had 100+ persisted table kinds.


app/Platform/Controllers/HubController.php
```diff
@@ -0,0 +1,121 @@
...
+            // If the user navigates directly to the central hub ID,
+            // we should redirect them to "/hubs" instead.
+            $centralHub = GameSet::centralHub()->first();
+            if ($centralHub && $gameSet->id === $centralHub->id) {
+                return redirect()->route('hub.index');
+            }
+        }
+
+        // Return a 404 if this game set isn't actually for a hub.
+        if ($gameSet->type !== GameSetType::Hub) {
```
gboquizosanchez, line 60: Consider use abort_if instead

wescopeland: Good idea. This is from #2923, but I've made the change there and then remerged that branch into this one.

resources/js/features/game-list/components/DataTableToolbar/DataTableDesktopToolbar.tsx
```diff
@@ -38,32 +42,52 @@ export function DataTableDesktopToolbar<TData>({
...
+        </div>
+
+        <BaseLabel className="flex items-center gap-2 text-menu-link">
+          <BaseCheckbox
+            checked={isCurrentlyPersistingView}
+            onCheckedChange={(checked: boolean) => setIsCurrentlyPersistingView(checked)}
           />
-        ) : null}
+
+          {t('Remember my view')}
```
Jamiras, line 89: While this does remember the view, the filters don't reflect the remembered values when returning to the page.
image

image

resources/js/features/game-list/hooks/useTableSync.ts
```diff
@@ -1,25 +1,61 @@
...
+  isUserPersistenceEnabled = false,
 }: UseAutoUpdatingQueryParamsProps) {
+  const { persistenceCookieName } = usePageProps<{ persistenceCookieName: string }>();
+
+  const [cookie, setCookie, deleteCookie] = useCookie(persistenceCookieName);
+
+  useUpdateEffect(() => {
+    if (isUserPersistenceEnabled) {
+      // Don't persist filtering by title.
+      const persistedFilters = columnFilters.filter((filter) => filter.id !== 'title');
```
Jamiras, line 43: This doesn't seem to be working. The value is still appearing in my cookie:
```
%22sorting%22:[{%22id%22:%22title%22%2C%22desc%22:false}]
```

resources/js/features/game-list/hooks/useTableSync.ts
```diff
@@ -1,25 +1,61 @@
...
+
+      const tableState: Partial<TableState> = {
+        columnVisibility,
+        sorting,
+        columnFilters: persistedFilters,
+        pagination: { ...pagination, pageIndex: 0 }, // don't persist the page index
+      };
+
+      setCookie(JSON.stringify(tableState), { expires: 180 }); // 180 day (6 month) expiry
```
Jamiras, line 52: JSON is a very inefficient storage mechanism for cookies because of all the escaping:
image

But I don't have any better solutions. Perhaps we could restructure the data to eliminate most of the structure?
```
{"columnVisibility":{"hasActiveOrInReviewClaims":false,"lastUpdated":false,"numUnresolvedTickets":false,"numVisibleLeaderboards":false,"playersTotal":false,"progress":true,"retroRatio":false},"sorting":[{"id":"title","desc":false}],"columnFilters":[{"id":"achievementsPublished","value":["either"]},{"id":"system","value":["27"]},{"id":"progress","value":["unstarted"]}],"pagination":{"pageIndex":0,"pageSize":25}};
```
to
```
{"columnVisibility":"!hasActiveOrInReviewClaims|!lastUpdated|!numUnresolvedTickets|!numVisibleLeaderboards|!playersTotal|progress|!retroRatio","sort":"title","columnFilters":[{"achievementsPublished":["either"]},{"system":["27"]},{"progress":["unstarted"]}],"pageIndex":0,"pageSize":25};
```
Although, any sort of custom mutation introduces a point of failure. Is there some way to better filter out default values?

resources/js/features/game-list/components/DataTableToolbar/DataTableDesktopToolbar.tsx
```diff
@@ -38,32 +42,52 @@ export function DataTableDesktopToolbar<TData>({
...
+        <div className="flex w-full flex-col gap-2 sm:w-auto sm:flex-row md:gap-3">
+          {doesColumnExist(allColumns, 'system') && filterableSystemOptions?.length > 1 ? (
+            <DataTableSystemFilter
+              table={table}
+              filterableSystemOptions={filterableSystemOptions}
+            />
+          ) : null}
+
+          {doesColumnExist(allColumns, 'achievementsPublished') ? (
+            <DataTableAchievementsPublishedFilter table={table} />
```
Jamiras, line 66: At MD or smaller, the filters always show "1", regardless of the value
image

resources/js/features/game-list/hooks/useTableSync.ts
wescopeland: The cookie value demonstrated (without encoding) is:
```
"sorting": [{"id":"title","desc":false}]
```
The logic on this line is for ensuring that a filter by `title` is not persisted in the cookie. For example, if you search for "zelda" on the All Games list, then navigate directly to `/games`, the search (title filter) should not persist.

resources/js/features/game-list/components/DataTableToolbar/DataTableDesktopToolbar.tsx
wescopeland: This also happens in prod (unrelated to this branch). May require some careful thought to solve. 

The filters are showing "1" to represent they currently hold a single value (try filtering by multiple systems and the number will increment on the system filter). I think what may make sense is for the System numeric to continue functioning as-is at these breakpoints, and maybe have a different treatment for the other filters. We're awfully cramped for space, though.

May be best to address this in a different branch.

resources/js/features/game-list/components/DataTableToolbar/DataTableDesktopToolbar.tsx
wescopeland: Should be fixed in latest.

resources/js/features/game-list/hooks/useTableSync.ts
Jamiras: base64 encoding should generate about the same rate of compression. It's a fixed 33% bloat, whereas JSON encoding is 200% bloat, but only for for the structure characters, so if the keys and data compose 83% of the message, you end up with the same 33% bloat.

I did actually try that the other day, and my cookie went from something like 570 characters (escaped) to 550 characters (base64 encoded).

I'd be fine if you feel more comfortable using the escaped methodology.

Jamiras (approved):

resources/js/features/game-list/hooks/useTableSync.ts
wescopeland: Undid `btoa()` in latest. If we'd like to pursue a compression algorithm in the future, here's a few benchmarks/comparisons for algorithms that are design to compress JSON: https://senryoku.github.io/smol-string/

PR #2950, opened 2024-12-18
feat(api): add API_GetUserSetRequests

drisc:
This PR adds an endpoint to fetch the current set request list of a user along with their total request count and the amount of points they need to unlock their next request.


public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,50 @@
...
+$userRequestInfo = UserGameListEntry::getUserSetRequestsInformation($user);
+
+$requestList = [];
+foreach ($requestedSets as $gameId) {
+    $gameData = Game::with('system')->find($gameId);
+    $requestList[] = [
+        'GameID' => $gameData['ID'],
+        'Title' => $gameData['Title'],
+        'ConsoleID' => $gameData['ConsoleID'],
+        'ConsoleName' => $gameData->system['Name'],
```
wescopeland, line 41: This kicks off 2 SQL queries for each of the user's set requests. Against some users in the database, this results in 500+ queries every time this endpoint gets called.

wescopeland (commented):
Hi @drisc!

Thanks for opening this PR! I've briefly scanned over the implementation and left one comment for you to chew on 🙂

Once it's resolved, I'm happy to do a deeper dive into the patchset implementation details. 

In the meantime, I will be temporarily moving this PR to draft. From our [latest contributing requirements](https://github.com/RetroAchievements/RAWeb/blob/master/docs/CONTRIBUTING.md#requirements), any API changes require a sister PR in [api-docs](https://github.com/RetroAchievements/api-docs). An example of a good one can be found at  https://github.com/RetroAchievements/api-docs/pull/68!

Once an api-docs PR is opened, please feel free to move this PR out of draft! Thanks for your understanding 🙏

public/API/API_GetUserSetRequests.php
drisc: As it stands none of the game data is fetched by the UserGameListEntry Model, it only provides the GameID from the SetRequests table which I then have to look up in order to get the data for the response.

Is the second SQL query that you are referring to the one for ConsoleName? The name is not stored in the GameData table so I have to call out to the Console table for it, I can remove that if it adds too much weight to the request.

wescopeland: The two queries are coming from these two lines:
```php
$gameData = Game::with('system')->find($gameId);

'ConsoleName' => $gameData->system['Name'],
```
If you want to see the queries firsthand, in _.env_, make sure this value is set:
```
DEBUGBAR_ENABLED=true
```
Then copy over most of your endpoint code to some random PHP page and use hardcoded values for the user. The debugbar will show all the SQL queries that get executed on the page - you will see two queries kicked off for every game on the user's set request list.

The [issue](https://www.appdynamics.com/blog/product/common-application-problems-and-how-to-fix-them-the-select-n-1-problem/) is the queries are being kicked off in the loop. You'll need to lift this data fetching out of the loop to resolve the performance issue.

drisc: I have lifted the queries out of the loop, after testing on a PHP page like you suggested each query is now only run once.

public/API/API_GetUserSetRequests.php
wescopeland, line 1: Query param input seems to not be using any sort of validation in this endpoint.

tests/Feature/Api/V1/UserSetRequestsTest.php
```diff
@@ -0,0 +1,72 @@
...
+        UserGameListEntry::factory()->create([
+            'user_id' => $user->ID,
+            'type' => 'achievement_set_request',
+            'GameID' => $game->ID,
+        ]);
+
+        $userRequestInfo = [
+            'total' => 1,
+            'pointsForNext' => 999,
+        ];
```
wescopeland, line 54: nit: Mental load for this test case could be reduced by deleting this variable and just hardcoding the values into the assert.

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,52 @@
...
+}
+
+$requestedSets = UserGameListEntry::where('user_id', $user->ID)
+    ->where('type', 'achievement_set_request')
+    ->pluck('GameID')
+    ->toArray();
+
+$games = Game::with('system')
+    ->whereIn('ID', $requestedSets)
+    ->get();
```
wescopeland, line 34: This is significantly better than before, but can be optimized even further.

If you combine these two queries into a single query, three benefits are unlocked:
* Memory use is reduced because we can remove `->toArray()`.
* We can avoid a large `WHERE IN` lookup.
* We only need to rely on a single composite index.

database/factories/UserGameListEntryFactory.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+class UserGameListEntryFactory extends Factory
+{
+    protected $model = UserGameListEntry::class;
+
+    public function definition(): array
+    {
+        return [
+            'user_id' => User::factory(),
+            'type' => '',
```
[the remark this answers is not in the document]
drisc: Default added in latest

app/Models/UserGameListEntry.php
```diff
@@ -5,12 +5,21 @@
...
 
 class UserGameListEntry extends BaseModel
 {
+    /** @use HasFactory<UserGameListEntryFactory> */
+    use HasFactory;
+
+    protected static function newFactory(): UserGameListEntryFactory
+    {
+        return UserGameListEntryFactory::new();
+    }
```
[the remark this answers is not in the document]
drisc: Position updated in latest

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,52 @@
...
+
+$userRequestInfo = UserGameListEntry::getUserSetRequestsInformation($user);
+
+$requestList = $games->map(function ($gameData) {
+    return [
+        'GameID' => $gameData['ID'],
+        'Title' => $gameData['Title'],
+        'ConsoleID' => $gameData['ConsoleID'],
+        'ConsoleName' => $gameData->system['Name'],
+        'ImageIcon' => $gameData['ImageIcon'],
```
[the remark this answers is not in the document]
drisc: I'll keep this in mind for the future, updated to suggestion in latest

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,52 @@
...
+ *  int        PointsForNext              number of points remaining until maximum request increase
+ */
+
+use App\Models\Game;
+use App\Models\User;
+use App\Models\UserGameListEntry;
+
+$user = User::firstWhere('User', request()->query('u'));
+if (!$user) {
+    return response()->json([]);
```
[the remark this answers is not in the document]
drisc: 404 response added, test updated to reflect this as well

public/API/API_GetUserSetRequests.php
drisc: Input validation has been added based on the implementation in other endpoints

tests/Feature/Api/V1/UserSetRequestsTest.php
drisc: Good catch, this was left over from when I was trying to figure out how to get predictable results from the User creator. Removed in latest

public/API/API_GetUserSetRequests.php
drisc: I have condensed this down to a single query and combined it with the map for the final response.

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,52 @@
...
+    ->get()
+    ->map(function ($gameData) {
+        return [
+            'GameID' => $gameData->game->id,
+            'Title' => $gameData->game->title,
+            'ConsoleID' => $gameData->game->system->id,
+            'ConsoleName' => $gameData->game->system->name,
+            'ImageIcon' => $gameData->game->ImageIcon,
+        ];
+    });
```
Jamiras, line 44: Recommend adding an orderBy for consistent results (most important for unit tests, but beneficial for consumers as well). The page is sorted by game title, so that probably makes the most sense, but requires specifying titles in the unit tests as the random game titles wouldn't have a guaranteed order.

Jamiras:
Users with inflated requests due to April Fools shenanigans are reporting "TotalRequests" as their supposed limit, not the actual number of requests they have.
image
```
...,"TotalRequests":13,"PointsForNext":4425}
```
This does match their page, so I think it's probably working as designed.

public/API/API_GetUserSetRequests.php
drisc: Nice, I’ll use that method in the future. I was using Content.php but that has a lot of other calls on each page.

public/API/API_GetUserSetRequests.php
drisc: I was wondering about that, by default it seems to be ordering the result by GameID, they are out of order in my SetRequests table. 

I’ll add the orderBy, I’ll need to make the test return more than one result to get an accurate assert.

public/API/API_GetUserSetRequests.php
wescopeland: May be best to use:
```php
    'u' => ['required', 'min:2', 'max:20', new CtypeAlnum()],
```

public/API/API_GetUserSetRequests.php
wescopeland: Much better. Can be optimized even further though!

```php
// no need for map here, the fields are already aliased

// ->get()
// ->toArray();
// ^^^ faster
->map(function ($gameData) {
    return [
        'GameID' => $gameData->GameID,
        'Title' => $gameData->Title,
        'ConsoleID' => $gameData->ConsoleID,
        'ConsoleName' => $gameData->ConsoleName,
        'ImageIcon' => $gameData->ImageIcon,
    ];
});
```

```php
->orderBy('GameData.sort_title', 'asc') // nit: 'asc' is the default
```

public/API/API_GetUserSetRequests.php
drisc: Updated in latest

public/API/API_GetUserSetRequests.php
drisc: Map has been removed along with explicit 'asc'. I thought I would include it for readability but have removed it based on the nit.

wescopeland (approved):

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,54 @@
...
+    'GameData.ConsoleID',
+    'Console.Name as ConsoleName',
+])
+    ->join('GameData', 'SetRequest.GameID', '=', 'GameData.ID')
+    ->join('Console', 'GameData.ConsoleID', '=', 'Console.ID')
+    ->where('SetRequest.user_id', $user->id)
+    ->where('type', 'achievement_set_request')
+    ->orderBy('GameData.sort_title')
+    ->get()
+    ->toArray();
```
Jamiras, line 46: I just noticed that this is returning completed requests (i.e. the list returned when clicking on "View All User Set Requests"). It should probably only return active requests (`where GameData.achievements_published = 0`)

Is there any reason we'd want to expose the completed requests? If so, we might want a flag for doing so.

drisc: The site page allows for viewing all past requests as well so adding a flag for it would be my vote. That way the endpoint can get a list of *all* a users requests if needed.

drisc: Type flag added in latest, I've also updated and added a test to check for achievements_published.

public/API/API_GetUserSetRequests.php
```diff
@@ -23,27 +24,49 @@
 
 $input = Validator::validate(Arr::wrap(request()->query()), [
     'u' => ['required', 'min:2', 'max:20', new CtypeAlnum()],
+    't' => ['nullable', 'min:1', 'max:1', new CtypeAlnum()]
```
wescopeland: Correct me if I'm wrong, this means 0 is not an accepted explicit value (some API clients might set the value regardless of whether it's optional).

drisc: Would changing it to `'t' => ['sometimes', 'integer', 'min:1', 'max:1', new CtypeAlnum()` be the fix here?

It still evaluates properly when set to 0 as well as when it is omitted entirely in the current code.

wescopeland: I think `min` and `max` may actually be checking the length of the given value rather than enforcing a subset of acceptable values. ie: I bet "9" is also valid. ref: https://laravel.com/docs/11.x/validation#rule-min

We may want:
```php
't' => ['nullable', 'in:0,1']
```

drisc: You are correct, 9 is indeed a valid type haha

Updated with the suggested code in latest, types are now constrained to 0 and 1.

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,76 @@
...
+    ])
+        ->join('GameData', 'SetRequest.GameID', '=', 'GameData.ID')
+        ->join('Console', 'GameData.ConsoleID', '=', 'Console.ID')
+        ->where('SetRequest.user_id', $user->id)
+        ->where('type', 'achievement_set_request')
+        ->where('GameData.achievements_published', '=', '0')
+        ->orderBy('GameData.sort_title')
+        ->get()
+        ->toArray();
+}
```
Jamiras, line 68: You can avoid duplicating most of this:
```
$query = UserGameListEntry::select([
    'GameData.id as GameID',
    'GameData.Title',
    'GameData.ImageIcon',
    'GameData.ConsoleID',
    'Console.Name as ConsoleName',
])
    ->join('GameData', 'SetRequest.GameID', '=', 'GameData.ID')
    ->join('Console', 'GameData.ConsoleID', '=', 'Console.ID')
    ->where('SetRequest.user_id', $user->id)
    ->where('type', 'achievement_set_request');
        
if ($type !== 1) {
    $query ->where('GameData.achievements_published', '=', '0');
} 

$requestedSets = $query->orderBy('GameData.sort_title')->get()->toArray();
```

drisc: Thanks, I have applied that code squash in latest.

Jamiras (approved):

tests/Feature/Api/V1/UserSetRequestsTest.php
```diff
@@ -0,0 +1,141 @@
...
+            'GameID' => $game->ID,
+        ]);
+        // Create the second user game list entry for the set request
+        UserGameListEntry::factory()->create([
+            'user_id' => $user->ID,
+            'type' => 'achievement_set_request',
+            'GameID' => $game2->ID,
+        ]);
+
+        // Note that only the second game is present in the assert. The  code will find the first game as well but only return the game with no published achievements.
```
wescopeland, line 124: nit: there is a small typo here.
```suggestion
        // Note that only the second game is present in the assert. The code will find the first game as well but only return the game with no published achievements.
```

tests/Feature/Api/V1/UserSetRequestsTest.php
```diff
@@ -0,0 +1,141 @@
...
+                ],
+                'TotalRequests' => 1,
+                'PointsForNext' => 999,
+            ]);
+    }
+
+    /**
+     * Test that the API returns only set requests with no published achievements for an existing user.
+     */
+    public function testGetUserSetRequests(): void
```
wescopeland, line 92: nit: On test failure, the method name is printed like:
"get user set requests"
But this doesn't accurately describe what is being tested in this test case. 

This might be better:
```
testGetUnpublishedUserSetRequests()
```

tests/Feature/Api/V1/UserSetRequestsTest.php
wescopeland, line 1: "achievement_set_request" is a magic string. It would be better to leverage `UserGameListType::AchievementSetRequest`.

public/API/API_GetUserSetRequests.php
```diff
@@ -0,0 +1,61 @@
...
+    'GameData.id as GameID',
+    'GameData.Title',
+    'GameData.ImageIcon',
+    'GameData.ConsoleID',
+    'Console.Name as ConsoleName',
+])
+    ->join('GameData', 'SetRequest.GameID', '=', 'GameData.ID')
+    ->join('Console', 'GameData.ConsoleID', '=', 'Console.ID')
+    ->where('SetRequest.user_id', $user->id)
+    ->where('type', 'achievement_set_request');
```
wescopeland, line 47:
```suggestion
    ->where('type', UserGameListType::AchievementSetRequest);
```

tests/Feature/Api/V1/UserSetRequestsTest.php
drisc: Function name changed in latest to suggestion.

tests/Feature/Api/V1/UserSetRequestsTest.php
drisc: Changed in latest.

public/API/API_GetUserSetRequests.php
drisc: Updated in latest, I'll remember this for next time.

wescopeland (approved):

PR #2952, opened 2024-12-19
feat(hub): add support for content warnings

wescopeland:
On deploy, set the "[Theme - Mature]" game_set to have mature content.

---

This PR adds support for flagging any hub as having a "content warning". If a hub is flagged as having a content warning, viewing any games in the hub and/or the hub itself will display a content/age gate.

In the management app, there is a new toggle on the `HubResource`, which can be set by admins and staff devs:

Screenshot 2024-12-18 at 7 43 14 PM

When set, viewing the hub exhibits a new behavior:

https://github.com/user-attachments/assets/b83ed471-a381-41d6-8036-bf35e3b1a49b

_gameInfo.blade.php_ has been adjusted to use this new flag rather than the hardcoded "[Theme - Mature]" legacy hub.


app/Filament/Resources/HubResource.php
```diff
@@ -85,6 +95,12 @@ public static function form(Form $form): Form
                             ->minLength(2)
                             ->maxLength(80)
                             ->rules([new NoEmoji()]),
+
+                        Forms\Components\Toggle::make('has_content_warning')
+                            ->label('Has Content Warning')
+                            ->helperText('CAUTION: If this is enabled, players will get a warning when opening any game in the hub!')
+                            ->default(false)
+                            ->visible(fn ($record) => $user->can('toggleHasContentWarning', $record)),
```
Jamiras, line 103: I can't decide if this is a poorly implemented feature, or a stub for future functionality.

As it stands, this is specifically for the "mature content warning", but it seems like it could be for any type of content warning (i.e. seizure warnings). I'm assuming that in the the future we want to break "mature content" into more categories, like: language, nudity, sex, abuse. Otherwise, there would be little benefit to allowing the value to be changed by users.

Maybe it should be renamed to `has_mature_content` or something similar. Or changed to an enum that can be set to "mature" or "seizure" or whatever other warnings we want to add in the future.

resources/js/common/components/ContentWarningDialog/ContentWarningDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+              <span>{t('Are you sure you want to view this page?')}</span>
+            </BaseAlertDialogDescription>
+          </BaseAlertDialogHeader>
+
+          <BaseAlertDialogFooter className="mt-2 flex flex-col gap-5 sm:flex-row sm:gap-1">
+            <BaseAlertDialogAction onClick={handlePermanentYesClick}>
+              {t("Yes, and don't ask again")}
+            </BaseAlertDialogAction>
+
+            <BaseAlertDialogAction onClick={handleYesClick}>{t('Yes')}</BaseAlertDialogAction>
```
Jamiras, line 76: I think we want to keep the "Yes, I'm an adult" for this button. I think it removes some sort of legal liability from us since it causes users who want to get past the warning to knowingly lie about their age.

resources/js/common/components/ContentWarningDialog/ContentWarningDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+            <BaseAlertDialogAction onClick={handleNoClick}>{t('No')}</BaseAlertDialogAction>
```
Jamiras, line 78: Clicking No makes the page visible while it's redirecting to home. Can we keep it blurred?

app/Filament/Resources/HubResource.php
wescopeland: > I can't decide if this is a poorly implemented feature, or a stub for future functionality.

The intention of this feature is to formalize our existing mature content handling in a more maintainable way, without a reliance upon comparing against a magic string. Ideally we can move away from stuff like:
```php
if ($hub['Title'] == '[Theme - Mature]') {
    if ($userDetails && BitSet($userDetails['websitePrefs'], $matureContentPref)) {
        break;
    }
    $gate = true;
}
```

Configuration of this value is backed by a policy:
```php
public function toggleHasMatureContent(User $user, GameSet $gameSet): bool
{
    return $user->hasAnyRole([
        Role::ADMINISTRATOR,
        Role::DEVELOPER_STAFF, // <---- maybe this should be removed and only allow admin?
    ]);
}
```

> As it stands, this is specifically for the "mature content warning", but it seems like it could be for any type of content warning (i.e. seizure warnings).

While I agree that we should support these, mature content has different handling requirements. We wouldn't block access to games with seizure risks, but would want to warn users on those respective game pages. The nuance is likely in how "elevated" a warning type needs to be.

For future content warning types, we could use string enum columns, JSONB, or even tags. Each approach has trade-offs that will need to be evaluated, and right now I'm not sure what the best choice is.

> Maybe it should be renamed to has_mature_content or something similar. Or changed to an enum that can be set to "mature" or "seizure" or whatever other warnings we want to add in the future.

I've renamed this field to `has_mature_content` in latest.

resources/js/common/components/ContentWarningDialog/ContentWarningDialog.tsx
wescopeland: Good point. Updated in latest.

resources/js/common/components/ContentWarningDialog/ContentWarningDialog.tsx
wescopeland: We can - this a bug. Fixed in latest.

Jamiras (approved):

PR #2960, opened 2024-12-20
perf: prefetch client-side route transitions

wescopeland:
This PR leverages new Inertia 2.0 functionality to dramatically improve the performance of client-side route transitions between Inertia.js pages.

**It's best to test this PR with SSR enabled.**
```
sail pnpm build
```

A new component has been added to the common components library: `<InertiaLink />`. This is a smart hyperlink with desktop & mobile prefetching capabilities.

This component is intended to be a drop-in replacement for `<a />` tags. **It does not work with legacy Blade routes.** Here's an implementation example:
```tsx
<InertiaLink href={route('game.index')}>{t('All Games')}</InertiaLink>
```

Here's a demonstration of the component at work:

https://github.com/user-attachments/assets/31908797-56e7-4b96-be5b-98b3a5c94853

Under non-distressed network conditions, pages appear to load instantly. The browser network tab tells the real story:

https://github.com/user-attachments/assets/95330f84-e704-4649-bdc1-b6e7cafbfb00

All client-side routes should use `<InertiaLink />`. It supports a few configuration options:
```ts
// link-prefetch-behavior.model.ts
export type LinkPrefetchBehavior =
  | 'never'
  | 'desktop-hover-only'
  | 'desktop-hover-and-mobile-intersect';

// InertiaLink.tsx
type InertiaLinkProps = Omit<OriginalInertiaLinkProps, 'prefetch'> & {
  /**
   * Controls prefetch behavior:
   * - never: Do not prefetch this link under any circumstance.
   * - desktop-hover-only: Default. Prefetch when the user hovers over the link on desktop.
   * - desktop-hover-and-mobile-intersect: Prefetch on hover for desktop, and on visible for mobile.
   */
  prefetch?: LinkPrefetchBehavior;
};
```


Jamiras:
Is this really important? Can users not wait an extra second or two for a page transition? Even with your 75ms delay, there's a non-zero chance that this will prefetch the data even if the user doesn't intend to go to that page. And that just puts more load on the server to address a client-side concern that I haven't heard anyone complain about.

wescopeland:
I can definitely see there being times in the future where we can be reasonably certain a user is going to click on X link on the page and it makes sense to wrap that link up in some kind of prefetch behavior.

However, I agree that perhaps defaulting to prefetching being enabled on the link components could be too assertive of a change.

I feel we should still have some wrapper around Inertia 2.0's client-side routing prefetching behavior - it is suboptimal for mobile and this component fixes that.

In latest, I've set the default `prefetch` prop value to "never".

resources/js/features/home/components/SeeMoreLink/SeeMoreLink.tsx
```diff
@@ -10,12 +12,18 @@ interface SeeMoreLinkProps {
    * route to improve the performance of loading that page.
    */
   asClientSideRoute?: boolean;
+
+  /**
+   * Optional. You can manually adjust the prefetch strategy of this link.
+   * Defaults to "desktop-hover-only".
```
wescopeland, line 18:
```suggestion
   * Defaults to "never".
```

Jamiras (approved):
Noticed that at 10 items (no scroll bar), clicking the next button immediately prefetches the next "next page". I expect this is okay, as a user proceeding to page 2 will likely proceed farther as well. Not sure if we want to require a mouse-out first?

wescopeland:
When using the table with the network tab open, you'll see all kinds of prefetches happening. For example, they occur when clicking a column header and hovering over a sort order.

This is actually done manually(!!) unrelated to the changes in this branch and has been occurring in prod since the table itself launched.

In _DataTablePagination.tsx_, we have:
```tsx
/**
 * Handles page change and optional prefetching of adjacent pages for a smoother user experience.
 *
 * @param {Array<'next' | 'previous'>} [prefetchDirections] - Optional. Specifies whether to prefetch
 * adjacent pages. 'next' prefetches the following page, and 'previous' prefetches the previous page.
 * This is helpful for loading pages before the user actually clicks the pagination button.
 *
 * @example
 * // Navigate to page 3 and prefetch the next page:
 * handlePageChange(3, ['next']);
 *
 * // Navigate to page 2 and prefetch both next and previous pages:
 * handlePageChange(2, ['next', 'previous']);
 */
const handlePageChange = (
  newPageIndex: number,
  prefetchDirections?: Array<'next' | 'previous'>,
) => {
  // Update the current page.
  table.setPageIndex(newPageIndex);

  // Handle prefetching of adjacent pages.
  const lastPageIndex = table.getPageCount() - 1;
  const canPrefetchNext = prefetchDirections?.includes('next') && newPageIndex < lastPageIndex;
  const canPrefetchPrevious = prefetchDirections?.includes('previous') && newPageIndex > 0;

  if (canPrefetchNext) {
    prefetchPagination({
      newPageIndex: Math.min(newPageIndex + 1, lastPageIndex),
      newPageSize: pagination.pageSize,
    });
  }
  if (canPrefetchPrevious) {
    prefetchPagination({
      newPageIndex: Math.max(newPageIndex - 1, 0),
      newPageSize: pagination.pageSize,
    });
  }

  scrollToTopOfPage();
};
``` 

I think the next huge optimization for the table is probably a "client-side only" mode.

Several weeks ago, it occurred to me that if the table has <100 games total when no filtering is applied (ie: Virtual Boy games list, small hubs, etc), there's probably no reason to filter, sort, search, or paginate on the server at all. Just have the server send the whole list of games and do all filtering, sorting, searching, and pagination completely in memory on the client.

I've gotten pretty far with a proof-of-concept, but it's not yet in a place I'm totally happy with it. It's probably something I'll shoot to land after the remaining tables, such as dev sets, have been migrated.

PR #2965, opened 2024-12-21
Notify clients of impending hardcore restriction

Jamiras:
Sends achievements to clients that we have deemed are outdated or unsupported so users will be informed that hardcore will no longer be supported in those emulators after a grace period. I've arbitrarily selected "sometime in April" - which gives us a bit more than three months to address any concerns.

Here's an example using RetroArch 1.11.0:
image

NOTE: The achievement description contains more information, but not all emulators show the description in the popup.
image

These achievements will be set as already unlocked in softcore, so they should only pop if the player has hardcore mode enabled in their emulator.

After this is merged, I'll create a forum post (and probably a news article linking to the forum post), to give this the most visibility possible. I don't think we'll get much push-back from people using outdated emulators, but I imagine there will be push-back from people using unsupported emulators (like PizzaBoy and AetherSX2).

Outdated versions have to be specified through the management application. The goal is to mark anything over two years old that has a newer version available as outdated. Unfortunately, this is currently a manual process (i.e. RetroArch 1.15.0 will become outdated in mid-March, so we'll have to remember to mark it as such then). I think we probably could automate this logic, but I'm concerned that something hasn't has a release in over two years would immediately invalidate the most common in-use version as soon as a new version is acknowledged.


wescopeland:
> RetroArch 1.15.0 will become outdated in mid-March, so we'll have to remember to mark it as such then

I wish there was a better way to do this, but I'm having trouble coming up with any good ideas. The only thing I can really think of is if maybe `emulator_releases` had some kind of timestamped column, like `outdated_at` or `supported_until`? Then it would just be a set-and-forget kind of thing.

tests/Feature/Platform/Services/UserAgentServiceClientSupportLevelTest.php
wescopeland, line 1: A question mostly driven by intellectual curiosity and ignorance: what would happen if `minimum_allowed_version` is set, no `minimum_hardcore_version` is set, and `active` for the emulator is 0 (false)? Is such a thing even possible?

app/Connect/Actions/InjectPatchClientSupportLevelDataAction.php
```diff
@@ -60,9 +60,15 @@ private function buildClientSupportWarningAchievement(ClientSupportLevel $client
...
-            'Description' => 'Hardcore unlocks cannot be earned using this emulator.',
+            'Title' => match ($clientSupportLevel) {
+                ClientSupportLevel::Outdated => 'Warning: Outdated Emulator (please update)',
+                ClientSupportLevel::Unsupported => 'Warning: Unsupported Emulator',
+                default => 'Warning: Unknown Emulator',
+            },
+            //            'Description' => 'Hardcore unlocks cannot be earned using this emulator.',
+            'Description' => ($clientSupportLevel === ClientSupportLevel::Outdated) ?
+                'Starting in April, all unlocks from this emulator version will be softcore' :
+                'Starting in April, all unlocks from this emulator will be softcore',
```
wescopeland, line 71: Do you think it would make sense to check if the current date is greater than or equal to May 1 2025, and if so, change this message?

Jamiras: There's going to be another PR to actually enable the downgrading (see TODOs in dorequest.php). We should just change the message then.

tests/Feature/Platform/Services/UserAgentServiceClientSupportLevelTest.php
Jamiras: If minimum allowed is set, and the version is greater than the minimum allowed, but no minimum hardcore is set, it'll be given full permission.

The active check is only a fallback is both minimum allowed and minimum hardcore are null, which will be the case for all of the unsupported clients. A supported client should provide at least one minimum value, but several weren't (I have updated the prod DB while working on this PR).

Jamiras:
> The only thing I can really think of is if maybe emulator_releases had some kind of timestamped column, like outdated_at or supported_until? Then it would just be a set-and-forget kind of thing.

The management page uses emulator_releases.created_at as a release date column, so the outdated_at would just be two years after that.

The problem is still going to be what happens if the next version is released three years after that? Maybe we could make the logic automatically expire if a newer version _has existed_ for three months in addition to the version being more than two years old?

wescopeland:
> The problem is still going to be what happens if the next version is released three years after that? Maybe we could make the logic automatically expire if a newer version has existed for three months in addition to the version being more than two years old?

That seems like a compelling idea. Definitely not urgent, but could be considered to reduce the management overhead and/or make things less reliant on human memory (or prone to human error).

wescopeland (approved):

PR #2966, opened 2024-12-22
feat: add dedicated SEO component, fix several SEO issues

wescopeland:
This PR remediates several SEO issues with the Inertia.js pages and adds a new heavily-documented `SEO` component.

The component can be used like:
```tsx
<SEO
  title={t('Comments - {{gameTitle}}', { gameTitle: game.title })}
  description={`General discussion about the achievement set for ${game.title}`}
  ogImage={game.badgeUrl}
/>
```

The component has several benefits:
* `title` now throws a type error when given an untranslated string.
* `description` automatically propagates to `og:description`.
* Twitter and other social embed fields are automatically handled.
* The page canonical URL is automatically set.
* `type` now defaults to "website", and is strongly-typed to only accept "website", "article", or "profile". These are the only correct values on the RA website.
* JSON+LD structured data is automatically populated which gives us an increased authority boost with Google.

If a developer attempts to import `Head` into a component, they'll now get an error from ESLint with a message telling them to use `@/common/components/SEO` instead.


resources/js/features/home/components/+root/FrontPageNews/NewsCard.tsx
```diff
@@ -90,6 +91,8 @@ const NewsCardImage: FC<NewsCardImageProps> = ({ src }) => {
 
       {isImageValid ? (
         <div
+          role="img"
+          aria-label="news post photo"
           className="h-28 w-full rounded bg-cover bg-center sm:w-[197px]"
```
wescopeland, line 96: Google is angry because this element doesn't have a role or aria-label. It's fine that this is a plain English string.

resources/js/pages/achievement/[achievement]/comments.tsx
```diff
@@ -10,17 +10,13 @@ const AchievementComments: AppPage<App.Community.Data.AchievementCommentsPagePro
...
-  const metaDescription = `General discussion about the achievement ${achievement.title}`;
-
   return (
     <>
-      <Head title={t('Comments - {{achievementTitle}}', { achievementTitle: achievement.title })}>
-        <meta name="description" content={metaDescription} />
-        <meta name="og:description" content={metaDescription} />
-
-        <meta property="og:image" content={achievement.badgeUnlockedUrl} />
-        <meta property="og:type" content="retroachievements:comment-list" />
```
Comment (wescopeland) [inline resources/js/pages/achievement/[achievement]/comments.tsx:22]: All of these og:type values have been replaced with "website".

User-defined types like "retroachievements:comment-list" are valid, but not officially recognized. The valid types can be found here: https://ogp.me/#no_vertical

resources/js/common/components/SEO/SEO.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const isArticle = type === 'article';
+
+  return (
+    <Head title={title}>
+      {/* Base metadata */}
+      <meta name="description" content={description} />
+      <link rel="canonical" href={resolvedCanonicalUrl} />
+
+      {/* Open Graph metadata */}
+      <meta property="og:site_name" content={ogSiteName} />
```
Jamiras, line 202: Are these all supposed to end with `inertia`?:
```
<meta property="og:site_name" content="RetroAchievements" inertia>
```

Jamiras (approved):

resources/js/common/components/SEO/SEO.tsx
wescopeland: Yea - Inertia does this on purpose as a runtime performance optimization.

On client-side navigation, Inertia needs to quickly identify which head elements it inserted during potential previous navigations, as well as determine new ones to add (if any). The `inertia` attribute helps Inertia quickly access those tags with a selector.

PR #2967, opened 2024-12-22
fix(ReportIssue): hide Create Ticket button if user cannot create tickets

wescopeland:
This PR fixes a minor issue on the Inertia.js Report Issue screen.

Now, if the current user does not have permission to create tickets, they will no longer see any "Create Ticket" button.


app/Policies/CommentPolicy.php
```diff
@@ -47,19 +47,6 @@ public function viewAny(?User $user, Model $commentable): bool
...
-            if ($commentable !== null && $commentable instanceof \App\Models\Ticket) {
-                $commentable->loadMissing(['achievement.developer']);
-
-                $didAuthorAchievement = $commentable->achievement->developer->id === $user->id;
-
-                return
-                    $didAuthorAchievement
-                    && $commentable->is_open
-                    && $user->hasAnyRole([Role::DEVELOPER_STAFF, Role::DEVELOPER]);
-            }
```
wescopeland, line 61: Taking the opportunity now to remove all this stuff.

app/Platform/Controllers/ReportAchievementIssueController.php
```diff
@@ -41,11 +42,14 @@ public function index(Request $request, Achievement $achievement): InertiaRespon
...
+
         $props = new ReportAchievementIssuePagePropsData(
             achievement: $achievementData,
             hasSession: $foundPlayerAchievement ? true : $user->hasPlayed($achievement->game),
             ticketType: $this->determineTicketType($foundPlayerAchievement, $allPlayerAchievements),
             extra: $request->input('extra'),
+            can: $can,
         );
 
         return Inertia::render('achievement/[achievement]/report-issue', $props);
```
Jamiras, line 55:
```
Too few arguments to function App\Policies\TriggerTicketPolicy::create(), 1 passed in /var/www/html/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 815 and exactly 2 expected
```

Trying to click on "Report Issue" from achievement page

wescopeland: Concerning that a test did not catch this.

I've fixed the issue and added a test case in latest.

app/Policies/CommentPolicy.php
Jamiras: Seems like it might still be necessary.
```
// Even when muted, developers may still comment on tickets for their own achievements.
```
After setting my `muted_until`, I'm unable to comment on tickets for my achievements.

wescopeland: Then it's working as intended.

Discussion here: https://discord.com/channels/476211979464343552/1002688331827658774/1319427748216967242

Jamiras (approved):

PR #2968, opened 2024-12-22
feat(hub): restore forum topics

wescopeland:
This PR adds a tiny migration for the `game_sets` table and includes a new script to run:
```bash
sail artisan ra:sync:game-sets:hub-forum-topic-ids
```

After some discussion with the maintainers of several hubs, it does sound like we'll need these forum topics after all, at least in the near-term.

If a hub has a forum topic, a URL appears both in the management app as well as the hub itself (visible to those who can manage the hub only):

Screenshot 2024-12-22 at 3 46 51 PM


app/Filament/Resources/HubResource.php
```diff
@@ -57,6 +58,12 @@ public static function infolist(Infolist $infolist): Infolist
                         Infolists\Components\TextEntry::make('title')
                             ->label('Title'),
```
Jamiras, line 59: ??
image

It doesn't appear to be the pipe that makes it angry

Just trying to save the record with a forum_topic_id.

resources/js/features/game-list/components/HubMainRoot/HubHeading/HubHeading.test.tsx
```diff
@@ -83,4 +83,32 @@ describe('Component: HubHeading', () => {
...
+
+    render<App.Platform.Data.HubPageProps>(<HubHeading />, {
+      pageProps: { hub, can: { manageGameSets: true } },
+    });
+
+    // ASSERT
+    expect(screen.queryByRole('link', { name: /view forum topic/i })).not.toBeInTheDocument();
+  });
+
+  it('given the user can manage hubs and there is an associated forum topic, shows the View Forum Topic button', () => {
```
Jamiras, line 99: I'm having trouble identifying where the code is for this comment:
> visible to those who can manage the hub only

But, is that what we really want? If a hub has a forum topic, presumably discussion is allowed, so why restrict it to moderators? 

"Console Wars I - Developed Sets" forum topic has 145 pages of comments, but it's also the same forum topic pointed to by the "Console Wars" event, so that's probably a mistake.

Also, there's nothing preventing users from accessing the forum topic if they have a direct URL.

wescopeland: > I'm having trouble identifying where the code is for this comment

The permission is being sent down to the UI via `UserPermissionsData`:
```php
manageGameSets: Lazy::create(fn () => $user ? $user->can('manage', \App\Models\GameSet::class) : false),
```

And then in _HubHeading.tsx_, we have:
```tsx
{can.manageGameSets ? (
  <div className="flex gap-2">
    {hub.forumTopicId ? (
      <a
        href={`/viewtopic.php?t=${hub.forumTopicId}`}
        className={baseButtonVariants({
          size: 'sm',
          className: 'gap-1',
        })}
      >
        <LuMessageCircle className="size-4" />
        {t('View Forum Topic')}
      </a>
    ) : null}
```

> But, is that what we really want? If a hub has a forum topic, presumably discussion is allowed, so why restrict it to moderators?

I suspect what we really want is role-based access control (RBAC) for forum topics and hubs, which should be handled by subsequent PRs.

Some meta hubs should only be editable for a specific team. The initial underpinnings of setting up this RBAC stuff is handled in https://github.com/RetroAchievements/RAWeb/pull/2972.

Additionally, some forum topics should probably only be accessible to users with certain roles.

However, that does beg the question of whether or not forum topics should be visible if one is attached to the hub and has no RBAC associated with it. Given these topics are already publicly visible, I've updated latest to surface these URLs. The URLs currently do not currently check legacy `Permissions` for visibility.

app/Filament/Resources/HubResource.php
wescopeland: In latest, I've lifted the pattern from _NewsCard.tsx_ into the `NoEmoji()` rule and added unit tests for the rule.

Jamiras (approved):

PR #2970, opened 2024-12-22
ci: enforce vitest coverage thresholds

wescopeland:
At the moment, the React UI sits at 100% statement, branch, function, and line coverage. With over 1,100 unique test cases, there are a lot of examples to look to. As the UI continues to scale, we should probably shoot to keep the coverage level absurdly high.

This PR sets coverage thresholds to:
* Lines - 98.5%
* Functions - 100%
* Branches - 98.5%
* Statements - 98.5%

Personally, I plan on keeping everything at 100% coverage for all commits going forward.


Jamiras (approved):
I think this is a lofty goal that may impede new contributors, but if we can manage to maintain it, that would be great.

wescopeland:
> may impede new contributors

I considered this as well. Over time it may be appropriate to lower the threshold, but given how much React code is in RAWeb now, 98.5% is probably appropriate for the time being.

When we approve code, we're not only saying "yes, this is fine to ship", but we're accepting responsibility for maintaining it over the long term, even if the new contributor disappears.

After doing a bunch of these site releases, I must confess I'm not happy to maintain untested code anymore. I feel way more at ease when whatever we're deploying has a high level of coverage. At least then when there's something to hotfix, I know the odds of causing a regression are fairly low.

PR #2971, opened 2024-12-22
fix(TrendingRightNow): resolve css issue when game title is long

wescopeland:
Resolves a CSS issue that can occur on the home page when the game title in the trending section is very long.

**Before**
Screenshot 2024-12-22 at 4 57 36 PM

Screenshot 2024-12-22 at 4 50 47 PM

**After**
Screenshot 2024-12-22 at 4 57 05 PM

Screenshot 2024-12-22 at 4 56 57 PM


resources/js/features/home/components/+root/TrendingRightNow/TrendingRightNow.tsx
```diff
@@ -20,9 +20,15 @@ export const TrendingRightNow: FC = () => {
...
           <div key={`trending-${trendingGame.game.id}`} className="rounded-lg bg-embed p-2">
-            <div className="flex w-full items-end justify-between">
-              <GameAvatar {...trendingGame.game} size={40} showSystemChip={true} />
-              <p className="text-2xs">
+            <div className="relative flex w-full items-end justify-between">
+              <GameAvatar
+                {...trendingGame.game}
+                size={40}
+                showSystemChip={true}
+                gameTitleClassName="line-clamp-1"
```
Jamiras, line 28: Not sure how it determines where to do the truncation, but I find it weird that it'll turn the last two characters of a word into "...", but one more is too much and it just nukes the word entirely.
image
image

Jamiras (approved):

resources/js/features/home/components/+root/TrendingRightNow/TrendingRightNow.tsx
wescopeland: For better or worse, this one is at the mercy of the browser vendor's CSS implementation. One benefit of doing it with a `line-clamp-*` classname though is we don't have to pull in a dependency to do the truncation based on the div width.

PR #2972, opened 2024-12-23
refactor: split Staff Dev role into separate roles, add CM vanity role

wescopeland:
This PR repurposes the currently-unused Staff Developer role, splitting it into the three staff dev roles that have emerged since the original permissions matrix was envisioned. Those roles are:

* Quality Assurance
* Developer Compliance
* Code Reviewer

Each of these three roles has distinct responsibilities, and should ultimately have distinct abilities. There is enough nuance in their day-to-day that this split should occur.

Also, the way Staff Developer was written implied that a user would be promoted from Developer to Staff Developer. However, in practice, QAM/DevC/CR are additive to Developer, giving that Developer additional privileges. A full developer may have 1 to N of these three roles.

Additionally, a new vanity role has been added: Community Manager. This can be attached to users, but it should not have any abilities. A Community Manager should inherit abilities from other roles like Moderator, etc.

---

SQL commands to run:
```sql
-- 1: Detach developer-staff from any users.
DELETE FROM auth_model_roles
WHERE role_id = (SELECT id FROM auth_roles WHERE name = 'developer-staff');

-- 2: Remove developer-staff from the roles table.
DELETE FROM auth_roles WHERE name = 'developer-staff';

-- 3: Insert new roles.
INSERT INTO auth_roles (name, display, guard_name, created_at, updated_at) VALUES
    ('dev-compliance', 4, 'web', NOW(), NOW()),
    ('quality-assurance', 4, 'web', NOW(), NOW()),
    ('code-reviewer', 4, 'web', NOW(), NOW()),
    ('community-manager', 3, 'web', NOW(), NOW());
```

---

Things to note regarding this PR:
* There should not be any behavior changes for current users.
* When a developer is demoted, any staff developer roles they have should be detached.
* This is ultimately leading to users being able to select their visible role, which is currently a read-only field on the Settings page.


app/Models/Role.php
```diff
@@ -145,7 +160,7 @@ public static function boot()
             if ($relationName === 'users') {
                 foreach ($pivotIds as $pivotId) {
                     $user = User::find($pivotId);
-                    activity()->causedBy(auth()->user())->performedOn($user)
+                    activity()->causedBy($user)->performedOn($user)
```
Jamiras, line 163: `$user` is overloaded here. it's already defined on line 162

app/Policies/CommentPolicy.php
```diff
@@ -47,19 +47,6 @@ public function viewAny(?User $user, Model $commentable): bool
...
-            if ($commentable !== null && $commentable instanceof \App\Models\Ticket) {
-                $commentable->loadMissing(['achievement.developer']);
-
-                $didAuthorAchievement = $commentable->achievement->developer->id === $user->id;
-
-                return
-                    $didAuthorAchievement
-                    && $commentable->is_open
-                    && $user->hasAnyRole([Role::DEVELOPER_STAFF, Role::DEVELOPER]);
-            }
```
Jamiras, line 61: This was removed in another PR where I suggested it was still needed. I realize there will be conflicts, but you should only remove the DEVELOPER_STAFF check at this time.

app/Filament/Resources/UserResource/Pages/Roles.php
```diff
@@ -123,4 +118,17 @@ public function table(Table $table): Table
...
+    {
+        if ($targetUser->hasRole(Role::QUALITY_ASSURANCE)) {
+            $targetUser->removeRole(Role::QUALITY_ASSURANCE);
+        }
+        if ($targetUser->hasRole(Role::DEV_COMPLIANCE)) {
+            $targetUser->removeRole(Role::DEV_COMPLIANCE);
+        }
+        if ($targetUser->hasRole(Role::CODE_REVIEWER)) {
+            $targetUser->removeRole(Role::CODE_REVIEWER);
+        }
```
Jamiras, line 132: These roles can be assigned to non-developers.

app/Policies/UserPolicy.php
```diff
@@ -16,7 +16,19 @@ class UserPolicy
...
+            Role::ROOT,
+            Role::ADMINISTRATOR,
+
+            // moderation
+            Role::MODERATOR,
+
+            // staff developers
+            Role::CODE_REVIEWER,
+            Role::DEV_COMPLIANCE,
+            Role::QUALITY_ASSURANCE,
```
Jamiras, line 30: What does "manage user" policy give to these roles? I can't seem to add or remove roles as a Developer with a QA role. It does seem to open up the Moderate tools on the non-admin user page, which doesn't seem desirable for these roles.

app/Policies/CommentPolicy.php
wescopeland: Per the discussion mentioned here: https://github.com/RetroAchievements/RAWeb/pull/2967#discussion_r1899637199 I'm thinking maybe we should keep this code all deleted in this branch.

app/Models/Role.php
wescopeland: Should be fixed in latest. I also added some safeguards to prevent server errors in case these code paths are entered from side effects (ie: a command, Horizon job, etc).

app/Policies/UserPolicy.php
wescopeland: This is cruft from a change that existed earlier in the branch that I ultimately decided against and removed (a promote/demote button on the manage user infolist). I've reverted this policy change in latest.

app/Filament/Resources/UserResource/Pages/Roles.php
wescopeland: As of latest, the staff dev roles can no longer be attached to a user unless they have the developer role first (may require a refresh when attaching the full dev role to a user).

app/Models/Role.php
Jamiras: I see you removed the assignment of `$user=Auth::user()`, but this statement previously was `causedBy(auth()->user())` and now it's `causedBy(User::find($pivotId))`.

The same change occurs a few lines up (151).

wescopeland: Ah, I see what's happening now. I got confused.

I try to replace those `auth()->user()` calls with `Auth::user()` -- for some reason my IDE yells about `auth()->user()`. I think we should be in a good place now with changes in latest.

Jamiras (approved):

PR #2973, opened 2024-12-23
perf: reduce excessive re-rendering

wescopeland:
No behavior changes.

After running the Inertia.js stuff through [a popular benchmarking tool](https://million.dev/), I observed that React is unnecessarily re-rendering nearly all of our page components, in some cases resulting in ~600ms of unnecessary UI processing time.

This PR mitigates these expensive re-renders by wrapping all root-level page components in [`memo()`, a built-in React helper.](https://react.dev/reference/react/memo)

I generally don't like reaching for `memo()` unless being yelled at by a tool, otherwise it's too easy to accidentally balloon the client's memory usage.

Additionally, an expensive CSS calculation on `BaseButton` is now optimized via the native `useMemo()` hook. Some base components were also being reported as having expensive re-renders, and they are now also wrapped with `memo()`.

**How to test this PR:**
Click around in our various UIs. Especially try to do interactions (stuff on datatables, paginating through things, etc). You should observe no new bugs.


Jamiras (approved):
Seems functional. Did not attempt to confirm performance changes.

PR #2974, opened 2024-12-23
refactor(head-analytics): migrate php code to an action, add coverage

wescopeland:
This PR migrates the URL path sanitization and props extraction code in _head-analytics.blade.php_ to a reusable action that is also covered by tests.

Currently, there is an issue in Plausible where our self-healing system URLs are not being sanitized correctly:
Screenshot 2024-12-23 at 1 25 35 PM

This will be a recurring problem going forward as more URLs use slugs. The changes in this PR address the issue:
Screenshot 2024-12-23 at 1 26 25 PM

**How to test this PR**
Open your network tab and navigate to various pages. Check the "event" call. The `p` and `u` parameters are the ones of interest.


app/Actions/ProcessPlausibleUrlAction.php
```diff
@@ -0,0 +1,173 @@
...
+
+        // Routes that just use a string parameter.
+        $this->addStringRoute('user', 'username');
+
+        // Routes that just use an ID.
+        $this->addIdRoute('ticket');
+
+        // Routes with nested ID-slug segments.
+        // TODO $this->addNestedRoute('forums', ['category', 'forum']);
+    }
```
wescopeland, line 47: I tried to make this as simple as possible for adding new routes. Nested routes are not yet implemented, but some TODOs are in place.

Ideally a developer should not need to write regex to add model-bound routes. It should be a fairly trivial thing.

resources/views/components/head-analytics.blade.php
```diff
@@ -10,57 +10,13 @@
...
-            break;
-        }
-    }
+@php
+    $result = (new ProcessPlausibleUrlAction())->execute(request()->path());
 
-    // Track what topic ID users are viewing at viewtopic.php.
-    if (strpos($url, 'viewtopic') !== false && isset($queryParams['t'])) {
-        $props['topicId'] = $queryParams['t'];
-    }
```
Jamiras, line 63: Looks like `viewtopic` is not handled.

wescopeland: Looks like viewtopic was not handled and neither were the default props. Should all be fixed in latest & tests have been updated.

I added this block for legacy routes:
```php
// Legacy routes that need special handling.
$this->addLegacyRoute('leaderboardinfo.php', ['i' => 'id']);
$this->addLegacyRoute('viewtopic.php', ['t' => 'topicId']);
$this->addLegacyRoute('viewforum.php', ['f' => 'forumId']);
$this->addLegacyRoute('forum.php', ['c' => 'categoryId']);
```
We have a few more that take in query params, but all the ones I can think of are pretty low volume. Happy to add anything else here if you'd like to track it, though.

Jamiras (approved):

PR #2977, opened 2024-12-24
fix(connect): add User type to allprogress query

drisc:
Reported by Stenzek, who would like to use it to get a list of all progress on Playstation games in Duckstation, `allprogress` in the connect api is currently broken due to a type error.

Current Error:
```GetAllUserProgress(): Argument #1 ($user) must be of type App\\Models\\User, string given, called in dorequest.php on line 239```

Fixed response (from my local instance):
```json
{
    "Success": true,
    "Response": {
        "1": {
            "NumAch": 8,
            "Earned": "8",
            "HCEarned": "5"
        },
        "2": {
            "NumAch": 9,
            "Earned": "2",
            "HCEarned": "0"
        },
        "3": {
            "NumAch": 10,
            "Earned": "0",
            "HCEarned": "0"
        }
    }
}
```


wescopeland (changes requested):
This endpoint seems to have been offline for roughly 14 months. I'm happy to support it, but if we do then it needs to be covered by tests.

drisc:
Closing as a better implementation can be found in #3062

PR #2978, opened 2024-12-24
feat: add GameSuggestionEngine class

wescopeland:
This PR lays the groundwork for migrating the two game suggestion pages to React. It focuses purely on implementing the back-end logic.

**How to test this PR:**
Navigate to both of these pages and refresh a bunch of times:
```
http://localhost:64000/game/1/suggestions
http://localhost:64000/games/suggestions
```

You'll see pretty-printed JSON output for the table results:
Screenshot 2024-12-24 at 2 46 44 PM

**How it works:**
The core of this implementation is a new `GameSuggestionEngine` class, which uses a weighted strategy pattern. Each strategy represents a different way to suggest games to users:

```php
// Example: Set strategies for game-specific suggestions.
if ($this->sourceGame) {
    $this->strategies = [
        [new Strategies\SimilarGameStrategy($this->sourceGame), 50],
        [new Strategies\SharedHubStrategy($this->sourceGame), 20],
        [new Strategies\CommonPlayersStrategy($this->user, $this->sourceGame), 20],
        [new Strategies\SharedAuthorStrategy($this->sourceGame), 10],
    ];
}
```
Notice that the array values are like `[strategy, weight]`.

Each suggestion strategy implements a `GameSuggestionStrategy` interface:
```php
interface GameSuggestionStrategy
{
    public function select(): ?Game; // what is being suggested?
    public function reason(): GameSuggestionReason; // why is it being suggested?
    public function reasonContext(): ?GameSuggestionContextData; // is there more to know about the suggestion?
}
```

This makes it easy to:
* Add new suggestion types.
* Adjust suggestion weights.
* Test each strategy independently.

Here's an example of an implemented strategy to fetch a suggested game from the user's Want to Play Games list:
```php
class WantToPlayStrategy implements GameSuggestionStrategy
{
    public function __construct(
        private readonly User $user
    ) {
    }

    public function select(): ?Game
    {
        return $this->user->gameListEntries()
            ->whereType(UserGameListType::Play)
            ->whereHas('game', function ($query) {
                $query->whereHasPublishedAchievements();
            })
            ->with('game')
            ->inRandomOrder()
            ->first()
            ?->game;
    }

    public function reason(): GameSuggestionReason
    {
        return GameSuggestionReason::WantToPlay;
    }

    public function reasonContext(): ?GameSuggestionContextData
    {
        return null;
    }
}
```

```php
public function testItSelectsGameFromWantToPlayList(): void
{
    // Arrange
    $user = User::factory()->create();
    $game1 = Game::factory()->create(['Title' => 'Game 1', 'achievements_published' => 6]);
    $game2 = Game::factory()->create(['Title' => 'Game 2', 'achievements_published' => 6]);

    $addGameToListAction = new AddGameToListAction();
    $addGameToListAction->execute($user, $game1, UserGameListType::Play);
    $addGameToListAction->execute($user, $game2, UserGameListType::Play);

    // Act
    $strategy = new WantToPlayStrategy($user);
    $result = $strategy->select();

    // Assert
    $this->assertTrue(in_array($result->id, [$game1->id, $game2->id]));
    $this->assertEquals(GameSuggestionReason::WantToPlay, $strategy->reason());
    $this->assertNull($strategy->reasonContext());
}
```

All of the strategies are covered by tests.

---

Next steps:
* Have the React datatable support a "client-only" mode, where if <N results are returned, all table operations are handled client-side in browser memory.
* Build out the React versions of these two pages, remove the associated PHP UIs.


app/Platform/Services/GameSuggestions/Strategies/RevisedGameStrategy.php
```diff
@@ -0,0 +1,50 @@
...
+        return Game::query()
+            ->whereHas('playerGames', function (Builder $query) {
+                $query->whereUserId($this->user->id)->whereNotAllAchievementsUnlocked();
+            })
+            ->whereHas('playerBadges', function ($query) {
+                $query->whereUserId($this->user->id)->where('AwardType', AwardType::Mastery);
+            })
+            ->whereHasPublishedAchievements()
+            ->inRandomOrder()
+            ->first();
```
wescopeland, line 38: In certain places, even though it's benign, I've used `query()` to make the line indention of the first function (ie: `whereHas()` more readable:

```php
// With query()
return Game::query()
  ->whereHas("playerGames", function (Builder $query) {
    $query->whereUserId($this->user->id)->whereNotAllAchievementsUnlocked();
  })
  ->whereHas("playerBadges", function ($query) {
    $query
      ->whereUserId($this->user->id)
      ->where("AwardType", AwardType::Mastery);
  })
  ->whereHasPublishedAchievements()
  ->inRandomOrder()
  ->first();

// Without query()
return Game::whereHas("playerGames", function (Builder $query) {
  $query->whereUserId($this->user->id)->whereNotAllAchievementsUnlocked();
})
  ->whereHas("playerBadges", function ($query) {
    $query
      ->whereUserId($this->user->id)
      ->where("AwardType", AwardType::Mastery);
  })
  ->whereHasPublishedAchievements()
  ->inRandomOrder()
  ->first();
```

app/Platform/Actions/BuildGameListAction.php
```diff
@@ -181,4 +172,105 @@ private function getUnfilteredResultsCount(Builder $query, array $filters): ?int
...
+
+            // Game suggestions have contextual data attached about why
+            // that particular game was suggested to the user.
+            if ($this->isSuggestionListType($listType)) {
+                $suggestionData = collect($this->suggestions)
+                    ->first(fn ($suggestion) => $suggestion->gameId === $game->id);
+
+                $suggestionEntryData = $baseData + [
+                    'suggestionReason' => $suggestionData->reason,
+                    'relatedGameContext' => $suggestionData->context,
```
Jamiras, line 267: I understand the desire for this on the full game list, but when there's a source game, should the source game be duplicated for every similar game entry?
image

app/Platform/Services/GameSuggestions/Strategies/SimilarGameStrategy.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        $this->selectedGame = Game::query()
+            ->whereHas('gameSets', function ($query) use ($gameSetIds) {
+                $query->whereIn('game_sets.id', $gameSetIds);
+            })
+            ->where('ID', '!=', $this->sourceGame->id)
+            ->whereHasPublishedAchievements()
+            ->inRandomOrder()
+            ->first();
```
Jamiras, line 40: Not sure how these are similar
image

image

It is an interesting choice though, as I could see them being related, but the UI doesn't reflect that.

They do both share a common related game (Boxxle).
Carrier (17588) and Pile Wonder (10793) were both also suggested as a similar game and are linked to Boxxle.

app/Platform/Services/GameSuggestions/Strategies/CommonPlayersStrategy.php
```diff
@@ -0,0 +1,103 @@
...
+            ->orderByRaw('COUNT(*) DESC')
+            ->limit(10)
+            ->value('game_id');
+
+        return $gameId ? Game::find($gameId) : null;
+    }
+
+    public function reason(): GameSuggestionReason
+    {
+        return GameSuggestionReason::CommonPlayers;
```
Jamiras, line 50: Missing relatedGameContext for "all games suggestion" results

image

app/Platform/Actions/BuildGameListAction.php
wescopeland: Probably not - should be removed from this case in latest.

app/Platform/Services/GameSuggestions/Strategies/SimilarGameStrategy.php
wescopeland: Looks like I messed up the query. I think it was pulling games from similar games 1 layer deep (similar games of similar games), which would explain a possible adjacent relationship. This should be adjusted in latest.

app/Platform/Services/GameSuggestions/Strategies/CommonPlayersStrategy.php
wescopeland: Fixed in latest.

app/Platform/Services/GameSuggestions/GameSuggestionEngine.php
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * @return array<GameSuggestionData>
+     */
+    public function selectSuggestions(int $limit = 10): array
+    {
+        // We won't show the user games they already have 100% completion for.
+        $masteredGameIds = PlayerGame::whereUserId($this->user->id)
+            ->whereAllAchievementsUnlocked()
+            ->pluck('game_id')
+            ->toArray();
```
wescopeland, line 88: I think I missed this in the original commit. I noticed when refreshing constantly in prod, I was never suggested a mastered game. This new logic should filter those out.

Jamiras (approved):

PR #2983, opened 2024-12-26
feat(multiset): handle `r=ping` correctly

wescopeland:
This PR modifies how game IDs are resolved between the `patch` and `ping` routines to ensure consistent behavior when multiset is enabled.

Key changes:
* A new action, `ResolveRootGameIdAction`, has been added. This action encapsulates game ID resolution logic for both endpoints.
* `BuildClientPatchDataAction` has been modified to use this action. Tests are unchanged and all remain passing.
* In _dorequest.php_, `ping` has been modified to use this action for game ID resolution if multiset is enabled.

A reminder of how game ID resolution works:
* For legacy clients, the loaded hash's direct game ID is used.
* When multiset is disabled, the loaded hash's direct game ID is used.
* When multiset is enabled:
  * For bonus set game hashes, resolve to the core/base game ID.
  * For specialty/exclusive set game hashes, maintain the subset game ID.
  * If the user has globally opted out of subsets, use the loaded hash's direct game ID.

The changes in this PR ensure that:
* The active game ID in `ping` matches the root-level game ID from `patch`.
* Legacy clients and users who have opted out of multiset maintain the existing behavior.

It's worth noting the changes in `ping` are controlled entirely by the `VITE_FEATURE_MULTISET` feature flag, as clients are already sending both the game ID (`g`) and the game hash MD5 (`x`).

---

Next steps:
* Handle `r=startsession` correctly.


public/dorequest.php
```diff
@@ -368,6 +369,12 @@ function LogDeprecatedUserAgent(string $requestType, ClientSupportLevel $clientS
                 }
             }
 
+            // If multiset is enabled, resolve the root game ID.
+            if (config('feature.enable_multiset')) {
+                $rootGameId = (new ResolveRootGameIdAction())->execute($gameHash, $game, $user);
```
Jamiras, line 374: Note that `gameHash` will be null for multidisc games (see line 368). We do this so we don't spawn a new session when updating the rich presence string.

The client will send the hash of the currently loaded disc. Presumably, all discs would map to the same root game ID.

wescopeland: I've updated a comment and added another test case. Beyond that, does anything logically need to change? Want to make sure I'm not misinterpreting.

Jamiras: TBH, I hadn't really dug into the code to figure out how, or even if, you were handling this case. I just wanted to make sure you knew about it.

I would expect the provided game id to be sufficiently accurate.

Jamiras (approved):

PR #2986, opened 2024-12-26
feat: improve 'Claimed' column UX in datatable

wescopeland:
This PR makes an improvement to the "Claimed" column and is a soft preparation for migrating the various claims table pages to React, namely with the new component shown in the screenshots below.

**Before**
Screenshot 2024-12-26 at 4 39 49 PM

**After**
Screenshot 2024-12-26 at 4 40 03 PM

The game list query now returns the active claimants for the game. In the column itself, a new component called `UserAvatarStack` is used.

As the name implies, this component stacks user avatars:
Screenshot 2024-12-26 at 4 42 09 PM

Screenshot 2024-12-26 at 4 42 31 PM

The column's sort order logic is unchanged: it's a simple yes or no. Mobile is unaffected by these changes.


app/Platform/Concerns/BuildsGameListQueries.php
```diff
@@ -34,7 +34,16 @@ private function buildBaseQuery(
         ?User $user = null,
         ?int $targetId = null,
     ): Builder {
-        $query = Game::with(['system'])
+        $query = Game::query()
```
wescopeland, line 37: `query()` is just here to move `with()` to the next line so line indentation makes sense.

resources/js/features/game-list/utils/column-definitions/buildHasActiveOrInReviewClaimsColumnDef.tsx
```diff
@@ -47,18 +44,22 @@ export function buildHasActiveOrInReviewClaimsColumnDef({
...
         <div>
-          {hasActiveOrInReviewClaims ? (
-            <BaseTooltip>
-              <BaseTooltipTrigger asChild>
-                <p>{strings.t_yes}</p>
-              </BaseTooltipTrigger>
+          {hasActiveOrInReviewClaims && row.original.game.claimants ? (
+            <div className="flex items-center gap-1.5">
+              <LuCheck className="size-4" />
+              <span className="sr-only">{strings.t_yes}</span>
```
Jamiras, line 50: I find the X misleading, especially right next to the add-to-list column
image

Similarly, the check mark is redundant. Either there will be one or more avatars, or there will be zero avatars. Do we really need the X/Check too?

wescopeland: We can get rid of the check and X. Though, I do like that all of our table cells have an empty state of some kind. To maintain consistency, I've reverted the empty state for this one back to what it was previously:

 
Screenshot 2024-12-31 at 8 46 18 PM

Jamiras (approved):

PR #2993, opened 2024-12-27
fix(leaderboard): change ->authorUser to ->developer

wescopeland:
No idea how this is working in prod.

Locally it doesn't appear to be possible to edit a leaderboard without RALibretro throwing an error. `->authorUser` was renamed to `->developer` and this line of code wasn't changed.

In prod, changes are made but `$originalAuthor` appears to not be set, which results in server comments not being recorded. I cannot explain why prod exhibits a different behavior from local development.

Screenshot 2024-12-27 at 3 35 52 PM


Jamiras:
> No idea how this is working in prod.

It's not, at least not on staging. I'm guessing most users don't update leaderboards after they've been published.

Jamiras (approved):

PR #3001, opened 2024-12-29
fix(GameListItems): prevent over-prefetching

wescopeland:
This PR fixes a bug with mobile game lists related to optimistic prefetching.

When approaching the "Load more" button, the UI prefetches the next page. This has a benefit of making the "Load more" tap feel instantaneous and also improves a core web vital that Google monitors.

However, the UI does not track if the user actually tapped the button. This can lead to an edge case where the UI actually overfetches for data if the user follows these steps:
* Scroll down to the Load More button.
* Do not tap the button.
* Scroll up about 100px.
* Scroll back down - page N+2 is fetched.

https://github.com/user-attachments/assets/488ba2f2-434d-4e1d-8bd7-07aafd9aa3bb


Jamiras (approved):

Jamiras:
Presumably unrelated to this PR, but if the "Load More" button is at the top of the screen, pressing it loads the new items above the "Load More" button, instead of pushing the "Load More" button down so you can see the new items.

wescopeland:
Indeed unrelated - that is the browser trying to be smart and pin the scroll position. I believe the behavior also varies between Chromium and Safari. I'll need to think carefully through if there's something we can do to remediate that.

PR #3004, opened 2024-12-31
include AotM achievements in AotW event

Jamiras:
https://discord.com/channels/310192285306454017/1305921595843285054/1323451219905085480


wescopeland (approved):
I agree with your assessment that relying on the date range subtext is less than ideal for determining if an achievement in the event is AotM instead of AotW.

Similar to how hubs are being extracted from games, we may want to think about what "event games" extracted from games might look like and if that's a route we should pursue.

I can't help but feel that perhaps overloading games with events logic will inevitably come back to bite us (and given that, I don't think we should adjust the subtext treatment for AotW stuff at this time).

PR #3005, opened 2024-12-31
use BuildAchievementOfTheWeekDataAction to eliminate duplicate code

Jamiras:
Ensures the achievement of the week API uses the same logic for determining the active event achievement as the home page widget.


public/API/API_GetAchievementOfTheWeek.php
```diff
@@ -47,66 +49,52 @@
...
  */
 
-$achievementOfTheWeek = EventAchievement::active()
-    ->whereHas('achievement.game', function ($query) {
-        $query->where('Title', 'like', '%of the week%');
-    })
-    ->with('achievement.game')
-    ->first();
+$staticData = StaticData::first();
+$data = (new BuildAchievementOfTheWeekDataAction())->execute($staticData);
```
wescopeland, line 53: nit: Recommend renaming to `$aotwData` or something else more descriptive. In my first read of the code, I had already forgotten what `$data` was when I landed on line 91 (obviously, all variables are data):
```php
'ID' => $data->forumTopicId->resolve() ?? null,
```

wescopeland (approved):

PR #3011, opened 2025-01-01
feat(hash): surface source URLs on the Supported Game Files page

wescopeland:
Implements https://github.com/RetroAchievements/RAWeb/discussions/2712.

If a hash has **both** an RAPatches URL and a Resource Page URL, the user is recommended to visit the resource page URL. The RAPatches URL is _always_ shown as a mirror (we can't constantly check the source URL to see if it's still online).

Screenshot 2025-01-01 at 4 21 49 PM

### DIFF
--- a/resources/js/features/games/components/HashesMainRoot/HashesList/HashesListItem.tsx
+++ b/resources/js/features/games/components/HashesMainRoot/HashesList/HashesListItem.tsx
@@ -2,6 +2,7 @@ import type { FC } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { buildTrackingClassNames } from '@/common/utils/buildTrackingClassNames';
+import { cn } from '@/common/utils/cn';
 
 interface HashListingProps {
   hash: App.Platform.Data.GameHash;
@@ -27,7 +28,27 @@ export const HashesListItem: FC<HashListingProps> = ({ hash }) => {
       <div className="flex flex-col border-l-2 border-neutral-700 pl-2 light:border-embed-highlight black:border-neutral-700">
         <p className="font-mono text-neutral-200 light:text-neutral-700">{hash.md5}</p>
 
-        {hash.patchUrl ? (
+        {/* Can show RAPatches as the mirror */}
+        {hash.source && hash.patchUrl ? (
+          <div className="mt-1 flex flex-col">
+            <a
+              href={hash.source}
+              className={cn(buildTrackingClassNames('Open Patch Source URL', { md5: hash.md5 }))}
+            >
+              {t('Download from Original Source (Recommended)')}
+            </a>
+
+            <a
+              href={hash.patchUrl}
+              className={cn(buildTrackingClassNames('Download Patch File', { md5: hash.md5 }))}
+            >
+              {t('Mirror')}
+            </a>
+          </div>
+        ) : null}
+
+        {/* Show RAPatches as the direct download link */}
+        {!hash.source && hash.patchUrl ? (
           <a
             href={hash.patchUrl}
             className={buildTrackingClassNames('Download Patch File', { md5: hash.md5 })}
--- a/app/Platform/Data/GameHashData.php
+++ b/app/Platform/Data/GameHashData.php
@@ -20,6 +20,8 @@ class GameHashData extends Data
         #[TypeScriptType('App\\Platform\\Data\\GameHashLabelData[]')]
         public array $labels,
         public ?string $patchUrl,
+        /** "Resource Page URL" */
+        public ?string $source,
     ) {
     }
 
@@ -31,6 +33,7 @@ class GameHashData extends Data
             name: $gameHash->name,
             labels: GameHashLabelData::fromLabelsString($gameHash->labels),
             patchUrl: $gameHash->patch_url,
+            source: $gameHash->source,
         );
     }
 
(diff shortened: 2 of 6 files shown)


resources/js/features/games/components/HashesMainRoot/HashesList/HashesListItem.tsx
```diff
@@ -27,7 +28,27 @@ export const HashesListItem: FC<HashListingProps> = ({ hash }) => {
...
-        {hash.patchUrl ? (
+        {/* Can show RAPatches as the mirror */}
+        {hash.source && hash.patchUrl ? (
+          <div className="mt-1 flex flex-col">
+            <a
+              href={hash.source}
+              className={cn(buildTrackingClassNames('Open Patch Source URL', { md5: hash.md5 }))}
+            >
+              {t('Download from Original Source (Recommended)')}
+            </a>
```
Jamiras, line 39: What if the original source gets a newer version? 

I thought we had our own repository in case the original went away _or got updated_.

wescopeland: That's a fair point. I don't know if we can have a good solution if and when that happens. @televandalist do you have any thoughts?

cIymax: Just chiming in with my two cents from the perspective of a hack author.  Better visibility of the hack source, if anything, can lead players to more easily identify and ask for a newer version to be linked on the RA side or for a legacy RA version to be rehosted at the hack source at least until the former happens.

If the hack author is active, the hack author should be able to rehost a legacy RA version if the hack author is willing.  If the hack author is no longer active, this means that the hack versioning has stabilized, and RA is more likely to be able to catch up in linked version or to have the stabilized version already linked.

I realize that devs are under no obligation to catch the set up in any timeframe (the same applies to hack authors rehosting legacy versions), but players on RA do tend to want to be able to play the latest version.

How about we label the hack source as "(Recommended; check the hack's version in case of updates)" and add a tooltip that says, "If there is a version mismatch, use the mirror link in the meantime and consider contacting the hack author or replying to the RA forum topic to get the versions to match."?

wescopeland: I'm not a fan of adding this to every hack link that has a resource URL.

Ideally, hack authors and RA developers collaborate on this sort of thing. I don't want RA users taking it upon themselves to reach out to hack authors to remediate version mismatches on RA. We have seen how RA users reaching out to emulator developers has historically gone.

I would like @televandalist to weigh in on this comment thread before I take any further action.

televandalist: > Better visibility of the hack source, if anything, can lead players to more easily identify and ask for a newer version to be linked on the RA side or for a legacy RA version to be rehosted at the hack source at least until the former happens.

I really don't think that this is something that can be universally applied. It ultimately falls on the set developer(s) to figure out and handle as they see fit. Some newer versions of hacks, translations, and even homebrews are intentionally ignored and some aren't linked simply because a set lacks an active dev/maintainer. While it's still new, we should encourage [compatibility testing](https://docs.retroachievements.org/guidelines/content/player-compatibility-testing.html#player-compatibility-testing) over flooding the forums, walls, and/or DMs with requests to update hashes. Most players aren't aware of what all goes into set development; which compatibility testing makes irrelevant.

***

> If the hack author is active, the hack author should be able to rehost a legacy RA version if the hack author is willing. If the hack author is no longer active, this means that the hack versioning has stabilized, and RA is more likely to be able to catch up in linked version or to have the stabilized version already linked.

I believe this would further complicate the entire process and adds way too much reliance on certain people not having a bad day. I have more than enough on my plate so adding "check if hack author is still active and if versioning has stabilized" to each RAPatches upload is not something I'd be willing to do. My interest in all of this starts and ends with whether hashes are accounted for, verified, and can be reproduced by someone without them pulling their hair out. I also believe that if having a patch in the repo is problematic to the author, then maybe the hash shouldn't even be linked.

***

> I realize that devs are under no obligation to catch the set up in any timeframe (the same applies to hack authors rehosting legacy versions), but players on RA do tend to want to be able to play the latest version.

Again, it's not something that's universal and players can check out [compatibility testing](https://docs.retroachievements.org/guidelines/content/player-compatibility-testing.html#player-compatibility-testing). 

***

> How about we label the hack source as "(Recommended; check the hack's version in case of updates)" and add a tooltip that says, "If there is a version mismatch, use the mirror link in the meantime and consider contacting the hack author or replying to the RA forum topic to get the versions to match."?

> I'm not a fan of adding this to every hack link that has a resource URL.

I think it would be better to just add more info to this section:
Capture

Maybe something like this:

"This page shows which ROM hashes are compatible with this entry. Additional information regarding these hashes may be listed in the [entry's official forum topic](https://retroachievements.org/viewtopic.php?t=448). Details on how hashes are generated for each system can be found [here](https://docs.retroachievements.org/developer-docs/game-identification.html).

Please contact active set developer(s) regarding any errors. If there is not an active developer associated with the entry, then [contact QATeam](https://retroachievements.org/messages/create?to=QATeam).

Information on contributing with compatibility testing can be found [here](https://docs.retroachievements.org/guidelines/content/player-compatibility-testing.html#player-compatibility-testing)."

televandalist: > What if the original source gets a newer version?
> 
> I thought we had our own repository in case the original went away _or got updated_.

> That's a fair point. I don't know if we can have a good solution if and when that happens. @televandalist do you have any thoughts?

Not really since the info is tied to individual hashes. Adding a note that patch sources may not be entirely up-to-date seems like it would defeat the purpose 😆

wescopeland:
I've grown convinced this change is not the best course of action for RA as a platform. It's increasingly unclear to me how this change leads to a better experience for our players.

PR #3012, opened 2025-01-01
fix(comment): respect single newlines

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1323464515400105994.

Single newlines are currently not being respected in comments.

**Before**
Screenshot 2025-01-01 at 6 55 14 PM

**After**
Screenshot 2025-01-01 at 6 55 07 PM

To repro the current issue, just try a comment input like:

```
- one
- two
- three
- four
```


app/Community/Actions/FormatLegacyCommentPayloadAction.php
```diff
@@ -40,14 +37,19 @@ public function execute(string $payload, bool $isTicketComment): string
...
-        return preg_replace(
-            '/<br\s*\/?>(\s*<br\s*\/?>)+/',
-            '<br /><br />',
-            nl2br($text)
-        );
+        // Convert newlines to <br /> tags without adding additional \n.
+        $text = str_replace("\n", '<br />', $text);
+
+        // Collapse multiple consecutive line breaks into exactly two.
+        $text = preg_replace('/(<br\s*\/?>\s*){2,}/', '<br /><br />', $text);
```
Jamiras, line 50: Not sure where this comes into play, but multiple line breaks seem to be preserved:
image

wescopeland: Fixed in latest.

Jamiras (approved):

PR #3013, opened 2025-01-02
fix(RecentForumPostItem): enforce max width

wescopeland:
I noticed an issue on the home page on my local env:
Screenshot 2025-01-01 at 7 37 30 PM

The width of the main content area can exceed what its bounds is supposed to be. The LG breakpoint is also broken.

It turns out this is actually being caused by a recent post on my local with the following content:

> AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

This PR enforces a maximum width on these post previews.

Note that `line-clamp-*` has no idea what to do with this input and sort of just gives up (which is probably the root cause of the problem):

Screenshot 2025-01-01 at 7 50 06 PM

Given that this is an edge case, it's probably not something to worry about.


resources/js/features/home/components/+root/RecentForumPosts/RecentForumPostItem.tsx
```diff
@@ -48,7 +48,7 @@ export const RecentForumPostItem: FC<RecentForumPostItemProps> = ({ post }) => {
         />
       </p>
 
-      <p className="line-clamp-1">{post.latestComment.body}</p>
+      <p className="line-clamp-1 lg:max-w-[580px] xl:max-w-[816px]">{post.latestComment.body}</p>
```
Jamiras, line 51: I don't have a monitor that supports large resolutions, but is 816px still appropriate at 4K?

Jamiras (approved):

resources/js/features/home/components/+root/RecentForumPosts/RecentForumPostItem.tsx
wescopeland: Yep, this is driven by the width of the container and independent of DPI or resolution.

PR #3015, opened 2025-01-03
perf(feed): speed up slow query

wescopeland:
There are 2 queries that are still kinda slow on the developer feed page: fetching the recent achievement unlocks and fetching the recent leaderboard entries.

The PR speeds up the leaderboard entries query.


app/Community/Actions/BuildDeveloperFeedDataAction.php
```diff
@@ -170,23 +170,12 @@ function ($join) {
...
-            ->join('LeaderboardDef', 'LeaderboardDef.ID', '=', 'leaderboard_entries.leaderboard_id')
-            ->where('LeaderboardDef.author_id', $targetUser->id)
+        return LeaderboardEntry::select('leaderboard_entries.*')
+            ->with(['leaderboard.game.system', 'user'])
+            ->join('LeaderboardDef as ld', 'ld.ID', '=', 'leaderboard_entries.leaderboard_id')
+            ->where('ld.author_id', $targetUser->id)
             ->whereNull('leaderboard_entries.deleted_at')
-            ->where('leaderboard_entries.updated_at', '>=', $thirtyDaysAgo)
-            ->select('leaderboard_entries.*')
+            ->where('leaderboard_entries.updated_at', '>=', now()->subDays(30))
```
Jamiras, line 178: Shouldn't this be `Carbon::now()->subDays(30)`? How is it finding the right `now()`. Doesn't PHP itself have a `now()` function?

app/Community/Actions/BuildDeveloperFeedDataAction.php
Jamiras, line 1: Other than changing the alias, this doesn't seem to change the query:
```
select `leaderboard_entries`.* from `leaderboard_entries`
 inner join `LeaderboardDef` on `LeaderboardDef`.`ID` = `leaderboard_entries`.`leaderboard_id`
 where `LeaderboardDef`.`author_id` = 43495 and `leaderboard_entries`.`deleted_at` is null
 and `leaderboard_entries`.`updated_at` >= '2024-12-04 14:20:42'
 and `leaderboard_entries`.`deleted_at` is null
 order by `leaderboard_entries`.`updated_at` desc limit 200
```

```
select `leaderboard_entries`.* from `leaderboard_entries`
 inner join `LeaderboardDef` as `ld` on `ld`.`ID` = `leaderboard_entries`.`leaderboard_id`
 where `ld`.`author_id` = 43495 and `leaderboard_entries`.`deleted_at` is null
 and `leaderboard_entries`.`updated_at` >= '2024-12-04 14:21:21'
 and `leaderboard_entries`.`deleted_at` is null
 order by `leaderboard_entries`.`updated_at` desc limit 200
```
So how does it improve the performance? Especially given the removal of the FORCE INDEX? Or was that causing the issue?

wescopeland: I didn't even realize after messing with this for a few rounds trying several different incantations that this is essentially where I ended up.

The only thing altered is the FORCE INDEX, which was slowing down the query. When I added it, it seemed to help a lot. Not sure why we've gone in the other direction now that it's removed.

app/Community/Actions/BuildDeveloperFeedDataAction.php
wescopeland: `now()` is provided by Laravel itself. This is the source code:
```php
if (! function_exists('now')) {
    /**
     * Create a new Carbon instance for the current time.
     *
     * @param  \DateTimeZone|string|null  $tz
     * @return \Illuminate\Support\Carbon
     */
    function now($tz = null)
    {
        return Date::now($tz);
    }
}
```

Jamiras (approved):

PR #3017, opened 2025-01-04
feat: use display_name everywhere

wescopeland:
**Pre-deploy SQL:**
```sql
UPDATE UserAccounts 
SET display_name = User 
WHERE display_name IS NULL;
```

---

This PR is a precursor to username change functionality. A subsequent PR will add a new panel to the Settings page for users to request a username change, and a new tool in Filament for moderators to approve username changes.

This PR makes the `display_name` field a first-class citizen on the site, having everywhere a user's name is displayed prefer it over `User` if a user's `display_name` value is non-empty.

I have tested:
* Every legacy PHP page.
* Every Blade UI page.
* Every React page.
* Every endpoint of the web API.
* The Connect API.
* Filament.
* All authentication flows (sign in using `display_name` instead of username, reset password, etc.).
* Email notifications.

Even though my manual testing has been extremely exhaustive, I have no doubt there are things here and there I may have missed. However, with the volume of users who are asking for username changes, I'd like to get this functionality done sooner rather than later.

**Implementation notes:**
* Pretend my `display_name` is "wes". This implementation allows you navigate to `/user/WCopeland` and `/user/wes`. This does not currently have a self-healing redirect (I don't think this is desirable until the respective pages are converted to React). In other words, `{user}`'s resolution logic is `$user->display_name ?? $user->User`.
* There is a large assumption baked in to this implementation that all usernames and display names are unique. In other words, no one can have a display name that matches an existing username, and vice versa. In the subsequent PR, we could even enforce this at the database level via a migration. 

**How to test this PR:**
Set your `display_name` value to anything that would be a valid username (min 3 chars, max 20 chars, no special chars). Perform any action anywhere on the site as you normally would.

Screenshot 2025-01-03 at 10 32 39 PM

Screenshot 2025-01-03 at 10 32 48 PM

Screenshot 2025-01-03 at 10 32 51 PM

Screenshot 2025-01-03 at 10 32 57 PM

Screenshot 2025-01-03 at 10 33 03 PM

---

Next steps:
* Write a migration to make `UserAccounts.display_name` non-nullable.


app/Helpers/database/user-relationship.php
Jamiras, line 1: The Message/Unfollow/Block links on the friends page use the User value.

Jamiras: Also the Message/Follow/Unfollow/Block links on the user's profile page.

app/Helpers/database/search.php
Jamiras, line 1: Searching on the User value returns the display_name record. This is particularly confusing if the User value is a substring of another user.

i.e. Searching on "test" could match "Tester" whose display name is now "QualityAssurance", but why would "QualityAssurance" be in a list when I searched for "test"?

You can also match User values this way in the Compare to Any user widget on game pages.

app/Helpers/database/player-game.php
```diff
@@ -313,11 +313,11 @@ function getUsersCompletedGamesAndMax(string $user): array
...
-        WHERE ua.User = :user
+        WHERE (ua.User = :user OR ua.display_name = :user2)
         AND gd.achievements_published > $minAchievementsForCompletion
         ORDER BY PctWon DESC, PctWonHC DESC, MaxPossible DESC, gd.Title";
 
-    return legacyDbFetchAll($query, ['user' => $user])->toArray();
+    return legacyDbFetchAll($query, ['user' => $user, 'user2' => $user])->toArray();
 }
 
 function getGameRecentPlayers(int $gameID, int $maximum_results = 10): array
```
Jamiras, line 323: Progress links in the Recent Players list use the User value.

Jamiras (commented):
Going to stop here for today. Have not tested public or connect APIs yet.

resources/views/pages-legacy/userInfo.blade.php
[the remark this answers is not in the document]
wescopeland: I think you're right. I've added a SQL command to the PR description to mirror `User` into `display_name`, and I've updated the register.php endpoint to write to `display_name` on sign up. We'll need to have a migration run after deploy to make the column non-nullable.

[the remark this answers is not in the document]
wescopeland: I don't think it's cause for concern. `User` should be considered stable.

resources/views/pages/user/[user]/tickets/index.blade.php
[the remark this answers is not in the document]
Comment (wescopeland) [reply resources/views/pages/user/[user]/tickets/index.blade.php:1]: Fixed in latest. The `withTrashed()` seems to be redundant.

resources/views/pages/user/[user]/developer/sets.blade.php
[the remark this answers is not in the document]
Comment (wescopeland) [reply resources/views/pages/user/[user]/developer/sets.blade.php:1]: Fixed in latest. The `withTrashed()` seems to be redundant.

app/Filament/Resources/UserResource.php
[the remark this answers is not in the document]
wescopeland: It should. Fixed in latest:
Screenshot 2025-01-05 at 8 05 58 PM

Additionally, if the user has a custom display name that differs from their stable username, we now show it on the infolist:
Screenshot 2025-01-05 at 8 06 25 PM

resources/views/pages-legacy/achievementList.blade.php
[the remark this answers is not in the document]
wescopeland: I may be getting confused by this. Here's what I see:

https://github.com/user-attachments/assets/7072e488-4d73-40c1-819c-58aeb8aa6a27

resources/views/pages-legacy/leaderboardinfo.blade.php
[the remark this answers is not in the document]
wescopeland: Fixed in latest.

public/request/forum-topic-comment/create.php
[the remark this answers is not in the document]
wescopeland: Fixed in latest.

app/Community/Components/UserProfileMeta.php
```diff
@@ -75,23 +75,23 @@ private function buildDeveloperStats(User $user, array $userMassData): array
...
+            'href' => $gameAuthoredAchievementsCount ? route('developer.sets', ['user' => $user->display_name]) : null,
             'isMuted' => !$gameAuthoredAchievementsCount,
         ];
 
         // Achievements unlocked by players
         $achievementsUnlockedByPlayersStat = [
             'label' => 'Achievements unlocked by players',
             'value' => localized_number($userMassData['ContribCount']),
-            'href' => $userMassData['ContribCount'] > 0 ? route('user.achievement-author.feed', ['user' => $user]) : null,
+            'href' => $userMassData['ContribCount'] > 0 ? route('user.achievement-author.feed', ['user' => $user->display_name]) : null,
```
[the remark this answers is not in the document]
wescopeland: Fixed in latest.

app/Helpers/database/user-relationship.php
wescopeland: Fixed both in latest.

app/Helpers/database/search.php
wescopeland: I agree this seems confusing.

The best UX I can think of at the moment is the search should just always respect the `display_name`. This particularly makes sense if the course we're going to pursue is that `display_name` will be non-empty & non-nullable.

app/Helpers/database/player-game.php
wescopeland: Fixed in latest.

resources/views/pages-legacy/achievementList.blade.php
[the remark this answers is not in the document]
Jamiras: This was supposed to be attached to `achievementInfo.blade.php`. For something I have unlocked, I'm not showing up in the Recent Unlocks list:
image

app/Models/User.php
```diff
@@ -348,19 +349,45 @@ public function registerMediaCollections(): void
...
+        if ($displayNameOrUsername === null) {
+            return static::query();
+        }
+
+        return static::query()
+            ->where(function ($query) use ($displayNameOrUsername) {
+                $query->where(function ($q) use ($displayNameOrUsername) {
+                    $q->whereNotNull('display_name')
+                      ->where('display_name', '!=', '')
+                      ->where('display_name', $displayNameOrUsername);
```
Jamiras, line 367: doesn't "where displayname=value" eliminate the need for the other two checks?

resources/views/pages-legacy/userInfo.blade.php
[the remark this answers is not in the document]
Jamiras: Yes, but if "Joe" doesn't want people to know he used to be "BigFatJoe", doesn't having "BigFatJoe.png" as his avatar kind of eliminate that possibility?

public/request/forum-topic-comment/create.php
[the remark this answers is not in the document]
Jamiras: I'm still seeing this.
```
[user=Jamiras]
[user=JamirasDisplay]
```
becomes:
```
[user=Jamiras]
[user=#####]
```

Both links show as "JamirasDisplay" after saving, but show as "[user=Jamiras]" when editing.

So, two problems:
1) [user=OldName] doesn't become [user=id]
2) [user=id] becomes [user=OldName] on edit

app/Helpers/database/user-relationship.php
Jamiras: Mutual follower still has a Follow link in their Following list:

this branch:
image

master:
image

app/Helpers/database/user-auth.php
```diff
@@ -82,7 +82,7 @@ function authenticateForConnect(?string $username, ?string $pass = null, ?string
 
     return [
         'Success' => true,
-        'User' => $user->User,
+        'User' => $user->display_name,
```
Jamiras, line 85: This should still return `User`. The display name should be returned in `DisplayName`. The client is already ready for that change: https://github.com/RetroAchievements/rcheevos/pull/191

Passing `DisplayName` as `User` causes the avatar to resolve incorrectly
image

And the request for the patchdata to fail:
```
{"Success":false,"Error":"Invalid user\/token combination.","Code":"invalid_credentials","Status":401}
```

Because `authenticateFromAppToken` is only checking for `User`.

resources/views/pages-legacy/userInfo.blade.php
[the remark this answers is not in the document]
wescopeland: The only thing I can imagine we can do for this is start writing these "UserPic" files using the display_name instead of username, and when we ultimately implement an action to assign a display_name, rename the file as part of the action.

However, this seems problematic, given that clients seem to explicitly expect the value to be the username: https://github.com/RetroAchievements/RAWeb/pull/3017#discussion_r1909599455

It seems like whatever we do to resolve this, it's going to break clients. Do you have any ideas?

app/Models/User.php
wescopeland: Strange, `->where('display_name', '!=', '')` appears to be eliminated in the branch head. I don't see that clause in latest.

resources/views/pages-legacy/achievementList.blade.php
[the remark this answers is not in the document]
wescopeland: Ah, good catch. Resolved in latest:
Screenshot 2025-01-09 at 8 04 56 PM

public/request/forum-topic-comment/create.php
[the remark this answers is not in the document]
wescopeland: Should now be resolved in latest.

For some reason, and I'm afraid to find out why, Livewire is showing a temporary flash of unfilled content in the textarea on page load.

That issue will not happen after https://github.com/RetroAchievements/RAWeb/pull/3021.

app/Helpers/database/user-relationship.php
wescopeland: Ah, I misinterpreted. This should now be resolved in latest.

resources/views/pages-legacy/admin.blade.php
```diff
@@ -15,7 +15,7 @@
 if ($action === 'alt_identifier') {
     $altsForUser = request()->input('u');
 
-    $forUser = User::where('User', $altsForUser)->first();
+    $forUser = User::whereName($altsForUser)->first();
```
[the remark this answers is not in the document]
wescopeland: I'm wondering if that may be confusing. `userAvatar()` doesn't have a way currently to override the label (and I'm not sure this is desirable).

Ideally, the log for username changes will always be exposed to mods/admins. Hopefully it won't take much digging to find that info if needed.

resources/views/pages-legacy/history.blade.php
```diff
@@ -6,13 +6,11 @@
 authenticateFromCookie($user, $permissions, $userDetails);
 
 $userPage = requestInputSanitized('u', $user);
+$userDetails = User::whereName($userPage)->first();
 
-if (!isset($userPage) || !isValidUsername($userPage)) {
-    abort(404);
-}
+$userPage = $userDetails->display_name;
```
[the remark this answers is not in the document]
wescopeland: Fixed in latest.

app/Helpers/database/user-auth.php
wescopeland: Added `DisplayName` return in latest.

Per https://github.com/RetroAchievements/RAWeb/pull/3017#discussion_r1909629207, this does seem to imply that clients are always anticipating `/UserPic/{$username}.png` as the avatar URL.

resources/views/pages-legacy/userInfo.blade.php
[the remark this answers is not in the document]
Jamiras: `achievementwondata`, `getfriendlist`, `lbinfo` haven't been plumbed for DisplayName support yet. As such, if they are changed to return the display name in the User field, any clients that show images for those records would try to use the DisplayName to find the image. 

`getfriendlist` is (afaik) only used by the DLL as it's not exposed through rcheevos yet, but it does display images. `achievementwondata` and `lbinfo` are exposed through rcheevos. Their usage in the DLL does not currently show images, but I don't know if any other emulator usages do show images for those lists.

Alternately, we could add a new DisplayName field to those APIs and recommend clients show the display name instead, while still using the Username to [resolve the image](https://github.com/RetroAchievements/rcheevos/wiki/rc_api_init_fetch_image_request#parameters).

In that case, `User` would continue to be exposed for locating the image, and `DisplayName` support for those lesser-used APIs would have to be implemented at a later time.

This is probably the better way forward, even though it does leave the User value exposed. If we were going to limit the exposure to the client APIs and not the webpage, that would limit the "leak", but would require being able to access the same image through more than one URL.

At this time, I'm going to recommend we proceed with the image URL still containing the User name.

public/dorequest.php
Jamiras, line 1: `achievementwondata`, `getfriendlist`, `lbinfo` should be modified to return `DisplayName` in addition to `User`.

wescopeland: I've made these changes in latest - please give them an extra look. While I've tested these changes, I'm not 100% confident in them. Just want to be extra sure we're okay surfacing the username in all of these.

I've made a change in _user-auth.php_ such that `u` now accepts both `User` and `display_name`.

`achievementwondata`, `getfriendlist`, and `lbinfo` have all been modified to return `DisplayName` values, ie:
```json
		"Entries": [
			{
				"User": "Bkid",
				"DisplayName": "Bkid",
```

Jamiras: > I've made a change in user-auth.php such that u now accepts both User and display_name.

I think we want the user to always log in with their username, don't we? The problem I reported earlier was that since DisplayName was being returned by the login API, future calls passed it as the user's Username.

Did we change the web login to support display name?

wescopeland: Yep, logging in on the site via display_name should already work on this branch.

I figured once a user sets a `display_name` it would probably be very counterintuitive for them to log in with their original username. For the Connect API it's probably less of a concern past the initial emulator login call, so I'd be happy to revert that if you think we should.

Jamiras: If they can log in to the site using their display name, they'll expect to be able to do so from the emulator too. I was under the impression that the User name was still part of their credentials, and the display name was just something that should be public-facing. Hence, my concerns about still exposing the Username through the avatar images on the webpages.

rcheevos uses the `User` value returned from the login call as the `u` parameter for all future calls, so it seems unlikely that the display name would be passed to any of them unless there's some client not using and not mimicking rcheevos.

I guess it doesn't hurt to check both.

app/Models/User.php
Jamiras: It's gone now, but my question remains. Can't this (365-369) be simplified to:
```
          $query->where('display_name', $displayNameOrUsername)
                ->orWhere('User', $displayNameOrUsername);
```

resources/views/pages-legacy/history.blade.php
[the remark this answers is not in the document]
Jamiras: I think it's reintroduced.
```
$userPage = $userDetails?->display_name ?? $userDetails->username;
```
If `$userDetails` is null, it'll cascade to the right side where there's no `?->`.

public/dorequest.php
Jamiras: > `achievementwondata`, `getfriendlist`, and `lbinfo` have all been modified to return `DisplayName`. [...] please give them an extra look.

The responses look correct to me. I'll get an rcheevos PR open soon to look for them.

> Just want to be extra sure we're okay surfacing the username in all of these.

We basically have three choices: 
1) Overload `User` with the DisplayName and somehow associate the user's avatar image with `/User/DisplayName.png`.
2) Send `User` and `DisplayName`, and expect clients to display the `DisplayName` while fetching the image using the `User`.
3) Add an `AvatarUrl` with the full path to the image (which would contain the User name at this time, but we could change it in the future), and overload `User` with the DisplayName.

Since we're making changes to the API, it's probably reasonable to do #3.

But we still have the issue of the primary user. Since we're not sending DisplayName yet in the login response, we could continue to not do so, send the DisplayName in the `User` field, and add an `AvatarUrl` to that as well. Existing clients would try to fetch the image through the DisplayName and get back the default avatar.

app/Models/User.php
wescopeland: Yep, it can. Updated in latest.

resources/views/pages-legacy/history.blade.php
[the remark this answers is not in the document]
wescopeland: Ah, I see. Updated in latest so the right side also uses `?->`.

public/dorequest.php
wescopeland: Implemented number 3 in latest. To reiterate, this means:
* `User` is overloaded with the user's `display_name` in `login2`, `achievementwondata`, getfriendlist`, and `lbinfo`.
* `AvatarUrl` has been added to all four of those.

I also see we have `codenotes2`. This is currently returning `User` and `DisplayName` - I assume this is used by the toolkit only - not sure if this should be changed or left as-is.

Jamiras: > I also see we have codenotes2. This is currently returning User and DisplayName - I assume this is used by the toolkit only - not sure if this should be changed or left as-is.

rcheevos is only [processing the `User` value](https://github.com/RetroAchievements/rcheevos/blob/3a91a58605c0fb05833a228dbb674357b0e65a09/src/rapi/rc_api_editor.c#L57-L61), and returning it as `author`. The only thing the toolkit uses is for is the warning that you're overwriting someone else's note. So, I think it's reasonable to return display name there too.

wescopeland: I've verified this is happening in latest:
```json
{
	"User": "WCopeland",
	"DisplayName": "wesweswes",
	"Address": "0x0001a0",
	"Note": "[8-bit] Total Lives Remaining\r\n\r\n02 = Two lives remaining\r\n00 = Player is on last life (not game over)\r\nFD = Game Over \/ Password screen\r\n\r\nSymbol: lives"
},
```

Jamiras: Sorry, I meant "I think it's reasonable to return display name [in the User field] there too".

Then all the APIs will be overloading User with the display name, and provide an AvatarUrl where we expect the client might want to display one.

wescopeland: Ah, I understand now. In latest, I've changed the query to do `SELECT ua.display_name AS User`, ie:
Screenshot 2025-01-16 at 8 56 38 PM

Jamiras (approved):

PR #3020, opened 2025-01-05
add event_awards table

Jamiras:
Adds the ability to associate badges to a number of unlocked achievements within an Event game.

For example, for AotW 2025:
image

This PR only touches the administration side of things. Future PRs will be opened to award and display the badges.


database/migrations/2025_01_05_000000_create_event_awards_table.php
```diff
@@ -0,0 +1,36 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::create('event_awards', function (Blueprint $table) {
+            $table->bigIncrements('id');
+            $table->unsignedBigInteger('game_id');
```
wescopeland, line 14: I have reviewed all of the code in this PR and it all looks logically sound. I want to steer the conversation towards this migration, as it drives the architectural underpinnings of the PR and could have profound impacts on how the system stores and manages events going forward.

We're creating increasingly tight coupling between events and games, and I'm not entirely sure this is desirable unless we want to establish that events will continue to live on indefinitely as games in the database.

My view could be changed, but I have a feeling that the path we should be driving towards is `events` existing as their own kind of entity. This would facilitate varied kinds of events, an actual events dashboard / home page, and potentially even a quests system in the future.

I know we don't have an `events` table yet, but it seems like this may actually be our best opportunity to create one. The migration below is me spitballing and have only thought about this for 5 minutes. The core idea here is Event Awards are associated to a new type of entity -- an `Event` -- not games:

```php
Schema::create('events', function (Blueprint $table) {
    $table->bigIncrements('id');
    // For now this will always point to a game with ConsoleID 101.
    // In the future this can be made nullable.
    $table->unsignedBigInteger('legacy_game_id');
    $table->string('name', 100);
    // `EventType` enum? achievement_hunt, quest, competition, etc
    $table->string('type', 50);
    // For storing event-specific config/metadata, ie: rules, etc
    $table->json('properties')->nullable();
    $table->dateTime('starts_at');
    $table->dateTime('ends_at');
    $table->timestamps();
    $table->softDeletes();

    // When we're ready to migrate away from events-as-games,
    // we can make this nullable and add games via a proper pivot table.
    $table->foreign('legacy_game_id')
        ->references('ID')
        ->on('GameData')
        ->onDelete('cascade');
});

Schema::create('event_awards', function (Blueprint $table) {
    $table->bigIncrements('id');
    $table->unsignedBigInteger('event_id');
    $table->integer('tier_index');
    $table->string('label', 40);
    // either nullable or defaults to 0? not all events may have achievements required
    $table->integer('achievements_required');
    $table->string('image_asset_path', 50);
    $table->timestamps();
    $table->softDeletes();

    $table->foreign('event_id')
        ->references('id')
        ->on('events')
        ->onDelete('cascade');

    $table->unique(['event_id', 'tier_index']);
});
```

With this schema, events are still associated to games via `legacy_game_id`, but event_awards are now associated to events. As more event stuff is built out, it can be attached to `events` rather than `GameData`, which will make extraction of events from games significantly easier in the future.

There's both an immediate and long-term benefit of pursuing this path in that we already beginning to break the coupling of events to games while maintaining backwards compatibility through `legacy_game_id`.

Jamiras: While I agree with the idea, I don't think we're in a place where it's practical yet. 

The way the AotW2025 event is functioning now, it relies on the `player_achievements` table to hold the unlocks and the GameMetrics aggregate jobs to track player counts and per-week participation. And because unlocks are associated to achievements, and achievements are associated to games, there's already a tight coupling that will be a lot harder to fix than adding an extra foreign key to the event_awards table to point an event instead of a game.

To do this effectively, we'd have to have a `player_event_achievements` table to track event achievement unlocks, and EventMetrics jobs to update player counts and participation. Then we'd have to duplicate a lot of the widgets on the game page to query and use the alternate data.

wescopeland: I do acknowledge these valid concerns regarding event achievement entities. My suggestion is that we begin taking an incremental approach with the `events` table:

* We can create the new table now but keep the game-based achievement tracking in place. We don't have to disrupt that at the moment.
* This gives us the foundation to gradually migrate event functionality rather than further tightening the coupling to games. We can do this right now with awards, and later tackle more complex stuff like the achievement tracking and metrics.
* Having events as a separate entity makes each future migration/decoupling step cleaner.
 
The goal isn't to decouple everything immediately, but to create a path for progressive decoupling as resources allow. Does that seem practical?

Jamiras: I'll think about this some more. It's probably reasonable to make an `event` object wrap a game, so we can keep all the game-derived functionality working while offloading some of the non-game dependencies.

Jamiras: These are now attached to an event record. Ready for review again.

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
```diff
@@ -0,0 +1,130 @@
...
+                            ->label('Badge')
+                            ->disk('livewire-tmp') // Use Livewire's self-cleaning temporary disk
+                            ->image()
+                            ->rules([
+                                'dimensions:width=96,height=96',
+                            ])
+                            ->acceptedFileTypes(['image/jpeg', 'image/png', 'image/gif'])
+                            ->maxSize(1024)
+                            ->maxFiles(1)
+                            ->previewable(true),
```
wescopeland, line 70: Looks like it's possible to submit without having uploaded a file. When doing so, `image_asset_path` is `EMPTY`. Should this field be set to required?

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
```diff
@@ -0,0 +1,130 @@
...
+        return $form
+            ->schema([
+                Forms\Components\TextInput::make('label')
+                    ->minLength(2)
+                    ->maxLength(40)
+                    ->required(),
+
+                Forms\Components\TextInput::make('achievements_required')
+                    ->numeric()
+                    ->required(),
```
wescopeland, line 47: Let's say tier 1 requires 10 achievements. Should it be possible for tier 2 to require 5 achievements?

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
wescopeland, line 1: If I create 3 awards (tiers 1, 2, and 3), I can delete a tier and end up in this situation:

Screenshot 2025-01-11 at 2 44 02 PM

Notice that tier 2 is now skipped, and it appears that the tier field is disabled, so I can't edit it on any of the existing awards.

I'm almost wondering if the tier should be a derived value rather than one recorded in the database. In other words, scan the event's associated awards and automatically infer that the one with the lowest achievements required threshold is tier 1, etc.

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
Jamiras: No. How do we enforce that?

wescopeland: This is what I came up with, give it a try:
```php
$minRequired = $event->awards()
    ->where('tier_index', '<', $nextTier)
    ->max('achievements_required') ?? 0;

return $form
    ->schema([
        Forms\Components\TextInput::make('label')
            ->minLength(2)
            ->maxLength(40)
            ->required(),

        Forms\Components\TextInput::make('achievements_required')
            ->numeric()
            ->rules([
                'required',
                'numeric',
                'min:' . ($minRequired + 1),
            ]),
```

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
wescopeland: Yeah, it probably makes the most sense to make it such that badges can't be deleted. That way there's less of a possibility that we'll accidentally take away something from someone.

Note that my solution in this comment https://github.com/RetroAchievements/RAWeb/pull/3020#discussion_r1912226068 may be disruptive towards editing the `achievements_required` value for a row. In other words, it may make this sort of edit impossible:
Screenshot 2025-01-11 at 3 48 44 PM

Jamiras: Done. Badges can no longer be deleted.

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
Jamiras: Expanded on this to set minimum _and_ maximum values to ensure values don't overlap.

app/Filament/Resources/EventResource/RelationManagers/EventAwardsRelationManager.php
Jamiras: It cannot be required for edit or the user has to upload a badge to change the label. I did figure out a way to make it required for new records.

wescopeland (approved):

PR #3027, opened 2025-01-08
fix(home): add max-height to 'new' label in news cards

wescopeland:
This PR resolves a very minor layout shift that can occur on the home page caused by the "new" label in the news cards.

The shift occurs when there are 3 lines of lead text in the card and the user navigates to the next page of cards.

**Root Cause:**
The height of the "new" label is slightly increasing the height of the news card itself.

Here is what the issue looks like (currently reproducible on prod):

https://github.com/user-attachments/assets/149c944d-b791-4fa7-a838-03460cb15cbb


Jamiras (approved):
Possibly worth noting that it's only a problem if the new post has at least three lines of text.

wescopeland:
Noted, though this may have been ambiguous:
> The shift occurs when there are 3 lines of lead text in the card

PR #3028, opened 2025-01-08
fix(home): remediate 'Ends in...' padding issue

wescopeland:
This PR resolves a minor issue being surfaced by the "smalldate" classname on the home page's AOTW component.

"smalldate" is overly greedy in applying a minimum width to its text content, which gives the illusion that the "Ends in" label has excessive right padding.

**Before**
Screenshot 2025-01-07 at 7 56 50 PM

**After**
Screenshot 2025-01-07 at 7 57 07 PM


wescopeland:
We sure can, here's how it looks in latest (with a hardcoded & obviously-wrong `activeUntil` value):
Screenshot 2025-01-09 at 8 55 45 PM

PR #2975, opened 2024-12-23
feat: fully migrate hash management to filament

wescopeland:
This PR drops the Blade manage hashes page, migrating hash management capabilities fully into Filament.

Hash management for a game is now a dedicated subpage (to facilitate proper linking) instead of a relation manager. However, all the existing relation manager code has been ported over 1:1.

Screenshot 2024-12-23 at 2 04 42 PM

Screenshot 2024-12-23 at 2 04 46 PM

PR #3032, opened 2025-01-08
feat(manage): move game sets management exclusively to filament

wescopeland:
This PR migrates `game_sets` management from the game page devbox to live exclusively in Filament. This is likely required for "Hubs 2.0" to be generally available, as right now hub pages point to `game_set` IDs, not game IDs.

**Before**
Screenshot 2025-01-08 at 5 58 24 PM

**After**
Screenshot 2025-01-08 at 5 58 35 PM

Similar to the hashes page introduced in https://github.com/RetroAchievements/RAWeb/pull/2975, these two game relation managers also have dedicated pages (so they have stable URLs which can be linked to):

Screenshot 2025-01-08 at 5 59 47 PM

Screenshot 2025-01-08 at 6 00 03 PM

Screenshot 2025-01-08 at 6 00 08 PM


app/Filament/Resources/GameResource/Pages/SimilarGames.php
```diff
@@ -0,0 +1,206 @@
...
+                    ->action(function (array $data): void {
+                        /** @var Game $game */
+                        $game = $this->getOwnerRecord();
+
+                        // This assumes there will only ever be one similar games set.
+                        $similarGamesSet = GameSet::whereType(GameSetType::SimilarGames)
+                            ->whereGameId($game->id)
+                            ->first();
+
+                        (new AttachGamesToGameSetAction())->execute($similarGamesSet, $data['game_ids']);
```
Jamiras, line 138: Adding a similar game makes it appear in the similar games list for both games, but it only appears in the Manager Similar Games page for the game it was added from.

Jamiras: Oddly, the count appears to include the added game, but the list does not show it (filters are set to Any):
image

Jamiras: Conversely, the game where the similar game was added from shows the game in the list, but not in the count:
image

app/Filament/Resources/GameResource.php
```diff
@@ -542,6 +544,8 @@ public static function getPages(): array
             'create' => Pages\Create::route('/create'),
             'view' => Pages\Details::route('/{record}'),
             'edit' => Pages\Edit::route('/{record}/edit'),
+            'hubs' => Pages\Hubs::route('/{record}/hubs'),
+            'similar-games' => Pages\SimilarGames::route('/{record}/similar-games'),
```
Jamiras, line 548: Out of curiosity: Why are these at the top navigation bar instead of in the lower navigation bar with achievements and such?

Related Hubs are in the lower navigation bar on individual hub pages.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -777,35 +777,14 @@ function resize() {
...
-                echo "<form class='mb-2' method='post' action='/request/game-relation/create.php'>";
-                echo csrf_field();
-                echo "<input type='hidden' name='game' value='$gameID'>";
-                echo "<div class='md:grid grid-cols-[180px_1fr_100px] gap-1 items-center mb-1'>";
-                echo "<label for='game_relation_add'>Add Related Games<br>(CSV of game IDs)</label>";
-                echo "<input type='text' name='relations' id='game_relation_add' class='w-full'>";
-                echo "<div class='text-right'><button class='btn'>Add</button></div>";
-                echo "</div>";
-                echo "</form>";
+                if ($userModel && $userModel->can('manage', [\App\Models\GameSet::class])) {
```
Jamiras, line 780: Do you still need/want to check `Permissions::Developer` on the previous line?

app/Filament/Resources/GameResource.php
wescopeland: Only intention here is to give these a stable URL to link to. When they're relation managers on the bottom, they can be linked to via a query param, but I believe the link can change at any time if relation managers are added/removed to&from the resource.

resources/views/pages-legacy/gameInfo.blade.php
wescopeland: Nope - removed in latest.

app/Filament/Resources/GameResource/Pages/SimilarGames.php
wescopeland: Fixed in latest - I needed to create a couple of new actions to ensure these bidirectional relationships are respected on attach/detach of similar games.

PR #3034, opened 2025-01-09
fix(SEO): change default twitterCard value

wescopeland:
This PR resolves an issue with Discord embeds on Inertia.js pages.

**Before**
Screenshot 2025-01-08 at 7 24 30 PM

**After**
Screenshot 2025-01-08 at 7 24 21 PM

Can be tested with the following extension if desired: https://chromewebstore.google.com/detail/discord-embeds/faeojpkidgnhcochgodeklokfimbencc


Jamiras (approved):
I don't see any differences looking at the page source for a hub, but the change looks reasonable.

PR #3037, opened 2025-01-09
fix(game): reduce claim info moderator memory usage

wescopeland:
Resolves an issue where moderators can no longer open https://retroachievements.org/game/228.

**Root Cause:**
```php
$achievementSetClaims->load(['user', 'game', 'game.playerSessions', 'game.achievements', 'game.memoryNotes']);
```
This is loading all of the game's player sessions, achievements, and memory notes in a single query. We only need to load data in targeted queries for each claimant.

### DIFF
--- a/resources/views/components/game/claim-info.blade.php
+++ b/resources/views/components/game/claim-info.blade.php
@@ -21,12 +21,7 @@ use Illuminate\Support\Facades\DB;
 $expirationText = '';
 
 if (!$achievementSetClaims->isEmpty()) {
-    // TODO use a policy
-    if ($userPermissions >= Permissions::Moderator) {
-        $achievementSetClaims->load(['user', 'game', 'game.playerSessions', 'game.achievements', 'game.memoryNotes']);
-    } else {
-        $achievementSetClaims->load(['user', 'game']);
-    }
+    $achievementSetClaims->load(['user', 'game']);
 
     $firstClaim = $achievementSetClaims->first();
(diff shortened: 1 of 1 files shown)


resources/views/components/game/claim-info.blade.php
```diff
@@ -76,20 +70,20 @@
             $activity = 'played this game';
         } else {
             // player_sessions only exist after 14 Oct 2023
-            $achievement = $achievementSetClaim->game->achievements
+            $achievement = Achievement::where('GameID', $achievementSetClaim->game->id)
```
Jamiras, line 73: I'm pretty sure we can remove this code, a claim had to be renewed every three months, and Oct 2023 is like 15 months ago. Wouldn't recommend doing that in this PR though.

Jamiras (approved):
Seems to work. On a game with 600 users, before:
image
After:
image

Also verified a game with a collaboration claim, and all claimants appeared to be reported correctly. It is generating more queries as each claimant is queried individually, but I think that's acceptable.

resources/views/components/game/claim-info.blade.php
wescopeland: Yea, the only way I could personally fall into this branch is if I started doing stuff without ever playing the game (like editing a code note on the website). I was wondering what the actual utility of it was now that we have player sessions. Probably best to just get this logic when we finally get to the game page React migration.

PR #3039, opened 2025-01-09
support uploading images for games and events from filament

Jamiras:


app/Filament/Resources/NewsResource/Actions/ProcessUploadedImageAction.php
```diff
@@ -41,6 +41,18 @@ public function execute(string $tempImagePath, ImageUploadType $imageUploadType)
...
+                    $imagePath = UploadGameImage($file, ImageType::GameBoxArt);
+                    break;
+                case ImageUploadType::GameTitle:
+                    $file = createFileArrayFromDataUrl($dataUrl);
+                    $imagePath = UploadGameImage($file, ImageType::GameTitle);
+                    break;
+                case ImageUploadType::GameInGame:
+                    $file = createFileArrayFromDataUrl($dataUrl);
+                    $imagePath = UploadGameImage($file, ImageType::GameInGame);
+                    break;
```
wescopeland, line 55: Looks like duplicated code may be starting to stack up here. I wonder if a private function may be beneficial?

```php
// untested pseudocode
private function processGameImage(string $dataUrl, string $imageType): string {
    $file = createFileArrayFromDataUrl($dataUrl);
    return UploadGameImage($file, $imageType);
}
```

app/Filament/Resources/GameResource/Pages/Edit.php
```diff
@@ -18,6 +20,46 @@ protected function mutateFormDataBeforeSave(array $data): array
...
+
+        if (isset($data['ImageBoxArt'])) {
+            $data['ImageBoxArt'] = (new ProcessUploadedImageAction())->execute(
+                $data['ImageBoxArt'],
+                ImageUploadType::GameBoxArt,
+            );
+        } else {
+            // If no new image was uploaded, retain the existing image.
+            unset($data['ImageBoxArt']);
+        }
```
wescopeland, line 61: May be able to cut down a lot of duplication with:

```php
$imageTypes = [
    'ImageIcon' => ImageUploadType::GameBadge,
    'ImageTitle' => ImageUploadType::GameTitle, 
    'ImageIngame' => ImageUploadType::GameInGame,
    'ImageBoxArt' => ImageUploadType::GameBoxArt,
];

foreach ($imageTypes as $field => $type) {
    if (isset($data[$field])) {
        $data[$field] = (new ProcessUploadedImageAction())->execute($data[$field], $type);
    } else {
        // If no new image was uploaded, retain the existing image.
        unset($data[$field]);
    }
}
```

app/Filament/Resources/GameResource/Pages/Edit.php
```diff
@@ -5,7 +5,9 @@
 namespace App\Filament\Resources\GameResource\Pages;
 
 use App\Filament\Concerns\HasFieldLevelAuthorization;
+use App\Filament\Enums\ImageUploadType;
 use App\Filament\Resources\GameResource;
+use App\Filament\Resources\NewsResource\Actions\ProcessUploadedImageAction;
```
wescopeland, line 10: We should probably lift this out of NewsResource. Maybe it can just live the an `App\Filament\Actions` namespace?

wescopeland (approved):
All nit feedback, verified the implementation locally.

app/Filament/Resources/GameResource/Pages/Edit.php
Jamiras: done.

app/Filament/Resources/GameResource/Pages/Edit.php
Jamiras: I like it. 

I wish there were some way to turn the inner lines of code into a function, as the `if isset {} else unset` occurs every place the `ProcessUploadedImageAction` is used.

app/Filament/Resources/NewsResource/Actions/ProcessUploadedImageAction.php
Jamiras: I thought about it, but it's just cutting two lines down to one for each case.

I've opted to use a mapping dictionary, similar to the one you proposed for the calling function.

wescopeland (approved):

PR #3041, opened 2025-01-10
ci: switch vitest pool mode to 'threads'

wescopeland:
Improves vitest execution time by roughly 15% on my local machine (43s -> 36s).


wescopeland:
Note that this appears to have virtually no impact on GitHub Actions vitest performance, likely due to how constrained those containers are. We'll probably need to reach for sharding to improve cloud CI performance.

Jamiras (approved):
Without change: 2m51.815s
With change: 1m21.863s

(only ran tests once, so there might be some statistical variance, but I doubt that would swing things enough to make this change not be beneficial).

PR #3042, opened 2025-01-10
add checkbox to hide inactive achievements on event page

Jamiras:
image


wescopeland (approved):
Verified locally.

The only thing I'm wondering is if the default behavior should be to always enable this checkbox? I could see pros and cons of either approach.

I _suspect_ (only based on vibes) that a user would want to keep this enabled after seeing the entire list the first time.

PR #3043, opened 2025-01-10
fix(useGameListState): ignore persisted page index of 0

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1326579342230159360/1327411224249368748.

**Root Cause**
We're actually always persisting a page index of 0 in the "Remember My View" cookie.

I think it's actually desirable to do this so the types between pagination state in the cookie and datatable are perfectly aligned. Our logic error is that we weren't throwing away this persisted 0.

In other words, when a user currently has "Remember My View" checked:
* They navigate to page N of the table.
* They click a game.
* They click the back button.
* The table now thinks they're on page 1 even though they aren't, and this messes up table state.

### DIFF
--- a/resources/js/features/game-list/hooks/useGameListState.test.ts
+++ b/resources/js/features/game-list/hooks/useGameListState.test.ts
@@ -230,7 +230,7 @@ describe('Hook: useGameListState', () => {
     });
 
     // ASSERT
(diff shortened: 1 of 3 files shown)


Jamiras (approved):
I couldn't reproduce (either before or after), but it doesn't seem to break anything.

PR #3048, opened 2025-01-11
feat(home): change AOTW href

wescopeland:
This PR:
* Changes the "Learn more about this event" label to "View this year's event", which links to the event "game" page.
* Rips out all the `StaticData` stuff from `BuildAchievementOfTheWeekDataAction` and `API_GetAchievementOfTheWeek`. I assume the event will follow the automated system going forward, so we're probably at a point that all this stuff can be removed.
* Deletes the legacy AOTW admin tool in _admin.blade.php_ and the corresponding routine in the legacy admin HTTP endpoint.

Screenshot 2025-01-11 at 12 32 42 PM


app/Http/Actions/BuildAchievementOfTheWeekDataAction.php
```diff
@@ -53,7 +34,8 @@ public function execute(?StaticData $staticData): ?EventAchievementData
             'sourceAchievement.game.badgeUrl',
             'sourceAchievement.game.system.iconUrl',
             'sourceAchievement.game.system.nameShort',
-            'forumTopicId',
+            'event',
+            'event.legacyGame',
```
Jamiras, line 38: Should this be `event.game`? I see you have a TODO: rename there, but as of right now, they don't match.

app/Models/Event.php
```diff
@@ -113,7 +115,7 @@ public function achievements(): HasManyThrough
             'GameID',         // Achievements.GameID
             'achievement_id', // event_achievements.achievement_id
             'ID',             // Game.ID
-            'ID',             // Achievement.ID
+            'ID',             // Achievements.ID
```
Jamiras, line 118: If you're going to fix this one, you should fix 117 as well: `GameData.ID`.

public/request/admin.php
```diff
@@ -82,23 +82,3 @@
...
-
-if ($action === 'aotw') {
-    $aotwAchID = requestInputSanitized('a', 0, 'integer');
-    $aotwForumID = requestInputSanitized('f', 0, 'integer');
-    $aotwStartAt = requestInputSanitized('s', null, 'string');
-
-    $query = "UPDATE StaticData SET
-        Event_AOTW_AchievementID='$aotwAchID',
-        Event_AOTW_ForumID='$aotwForumID',
-        Event_AOTW_StartAt='$aotwStartAt'";
```
Jamiras, line 94: Should we have a migration to drop these columns? Or just let them stagnate? I don't think the table will be going away any time soon.

app/Http/Actions/BuildAchievementOfTheWeekDataAction.php
wescopeland: In latest, I went ahead and implemented the TODO. This is just so the relation name and DTO field closely match the column name. Should also make it easier to find these references later.

app/Models/Event.php
wescopeland: I didn't even notice that one - fixed in latest.

public/request/admin.php
wescopeland: I think we can probably drop around half of that table's columns. I think we should do it, but let's push that off to a subsequent PR.

Jamiras (approved):

app/Http/Actions/BuildAchievementOfTheWeekDataAction.php
```diff
@@ -24,25 +21,8 @@ public function execute(?StaticData $staticData): ?EventAchievementData
             ->with(['achievement.game', 'sourceAchievement.game'])
             ->first();
 
-        if (!$achievementOfTheWeek || !$achievementOfTheWeek->source_achievement_id) {
```
Jamiras, line 27: This still needs to ensure there's a source achievement, or an exception occurs on the main page.

wescopeland: Resolved in latest. I've also added another test case to `HomeControllerTest` to ensure this doesn't accidentally get broken again.

PR #3056, opened 2025-01-13
test(GameSuggestionEngine): resolve flake from randomness

wescopeland:
`GameSuggestionEngine` is using weighted random selection. While this works well in userland, it is leading to test flake since even high-weight strategies aren't guaranteed to be selected within the max attempts limit.

To fix this, I've added a test-only override that forces the engine to always select the first strategy in the list. This makes our tests deterministic while still validating the core suggestion logic. This override is intentionally named with `unsafe_` and `dangerously` prefixes to prevent accidental usage in production code.

Additionally, I've modified `SharedHubStrategyTest` to use more hubs and samples, making random selection more reliable.


app/Platform/Services/GameSuggestions/GameSuggestionEngine.php
```diff
@@ -26,6 +27,14 @@ public function __construct(
         $this->initializeStrategies($user);
     }
 
+    /**
+     * For testing purposes only.
+     */
+    public function dangerouslyEnableFixedStrategyForTesting(): void
+    {
+        $this->unsafe_useFixedStrategyForTesting = true;
+    }
```
Jamiras, line 36: I have a feeling that you're going to want to specify an explicit strategy to test. Wouldn't it make more sense to have this function just set `$this->strategies` to a single strategy (possibly passed in)?

wescopeland: Yeah, I like that more. Updated in latest.

Jamiras (approved):

PR #3057, opened 2025-01-13
allow creating events directly from events list

Jamiras:
Removes the "create event" widget from the Admin page, and integrates the functionality into the Event console.


app/Filament/Resources/EventResource.php
```diff
@@ -105,6 +106,8 @@ public static function infolist(Infolist $infolist): Infolist
 
     public static function form(Form $form): Form
     {
+        $isNew = !is_a($form->model, Event::class);
```
gboquizosanchez, line 109: https://stackoverflow.com/questions/3017684/what-is-the-difference-between-is-a-and-instanceof

app/Filament/Resources/EventResource/Pages/Create.php
wescopeland, line 1: Recommend adding a `->placeholder()` to the "URL alias" field so the user has an example without needing to check another page first.

wescopeland, line 1: I'm wondering if it would be best to hide the sort order field in the form altogether. Seems like things have a high probability to go wrong with editing that field due to a misunderstanding of how it works.

wescopeland: Apologies for the confusion. I was going a little too fast as usual, typo'd, and was actually referring to the sort _title_ field:

Screenshot 2025-01-16 at 5 59 12 PM

wescopeland: 😅

I think it may be ambiguous for someone who is non technically sophisticated. The validator is indeed a great safeguard. For example, it keeps me from doing this:

Screenshot 2025-01-16 at 6 02 17 PM

But having a placeholder of something like "aotw-2025" is even better UX imo, as I don't need the safeguard to yell at me in order to understand the field. It also provides a good example of an alias up front that the helper text can't do.

This feedback is based on prior experience with fields that have helper text nearly identical to this one, where about 1 in 6 users will skim it and then try to enter "/events/foo-bar-baz".

Jamiras: added 'event-alias' as placeholder.

wescopeland: Yea, my main concern was the one on the game resource requires very elevated privs to edit, whereas I think that one was open to everyone with the role to access the page.

wescopeland (approved):

PR #3059, opened 2025-01-15
fix(BuildsGameListQueries): remediate issue with release date sorting

wescopeland:
This PR resolves a minor issue with release date sorting on game lists. It's currently possible for this scenario to occur while sorting:

Screenshot 2025-01-14 at 7 13 28 PM

Notice that something with a granularity of month is mixing with stuff that has a granularity of year. The year granularity rows should always come first.

This is occurring because these 3 rows all have the exact same 1978-01-01 release date and priority is not given to granularity values.


app/Platform/Concerns/BuildsGameListQueries.php
```diff
@@ -323,18 +323,25 @@ private function applyReleasedAtSorting(Builder $query, string $sortDirection =
         // and SQLite. This is preferable to altering the query specifically for
         // SQLite, because if we do so then we can't actually trust any test results.
         $query
-            ->selectRaw(
-                "GameData.*,
+            ->selectRaw(<<<SQL
```
wescopeland, line 326: heredoc syntax used for IDE syntax highlighting:
Screenshot 2025-01-14 at 7 15 37 PM

gboquizosanchez: Maybe...

```php
$query->select('GameData.*')
    ->selectRaw(<<<SQL
        CASE
            WHEN GameData.released_at_granularity = 'year' THEN
                DATE(CONCAT(SUBSTR(GameData.released_at, 1, 4), '-01-01'))
            WHEN GameData.released_at_granularity = 'month' THEN
                DATE(CONCAT(SUBSTR(GameData.released_at, 1, 7), '-01'))
            ELSE
                COALESCE(GameData.released_at, '9999-12-31')
        END AS normalized_released_at
    SQL)
    ->selectRaw(<<<SQL
        CASE GameData.released_at_granularity
            WHEN 'year' THEN 1
            WHEN 'month' THEN 2
            WHEN 'day' THEN 3
            ELSE 4
        END AS granularity_order
    SQL)
    ->orderBy('normalized_released_at', $sortDirection)
    ->orderBy('granularity_order', $sortDirection);

```

wescopeland: I'm not sure on this one - seems more complicated for me to read? Could just be me though 🤔

Jamiras: I'm pretty sure `select()` replaces any previous `select()` clause. I would assume `selectRaw` would do something similar. If we did want to break it up, we'd probably need to do something like:
```
->select(['GameData.*',
    DB::raw(<<<SQL
    ...
    SQL),
    DB::raw(<<<SQL
    ...
    SQL)])
```

Personally, I find the single raw SQL statement slightly easier to read just because I don't have to toggle back and forth between SQL syntax and Eloquent syntax.

app/Platform/Concerns/BuildsGameListQueries.php
```diff
@@ -323,18 +323,25 @@ private function applyReleasedAtSorting(Builder $query, string $sortDirection =
...
-                END AS normalized_released_at"
-            )
-            ->orderBy('normalized_released_at', $sortDirection);
+                END AS normalized_released_at,
+                CASE GameData.released_at_granularity
+                    WHEN 'year' THEN 1
+                    WHEN 'month' THEN 2
+                    WHEN 'day' THEN 3
+                    ELSE 4
+                END AS granularity_order
```
Jamiras, line 341: I was momentarily confused by this putting day-granular results in the middle of month-granular results:
image
Notice the day granular items in between the January and September monthly granularities?

Digging into it, it's caused by the primary sort on `normalized_released_at`:
image

Which makes sense - the granularity sort is just being used as a tie-breaker.

However, are there concerns about the day-granular values having timestamp information?

Jamiras (approved):

app/Platform/Concerns/BuildsGameListQueries.php
wescopeland: No concerns from me. In MariaDB, if the time isn't set, it's implied as 00:00:00. I'm not sure where the nuance is on how it's getting set - perhaps a difference from the sync script and Filament modifications.

Jamiras: It's from the query. When you construct the dates for the less granular options, you don't include the timestamp. The most granular option is read directly from the DB without any alterations.

wescopeland: Ah, I see what's happening now.

I do have a change locally which causes the query to always return timestamps:
Screenshot 2025-01-16 at 7 24 40 PM

but, if the query is currently causing results to be displayed in the expected fashion, I'll hang on to this in a stash unless we'd like to have it in `master`.

PR #3062, opened 2025-01-15
fix(connect): remediate allprogress endpoint exception

wescopeland:
Resolves https://discord.com/channels/476211979464343552/757767535293890682/1328924126114480149.

Converts the helper function to Eloquent ORM and adds coverage.


app/Helpers/database/player-game.php
```diff
@@ -276,20 +277,33 @@ function reactivateUserEventAchievements(User $user, array $userUnlocks): array
...
+        ->get()
+        ->keyBy('game_id');
 
-    foreach (legacyDbFetchAll($query) as $row) {
-        $id = $row['ID'];
-        unset($row['ID']);
+    /** @var Collection<int, Game> $games */
+    $games = Game::where('ConsoleID', $consoleID)
+        ->where('achievements_published', '>', 0)
+        ->get();
```
Jamiras, line 294: Would it make sense to reuse this query for the filter above? (psuedo-code incoming)
```
$playerGames = $user->playerGames()
    ->whereIn('game_id', $games->pluck('ID'))
```

app/Helpers/database/player-game.php
```diff
@@ -276,20 +277,33 @@ function reactivateUserEventAchievements(User $user, array $userUnlocks): array
...
+    /** @var array<int, array{NumAch: int, Earned: int, HCEarned: int}> $retVal */
+    $retVal = [];
+    foreach ($games as $game) {
+        /** @var ?PlayerGame $playerGame */
+        $playerGame = $playerGames->get($game->id);
+
+        $retVal[$game->id] = [
+            'NumAch' => $game->achievements_published,
+            'Earned' => $playerGame?->achievements_unlocked ?? 0,
+            'HCEarned' => $playerGame?->achievements_unlocked_hardcore ?? 0,
```
Jamiras, line 305: Since this API has been broken for an indeterminate amount of time, should we take the opportunity to rename these fields?

tests/Feature/Connect/AllProgressTest.php
Jamiras, line 1: Suggest having a test for a game where the user doesn't have progress, and one where "ItIgnoresGamesFromOtherConsoles"

app/Helpers/database/player-game.php
wescopeland: I think so. Maybe what would be best is to not alias them at all:
```json
'AchievementsPublished' => ...,
'AchievementsUnlocked' => ...,
'AchievementsUnlockedHardcore' => ...,
```

Jamiras: That might be too verbose. Remember that this API returns entries for every game on the system and several of our systems have more than 2000 entries.

Maybe `Achievements`, `Unlocked`, and `UnlockedHardcore`?

wescopeland: To be honest, I can't help but feel that if we're going to bring this endpoint back after it has been offline for so long, the current response shape seems kinda undesirable.

For example, if I set `c` to 12, I get 723 objects returned in the response. It may be better to just exclude values where the user has no progress. I'm going to optimistically assume this is okay and do this in latest. If you're against the change, I will roll it back.

Alternatively, if we actually do need all the game IDs in the response, maybe we can cut "Unlocked" and "UnlockedHardcore" if those values are 0.

tests/Feature/Connect/AllProgressTest.php
wescopeland: Added in latest.

app/Helpers/database/player-game.php
wescopeland: It would -- in latest I've actually condensed this down to a single query.

app/Helpers/database/player-game.php
Jamiras: > I'm going to optimistically assume this is okay and do this in latest.

I'm not sure exactly what the requirements are, but it seems like filtering is undesirable.
https://discord.com/channels/476211979464343552/757767535293890682/1321014964453904419

wescopeland: Gotcha. I've fallen back to this strategy in latest:
> Alternatively, if we actually do need all the game IDs in the response, maybe we can cut "Unlocked" and "UnlockedHardcore" if those values are 0.

Jamiras (approved):

PR #3018, opened 2025-01-04
add achievement checklist page

Jamiras:
Designed to allow users to check their progress towards a specific set of achievements. Can also be used to check another user's progress towards a specific set of achievements.

For example, here's the page for me and the first round of Challenge League 7: http://localhost:64000/user/Jamiras/achievement-checklist?list=Tiddleywink%3A255909%2C344447%2C406776%7CBathamption%3A81425%2C476492%2C7743%7CAlton%20Barnes%3A180396%2C282301%2C17471%7CMelksham%3A138052%2C360656%2C429119%7CWinterslow%3A56987%2C264417%2C172473%7CNettleton%3A28343%2C38335%2C383564%7CFittleton%3A468071%2C371883%2C366263%7CZeals%3A310782%2C347836%2C316173

image

And [some achievements](http://localhost:64000/viewtopic.php?t=24927&c=247936#247936) I did for the final board of last year's Challenge League: [with](http://localhost:64000/user/Jamiras/achievement-checklist?list=April%20Fool%27s%20Day%3A122685%7CPlayJam%3A416446%7CAotW%202023%3A59380%7CAotW%202022%3A191627) and [without labels](http://localhost:64000/user/Jamiras/achievement-checklist?list=122685%2C416446%2C59380%2C191627).

The basic format of the list parameter is a CSV of achievement IDs. A prefix can be added to the list to add a header: `Header:1,2,3`, and multiple groups can be joined with pipes: `Group1:1,2,3|Group2:4,5,6`. The browser (at least my browser) automatically escapes the comma, colon, and pipe characters, resulting in a valid URL that can be shared.

PR #3063, opened 2025-01-15
fix(UnlockableAchievementAvatar): swap `t()` for `<Trans />`

wescopeland:
Translators are reporting some trouble with the "from" string. This PR swaps `t('from')` with a `<Trans />` string that has the achievement title and game title set to "sr-only". This should provide them with the context they need without causing any disruption on the page.

Additionally, I implemented the TODO for showing the points in the title:
Screenshot 2025-01-15 at 8 33 17 AM


resources/js/features/achievements/components/UnlockableAchievementAvatar/UnlockableAchievementAvatar.tsx
Jamiras, line 1: `UnlockableAchievementAvatar` is currently only used by the `achievement-checklist` (#3018).

The event page is handled here:
https://github.com/RetroAchievements/RAWeb/blob/1d49bb99ecd795c17f8a0db01d47a452277df831/resources/views/components/game/achievements-list/achievements-list-item.blade.php#L84

And the event achievement is handled here (this one might be real tricky):
https://github.com/RetroAchievements/RAWeb/blob/1d49bb99ecd795c17f8a0db01d47a452277df831/resources/views/pages-legacy/achievementInfo.blade.php#L246

lang/en_US.json
```diff
@@ -613,5 +612,6 @@
     "Similar masteries": "Similar masteries",
     "Beaten by players of": "Beaten by players of",
     "Mastered by players of": "Mastered by players of",
-    "Roll again": "Roll again"
+    "Roll again": "Roll again",
+    "<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>": "<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>"
```
Jamiras, line 616: If I customize this, it doesn't seem to update the checklist page.

wescopeland: You should see an update if you try something like:
```json
"<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>": "<1>{{achievementTitle}}</1> 123from <2>{{gameTitle}}</2>"
```

Notice the "123from" in the value. The achievement title and game title are invisible.

{{achievementTitle}} and {{gameTitle}} come from the `values` prop:
```tsx
<Trans
  i18nKey="<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>"
  components={{
    1: <span className="sr-only" />,
    2: <span className="sr-only" />,
  }}
  values={{ achievementTitle: achievement.title, gameTitle: achievement.game.title }}
/>
```

Jamiras: Yes, I'm aware that I should just be able to change the right side.
```
    "<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>": "<1>{{achievementTitle}}</1> xxx <2>{{gameTitle}}</2>"
```

But I'm not seeing it in the UI.
image

Ideally, I'd like to test swapping the order: `"<2>{{gameTitle}}</2> featuring <1>{{achievementTitle}}</1>"`

wescopeland: In latest, the `<Trans />` component now manipulates the label placement and your example should work.

In the original implementation, it was still moving elements, but the elements were invisible:
Screenshot 2025-01-16 at 7 41 29 PM

Jamiras: Ah, figured it out. Need to run `npm run build` to get it to pick up the changes.

Jamiras (approved):
Approved for achievement checklist. Did you have any thoughts on the event page or event achievement page?

wescopeland:
There's probably not much we can do for localization on those pages until they're migrated to React.

Finishing up ripping out hubs from game pages will be a big step towards migration of game pages. Once that's done, I plan on doing a phased conversion of game pages to React very similar to what happened with the home page. First, a PR just with UI components containing hardcoded values. Then, a PR to connect real data to the page. Then, subsequent PRs to stamp out all the ancillary things that come up (handling events, component edge cases, etc).

PR #3065, opened 2025-01-16
feat: make modern hubs generally available

wescopeland:
This PR cuts over legacy hubs to the new Inertia.js hubs:
* All hub links on the game page now route directly to `/hub/{gameSet}`.
* Legacy hub URLs like `/game/{game}` 302 redirect to their corresponding modern hub page.
* Navbar hub links now route to either `/hubs/` or `/hub/{2-5}`.
* Game pages no longer query for game alts.
* The `FEATURE_ENABLE_MODERN_HUBS` env var has been removed.

Additionally, I've removed all legacy hub code from _gameInfo.blade.php_ (game list service stuff).

I've also relocated related events to the sidebar:
Screenshot 2025-01-15 at 8 21 07 PM

Next steps:
* Drop `GameAlternatives`.
* Update search to add "Hub" as a `SearchType` and break hub games out of search.

PR #3066, opened 2025-01-16
perf(site-award): query for game_sets in SeparateAwards()

wescopeland:
This PR migrates `SeparateAwards()` to use `game_sets` instead of `GameAlternatives`.

After #3065 is merged, to the best of my knowledge, this is the only place actually querying for `GameAlternatives`.

While working on this function, I couldn't help but notice that `SeparateAwards()` is wildly inefficient. With the changes in this PR, I was able to drop the number of queries on at least one developer's user profile by ~40.

This PR:
* Consolidates all the queries into a single optimized query using `whereIn`.
* Uses array flipping for O(1) lookups instead of repeated array searches.
* Pre-calculates award type checks to avoid redundant calls.
* Processes awards in a single execution pass wherever possible.
* Uses early returns to skip unnecessary processing for empty event awards.


wescopeland:
After https://github.com/RetroAchievements/RAWeb/pull/3067 is merged, I will fix this branch.

app/Helpers/render/site-award.php
```diff
@@ -3,38 +3,69 @@
...
+    $devEventGamesLookup = array_flip(GameSet::query()
+        ->whereHas('games', fn ($q) => $q->whereIn('game_id', array_column($eventAwards, 'AwardData')))
+        ->where(fn ($q) => $q->where('id', 5)->orWhere('title', 'like', '[Dev Events - %'))
+        ->with(['games' => fn ($q) => $q->select('GameData.ID as id')])
+        ->get()
+        ->pluck('games.*.id')
+        ->flatten()
+        ->unique()
+        ->values()
+        ->all());
```
Jamiras, line 47: I'm not sure what this is doing, but it generates `Array ( [1] => 0 )` for my user. I have two manually inserted event badges. One is appearing as a Site Award, and the other as an Event Award (both should be Event Awards). 

Similarly, all my dev badges migrated from Site Awards to Event Awards. This is most likely due to me starting work on converting the existing event games to actual event records. You might be able to simulate it with a simple query:
```
INSERT INTO events (legacy_game_id, image_asset_path) SELECT ID, ImageIcon FROM GameData WHERE ConsoleID=101;
```

Prod:
image

This branch:
image

The "Mario Paint" badge is the AotW 2025, and the Chrono Trigger badge is another test event badge.

Jamiras: For contrast, this is what my master looks like (it seems to be affected by the dev badges being event badges too - but only for the ones I've already migrated):
image

Jamiras (commented):
For my user, queries are down by 34, but models are more than doubled. This might be related to the aforementioned reclassification of some of my badges to events. Overall response down is down ~40ms.

app/Helpers/render/site-award.php
wescopeland: Thanks for pointing that out - I was indeed able to repro this issue on my local.

I've redone all the logic to keep the optimizations, continue using `game_sets`, but also ensure the award sections & ordering remain intact.

Jamiras (approved):

PR #3067, opened 2025-01-17
Automatically award event awards

Jamiras:
Based on achievements required for each tier. Or if no tiers are defined, having all the achievements in the event unlocked.

Supports reordering.

First image is first tier of AotW 2025. Second image is an event I created for #3057 that has been fully unlocked.

image

Here, the AotW badge has been upgraded by manually unlocking a bunch more achievements.

image

Clicking on the badges takes you to the associated event page.

The tooltips could probably use some flair-up (maybe their own card), but I'll leave that to someone with more UX experience.


app/Helpers/render/site-award.php
```diff
@@ -197,6 +199,39 @@ function RenderAward(array $award, int $imageSize, string $ownerUsername, bool $
...
+
+                    if ($eventAward->achievements_required < $event->game->achievements_published) {
+                        $tooltip = "Awarded for earning at least {$eventAward->achievements_required} achievements in the {$event->title} event";
+                    }
+                }
+            }
+
+            echo avatar('event', $event->id,
+                link: route('game.show', $event->game->id),
+                tooltip: "<div class='px-2 py-2'><span>$tooltip</span><p><span>{$awardDate}</span></div>",
```
wescopeland, line 224: Not overly concerned with UX at the moment - this will all get migrated to React later anyway. This may be a marginal improvement though:
```php
                tooltip: "<div class='p-2 max-w-[320px] text-pretty'><span>{$tooltip}</span><p class='italic'>{$awardDate}</p></div>",
```
Changes:
* Combines "px-2 py-2" into "p-2" (p-* applies padding on all 4 sides).
* Adds a nominal max width and text balancing class.
* Fixes an invalid HTML issue (missing a closing p tag).
* Italicizes the date.

Screenshot 2025-01-16 at 9 21 49 PM

We can take some of these, all of these, or none of these. Just suggestions!

wescopeland (approved):

app/Helpers/render/site-award.php
Jamiras: Applied suggested changes. Also remove timestamp from date (it doesn't appear in any other badge tooltips)

wescopeland (approved):

PR #3069, opened 2025-01-17
Allow event managers to unlock event achievements

Jamiras:
Adds a drop-down menu to the event achievements list on the event admin page, which has an Unlock menu option.
image

This opens a dialog where the user can enter a CSV of usernames.
image

For events, a single achievement is usually going to be awarded to a number of users due to the events only being active for certain periods of time and bookkeeping for multiple users has to be applied to each active achievement. The existing manual unlock tool allows unlocking multiple achievements but only for a single user.

Upon hitting Submit, jobs are kicked off to award the achievements. If the user already has the achievement, the job will do nothing, otherwise, it will add a manual unlock for the user for the achievement. Unknown users are reported in an error notification, but the user may not get all the names from it before it disappears.

image


gboquizosanchez:
Consider [this](https://filamentphp.com/docs/3.x/forms/fields/tags-input/) in your feature.

app/Filament/Resources/EventResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -77,7 +82,68 @@ public function table(Table $table): Table
...
+                            $timestamp = Carbon::now();
+                            $lastFoundUser = null;
+                            foreach (explode(',', $data['users']) as $username) {
+                                $username = trim($username);
+                                if (empty($username)) {
+                                    continue;
+                                }
+
+                                // TODO: whereName
+                                $forUser = User::where('User', $username)->first();
```
wescopeland, line 113: I'm sure it's already on your radar, just leaving a breadcrumb here that we should be good to cut over to `whereName` now.

wescopeland (approved):
Verified locally; I also wasn't able to break it regardless of my efforts.

Notes:
* I have no opinion on @gboquizosanchez's suggestion of using a `TagsInput` control, as this is primarily for back-office use and will (hopefully) not be invoked very often.
* This action currently isn't policy-gated, so the implied policy is `EventPolicy::manage()`. Given that any underlying achievement isn't unlocked, this is probably okay, but still feels worth noting nonetheless.
* In the future, if the inner function within `->action()` grows in complexity, it may be worth breaking the logic out to a separate action class with an `->execute()` method.

Jamiras:
> In the future, if the inner function within ->action() grows in complexity, it may be worth breaking the logic out to a separate action class with an ->execute() method. This would have the benefit of giving the logic testability w/ PHPUnit.

I had considered that, but half the function is just showing the success/failure messages.

> I have no opinion on @gboquizosanchez's suggestion of using a TagsInput control, as this is primarily for back-office use and will (hopefully) not be invoked very often.

`->separator()` is a function on `TagsInput`. I can't seem to make it work on `Textarea`.

It would be nice to have a regex validate that it's a CSV of valid username characters, but the error message for not matching a regex is not helpful ("format is invalid").

PR #3072, opened 2025-01-18
fix(EditPostForm): resolve untranslated string

wescopeland:
The `placeholder` prop currently uses hardcoded text.


Jamiras (approved):
Not sure if you want to try to rush something in, but the page for creating new forum posts is still using the old UI, so the likelihood of someone seeing this watermark is pretty ow.

wescopeland:
No critical rush on my end - just wanted to make sure this string was delivered to the localization team as soon as I noticed the issue.

PR #3074, opened 2025-01-18
shortcode support for [quote]

Jamiras:
Provides a separate [quote] shortcode tag, so the user doesn't have to use [code] or [i] as a substitute.

Before:
image

After:
image

Nested formatting example (though copying text to use in a quote won't copy the formatting tags):
image

There are already a [few](https://retroachievements.org/viewtopic.php?t=2508&c=13060#13060) [examples](https://retroachievements.org/viewtopic.php?t=609&c=7686#7686) [of](https://retroachievements.org/viewtopic.php?t=16757) people trying to use it in the database. Was it maybe sort of supported at some point in the past?


wescopeland:
I was hoping to pick this off, it seems you beat me to it :-)

For quote support, I think something like this is desirable:
```
[quote author=Dk_madness comment={forum-topic-comment-id}]
this is the quote
[/quote]
```

This allows us to standardize authorship and linkbacks in a dedicated "quote" button on posts. In other words, if I want to quote a post, I will click a new "quote" button on the post itself, which starts an editing flow with these values pre-filled. I don't think we should support that interaction right now, as I'm in the process of migrating the forum to Filament, but we should at least support `author` and probably `comment` props in the shortcode.

Jamiras:
Correct me if I'm wrong, but do any of the existing tags support properties? Everything is either `[tag]content[/tag]` or `[tag=value]content[/tag]`.

I would expect those extra properties would be optional, as the user might want to quote a source that's not another forum post.

wescopeland:
You're correct - none of them do. You make a good point.

Given that those values should be optional, we can add them later.

app/Support/Shortcode/Shortcode.php
wescopeland, line 1: Comment unrelated to this file; I wasn't sure where it might best be placed.

I'm unsure of where the nuance is, but [quote] and [code] are inconsistent in how they treat whitespace:
Screenshot 2025-01-18 at 9 11 16 PM

I can repro the issue in both the PHP UI and the React UI.

I can help hunt this down if it proves overly cumbersome to solve.

resources/css/devbox.css
```diff
@@ -24,6 +24,15 @@
...
 }
 
+.quotedtext {
+  border-left: 3px solid var(--text-color-muted);
+  padding: 4px 6px 5px 8px;
+  margin: 3px 0px;
+  background: var(--box-bg-color);
+  width: 100%;
+  display: inline-block;
+}
```
wescopeland, line 34: There's a minor performance benefit to putting this within the `@layer components` block in _resources/css/utilities.css_, namely, we can reach for Tailwind's sane defaults and benefit from the Tailwind compiler's dead code elimination ([docs](https://tailwindcss.com/docs/functions-and-directives)).

I recommend:
```css
@layer components {
  .quotedtext {
    @apply border-l-[3px] border-text-muted;
    @apply pt-1 pb-1.5 pr-1.5 pl-2;
    @apply my-2;
    @apply bg-box-bg;
    @apply w-full;
    @apply inline-block;
  }
```

resources/js/features/forums/components/ShortcodeRenderer/ShortcodeQuote/ShortcodeQuote.tsx
wescopeland, line 1: Quotes within quotes work great in the React UI, but break in the PHP UI. I consider this non-blocking feedback, just commenting about it here for awareness. This will naturally fix itself soon.

Screenshot 2025-01-18 at 9 24 54 PM

Screenshot above is the PHP UI with the input:
```
[quote]
this is my awesome quote

https://www.youtube.com/watch?v=508wu-JxT2A

asdf asdf hausdfh ausdfh ausdhf a

[s]asdffoooofdasasd fasdf asdf[/s]

[user=Scott]

[quote]how about this!!

[quote][quote][quote][quote]one more for science


asdfasdfasdfa

asdfasdf[/quote]anotasdfasdf
a
dsf
asdf
as








df
af
f
fdsa asdf[/quote]woooo[/quote]again!!!![/quote]
[/quote]
[/quote]
```

app/Support/Shortcode/Shortcode.php
wescopeland: I can help. This seems to work on my local:
```tsx
export const ShortcodeQuote: FC<ShortcodeQuoteProps> = ({ children }) => {
  // Remove leading <br>s and empty strings until we find content.
  const processedChildren = (Array.isArray(children) ? children : [children]).filter((node) => {
    const isObjectNode = typeof node === 'object' && node !== null;
    const isEmptyObject = isObjectNode && !Object.keys(node as object).length;
    const isBRElement = isObjectNode && 'type' in node && node.type === 'br';
    const isEmptyString = typeof node === 'string' && node.trim() === '';

    const isLeadingWhitespace = isEmptyObject || isBRElement || isEmptyString;

    // If it isn't leading whitespace, it can stay.
    // Otherwise, it's filtered out.
    return !isLeadingWhitespace;
  });

  return <span className="quotedtext">{processedChildren}</span>;
};
```

This test should be able to cover it:
```tsx
it('removes leading line breaks', () => {
  // ARRANGE
  render(
    <ShortcodeQuote>
      <br />
      test content
    </ShortcodeQuote>,
  );

  // ASSERT
  const spanEl = screen.getByText(/test content/i);
  expect(spanEl.firstChild?.nodeType).not.toBe(Node.ELEMENT_NODE);
});
```

Jamiras: Thanks. But this strips <br>s in the middle too. I've made adjustments.

I see something similar in the `ShortcodeCode`. I'm not sure why that's not working for me. I've applied the changes to both.

resources/js/features/forums/components/ShortcodeRenderer/ShortcodeQuote/ShortcodeQuote.tsx
```diff
@@ -0,0 +1,26 @@
+import type { FC, ReactNode } from 'react';
+
+interface ShortcodeQuoteProps {
+  children: ReactNode;
+}
+
+export const ShortcodeQuote: FC<ShortcodeQuoteProps> = ({ children }) => {
+  // Remove leading <br>s and empty strings until we find content.
+  var isLeadingWhitespace = true;
```
wescopeland, line 9: nit: Recommend using `let`.
```suggestion
  let isLeadingWhitespace = true;
```
This is supposed to be caught by a linting rule ... I will investigate why it isn't later today.

wescopeland: Sorry, I didn't provide context. `var` has generally fallen out of favor in modern JS/TS because it doesn't respect block-scoping rules.

app/Support/Shortcode/Shortcode.php
wescopeland: Whitespace between the React UI and PHP UI now looks considerably different. I don't think it's related to these changes 🤔 

**React Preview**
Screenshot 2025-01-19 at 1 50 17 PM

**PHP UI**
Screenshot 2025-01-19 at 1 50 26 PM

Looks like whitespace is considerably different. Here's the input I used:
```
bbbbbbbbbb

[quote][/quote]

[code]asdfasdfa



[/code]


[quote]more stuff[/quote]

[quote]
stuff
[/quote]

[quote]
stuff



[/quote]

[quote]

stuff


[/quote]
```

Jamiras: It's not. I can reproduce on master by changing all your `quote` tags to `code` tags. The odd thing is that if I "Preview" without modifying it (i.e. save and re-edit), then it shows up fine. It's only after editing and hitting preview that the extra space between elements disappears.

wescopeland: Currently investigating on my local, will mark resolved as it's a bug unrelated to this PR.

wescopeland:
Sorry, not sure why this was closed. I think GitHub did it automatically?

wescopeland (approved):
Poked around a bit with the latest changes. I think things are looking good now 👍

Jamiras:
Yep.

Closed by 3083
image

Which says it resolves a comment here.
image

Surprised it did it for a comment reference though.

PR #3083, opened 2025-01-19
fix(forums): normalize newlines from synthetic events

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/pull/3074#discussion_r1921615916.

Also, adds margin bottom to code tags (we should probably also apply the same margin bottom to quote tags).

**Root Cause**
React uses its own synthetic event system to maximize browser compatibility. For example, when a component has `onClick={(event) => {...}}`, this is not a native `onclick="..."` event. It's running through React's synthetic event system that _emulates_ native events.

I've worked with React for a number of years and haven't hit a bug caused by synthetic events until now.

In our case, newlines were being sent to the preview component as escaped sequences (↵\n) after making any edit (ie: when initiating a synthetic event), but are served as actual newlines (\n) on load. I was able to observe this after putting a ton of logging in all the BBCode processors.

This fix normalizes all line endings to actual newlines (\n) during the preprocessing stage, ensuring consistent handling of whitespace regardless of how the newlines are represented in the input.

### DIFF
--- a/resources/js/features/forums/components/ShortcodeRenderer/ShortcodeCode/ShortcodeCode.tsx
+++ b/resources/js/features/forums/components/ShortcodeRenderer/ShortcodeCode/ShortcodeCode.tsx
@@ -15,8 +15,8 @@ export const ShortcodeCode: FC<ShortcodeCodeProps> = ({ children }) => {
       return true;
     });
 
-    return <span className="codetags font-mono">{processedChildren}</span>;
+    return <span className="codetags mb-3 font-mono">{processedChildren}</span>;
   }
 
-  return <span className="codetags">{children}</span>;
+  return <span className="codetags mb-3 font-mono">{children}</span>;
 };
--- a/resources/js/features/forums/components/ShortcodeRenderer/ShortcodeCode/ShortcodeCode.tsx
+++ b/resources/js/features/forums/components/ShortcodeRenderer/ShortcodeCode/ShortcodeCode.tsx
@@ -15,8 +15,8 @@ export const ShortcodeCode: FC<ShortcodeCodeProps> = ({ children }) => {
       return true;
     });
 
-    return <span className="codetags mb-3 font-mono">{processedChildren}</span>;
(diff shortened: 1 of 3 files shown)


resources/js/features/forums/components/ShortcodeRenderer/ShortcodeCode/ShortcodeCode.tsx
```diff
@@ -15,8 +15,8 @@ export const ShortcodeCode: FC<ShortcodeCodeProps> = ({ children }) => {
       return true;
     });
 
-    return <span className="codetags font-mono">{processedChildren}</span>;
+    return <span className="codetags mb-3 font-mono">{processedChildren}</span>;
   }
 
-  return <span className="codetags">{children}</span>;
+  return <span className="codetags mb-3 font-mono">{children}</span>;
```
Jamiras, line 21: I've applied the `mb-3` to the #3074 PR (for both this and ShortcodeQuotes. You should remove it from here to eliminate the merge conflict (I've touched this file for trimming the leading <br>.

resources/js/features/forums/utils/preProcessShortcodesInBody.ts
```diff
@@ -30,7 +30,11 @@ const createPatterns = (type: string) => [
 ];
 
 export function preProcessShortcodesInBody(body: string): string {
-  let result = body;
+  // First, normalize any escaped newlines back to actual newlines.
+  let result = body.replace(/↵\n/g, '\n');
```
Jamiras, line 34: A unicode character for escaping a newline? I've never seen that before.

Should you use an escape sequence for it (`\u21B5`) to prevent possible encoding issues?

Jamiras (approved):

resources/js/features/forums/utils/preProcessShortcodesInBody.ts
wescopeland: No harm in adding an escape sequence. Updated in latest - I'll go ahead and merge so the changes in this branch can safely land in #3074.

PR #3084, opened 2025-01-19
feat: add hub shortcode

wescopeland:
Related to https://github.com/RetroAchievements/RAWeb/pull/3065 and should likely deploy simultaneously.

This PR adds a `[hub=N]` shortcode for both the forums and messages, as well as an Artisan command to migrate legacy hub game shortcodes to the new `[hub]` shortcode.

**How to test this PR:**
```
sail artisan ra:community:migrate-game-shortcodes
```
This will convert all existing legacy hub game shortcodes to the new hub shortcode.
You can also roll it back with:
```
sail artisan ra:community:migrate-game-shortcodes --undo
```

This command is largely based on the one we used for user IDs.

Now, you should be able to check any forum topic that uses a hub shortcode, ie: http://localhost:64000/viewtopic.php?t=1287.

You can also try using and previewing the hub shortcode directly with something like `[hub=1]`.

Currently, the new hub shortcodes look and feel exactly like the existing legacy game shortcodes.

Screenshot 2025-01-19 at 6 29 42 PM

Screenshot 2025-01-19 at 6 29 54 PM


app/Community/Commands/ConvertGameShortcodesToHubs.php
```diff
@@ -0,0 +1,267 @@
...
+                }, $message->body);
+
+                // Update the message in the DB only if it has actually changed.
+                // Use `DB` so we don't change the `updated_at` value.
+                if ($originalBody !== $updatedBody) {
+                    DB::table('messages')
+                        ->where('id', $message->id)
+                        ->update([
+                            'body' => $updatedBody,
+                            'updated_at' => DB::raw('updated_at'), // don't change the value
```
Jamiras, line 150:
```
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'updated_at' in 'SET' (Connection: mysql, SQL: update `messages` set `body` = ... , `updated_at` = updated_at where `id` = 19)
```
Is there another PR that adds/removes this column and I'm just out of sync?

Jamiras: Also, why use a direct query to update a single record?
```
$message->body = $updatedBody;
$message->timestamps = false;
$message->save();
```
(same comment for forums topic comments above)

Jamiras (commented):
I've been holding off reviewing this because of the conflict. Did you want to try to get this in?

app/Community/Commands/ConvertGameShortcodesToHubs.php
wescopeland: > Is there another PR that adds/removes this column and I'm just out of sync?

Nope, this is a bad copy+paste job after fixing a bug in the command.

> Also, why use a direct query to update a single record?

I have no idea why I landed on this solution anymore. I tried `->timestamps` and for some reason, found on my local it wasn't working. Now it is. I must have had a logic error somewhere.

Should all be fixed up now in latest (along with the merge conflict resolved).

app/Community/Commands/ConvertGameShortcodesToHubs.php
```diff
@@ -0,0 +1,253 @@
...
+
+                    return $matches[0];
+                }, $forumTopicComment->body);
+
+                // Update the comment in the DB only if it has actually changed.
+                // Use `DB` so we don't change the `updated_at` value.
+                if ($originalPayload !== $updatedPayload) {
+                    $forumTopicComment->body = $updatedPayload;
+                    $forumTopicComment->timestamps = false;
+                    $forumTopicComment->save();
```
Jamiras, line 85: `timestamps=false` didn't seem to work:
image

wescopeland: So strange - I didn't see it working on the original iteration and did see it working after I swapped over.

Reverted to the `DB::` calls in latest.

Jamiras (approved):
If this wasn't a one-time command, I'd recommend pursuing the `timestamps=false` route. Really confused why it wasn't working.

PR #3085, opened 2025-01-20
hide future event achievement selection

Jamiras:
Allows event managers to assign an achievement prior to it becoming active without exposing it to players until it does become active.

https://discord.com/channels/310192285306454017/1328443350889795635/1330687821429608500

If an event achievement is associated to a source achievement, it's title and description and image will be obscured on the event page and achievement page.

image
image

There are still ways to get to the unobscured data, but this should be sufficient for the largest portion of the playerbase.

### DIFF
--- a/app/Support/Shortcode/Shortcode.php
+++ b/app/Support/Shortcode/Shortcode.php
@@ -345,9 +345,9 @@ final class Shortcode
 
         if ($data['ConsoleID'] === System::Events) {
             $achievement = Achievement::find($id);
-            if ($achievement->eventData?->source_achievement_id &&
-                $achievement->eventData?->active_from > Carbon::now()) {
-                $data['Title'] = $data['AchievementTitle'] = '?????';
+            if ($achievement->eventData?->source_achievement_id
+                && $achievement->eventData->active_from > Carbon::now()) {
+                $data['Title'] = $data['AchievementTitle'] = 'Upcoming Challenge';
                 $data['Description'] = '?????';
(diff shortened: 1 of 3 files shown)


wescopeland:
https://discord.com/channels/310192285306454017/1328443350889795635/1330694768044277783

I could be missing something, but I think it would make more sense to consistently show "Week N" for all future achievements, regardless of whether an underlying achievement has been selected internally or not.

The internal state (whether the events team has picked an achievement) seems like it shouldn't affect what end-users see before the go-live date. Using "?????" could be confusing for users who might not know if it's intentional, a bug, or something else.

From a user's perspective, this has the benefit of all future weeks looking consistent until the go-live date has been reached.

Jamiras:
"Week N" is just the placeholder prepopulated by the `ra:platform:event:create-aotw` command. When an actual achievement is selected, it's properties get copied to the event achievement, overwriting the "Week N"

wescopeland:
I see. It's actually a pretty good placeholder, to be honest.

I've verified the change locally, but my reservations about the "?????" label still give me pause. This could be too ambiguous. How about any of these alternatives?:

* Coming Soon
* Future Challenge
* Future Achievement
* Upcoming Challenge

Jamiras:
Changed Title to "Upcoming Challenge" per popular vote: https://discord.com/channels/310192285306454017/1328443350889795635/1330716794255839262
Left Description as "?????"

wescopeland (approved):

PR #3087, opened 2025-01-21
feat(user): fully support the Visible Role field

wescopeland:
This PR adds full support for the Visible Role field in a user's settings. A migration has been included which adds a `UserAccounts.visible_role_id` field.

**How it works:**
Given a user has 2 or more roles attached, they may select which one is publicly visible from their settings:
Screenshot 2025-01-20 at 7 57 23 PM

Once a `visible_role_id` is set, it is respected on the user profile, forum author box, and user card tooltips:
Screenshot 2025-01-20 at 7 57 35 PM

Screenshot 2025-01-20 at 7 57 45 PM

Screenshot 2025-01-20 at 7 57 58 PM

**Implementation details:**
The meat of this PR is `ActsAsCommunityMember::getVisibleRoleAttribute()`.

`visible_role_id` is nullable. By default, `getVisibleRoleAttribute()` is going to return the user's first `display`-ordered role. For the great majority of folks who have roles, this will either be their Developer or Junior Developer roles.

If the user has a defined `visible_role_id` value, but they've since lost the role, then the attribute falls back to the first `display`-ordered role they have attached, or just null.

We'll obviously need to do some kind of mass assignment of team roles after deploy. I can figure out those details later.


app/Community/Concerns/ActsAsCommunityMember.php
```diff
@@ -28,6 +29,30 @@ public static function bootActsAsCommunityMember(): void
...
+            $explicitRole = $displayableRoles->find($this->visible_role_id);
+            if ($explicitRole) {
+                return $explicitRole;
+            }
+        }
+
+        // Otherwise, fall back to highest ordered displayable role.
+        // For most users, this will return null.
+        return $displayableRoles->first();
+    }
```
Jamiras, line 54: Not sure exactly what's going on here, but it looks like this query gets run 6 times (for like 60% of the page time).
image

app/Community/Controllers/UserSettingsController.php
```diff
@@ -54,7 +57,18 @@ public function show(): InertiaResponse
             'updateMotto'
         );
 
-        $props = new UserSettingsPagePropsData($userSettings, $can);
+        /** @var Collection<int, Role> $displayableRoles */
+        $displayableRoles = $user->displayableRoles()->orderBy('display')->orderBy('name')->get();
```
Jamiras, line 61: Would it make sense to just order these alphabetically instead of by display order? When I first saw the dropdown I assumed it wasn't sorted at all.

app/Community/Concerns/ActsAsCommunityMember.php
wescopeland: Fixed in latest.

app/Community/Controllers/UserSettingsController.php
wescopeland: Yeah, I like that approach more. Updated in latest.

Jamiras (approved):

wescopeland:
Good point - probably never.

I'm pretty sure each admin _should_ have more specific roles under the current role offerings we have. For example, you and I (and anyone else with write access to our GitHub org) should probably have the "Engineer" role.

PR #3103, opened 2025-01-23
don't show Awarded for an event unless an event badge has been awarded

Jamiras:
https://discord.com/channels/310192285306454017/1017568867574362283/1331997952088211489

When we redo this page in react, it would be nice to show which tier was awarded.


wescopeland (approved):
> When we redo this page in react, it would be nice to show which tier was awarded.

Agreed. If we've nearly settled on feature+function for event pages, I will get started on this soon.

PR #3113, opened 2025-01-24
feat(messages): migrate message threads index page to React

wescopeland:
This PR migrates the user messages index page to React.

Similar to previous PRs, both the legacy PHP UI and React UI are present in this branch. A subsequent small PR will be opened afterwards to remove the PHP UI.

**How to test this PR:**
http://localhost:64000/messages
http://localhost:64000/messages2

Compare and contrast.

The mobile UI has also been improved to reduce the need for horizontal scrolling.

Screenshot 2025-01-23 at 9 11 25 PM


resources/js/pages/messages.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type { AppPage } from '@/common/models';
+import { MessagesRoot } from '@/features/messages/components/+root';
+
+const Messages: AppPage = () => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <SEO title={t('Messages')} description="Your inbox messages" />
+
```
Jamiras, line 14: Would it be appropriate to include a user breadcrumb?

We'd definitely want some sort of discriminator if we open up the team account message lists in the future.

resources/js/features/messages/components/+root/MessagesRoot.tsx
```diff
@@ -0,0 +1,50 @@
...
+          })}
+        </p>
+      </div>
+
+      <div className="flex w-full flex-col items-end justify-between gap-2 sm:flex-row sm:items-center">
+        <a href={route('message.create')} className={baseButtonVariants({ size: 'sm' })}>
+          {t('New Message')}
+        </a>
+
+        <FullPaginator paginatedData={paginatedMessageThreads} />
```
Jamiras, line 34: The dropdown to select pages doesn't seem to work. The arrow navigation does.

resources/js/pages/messages.tsx
wescopeland: I think a user breadcrumb makes sense as a way to potentially solve this problem. Added in latest:

Screenshot 2025-01-25 at 2 59 17 PM

resources/js/features/messages/components/+root/MessagesRoot.tsx
wescopeland: Fixed in latest.

resources/js/pages/messages.tsx
Jamiras: I was kind of expecting "All Users > User > Messages", but this does seem consistent (at least for now) with the old messages that are just "Messages > New Message" or "Messages > Subject". Maybe we can revisit when they get converted.

Jamiras (approved):

resources/js/pages/messages.tsx
wescopeland: I did waffle around on this a bit with the same thought.

Ultimately, I was afraid All Users as the base nav could be an accessibility issue, as it may imply I can navigate to N users' inboxes, whereas realistically for 99.9% of users this is going to be completely locked down.

PR #3114, opened 2025-01-24
fix associating source achievement to event achievement

Jamiras:


wescopeland (approved):
Verified the observer changes locally.

PR #3121, opened 2025-01-24
fix(mail): remediate stable username leak

wescopeland:
Resolves an issue where, if I comment on my own tickets, I'll receive an email which also leaks my stable username:

Screenshot 2025-01-24 at 6 30 23 PM


app/Helpers/util/mail.php
```diff
@@ -286,7 +286,17 @@ function informAllSubscribersAboutActivity(
         $isThirdParty = ($subscriber['User'] != $activityAuthor && ($subjectAuthor === null || $subscriber['User'] != $subjectAuthor));
 
         if (isset($subscriber['EmailAddress'])) {
-            sendActivityEmail($subscriber['User'], $subscriber['EmailAddress'], $articleID, $activityAuthor, $articleType, $articleTitle, $urlTarget, $isThirdParty, $payload);
+            sendActivityEmail(
+                isset($subscriber['display_name']) ? $subscriber['display_name'] : $subscriber['User'],
```
Jamiras, line 290: I think it's important to note that this fix doesn't correct the email being sent to the user. Instead, it correctly fixes the filter that detects the email is being sent to the originator as the originator should not receive the email.
https://github.com/RetroAchievements/RAWeb/blob/954201bd9259d2fbbc216425ce9fbe55b7c9d109/app/Helpers/util/mail.php#L305

Jamiras (approved):

app/Helpers/util/mail.php
wescopeland: Yes, I was a bit ambiguous in my wording in the PR description.

The email shouldn't be sent in the first place, but if someday were supposed to be, the issue would resurface.

PR #3125, opened 2025-01-25
feat(manage): add has_box_art filter to GameResource

wescopeland:
Adds a new filter for games in the management app: "Has box art". This is a feature requested by the site cleanup team.

Screenshot 2025-01-24 at 9 31 35 PM

Screenshot 2025-01-24 at 9 31 29 PM


app/Filament/Resources/GameResource.php
```diff
@@ -496,6 +496,25 @@ public static function table(Table $table): Table
...
+                            ->whereNotIn('ConsoleID', System::getNonGameSystems())
+                            ->where('ImageBoxArt', '!=', '/Images/000002.png'),
+                        false: fn (Builder $query): Builder => $query
+                            ->whereNotIn('ConsoleID', System::getNonGameSystems())
+                            ->where(fn (Builder $query): Builder => $query
+                                ->whereNull('ImageBoxArt')
+                                ->orWhere('ImageBoxArt', '/Images/000002.png')
+                            ),
+                        blank: fn (Builder $query): Builder => $query,
+                    ),
```
Jamiras, line 517: Should this be restricted to just box art? I have a local script that I run that returns a handful of random games that are missing _any_ media (box art, icon, title/in-game image). I try to add media (particularly icons) to at least 3 or 4 games a week. A game is more likely to attract a developer from media than its title.

I can definitely see the use case of finding games with published sets that are missing box art, but isn't it equally important to see if they're missing other media too?

Outside of that particular use case, I think the other thing that would make this feature useful is the ability to filter on active systems. Because of the way IDs were recycled, selecting "no box art" with the default filters provides 20ish Wii games, then a couple Dreamcast games, then 20 ZX spectrum games before getting into systems that are easily accessible for players.

wescopeland: I think this should all be covered in latest. Here's where I've landed:

Screenshot 2025-01-25 at 3 43 04 PM

There's a new system filter option for only active systems. Additionally, new filters have been added for each media kind. These things can all be mixed and matched.

Jamiras: That many menu options is undesirable for something that won't be used very frequently, or in all the available combinations. Can we come up with good descriptors for a single option that would have multiple levels of filter?
```
Missing Media
* Don't Care          <- This needs a better name
* None
* All                 <- Is this desirable, or is "Any" good enough?
* Any
* Badge
* Box Art             <- Is this really desirable by itself?
```

wescopeland: Here's where I've landed in latest: a single multi-select filter with all the media options we currently support. The user can pick 0 to N choices.

Screenshot 2025-01-26 at 10 34 51 AM

.github/workflows/ci.yml
wescopeland, line 1: GitHub Actions decided to be combative again with pnpm, this time related to the cache.

To force it to pass, I removed the node-setup step and have the 3 Node.js tasks each do their own pnpm installs. Interestingly, this executes so fast it's actually faster than having a node-setup step.

app/Filament/Resources/GameResource.php
Jamiras: The default option "All" implies having the other four selected. It should be "None", but in reality it's still the "Don't Care" option, so how do I find stuff that isn't missing any media?

wescopeland: Here's where I've landed in latest. It's now back to a single-select modeled after the options you've provided in your feedback:

Screenshot 2025-01-29 at 7 32 14 PM

Screenshot 2025-01-29 at 7 32 18 PM

Jamiras: I'm still not sold on the "Select a Value". It makes it feel like a required filter the first time you see it, and it's not obvious that it'll clear the filter if you had another value selected.

Maybe "Unfiltered"?

Jamiras (approved):

PR #3126, opened 2025-01-25
feat(hub): allow some roles to create new hubs

wescopeland:
Depends on https://github.com/RetroAchievements/RAWeb/pull/3065 and https://github.com/RetroAchievements/RAWeb/pull/3084.

Allows admins, game hash managers, and game editors to create new hubs.

Screenshot 2025-01-24 at 9 36 52 PM

Screenshot 2025-01-24 at 9 36 58 PM


app/Policies/GameSetPolicy.php
```diff
@@ -37,13 +37,11 @@ public function view(?User $user, GameSet $gameSet): bool
...
-        // TODO enable after dropping GameAlternatives
-        // return $user->hasAnyRole([
-        //     Role::ADMINISTRATOR,
-        //     Role::GAME_EDITOR,
-        // ]);
+        return $user->hasAnyRole([
+            Role::ADMINISTRATOR,
+            Role::GAME_HASH_MANAGER,
+            Role::GAME_EDITOR,
+        ]);
```
Jamiras, line 44: I'm not sure if we should have a separate role, or if this should be opened more.

I know we have hubs dedicated to events, and there's all the `Meta|**` hubs that are used by other teams.

Jamiras (approved):

app/Policies/GameSetPolicy.php
wescopeland: I'm not too sure either. I don't think this should be opened up to events managers or the 3 staff dev roles. Right now I'm inclined to think if any of those folks regularly find themselves creating new hubs, they should be given the game editor role, otherwise they can ask one of these 3 roles who are ultimately responsible for the curation of the hubs.

Issue #3116, opened 2025-01-24
Release Date Sorting Inconsistency


Hoff16:
**Describe the bug**
When I sort games with the new UI framework when a release date is unknown it either appears at the beginning or at the end of the list.

**To Reproduce**
1. Go to Want to Play Games page
2. Click on Customize View, sort by release date earliest or latest
3. See unknown games at either end

**Expected behavior**
I personally expect unknown games to always appear at the latest date possible.

**Screenshots**
![image](https://github.com/user-attachments/assets/960cdcdd-ff08-4e76-8694-49e57da693c1)
![image](https://github.com/user-attachments/assets/4246357b-1c9b-461b-8d07-14bf1c6ccd48)

**Smartphone (please complete the following information):**
 - Device: iPhone 13 Pro Max
 - OS: iOS 18.2.1
 - Browser Chrome
 - Version 132.0.6834.100

bhimbho:
I have thrown in a possible fix for this [#3127](https://github.com/RetroAchievements/RAWeb/pull/3127)

wescopeland:
Resolved by https://github.com/RetroAchievements/RAWeb/pull/3127.

Issue #2957, opened 2024-12-19
Null values should be beneath non-null values when sorting by a column


hb-robo:
**Describe the bug**
Not sure if this is a bug or just a feature request, but when sorting by columns in Games List tables, null values appear as the top result when sorting in descending order in that column.
For example, I want to see the achievement set for the game released most recently. I sort by release date descending, but null values are above recent dates, obstructing the search.

**To Reproduce**
1. Go to 'All Games' page or any other page with a games list table
2. Click on 'Release date' column header and then click 'Latest'
3. See error/behavior

**Expected behavior**
I expected the most recent non-null date to be the top of the list, essentially the behavior of a "NULLS LAST" clause in SQL.

**Desktop:**
 - OS: Windows 10
 - Browser: Firefox
 - Version 133.0.3 64-bit (most recent build)

**Additional context**
Arguably, this behavior could be said to apply for some 0 values in numeric columns. For example, all sets without achievements have 0 points to earn, which is sensible. However, in a view where I am looking through games both with and without achievements, I see games without sets before I see [the singular achievement set with 0 points](https://retroachievements.org/game/14736). I am not sure what preferred behavior is in this scenario, but to me it seems related.

wescopeland:
This is now fixed for release date via https://github.com/RetroAchievements/RAWeb/pull/3127, but still needs to be checked/fixed for other columns.

wescopeland:
It appears this is fully resolved. The Last Updated column doesn't seem to include null values in the response.

PR #3127, opened 2025-01-25
fix: always prioritize games with release date when sorting by release date

bhimbho:


app/Platform/Concerns/BuildsGameListQueries.php
```diff
@@ -340,6 +340,7 @@ private function applyReleasedAtSorting(Builder $query, string $sortDirection =
                     ELSE 4
                 END AS granularity_order
             SQL)
+            ->orderByRaw('released_at IS NULL')
```
wescopeland, line 343:
```suggestion
            // Ensure NULL release dates always sort to the end, regardless of sort direction.
            ->orderByRaw('released_at IS NULL')
```
Recommend adding a comment explaining the "why", otherwise the intent of this sort may be lost.

wescopeland (approved):
Thanks!

app/Platform/Concerns/BuildsGameListQueries.php
bhimbho: suggestion makes sense. Thanks

Jamiras:
We briefly discussed this behavior in https://github.com/RetroAchievements/RAWeb/pull/2706#discussion_r1775335057.

> **Jamiras:**
> As a player, I'd most likely look for either the oldest game on the list, or the newest, and wouldn't care at all about the ones without a date. As such, maybe it's better the way you had it.

and

> **wescopeland:**
> To be clear, I'm only seeing null dates appear at the bottom of the list if the sort is by earliest (ascending).
> ...
> However, based on the findings reported in my previous comment, this is a behavior I'm in favor of.

Both #3116 and #2957 seem to want nulls to always be at the end, which this seems to address. I just want to make sure everyone is in favor of this change.

wescopeland:
I've grown to think the change in this PR is most likely desirable for two reasons:
* It's one we keep hearing about on Discord and elsewhere.
* From my own day-to-day using the site as a player, seeing null dates first in the sort order is unexpected.

On the other hand, as an achievement developer, I _do_ want to see the null dates first when I do this particular sort.

For the achievement developer persona, we should probably handle that use case in Filament.

bhimbho:
> I've grown to think the change in this PR is most likely desirable for two reasons:
> 
> * It's one we keep hearing about on Discord and elsewhere.
> * From my own day-to-day using the site as a player, seeing null dates first in the sort order is unexpected.
> 
> On the other hand, as an achievement developer, I _do_ want to see the null dates first when I do this particular sort.
> 
> For the achievement developer persona, we should probably handle that use case in Filament.

wescopeland:
Hey @bhimbho 👋 

We'll get around to merging this before we do our next release. Thanks!

Jamiras:
> On the other hand, as an achievement developer, I do want to see the null dates first when I do this particular sort.
>
> For the achievement developer persona, we should probably handle that use case in Filament.

I don't think that's necessary. It might even be confusing for someone who is both a player and a developer. There's a page selector. If you're actively seeking the null values, you can just jump to the last page.

Jamiras (approved):

PR #3128, opened 2025-01-25
detect alternate core label

Jamiras:
Apparently, IOS labels the cores "core.libretro" instead of "core_libretro"

https://discord.com/channels/476211979464343552/1002689485005406249/1332717032055115897


wescopeland (approved):
I haven't tested the change locally, but it seems reasonable and the tests prove that it works.

PR #3129, opened 2025-01-25
fix: resolve 404 with deleted user sets and tickets

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1002689485005406249/1332744554721312839.

The solution required a `resolveSoftDeletableRouteBinding()` method to be added to the `User` model to play nicely with Folio's `withTrashed()` middleware.

I'm not certain if the `->orWhere('User', $value)` is required to be in there. I verified we do still seem to need it for `resolveRouteBinding()`.


Jamiras (approved):
Should the username still be clickable in the breadcrumb on these pages?

wescopeland:
Updated in latest. Now, if the user is deleted, their name will have a strikethrough and not be hyperlinked in the breadcrumbs.

Jamiras (approved):

PR #3131, opened 2025-01-25
fix(SharedHubStrategy): exclude Meta and Misc hubs

wescopeland:
This PR causes the game suggestions pages to exclude Meta and Misc hubs from their shared hub strategy. This should hopefully make the hub selection ever-so-slightly more sensical (ie: no more recommendations because the game happens to be in the Noncompliant Writing hub).


app/Platform/Services/GameSuggestions/Strategies/SharedHubStrategy.php
```diff
@@ -36,6 +36,8 @@ public function select(): ?Game
             ->whereHas('games', function ($query) {
                 $query->where('game_id', '!=', $this->sourceGame->id); // needs other games too!
             })
+            ->where('Title', 'NOT LIKE', "%Meta%")
```
Jamiras, line 39: This should be `NOT LIKE "[Meta%"` or you'll exclude hubs like "[Series - Metal Gear]" and "[Series - Twisted Metal]".

app/Platform/Services/GameSuggestions/Strategies/SharedHubStrategy.php
```diff
@@ -36,6 +36,8 @@ public function select(): ?Game
...
+            ->where('Title', 'NOT LIKE', "%Misc%")
```
Jamiras, line 40: This looks safe for now, but I'd recommend changing it in a similar manner.

wescopeland: Updated to "%Misc. - %" in latest.

app/Platform/Services/GameSuggestions/Strategies/SharedHubStrategy.php
Jamiras: The `%Misc. -%` filter will work, but `%Meta -%` will miss anything `[Meta|QA -`, `[Meta|Art - `, or `[Meta|DevComp - `

wescopeland: Good catch, I forgot about those.

I have reservations about adding more "NOT LIKE" filters on the query. I've combined everything into a single:
```php
->whereRaw("game_sets.title NOT REGEXP '^(Meta[| -]|Misc\\.[ -])'")
```
which should be a bit more performant. Note that the square brackets in the pattern are not checking for the presence of square brackets in the title. They're defining character classes for conditionals in the pattern.

SQLite doesn't support `REGEXP`, so I needed to add some conditional logic for tests.

Jamiras: Unless I made a type, this is still matching the aforementioned series names.
```
select title from game_sets where title like '%meta%'
and title not regexp '^(Meta[| -]|Misc\\.[ -])';
```
I think it's the leading `^`. Doesn't that imply that the `Meta` must start the string? And we know the strings all start with an open bracket.

wescopeland: No typo on your end, I messed up. I think I have things in a better place now in latest.

Jamiras (approved):

PR #3132, opened 2025-01-25
perf(CommonPlayersStrategy): add index to reduce suggestions bottleneck

wescopeland:
Adds an index to `player_games` which dramatically improves the performance of `CommonPlayersStrategy`.

**Before**
Screenshot 2025-01-25 at 5 01 34 PM

**After**
Screenshot 2025-01-25 at 5 01 03 PM


database/migrations/2025_01_25_000000_update_player_games_table.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        Schema::table('player_games', function (Blueprint $table) {
+            $table->index([
+                'user_id',
+                'achievements_unlocked',
+                'achievements_total',
+                'deleted_at',
+                'game_id',
+            ], 'idx_player_games_suggestions'); // custom name needed because the auto-generated one is too long
+        });
```
Jamiras, line 20: This looks like the `player_games_completion_sample_idx`, but with the additional `deleted_at` column. Do we need both?

Jamiras: There's currently **0** deleted player_games records.

Maybe you could remove the `deleted_at` check form the query and leverage the existing index? A little deleted data isn't going to matter for the outcome of random selection.

wescopeland: I believe it's unfortunately still needed:

Screenshot 2025-01-26 at 10 48 44 AM

The screenshot above is with the index removing that column. Eloquent deciding to automagically include the column is what necessitates it being part of the compound index.

Jamiras: `->withTrashed()`?

wescopeland: That works - fixed in latest.

Jamiras: Is the new index still needed?

wescopeland: Yes, I think so. As soon as I delete it, I see these monster queries appear:
Screenshot 2025-01-26 at 10 59 31 AM

Jamiras: If we're not leveraging the old index, there's no reason to remove the `deleted_at` clause.

wescopeland: Sorry, I'm happy to implement feedback, but I'm getting confused. What specifically would you like me to do?

My interpretation is you'd like me to add back `deleted_at` to the index as well as the queries, but we don't have any deleted records.

Jamiras: I was hoping you'd be able to use the existing `player_games_completion_sample_idx` index, but it sounds like that's not being picked up. 
```
 KEY `player_games_completion_sample_idx` (`game_id`,`achievements_unlocked`,`achievements_total`,`user_id`,`id`),
```
I see now that that index has `id` in it - why would you ever want id in an index? Maybe we can drop it and replace it with yours.

If you're going to have to add a new index, do whatever you feel is necessary with it - like filtering out the deleted items.

wescopeland: `player_games_completion_sample_idx` is a [covering index](https://stackoverflow.com/questions/62137/what-is-a-covered-index), but `id` is probably redundant.

Based on what I'm seeing in my local, we don't even need that index anymore and a lot of the code in the strategy can be simplified.

I've updated the new migration and the strategy in latest.

Jamiras (approved):

PR #3141, opened 2025-01-26
feat(api): allow querying for user profile by ULID

wescopeland:
https://discord.com/channels/310192285306454017/1017568867574362283/1333195457597411460

API consumers need a stable value to query by, as a user's name is no longer stable and we probably don't want them querying by user ID. Therefore, we add a new column, `ulid`, to the user accounts table. `API_GetUserProfile` has been updated to return a `ULID` field as well as be queryable by `?i={ulid}`.

Documentation for this API change will come after we've agreed this is the right direction & before merging.

This PR includes a migration and a mandatory sync command:
```
sail artisan ra:sync:user-ulids
```

If we're happy with this direction, I'll open an api-docs PR, as well as update the other endpoints.

**Example**
```
GET /API/API_GetUserProfile.php?y=redacted&i=00003EMFWR7XB8SDPEHB3K56ZQ
```
```json
{
  "User": "wesweswes",
  "ULID": "00003EMFWR7XB8SDPEHB3K56ZQ",
  "UserPic": "\/UserPic\/WCopeland.png",
  "MemberSince": "2020-02-02 20:10:35",
  "RichPresenceMsg": "Playing Sonic the Hedgehog [Subset - Perfect Bonus]",
  "LastGameID": 29895,
  "ContribCount": 31962,
  "ContribYield": 255792,
  "TotalPoints": 27503,
  "TotalSoftcorePoints": 0,
  "TotalTruePoints": 164850,
  "Permissions": 4,
  "Untracked": 0,
  "ID": 117089,
  "UserWallActive": true,
  "Motto": "https:\/\/i.imgur.com\/ov30jeD.jpg"
}
```


app/Community/Commands/SyncUserUlids.php
```diff
@@ -0,0 +1,71 @@
...
+use Illuminate\Console\Command;
+use Illuminate\Support\Str;
+
+class SyncUserUlids extends Command
+{
+    protected $signature = 'ra:sync:user-ulids';
+
+    protected $description = 'Sync user ulid field values';
+
+    public function __construct()
```
gboquizosanchez, line 19: I think this is not necessary.

wescopeland: Thanks for the feedback!

It's a one-time command. Happy to resolve any bugs found in the logic, but as long as this command works, I'd prefer not to spend any time hunting down stylistic issues.

gboquizosanchez: Well, as you wish, but it's not a stylistic issue, it's a non necessary override from an inheritance

luchaos:
Heads up: users maintain a Hash ID attribute for permalinks that probably serve the same purpose but are not exposed in the API yet. They were meant to replace references for user IDs and usernames in rich/shortcode text without exposing IDs (which would allow for enumeration) nor relying on "static" usernames.

You can find them in Filament being genereated on the fly, are based on the actual user ID, calculated using Knuth's integer hash via the "Optimus" (Prime) package: https://github.com/jenssegers/optimus

As an example, this is my user's permalink: https://retroachievements.org/u/1354722784957298787

The reason I bring this up here is that if we choose to go with stored ULIDs there is no need for Hash IDs and vice versa. I don't hold a strong opinion on which one to move forward with.

wescopeland:
Hey @luchaos! 👋 

This does indeed add some nuance to the discussion.

I agree that both approaches are sound. It isn't a strong opinion, but I lean towards ULIDs only because Knuth prime hashes are reversible. Right now, it appears `HasHashId` is hashing against the primary key, which ultimately means someone clever could take the hash and calculate what your real user ID is.

We could change the hashing field to be `User`, but this could have the same problem if it's undesirable to not leak an original username.

wescopeland:
Actually, it looks like Optimus includes a random integer in the calculation. Given that, maybe it's better to not use ULIDs and use those hash IDs instead.

app/Community/Commands/SyncUserUlids.php
```diff
@@ -0,0 +1,66 @@
...
+
+        User::withTrashed()
+            ->whereNull('ulid')
+            ->chunkById(4000, function ($users) use ($progressBar, &$lastTimestamp) {
+                $updates = [];
+
+                /** @var User $user */
+                foreach ($users as $user) {
+                    // Increment timestamp for each ULID to ensure some variance.
+                    $lastTimestamp++;
```
Jamiras, line 40: Suggest using `$user->Created` for even more variance (and for more consistency with any users created after this point).

Note that this does decrease the variance for users who do share the same timestamp, but that can be adjusted by adding a random number of milliseconds:
```
      $milliseconds += rand(1,20);
      $milliseconds %= 1000;

      $timestamp = $user->Created->clone()->addMilliseconds($milliseconds);
      $ulid = (string) Str::ulid($timestamp);
```

luchaos:
I'm very fine with going with ULIDs and support that decision.

Another downside to hash IDs is that if the configured prime number are lost all of the permalinks would change. Which is certainly not what a permalink should ever do 😅

In this case, I think removing the Optimus package and transitioning permalinks to utilize the new ULIDs would be appropriate.

Thanks for considering my input and maintaining the focus on the optimal solution, as usual 😊

wescopeland:
> Another downside to hash IDs is that if the configured prime numbers are lost for whatever reason all of the permalinks would change. Which is certainly not what a permalink should ever do 😅

I hadn't considered that either. ULIDs it is, then!

Jamiras:
> Another downside to hash IDs is that if the configured prime numbers are lost for whatever reason all of the permalinks would change. Which is certainly not what a permalink should ever do 😅

This is only true if the hashes aren't stored in the DB (which they currently aren't).

We'd have the same problem with ULIDs. If we generate a ULID from the Created timestamp, there's still a possibility of multiple users joining at the same second. ULIDs are unique to the millisecond, and claim to handle clash resolution (though I'm not certain that's true for arbitrarily requesting a ULID for some timestamp in the past).

Even if we faked the millisecond value, there's no way to derive a millisecond from the user ID to ensure 100% uniqueness.

luchaos:
> Actually, it looks like Optimus includes a random integer in the calculation. Given that, maybe it's better to not use ULIDs and use those hash IDs instead.

Whichever you prefer. I thought about storing hash IDs to never run into the loss problem but it's unlikely to happen.

The hashed IDs _should_ be resilient. As you said, someone woth enough compute power could theoretically fogure it out but wouldn't gain much. It would be securitx by obscurity against enumeration attacks - which are in turn unlikely to gain much insight.

luchaos:
> > Another downside to hash IDs is that if the configured prime numbers are lost for whatever reason all of the permalinks would change. Which is certainly not what a permalink should ever do 😅
> 
> 
> 
> I hadn't considered that either. ULIDs it is, then!

Yeah, the more I think about it the more I lean towards ULIDs myself. Much easier to reason about and well supported in Laravel. Less esoteric, too. I just liked the idea of hashed IDs and played around with it to be honest.

luchaos:
> > Another downside to hash IDs is that if the configured prime numbers are lost for whatever reason all of the permalinks would change. Which is certainly not what a permalink should ever do 😅
> 
> 
> 
> This is only true if the hashes aren't stored in the DB (which they currently aren't).
> 
> 
> 
> We'd have the same problem with ULIDs. If we generate a ULID from the Created timestamp, there's still a possibility of multiple users joining at the same second. ULIDs are unique to the millisecond, and claim to handle clash resolution (though I'm not certain that's true for arbitrarily requesting a ULID for some timestamp in the past).
> 
> 
> 
> Even if we faked the millisecond value, there's no way to derive a millisecond from the user ID to ensure 100% uniqueness.

Very true. I guess the next step could be snowflakes - I had an implementation for that, too 😅 Doesn't make much sense without sharding though.

Interesting problem - we do have quite a few users sharing the same creation date and time iirc. For future signups some uniqueness check during creation with a simple loop might be enough.

Jamiras:
> We need to handle two collision scenarios:
> * Historical data (the sync command)
> * New user signups / registration flow

I don't think you have to worry about either. https://github.com/ulid/spec
> Monotonic sort order (correctly detects and handles the same millisecond)

In testing the command, I naively just used Created, and you could see the dozen or so Scott alts created on 11/21/12 had ULIDs that only differed by one character. Adding the random milliseconds made the "random" part of the ULID truly unique between records.

I'm assuming the ULID implementation just has an internal counter that gets incremented for each call, so it will handle collisions when passed the same timestamp. And I believe that would prevent us from being able to consistently generate the same ULID from a given timestamp.

I think storing the values in the database solves this problem. And if two users do sign up at the exact same millisecond, they may get neighboring ULIDs, but not conflicting ones.

luchaos:
> I don't think you have to worry about either. https://github.com/ulid/spec
> 
> > Monotonic sort order (correctly detects and handles the same millisecond)

nice! learned something today, thank you!

app/Community/Commands/SyncUserUlids.php
wescopeland: Updated in latest to use this improved technique for generating variance.

app/Community/Commands/SyncUserUlids.php
```diff
@@ -18,28 +17,22 @@ class SyncUserUlids extends Command
...
     public function handle(): void
     {
-        $total = User::whereNull('ulid')->count();
-
-        if ($total === 0) {
-            $this->info('No records need ULIDs.');
-
-            return;
-        }
+        $total = User::count();
```
Jamiras, line 20: I realize this is meant to be a one-time command, but without the "whereNull" check, it must only be run once. The random millisecond salt will cause (statistically) every user to get a new ULID.

app/Community/Commands/SyncUserUlids.php
```diff
@@ -18,28 +17,22 @@ class SyncUserUlids extends Command
...
-            ->whereNull('ulid')
-            ->chunkById(4000, function ($users) use ($progressBar, &$lastTimestamp) {
+            ->chunkById(4000, function ($users) use ($progressBar) {
                 $updates = [];
 
                 /** @var User $user */
                 foreach ($users as $user) {
-                    // Increment timestamp for each ULID to ensure some variance.
-                    $lastTimestamp++;
+                    $milliseconds = rand(1, 20);
```
Jamiras, line 31: This should be `+=`. We still want to simulate increasing timestamps for entries created in the same second.

Without the `+=`, the `%=` on the next line is unnecessary.

app/Community/Commands/SyncUserUlids.php
wescopeland: Re-added these checks in latest.

app/Community/Commands/SyncUserUlids.php
wescopeland: Fixed to `+=` in latest.

Jamiras (approved):

PR #3143, opened 2025-01-27
feat: set up notifications for discord #name-change-log channel

wescopeland:
Adds support for logging to `#name-change-log`: https://discord.com/channels/310192285306454017/1333550040509059102.

Tested & verified at https://discord.com/channels/476211979464343552/894379452631490610/1333548825318654013.

On a username change approval, a message will be logged like:
[Scott](https://retroachievements.org/user/Scott): Display name changed from **Foo** to **Bar**.

`#name-change-log` is a privileged channel, with access granted to all devs and a handful of teams.


wescopeland:
I've removed both Developer and Engineer. I think Art Helper should probably stay, as they may need to hunt specific users down to attribute art credit at some point.

PR #3082, opened 2025-01-19
show award tiers on event page

Jamiras:
Adds a widget to the side panel that shows available tiered badges, their requirements, and the number of people who have earned them.

image

image

PR #3145, opened 2025-01-28
properly classify converted dev events

Jamiras:
Fixes an issue that hasn't come up yet where events in the Dev Events hubs weren't being put in the Site Awards bucket.


app/Helpers/render/site-award.php
```diff
@@ -8,39 +8,72 @@
 
 function SeparateAwards(array $userAwards): array
 {
-    // Get all dev event game IDs in a single optimized query.
-    $devEventGameIds = GameSet::query()
-        ->whereHas('games', fn ($q) => $q->whereIn('game_id', array_column($userAwards, 'AwardData')))
```
Jamiras, line 13: Prefiltering the list reduces the number of records in the `whereIn` clause, which provides a sizable performance increase.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -1113,7 +1113,7 @@ function resize() {
             </div>
         @endif
 
-        @if ($numAchievements > 0 && $isOfficial)
+        @if ($numAchievements > 1 && $isOfficial)
```
wescopeland, line 1116: I've noticed this bug as well.

I'm actually wondering if 1 is still too low. Might be worth bumping to `>= 6`?

wescopeland (approved):
Verified locally.

resources/views/pages-legacy/gameInfo.blade.php
Jamiras: I noticed this when collapsing some of the existing events. Events don't require 6 achievements for a badge, so some of the existing events will be collapsed to having a single achievement after the conversion. A bar chart with a single bar isn't very informative. Other events will be collapsed to one achievement per tier, and a two-tier event would have two bars. Arguably, the tier distribution is handled by #3082, so the limit could be higher.

PR #3151, opened 2025-01-30
fix(game): remediate error on set request after username change

wescopeland:
This PR remediates an error where if a user's display name differs from their username, an error message appears when clicking the "Request Set" button on the game page.


Jamiras (approved):

Jamiras:
Just out of curiosity, why didn't the `whereName` work here? Or is it just that `->User` isn't populated?

wescopeland:
The _set-requests.php_ endpoint was updated, but it's receiving the wrong field value from the UI in the HTTP call.

PR #3152, opened 2025-01-30
Allow weighting event achievements

Jamiras:
Allows assigning a point value to an event achievement (1-5 [1=default]). Changes the wording for tiered achievements from "achievements required" to "points required".

SQL to run after upgrade:
```
UPDATE Achievements SET Points=1 WHERE ID IN (SELECT achievement_id FROM event_achievements);
```

Then the following command should be run to regenerate the aggregate data for the event games:
```
sail artisan ra:platform:game:update-metrics 28429,29830,31745,32718,32827,32828,32829,32830,32831,32832,32833,32834,32835,32836,32838,32847
```


wescopeland (approved):
Note that this component still does display RetroPoints:
Screenshot 2025-01-30 at 5 31 56 PM

I'm in favor of not touching this though. We can fix it in the React migration.

Jamiras:
I've updated it to not display the true points if they're 0 (I did the same thing for the achievement list item).

I've also updated the order of the event-specific processing so it occurs after the point totals are tallied so we can explicitly zero out the total true points.

With the React conversion we should hide the achievement progress if it matches the point total (or maybe always?).

televandalist:
Nice. I was wondering if there would be something like this. I'm planning something where it would definitely get utilized. 😏

wescopeland (approved):

PR #3156, opened 2025-01-30
feat(api): Add sortOrder to API_GetComments

ioslife:
Alter the GetComments endpoint from #2552 to add a sortOrder query param

```
/*
*  API_GetComments - returns the comments associated to a game or achievement
*    i : game or achievement id or username
*    t : 1 = game, 2 = achievement, 3 = user
*    s : sortOrder - sort comments. 'submitted' = ascending, '-submitted' = descending (default: 'submitted')
*
*  int         Count                       number of comment records returned in the response
*  int         Total                       number of comment records the game/achievement/user actually has overall
*  array       Results
*   object      [value]
*    int         User                      username of the commenter
*    string      Submitted                 date time the comment was submitted
*    string      CommentText               text of the comment
*/
```


ioslife:
FYI – I am open to a different query letter here

public/API/API_GetComments.php
```diff
@@ -15,6 +15,7 @@
 *    t : 1 = game, 2 = achievement, 3 = user
 *    o : offset - number of entries to skip (default: 0)
 *    c : count - number of entries to return (default: 100, max: 500)
+*    s : sortOrder - sort comments. 0 = ascending, 1 = descending (default: 0)
```
wescopeland, line 18: This could be too ambiguous in terms of what the `orderBy()` is. I'm also thinking integers as possible values may be inflexible if this pattern is extended to other API endpoints.

I think we may want to change this to use JSON:API-like enum values, ie: `createdAt` and `-createdAt`.

Jamiras: We do have a few public APIs where filter parameters are effectively booleans (0=all,1=filtered), and some that are direct mapping of enums (3=core,5=unofficial).

But as this is not an enum, and it's not being implemented as a boolean (`s: sort descending (0=no,1=yes)`), I agree that it makes sense to have it be more free-form.

public/API/API_GetComments.php
```diff
@@ -84,8 +87,13 @@
         $query->whereNull('banned_at');
     })
     ->offset($offset)
-    ->limit($count)
-    ->get();
+    ->limit($count);
+
+if ($sortOrder == 1) {
+    $commentsQuery->orderBy('Submitted', 'DESC');
+}
```
Jamiras, line 94: You'll probably want an explicit `orderBy` for ascending to. It's unsafe to assume that an unordered query returns sorted data.

public/API/API_GetComments.php
ioslife: Cool, I will check on this next week. Thanks for looking at it

ioslife: Hello, it is I! Finally making finishing this a priority, apologies for the PR and run. 

I just pushed a commit that accepts `submitted` and `-submitted` as the s param. I am willing to add others to the sortOptions enum, but wanted to verify this is the pattern you were thinking @wescopeland.

wescopeland: Yep, I think so. I would rename it to `sort` to align with some of our other paths like: https://retroachievements.org/games?sort=-achievementsPublished

tests/Feature/Api/V1/CommentsTest.php
```diff
@@ -280,12 +351,14 @@ public function testGetCommentsForUserWithDisabledWall(): void
             'ArticleType' => 3,
             'user_id' => $user2->ID,
             'Payload' => 'This is my first comment.',
+            'Submitted' => "2024-01-18T15:01:04.000000Z",
```
Jamiras, line 354: There's another test between these two `testGetCommentsForUserByUlid` that is failing for me. It looks like it needs static submitted timestamps too.

ioslife: I'll check it out and add a couple of other sort options

ioslife: Added static timestamps to this test, thanks for catching. Looks like I had just gotten lucky previously.

I also looked at the `Comment.php` model and I don't think there is a good option for expanding sort capabilities. I could see _maybe_ sort by userId, but that doesn't seem like a great use case to me. Open to feedback from you and @wescopeland on that decision.

wescopeland (approved):
LGTM - we just need an accompanying docs PR before we can ship this in a release.

ioslife:
> LGTM - we just need an accompanying docs PR before we can ship this in a release.

Awesome, thanks! Will push docs PR tomorrow.

ioslife:
@wescopeland added in https://github.com/RetroAchievements/api-docs/pull/79

PR #3159, opened 2025-01-30
ci: run tasks conditionally

wescopeland:
This PR makes an adjustment to our pull request CI workflow.

The meat of the PR is here:
```yml
changes:
  runs-on: ubuntu-22.04
  outputs:
    php: ${{ steps.filter.outputs.php }}
    node: ${{ steps.filter.outputs.node }}
  steps:
    - name: Check changed files
      uses: dorny/paths-filter@v2
      id: filter
      with:
        filters: |
          php:
            - '**/*.php'
            - 'composer.json'
            - 'composer.lock'
            - 'phpstan.neon'
            - 'pint.json'
          node:
            - '**/*.js'
            - '**/*.ts'
            - '**/*.tsx'
            - 'package.json'
            - 'pnpm-lock.yaml'
            - '*eslint*'
            - 'tailwind.config.json'
```

If something in the back-end layer is touched, PHP tasks will run.
If something in the front-end layer is touched, Node.js tasks will run.
Otherwise, tasks are marked as skipped (verifiable by clicking on the task icons in this PR):

Screenshot 2025-01-30 at 6 12 33 PM

The output of `composer types` is the only glue between Laravel and the React UI, and changes there will always be caught by this filter (requiring all Node.js tasks to pass).

We should probably make a similar change to our pre-push job, but given the logic there will likely be very different to achieve the same functionality, let's push that to a subsequent PR.


wescopeland:
I've verified this is functional and now ready for review.

Jamiras:
Is this going to block merging?
image

wescopeland:
I don't think so. Based on my research, GitHub Actions is smart enough to detect that a "depends on" task has been skipped. Things may look different after this PR is merged to `master`, and if it's still complaining I can investigate further.

Jamiras (approved):

wescopeland:
I've made a follow-up tweak in https://github.com/RetroAchievements/RAWeb/pull/3125/commits/cc82868e3d5c5cec82679a807c0fc505b4fa7878 to ensure we can keep our 6 jobs as required.

Rather than skipping the required jobs, they now execute and instantly succeed if they're not needed for the branch.

wescopeland:
Verified the tweak as working in https://github.com/RetroAchievements/RAWeb/actions/runs/13081868701 and https://github.com/RetroAchievements/RAWeb/actions/runs/13081790730.

PR #3163, opened 2025-01-31
ci: use phpstan result cache

wescopeland:
Another PR to improve CI performance. This specifically targets `analyse` (PHPStan) while running in GitHub Actions.

As the project grows and grows, the `analyse` step is going to take exponentially longer because, currently, there is no cache to leverage in the actions containers.

This branch incorporates the official instructions for setting up the PHPStan cache in GitHub Actions from https://phpstan.org/user-guide/result-cache#setup-in-github-actions.


wescopeland:
https://github.com/RetroAchievements/RAWeb/actions/runs/13082072305/job/36507524947?pr=3163 confirms a successful cache hit. The cache will be available across all branches.

Jamiras (approved):

PR #3166, opened 2025-02-01
only allow event managers to view future events

Jamiras:
If an event has a start date in the future, anyone without access to the event management console will get an Unauthorized message trying to view it.

https://discord.com/channels/310192285306454017/1334630307915104316/1335134440359329872

### DIFF
--- a/app/Policies/EventPolicy.php
+++ b/app/Policies/EventPolicy.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace App\Policies;
 
+use App\Models\Event;
 use App\Models\Role;
 use App\Models\User;
+use Carbon\Carbon;
 use Illuminate\Auth\Access\HandlesAuthorization;
 
 class EventPolicy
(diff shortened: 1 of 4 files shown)


resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -110,6 +110,13 @@
 $isFullyFeaturedGame = System::isGameSystem($gameData['ConsoleID']);
 $isEventGame = $gameData['ConsoleID'] == System::Events;
 
+// future events can only be viewed by users who can manage events.
+if ($isEventGame && $gameModel->event?->active_from != null
+    && $gameModel->event->active_from > Carbon::now()
+    && !$userModel?->can('manage', $gameModel->event)) {
+    abort(401);
+}
```
wescopeland, line 118: It would probably be best for the majority of this logic to live in `EventPolicy::view()`.

Jamiras: Good idea. Done.

app/Policies/EventPolicy.php
```diff
@@ -25,8 +27,15 @@ public function viewAny(?User $user): bool
         return true;
     }
 
-    public function view(?User $user): bool
+    public function view(?User $user, Event $event): bool
     {
+        if ($event->active_from != null && $event->active_from > Carbon::now()) {
```
wescopeland, line 32: nit: Recommend using `!==` to enforce type safety.

Here's a contrived example:
```php
var_dump("" == null); // true ?!
var_dump("" === null); // false
```

wescopeland (approved):

app/Policies/EventPolicy.php
Jamiras: done

PR #3167, opened 2025-02-01
feat(game): add game_sets attach/detach actions to game audit logs

wescopeland:
This PR makes it so that when a hub or similar game is attached/detached on a game, it's logged to the game's activitylog.

In a subsequent PR, I will make these log messages look prettier (this will require some Blade UI work, better to get the data collection piece out of the way as a separate unit of work).

Screenshot 2025-02-01 at 12 40 17 PM


app/Models/GameSet.php
```diff
@@ -78,6 +78,20 @@ public static function boot()
...
+
+                    activity()->causedBy($user)->performedOn($game)
+                        ->withProperty('old', [$attribute => null])
+                        ->withProperty('attributes', [$attribute => [
+                            'id' => $model->id,
+                            'title' => $model->title,
+                        ]])
+                        ->event('pivotAttached')
+                        ->log('pivotAttached');
+                }
```
Jamiras, line 94: It probably doesn't matter, but multiple add/removes in a single commit are creating separate activity records:
image

Jamiras (approved):

app/Models/GameSet.php
wescopeland: Probably not a huge deal. If this becomes cumbersome, we can revisit it later (or alternatively come up with some kind of squashing treatment when we display these in the UI).

PR #3173, opened 2025-02-03
fix(AchievementsRelationManager): gracefully handle bulk re-inserts of trashed credit

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1335738485273002056.

Currently, a server error is thrown when trying to re-insert a soft deleted credit record using this tool:
Screenshot 2025-02-03 at 4 48 05 PM

**Steps to repro the server error:**
1. Visit http://localhost:64000/manage/games/1.
2. Select every achievement on the list, bulk add some kind of credit for them.
3. Now, navigate to the achievement and delete that credit record.
4. Navigate back to http://localhost:64000/manage/games/1, and with the bulk tool attempt to reinsert the same type of credit for the same user.

Now, similar to `AuthorshipCreditsRelationManager`, if we're reinserting a trashed record, we'll restore it.


Jamiras:
> Now, navigate to the achievement and delete that credit record.

I can't figure out how to do that. I can edit the credit record, but there's no delete option.

wescopeland:
Interesting! Here is what I see:


https://github.com/user-attachments/assets/bdf4c312-7d05-4962-995f-88b5d7a91d8b


Could be perms-related? `AchievementAuthorPolicy::delete()`:
```php
public function delete(User $user, AchievementAuthor $achievementAuthor): bool
{
    return $user->hasAnyRole([
        Role::MODERATOR,
        Role::TEAM_ACCOUNT,
    ]);
}
```

Jamiras (approved):
Unrelated: Noticed that trying to add a user with a custom display_name by their display_name doesn't work. Entering their username shows the display_name in the list, and then that can be selected.

wescopeland:
Strange, I am able to reproduce that issue for either the resource or the bulk add tool:


https://github.com/user-attachments/assets/37ce8e89-f163-4b01-a94e-73c3c27c2aa3


Possible I am doing something wrong? Going to go ahead and merge this, but happy to hunt down a bug in a subsequent PR.

PR #3176, opened 2025-02-05
feat(messages): migrate message thread create page to React

wescopeland:
This PR finalizes the messaging system's migration to React. In this branch, both the PHP UI and React UI for the new create page are present.

**How to test this PR:**
```
sail artisan migrate

http://localhost:64000/messages/create2
http://localhost:64000/messages/create
```

You should also test various template scenarios, ie:
```
http://localhost:64000/messages/create2?message=I%27d%20like%20to%20report%20an%20unwelcome%20concept%20in%20%5Bach%3D24%5D.%0A%20%20%0A-%20Which%20Unwelcome%20Concept%3A%0A%28Insert%20which%20concept%20from%20the%20docs%20here%29%0A%20%20%0A-%20Detailed%20Explanation%3A%0A%28Provide%20as%20much%20detail%20as%20possible%20here.%20Assume%20the%20reader%20may%20not%20have%20played%20the%20game%20before.%20The%20more%20detail%20you%20provide%2C%20the%20better%20your%20case.%29&subject=Unwelcome%20Concept%3A%20Bouncing%20Combo%20Master%20%5B24%5D%20%28Sonic%20the%20Hedgehog%29&templateKind=unwelcome-concept&to=DevCompliance
```

One important technical note is the use of a fulltext index for user searches. Each search query takes approximately 10ms at a result, down from 3,000ms when using `LIKE`.

**Before**
Screenshot 2025-02-04 at 7 02 18 PM

**After**
Screenshot 2025-02-04 at 7 02 12 PM


app/Models/User.php
```diff
@@ -609,4 +607,52 @@ public function scopeTracked(Builder $query): Builder
...
+            ", [$keyword, $keyword, "{$keyword}%", "{$keyword}%"])
+            ->where(function ($q) use ($keyword) {
+                $q->where('User', '=', $keyword)
+                  ->orWhere('display_name', '=', $keyword)
+                  ->orWhere('User', 'LIKE', "{$keyword}%")
+                  ->orWhere('display_name', 'LIKE', "{$keyword}%");
+            })
+            ->orderBy('search_rank', 'desc')
+            ->orderBy('achievements_unlocked', 'desc')
+            ->limit(10);
```
Jamiras, line 656: The documentation for a FULLTEXT index says you have to use match/against. Is this smart enough to do that?

app/Models/User.php
```diff
@@ -609,4 +607,52 @@ public function scopeTracked(Builder $query): Builder
...
+            return $baseQuery;
+        }
+
+        // Common ranking calculation for both queries.
+        $activityRank = "
+            CASE
+                WHEN achievements_unlocked > 0 THEN 40
+                WHEN LastLogin > DATE_SUB(NOW(), INTERVAL 30 DAY) THEN 20
+                WHEN LastLogin > DATE_SUB(NOW(), INTERVAL 90 DAY) THEN 10
+                ELSE 0
```
Jamiras, line 634: Should LastLogin be given more weight? It seems like I'd be more likely to be searching for an active user than one with a lot of points.

resources/js/features/messages/components/CreateMessageThreadForm/useCreateMessageThreadForm.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      return axios.post<{ threadId: number }>(route('api.message.store'), normalizedPayload);
+    },
+  });
+
+  const onSubmit = async (formValues: FormValues) => {
+    await toastMessage.promise(mutation.mutateAsync(formValues), {
+      loading: t('Submitting...'),
+      success: ({ data }) => {
+        router.visit(route('message-thread.show', { messageThread: data.threadId }));
```
Jamiras, line 47: This seems to be redirecting to a popup window?
image

wescopeland: Ah, I think this may be automatically fixed by https://github.com/RetroAchievements/RAWeb/pull/3206. This is another case (similar to what happened with the Report Issue page) where an Inertia route is redirecting to a Blade route.

app/Models/User.php
wescopeland: Nope! Thanks for catching this, I messed up before I pushed and forgot to revert back to the correct query. Fixed in latest.

app/Models/User.php
wescopeland: I agree. I've made some adjustments in latest so equally-weighted results will now be ordered by LastLogin.

app/Models/User.php
```diff
@@ -609,4 +607,50 @@ public function scopeTracked(Builder $query): Builder
...
+            ->whereRaw('MATCH(User, display_name) AGAINST (? IN BOOLEAN MODE)', [$fullTextQuery])
+            ->orderByRaw("
+                CASE 
+                    WHEN User = ? THEN 100
+                    WHEN display_name = ? THEN 90
+                    WHEN User LIKE ? THEN 80
+                    WHEN display_name LIKE ? THEN 70
+                    ELSE MATCH(User, display_name) AGAINST (? IN BOOLEAN MODE) * 10
+                END DESC",
+                [$keyword, $keyword, $keyword . '%', $keyword . '%', $fullTextQuery]
```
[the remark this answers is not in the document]
Jamiras: It looks like everything is falling back to the LastLogin sort.

I think this is more-or-less the query that's being constructed, but I'm not seeing 
```
SELECT User, display_name, MATCH(User, display_name) AGAINST ('+telev*' IN BOOLEAN MODE) * 10 as MatchRating, 
       CASE WHEN User='telev' THEN 100 WHEN display_name='telev' THEN 90 WHEN User LIKE 'telev%' THEN 80 WHEN display_name LIKE 'telev%' THEN 70 ELSE MATCH(User, display_name) AGAINST ('+telev*' IN BOOLEAN MODE) * 10 END as SortRank
  FROM UserAccounts
 WHERE MATCH(User, display_name) AGAINST ('+telev*' IN BOOLEAN MODE)
 ORDER BY SortRank DESC, LastLogin DESC;
+------------------+------------------+--------------------+----------+
| User             | display_name     | MatchRating        | SortRank |
+------------------+------------------+--------------------+----------+
| Telev            | Telev            | 472.69412994384766 |      100 |
| televandalist    | tele             | 236.34706497192383 |       80 |
| Televisionia     | Televisionia     | 472.69412994384766 |       80 |
| televisionsputty | televisionsputty | 472.69412994384766 |       80 |
| televisionchild  | televisionchild  | 472.69412994384766 |       80 |
| Telev11          | Telev11          | 472.69412994384766 |       80 |
| televizoranet    | televizoranet    | 472.69412994384766 |       80 |
| Television7      | Television7      | 472.69412994384766 |       80 |
| televisedBow     | televisedBow     | 472.69412994384766 |       80 |
| televisionquest  | televisionquest  | 472.69412994384766 |       80 |
| Televasion       | Televasion       | 472.69412994384766 |       80 |
| televarolus      | televarolus      | 472.69412994384766 |       80 |
| Television       | Television       | 472.69412994384766 |       80 |
| televandalisttt  | televandalisttt  | 472.69412994384766 |       80 |
| televandalistt   | televandalistt   | 472.69412994384766 |       80 |
+------------------+------------------+--------------------+----------+
```

Also, I'm not sure exactly how the MATCH value is calculated, but the values generated by the query greatly outrank the exact and starts with constants. And I couldn't get a mid-word match to match. Try searching on "elevan" (televandalist) or "amiras" (Jamiras)

app/Models/User.php
```diff
@@ -609,4 +607,50 @@ public function scopeTracked(Builder $query): Builder
...
+        $fullTextQuery = '+' . $escapedKeyword . '*';
+
+        /**
+         * Search users with FULLTEXT index and rank results by:
+         * 1. Exact username match (100 points).
+         * 2. Exact display name match (90 points).
+         * 3. Username prefix match (80 points).
+         * 4. Display name prefix match (70 points).
+         * 5. Username infix match (weighted by FULLTEXT relevance).
+         * 6. Display name infix match (weighted by FULLTEXT relevance).
```
Jamiras, line 635: As User is not supposed to be visible for users with a custom display_name, display_name should be prioritized over User and "User like" should be eliminated in favor of just letting the match handle it. (display_name=100, User=90, like display_name%=80, match result)

resources/js/features/messages/components/CreateMessageThreadForm/useCreateMessageThreadForm.ts
wescopeland: Confirmed this should be fixed in latest as a result of #3206 being merged.

app/Models/User.php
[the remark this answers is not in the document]
wescopeland: In latest I decided to abandon the FULLTEXT index and remove the associated migration.

Searches now take roughly 450ms; if we want to push faster (we inevitably will when migrating the search feature to React), we should set up Laravel Scout so we don't have to worry about making any concessions in the flexibility of our searches.

As a result of this change, MATCH clauses have been removed. This is the query I now see when logging it out:
```sql
SELECT
  *
FROM
  `UserAccounts`
WHERE
  `email_verified_at` IS NOT NULL
  AND `Deleted` IS NULL
  AND `banned_at` IS NULL
  AND (
    `display_name` LIKE '%amiras%'
    OR `User` LIKE '%amiras%'
  )
  AND `UserAccounts`.`Deleted` IS NULL
ORDER BY
  CASE
    WHEN display_name = 'amiras' THEN 100
    WHEN display_name LIKE 'amiras%' THEN 90
    WHEN display_name LIKE '%amiras%' THEN 80
    WHEN User = 'amiras' THEN 70
    WHEN User LIKE '%amiras%' THEN 60
  END DESC,
  `LastLogin` DESC
LIMIT
  10
```

Mid-word matches should now work in latest.

app/Models/User.php
wescopeland: Tweaked this in latest - here's the new scoring:
```php
/**
 * Search users and rank results by:
 * 1. Exact display_name match (100 points).
 * 2. display_name prefix match (90 points).
 * 3. display_name contains match (80 points).
 * 4. Exact username match (70 points).
 * 5. username contains match (60 points).
 *
 * Finally, sort equally-ranked results by last login date.
 */
```
#5 (`User LIKE`) is there as a result of the removal of the FULLTEXT index.

Jamiras (approved):

PR #3181, opened 2025-02-06
sync achievement title/description/badge to active event achievements

Jamiras:
https://discord.com/channels/310192285306454017/1336000692371853404/1336000692371853404


app/Filament/Resources/AchievementResource/Pages/Edit.php
wescopeland, line 1: Functionally the code works great. My only concern is everything within the conditionals is duplicative. I'm wondering if it'd be best to extract it to either a method on the model or an action, eg:
```php
if ($record->isDirty(['BadgeName', 'Title', 'Description'])) {
    /**
     * reusable code
     * $record->syncMetadataWithEventAchievements()?
     * (new SyncMetadataWithEventAchievementsAction())->execute(...)?
     */
}
```
```php
$shouldResyncEventAchievementMetadata = (
    in_array('badge', $fields)
    || in_array('title', $fields)
    || in_array('description', $fields)
);
if ($shouldResyncEventAchievementMetadata) {
    /**
     * reusable code
     * $record->syncMetadataWithEventAchievements()?
     * (new SyncMetadataWithEventAchievementsAction())->execute(...)?
     */
}
```

Jamiras: I had considered that, but because one used an achievement record and the other used a fields collection, I didn't think it would be easy to consolidate them.

By rearranging when the action is called, I've managed to make both use the achievement record.

wescopeland (approved):

Issue #3185, opened 2025-02-06
Hubs appear in the results when searching for games


iTom33ro:
**Describe the bug**
Since the hubs got their new pages, they can still appear on the search page when having **Games** selected

**To Reproduce**
1. Go to the search page
2. Search for any terms that is in a hub
3. Notice that hubs can appear

**Expected behavior**
There should be a new category for hubs.

**Screenshots**

![Image](https://github.com/user-attachments/assets/b105c42e-5c54-4be0-855c-8df4284fb03f)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Brave
 - Version 1.74.51

Issue #3184, opened 2025-02-06
Search results still use the old links for hubs


SuperGamer18:
**Bug:**
Whenever I search for any hub and I hold my finger on that hub (or just hover my cursor over that result) it shows the old link that the hub used to occupy (aka it uses game instead of hub).

**Steps to reproduce the behavior:**
1. Search for any hub.
2. Hold your finger on the result (or hover your cursor over it).
3. See how the old link appears instead of the new one.
4. If you go to the link, it redirects you to the new hub link.

**Expected behavior:**
It should show the new link instead and without any redirects.

**Screenshots:**
![Image](https://github.com/user-attachments/assets/628f95c9-9d20-4e38-9885-fe0471ccae57)

**Information:**
 - Device: Xiaomi Redmi 12C
 - OS: Android 14
 - Browser: Firefox
 - Version: 134/5 I think

**Additional context**
As usual, this isn't platform/browser specific. While the screenshot and info come from a phone, I also found this issue on a laptop. I think this might have happened because the search results don't have the "HasSelfHealingURLs" function implemented (which is what was used to solve a previous issue of mine, #3075).

PR #3190, opened 2025-02-06
feat(search): add support for modern hubs

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3185 and https://github.com/RetroAchievements/RAWeb/issues/3184.

Excludes "hub games" from search results and uses results from `game_sets` instead:

Screenshot 2025-02-06 at 6 46 20 PM

Screenshot 2025-02-06 at 6 46 25 PM


public/request/search.php
```diff
@@ -25,7 +25,7 @@
 } elseif ($source == 'user' || $source == 'game-compare') {
     $order = [SearchType::User];
 } else {
-    $order = [SearchType::Game, SearchType::Achievement, SearchType::User];
+    $order = [SearchType::Game, SearchType::Hub, SearchType::Achievement, SearchType::User];
```
Jamiras, line 28: I was going to suggest adding Hubs to the Game case and maybe the Achievement case, but I can't figure out if they're even used any more.

I think the source prioritization has been lost to the blade conversion. If I recall correctly, it used to be that searching on a game page would only return games and searching on a user page would only return users, and this catch-all was for everything else.

Jamiras (approved):

Issue #1927, opened 2023-10-25
Global ranking: wrong rank for first row


Mindhral:
**Description**
On Global Ranking page, when using pagination, the first row shows the rank of the player on hardcore points ranking instead of the selected ranking.

**To Reproduce**
1. Go to 'Global Ranking' page
2. Stay on 'All Time' type
3. Select any ranking (sort) other than 'Hardcore Points' and 'Softcore Points'
4. From the code, I believe the option 'All Users' must be selected, but I don't have enough followed users to check
5. Go to next page(s), the first row shows the rank by points

**Expected behavior**
All rows ranked according to the same logic.

**Screenshots**
![image](https://github.com/RetroAchievements/RAWeb/assets/148589639/b87e57e1-4076-4e7b-81e6-b42bc74d946c)

**Additional context**
Seems to be a consequence of #1328 which handles tied ranks.
The function getUserRank, used to retrieve the rank for the first row, only handles Points (hardcore, softcore or "retro"), and the call in globalRanking.php (l.238 on current master) only use the first two anyway.

PR #3200, opened 2025-02-08
fix(globalRanking): resolve incorrect rank shown for 1st row on non-default sort

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/1927.

To reproduce the, navigate to http://localhost:64000/globalRanking.php, sort by "Hardcore Achievements", and navigate to page 2. Take a look at the "Rank" cell for the 1st row.

**Before**
Screenshot 2025-02-08 at 3 33 27 PM

**After**
Screenshot 2025-02-08 at 3 33 20 PM

**Root Cause**
When showing ranks on subsequent pages, we're using `getUserRank()` for calculating the first user's rank regardless of sort type.

`getUserRank()` only works for points-based rankings (hardcore/softcore/Retro points), as it counts users with higher points to determine rank. For other sorting types like achievement unlocks count, we should use the sequential rank from the ordered query results instead.


wescopeland:
> We could use a rank calculation for Hardcore achievements too

Agreed - I have a change ready for this. Going to merge this and then put it in a follow-up PR.

Issue #2743, opened 2024-10-07
Resetting a game should also update `player_stats`


wescopeland:
**Describe the bug**
Currently, if a game is a beaten (hardcore) and then reset, the player keeps the beaten game credit on `player_stats` (the global beaten games leaderboard).

**To Reproduce**
1. On a fresh account, beat a game in hardcore mode.
2. Observe that there is now 1 game for "Total beaten games" on the user profile. 
3. Reset the game.
4. Observe that the count is still 1.

**Expected behavior**
The count should be 0.

**Screenshots**
![Screenshot 2024-10-07 at 11 25 42 AM](https://github.com/user-attachments/assets/0498a73c-95d0-4ecd-afd5-b8b168e5a0b8)

**Additional context**
https://discord.com/channels/476211979464343552/1292851127095791647

PR #3202, opened 2025-02-10
fix(UpdatePlayerBeatenGamesStatsAction): handle resets to zero correctly

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/2743. Addresses https://retroachievements.org/viewtopic.php?t=29707&c=265715#265715.

Currently, when resetting achievements which also cause beaten awards to get deleted, it's possible to fall into a situation where `player_stats` is not correctly updated and hangs on to old data.

**How to reproduce:**
* Be sure Horizon is running locally.
* Create a new account.
* On admin.php, give the new account a hardcore unlock for achievement ID 3317.
* Observe on the new account's profile it now has a "Total Games Beaten" value of 1.
* While logged in as the new account, reset that 1 achievement on your settings page.
* Observe that "Total Games Beaten" on the account's profile is set to 1.

**Root Cause**
`UpdatePlayerBeatenGamesStatsAction` has safeguards to prevent writing rows where data is zero. However, these same safeguards are preventing transitions from 1 -> 0.

### DIFF
--- a/app/Platform/Actions/UpdatePlayerBeatenGamesStatsAction.php
+++ b/app/Platform/Actions/UpdatePlayerBeatenGamesStatsAction.php
@@ -11,6 +11,7 @@ use App\Models\User;
 use App\Platform\Enums\PlayerStatType;
 use App\Platform\Enums\UnlockMode;
 use App\Platform\Events\PlayerBeatenGamesStatsUpdated;
+use Illuminate\Support\Collection;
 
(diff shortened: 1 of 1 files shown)


Jamiras (approved):
There is a Test file for this action. How hard would it be to add a case for this scenario?

wescopeland:
Not very hard. Added a new test case in latest.

PR #3207, opened 2025-02-11
feat: have RABot automatically change Discord nicknames on name change

wescopeland:
This PR adds a new action which is invoked during `ApproveNewDisplayNameAction`.

Now, when a user's new display name is approved, a call is made to Discord's API to instruct RABot to:
* Attempt to find a user in the main Discord server with a nickname matching the old display name.
* If found, change the nickname to the new display name.

This should hopefully remove a lot of the manual work currently involved in the username change approval process.

I've verified this action works in the main server using a hardcoded guild (server) ID and bot token. If you'd like to try yourself, I've populated the new _.env_ values already in prod's environment file.


Jamiras (approved):
Untested, but I remember you doing some testing of this. Changes look reasonable.

PR #3211, opened 2025-02-12
fix: bump username approval cooldown from 30 to 90 days

wescopeland:
Based on feedback from several teams. The reject/expiry cooldown remains at 30 days.


lang/en_US.json
```diff
@@ -634,7 +634,7 @@
     "Mastered by same players": "Mastered by same players",
     "<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>": "<1>{{achievementTitle}}</1> from <2>{{gameTitle}}</2>",
     "New usernames must match.": "New usernames must match.",
-    "You can only request a new username once every 30 days, even if your new username is not approved. Are you sure you want to do this?": "You can only request a new username once every 30 days, even if your new username is not approved. Are you sure you want to do this?",
+    "Username change requests have cooldown periods: 90 days after an approval, or 30 days after a denial. Are you sure you want to request a change?": "Username change requests have cooldown periods: 90 days after an approval, or 30 days after a denial. Are you sure you want to request a change?",
```
Jamiras, line 637: The implementation seems to require 90 days for both approvals and denials.

wescopeland: Fixed in latest with an update to the policy.

lang/en_US.json
```diff
@@ -660,7 +660,7 @@
     "New username must be different from current username.": "New username must be different from current username.",
     "Your username change request is being reviewed.": "Your username change request is being reviewed.",
     "Your username cannot be changed right now.": "Your username cannot be changed right now.",
-    "You can request another change after your previous request's 30-day cooldown period has ended.": "You can request another change after your previous request's 30-day cooldown period has ended.",
+    "You can request another change after your previous request's cooldown period has ended.": "You can request another change after your previous request's cooldown period has ended.",
```
Jamiras, line 663: I wonder if it would be worth including the cooldown times here. After a request has been submitted, the user has no idea how much longer they have to wait.

Jamiras (approved):

wescopeland:
Holding off on this change for the time being. I might revive this branch later.

PR #3218, opened 2025-02-14
feat: add a dedicated Hub tooltip hover card

wescopeland:
**Post-deployment command**
```
sail artisan ra:community:migrate-game-shortcodes
```
---

Context: I needed to disable the create hub button in the management app in https://github.com/RetroAchievements/RAWeb/pull/3215, as creating hubs without a "backing game" was flooding the logs with server errors.

The current game card implementation assumes all hubs have backing games. This obviously won't be the case for new hubs going forward.

This PR creates a hub card, which only relies on data from the game set model. It also re-enables the creation of new hubs.

412181151-12122688-3a8e-4f09-acee-bd5016efe2c1


app/Helpers/render/game.php
```diff
@@ -50,10 +50,10 @@ function gameAvatar(
     }
 
     return avatar(
-        resource: 'game',
+        resource: $isHub ? 'hub' : 'game',
         id: $id,
         label: $label !== false && ($label || !$icon) ? $label : null,
-        link: $href ?? route('game.show', $id),
+        link: $isHub ? route('hub.show', ['gameSet' => $id]) : route('game.show', $id),
```
Jamiras, line 56: When `$isHub`, it should change the tooltip too.

[game=]
image

[hub=]
image

wescopeland: This may be a bigger issue. I think what we really want to do is convert [game=31277] to [hub=28019] on submit.

I've modified this to happen in latest (we'll probably need to run our game -> hub shortcode convert command again on deploy).

Preview is still affected by this issue, but not sure if that's worth hunting down.

Jamiras (approved):

PR #3228, opened 2025-02-16
fix(ticket): soft delete tickets when achievement is soft deleted

wescopeland:
**Pre-deploy SQL**
```sql
UPDATE Ticket t
INNER JOIN Achievements a ON t.AchievementID = a.ID 
SET t.deleted_at = a.deleted_at
WHERE a.deleted_at IS NOT NULL 
AND t.deleted_at IS NULL;
```
---

Resolves https://retroachievements.org/log-viewer?file=bd7e48b7-laravel-2025-02-16.log&query=log-index%3A116.

> achievementAvatar(): Argument #1 ($achievement) must be of type App\Models\Achievement|array|string|int, null given, called in /home/forge/retroachievements.org/storage/framework/views/4dea0b2213cf55ac278d69d52717c821.php on line 119 (View: /home/forge/retroachievements.org/releases/2025-02-16T152149-2025.02.06-B-eb3f967f/resources/views/components/ticket/ticket-list.blade.php) (View: /home/forge/retroachievements.org/releases/2025-02-16T152149-2025.02.06-B-eb3f967f/resources/views/components/ticket/ticket-list.blade.php)

If a soft-deleted achievement has any tickets associated with it, when someone tries to view a ticket list containing that ticket, the page will crash.

With these changes, on delete, the tickets are also soft deleted. They're restored if the achievement is ever restored.


Jamiras (approved):
It took me longer than I'd like to admit to figure out how to test this, as you can't delete achievements that have unlocks. And unless an achievement is super-broken, it will have some number of unlocks before it gets ticketed.

wescopeland:
Added a pre-deploy SQL command to ensure data integrity.

PR #3229, opened 2025-02-19
fix(connect): remediate race condition on submitlbentry causing server error

wescopeland:
Resolves https://retroachievements.org/log-viewer?file=453f5bef-laravel-2025-02-18.log&query=log-index%3A39.

**Root Cause**
I _believe_ what's happening is:
* A leaderboard entry for a user exists but is soft deleted.
* On submit by the user to the same leaderboard, the server attempts to restore the leaderboard entry.
* If the client submits two scores for the same leaderboard+user in rapid succession, both requests see the soft deleted entry, both try to restore and update the existing entry, one request restores it, and the 2nd request fails due to a duplicate key error.

Swapping the other side of the condition to use `updateOrCreate` with `'deleted_at' => null,` should keep the same business logic but gracefully handle this race condition.


app/Helpers/database/leaderboard.php
```diff
@@ -49,6 +49,7 @@ function SubmitLeaderboardEntry(
         timestamp: $timestamp,
     );
 
+    // First check if there's an existing entry (including soft-deleted)
```
Jamiras, line 52: The soft-deleted record should be found by the `withTrashed` clause. But there's still a possibility that two requests will both get to the `else` before either finishes adding the row.

Jamiras (approved):

Issue #3234, opened 2025-02-22
Broken emulator tab when reporting tickets while using BizHawk


SuperGamer18:
**Bug:**
Whenever I report a ticket while using BizHawk I get a bugged emulator tab which shows BizHawk twice and nothing else (they are both ticked, aka they have the ✓ icon next to the name).

**Steps to reproduce the behavior:**
1. Play a game using BizHawk (in this case, I played Mario Calculator for DSi).
2. Report an issue on a cheevo from that game (in this case, this one: https://retroachievements.org/achievement/492447).
3. Pick any ticket type (aka https://retroachievements.org/achievement/492447/tickets/create?type=1 and https://retroachievements.org/achievement/492447/tickets/create?type=2).
4. See the rather uncanny emulator tab

**Expected behavior:**
It should list BizHawk only once, as well as both RetroArch and RALibretro.

**Screenshots:**
(the following screenshot is in Spanish, but this should be easily recognizable)
![Image](https://github.com/user-attachments/assets/16b58857-8f0c-4649-851d-f499cf66e73a)

**Information:**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 135.0.1

**Additional context:**
I genuinely have no idea about what happened here this time. Also keep in mind that this does NOT happen when using RetroArch.

PR #3236, opened 2025-02-22
fix(BuildTicketCreationDataAction): deduplicate emulators

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3234.

**Root Cause**
I have never been able to repro this issue locally, but I can see how it would happen. Because emulator names are not normalized in `addInactiveEmulators`, subtle nuances in the decoded client string could plausibly cause a duplicate emulator to be inserted into the `$emulators` collection.

We could alternatively solve this issue in the front-end, but that feels too far downstream when it's technically a logic error in the back-end action.


app/Platform/Actions/BuildTicketCreationDataAction.php
```diff
@@ -92,6 +92,9 @@ private function addInactiveEmulators(Collection &$emulators, Achievement $achie
         if ($needsOther) {
             $emulators->add(new EmulatorData(0, 'Other (please specify in description)'));
         }
+
+        // Deduplicate without needing to normalize decoded client strings.
+        $emulators = $emulators->unique('name');
```
Jamiras, line 97: I am able to reproduce this by changing some data in the DB and then trying to create a ticket.
```
update player_sessions set user_id=[me] where game_id=22308 and user_id=13448;
update player_games set user_id=[me] where game_id=22308 and user_id=13448;
```

http://localhost:64000/achievement/285575/tickets/create?type=2

The problem is the `contains` check on line 82 is case sensitive. The DB contains "Bizhawk" as an emulator, but the User-Agent is "BizHawk". We should probably update the DB to be "BizHawk", but there's going to be a secondary issue with the new "EmuHawk" user agent. I think the correct solution is to double check the collection after resolving the emulator:

```
if (!$emulators->contains('name', $decoded['client'])) {
    $emulatorUserAgent = EmulatorUserAgent::firstWhere('client', $decoded['client']);
    if (!$emulatorUserAgent) {
        $needsOther = true;
    } else if (!$emulators->contains('name', $emulatorUserAgent->emulator->name)) { // <---
        $emulators->add(EmulatorData::fromEmulator($emulatorUserAgent->emulator));
    }
}
```

wescopeland: Repro'd and confirmed this change does resolve the issue. Updated in latest.

Jamiras (approved):

Issue #3081, opened 2025-01-19
Expanding any dropdown menu in the Settings causes the frame of the entire page to change size horizontally


FIareon:
**Describe the bug**
Expanding any Dropdown menu in the Settings page and viceversa makes the page change size. The effect seems to revolve around pixels, so the effect is significantly more drastic the smaller your screen/window size is.

**To Reproduce**
1. Go to the Settings page
2. Click on any dropdown button
3. The entire page has changed size

**Expected behavior**
The dropdown should expand and collapse without any changes to the page's size.

**Video**

https://github.com/user-attachments/assets/a8f0b729-5d1e-44a8-857a-1ad75842a3b0

**Smartphone (please complete the following information):**
 - OS: Android 10
 - Browser Chrome
 - Version 131

**Additional context**
This can be really bad on really small screens.

wescopeland:
I've confirmed this issue only affects Android. I don't have an Android device, so unfortunately I am unable to debug.

gboquizosanchez:
There is a horizontal overflow in the whole settings page, probably it's related with it.

I'll attempt to take a look.

PR #3237, opened 2025-02-22
fix(BaseSelect): remediate Android overflow issue

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3081.

This issue **only** occurs on Android Chrome - no other browser or environment.

I discovered a solution here: https://stackoverflow.com/a/76771880. From the Radix docs:

> When using position="popper" on Select.Content, you may want to constrain the width of the content so that it matches the trigger width. You may also want to constrain its height to not exceed the viewport.
> 
> We expose several CSS custom properties such as --radix-select-trigger-width and --radix-select-content-available-height to support this. Use them to constrain the content dimensions.


Jamiras (approved):
Untested. I'll take your word for it.

PR #3238, opened 2025-02-23
feat(globalRanking): support hardcore achievements & RetroPoints rankings

wescopeland:
Addresses https://github.com/RetroAchievements/RAWeb/pull/3200#pullrequestreview-2635613788.

Now, ties are fully supported when sorting by Hardcore Achievements or RetroPoints.

Screenshot 2025-02-23 at 12 51 43 PM

Screenshot 2025-02-23 at 12 51 56 PM


resources/views/pages-legacy/globalRanking.blade.php
```diff
@@ -245,7 +256,7 @@
                 $rank = $rowRank;
             }
 
-            $rankPoints = $dataPoint['Points'];
+            $rankValue = $dataPoint[$rankField];
```
Jamiras, line 259: This works fine for matches within a page, but doesn't carry over between pages.

o=10000:
image

o=10025:
image

The switch on line 253 would need to be updated for the additional cases.

I'm not certain it's important enough to address.

wescopeland: I've picked away at this one during two different working sessions on the branch, and I'm struggling to come up with a solution that both works and doesn't result in an extremely long query execution time.

I'll probably give it another shot over the weekend, but this is proving to be tough to resolve.

Jamiras: If it's too much trouble, it's probably not worth the effort. How often are people going to be looking for an exact rank by number of achievements unlocked where there's a discrepancy caused by pagination?

Jamiras (approved):

PR #3256, opened 2025-02-27
fix(ShortcodeVideo): remediate hydration error on twitch clips

wescopeland:
Resolves https://retroachievements.org/log-viewer?file=62fa1cce-laravel-2025-02-27.log&query=log-index%3A332.

**How to reproduce this issue:**
* Send a DM to anyone that contains a Twitch URL. Here's an example:
```
https://www.twitch.tv/videos/2391365391
```
* `sail pnpm build && sail artisan inertia:start-ssr`
* View the message you sent while SSR is running. In the Inertia SSR process, you should see errors. You should also see errors in the browser devtools.

**Root Cause**
`window` is not available during SSR.

When testing the fix, make sure your `VITE_BASE_URL` env var is set.

The fix should be verified on stage before deploying to prod. I have already updated the env configs of our 3 app servers.


.env.example
```diff
@@ -21,6 +21,7 @@ VITE_FEATURE_MULTISET=true
 
 APP_NAME=RetroAchievements
 APP_ENV=local
+VITE_BASE_URL=localhost
```
Jamiras, line 24: Shouldn't this be down with the other VITE settings?

https://github.com/RetroAchievements/RAWeb/blob/416f2bdfa1280e1bd227e3b4afae814c03849b6a/.env.example#L124

Jamiras (approved):

.env.example
wescopeland: Forgot that was down there. Updated in the branch and on the server.

PR #3259, opened 2025-03-01
chore: update a bunch of node.js dependencies

wescopeland:
No functional changes, just some housekeeping. I've bumped all of our packages that I know are safe to update.

The only notable thing here is we had a version mismatch where `@inertiajs/core` was on 1.x and `@inertiajs/react` was on 2.x. `php-array-reader` was also dropped, as it's unused.

We'll eventually want to migrate to React 19 and Tailwind 4 for some front-end performance improvements, but I don't think we're ready for either of those jumps yet.

**How to test this PR:**
```
sail pnpm i
sail pnpm build
```
All tests are still passing, so confidence is fairly high. Just click around on the home page a bit. You shouldn't notice any differences.


Jamiras (approved):
Didn't see any obvious issues with brief smoketesting.

PR #3260, opened 2025-03-01
feat(connect): label AOTW and AOTM in the patch response

wescopeland:
This PR makes a minor quality of life improvement to Connect API `patch` responses.

Now, if a `patch` response contains either the Achievement of the Week or the Achievement of the Month, the description of the achievement will be updated, ie:

```json
{
    "ID": 2508,
    "MemAddr": "0xH000652=1.1._0xH000652=0.100._R:0xH0006b1<10.1._0xH0006b1>=10.1._R:0x 000e94=0.1._0xH0006b1<=14.1._R:0xH0006b1>14.1._R:0xH00064b=0_0xH0006a6=2_0xH00001c=31",
    "Title": "Something's Fishy",
    "Description": "[Achievement of the Month] Defeat Kaptain K. Rool in the Lost World",
    "Points": 25,
    "Author": "Brian",
    "Modified": 1738858114,
    "Created": 1377544462,
    "BadgeName": "123989",
    "Flags": 3,
    "Type": null,
    "Rarity": 12.76,
    "RarityHardcore": 9.9,
    "BadgeURL": "http:\/\/media.retroachievements.org\/Badge\/123989.png",
    "BadgeLockedURL": "http:\/\/media.retroachievements.org\/Badge\/123989_lock.png"
},
```

Oftentimes when I start going for Achievement of the Week, I boot up the game on my couch and can't remember which achievement I actually am going for. This leads to me having to pull my phone out and jump around on our website. Inserting these "[Achievement of the Week]" and "[Achievement of the Month]" labels solves this inconvenience.

### DIFF
--- a/app/Connect/Actions/InjectPatchAotwEventDataAction.php
+++ b/app/Connect/Actions/InjectPatchAotwEventDataAction.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Connect\Actions;
+
+use App\Models\EventAchievement;
+use Illuminate\Support\Facades\Cache;
+
+class InjectPatchAotwEventDataAction
+{
+    /**
+     * Cache keys and TTL configuration for stale-while-revalidate pattern.
+     * @see https://laravel.com/docs/11.x/cache#swr
+     */
+    private const CACHE_KEY_AOTW = 'aotw_achievement_data';
+    private const CACHE_KEY_AOTM = 'aotm_achievement_data';
+    private const CACHE_TTL_CONFIG = [30, 60]; // [fresh for 30s, stale but usable for 60s]
+
+    /**
+     * Add special achievement of the week labels to AOTW/AOTM achievement descriptions.
+     *
+     * @param array<string, mixed> $response Patch response data
+     * @return array<string, mixed> modified response with AOTW/AOTM labels in descriptions
+     */
+    public function execute(array $response): array
+    {
+        $eventAchievements = $this->getEventAchievements();
+
+        // If no event achievements were found, return an unmodified response.
+        if (empty($eventAchievements)) {
+            return $response;
+        }
+
+        // Update the main achievements list.
+        if (!empty($response['PatchData']['Achievements'])) {
+            $response['PatchData']['Achievements'] = $this->processAchievements(
+                $response['PatchData']['Achievements'],
+                $eventAchievements
+            );
+        }
+
+        // Update the sets achievement list(s).
+        if (!empty($response['PatchData']['Sets'])) {
+            foreach ($response['PatchData']['Sets'] as &$set) {
+                if (!empty($set['Achievements'])) {
+                    $set['Achievements'] = $this->processAchievements(
+                        $set['Achievements'],
+                        $eventAchievements
+                    );
+                }
+            }
+        }
+
+        return $response;
+    }
+
+    /**
+     * Load the current AOTW and AOTM achievements.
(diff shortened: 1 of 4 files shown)


app/Connect/Actions/InjectPatchAotwEventDataAction.php
```diff
@@ -0,0 +1,108 @@
...
+        $aotwAchievement = Cache::flexible(self::CACHE_KEY_AOTW, self::CACHE_TTL_CONFIG, function () {
+            return EventAchievement::currentAchievementOfTheWeek()
+                ->with(['achievement', 'sourceAchievement'])
+                ->first();
+        });
+        $aotmAchievement = Cache::flexible(self::CACHE_KEY_AOTM, self::CACHE_TTL_CONFIG, function () {
+            return EventAchievement::currentAchievementOfTheMonth()
+                ->with(['achievement', 'sourceAchievement'])
+                ->first();
+        });
```
wescopeland, line 76: Both of these queries are already super fast, but for a high-traffic endpoint such as this it makes sense to use a stale-while-revalidate caching strategy.

Jamiras:
Before I even look at this, what about all the other automatic events? Should RAWR achievements be decorated, what about evergreen achievements?

wescopeland:
I had considered that, but no - I don't think so. The reason being is AOTW is the only event we feature prominently on our home page and has extra special treatment baked into our code.

Jamiras (commented):
I'm not sold on this. Have you run it by the events-team?

People who are participating in the achievement of the week/month event will get the active achievement from the website and actively focus on getting it. There's no reason to decorate the achievement for them.

For players not actively participating in the event, the additional information may not mean anything to them - especially if we choose to use the abbreviation.

Finally, the change relies on the user actually looking at the achievement list (at least in RetroArch) as the description isn't displayed on the popups.
image
image

Players on the standalone emulators will see the text in the popup, but aren't likely the target audience of this change.
image

Jamiras:
Also, this indirectly modifies the editable content, which could be sent back to the server if a fix has to be made to the achievement logic while it's active.
image

wescopeland:
> People who are participating in the achievement of the week/month event will get the active achievement from the website and actively focus on getting it. There's no reason to decorate the achievement for them.

I disagree, for the use case I described in the PR description:
> Oftentimes when I start going for Achievement of the Week, I boot up the game on my couch and can't remember which achievement I actually am going for.

I know my memory is poor, and I could be the only person who doesn't memorize what the AotW is before I boot the game, but I suspect I'm not the only one.

> For players not actively participating in the event, the additional information may not mean anything to them - especially if we choose to use the abbreviation.

It may be worth us checking to see what events the player is participating in (probably would have to be non-evergreen). If a user has at least 1 unlock in the event, maybe we add the decoration to any source achievements in the patch response.

> the change relies on the user actually looking at the achievement list (at least in RetroArch) as the description isn't displayed on the popups.

This is exactly what I want. I've launched the AotW game and played for a few minutes, but I've already forgotten which achievement I'm going for as soon as I start getting immersed in the game. I don't want to unlock my phone, go through Cloudflare verification, make sure I'm logged in, then scroll the homepage to look. I want to open the achievements list in my emulator and see the label.

wescopeland:
> Also, this indirectly modifies the editable content, which could be sent back to the server if a fix has to be made to the achievement logic while it's active.

Ah, unfortunately, I hadn't considered this. I don't think I'll have a good solution. It's probably more trouble than it's worth.

PR #3261, opened 2025-03-01
fix(EventAchievementSection): remediate 'pop' effect in collapse animation

wescopeland:
Follows up on the comment here https://github.com/RetroAchievements/RAWeb/pull/3235#discussion_r1971588240:
> I'll probably take some time in a future PR to tune this animation further. I'm still not 100% pleased with it, but it's at least passable in its current state.

This PR mostly focuses on mitigating a "pop" effect which occurs in the open animation. It's probably impossible to resolve 100%, but it can be greatly concealed. The close animation has also been slightly adjusted, but not to remediate any issues.

The open "pop" occurs right at the end of the animation, where on `master`, the final bit of content seemingly teleports onto the page. There's a single frame of animation where the easing curve abruptly ends and ~10px of content instantly appears. It's most noticeable if you focus your eyes on the "Previous" text of the collapse heading below the one being clicked on.

**Before**

https://github.com/user-attachments/assets/57b5c91b-9ce7-4453-bf01-24b0fcd9782e

**After**

https://github.com/user-attachments/assets/573c90bd-b452-44fa-929a-75f3460582a5

### DIFF
--- a/resources/js/features/events/components/AchievementSet/EventAchievementSection/useEventAchievementSectionAnimation.test.ts
+++ b/resources/js/features/events/components/AchievementSet/EventAchievementSection/useEventAchievementSectionAnimation.test.ts
@@ -59,8 +59,41 @@ describe('Hook: useEventAchievementSectionAnimation', () => {
     expect(styleSpy).not.toHaveBeenCalled();
   });
 
-  it('given opening animation, sets and cleans up styles after transition', () => {
+  it('given isInitiallyOpened is false, applies initial closed styling on mount', () => {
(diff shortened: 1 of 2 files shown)


Jamiras (approved):
Honestly, I don't think I'd be able to spot the difference if you didn't explain the change.

wescopeland:
Yea, I understand. Motion is one of the very minor details I tend to get obsessive over 😵‍💫 This would probably have bothered me forever if I didn't take another pass at fixing it.

PR #3265, opened 2025-03-02
update checklist page to use new widgets

Jamiras:
Pulls the user picker from the create message page and the achievement list grouping from the event page.

image


wescopeland:
Do you have a good sample URL I can try which will exhibit the bug we were discussing earlier?

resources/js/features/events/components/AchievementSet/AchievementGameTitle/AchievementGameTitle.tsx
```diff
@@ -19,7 +19,7 @@ export const AchievementGameTitle: FC<AchievementGameTitleProps> = ({ game }) =>
 
   return (
     <a href={route('game.show', { game })} {...cardTooltipProps}>
-      <GameTitle title={game.title} isWordWrappingEnabled={true} /> {`(${game.system!.nameShort})`}
+      <GameTitle title={game.title} isWordWrappingEnabled={true} /> {`(${game.system?.nameShort})`}
```
wescopeland, line 22: Thinking this over, maybe we should do:
```tsx
<a href={route('game.show', { game })} {...cardTooltipProps}>
  <GameTitle title={game.title} isWordWrappingEnabled={true} />
  {game.system?.nameShort ? ` (${game.system.nameShort})` : null}
</a>
```

Now, there's no possibility of displaying "(undefined)" onto the page. Can be verified by adding this test in _AchievementGameTitle.test.tsx_:
```tsx
it('given there is no system short name, does not crash or display an undefined system', () => {
  // ARRANGE
  const system = createSystem({ nameShort: undefined });
  const game = createGame({ title: 'StarCraft 64', system });

  const { container } = render(<AchievementGameTitle game={game} />, {
    pageProps: { auth: { user: createAuthenticatedUser({ displayName: 'Scott' }) } },
  });

  // ASSERT
  expect(container).toBeTruthy();
  expect(screen.queryByText(/undefined/i)).not.toBeInTheDocument();
});
```

resources/js/features/events/components/AchievementSet/AchievementsListItem/AchievementsListItem.tsx
```diff
@@ -101,61 +101,63 @@ export const AchievementsListItem: FC<AchievementsListItemProps> = ({
...
+              </span>
+              <span className="md:hidden">
+                {'–'} {formatPercentage(unlockHardcorePercentage)}
+              </span>
+              <span className="hidden sm:inline md:hidden">{t('unlock rate')}</span>
+            </p>
+
+            <BaseProgress
+              className="h-1"
+              max={1521}
```
wescopeland, line 147: Looks like I wrote a bug here. I bet this is also what's causing the progress bar scale issue. I think we can fix this by doing `max={playersTotal}`, since it can't be `null` at this point in the tree.

resources/js/pages/user/[user]/achievement-checklist.tsx
Comment (wescopeland) [inline resources/js/pages/user/[user]/achievement-checklist.tsx:1]: Doesn't need to be addressed in this PR, just a general comment.

If we find ourselves adding more complexity in here, we'll probably want everything within `<AppLayout.Main>` to be broken out of the page component and into a new component in the events feature module.

Comment (Jamiras) [reply resources/js/pages/user/[user]/achievement-checklist.tsx:1]: The User picker should probably be a tested component, and I expect we'll have more use cases for it in the future (it already appears again in #3266).

I'm not sure the nested array expansion needs testing as there's no logic there.

resources/js/features/events/components/AchievementSet/AchievementGameTitle/AchievementGameTitle.tsx
wescopeland: I get what you're saying. My only reservation is `undefined` is technically an implementation detail of Node.js. If we want to have a visual indicator, that would be okay, but it should be a translatable string.

Jamiras: How do you feel about empty string? It provides feedback that something is missing, and doesn't need to be translated.
image

wescopeland: That seems fine

resources/js/pages/user/[user]/achievement-checklist.tsx
Comment (wescopeland) [reply resources/js/pages/user/[user]/achievement-checklist.tsx:1]: It's mostly for the sake of defensiveness given underlying component logic changes & prop types don't change (no `tsc` errors).

Right now, stuff is tightly coupled to events just to keep complexity minimal. This inevitably will no longer be the case when we start porting game pages to React, which will likely start happening as soon as we cut events over.

For example, there isn't any automated sanity check at the moment that, given the expected props from Inertia, the page component doesn't crash on mount.

However, to reiterate, I don't think that's a huge deal on this specific page unless complexity grows, especially given this is mostly intended for internal use.

PR #3263, opened 2025-03-01
perf(game): limit Compare Progress, resolve N+1 issue

wescopeland:
It's currently possible to open some game pages and kick off thousands of queries to the database. This issue is easily reproducible with user 886541 on game 228.

This PR:
* Limits the Compare Progress component to a maximum of 50 users, sorted by most to least progress.
* Resolves an N+1 query problem within the component template where each loop iteration fetches a user model from the DB.

PR #3266, opened 2025-03-02
feat(events): add Compare Progress sidebar component

wescopeland:
This PR adds the "Compare Progress" component to the event page sidebar.

The back-end code includes optimizations from #3263.

An additional nuance of this new UI is the "See N more" button. Returning 50 results is still probably fine, but showing them all on load is very likely undesirable, as it forces the component to occupy the great majority of sidebar space.

The button will never read "See 1 more":
* If there are 50 results, 10 will be shown and 40 will be inside the collapse.
* If there are 11 results, all 11 will be shown.
* If there are 12 results, 10 will be shown and 2 will be inside the collapse.

I don't like buttons that expand to show just 1 thing, so trying to avoid that here.

https://github.com/user-attachments/assets/7ef66bd2-3091-488b-8ad6-48cfad31e78c

### DIFF
--- a/app/Models/UserRelation.php
+++ b/app/Models/UserRelation.php
@@ -5,9 +5,14 @@ declare(strict_types=1);
 namespace App\Models;
 
 use App\Support\Database\Eloquent\BaseModel;
+use Database\Factories\UserRelationFactory;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
 
 class UserRelation extends BaseModel
 {
+    /** @use HasFactory<UserRelationFactory> */
+    use HasFactory;
+
     // TODO rename Friends table to user_relations
     // TODO migrate Friendship column to status, remove getStatusAttribute()
     protected $table = 'Friends';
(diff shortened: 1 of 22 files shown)


Jamiras (approved):
Seems functional, but I can't actually access the compare pages for any user. That functionality seems to be broken on my master branch as well.

As best as I can tell, the UserPolicy is non-restrictive at this point, so I don't know what's going on.

You can see this on stage too. Just go to some random game, and click on the progress bar of a user in the recent players list.

wescopeland:
@Jamiras I can repro on stage, but not locally.

Since you're able to repro locally, do you see anything peculiar in the "Gate" tab of the debugbar?

Screenshot 2025-03-05 at 8 32 44 PM

Jamiras:
> do you see anything peculiar in the "Gate" tab of the debugbar

Everything is green except for one thing:

image

URL for that particular error: http://localhost:64000/user/OldSport/game/30824/compare

wescopeland:
I'm still unable to repro this locally.

However, a strange thing I notice is my `view` which is passing (and yours failing) passes in a model rather than a display name:
```
view [▼
  "ability" => "view"
  "result" => true
  "user" => 117089
  "arguments" => "[0 => Object(App\Models\User)]"
]
```

I'm going to go ahead and merge this branch, and will debug on my next DB backup from prod.

PR #3268, opened 2025-03-02
perf: reduce supervisor-2 maxProcesses to 8

wescopeland:
https://discord.com/channels/476211979464343552/718091467893243937/1339221412661170289

If this is too slow, we can try bumping from 8 to 10. Especially on Sundays, this is putting too much pressure on the server.


Jamiras (approved):
🤷‍♂️ (shrug). I have no opinion on this change. If you think it's beneficial, go for it.

PR #3270, opened 2025-03-05
fix: remediate another display_name issue

wescopeland:
This internal tool is going to eventually live in Filament -- just getting in a quick fix on this for now. At the moment, someone using this tool could get disoriented/confused if we only show `->User`.


resources/views/pages-legacy/admin.blade.php
```diff
@@ -19,7 +19,7 @@
     if ($forUser == null) {
         $message = "Unknown user: $altsForUser";
     } else {
-        $altsForUser = $forUser->User;
```
Jamiras, line 22: This should only display both if they differ.

wescopeland: Fixed in latest.

Jamiras (approved):

PR #3273, opened 2025-03-06
feat(events): add Achievement Distribution chart

wescopeland:
This PR adds the Achievement Distribution chart to the event page sidebar.

Because there isn't much data to show, these screenshots simulate usage when data is pulled from various game IDs. This can be tested locally by changing two lines of code:
```
First, pick a game ID.

---

In EventController.php, hardcode the game ID around line 58:

$playerGame = $user
    ? $user->playerGames()->whereGameId(2362)->first()
    : null;

---

In BuildGameAchievementDistributionAction.php, override $game at the start of execute():

public function execute(Game $game, ?User $user): Collection
{
    $game = Game::find(2362);
```

The action reuses our existing helper utils to build the distribution buckets. As a result, most of the logic for this component remains the same. There are a few functional changes to keep in mind:

* There isn't a y-axis anymore.
* If authenticated and the player has progress, their current hardcore & softcore positions are shown via [reference lines](https://recharts.org/en-US/api/ReferenceLine). The reference line is labeled in the hover tooltip.
* If there are 0 buckets, the component doesn't render at all.

Screenshot 2025-03-05 at 8 21 41 PM

Screenshot 2025-03-05 at 8 21 50 PM

Screenshot 2025-03-05 at 8 21 59 PM

Screenshot 2025-03-05 at 8 24 09 PM

Screenshot 2025-03-05 at 8 24 28 PM


Jamiras (approved):
Tested functionality. Did not review code.

PR #3278, opened 2025-03-07
fix: resolve user route binding issues

wescopeland:
This PR:
* Removes `getDisplayNameAttribute()`, as it's now redundant.
* Modifies `getRouteKey()` to not check if `$this->display_name` is nullable.
* Adjusts _compare-progress.blade.php_ to not use a temporary user model.
* Updates `UserFactory` so, similar to `GameFactory`, generated user models always have `display_name` values.


wescopeland:
Latest renames the `[user]` folder to `[User]`.

Jamiras (approved):

PR #3279, opened 2025-03-07
perf(game): resolve N+1 query problem with top achievers

wescopeland:
Downstream consumers are querying for `User` models in loops, derived from `user_id` values.

This PR adds a `user` field to `$retval`.

`user_id` remains in `$retval`, as it's possible the `User` model is trashed (there are checks for this at the component level).


app/Platform/Services/GameTopAchieversService.php
```diff
@@ -171,8 +172,13 @@ private function convertPlayerGames(Collection $playerGames): array
     {
         $retval = [];
         foreach ($playerGames as $playerGame) {
+            if (!$playerGame->relationLoaded('user') && $playerGame->user_id) {
+                $playerGame->load('user');
+            }
```
Jamiras, line 177: Is this different than `$playerGame->loadMissing('user');`?

wescopeland: It is, but we can simplify this to use `->loadMissing()`. Updated in latest.

PR #3281, opened 2025-03-08
chore: delete dead react component

wescopeland:
Tiny cleanup PR to remove a component that is no longer used. Discovered due to now being uncovered by tests:
```
 features/achievements/components/UnlockedAtLabel                                               |       0 |      100 |     100 |       0 |
  UnlockedAtLabel.tsx                                                                           |       0 |      100 |     100 |       0 | 2-18
```


resources/js/features/events/components/EventAchievementSets/EventAchievementSets.test.tsx
```diff
@@ -5,6 +5,9 @@ import { createAchievement, createEventAchievement, createRaEvent } from '@/test
 
 import { EventAchievementSets } from './EventAchievementSets';
 
+// Suppress "Invalid prop `max` of value `0` supplied to `Progress`".
```
Jamiras, line 8: Where is this error coming from? We should fix that instead of suppressing it. It's completely valid for a game/event to have no players and show "0 of 0". 

This error also seems to be getting spit out to the console. (see event 11).

wescopeland: I confirmed moments ago this is coming from elsewhere - I believe the Achievement Distribution chart. I've removed this comment from this branch and will resolve that issue in a separate PR.

Jamiras (approved):

PR #3285, opened 2025-03-08
feat(events): add Top Players sidebar component

wescopeland:
To the best of my knowledge, this is the final component remaining for the events page React migration. After this we just need to compare the two, fix any issues that come up, and make sure the page has plenty of polish before cutting over.

There are three possible states for the component to be in.

**Most Achievements Earned**
Appears when:
* The event only has achievements worth 1 point (ie: AOTW).
* There are <11 masters.

Screenshot 2025-03-08 at 9 44 23 AM

**Most Points Earned**
Appears when:
* The event has any achievement worth more than 1 point.
* There are <11 masters.

Screenshot 2025-03-08 at 9 46 13 AM

**Latest Masters**
Appears when:
* There are >10 players who have mastered the event.

Screenshot 2025-03-08 at 9 46 53 AM


resources/js/features/events/components/TopEventPlayers/TopEventPlayers.tsx
```diff
@@ -0,0 +1,98 @@
...
+        {listKind === 'latest-masters' ? t('Latest Masters') : null}
+        {listKind === 'most-achievements-earned' ? t('Most Achievements Earned') : null}
+        {listKind === 'most-points-earned' ? t('Most Points Earned') : null}
+      </h2>
+
+      <div className="flex flex-col gap-2 rounded-lg bg-embed p-2">
+        <BaseTable className="table-highlight overflow-hidden rounded-lg outline outline-1 outline-neutral-800">
+          <BaseTableHeader className="border-neutral-800">
+            <BaseTableRow className="do-not-highlight text-menu-link">
+              <BaseTableCell className="text-right">{'#'}</BaseTableCell>
```
Jamiras, line 46: I realize this matches existing behavior, and I'm assuming it's using the same logic/cache to retrieve the data, so that's understandable.

But should this be updated to share ranks for ties like is done on the "See More" page?
image
image

app/Platform/Data/GameTopAchieverData.php
```diff
@@ -5,17 +5,34 @@
...
+        public Carbon $lastUnlockHardcoreAt,
+        public ?Carbon $beatenHardcoreAt,
     ) {
     }
+
+    public static function fromTopAchiever(array $topAchiever): self
+    {
+        return new self(
+            userId: $topAchiever['user_id'],
+            user: UserData::from($topAchiever['user']),
```
Jamiras, line 29: > Undefined array key "user"

http://localhost:64000/event/21

Strangely, I get a similar error accessing the backing game page:

http://localhost:64000/game/6189

wescopeland: I seem to be unable to repro this locally or on stage https://stage.retroachievements.org/game/6189.

Possible something could be wrong in your local DB?

wescopeland:
> It would be nice to have the URLs use the slugs.
>
> Also, the management page permalink still points to the backing game page instead of the event page.

Yeah, we should get both of these knocked out. May need to also support a "Similar Events" component in the sidebar.

resources/js/features/events/components/TopEventPlayers/TopEventPlayers.tsx
wescopeland: Updated in latest so ties are properly handled in the UI.

app/Platform/Data/GameTopAchieverData.php
Jamiras: It's a caching issue. Check out `release`. Visit a random game page. Check out this branch. Refresh the page. The cached data doesn't have the `user` subrecord added in #3279. https://github.com/RetroAchievements/RAWeb/pull/3279/files#diff-91f62e6ba07e1193c252fed5dc40e57518c9d09705b12119af186956ffadfaa0R138

wescopeland: Ah, now it makes sense.

I think the right solution here is probably to change the cache key. Updated in latest.

Jamiras: This does fix the error, but I can't help but wonder if caching the whole user record will lead to some sort of stale data showing up in the list. Most likely the `display_name`. I think the card is dynamically generated on mouseover.

Which fields from user are actually needed?

wescopeland: I think we only really need the user display name and avatar URL.

Updated in latest to only send those two values to the UI.

app/Platform/Services/GameTopAchieversService.php
```diff
@@ -178,7 +178,8 @@ private function convertPlayerGames(Collection $playerGames): array
 
             $retval[] = [
                 'user_id' => $playerGame->user_id,
-                'user' => $playerGame->user,
+                'user_display_name' => $playerGame->user->display_name,
+                'user_avatar_url' => $playerGame->user->avatar_url,
```
Jamiras, line 182: I can't access any game pages now. Both 'lastest-masters' and 'most-points' files are still referencing `user`.

app/Platform/Services/GameTopAchieversService.php
```diff
@@ -134,7 +134,7 @@ public static function expireTopAchieversComponentData(int $gameId): void
 
     public function getTopAchieversComponentData(): array
     {
-        $cacheKey = "game:{$this->gameId}:top-achievers";
+        $cacheKey = "game:{$this->gameId}:top-achievers:v2-1";
```
Jamiras, line 137: Additional changes to this key should only affect you and I. Recommend keeping it as `:v2`.

wescopeland: Reverted back to `:v2` in latest.

app/Platform/Services/GameTopAchieversService.php
wescopeland: Fixed in latest with a few `whereIn()` queries in those components to avoid an N+1 query problem.

Jamiras (approved):

PR #3288, opened 2025-03-08
fix(user): remediate bug with toggling tracked status for players

wescopeland:
This PR resolves an issue where, on user profiles, if you attempt to toggle the tracked status for a player who is using a custom display name, the function leaves a system comment, but doesn't actually toggle their tracked status.


Jamiras (approved):
Might want to look at the `remove-avatar` component. It's still passing the User value instead of the display_name.

wescopeland:
Should be ok, looks like that's what the input expects:
```php
    'user' => 'sometimes|string|exists:UserAccounts,User',
]);
```
This makes sense given the implementation details of `removeAvatar()`:
```php
function removeAvatar(string $user): void
{
    $avatarFile = storage_path('app/media/UserPic/' . $user . '.png');
    if (file_exists($avatarFile)) {
        unlink($avatarFile);
    }
}
```

PR #3289, opened 2025-03-09
feat: add 'Set type' filter for game datatables

wescopeland:
This PR adds a "Set type" filter for game datatables, removes _gameSearch.blade.php_, and links "Hardest Games" to the All Games list, pre-sorted by most to least ratio.

By default, every list except the "All Games" list defaults to the "All Sets" option (this mirrors the current behavior in prod).

Screenshot 2025-03-09 at 11 23 12 AM

Screenshot 2025-03-09 at 11 23 18 AM

Mobile is also able to use the new filter:
Screenshot 2025-03-09 at 11 24 17 AM


resources/views/components/menu/main.blade.php
```diff
@@ -77,7 +77,7 @@
             <x-dropdown-header>Miscellaneous</x-dropdown-header>
             <x-dropdown-item :href="route('game.index')">All Games</x-dropdown-item>
             {{--<x-dropdown-item href="popularGames.php">Most Played</x-dropdown-item>--}}
-            <x-dropdown-item :href="url('gameSearch.php?p=0')">Hardest Games</x-dropdown-item>
```
Jamiras, line 80: Not sure how it's working without an explicit redirect, but hitting the old URL (http://localhost:64000/gameSearch.php?p=0) redirects to the All Games page without a sort.

resources/views/pages-legacy/gameSearch.blade.php
```diff
@@ -1,143 +0,0 @@
...
-$consoleList = System::gameSystems()->active()->get(['ID', 'Name'])->keyBy('ID')->map(fn ($system) => $system['Name']);
-$consoleList = $consoleList->sort();
-$consoleList->prepend('All Consoles', 0);
-
-$count = requestInputSanitized('c', $maxCount, 'integer');
-$offset = requestInputSanitized('o', 0, 'integer');
-$method = requestInputSanitized('p', 0, 'integer');
-$consoleID = requestInputSanitized('i', 0, 'integer');
-
-$gameData = getGameListSearch($offset, $count, $method, $consoleID);
```
Jamiras, line 18: This function is no longer referenced anywhere and can be removed.

resources/views/components/menu/main.blade.php
wescopeland: Fixed in latest.

resources/views/pages-legacy/gameSearch.blade.php
wescopeland: Removed in latest.

Jamiras (approved):

PR #3295, opened 2025-03-11
fix(HasSelfHealingUrls): swap slug-ID to ID-slug

wescopeland:
Precursor to setting up event (and other) self-healing URLs w/ slugs.

This PR swaps slug-ID to ID-slug. The only page affected by this change is `/system/{system}/games`.

There are a few subtle reasons why I've come to believe making this swap is beneficial, and it seems like it'd be better to do it sooner rather than later.

* It's better for SEO, as search engine crawlers may truncate slugs, which just so happens to chop the ID off at the end of the path, breaking self-healing. [ref](https://webmasters.stackexchange.com/a/71692)
* It supports better caching organization at the CDN level. Cloudflare uses cache keys based on URL patterns, and having consistent ID prefixes can make cache rule creation and management more predictable across resources.
* It provides cleaner and more consistent URL patterns when we have nested resources (such as forum topics).

**Before**
`/system/playstation-12/games`
... for a while now, I've found this to be kind of funny, because it reads like we support a mythical "PlayStation 12".

**After**
`/system/12-playstation/games`.

Redirect and analytics handling are also included, so the old pattern will redirect to the new one.


Jamiras:
I never really understood why we even want these. If it's just extracting the ID out of the string, why have the slug at all?

These are all effectively the same URL:
http://localhost:64000/system/4/games
http://localhost:64000/system/game-boy-4/games
http://localhost:64000/system/this-url-is-wrong-4/games
http://localhost:64000/system/1-2-3-4/games

wescopeland:
Events a good example of why self-healing URLs are desirable. Even if the slug changes, URLs will remain unbroken.

wescopeland:
I'm not sure I actually gave a good answer to your question.

Events are the obvious use case: we want event managers to have the ability to change the slug at some indeterminate point in the future without any worry of repercussions.

A primary reason websites use self-healing URLs is to improve social sharing clickthrough. Several popular channels, such as Reddit, do not actually do social embeds like what we see on Discord.

Take this URL as an example. It's impossible to know what this actually is.
https://www.amazon.com/dp/B09M4FHV4Z

But these all work, and I have context/added confidence in what I'm clicking on:
https://www.amazon.com/Retro-Fighters-Brawler64-Wireless-N64-Controller/dp/B09M4FHV4Z
https://www.amazon.com/Brawler64-Wireless-Controller/dp/B09M4FHV4Z
https://www.amazon.com/N64-Wireless-Controller/dp/B09M4FHV4Z

Similarly:
https://retroachievements.org/game/1

This isn't as cryptic as "dp" (details page), but given no embed, I'm not really sure what this may be before I click on it. https://retroachievements.org/game/1-sonic-the-hedgehog is more informative and more likely to entice.

Jamiras:
> these all work, and I have context/added confidence in what I'm clicking on

I guess that makes sense.

> Events is a good example of why self-healing URLs are desirable. Even if the slug changes, URLs will remain unbroken.

Then why even have a customizable slug? Just let them auto-generate like the other models.

When I added a slug field to events, I was picturing completely obscuring the ID:
http://localhost:64000/event/aotw-2025

Not:
http://localhost:64000/event/1-aotw-2025

So why not just have (auto-generated):
http://localhost:64000/event/1-achievement-of-the-week-2025

Jamiras:
Also, from the page you linked.
> From the SEO prospective, i bet there might be a difference:
> 1. In you first example, the ID is separated from the title, making it clear to the crawler it is a different resource (as the / character does it naturally).
> 2. In your second example, the ID is mixed with the title. It requires more brains from the crawler to determine the meaning of it.

Would it be possible to use a different separator between the ID and slug? I understand that the most desirable ones are reserved or would get escaped.

Jamiras (approved):
Approved for what it does. Still have a couple questions about the general functionality.

wescopeland:
> Then why even have a customizable slug? Just let them auto-generate like the other models.

We can drop the column if that's desirable. I don't have any strong opinions on supporting customizable slugs.

> Would it be possible to use a different separator between the ID and slug?

Underscores come to mind, ie:
```
retroachievements.org/game/1_sonic-the-hedgehog
```

Jamiras:
>> Then why even have a customizable slug? Just let them auto-generate like the other models.

> We can drop the column if that's desirable. I don't have any strong opinions on supporting customizable slugs.

Customizable slugs were meant to provide a URL that players could remember (i.e. /event/aotw-2025). If the player still has to remember the ID, I see little benefit to keeping them.

>> Would it be possible to use a different separator between the ID and slug?

> Underscores come to mind, ie:
> `retroachievements.org/game/1_sonic-the-hedgehog`

I'm not sure that's any better, but I don't have any other suggestions either.

wescopeland:
> I'm not sure that's any better, but I don't have any other suggestions either.

One benefit of the self-healing nature of these URLs is if we do come up with a better pattern, any URLs in the wild will correctly redirect to the new path we settle on.

I'm going to go ahead and merge this for now so I'm in good shape to continue with the events URL work.

PR #3298, opened 2025-03-12
fix(EventAwardTiers): display even when only 1 tier

wescopeland:
Given an event with only a single award tier (eg: http://localhost:64000/event/3), the award tiers component will now display in the sidebar on React event pages.


resources/js/features/events/components/EventAwardTiers/EventAwardTiers.tsx
```diff
@@ -10,7 +10,7 @@ interface EventAwardTiersProps {
 export const EventAwardTiers: FC<EventAwardTiersProps> = ({ event }) => {
   const { t } = useTranslation();
 
-  if (!event.eventAwards?.length || event.eventAwards.length === 1) {
+  if (!event.eventAwards?.length) {
```
Jamiras, line 13: This works fine if the event has a single tier, but several events don't have any tiers and just award a "tier 0" badge at mastery (see events 2, 13, and 19).

Here's where the badge gets awarded: https://github.com/RetroAchievements/RAWeb/blob/b4eace3d02fdd1afd3cd4dad84f2c60151e586e1/app/Platform/Actions/RevalidateAchievementSetBadgeEligibilityAction.php#L177-L181

wescopeland: I hadn't considered this "virtual tier" case. I decided to handle this completely in React rather than messing with the back-end DTO(s).

Here's what it looks like in latest:
Screenshot 2025-03-12 at 8 11 21 PM

I removed the award title (it's redundant) and points/achievements required if the award tier is a virtual one. The points/achievements required badge has some inner padding which makes it look weird when positioned directly above the earners count.

Jamiras: It doesn't reflect the player having earned it.
image

wescopeland: Fixed in latest.

Screenshot 2025-03-13 at 5 36 41 PM

Jamiras (approved):

PR #3300, opened 2025-03-12
refactor(events): drop the slug column

wescopeland:
Per https://github.com/RetroAchievements/RAWeb/pull/3295#issuecomment-2718120421, drops `events.slug` in favor of ultimately switching over to automated slugs.

As best as I can tell, this value currently isn't exposed anywhere other than the management app.


database/migrations/2025_03_12_000000_update_events_table.php
```diff
@@ -0,0 +1,23 @@
...
+            $table->dropUnique('events_slug_unique');
+            $table->dropColumn('slug');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('events', function (Blueprint $table) {
+            $table->string('slug', 20)->after('image_asset_path');
+            $table->unique('slug', 'events_slug_unique');
```
Jamiras, line 20: This won't work with existing records in the table. I'm fine with leaving the unique constraint off for the reverse migration step.

Jamiras (approved):

PR #3313, opened 2025-03-14
refactor(EventController): thin out show() in favor of action-oriented logic

wescopeland:
No functional changes. Thins out `EventController::show()` by sweeping the logic into a handful of actions.


Jamiras (approved):
While I could see breaking these out into helper functions, I don't see the benefit of making them separate actions. These are tightly coupled to the EventController. Nothing else is using them yet, and I'm not sure anything else would. But now they're unique entities that are given their own nodes in the file tree.

The only other reason I can think of for making them separate is unit testing, and pretty much the only logic in any of them is null checks.

wescopeland:
You raise a valid point about the cost-benefit tradeoff of the action class extraction from the controller. In my view:

* Ideally, `LoadGameTopAchieversAction` is reused by `GameController`.
* Other controllers, such as `HomeController`, are following a clear action-oriented arch pattern.
* While the actions could be private methods in the controller, having them as separate classes has a minor benefit of keeping the controller focused on HTTP-domain concerns.

I think the core thing I'm shooting for here is, similar to `HomeController`, if someone wants to pick away at a certain piece of back-end logic on the page, they don't have to navigate a labyrinth of other code to find it. Similarly, if someone wants to add new logic to the page, a pattern is already established in the controller on how to isolate that logic.

Jamiras:
Unrelated. I updated some data in the database and didn't force the metrics to be updated.

PR #3314, opened 2025-03-15
feat(events): use slug urls

wescopeland:
This PR sets up slug URLs on event show pages.

To test this, navigate to http://localhost:64000/event/1.

Additionally, the permalink in the management app has been updated.


Jamiras (approved):
Not specifically related to this PR: The "View this year's event" link on the main page still points to the game record.

Jamiras:
So do `[game=32718]` style links, though I expect the latter to still evaluate to a game page, which means that we still need to hook up game pages redirecting to their event wrapper counterparts.

wescopeland:
Yep, the only thing this branch does is enable the slug URLs.

https://github.com/RetroAchievements/RAWeb/pull/3318 begins working on the cutover, and has a few open questions we'll need to think through.

PR #3318, opened 2025-03-15
feat(events): cut over from game.show to event.show

wescopeland:
All the open PRs probably put us in a good position where we can cut over from gameInfo.blade.php to our React event page.

This PR rips automated events stuff out of gameInfo.blade.php and sets up a redirect for all existing links.

There are a few other things we may also want to think about:
* `[event]` shortcode in the forum and DMs? This may also imply we should have a hover card for events.
* How do we handle adding events to hubs if the game ID is obfuscated? I guess it still works as-is, but could be cumbersome.


Jamiras:
> How do we handle adding events to hubs if the game ID is obfuscated? I guess it still works as-is, but could be cumbersome.

Hubs can be added to events without needing to know the event's game ID. The only thing that would be difficult is linking an event to another event - which, while possible, is not currently reflected on the event page.

Game page:
image

Event page:
image

Jamiras:
> [event] shortcode in the forum and DMs? This may also imply we should have a hover card for events.

That would make sense. We did the same thing for hubs.

Jamiras (approved):

wescopeland:
> The only thing that would be difficult is linking an event to another event - which, while possible, is not currently reflected on the event page.

Do you have a link to the URL that shows this? I can't seem to find it.

Jamiras:
I'm not sure if any of the existing events have the association, but the DevQuests that have `-II`, `-III` do. Here's one: http://localhost:64000/game/15940

If you've converted it to an event via the command, you can click the Manage button, Details tab, and Permalink to get to the wrapper event page.

PR #3319, opened 2025-03-15
chore: set up Sentry on the server

wescopeland:
If you haven't taken a peek at our Sentry dashboard yet, an invite should have landed in your inbox. You may even be able to log in with GitHub. [Here's a sample error](https://retroachievementsorg.sentry.io/issues/33510952/?referrer=alert_email&alert_type=email&alert_timestamp=1742055154815&alert_rule_id=128347&notification_uuid=3b7cd285-d489-437d-9244-c48032ab0123&environment=local).

Sentry ultimately should get us full-stack error traces when something goes wrong in our Inertia stuff (ie: a client-side or hydration error). Right now these are nearly impossible to trace in log-viewer, and I have to take educated guesses. Additionally, Sentry is smart enough to pinpoint the likely commit or PR that caused an error.

This branch doesn't integrate yet with our Inertia/React stuff - I want to see events from PHP and Blade UIs flowing into Sentry before tackling that.

I've tested this locally using our SENTRY_LARAVEL_DSN and was able to successfully record an error with a full stack trace (linked above). It's tough to say if the config here is perfect though until it's actually deployed, so I want to keep the surface area for something to go wrong minimal.

Not much to do here to test locally other than land on a non-Inertia page and verify it doesn't explode.


Jamiras (approved):
Untested

PR #3320, opened 2025-03-15
feat: add 'Game type' filter for game datatables

wescopeland:
This PR adds a "Game type" filter for game datatables. This new filter leverages `tags` and `taggables` (from https://github.com/RetroAchievements/RAWeb/pull/2963) for optimal performance.

The filter is available on both desktop and mobile.

Screenshot 2025-03-15 at 3 49 16 PM


app/Platform/Concerns/BuildsGameListQueries.php
```diff
@@ -151,6 +151,26 @@ private function applyFilters(Builder $query, array $filters, ?User $user = null
                 continue;
             }
 
+            /*
+             * only show games based on their tags
+             */
+            if ($filterKey === 'game-type' && !empty($filterValues)) {
+                if (in_array('retail', $filterValues)) {
```
Jamiras, line 158: While the UI doesn't allow it, you can pass multiple filter values by modifying the URL. (i.e. http://localhost:64000/system/4-game-boy/games?filter%5Bgame-type%5D=demo%2Chomebrew). This logic assumes that retail excludes all the other values (which may be semantically true, but a user might want to search on retail,demo,prototype and ignore hacks and homebrews).

And on that note, should the UI allow multi-select?

app/Platform/Concerns/BuildsGameListQueries.php
```diff
@@ -151,6 +151,26 @@ private function applyFilters(Builder $query, array $filters, ?User $user = null
...
+                    $query->whereNotExists(function ($subquery) {
+                        $subquery->select(DB::raw(1))
+                            ->from('tags')
+                            ->join('taggables', 'tags.id', '=', 'taggables.tag_id')
+                            ->whereColumn('taggables.taggable_id', 'GameData.ID')
+                            ->where('taggables.taggable_type', 'game')
+                            ->where('tags.type', 'game');
+                    });
+                } else {
+                    $query->withAnyTags($filterValues, 'game');
```
Jamiras, line 169: This results in something extracting the english json tags from the tags table. Does it work in other languages? As far as I can tell, the json in the table only has english entries.

image

app/Platform/Concerns/BuildsGameListQueries.php
wescopeland: Looks like our other single-select filters are affected by this "URL injection" bug as well:

Screenshot 2025-03-23 at 1 50 40 PM

I lean towards keeping this a single-select right now for two reasons:
* I'm not sure how to handle when users will want a logical AND vs a logical OR, especially if @televandalist's tag revamps go through at some point in the foreseeable future.
* A single-select likely covers 80%+ of the current use cases for the control. If/when people clamor for additional functionality in the filter, we can build it based on feedback.

app/Platform/Concerns/BuildsGameListQueries.php
wescopeland: This is an unfortunate quirk of `spatie/laravel-tags`. There was some prior discussion in #2963:

> Another thing to keep in mind is spatie/laravel-tags localizes tags, and there is no way to disable this behavior:
> We should probably only store "en" values in the database, and let the React UI do the presentation/localization work.

I've confirmed other locales can use this filter without any issues.

app/Platform/Concerns/BuildsGameListQueries.php
wescopeland: https://github.com/RetroAchievements/RAWeb/pull/3344 resolves the injection bug on single select filters.

Jamiras (approved):

PR #3321, opened 2025-03-15
fix(SetTypeFilter): don't show label when default value is set

wescopeland:
This PR makes a minor tweak to the new "Set type" filter on game datatables.

Now, when "All Sets" is undefined (which causes the back-end to fall-through on the filter itself), a corresponding label is not shown in the filter. This saves a little bit of space in the toolbar on desktop.

**Before**
Screenshot 2025-03-15 at 4 08 01 PM

**After**
Screenshot 2025-03-15 at 4 07 06 PM


Jamiras:
This doesn't seem to do anything for me. I still see "All Sets" on initial landing of the system games list or want to play games list. Changing the filter to something else and back to "All Sets" also still shows "All Sets".

Jamiras (approved):
Nevermind. I was apparently testing on prod :(

Issue #3091, opened 2025-01-22
Code notes created link takes you back to the dev's page


SuperGamer18:
**Bug:**
Every time I try to access AlexHidanBR's code notes, it takes me back to his user page, but not before going to this link (https://retroachievements.org/individualdevstats.php?u=AlexHidanBR#code-notes) and ending up on this one (https://retroachievements.org/user/AlexHidanBR#code-notes). Keep in mind that AlexHidanBR is a Jr. Dev and only has 2 code notes, both of which are for his first set (https://retroachievements.org/game/10282). I tried this with another developer (tho this user a full dev and not a Jr.), TeddyWestside, and it worked normally (it only takes me to https://retroachievements.org/individualdevstats.php?u=TeddyWestside#code-notes this time and actually shows the dev stats).

**Steps to reproduce the behavior:**
1. Go to AlexHidanBR's userpage (https://retroachievements.org/user/AlexHidanBR).
2. Click on the number of code notes created.
3. See how it just takes you back to the userpage.

**Expected behavior:**

**Screenshots:**
Screenshots will not work for this. At best, a couple of GIFs.
![Image](https://github.com/user-attachments/assets/b565816c-c7c5-4e21-9961-edda38c820e3)
![Image](https://github.com/user-attachments/assets/8e716592-41e1-4b7e-8c4e-126b3d42436b)
**Information**
 - OS: Windows 10
 - Browser: Firefox
  - Version: 134.0.1 (64-bit)

**Additional context:**
Same as before, might not be platform/browser specific.

wescopeland:
This is occurring because the user has no achievements developed, so there are no stats to display. At best, we should remove the hyperlink in this case.

PR #3330, opened 2025-03-18
fix(user): remediate some jrdevs having dead hrefs on their profile

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3091.

Individual dev stats cannot be viewed if the user has no published achievements on the website.

**Before**
Screenshot 2025-03-18 at 4 32 20 PM

**After**
Screenshot 2025-03-18 at 4 32 10 PM

### DIFF
--- a/app/Community/Components/UserProfileMeta.php
+++ b/app/Community/Components/UserProfileMeta.php
@@ -100,7 +100,7 @@ class UserProfileMeta extends Component
         $codeNotesCreatedStat = [
             'label' => 'Code notes created',
             'value' => localized_number($totalAuthoredCodeNotes),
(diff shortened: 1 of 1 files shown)


Jamiras (approved):
While this is definitely a solution, I don't think it's the correct one. The individual dev stats page should gracefully handle the case where the user has no published achievements.

There are three Jr. Devs with claims that have both code notes and leaderboards and just haven't published their sets yet.
```
select User from UserAccounts where ID in (select distinct author_id from LeaderboardDef) and ID not in (select distinct user_id from Achievements) and Permissions >= 2;
```

And, while unlikely, a developer could choose to just add leaderboards to existing sets and never publish any achievements.

PR #3333, opened 2025-03-19
feat(emulators): add new tables to support a revamped Downloads page

wescopeland:
This PR adds a new migration to create several new DB tables to support a revamped downloads page.

The screenshot below is a concept and not currently driven by real data. This component is not part of the PR. This is provided for illustrative purposes to better convey where I'd like the downloads page to get to:

Screenshot 2025-03-19 at 7 33 49 PM

I'm envisioning a downloads page that allows the user to search for a client in multiple ways. Ideally, the user can search for clients that support their platform (Windows, Linux, Android, etc - the page could even auto-detect this via user agent). It'd also make sense to be able to select a system and see all emulators for that system. We should also list front-ends who integrate with us.

We should also be able to support all this rich functionality while simultaneously being able to manage it all in Filament. Ideally, we aren't constantly making code updates to the page and it's mostly DB-driven.

Screenshot 2025-03-19 at 7 36 01 PM

To drive all this functionality, we need several new tables & accompanying relations to `emulators`.

* `platforms`
Backed by a new resource in Filament. This lets us connect the dots between a client and the platforms it can run on.

Screenshot 2025-03-19 at 7 39 34 PM

* `emulator_platforms`
Backed by a new relation manager in Filament on `EmulatorResource`. This allows a release manager to connect platforms to an emulator, with the assumption that the platform being connected is one where a RetroAchievements connection exists. In other words, we wouldn't connect "Android" to "Dolphin" at this time.

Screenshot 2025-03-19 at 7 41 57 PM

* `emulator_downloads`
Allows overriding the download URL on a platform-basis for an emulator. Ideally, the download URL in `emulators` is the canonical one, but this table can be used when it's preferable to override that value, such as to an App Store URL for an emulator that's also available on Windows/Linux.

This branch doesn't include any code in Filament to support `emulator_downloads` at the moment.


app/Platform/Enums/PlatformExecutionEnvironment.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace App\Platform\Enums;
+
+enum PlatformExecutionEnvironment: string
+{
+    /** Windows, macOS, Linux */
+    case Desktop = 'desktop';
+
+    /** Android, iOS */
+    case Mobile = 'mobile';
```
Jamiras, line 13: Would the SteamDeck be desktop or mobile? Does it matter? I think it qualifies as Linux.

app/Platform/Enums/PlatformExecutionEnvironment.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /** Xbox, PlayStation, Switch */
+    case Console = 'console';
+
+    /** Raspberry Pi, similar SBCs */
+    case SingleBoard = 'single_board';
```
Jamiras, line 19: Don't these usually run a flavor of Linux too?

app/Policies/PlatformPolicy.php
```diff
@@ -0,0 +1,70 @@
...
+    public function update(User $user, Platform $platform): bool
+    {
+        return $user->hasAnyRole([
+            Role::RELEASE_MANAGER,
+        ]);
+    }
+
+    public function delete(User $user, Platform $platform): bool
+    {
+        return false;
```
Jamiras, line 58: Can still delete platforms using Bulk actions

app/Platform/Enums/PlatformExecutionEnvironment.php
wescopeland: I also think we'd treat it as Linux.

app/Platform/Enums/PlatformExecutionEnvironment.php
wescopeland: Not necessarily, for example: https://www.replayos.com/

wescopeland: Granted - that's still a Linux distro. But it only runs on SBCs.

app/Policies/PlatformPolicy.php
wescopeland: Weird, I'm surprised Filament didn't pick up on the policy here. Updated in latest to remove the bulk action from the resource.

Jamiras (approved):
Changes approved. Still have some questions on how this data will be used.

app/Platform/Enums/PlatformExecutionEnvironment.php
wescopeland: I'm thinking `PlatformExecutionEnvironment` mostly exists to help inform UI concerns, such as what icon gets displayed here:
Screenshot 2025-03-24 at 4 11 27 PM

It's also nullable, as I suspect there may be cases where some platform may not fit neatly into one of these boxes.

For Steam Deck specifically, it's probably classed best as a desktop device, as despite its form factor, it is running a full desktop operating system.

app/Platform/Enums/PlatformExecutionEnvironment.php
wescopeland: We probably won't be able to detect if a user is hitting the downloads page directly with their Raspberry Pi. Raspian sets it user agent correctly, but it doesn't look like `jenssegers/agent` has any logic to sniff it out.

That's probably okay. There are still UI/UX benefits to setting the field. For example, we could use different iconography for SBC platforms, as well as visually separate them from Desktop, Mobile, etc.

Issue #3304, opened 2025-03-13
URL tag falsely converted to game tag


samtimes1:
When using the URL tag in a forum post for certain URLs it gets converted to a Game tag, even if it is an external URL.

**To Reproduce**
1. Go to any Forum (I noticed and mentioned it [here](https://retroachievements.org/viewtopic.php?t=16425&c=272449#272449))
2. Write the following as a new post: `[url=https://community.arduboy.com/t/longcat-puzzle-game/10331]link[/url]`
3. Click "Preview"

**Expected behavior**
The URL tag should not be converted to a game tag and remain. There should be text saying "link" that is a link to the linked website.

PR #3338, opened 2025-03-21
fix: remove regex for external URL shortcodes normalization

Danielopes7:
This PR ensures that the normalize_shortcodes function does not modify [url] tags containing external URLs. 

bug from #3304


wescopeland:
Hi @Danielopes7 👋 

Thanks for taking the time to attack this issue!

Unfortunately, this change doesn't work in Inertia+React contexts, which we're currently in the process of transitioning all this stuff to. Right now, some of the shortcode processors are in PHP, while others are in React, and we're trying to briefly maintain functional parity between the two.

For example: http://localhost:64000/message-thread/176479
Screenshot 2025-03-21 at 7 55 30 PM

Danielopes7:
> Hi @Danielopes7 👋
> 
> Thanks for taking the time to attack this issue!
> 
> Unfortunately, this change doesn't work in Inertia+React contexts, which we're currently in the process of transitioning all this stuff to. Right now, some of the shortcode processors are in PHP, while others are in React, and we're trying to briefly maintain functional parity between the two.
> 
> For example: http://localhost:64000/message-thread/176479 Screenshot 2025-03-21 at 7 55 30 PM

Hi @wescopeland,

That's true, I didn't cover this scenario before so I have just pushed a commit to fix it.

wescopeland (approved):
Verified locally.

I can't remember what that pattern was originally there for. But I also can't seem to induce any bugs with it being removed.

Thanks!

Jamiras:
> I can't remember what that pattern was originally there for. But I also can't seem to induce any bugs with it being removed.

Prior to [2361](https://github.com/RetroAchievements/RAWeb/commit/ff7a53d8a1e75f2316eb11e138d2a1961d5d68fa#diff-ae7b790313cc35697a8b3021d02308696761e42ef1443adf13ca7816ae5eb68cL81), it was only matching `user/`
```
        "~\[url[^\]]*?user/([\w]{1,20})(/?(?![\w/?])).*?\[/url\]~si",
```

Since it's not matching a domain, I can only assume it's matching things like `[url=...]user/Jamiras[/url]`, but I don't know why someone would construct something like that.

PR #3341, opened 2025-03-22
refactor: convert UserGameListType to a native enum

wescopeland:
Prep PR for migrating the Want to Develop page to React. This PR converts `UserGameListType` from an abstract class to a native enum. No functional changes.


resources/js/common/hooks/mutations/useAddToGameListMutation.ts
Jamiras, line 1: Not sure if there's anything we can do about it, or if its worth the effort, but any pages that are open before switching to this code will generate an error when adding/removing items to the want to play list.

Jamiras (approved):

resources/js/common/hooks/mutations/useAddToGameListMutation.ts
wescopeland: It's good to point out.

I've slapped together a solution locally that does respect the old interface, but after putting it together, TBH I think I'd rather just rip the band-aid off for this time around.

PR #3346, opened 2025-03-24
Add command for converting existing event games to event records

Jamiras:
Provides several different conversions to wrap games of console 101 with event records.

This migration is uni-directional. There's no command to delete the event wrapper.

The list of event games and the conversion I've chosen for each is listed [here](https://docs.google.com/spreadsheets/d/1QHDl3Zv9GlVMoa5NJzuCSCHNcm4aJWnmUlaxu2bGwrM/edit?usp=sharing). I've had Sutarion give his approval. I'll probably ping the rest of the event team in the next couple days.

To convert everything:
```
sail artisan ra:sync:events
```

To convert a single game:
```
sail artisan ra:sync:events 15949 --test
```
The `--test` flag is optional and outputs the result of the conversion without committing it.

-----

Before doing the full conversion, a little data sanitization is recommended:

This user had his unlocks transferred between accounts, but some of the player_games records weren't updated or created at a later point. (I ran this on prod earlier today)
```
update player_games set user_id=42326 where user_id=98352 and game_id in (8000, 18858, 15937, 862, 22097, 9858, 3904, 519, 10830, 15907, 15953);
```

This event has achievements waiting in unofficial. They should be promoted before running the conversion. (I have not run this in production yet)
```
update Achievements set Flags=3 where GameID=31744;
```

Several additional touch-ups are listed in the document linked above. They're not required for testing.

-----

There will be at least one more commit once I fix the conversion for 7970. The chosen conversion left softcore unlocks behind.

I also need to look into why the April Fools' sets aren't showing the Award Tiers widget.


app/Platform/Commands/SyncEvents.php
```diff
@@ -0,0 +1,2199 @@
...
+
+// Base class - provides helper functions used by the individual converters below.
+class ConvertGame
+{
+    protected string $slug;
+    protected ?string $activeFrom = null;
+    protected ?string $activeThrough = null;
+    protected bool $noWinners = false;
+
+    public function convert(Command $command, int $gameId): void
```
wescopeland, line 594: Still in the process of reviewing, but wanted to fire one comment off before a review is complete.

I'm wondering if this function in particular should leverage transactions, ie:
```php
public function convert(...): void
{
  DB::beginTransaction();

  try {
    $game = Game::find($gameId);
     // ... stuff ...

    $this->process($command, $event);
    DB::commit();
  } catch (Exception $e) {
    DB:rollBack();
    // add some logging
  }
}
```

This way in the off-chance something goes wrong in the prod transformations, it's easier to recover from without manual intervention. This might make sense given the transformations are both complex and unidirectional.

Jamiras: Convert is already wrapper in a transaction, but only in `--test` mode:
```
            if ($test) {
                $before = $conversion->captureBefore($gameId);
                DB::beginTransaction();
            }

            try {
                $conversion->convert($this, $gameId);

                if ($test) {
                    $this->dumpConverted($this, $gameId);
                    $conversion->validate($this, $gameId, $before);
                }
            } catch (Exception $e) {
            }

            if ($test) {
                DB::rollBack();
            }
```
I could expand that for non-test mode too.

wescopeland (approved):
Tested several events locally. Happy to deploy, but I'll feel most comfortable leaving the conversions in prod to you.

PR #3351, opened 2025-03-26
feat(connect): add set id param to uploadachievement and uploadleaderboard

wescopeland:
This PR:
* Adds a new `game_achievement_sets.id` param to `uploadachievement` ("s") and `uploadleaderboard ("p").
* Modifies the patch responses to always include `GameAchievementSetID`, even for root-level data.

Now, `g` is optional for both of these routines. If `s`/`p` are provided instead, the game ID is pulled from the game achievement set.


app/Connect/Actions/BuildClientPatchDataAction.php
```diff
@@ -252,13 +252,18 @@ private function buildAchievementsData(
...
+        ?string $richPresencePatch,
+        ?Game $titleGame,
+        ?int $coreGameAchievementSetId = null,
+    ): array {
         // If a title game is provided, use its title and image.
         $titleGame = $titleGame ?? $game;
 
         return [
             'ID' => $game->id,
+            'GameAchievementSetID' => $coreGameAchievementSetId,
```
Jamiras, line 266: I'm not sure this is necessary. Top-level records will always submit via game ID and nested records will always submit via set id.

* Patch data for a core hash returns the core set at the top level and the subset as a subset.
* Patch data for a subset hash returns the core set at the top level and the subset as a subset.
* Patch data for an exclusive subset hash returns the subset at the top level and no subsets.

This is particularly important for set development as the local changes are stored in an `GAMEID-User.txt` file. The subset data is also currently stored in a `SUBSETGAMEID-User.txt`, which will still work for the exclusive subsets using the top-level game ID paradigm, but won't work for basic subsets which no longer provide a game ID.

I still need to figure out how to store local data for a subset. Should it be directly in the `GAMEID-User.txt` file with some sort of subset identifier on the subset achievements, or in a `Subset-SUBSETID-User.txt`? 

What if users already have a `SUBSETGAMEID-User.txt` file? How can we find that to migrate it to the new location?

If we want to completely eliminate the game ID for subsets, we need some way to know that a top-level record is a subset so we can submit updates using the subset ID and look for the `Subset-SUBSETID-User.txt` file, which would imply that subset data cannot be stored in the parent game's User.txt.

At that point, maybe we could also always submit the top-level records via subset ID, but I still think games need to use `GAMEID-User.txt` files.

app/Helpers/database/achievement.php
```diff
@@ -212,12 +214,23 @@ function UploadNewAchievement(
...
+    ?int $gameAchievementSetID
 ): bool {
+    if (!$gameAchievementSetID && !$gameID) {
+        $errorOut = "You must provide a game ID or a game achievement set ID.";
+
+        return false;
+    }
+
+    if ($gameAchievementSetID) {
+        $gameAchievementSet = GameAchievementSet::findOrFail($gameAchievementSetID);
```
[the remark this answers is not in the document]
wescopeland: Updated in latest to return an error output.

app/Helpers/database/leaderboard.php
```diff
@@ -308,15 +309,27 @@ function SubmitNewLeaderboard(int $gameID, ?int &$lbIDOut, User $user): bool
...
+    ?int $gameAchievementSetID
 ): bool {
+    if (!$gameAchievementSetID && !$gameID) {
+        $errorOut = "You must provide a game ID or a game achievement set ID.";
+
+        return false;
+    }
+
+    if ($gameAchievementSetID) {
+        $gameAchievementSet = GameAchievementSet::findOrFail($gameAchievementSetID);
```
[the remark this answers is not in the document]
wescopeland: Updated in latest to return an error output.

app/Connect/Actions/BuildClientPatchDataAction.php
wescopeland: Removed this `GameAchievementSetID` field from the base set in latest.

app/Helpers/database/achievement.php
```diff
@@ -212,12 +214,23 @@ function UploadNewAchievement(
...
+        $gameID = $gameAchievementSet->game_id;
+    }
+
     $gameData = getGameData($gameID);
     $consoleID = $gameData['ConsoleID'];
-    $consoleName = $gameData['ConsoleName'];
-    $isEventGame = $consoleName == 'Events';
+    $isEventGame = $gameData['ConsoleID'] == System::Events;
```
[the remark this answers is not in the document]
wescopeland: Went ahead and added an error out case here if `$gameData` is null.

app/Connect/Actions/BuildClientPatchDataAction.php
Jamiras: I think this is fine for our current purposes, but does it address future needs?

Do you see us ever getting to a point where a subset won't have a backing game record? I know that we have a similar problem with events right now. Anything with achievements has to have a game record for all the foreign key relationships and aggregation logic.

If that's possible, maybe we need to move to a paradigm where exclusive subsets are still returned as children of an empty parent game? This might be desirable anyway so that the parent game's code notes can be downloaded and used, though it might have the undesirable side effect of only running the parent game's rich presence...

wescopeland: I do think it's probably desirable for us to get to a point where a subset doesn't have a backing game record.

Events should likely do the same, but events have the benefit of not needing to worry about backwards compatibility concerns (at least, from the POV of dealing with a client).

This was my original motivation for adding the field, but us actually getting to this point is probably at least a couple of years away due to client backwards compatibility concerns.

Do you think we should keep `GameAchievementSetID` removed for the time being, or re-add it and just pretend it doesn't exist until we finally need it somehow?

wescopeland: In latest, I believe I've implemented what we discussed earlier.

Everything at the `Sets` level now has a new `GameID` field.
At the root level, a `ParentID` field is always included. If `m` is a hash associated with game ID 1, both "ID" and "ParentID" will both be 1. If `m` is a hash associated with a specialty or exclusive game ID, "ID" and "ParentID" will differ.

A good way to test the behavior locally is to modify the "type" value of game_achievement_set id 20022:
Screenshot 2025-04-06 at 11 07 58 AM
Rotate it between "bonus", "specialty", and "exclusive".

There is only one MD5 hash associated with the subset "game" ID: "09dadb5071eb35050067a32462e39c5f".
```
GET http://localhost:64000/dorequest.php?u=...&t=...&r=patch&m=09dadb5071eb35050067a32462e39c5f
```
While rotating the set type, you can see what the various responses look like:
Screenshot 2025-04-06 at 11 11 19 AM

wescopeland:
I didn't realize this branch had a merge conflict. That should now be resolved. @Jamiras please let me know if there's anything else that would be good to work through here.

Jamiras (approved):

PR #3352, opened 2025-03-26
fix(UpdateGameClaimAction): don't email banned users

wescopeland:
Resolves an issue where banned users are receiving email after claims are completed.

### DIFF
--- a/app/Community/Actions/UpdateGameClaimAction.php
+++ b/app/Community/Actions/UpdateGameClaimAction.php
@@ -109,7 +109,6 @@ class UpdateGameClaimAction
         $game = Game::find($claim->game_id);
         $game->achievementSetClaims()->active()->update(['Status' => $claim->Status, 'Finished' => $claim->Finished]);
 
-        // TODO: these emails should be queued and sent asynchronously
         if ($claim->SetType === ClaimSetType::Revision) {
             // Send email to users who had previously mastered the set
             $userAwards = PlayerBadge::with('user')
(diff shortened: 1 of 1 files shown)


app/Community/Actions/UpdateGameClaimAction.php
```diff
@@ -118,7 +117,7 @@ private function processCompletedClaim(AchievementSetClaim $claim, User $current
                 ->get();
 
             foreach ($userAwards as $userAward) {
-                if ($userAward->user) {
+                if ($userAward->user && !$userAward->user->banned_at) {
```
Jamiras, line 120: `sendActivityEmail` uses a permissions check:
```
getUserPermissions($user) < Permissions::Unregistered
```
Should this?

What about other emails? I don't see any check on the `PrivateMessageReceivedMail` or changing friend status.

wescopeland: I checked all the other email sources, but I believe they're fine to leave as-is for the time being.

`sendActivityEmail` likely requires a larger refactor to avoid an N+1 query issue.

I don't believe the other mailers' code paths can be entered if the target user is banned (eg: you can't friend them, you can't DM them, etc).

Jamiras: Can you reply to a conversation with a banned user? Does unfriending a banned user send an email?

wescopeland: Does look like it's possible to reply to a banned user after all, but I've verified it doesn't send an email. At the moment, to the best of my knowledge, it isn't possible to unfriend someone who is banned.

Jamiras (approved):

PR #3353, opened 2025-03-27
fix(game): don't show banned users in the Compare Progress section

wescopeland:
This PR fixes a minor issue on game and event pages where followed banned users still appear in the "Compare Progress" section. With no way to unfollow these users, they are permanently fixed in the section.


resources/views/components/game/compare-progress.blade.php
```diff
@@ -48,7 +48,12 @@
         ->toArray();
 
     $userIds = array_column($followedUserCompletion, 'user_id');
-    $friends = User::whereIn('ID', $userIds)->get()->keyBy('ID');
+    $friends = User::whereIn('ID', $userIds)->whereNull('banned_at')->get()->keyBy('ID');
+
+    // Filter out completion data for banned users.
+    $followedUserCompletion = array_filter($followedUserCompletion, function ($item) use ($friends) {
+        return isset($friends[$item['user_id']]);
+    });
```
Jamiras, line 56: I get the feeling that at least one of these queries can be eliminated, but I don't think it's worth the effort right now.

The `whereNull('banned_at')` clause should probably be added to `$followedUsers` query as that feeds into `$followedUserCompletion`, which is used to build `$userIds` and `$friends`.

Jamiras: These queries also look like the ones in `BuildFollowedPlayerCompletionAction`. Maybe this should call that?

wescopeland: Don't think it's possible at the moment (without a lot of legwork) for the Blade component to call `BuildFollowedPlayerCompletionAction`. That action is specifically designed for the Inertia UI, as it's leveraging and returning laravel-data DTOs.

It is true that I built the action specifically to usurp the back-end logic this Blade component is doing for both the event page and game page. It probably won't be long until we drop this Blade component in favor of a React game page altogether.

Jamiras: > The `whereNull('banned_at')` clause should probably be added to `$followedUsers` query as that feeds into `$followedUserCompletion`, which is used to build `$userIds` and `$friends`.

This should also provide the benefit of not fetching the additional records in the intermediary queries.

wescopeland: Sorry, I missed that bit earlier. Addressed in latest.

Jamiras (approved):

PR #3364, opened 2025-03-27
fix(getUserProgress): return all game achievements

wescopeland:
https://discord.com/channels/310192285306454017/453242743292952578/1354951345219571965


wescopeland:
This is a regression from https://github.com/RetroAchievements/RAWeb/pull/3356.

PR #3365, opened 2025-03-28
fix(email): remediate wrap and line break issues

wescopeland:
This PR remediates two issues in Blade email templates:

* `<br />` tags are no longer printed into the email content.
* Sane line wrapping styles are now applied.

### DIFF
--- a/resources/views/vendor/mail/html/themes/default.css
+++ b/resources/views/vendor/mail/html/themes/default.css
@@ -270,6 +270,7 @@ img {
     background-color: #edf2f7;
     color: #718096;
     padding: 16px;
+    word-break: break-all;
 }
 
 .panel-content p {
--- a/resources/views/mail/community/activity.blade.php
+++ b/resources/views/mail/community/activity.blade.php
@@ -3,11 +3,10 @@
 @php
     $body = '';
     if (!empty($payload)) {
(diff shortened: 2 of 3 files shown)


resources/views/vendor/mail/html/themes/default.css
```diff
@@ -270,6 +270,7 @@ img {
     background-color: #edf2f7;
     color: #718096;
     padding: 16px;
+    word-break: break-all;
```
Jamiras, line 273: For some reason, the original test case from [here](https://retroachievements.org/viewtopic.php?t=29236&c=275728#275728) has some correct newlines on master, and some double `<br>`s:
image

Whereas, everything is single `<br/>`s (but properly newlined) when submitted on this branch:
image

Jamiras: But the wrapping does seem to be fixed.
image

Presumably this will prevent Outlook from trying to make the whole thing visible by shrinking it.

wescopeland: Think I have both line break cases fixed in latest:

Screenshot 2025-03-28 at 6 47 29 PM

Screenshot 2025-03-28 at 6 47 36 PM

Jamiras (approved):

PR #3366, opened 2025-03-28
perf(achievementInfo): resolve N+1 query problem

wescopeland:
Resolves https://retroachievementsorg.sentry.io/issues/35231528/?environment=production&query=is%3Aunresolved%20is%3Afor_review%20assigned_or_suggested%3A%5Bme%2C%20my_teams%2C%20none%5D&referrer=issue-stream&sort=inbox&stream_index=0.

Easily observable at a URL like http://localhost:64000/achievement/237679.

This drops my local query count on the page from 82 to 33.

### DIFF
--- a/resources/views/pages-legacy/achievementInfo.blade.php
+++ b/resources/views/pages-legacy/achievementInfo.blade.php
@@ -470,18 +470,31 @@ if ($game->system->id === System::Events) {
     if ($unlocks->isEmpty()) {
         echo "Nobody yet! Will you be the first?!<br>";
     } else {
+        $userWinners = $unlocks->pluck('User')->filter()->unique()->values()->toArray();
+        $usersMap = [];
+
+        if (!empty($userWinners)) {
+            $users = User::whereIn('User', $userWinners)->get();
+
+            foreach ($users as $user) {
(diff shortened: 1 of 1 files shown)


resources/views/pages-legacy/achievementInfo.blade.php
```diff
@@ -470,18 +470,31 @@ function ResetProgress() {
     if ($unlocks->isEmpty()) {
         echo "Nobody yet! Will you be the first?!<br>";
     } else {
+        $userWinners = $unlocks->pluck('User')->filter()->unique()->values()->toArray();
```
Jamiras, line 473: Why is `->filter()->unique()->values()` necessary? Can't you just `pluck('User')->toArray()`?

Jamiras (approved):

resources/views/pages-legacy/achievementInfo.blade.php
wescopeland: Taking a 2nd look at `getAchievementUnlocksData()`, I way overcomplicated this. We don't need this much defensiveness given the implementation details of that function. Simplified in latest to `pluck('User')->toArray()`.

PR #3368, opened 2025-03-29
fix(redirect): decode HTML entities in redirect URL

### LINKED ISSUE #3339: Auto generated resources URL in game forum dropping part of the search if there's an apostrophe
**Describe the bug**
If a game title has an apostrophe (and I imagine this will possibly include other similar characters) then when clicking into any of the links under the resources section of the OP, it drops the apostrophe in the google search along with everything after it.

**To Reproduce**
1: Go to any game that has an apostrophe in its title
2: Navigate to game forum page
3: Click any of the automatically generated links under the resources section of the initial post.
Example here: https://retroachievements.org/viewtopic.php?t=10674

Example of a game page where this does not happen: https://retroachievements.org/viewtopic.php?t=29692

**Expected behavior**
The google search results to display "gamefaqs", "wikipedia" etc followed by the full game name

**Additional context**
This looks to be caused by how the special characters are handled on site which google then drops if included in their search.

Danielopes7:
This PR resolves #3339 

Added html_entity_decode function to convert HTML entity quotes to their corresponding characters, preventing redirect issues


wescopeland:
Hi @Danielopes7 👋 

Thanks for taking this issue on!

Looking over the change, I can't help but wonder if this is treating a symptom, and not an underlying cause.

For example, I'm noticing it appears the back-end is automatically inserting two single apostrophes into these URLs. Additionally, if I change the URL to something like https://www.google.com/search?q=site:www.gamefaqs.com+King%27s+Field+II+PlayStation, it works correctly.

Based on this, I'm theorizing we may be inserting bad data into the DB which is causing this bug to surface.

Danielopes7:
> Hi @Danielopes7 👋
> 
> Thanks for taking this issue on!
> 
> Looking over the change, I can't help but wonder if this is treating a symptom, and not an underlying cause.
> 
> For example, I'm noticing it appears the back-end is automatically inserting two single apostrophes into these URLs. Additionally, if I change the URL to something like https://www.google.com/search?q=site:www.gamefaqs.com+King%27s+Field+II+PlayStation, it works correctly.
> 
> Based on this, I'm theorizing we may be inserting bad data into the DB which is causing this bug to surface.


I got it, btw I'm testing using the URL tag, is it the same thing using automatically links generate? otherwise, how can I simulate using the automatically generated links, I have tried to reproduce, but I don't know how genererate it.

it seems like URL tag is inserting properly on db.

wescopeland:
To generate these URLs, first ensure your account has a `DEVELOPER` role attached.

Then find a game with an apostrophe in the title that doesn't have a `ForumTopicID` set.

Then click this button its game page:
Screenshot 2025-04-03 at 7 23 28 AM

Danielopes7:
I found this line that replaces one apostrophe with two. This fixes the bad inserting on db. However, I still use the commit that I did before to fix redirect issues. Do you think it works? Or it's better save with (%27s) instead (') on db, that way I don't need use the redirect issues commit.

PR #3370, opened 2025-03-29
fix(vite): don't assume Valet configs will use SSL

wescopeland:
On my local, I'm giving Valet/Herd a try for an increased performance boost.

Vite assumes that if the local environment is running under a Valet configuration, SSL will always be enabled. This is not the case for me. Without these config changes, if I don't enable SSL, I can't spin up a dev server.

**How to test this PR:**
These steps assume you're using Sail.
* Ensure after running `sail pnpm build` you're still able to visit the home page.
* In another terminal instance, start `sail pnpm dev`. Visit the home page again, you should still see content (though will probably get a hydration error - that's normal).
* While `sail pnpm dev` is running, open _HomeRoot.tsx_ and comment a component out. You should see the hot reload occur in the browser in real-time.


Jamiras (approved):
I have horrible luck getting `pnpm run dev` to behave. Here's what my homepage looks like when using it:
image

I trust this does what you say it does.

wescopeland:
Any luck doing it through sail (`sail pnpm dev`)?

Usually this sort of thing is caused by a Vite misconfiguration. If it's still broken, would you mind sharing any console errors you may be seeing in the browser devtools? Wonder if it's something I can debug.

Jamiras:
`sail pnpm dev` fails with permissions errors.
```
Error: EACCES: permission denied, open '/var/www/html/node_modules/.vite-temp/vite.config.ts.timestamp-1743435508895-9db51b6e691ce.mjs'
```

I do see this in the console:
```
Port 64009 is in use, trying another one...
Port 64010 is in use, trying another one...
```
I assume the two are related. Sail is managing its own npm, so running it outside of sail causes port conflicts, but trying to run it within sail somehow causes permissions errors with the already running instance?

PR #3375, opened 2025-03-30
add claim expiration email notification

Jamiras:
Sends a claim notification email when less than 7 days are left before a claim expires. Sends another when less than 1 day is left before a claim expires.

Should hopefully remind users who aren't actively logging in that they need to as the expiration warning is currently only shown on the notifications widget when logged in.


app/Platform/Actions/SendClaimExpirationWarningEmailsAction.php
wescopeland, line 1: The only reservation I have with the current implementation details is how data is cached. Let's say for some reason Redis gets purged. I think that means we may wind up blasting all these emails out again. Also, I may be mistaken, but I believe as-implemented, the current Redis cache entry will grow infinitely in size, as it has no TTL and stuff gets added to it, but never removed.

Would it be possible to use a time-based approach instead, driven by `Finished` in the query?

eg:
```php
// untested pseudocode

// ... stuff ...
->where('Finished', '>=', Carbon::now()->addDays(6))
->where('Finished', '<', Carbon::now()->addDays(7))
// ... stuff ...

// ... stuff ...
->where('Finished', '>=', Carbon::now()->addHours(12))
->where('Finished', '<', Carbon::now()->addHours(24))
// ... stuff ...
```

wescopeland (commented):
Great idea. Just one piece of feedback to think through.

app/Platform/Actions/SendClaimExpirationWarningEmailsAction.php
Jamiras: The cache is rebuilt on every execution (`$newNotificationsSent` is written). Things not returned by the query (expirations more than a week old and extended claims) will fall out of the cache.

I don't have a good solution for if the cache is purged.

wescopeland (approved):

PR #3378, opened 2025-04-01
fix(user): don't count event points in the Recent Progress chart

wescopeland:
This PR resolves an issue where event points are being counted in the user profile Recent Progress chart.

**Before**
Screenshot 2025-03-31 at 9 26 53 PM

**After**
Screenshot 2025-03-31 at 9 26 47 PM


app/Helpers/database/player-history.php
```diff
@@ -102,6 +103,7 @@ function getAchievementsEarnedOnDay(int $unixTimestamp, User $user): array
 
 function getAwardedList(
     User $user,
+    bool $excludeEvents = true,
```
Jamiras, line 106: new optional parameters should be added at the end of the list to not break any other existing calls that rely on the existing optional parameters having specific indices.

it looks like this function is only called twice, and the other case isn't passing any optional parameters, so this is probably fine.

Jamiras (approved):

PR #3386, opened 2025-04-03
chore(deps): upgrade to carbon 3

wescopeland:
This PR upgrades the project from Carbon 2 to Carbon 3, which is a prerequisite for upgrading to Laravel 12.

I followed the upgrade guide here: https://carbon.nesbot.com/docs/#api-carbon-3

There should be no noticeable functional changes.


tests/Feature/Api/V1/AchievementSetClaimTest.php
```diff
@@ -168,7 +169,7 @@ public function testGetUserClaims(): void
                     'GameIcon' => '/Images/000001.png',
                     'GameTitle' => $game->Title,
                     'ID' => $claim->ID,
-                    'MinutesLeft' => Carbon::now()->diffInRealMinutes($claim->Finished),
+                    'MinutesLeft' => (int) $now->diffInMinutes($claim->Finished, true) + 1,
```
wescopeland, line 172: Carbon 3 seems to calculate some timestamps slightly differently. The values in this test suite all have a "drift" of 1 minute.

app/Models/User.php
```diff
@@ -402,7 +402,7 @@ public function isModerated(): bool
 
     public function isNew(): bool
     {
-        return Carbon::now()->diffInMonths($this->created_at) < 1;
+        return ((int) Carbon::now()->diffInMonths($this->created_at, true)) < 1;
```
Jamiras, line 405: It probably isn't necessary to convert to int for comparisons.

app/Community/Components/UserProfileMeta.php
```diff
@@ -477,7 +477,7 @@ private function calculateRecentPointsEarned(User $user, string $preferredMode =
 
         foreach ($achievements as $playerAchievement) {
             $achievementDate = $playerAchievement->{$dateColumn};
-            $daysAgo = abs($now->diffInDays($achievementDate, false));
+            $daysAgo = abs((int) $now->diffInDays($achievementDate, false));
```
Jamiras, line 480: If I'm reading the docs correctly, if the second parameter is `true`, an absolute value will be returned, so the `abs` can be eliminated by passing `true`.

app/Community/Components/UserProfileMeta.php
```diff
@@ -447,7 +447,7 @@ private function calculateAveragePointsPerWeek(User $user, bool $doesUserPreferH
             return 0;
         }
 
-        $weeksSinceFirstUnlock = $startingDate->diffInWeeks(Carbon::now());
+        $weeksSinceFirstUnlock = (int) $startingDate->diffInWeeks(Carbon::now(), true);
```
Jamiras, line 450: Why are you adding `true` to all of these calls? The [documentation](https://carbon.nesbot.com/docs/#api-difference) suggests that makes it an absolute value, which will cause the logic on line 453 to never be true.

> All can take 2 optional arguments: date to compare with (if missing, now is used instead), and an **absolute boolean option** (false by default), it returns negative value when the instance the method is called on is greater than the compared date (first argument or now). **If set to true, that makes the method return an absolute value no matter which date is greater than the other**.

tests/Feature/Api/V1/AchievementSetClaimTest.php
Jamiras: Is it a rounding thing? What is the float value being returned?

wescopeland: It does seem to be a rounding thing. Maybe it's due to the test execution time itself?
```
  -    'MinutesLeft' => 131039.98916603332,
  +    'MinutesLeft' => 131040,
```

app/Community/Components/UserProfileMeta.php
wescopeland: My perception is the migration guide implies that's what needed to maintain the same logical behavior in Carbon 3:

> In Carbon 3, using `(int) $after->diffInSeconds($before, true)` or `(int) abs($after->diffInSeconds($before))` allows to get explicitly an absolute and truncated value, so the same result as in v2.

tests/Feature/Api/V1/AchievementSetClaimTest.php
wescopeland: I've changed most of these in latest to use a pattern like:
```php
round(Carbon::now()->diffInMinutes($claim->Finished, true))
```
There was one that still required a `+ 1` in order to pass.

app/Models/User.php
wescopeland: Updated in latest.

app/Community/Components/UserProfileMeta.php
wescopeland: Updated in latest.

tests/Feature/Api/V1/AchievementSetClaimTest.php
Jamiras: Yea, it's a rounding error.

Since `$now` is not aligned to the start of a second, the value inserted into the database gets truncated.
```
$now = '2025-04-06 17:24:59.678713'
$dbval = '2025-04-06 17:24:59'
$dbExpireAt = $dbVal + 30 days
```
Then, when the API call is made, the value returned is calculated from the database values.
```
$dbElapsed = $dbExpireAt - $dbval = exactly 30 days
```
But the assertion is calculated using the raw `$now` value.
```
$expected = $dbExpireAt - $now = .678 seconds less than 30 days
```
Truncating `$expected` to an integer results in the off-by-one error.

It looks like we may have encountered something similar before: https://github.com/RetroAchievements/RAWeb/blob/f560b7bc2a04d17fbfad292243f9dc6170a2992c/tests/Feature/Community/Controllers/AchievementSetClaimControllerTest.php#L44-L45

app/Community/Components/UserProfileMeta.php
Jamiras: For an _absolute_ value, yes.
> In Carbon 3, using (int) $after->diffInSeconds($before, true) or (int) abs($after->diffInSeconds($before)) allows to get explicitly **an absolute** and truncated value, so the same result as in v2.

wescopeland: My interpretation of this doc is in Carbon 2 they're implied to be absolute. Local testing seems to support this.

For example, in the `User` model:
```php
return Carbon::now()->diffInMonths($this->created_at) < 1;
```

With Carbon 2 for user 255178, this returns 34.
With Carbon 3, the exact same code returns -34.853889037349.

tests/Feature/Api/V1/AchievementSetClaimTest.php
wescopeland: Thanks, using `->startOfSecond()` worked and enabled me to remove the `+ 1`'s and `round()` calls.

app/Community/Components/UserProfileMeta.php
Jamiras: Ah, I see now. So these migration conversions maintain parity, but the logic might be wrong in a few use cases.

app/Policies/TicketPolicy.php
```diff
@@ -35,7 +35,7 @@ public function view(User $user, Ticket $ticket): bool
 
     public function create(User $user): bool
     {
-        if ($user->created_at->diffInDays() < 1 || $user->is_muted || $user->banned_at) {
+        if ($user->created_at->diffInDays(now(), true) < 1 || $user->is_muted || $user->banned_at) {
```
Jamiras, line 38: Should this be `Carbon::now()` or is the function overloaded to accept a `time_t`?

Also applies to next two files.

tests/Feature/Platform/Services/PlayerGameActivityServiceTest.php
```diff
@@ -32,12 +32,12 @@ public function testActivity(): void
         $game = $this->seedGame(achievements: 5, withHash: false);
 
         // ==== Start session at time1 ====
-        $now = Carbon::now()->floorSecond();
-        $time1 = $now->clone()->subMinutes(100);
+        $now = Carbon::now()->floorSeconds();
+        $time1 = $now->clone()->subMinutes(100)->startOfSecond();
```
Jamiras, line 36: Recommend using `startOfSecond` when calculating now so you don't have to apply it to each derived value. Besides, isn't that what `floorSeconds` is supposed to be doing?

app/Policies/TicketPolicy.php
wescopeland: No harm in using `Carbon::now()`. I've also switched some of these to instead use `->diffInHours()`. Updated in latest.

tests/Feature/Platform/Services/PlayerGameActivityServiceTest.php
wescopeland: Fixed in latest.

Jamiras (approved):

PR #3388, opened 2025-04-03
feat(emulators): add supports_toolkit field and ticket creation warning

wescopeland:
**Post-deploy SQL:**
```sql
UPDATE `emulators`
SET `supports_toolkit` = 0
WHERE `id` IN (19, 20, 21, 22, 24, 25, 26, 27, 29, 30, 32, 33);
```

---

This PR adds a new nullable boolean column to the `emulators` table: `supports_toolkit`.

We can use this to drive various behaviors on the site, notably warning users who are about to create tickets on emulators which developers cannot easily debug from:

Screenshot 2025-04-03 at 7 36 28 PM

This data may also be useful in other contexts, such as the downloads page, but this ticketing context feels like the current immediate need.


app/Filament/Resources/EmulatorResource.php
```diff
@@ -116,6 +116,9 @@ public static function infolist(Infolist $infolist): Infolist
 
                         Infolists\Components\IconEntry::make('active')
                             ->boolean(),
+
+                        Infolists\Components\IconEntry::make('supports_toolkit')
+                            ->boolean(),
```
Jamiras, line 121: The initial value for this field is null, and appears as blank
image

app/Filament/Resources/EmulatorResource.php
```diff
@@ -209,6 +214,11 @@ public static function table(Table $table): Table
                     ->default(false)
                     ->alignCenter(),
 
+                Tables\Columns\IconColumn::make('supports_toolkit')
+                    ->boolean()
+                    ->default(false)
+                    ->alignCenter(),
+
```
Jamiras, line 221: Unrelated to this PR, but I can't edit the emulator records where I'd most likely want to set this because the download links aren't complete URLs.

image

app/Filament/Resources/EmulatorResource.php
wescopeland: Updated in latest to give the new field (in the migration) a default value of `true`. Additionally, I've added a SQL query in the PR description to set some of the emulators' values to false.

app/Filament/Resources/EmulatorResource.php
wescopeland: I think this current behavior is probably desirable. Absolute URLs are a bit safer for the view layer to handle, as they don't come with any assumptions about relative base paths.

Jamiras:
> UPDATE `emulators`
SET `supports_toolkit` = 0
WHERE `id` IN (19, 20, 21, 22, 24, 25, 26, 27, 29, 30, 32, 33);

You missed 12 (retroarch), 23 (melonds), and 28 (flycast). Or is the assumption that the libretro cores can be analyzed in RALibretro sufficient?

app/Filament/Resources/EmulatorResource.php
Jamiras: Maybe so, but we currently rely on that logic to determine whether to display an external or internal download link.
https://github.com/RetroAchievements/RAWeb/blob/f560b7bc2a04d17fbfad292243f9dc6170a2992c/resources/views/pages-legacy/download.blade.php#L49

image

Jamiras (approved):

Jamiras:
> What would you think of `are_triggers_debuggable` instead?

Better. Maybe `can_debug_triggers`?

wescopeland:
> Maybe `can_debug_triggers`?

I like that a lot. Renamed in latest.

PR #3392, opened 2025-04-04
fix(DiffTimestamp): suppress hydration errors

wescopeland:
DiffTimestamp is reporting hundreds of hydration errors to Sentry. However, the React team has specified these specific kinds of hydration errors are safe to suppress, as they are unavoidable: https://react.dev/reference/react-dom/client/hydrateRoot#suppressing-unavoidable-hydration-mismatch-errors

I've confirmed the hydration errors coming from this component do not cause the component tree to panic and remount.

### DIFF
--- a/resources/js/common/components/DiffTimestamp/DiffTimestamp.tsx
+++ b/resources/js/common/components/DiffTimestamp/DiffTimestamp.tsx
@@ -36,13 +36,19 @@ export const DiffTimestamp: FC<DiffTimestampProps> = ({
   }
 
   if (!enableTooltip) {
-    return <span className={className}>{diffForHumans(at, renderedAt)}</span>;
+    return (
+      <span className={className} suppressHydrationWarning>
+        {diffForHumans(at, renderedAt)}
+      </span>
+    );
   }
 
   return (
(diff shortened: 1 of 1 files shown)


resources/js/common/components/DiffTimestamp/DiffTimestamp.tsx
```diff
@@ -36,13 +36,19 @@ export const DiffTimestamp: FC<DiffTimestampProps> = ({
   }
 
   if (!enableTooltip) {
-    return <span className={className}>{diffForHumans(at, renderedAt)}</span>;
+    return (
+      <span className={className} suppressHydrationWarning>
```
Jamiras, line 40: The linked document says to use `suppressHydrationWarning={true}`. Is the `{true}` implied?

wescopeland: Yep -- React automagically passes boolean props without a given value as `true`.

Looking at this again though, I think I'd rather have it explicit than leave it to assumption. Updated in latest.

Jamiras (approved):

PR #3404, opened 2025-04-07
Support compatibility flag for game hashes

Jamiras:
SQL to run:
```
UPDATE game_hashes SET compatibility='compatible';
```

Implements the "Current Rule" portion of #717
image

image

image
(I would like to de-emphasize the incompatible hashes or put them inside a collapsed section)

-----

When the client asks for patchdata for an incompatible hash, an empty set is returned with the associated game icon and a warning achievement telling the user to view the Supported Game Files to find an appropriate version of the game to use.

When the client asks to identify an incompatible hash via the `gameid` API, a virtual ID is returned with a flag indicating compatibility. Attempting to fetch the patch data for that virtual ID will return the incompatible achievement set as described above.

image
image

image
image


wescopeland:
It seems like we need a SQL query to mark all the current hashes in the database as compatible, otherwise we end up with this:
Screenshot 2025-04-07 at 7 20 55 AM

Additionally, it may make sense when RAIntegration pushes a new hash to the database for it to default to "compatible".

I agree that everything which isn't in a "compatible" state should be suppressed within a collapse. Users are coming to this page to see what files to use. Showing files they can't use is dilutive to the page's UX. _TopicOptions.tsx_ may have a good example of an animated collapsible which could help us get started. I would put everything that isn't compatible all within a single collapse, maybe labeled "Other Files" (not 100% sure on this wording).

app/Platform/Controllers/GameHashController.php
```diff
@@ -27,10 +28,13 @@ public function index(Request $request, Game $game): InertiaResponse
         $this->authorize('viewAny', $this->resourceClass());
 
         $gameData = GameData::fromGame($game)->include('badgeUrl', 'forumTopicId', 'system');
-        $hashes = GameHashData::fromCollection($game->hashes);
+        $hashes = GameHashData::fromCollection($game->hashes->where('compatibility', GameHashCompatibility::Compatible));
+        $incompatibleHashes = GameHashData::fromCollection($game->hashes->where('compatibility', GameHashCompatibility::Incompatible));
+        $untestedHashes = GameHashData::fromCollection($game->hashes->where('compatibility', GameHashCompatibility::Untested));
+        $patchRequiredHashes = GameHashData::fromCollection($game->hashes->where('compatibility', GameHashCompatibility::PatchRequired));
```
wescopeland, line 34: May be good to think about using scopes for these if we find ourselves reaching for compatibility values often elsewhere.

resources/js/features/games/components/HashesMainRoot/HashesMainRoot.tsx
```diff
@@ -75,9 +96,78 @@ export const HashesMainRoot: FC = memo(() => {
             />
           </p>
 
-          <HashesList />
+          <HashesList hashes={hashes} />
+        </div>
+
+        {hasOtherHashes ? (
+          <BaseCollapsible open={isOpen} onOpenChange={setIsOpen}>
```
wescopeland, line 103: nit: It may be worth considering extracting this into a child component -- it's taking up quite a bit of the HashesMainRoot template.

wescopeland (approved):
Verified locally. I think the addition of `CreateWarningAchievementAction` is also a nice touch.

app/Platform/Controllers/GameHashController.php
Jamiras: I added a `compatibleHashes` property on game, and a `scopeCompatible` on `GameHash`. I couldn't figure out how to use the scope on the `game->hashes` collection.

resources/js/features/games/components/HashesMainRoot/HashesMainRoot.tsx
Jamiras: done

PR #3408, opened 2025-04-11
feat(downloads): prepare more on the back-end for React page rewrite

wescopeland:
This patchset is part of a much larger change I'm working on locally, but these changes are probably good enough they can stand on their own and be reviewed separately from the whole.

This PR:
* Adds a new `ra:cache-most-popular-emulators` Artisan command.
* Adds a new `ra:cache-most-popular-systems` Artisan command.
* Sets up an `EmulatorDownloadsRelationManager` that can be used from the emulator resource in Filament.
* Adds a migration to add a `platforms.order_column` column, as well as enforce `emulator_id,platform_id` uniqueness in `emulator_downloads`.
* Adds a `->has_official_support` attribute to the `Emulator` model.
* Prepares our DTOs to support props a new Inertia-powered Downloads page will be anticipating.

---

### The new Artisan commands
As mentioned above, two new commands have been added. These commands determine the most popular systems and most popular emulators (overall and by system) on the site.

These are ultimately used in the new UI (not included in this branch) like:
Screenshot 2025-04-11 at 5 24 31 PM
In the screenshot above, the most popular systems are immediately shown, with the others buried behind a click-through.

Users will be able to sort the list of displayed emulators either by popularity or in alphabetical order:
Screenshot 2025-04-11 at 5 25 56 PM

### emulator_downloads
The list of emulators will be cards that look something kind of like this:
Screenshot 2025-04-11 at 5 29 08 PM

The `emulator_downloads` table is used to override `emulators.download_url` on a platform-specific basis.


app/Models/Emulator.php
```diff
@@ -148,6 +149,33 @@ public function registerMediaCollections(): void
...
+            // version of the emulator out there which we fully support.
+            if ($userAgent->minimum_hardcore_version) {
+                return true;
+            }
+
+            // Case 2: Emulator is active and either doesn't have minimum version requirements
+            // or meets the minimum allowed version requirement.
+            if (
+                $this->active
+                && (!$userAgent->minimum_allowed_version || $userAgent->minimum_allowed_version)
```
Jamiras, line 170: Am I reading this correctly? && `(!A || A)`? Won't that always be true?

app/Providers/AppServiceProvider.php
```diff
@@ -69,9 +73,14 @@ public function boot(): void
...
         $this->app->booted(function () {
             $schedule = $this->app->make(Schedule::class);
+
             $schedule->command(LogUsersOnlineCount::class)->everyThirtyMinutes();
+
             $schedule->command(DeleteExpiredEmailVerificationTokens::class)->daily();
             $schedule->command(DeleteOverdueUserAccounts::class)->daily();
+
+            $schedule->command(CacheMostPopularEmulators::class)->weeklyOn(4, '8:00'); // Thursdays, ~3:00AM US Eastern
+            $schedule->command(CacheMostPopularSystems::class)->weeklyOn(4, '8:30'); // Thursdays, ~3:30AM US Eastern
```
Jamiras, line 83: Is it sane to only run the command weekly if the cache expires daily?
https://github.com/RetroAchievements/RAWeb/pull/3408/files#diff-2dd96ecf88d62ed5d40d46165162389c8c0cfaa9608e1a0f4243a8e7d38cc20dR59

wescopeland: Nope! Looks like I updated the other command, but forgot to update the one you pointed to. Updated in latest.

app/Models/Emulator.php
wescopeland: Good catch - I made a typo here. With a fresh set of eyes, I've changed up this logic in latest.

I used this code for local debugging:
```php
$emulators = Emulator::with("userAgents")->get();

$results = [];

foreach ($emulators as $emulator) {
  $results[] = [
    "id" => $emulator->id,
    "name" => $emulator->name,
    "has_official_support" => $emulator->has_official_support
  ];
}

dd($results);
```

PR #3410, opened 2025-04-13
fix(report-issue): remove generic Message QATeam button

wescopeland:
At the request of the team: https://discord.com/channels/476211979464343552/1360729249929101535/1360803113275097222

Only the en_US.json string needs to be removed - it'll naturally propagate through Crowdin to the other files.


wescopeland:
Temporarily closing -- looks like they may want to keep it after all.

PR #3412, opened 2025-04-13
fix: always soft delete comments in the PHP UI

wescopeland:
The React UI is soft deleting comments, but the PHP UI is hard deleting them. This PR unifies the soft delete behavior.


app/Helpers/database/user-activity.php
```diff
@@ -9,31 +9,16 @@
...
-
+    if (!$user->can('delete', $comment)) {
         return false;
     }
 
-    return mysqli_affected_rows($db) > 0;
+    $comment->timestamps = false;
+    $comment->delete();
+
+    return true;
```
Jamiras, line 21: Is there even a reason to keep this function any more? Can't this logic just be inlined in the other file?

Jamiras (approved):

app/Helpers/database/user-activity.php
wescopeland: Updated in latest.

PR #3413, opened 2025-04-13
feat(GameTypeFilter): make it a multi-select

wescopeland:
Desktop now has the ability to select multiple options using a logical OR:
Screenshot 2025-04-13 at 1 03 17 PM

Mobile behavior remains unchanged as a single-select (this filter doesn't get much use on mobile).

Addresses https://discord.com/channels/310192285306454017/1349706269388177438/1359138264740532317.


Jamiras (approved):
Possible future feature: some way to select/deselect all. I see this most likely used as an "exclude hacks" feature

PR #3429, opened 2025-04-19
denormalize playtime metrics

Jamiras:
Maintains a running **playtime** metric on each `player_games` record, and fully recalculates it, as well and **time to beat** and **time to master** metrics any time an achievement is unlocked. These are further aggregated into median metrics on the `GameData` table, which can be shown on the individual game pages:
image
The small number in parenthesis is the number of data points used to find the median.

This value is calculated for each player using the actual sessions tracked in the database starting with the datetime of the first published achievement. Times prior to Oct 2023 (when we started tracking sessions) are estimated from achievement unlocks, but are still included in the metrics. It is hoped that the estimated values are statistically similar to real values and/or will be diminished by having more real values than estimated ones. 

We may wish to only aggregate values newer than Oct 2023 when calculating the median metrics on the `GameData` table to eliminate the estimated values. We could also consider only aggregating the last year's worth of data to minimize the impact of revisions.

For many games, the softcore times don't differ that much from the hardcore times. I'm wondering if we want to just show the hardcore times and maybe eliminate the softcore times or move them to a tooltip?

Future PRs will use this data for the recently played games list and annual recap command. It could probably also be displayed in the player progress widget on the game page.

-----

To recalculate the metrics for a single game:
```
sail artisan ra:platform:game:update-player-games [gameid]
```

To recalculate all metrics (warning: this took around 40 hours on my VM to process all 18M records):
```
sail artisan ra:platform:player:update-estimated-times
sail artisan ra:platform:game:update-metrics
```
-----

For prod, I plan to just let this run naturally for a couple months to fill any `player_games` being actively played. Then, add a scheduled action that processes a handful of records every few minutes. If the count of unprocessed backfill is low enough (unlikely), then we could just use the commands.

Since data is going to be sparse (at least initially), the changes to the mastery widget are only displayed in local environments.


tests/Feature/Connect/SubmitLeaderboardEntryTest.php
```diff
@@ -60,9 +60,9 @@ public function testSubmitLeaderboardEntryEmptyLeaderboard(): void
         Carbon::setTestNow($now);
 
         /** @var Game $game */
-        $game = Game::factory()->create();
+        $game = $this->seedGame();
```
Jamiras, line 63: This ensures the game is associated to a valid console ID, which prevents null exceptions trying to read `$game->system->id` during the session management/player metrics processing.

database/migrations/2025_04_12_000000_update_gamedata_table.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('GameData', function (Blueprint $table) {
+            $table->dateTime('achievements_published_at')->nullable()->default(null)->after('RichPresencePatch');
```
Jamiras, line 13: This field could be used for https://discord.com/channels/310192285306454017/1362385885513126150/1362385885513126150, but doesn't help with revisions.

resources/views/components/game/top-achievers/index.blade.php
```diff
@@ -21,7 +21,20 @@
 @endif
 
 @if ($game->players_hardcore > 10)
-    <div class="text-right">
+    <div class="text-right mb-4">
         <a class="btn btn-link" href="{{ route('game.top-achievers.index', ['game' => $game]) }}">more...</a>
     </div>
 @endif
+
+@if (app()->environment('local'))
```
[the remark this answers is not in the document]
wescopeland: That makes sense. I'm going to drop a PR in a few minutes that stands up the first inklings of a (very incomplete) React game page. We can figure out over time how to get this data on there.

database/migrations/2025_04_12_000000_update_gamedata_table.php
wescopeland, line 1: I can't help but wonder what a lot of these fields look like when achievements are coming from sets rather than being tightly coupled to games themselves.

Does this imply subset games will stick around forever?

Jamiras: https://github.com/RetroAchievements/RAWeb/pull/3429#discussion_r2051471032

wescopeland: I wrote this before reading your comment here: https://github.com/RetroAchievements/RAWeb/pull/3429/files#r2051471032

PR #3431, opened 2025-04-19
feat(game): stand up a very basic React game page

wescopeland:
This PR creates a very basic React game page. There is **_a lot_** of functionality missing here, more than I can even remember to list (though I will try near the bottom of this PR description). But we have to start somewhere.

The new pages are simple to access. Rather than http://localhost:64000/game/668, navigate to http://localhost:64000/game2/668.

Screenshot 2025-04-19 at 9 51 40 AM

The page is designed with multiset in mind. Given that, achievements are pulled from game -> gameAchievementSets -> achievementSets -> achievements. Additionally, if a game has any subset that is not prefixed by "will_be_", that set will appear alongside the core set on the game page (as a proof of concept, this is not the final design):

Screenshot 2025-04-19 at 9 53 26 AM

A lot of components on the page are reused from event pages. There is a ton of functionality missing, and virtually everything on the page is going to need to be revisited in focused separate PRs. I think it's probably also safe to say nothing on the page in this PR should be considered "finalized" or a final design - this is merely intended as a starting point so we can reasonably progress forward from here with focused tasks and improvements.

Missing functionality:
* Achievement rows are missing a lot of metadata, such as points and type indicators.
* There's no way for developers to update RP, or access any of the functionality not migrated yet to Filament.
* Meta buttons are missing.
* Authorship/credit metadata is missing.
* There's no way to request a set.
* Missable and unlocked achievement filtering isn't present.
* The beaten games dialog is missing.
* User award displays (beaten / mastered) is missing.
* Subsets, similar games, latest masters, and leaderboards aren't visible.
* There's no recent players shown.
* Comments is missing.
* Want to Play / Want to Dev toggles are missing.

There's probably a lot more stuff I'm forgetting too. This will take quite a bit of iteration, and this PR feels as good of a place as any to "break ground".


app/Platform/Controllers/GameController.php
```diff
@@ -86,76 +82,23 @@ public function index(GameListRequest $request): InertiaResponse
...
-        return view($this->resourceName() . '.edit')->with('game', $game);
-    }
+        /** @var ?User $user */
+        $user = $request->user();
 
-    public function update(GameRequest $request, Game $game): RedirectResponse
-    {
-        $this->authorize('update', $game);
+        $game = $loadGameWithRelationsAction->execute($game, AchievementFlag::OfficialCore);
+        $props = $buildGameShowPagePropsAction->execute($game, $user);
```
Jamiras, line 97: Cannot view `/game2/` for a game without achievements.
http://localhost:64000/game2/26960
http://localhost:64000/game2/32412

Or maybe it's something else? I get the same error trying to view the `/game2/` for any of the "New Releases" games on the main page.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetHeader/GameAchievementSetHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+        {!isOnlySetForGame ? (
+          <LuChevronDown
+            data-testid="chevron"
+            className={cn(
+              'size-5 transition-transform duration-300',
+              isOpen ? 'rotate-180' : 'rotate-0',
+            )}
+          />
+        ) : null}
```
Jamiras, line 60: Changing the sort order while a set is collapsed causes it to refresh expanded, but with the button still in a collapse state.

app/Platform/Controllers/GameController.php
wescopeland: It seems `created_at` and `updated_at` for sets is nullable. This may be worth investigating in a future PR, for now the pages that are impacted won't crash anymore.

Jamiras (approved):
Approving this to unblock additional work. I understand there's stuff in here that we want to redo before making it available for consumers.

PR #3433, opened 2025-04-19
return adjusted virtualized ids for games in patch response

Jamiras:
the `gameid` response returns different virtual game IDs depending on how the hash is incompatible. When requesting `patch` data via a game hash, a singular virtual game ID is always returns. This modifies the `patch` response to also virtualize the game ID so the toolkit can detect which type of incompatibility was found.


app/Connect/Actions/IdentifyGameHashAction.php
wescopeland, line 1: This isn't a conventional action class anymore if it's exposing multiple public methods.

I believe `virtualizeGameId()` should be a new action, or `IdentifyGameHashAction` should be converted to a service.

Jamiras: I made it a service.

wescopeland (approved):

PR #3435, opened 2025-04-19
feat(messages): add support for selectable team inboxes

wescopeland:
This PR allows users to change their currently-viewed inbox on the messages index page. From there, they can read, compose, and reply to messages on behalf of that team.

### UX Walkthrough

A new button has been added for users that should have more than one available inbox:
Screenshot 2025-04-19 at 5 32 52 PM

This logic is driven by a new prop: `selectableInboxDisplayNames`. The prop itself is driven by `MessageThreadPolicy`:
```php
private function getAccessibleInboxes(User $user): array
{
    $policy = new MessageThreadPolicy();
    $accessibleTeamInboxes = $policy->getAccessibleTeamInboxes($user);

    return [$user->display_name, ...$accessibleTeamInboxes];
}
```
The list of accessible inboxes is matched against a given user's roles:
```php
/**
 * Role => display_name
 */
protected const ROLE_INBOX_MAP = [
    Role::ADMINISTRATOR => 'RAdmin',
    Role::DEV_COMPLIANCE => 'DevCompliance',
    Role::QUALITY_ASSURANCE => 'QATeam',
    Role::ARTIST => 'RAArtTeam',
    Role::CHEAT_INVESTIGATOR => 'RACheats',
    Role::WRITER => 'WritingTeam',
    Role::EVENT_MANAGER => 'RAEvents',
];
```

Notice in the screenshot above, the breadcrumb label is "Your Inbox". This is what the user sees when they click the Change Inbox button:
Screenshot 2025-04-19 at 5 35 23 PM

Upon confirming, the user is navigated to a slightly different route that might look something like http://localhost:64000/messages/RAdmin.
Screenshot 2025-04-19 at 5 36 00 PM

The user can now read all of RAdmin's messages, and any new threads or replies will be sent as RAdmin.

Note: there is no audit trail of who sent what on what team account's behalf. This is because functionally, such auditing doesn't actually exist today. For this to work properly, we'd need to lock down all the team account passwords, otherwise users could log in under the team account directly to circumvent audit logging.


app/Community/Controllers/Api/MessageApiController.php
```diff
@@ -30,41 +30,81 @@ public function store(MessageRequest $request): JsonResponse
...
             }
 
+            $senderUser = User::firstWhere('ID', $authorId);
+
             foreach ($thread->users as $threadUser) {
-                if (!$threadUser->is($user) && !$user->can('sendToRecipient', [Message::class, $threadUser])) {
+                if (!$threadUser->is($senderUser) && !$senderUser->can('sendToRecipient', [Message::class, $threadUser])) {
                     return response()->json([
-                        'error' => $user->isMuted() ? 'muted_user' : 'cannot_message_user',
+                        'error' => $senderUser->isMuted() ? 'muted_user' : 'cannot_message_user',
```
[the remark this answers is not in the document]
Jamiras: I also get this error trying to send to a generic user from my account, and trying to send to the team account from my account. Sending from the team account works fine to both myself and the generic user.

resources/js/features/messages/components/CreateMessageReplyForm/CreateMessageReplyForm.tsx
```diff
@@ -68,7 +71,9 @@ export const CreateMessageReplyForm: FC<CreateMessageReplyFormProps> = ({ onPrev
               </BaseButton>
 
               <BaseButton type="submit" disabled={!form.formState.isValid || mutation.isPending}>
-                {t('Submit')}
+                {auth!.user.displayName === senderUserDisplayName
+                  ? t('Submit')
+                  : t('Submit (as {{username}})', { username: senderUserDisplayName })}
```
Jamiras, line 76: Would it be worth tracking which user actually sent the message? (add a `sent_by_id` column to the `messages` table)

app/Policies/MessageThreadPolicy.php
```diff
@@ -6,33 +6,86 @@
...
+     */
+    protected const ROLE_INBOX_MAP = [
+        Role::ADMINISTRATOR => 'RAdmin',
+        Role::DEV_COMPLIANCE => 'DevCompliance',
+        Role::QUALITY_ASSURANCE => 'QATeam',
+        Role::ARTIST => 'RAArtTeam',
+        Role::CHEAT_INVESTIGATOR => 'RACheats',
+        Role::WRITER => 'WritingTeam',
+        Role::EVENT_MANAGER => 'RAEvents',
+    ];
```
Jamiras, line 28: What about DevQuest, QualityQuest, TheUnwanted, and/or RANews?
https://retroachievements.org/manage/roles/23
https://github.com/RetroAchievements/RAWeb/blob/32443e37dbbb718fa0bd398a40d33a27230de48b/config/services.php#L38-L85

resources/js/features/messages/components/+index/MessagesIndexRoot.tsx
Jamiras, line 1: Using the dropdown selector to change pages switches back to the non-team account

app/Policies/MessageThreadPolicy.php
wescopeland: These are tricky because (to my knowledge) we don't have any roles or permissions associated with them.

Short of creating new roles for these inboxes, I'm not sure how best to accommodate for this.

app/Community/Controllers/Api/MessageApiController.php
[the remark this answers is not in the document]
wescopeland: Resolved in latest.

resources/js/features/messages/components/CreateMessageReplyForm/CreateMessageReplyForm.tsx
wescopeland: Went ahead and added this field in latest.

resources/js/features/messages/components/+index/MessagesIndexRoot.tsx
wescopeland: Good catch, fixed in latest.

app/Community/Actions/AddToMessageThreadAction.php
```diff
@@ -12,11 +12,16 @@
...
+        MessageThread $thread,
+        User $userFrom,
+        User $trueSenderUser,
+        string $body
+    ): void {
         $message = new Message([
             'thread_id' => $thread->id,
-            'author_id' => $userFrom->ID,
+            'author_id' => $userFrom->id,
+            'sent_by_id' => $trueSenderUser->id,
```
Jamiras, line 24: Rather than require the upgrade step, we could assume that null means same as author (like we do for achievement unlocks).

app/Community/Controllers/Api/MessageApiController.php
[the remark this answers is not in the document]
Jamiras: Still seeing this.
image

app/Community/Actions/AddToMessageThreadAction.php
Jamiras: I think this be displayed when using the share account (if different than author_id).

app/Community/Controllers/Api/MessageApiController.php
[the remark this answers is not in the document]
Jamiras: I've tracked it down to this line: https://github.com/RetroAchievements/RAWeb/pull/3435/files#diff-d700ee43f90bc0657afd7f525009267f6d80d5dbab9cd927eddd7c1acb12ecb9R88

It's trying to verify that I have permission to use the team account "Jamiras", which doesn't appear in the team account mapping, so I don't have the necessary role.

Maybe there's some setup needed where I switched back and forth between available inboxes?

app/Community/Actions/AddToMessageThreadAction.php
wescopeland: Added in https://github.com/RetroAchievements/RAWeb/pull/3435/commits/f7c8b842aa6d9bf4f0db1dd9d425d9ae064965aa.

Screenshot 2025-04-27 at 9 42 58 AM

Special care has been taken to not accidentally leak this value to the other user. It is only added to props if the user is authorized to see it.

I've also tweaked the logic for writing `sent_by_id` per your suggestion.

app/Community/Controllers/Api/MessageApiController.php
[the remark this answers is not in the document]
wescopeland: Still having some trouble reproducing this, but I suspect I know why it would be happening. I've made a small adjustment in https://github.com/RetroAchievements/RAWeb/pull/3435/commits/153469aa4e459cfddb9df1673ed3938a44514a06. Give latest a try.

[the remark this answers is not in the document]
wescopeland: Also, definitely try a different recipient other than myself. I've only recently re-enabled my inbox after having it disabled for several months.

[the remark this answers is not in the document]
Jamiras: That does indeed allow me to send mails. And I was using a different recipient. I just didn't want some random user appearing in the screenshot.

Issue #3436, opened 2025-04-20
Hub issues with [Series - Congo Bongo]


tomojira:
**Describe the bug**
I noticed [this hub](https://retroachievements.org/hub/6588) seems to have 2 unrelated(?) problems.
1. It's sorted at the very bottom of [Central - Series](https://retroachievements.org/hub/3). So maybe there is something wrong with the sorting title - however, unlike game pages, there doesn't seem to be a way to view the sorting title. Note that this hub used to be Series - Nightmare Creatures before it was recycled (it had a forum topic for that linked to it, but I removed it).
2. The [management page](https://retroachievements.org/manage/hubs/6588) claims that it's an orphaned hub. However, it is indeed in Central - Series.

**Expected behavior**
That error shouldn't appear, and the hub should be sorted properly in the list. Perhaps the sorting title should be viewable for hubs as well.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 137.0.2

wescopeland:
Hi @tomojira!

I think the problem is the hub name wasn't encased in square brackets. I've made this adjustment and I believe the problems are now fixed.

tomojira:
Ohhh! Well that was simple, sorry about that.

wescopeland:
No problem! It's a quirk of the legacy hubs system that is retained in Hubs 2.0, and it's a non-obvious trap. 

I'm keeping this issue open for now, as ideally the management app should wrap those titles with square brackets automatically if they're missing until we're in a position to remove this quirk.

PR #3437, opened 2025-04-20
fix(HubResource): warn when title isn't wrapped with square brackets

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3436.

Screenshot 2025-04-20 at 7 45 03 AM

We'll get rid of these square brackets eventually, but for now, a warning is the 2nd best option. Having a hub that doesn't include the square brackets causes all kinds of bizarre side effects throughout the site.


app/Filament/Resources/HubResource.php
```diff
@@ -58,11 +59,21 @@ public static function infolist(Infolist $infolist): Infolist
...
+                                    return new HtmlString('<span style="color: #f59e0b;">Missing wrapping square brackets!</span>');
+                                }
+
+                                return null;
+                            }),
 
                         Infolists\Components\TextEntry::make('forumTopic.id')
                             ->label('Forum Topic ID')
-                            ->url(fn (?int $state) => route('forum-topic.show', ['topic' => $state]))
+                            ->url(fn (?int $state) => $state ? route('forum-topic.show', ['topic' => $state]) : null)
```
wescopeland, line 76: This fixes a crash related to the React forum topic cutover.

Jamiras (approved):
It took me a while to realize this warning is appearing on the details page. I was expecting it to be a validation error. Should it be a validation error?

wescopeland:
It's a fair question. My gut reaction was "yes" and I started coding the change, but then I remembered there are hubs in the DB that don't have this and probably shouldn't ("Available Hub XXXXX") which would fail validation if they were ever updated or new ones were created.

PR #3438, opened 2025-04-20
fix: remediate null topic id forum-topic.show route crashes

wescopeland:
Resolves a few cases where pages will crash because topic ID is nullable. This seems to only be an issue in Filament, as far as I can tell.


app/Filament/Resources/EventResource.php
```diff
@@ -70,7 +70,7 @@ public static function infolist(Infolist $infolist): Infolist
 
                         Infolists\Components\TextEntry::make('legacyGame.forumTopic.id')
                             ->label('Forum Topic ID')
-                            ->url(fn (?int $state) => route('forum-topic.show', ['topic' => $state]))
+                            ->url(fn (?int $state) => $state ? route('forum-topic.show', ['topic' => $state]) : null)
```
Jamiras, line 73: recommend adding `->placeholder('none')` to mimic behavior on hubs:
image

wescopeland: Updated both infolists in latest to use the "none" placeholder.

Jamiras (approved):

PR #3442, opened 2025-04-21
feat(AchievementsListItem): display points for game achievements

wescopeland:
Achievements in games should now display their point totals:

Screenshot 2025-04-21 at 11 28 13 AM


resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
```diff
@@ -71,19 +70,29 @@ export const AchievementsListItem: FC<AchievementsListItemProps> = ({
...
+                    }}
+                  />
+                ) : null}
+              </span>
+
+              {/* Points */}
+              {achievement.points && achievement.pointsWeighted ? (
+                <span className="inline-flex gap-1 text-[12px]">
+                  <span>{`(${achievement.points})`}</span>
+                  <WeightedPointsContainer>{`(${achievement.pointsWeighted})`}</WeightedPointsContainer>
```
Jamiras, line 94: Commas have been lost.

Old:
image

New:
image

resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
```diff
@@ -71,19 +70,29 @@ export const AchievementsListItem: FC<AchievementsListItemProps> = ({
...
+                    components={{
+                      1: <span />,
+                      2: <AchievementGameTitle game={game} />,
+                    }}
+                  />
+                ) : null}
+              </span>
+
+              {/* Points */}
+              {achievement.points && achievement.pointsWeighted ? (
```
Jamiras, line 91: Event achievements are never showing points. We should probably show points (but not pointsWeighted) for event achievements worth more than 1 point.

resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
wescopeland: Fixed in latest.

resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
wescopeland: In latest, I've extracted the points display logic to a new child component: `<AchievementPoints />`.

This should now be satisfied by the following conditional within the component:
```tsx
if (!points || (isEvent && points === 1)) {
  return null;
}
```

Jamiras (approved):

PR #3458, opened 2025-04-25
feat(forums): animate the spoiler button

wescopeland:
This PR:
* Extracts animated collapse orchestration into a reusable hook.
* Animates the `[spoiler]` shortcode collapse.

https://github.com/user-attachments/assets/928025f9-d26f-4a4b-82c6-02b615c8e98e

### DIFF
--- a/resources/js/common/components/ShortcodeRenderer/ShortcodeSpoiler/ShortcodeSpoiler.tsx
+++ b/resources/js/common/components/ShortcodeRenderer/ShortcodeSpoiler/ShortcodeSpoiler.tsx
@@ -1,5 +1,11 @@
-import type { FC, ReactNode } from 'react';
+import { AnimatePresence } from 'motion/react';
+import * as m from 'motion/react-m';
+import { type FC, type ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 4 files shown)


Jamiras (approved):
This is a good change, but I wonder if it can't be taken further. Could we somehow have a 
```
<CollapsibleComponent buttonText='Spoiler'>
  <div ref={contentRef} className="rounded-b-lg rounded-tr-lg bg-embed px-3 py-2">
    {stripLeadingWhitespaceFromChildren(children)}
  </div>
</CollapsibleComponent>
```
?

wescopeland:
I've tried a few iterations of this over a few working sessions, but I'm having trouble landing on a solution I'm happy with.

This is ultimately what I'd prefer:
```tsx
<BaseCollapsible animated={true}>
  <BaseCollapsibleTrigger>...</BaseCollapsibleTrigger>
  <BaseCollapsibleContent>...</BaseCollapsibleContent>
</BaseCollapsible>
```

This retains the flexibility of using the underlying Radix components as a base for composition. The problem is, the child components aren't aware of whether or not they're animated, which causes a handful of issues that I'm having trouble resolving.

I may chip away at this just a little bit more before throwing in the towel and merging what I have.

PR #3461, opened 2025-04-25
feat: redesign FullPaginator

wescopeland:
This PR makes some subtle adjustments to the FullPaginator component's design, with the intention of solving a minor UX annoyance the component currently has.

**Before**
Screenshot 2025-04-25 at 7 39 14 PM

**After**
Screenshot 2025-04-25 at 7 39 27 PM

Previously, the buttons would conditionally render based on the user's current page. Now all the buttons render regardless of the page, with buttons being conditionally enabled/disabled.

With this change, the paginator doesn't subtly shift its layout while the user iterates through pages.


Lamwaunga:
A suggestion: is it possible to somehow keep the number of total page visible at a glance?

It allows users to have a clear expectation before start navigating / clicking the drop down button. I'm not sure how to fit that in the new  design but removing it does mean less information compared to the current design.

wescopeland:
Sure. I've re-added that span element in latest with a subdued text color:

Screenshot 2025-04-26 at 8 37 14 AM

Jamiras (commented):
Shouldn't this have the same style/replace the paginator on the game page lists?
image

wescopeland:
The controls are doing wildly different things - one uses anchor tags with client-side routing, another uses buttons which kick off prefetches and scroll events. It'll be fewer headaches to keep them as separate components.

I've updated `<DataTablePagination />` to mirror the new styles and icons though:
Screenshot 2025-04-29 at 4 28 43 PM

Jamiras:
I see the new icons on the old pages, but the overall styles still differ.

Messages/Forums/Top Achievers page:
image

NES Games page:
image

wescopeland:
I've made this update in latest:

Screenshot 2025-04-30 at 5 19 04 PM

This is about as close as it's realistically going to get without needing to trigger a retranslation of these strings if we apply additional styles.

Jamiras (approved):
Thanks for working through this. I didn't mean to make it such a big deal, I was just pointing out the inconsistency between the two pages.

wescopeland:
No worries at all @Jamiras. I appreciate your feedback; it always pushes me to meet a higher quality bar.

PR #3464, opened 2025-04-26
fix(forums): generate topic meta descriptions on the back-end

wescopeland:
This PR shifts the responsibility of generating SEO/embed descriptions for forum topic pages from Node.js to Laravel. By doing this, we can leverage `Shortcode::stripAndClamp()` so our SEO descriptions aren't littered with raw shortcodes (especially for events).


Jamiras (approved):
> especially for events

I found it easier to test with official forum topics.

PR #3465, opened 2025-04-26
fix(Emulator): don't show RAGens, etc on the new downloads page

wescopeland:
This PR prevents a number of outdated emulators (eg: RAGens) from appearing on the new Downloads page.


app/Models/Emulator.php
```diff
@@ -161,7 +161,10 @@ public function getHasOfficialSupportAttribute(): bool
         foreach ($this->userAgents as $userAgent) {
             // Case 1: Has minimum_hardcore_version. This implies there's at least one
             // version of the emulator out there which we fully support.
-            if ($userAgent->minimum_hardcore_version || $userAgent->minimum_allowed_version) {
+            if (
+                ($userAgent->minimum_hardcore_version && $userAgent->minimum_hardcore_version < 99999)
+                || ($userAgent->minimum_allowed_version && $userAgent->minimum_allowed_version < 99999)
+            ) {
```
Jamiras, line 167: The `->active` flag is meant to drive this behavior. There's a lot of stuff not marked as active because we didn't want it showing up on the Downloads page. Since we do want it showing up on the new Downloads page, shouldn't this just look at `->active` and we can update `->active` after switching over?

wescopeland: Yep - fixed in latest.

Jamiras (approved):

PR #3467, opened 2025-04-27
Support virtual game IDs on editor APIs

Jamiras:
Addresses code notes not being visible for compatibility tester: https://discord.com/channels/310192285306454017/310195377993416714/1365529380654288956

I've taken this opportunity to convert two of the four affected APIs to use Command classes that I've added to the Connect namespace (instead of the code being split between the dorequest file and the old helper files). If this approach is acceptable, I'll start converting additional APIs in the future. If it's not, this PR can be greatly simplified.


public/dorequest.php
```diff
@@ -805,6 +787,10 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
             break;
         }
 
+        if (VirtualGameIdService::isVirtualGameId($gameID)) {
+            [$gameID, $compatibility] = VirtualGameIdService::decodeVirtualGameId($gameID);
+        }
```
Jamiras, line 792: This remaps the virtual ID back to the regular ID for the `uploadachievement` API. The same change has been applied to `codenotes2` and `submitcodenote` in the new classes.

app/Connect/Commands/ApiHandlerBase.php
```diff
@@ -0,0 +1,70 @@
...
+    public function execute(Request $request): JsonResponse
+    {
+        $result = $this->initialize($request);
+
+        if (!$result) {
+            $result = $this->process();
+        }
+
+        return $this->buildResponse($result);
+    }
```
Jamiras, line 25: A normal Connect API flow will be to call this `execute` method, which uses `initialize` to validate the parameters, `process` to do the work, and `buildResponse` to turn the output into a JSON response.

app/Connect/Commands/ApiHandlerBase.php
```diff
@@ -0,0 +1,70 @@
...
+
+    protected function gameNotFound(): array
+    {
+        return [
+            'Success' => false,
+            'Status' => 404,
+            'Code' => 'not_found',
+            'Error' => 'Unknown game',
+        ];
+    }
```
Jamiras, line 69: Helper function for common errors are provided.

app/Connect/Commands/AuthenticatedApiHandlerBase.php
```diff
@@ -0,0 +1,53 @@
...
+    public User $user;
+
+    public function execute(Request $request): JsonResponse
+    {
+        if (!$this->authenticate($request)) {
+            return $this->buildResponse($this->accessDenied());
+        }
+
+        return parent::execute($request);
+    }
```
Jamiras, line 22: An `AuthenticatedApiHandlerBase` extends the `ApiHandlerBase` to validate the user's credentials before doing anything else.

app/Connect/Commands/SubmitCodeNote.php
```diff
@@ -0,0 +1,85 @@
...
+use App\Models\Game;
+use App\Models\MemoryNote;
+use App\Platform\Services\VirtualGameIdService;
+use Illuminate\Http\Request;
+
+class SubmitCodeNote extends AuthenticatedApiHandlerBase
+{
+    public int $gameId;
+    public int $address;
+    public string $note;
```
Jamiras, line 16: Commands expose their inputs as public fields so they can be used without instantiating a request (see `update-code-note.php` below).

app/Helpers/database/code-note.php
```diff
@@ -30,60 +28,13 @@ function loadCodeNotes(int $gameId): ?array
...
-        return false;
-    }
-
-    $addressHex = '0x' . str_pad(dechex($address), 6, '0', STR_PAD_LEFT);
-    $currentNotes = getCodeNotesData($gameID);
-    $i = array_search($addressHex, array_column($currentNotes, 'Address'));
-
-    // TODO use Eloquent ORM to determine if the operation is an update, and
-    // if so, use MemoryNotePolicy::update() instead of a legacy Permissions check.
-    $permissions = (int) $user->getAttribute('Permissions');
```
Jamiras, line 62: The command uses the policy now.

public/dorequest.php
```diff
@@ -26,6 +28,16 @@
...
 
+$requestType = request()->input('r');
+$handler = match ($requestType) {
+    'codenotes2' => new GetCodeNotes(),
+    'submitcodenote' => new SubmitCodeNote(),
+    default => null,
+};
+if ($handler) {
+    return $handler->execute(request());
+}
```
Jamiras, line 39: Once everything is converted, default will return an error handler, and this file would end here.

public/request/game/update-code-note.php
```diff
@@ -14,14 +16,16 @@
...
-$address = $input['address'];
-$note = $input['note'] ?? "";
+$command = new SubmitCodeNote();
+$command->gameId = $input['gameId'];
+$command->address = $input['address'];
+$command->note = $input['note'] ?? '';
+$command->user = User::whereName($user)->first();
 
-$success = submitCodeNote2($user, $gameId, $address, $note);
+$result = $command->process();
```
Jamiras, line 25: This is an example of using the command outside of a request.

app/Connect/Commands/ApiHandlerBase.php
wescopeland: I like the pattern you've implemented a lot. The only thing I'm having trouble getting past is these aren't "commands" in a Laravel sense. In other words, they aren't Artisan commands.

These things behave more like action classes:
* They handle API requests, not CLI input.
* They don't extend `Command`.
* They're built to follow a single responsibility focused on one domain task of the Connect API.

With this in mind, I'd propose largely keeping the same pattern you have here, but:
* Moving both "Base" files to _app/Connect/Support_, similar to other base interfaces in the project.
* Renaming `ApiHandlerBase` to `BaseApiAction` (consistent with `BaseModel`).
* Renaming `AuthenticatedApiHandlerBase` to `BaseAuthenticatedApiAction`.
* Moving both code notes classes to _app/Connect/Actions_ and suffixing their file and class names with "Action" for consistency (eg: `GetCodeNotesAction`, _GetCodeNotesAction.php_).

Jamiras: I made the suggested changes. In trying to make it more action-like, I've rename `execute` to `handleRequest` and added a new `execute` method to the actual implementation class, that accepts parameters instead of relying on all the fields being initialized.

wescopeland (approved):

PR #3473, opened 2025-04-30
chore: cut over downloads to the React UI

wescopeland:
Deletes _download.blade.php_ and cuts over links to the new React UI.


Jamiras (approved):
I was a bit concerned testing this as RAGens was appearing in the list again. It took me a bit, but I finally realized this branch doesn't have that code yet.

While I was digging around, I noticed that `getHasOfficialSupportAttribute` can be greatly simplified now.

https://github.com/RetroAchievements/RAWeb/blob/31754a28ccdcc5a49b11abe85ffabcec25337347/app/Models/Emulator.php#L154-L176

Because `$this->active` must be true on line 170 (because of line 158), and the collection cannot be empty (also because of line 158), the foreach loop will always return true. The whole function can be collapsed to:
```
    public function getHasOfficialSupportAttribute(): bool
    {
        // Officially supported emulators must be set to active and
        // have at least one user agent registered in the DB.
        return ($this->active && !$this->userAgents->isEmpty());
    }
```

wescopeland:
> While I was digging around, I noticed that getHasOfficialSupportAttribute can be greatly simplified now.

Thanks - went ahead and made this change in latest.

Issue #2402, opened 2024-05-03
History Best Days graph sending to the wrong date


Mindhral:
**Describe the bug**
I'm guessing this happens when we live east to the UTC line.
On the history page, if I click on one of the bars in the "Best Days" graph, I get sent to the day before the one I clicked.

**To Reproduce**
1. Set your system timezone to UTC+1 (unless you're on another UTC+n)
2. Go to your [user history](https://retroachievements.org/history.php)
3. Click on any of the best days bars.
4. You get sent to the day before the one you clicked (as in before in the calendar, not before on the graph)

**Expected behavior**
Arriving on the page for the day we clicked

**Screenshots**

**Environment**
Checked on:
- Windows 10 + Firefox
- Android table + Firefox
- Android phone + Samsung Browser

**Additional context**
All the other ways to get to a date on the page work (for me).
The timestamp used in the URL correspond to a different hour of the day depending on the widget we use:
- "Go to date" button: 00:00:00
- "History" graph: 10:00:00 (12:00:00 for GMT)
- "Best days" graph: the day before 22:00:00 (midnight for GMT)
- "Best days" links: 00:00:00

I'm guessing using the same conversion for the graphs as for the links would give a more consistent behavior?

sethxd:
It looks like we need to update two functions in history.blade.php to address this:

```php
function selectHandlerBestDays(e) {
      if (chartBestDays.getSelection().length >= 1) {
        var dateAbbr = dataBestDays.getFormattedValue(chartBestDays.getSelection()[0].row, 0);
        var dateParsed = Date.parse(dateAbbr) / 1000;
        window.location = '/historyexamine.php?d=' + dateParsed + '&u=<?= $userPage ?>';
      }
    }
``` 

```php
function selectHandlerScoreProgress(e) {
      if (chartScoreProgress.getSelection().length >= 1) {
        var dateFormatted = dataTotalScore.getFormattedValue(chartScoreProgress.getSelection()[0].row, 0);

        var d = new Date(Date.parse(dateFormatted));
        var dAdj = new Date(d.getTime() + 60000 * 60 * 12);	// Adjusted by 60000 (min) times 60 (hour) times 12 (middle of day)

        var nUnix = parseInt(dAdj.getTime() / 1000);

        window.location = '/historyexamine.php?d=' + nUnix + '&u=' + <?php echo "'$userPage'"; ?>;
      }
    }
```

I can take a look once I finished getting my local environment up and running.

PR #3481, opened 2025-05-02
fix(history): fixed wrong time sent as url parameter

gilramot:
The 2 functions were written for the same purpose, but for some reason implemented differently. This results in unexpected behavior for different locales (the unix time might be off by 1 day).


wescopeland (commented):
Hi @gilramot! 👋 

This change works, although I think you've touched on why the current implementation is problematic:

> The 2 functions were written for the same purpose, but for some reason implemented differently. 

Now both functions are all but identical, and if one changes, the other will probably break again.

It seems like we could add an arg and trim this down to a single function.

gilramot:
Ok @wescopeland, fixed in #[152c274](https://github.com/RetroAchievements/RAWeb/pull/3481/commits/152c274b4e497ce131f8ff82461f3c0c1831e413)

gilramot:
Ok @wescopeland, fixed in [9ede3a3](https://github.com/RetroAchievements/RAWeb/pull/3481/commits/9ede3a3a5528ab024042c2c6fac5be511e7a1821).

wescopeland (approved):
Thanks for making that adjustment @gilramot. I'd be happy to accept this change as-is, however, if you'd like to go deeper, I think there's still an inconsistency on the page we should fix before I'd feel good about closing out #2402.

When I click on one of the bars in the "Best Days" chart, it sends me to a different places than if I click on the same row in the table below it.

For example, on my profile, when I click a bar for 13 Aug 2022, I get sent to:
http://localhost:64000/historyexamine.php?d=1660406400&u=WCopeland

However, when I click on the "Num Achievements" value for the same date in the table, I get sent to:
http://localhost:64000/historyexamine.php?d=1660348800&u=wesweswes

gilramot:
@wescopeland alright, now it should work.

gilramot:
@wescopeland alright, now it should work. Although now that I think about it, why does it have to be midday? This can create more unexplained behavior when trying to reproduce bugs (for example, when there is a >12h difference between local time and GMT, and a unix time is converted to date using local time). In your opinion, should the time part of the unix time parameter be set to 00:00 or 12:00?

wescopeland:
@gilramot In my opinion (which is loosely-held), it should be set to 00:00. I haven't verified, but I believe this would align with when the "Followed Users Ranking" on your own profile updates:
Screenshot 2025-05-04 at 7 18 11 AM

gilramot:
@wescopeland alright, changed in [e49fff1](https://github.com/RetroAchievements/RAWeb/pull/3481/commits/e49fff19983195a16a849b880e1eec36b6a10a35).

gilramot:
@wescopeland alright, changed in [7cfceec](https://github.com/RetroAchievements/RAWeb/pull/3481/commits/7cfceece3e778dd5b9c8a5b2c1eb3b99817edea1)

wescopeland:
I am having a lot of trouble following what is changing between deleted comments and force pushes.

I should be able to get a final review in later this evening.

gilramot:
Sorry about that, should be fine now

wescopeland (changes requested):
Clicking any bar on the graph freezes the UI and throws a console error.

PR #3486, opened 2025-05-03
fix(code-notes): don't show soft-deleted notes in the php ui

wescopeland:
Soft deleted code notes are appearing in the legacy PHP UI (http://localhost:64000/codenotes.php?g=1) due to the legacy SQL query not considering the `deleted_at` field.

### DIFF
--- a/app/Helpers/database/code-note.php
+++ b/app/Helpers/database/code-note.php
@@ -1,10 +1,14 @@
 <?php
 
+use App\Models\MemoryNote;
 use App\Models\User;
 
 function loadCodeNotes(int $gameId): array
 {
-    $codeNotes = App\Models\MemoryNote::with('user')
+    $codeNotes = MemoryNote::query()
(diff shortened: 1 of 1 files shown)


app/Helpers/database/code-note.php
```diff
@@ -2,30 +2,22 @@
...
-            $codeNotesOut[] = $db_entry;
-        }
-
-        return $codeNotesOut;
-    }
-
-    log_sql_fail();
-
-    return null;
+    $codeNotes = App\Models\MemoryNote::with('user')
```
Jamiras, line 7: I believe this needs the same `withTrashed` treatment added in #3479 (thanks for doing that btw)

wescopeland: No problem at all.

I believe you're right - updated in latest.

Jamiras (approved):

PR #3487, opened 2025-05-03
fix note deletion from client

Jamiras:
Fixes first part of https://discord.com/channels/310192285306454017/310195377993416714/1367961651156095067 (ability to delete code notes from DLL). I believe the second part (them magically reappearing when deleted from the site) is handled by #3486.

### DIFF
--- a/app/Connect/Actions/SubmitCodeNoteAction.php
+++ b/app/Connect/Actions/SubmitCodeNoteAction.php
@@ -29,7 +29,7 @@ class SubmitCodeNoteAction extends BaseAuthenticatedApiAction
 
     protected function initialize(Request $request): ?array
     {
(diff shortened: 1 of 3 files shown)


wescopeland (approved):
LGTM. 

The only edge case I can think of is what happens if two users are viewing notes for the same game, and User A deletes Note X a few minutes before User B deletes Note X. I _think_ in that circumstance, User B may get an unhandled exception, but it's probably not a big deal and unlikely to happen.

Jamiras:
`SubmitCodeNoteAction` would fetch the deleted note, verify that the user has permission to delete it, then mark it as deleted. As long as `$memoryNote->delete()` is re-entrant, there shouldn't be a problem.

PR #3490, opened 2025-05-04
fix: remove the `@routes` directive from blade views

wescopeland:
This PR removes `@routes` from the Blade UI, which should hopefully resolve our double-deploy conundrum.

Now, route definitions are injected into the JS bundles directly. They are no longer attached to the document head, which should clear up any race condition between regenerating them and caching views.

**How to test this PR:**
```
sail pnpm build
```
You should be able to land on the home page without it crashing, and you should get no hydration errors.

Additionally, all tests and `tsc` should continue passing (as will be verified by CI on this PR).


Jamiras (approved):
Don't see any obvious errors navigating around the site.

PR #3491, opened 2025-05-04
chore(search): begin setting up meilisearch

wescopeland:
This PR begins the process of setting up [Laravel Scout](https://laravel.com/docs/12.x/scout) / Meilisearch. The intent of this PR is to prove we can get the service deployed and a few indexes populated.

I've ripped out nearly all of the V2 search/Scout code. I think for this stuff it's probably a better idea to start from scratch, especially given Scout will ultimately integrate with Inertia, which was not originally planned in V2.

**How to test this PR**
Find the `SCOUT_DRIVER` portion of your _.env_ file, and update it with the new content from _.env.example_:
```env
SCOUT_DRIVER=meilisearch
SCOUT_PREFIX=
SCOUT_QUEUE=true
MEILISEARCH_NO_ANALYTICS=true
MEILISEARCH_HOST=http://meilisearch:7700
MEILISEARCH_KEY=c0f9L4BrnVg9CoWXzy70olfWgrsct888
```

Only users and games are searchable. Background jobs+queues are used to populate the search indexes, so you'll need a Horizon process running.
```bash
sail up # you'll see some brand new meilisearch stuff appear
sail artisan horizon
```
```bash
sail artisan scout:import "App\Models\Game"
sail artisan scout:import "App\Models\User"
```
Those two commands will kick off a bunch of jobs to populate the search indexes. The jobs shouldn't take very long to run.

Once the jobs are completed, a new demo page is available:
http://localhost:64000/demo/scout

The code for the page is very simple, and you can try different searches by changing the queries:
```php
@use('App\Models\Game')
@use('App\Models\User')

@php
    $games = Game::search('poke')->get();
    $users = User::search('scott')->get();
@endphp

<x-app-layout>
    <div className="flex flex-col gap-6">
        <p>Games</p>
        @dump($games->toArray())

        <p class="mt-8">Users</p>
        @dump($users->toArray())
    </div>
</x-app-layout>
```

Weighted searches for users is outside the scope of this PR. Additionally, `User::scopeSearch` has been renamed to `User::scopeSqlSearch` and marked as deprecated.

Before merging, I will set up the Meilisearch process on stage. Forge provides a simple toggle switch to enable it.


app/Models/Game.php
```diff
@@ -292,17 +292,10 @@ public function registerMediaCollections(): void
...
-        ]);
-    }
-
-    public function shouldBeSearchable(): bool
-    {
-        // TODO return $this->isPublished();
-        // TODO return true;
-        return false;
+        return [
+            'id' => (int) $this->ID,
```
Jamiras, line 296: Is this given precedence because it's first? If I search for "2025", I get like 10 matches where 2025 is part of the ID before I get any of the 2025 events.

app/Models/User.php
```diff
@@ -338,6 +309,25 @@ public function getActivitylogOptions(): LogOptions
...
     }
 
+    // == search
+
+    public function toSearchableArray(): array
+    {
+        return [
+            'id' => (int) $this->ID,
+            'display_name' => $this->display_name,
+        ];
```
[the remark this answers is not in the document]
wescopeland: I'm waffling back and forth on this, but I've added it in latest (under `username`, the proper name of the column per the model's TODOs).

However, I've also removed "id", as it probably shouldn't be searchable.

You'll need to reindex the `User` model with these commands:
```
sail artisan scout:flush "App\Models\User"
sail artisan scout:import "App\Models\User"
```

app/Models/Game.php
```diff
@@ -292,17 +292,10 @@ public function registerMediaCollections(): void
...
+            'title' => $this->title,
+        ];
```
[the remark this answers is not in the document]
wescopeland: I've added `Game::shouldBeSearchable` in latest. Docs reference [here](https://laravel.com/docs/11.x/scout#conditionally-searchable-model-instances).

My sense is "hub games" and "event games" should not be searchable, and "subset games" _temporarily_ should be. There needs to be a cutover period after React game pages where subset games are made non-searchable and non-core game achievement sets are searchable instead.

You'll need to reindex the `Game` model with these commands:
```
sail artisan scout:flush "App\Models\Game"
sail artisan scout:import "App\Models\Game"
```

app/Models/Game.php
wescopeland: It is not given precedence because it's first. Anything returned by `toSearchableArray()` is a searchable value. Docs [here](https://laravel.com/docs/11.x/scout#configuring-searchable-data).

Ranking rules for a model are set up in the _config/scout.php_ file. [Docs reference](https://laravel.com/docs/12.x/scout#configuring-filterable-data-for-meilisearch). [Meilisearch settings reference](https://www.meilisearch.com/docs/reference/api/settings).

I believe `searchableAttributes` is the setting we're after, based on this text from the Meilisearch settings docs:
> Fields in which to search for matching query words sorted by order of importance

I've added it in latest with `title` prioritized over `id`.

wescopeland:
@Jamiras Is there anything else I can address on this PR?

app/Console/Commands/SyncUsers.php
Jamiras, line 1: I have no idea if this is just reporting things weirdly, but the "me" user is getting indexed every few seconds, even if I'm not actively doing anything.
image

app/Models/Game.php
Jamiras: Searching on "2000" is still preferring game IDs:
image
image

app/Models/User.php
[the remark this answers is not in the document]
Jamiras: The flush/import is still matching IDs.
image

Jamiras (commented):
I'm not sure if I screwed something up, but I'm only seeing half the changes.
* User is now searchable
* Events are no longer returned
I'm not seeing these changes:
* ID is still searchable for users
* ID is still preferred over titles for games

app/Console/Commands/SyncUsers.php
wescopeland: Confirmed this was a bug caused by us constantly updating `LastLogin` on the user value.

In latest, I've updated that line to:
```php
$user->saveQuietly(); // don't trigger search reindexing
```

app/Models/User.php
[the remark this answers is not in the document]
wescopeland: After a ton of digging, I managed to figure this out. I didn't realize that when an index setting is changing, the settings have to be synced, even if the index is completely flushed.

Try running:
```
```

Now I see username "123" as the first result.

app/Models/Game.php
wescopeland: https://github.com/RetroAchievements/RAWeb/pull/3491#discussion_r2083270336

After running the same command referenced in this other comment, I now see:
Screenshot 2025-05-10 at 3 00 01 PM

app/Models/User.php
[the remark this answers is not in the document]
Jamiras: Was it supposed to sync `UserAccounts` too?
```
jamiras@ubuntu24 pr/3491:~/source/RAWeb$ sail artisan scout:sync-index-settings
Settings for the [GameData] index synced successfully.
jamiras@ubuntu24 pr/3491:~/source/RAWeb$
```

app/Models/Game.php
Jamiras: This appears to be working, but I'm still having issues with users. I don't think the sync command actually did anything.

app/Console/Commands/SyncUsers.php
Jamiras: This appears to still be happening. Do I need to do something more than just `sail down`/`sail up`?

app/Models/User.php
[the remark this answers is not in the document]
wescopeland: Yea, it should be syncing them both:
```
❯ sail artisan scout:sync-index-settings
Settings for the [GameData] index synced successfully.
Settings for the [UserAccounts] index synced successfully.
```

I wonder if restarting Horizon and then running `sail artisan scout:delete-all-indexes` works.

Followed by doing that, you'll need to import the two models again:
```
sail artisan scout:import "App\Models\User"
sail artisan scout:import "App\Models\Game"
```

app/Console/Commands/SyncUsers.php
wescopeland: Restart Horizon - it holds all the previous app code in memory unless it's manually reset.

app/Models/User.php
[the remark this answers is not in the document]
Jamiras: I fully restarted my VM and it's still not seeing UserAccounts:
```
jamiras@ubuntu24 pr/3491:~/source/RAWeb$ sail artisan scout:delete-all-indexes
All indexes deleted successfully.
jamiras@ubuntu24 pr/3491:~/source/RAWeb$ sail artisan scout:sync-index-settings
Settings for the [GameData] index synced successfully.
jamiras@ubuntu24 pr/3491:~/source/RAWeb$
```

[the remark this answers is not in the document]
wescopeland: Just as a sanity check, can you confirm in _config/scout.php_ you see these lines (specifically the `User` config stuff):
```php
'meilisearch' => [
    'host' => env('MEILISEARCH_HOST', 'http://localhost:7700'),
    'key' => env('MEILISEARCH_KEY'),
    'index-settings' => [
        Game::class => [
            'filterableAttributes' => ['id', 'title'],
            'searchableAttributes' => ['title', 'id'],
            'sortableAttributes' => ['id', 'title'],
        ],

        User::class => [
            'searchableAttributes' => ['display_name', 'username'],
            'sortableAttributes' => ['display_name', 'username'],
            'filterableAttributes' => ['display_name', 'username'],
        ],
    ],
],
```

Jamiras (approved):

PR #3493, opened 2025-05-06
fix(SubscribeToggleButton): update mobile styles

wescopeland:
Resolves an issue on mobile where the pagination label is truncating on the forum.

The subscribe button now has a bell icon. On mobile, only the bell icon is visible.

**Before**
Screenshot 2025-05-05 at 8 18 47 PM

**After**
Screenshot 2025-05-05 at 8 17 36 PM


resources/js/common/components/SubscribeToggleButton/SubscribeToggleButton.tsx
```diff
@@ -49,8 +54,10 @@ export const SubscribeToggleButton: FC<SubscribeToggleButtonProps> = ({
   };
 
   return (
-    <BaseButton size="sm" onClick={handleClick}>
-      {isSubscribed ? t('Unsubscribe') : t('Subscribe')}
+    <BaseButton size="sm" onClick={handleClick} className="gap-1.5" aria-label={label}>
+      {isSubscribed ? <LuBellOff className="size-4" /> : <LuBell className="size-4" />}
+
+      <span className="hidden sm:block">{label}</span>
```
Jamiras, line 60: The dropdown is still truncated if there's a sufficient number of pages.
image

Jamiras (approved):

resources/js/common/components/SubscribeToggleButton/SubscribeToggleButton.tsx
wescopeland: Good catch. Should be fixed in latest.

PR #3496, opened 2025-05-07
denormalize playtime metrics

Jamiras:
Reimplements #3429, leveraging the `player_achievement_sets` table.

-----

Maintains a running **playtime** metric on each `player_games` record, and fully recalculates it, as well and **time to beat** and **time to master** metrics any time an achievement is unlocked. These are further aggregated into median metrics on the `GameData` table, which can be shown on the individual game pages:
image
The small number in parenthesis is the number of data points used to find the median.

This value is calculated for each player using the actual sessions tracked in the database starting with the datetime of the first published achievement. Times prior to Oct 2023 (when we started tracking sessions) are estimated from achievement unlocks, but are still included in the metrics. It is hoped that the estimated values are statistically similar to real values and/or will be diminished by having more real values than estimated ones. 

We may wish to only aggregate values newer than Oct 2023 when calculating the median metrics on the `GameData` table to eliminate the estimated values. We could also consider only aggregating the last year's worth of data to minimize the impact of revisions.

For many games, the softcore times don't differ that much from the hardcore times. I'm wondering if we want to just show the hardcore times and maybe eliminate the softcore times or move them to a tooltip?

Future PRs will use this data for the recently played games list and annual recap command. It could probably also be displayed in the player progress widget on the game page.

-----

To recalculate the metrics for a single game:
```
sail artisan ra:platform:game:update-player-games [gameid]
```

To recalculate all metrics (I was only about 20% through this after 40+ hours):
```
sail artisan ra:platform:player:update-estimated-times
sail artisan ra:platform:game:update-metrics
```
-----

For prod, I plan to just let this run naturally for a couple months to fill any `player_games` being actively played. Then, add a scheduled action that processes a handful of records every few minutes. If the count of unprocessed backfill is low enough (unlikely), then we could just use the commands.

Since data is going to be sparse (at least initially), the changes to the mastery widget are only displayed in local environments.


database/migrations/2025_05_05_000003_update_gamedata.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::table('GameData', function (Blueprint $table) {
+            $table->integer('times_beaten')->default(0)->after('players_hardcore');
+            $table->integer('times_beaten_hardcore')->default(0)->after('times_beaten');
+            $table->integer('median_time_to_beat')->nullable()->default(null)->after('times_beaten_hardcore');
+            $table->integer('median_time_to_beat_hardcore')->nullable()->default(null)->after('median_time_to_beat');
```
Jamiras, line 16: These are the beaten aggregated metrics that would be displayed on the game page.

image

database/seeders/Concerns/SeedsUsers.php
```diff
@@ -25,23 +23,7 @@ protected function seedUserByUsername(string $username = 'verified', array $attr
...
-                'EmailAddress' => config('mail.from.address'),
-                'email_verified_at' => Carbon::now(),
-                'APIKey' => $username . '-secret',
-                'Motto' => 'I am ' . $username,
-                'RAPoints' => 1000,
-                'Password' => Hash::make($safeUsername),
-                'TrueRAPoints' => 1010,
-                'User' => $safeUsername,
-                'display_name' => $safeUsername,
-            ], $attributes)
```
Jamiras, line 42: These fields are populated by the factory.

tests/Feature/Api/V1/UserRecentlyPlayedGamesTest.php
```diff
@@ -40,35 +39,40 @@ public function testGetUserRecentlyPlayedGamesUnknownUser(): void
...
 
-    public function testGetUserRecentlyPlayedGames(): void
+    public function testGetUserRecentlyPlayedGamesX(): void
     {
-        /** @var System $system */
-        $system = System::factory()->create();
-        /** @var Game $game */
-        $game = Game::factory()->create([
-            'ConsoleID' => $system->ID,
+        $game = $this->seedGame(achievements: 3);
```
Jamiras, line 43: `seedGame` is required for anything that regenerates metrics as the core achievement set also has to be populated.

app/Platform/Commands/UpdatePlayerEstimatedTimes.php
```diff
@@ -0,0 +1,101 @@
...
+                foreach ($playerGames as $playerGame) {
+                    $game = $playerGame->game;
+                    if (System::isGameSystem($game->system->id)) {
+                        if (mb_strpos($game->title, "[Subset")) {
+                            // do a full execution against the parent game
+                            $parentGame = getParentGameFromGameTitle($game->title, $game->system->id);
+                            if ($parentGame !== null) {
+                                $parentPlayerGame = PlayerGame::where('game_id', $parentGame->id)->where('user_id', $playerGame->user->id)->first();
+                                if ($parentPlayerGame) {
+                                    (new UpdatePlayerGameMetricsAction())->execute($parentPlayerGame, silent: true);
```
wescopeland, line 41: This makes me nervous. On my local, I need to execute the action roughly 17,000,000 times before the command can complete.

I whipped up a POC of batched jobs for this on my local machine which has numerous advantages:
* Parallel processing - multiple jobs can run simultaneously, giving us better throughput.
* The queue is tightly provisioned, so there's limited risk of pegging prod's CPU at 100%.
* Any individual failed job won't stop the entire process, and the queue can be paused/resumed.
* We have direct visibility into the queue progress on Horizon itself.

With this change, I could reach 10% completion in roughly 5 minutes:
Screenshot 2025-05-07 at 6 05 58 PM

I have a rough POC on this branch in my fork here: https://github.com/wescopeland/RAWeb/commit/151c808171687f964f89712eff7e25180c306c02

At the time of this writing, I haven't finished reviewing or even checked `sail artisan ra:platform:game:update-metrics` yet. This comment just felt important enough to send off while I'm mid-review.

Jamiras: If we're going to do this, wouldn't we want to just queue up `UpdatePlayerGameMetricsJob` jobs?

(untested)
```
            $jobs = $chunk->map(function ($playerGame) {
                return new UpdatePlayerGameMetricsJob($playerGame->user->id, $playerGame->game->id);
            })->all();
```

That would avoid the partial code duplication I was trying to use to minimize the execution time.

Jamiras: As noted elsewhere, the expectation is not to run this command for some period of time to not overload the server as every active session will immediately try to rebuild its playtime metrics.

> For prod, I plan to just let this run naturally for a couple months to fill any player_games being actively played. Then, add a scheduled action that processes a handful of records every few minutes. If the count of unprocessed backfill is low enough (unlikely), then we could just use the commands.

wescopeland: Understood - I misinterpreted the "let this run naturally for a couple months". I'll mark this feedback as resolved.

app/Platform/Actions/UpdateGameMetricsAction.php
wescopeland, line 1: Having some trouble here, not sure what to do.

I'm unable to invoke this action manually without it crashing:
```php
$model = Game::find(1);
$action = new UpdateGameMetricsAction();
$action->execute($model);
```
```
  Allowed memory size of 134217728 bytes exhausted (tried to allocate 48250880 bytes)
```

When I queue up `sail artisan ra:platform:player:update-estimated-times`, the app freezes and the connection to my local DB is terminated:
```
mariadb-1       | 2025-05-09 20:50:27 141 [Warning] Aborted connection 141 to db: 'retroachievements-web' user: 'retroachievements' host: '192.168.97.7' (Got an error writing communication packets)
mariadb-1       | 2025-05-09 20:50:34 142 [Warning] Aborted connection 142 to db: 'retroachievements-web' user: 'retroachievements' host: '192.168.97.7' (Got an error writing communication packets)
```
Screenshot 2025-05-09 at 4 56 31 PM

wescopeland: I can confirm the problem is coming from `UpdateGameMetricsAction::getMedian()`.

This tweak fixes the issue on my local machine, but may need an extra pair of eyes to confirm it doesn't introduce a logic error:
```php
if (($count % 2) == 1) {
    // odd. just get the middle item
    $median = $query->select($field)->orderBy($field)
        ->offset((int) ($count / 2))->take(1)->value($field);
} else {
    // even. get the two items in the middle and average them together
    $values = $query->select($field)->orderBy($field)
        ->offset((int) ($count / 2) - 1)->take(2)->pluck($field)->toArray();
    $median = count($values) === 2 ? ($values[0] + $values[1]) / 2 : null;
}
```
Key changes:
* `->select($field)` keeps us from loading entire model objects.
* `->value($field)` keeps up from needing to load/hydrate full models.
* I added an extra `count($values) === 2`, but this may not actually be needed.

Jamiras: > When I queue up sail artisan ra:platform:player:update-estimated-times, the app freezes and the connection to my local DB is terminated

The conversion to use the queue overloads my machine too if I let it queue up more than a million or so jobs. I think redis is running out of space. I have to reboot my VM after it locks up.

The command isn't meant to be run now, and could probably be eliminated entirely. It was more of a for-debugging thing I introduced in the original PR that has mutated several times into its current state. It's much more prudent to just run the single game command so you can test that single game.

I have a feeling that we'll want a scheduled job that slowly backfills the unprocessed records rather than a single command that tries to backfill all unprocessed records at once.

wescopeland: I'm confused. What I'm saying is I can't invoke `UpdateGameMetricsAction` anymore as-is under any circumstance due to its high memory usage. On this branch, all I have to do to cause things to break is demote an achievement from a popular game.
```
  2025-05-10 13:14:42 App\Platform\Listeners\DispatchUpdateGameMetricsJob  RUNNING
  2025-05-10 13:14:42 App\Platform\Listeners\DispatchUpdateGameMetricsJob  46.49ms DONE
  2025-05-10 13:14:42 App\Platform\Listeners\DispatchUpdateDeveloperContributionYieldJob  RUNNING
  2025-05-10 13:14:42 App\Platform\Listeners\DispatchUpdateDeveloperContributionYieldJob  10.32ms DONE
  2025-05-10 13:14:42 App\Platform\Listeners\UpdateTotalGamesCount ... RUNNING
  2025-05-10 13:14:42 App\Platform\Listeners\UpdateTotalGamesCount  21.21ms DONE
  2025-05-10 13:14:42 App\Mail\CommunityActivityMail ................. RUNNING
  2025-05-10 13:14:42 App\Mail\CommunityActivityMail ........... 121.85ms DONE
  2025-05-10 13:14:42 App\Mail\CommunityActivityMail ................. RUNNING
  2025-05-10 13:14:42 App\Mail\CommunityActivityMail ............ 46.59ms DONE
  2025-05-10 13:14:45 App\Platform\Jobs\UpdateGameMetricsJob ......... RUNNING
  2025-05-10 13:14:45 App\Platform\Jobs\UpdateDeveloperContributionYieldJob  RUNNING

   Symfony\Component\ErrorHandler\Error\FatalError

  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:414
    410▕             $this->bindValues($statement, $this->prepareBindings($bindings));
    411▕
    412▕             $statement->execute();
    413▕
  ➜ 414▕             return $statement->fetchAll();
    415▕         });
    416▕     }
    417▕
    418▕     /**


   Whoops\Exception\ErrorException

  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:414
    410▕             $this->bindValues($statement, $this->prepareBindings($bindings));
    411▕
    412▕             $statement->execute();
    413▕
  ➜ 414▕             return $statement->fetchAll();
    415▕         });
    416▕     }
    417▕
    418▕     /**

      +1 vendor frames

  2   [internal]:0
      Whoops\Run::handleShutdown()

PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php on line 414
  2025-05-10 13:14:46 App\Platform\Jobs\UpdateDeveloperContributionYieldJob  1s DONE
```

wescopeland: What is your `memory_limit` set to?
```php
echo ini_get("memory_limit");
```
I get "128M".

Jamiras: I also get "128M".

wescopeland: I think something may be different between how the Horizon instance manages memory and how the Laravel server instance itself manages memory.

For example, I can do:
```
sail artisan ra:platform:game:update-player-games 1
```
and this runs several thousand jobs before I finally see:
```
  2025-05-10 14:42:44 App\Platform\Jobs\UpdateGameMetricsJob ......... RUNNING

   Symfony\Component\ErrorHandler\Error\FatalError

  Allowed memory size of 134217728 bytes exhausted (tried to allocate 73416704 bytes)
``` 

However, if I go to `HomeController::index()` and at the top of the method add:
```php
$model = Game::find(1);
$action = new UpdateGameMetricsAction();
$action->execute($model);
```
When I visit the home page, it immediately crashes:
```
mariadb-1       | 2025-05-10 14:42:11 1603 [Warning] Aborted connection 1603 to db: 'retroachievements-web' user: 'retroachievements' host: '192.168.97.7' (Got an error writing communication packets)
laravel.test-1  |   PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 83902464 bytes) in /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php on line 412
```

Jamiras: Seems to be the `first()` clause. Changing the above to this:
```
$median = $query->pluck($field)->toArray()[0];
```
makes it work. As does the `value` change you proposed:
```
$median = $query->value($field);
```

database/migrations/2025_05_05_000001_update_achievement_sets.php
```diff
@@ -0,0 +1,44 @@
...
+        Schema::table('achievement_sets', function (Blueprint $table) {
+            $table->dropForeign(['user_id']);
+            $table->dropIndex(['user_id']);
+            $table->dropColumn('user_id');
+
+            $table->dateTime('achievements_published_at')->nullable()->default(null)->after('players_hardcore');
+            $table->integer('times_completed')->default(0)->after('points_weighted');
+            $table->integer('times_completed_hardcore')->default(0)->after('times_completed');
+            $table->integer('median_time_to_complete')->nullable()->default(null)->after('times_completed_hardcore');
+            $table->integer('median_time_to_complete_hardcore')->nullable()->default(null)->after('median_time_to_complete');
```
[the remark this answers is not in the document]
wescopeland: I'll brainstorm ways we can work this stuff into the React game pages.

For user-facing language, we may want to use the word "Average" (even though this nomenclature functionally represents a mean, it's more likely to be understood).

app/Models/AchievementSet.php
```diff
@@ -37,6 +37,7 @@ class AchievementSet extends BaseModel
         'user_id',
         'players_total',
         'players_hardcore',
+        'achievements_published_at',
```
wescopeland, line 40: nit: Recommend renaming this column to `achievements_first_published_at`. Initially, I erroneously assumed this would update every time an achievement was published to the set.

app/Platform/Actions/ComputeAchievementsSetPublishedAtAction.php
wescopeland, line 1: At some point, it may be desirable to also query the activitylog rather than automated comments, but I'm only noting this here as a future-facing concern (not relevant to today's implementation details of the PR).

wescopeland (approved):
No more concerns from my end. I ran several million jobs and spot-checked the results in the DB and the game page sidebar component. The results seem ok.

I'm still having trouble running a full cycle of all the jobs locally (would like to do a full cycle just to see what happens), but I think that's more of a Laravel Sail issue than anything with the implementation details of the PR.

app/Models/AchievementSet.php
Jamiras: Done. You'll want to manually rename the column in your database. It's much faster than doing a `migrate:rollback`/`migrate`, and you won't lose any data you may have already aggregated.

app/Platform/Actions/ComputeAchievementsSetPublishedAtAction.php
Jamiras: I briefly looked into this. This is the query that returns the appropriate data and it seems fairly performant:
```
SELECT created_at FROM audit_log WHERE subject_type='achievement' 
AND subject_id IN (SELECT ID FROM Achievements WHERE GameID=? AND Flags=3)
AND properties LIKE '%"Flags":3%"Flags":5%' ORDER BY created_at LIMIT 1;
```
But `audit_log` isn't exposed through Eloquent yet, and I'm not sure how to do the `subject_type` mapping.

We might want to get this in sooner rather than later, as promotions made on the management pages don't generate audit comments, so the date would be erroneously calculated based on the first unlock or claim completion (which would be close, but not exact).

Jamiras: That didn't help much, but pointed out that the model for the audit log is `Activity`. Using that, I found the source under the vendor directory, and determined it was using `getMorphClass` to map the `subject_type` and I was able to construct an Eloquent query that seems to work.

PR #3504, opened 2025-05-10
fix: make text highlighting more legible

wescopeland:
This PR adjusts the text highlight styles for dark, light, and system schemes. Only the default palette is affected.

**Before**
Screenshot 2025-05-10 at 5 12 52 PM

**After**
Screenshot 2025-05-10 at 5 12 15 PM


Jamiras (approved):
Is this browser specific?

Before (me):
image

After (me):
image

wescopeland:
I didn't consider it originally, but it's OS-specific. Probably best to make it consistent across all OSes given our default text color.

PR #3507, opened 2025-05-11
feat(search): swap sqlSearch with scout search for users

wescopeland:
🔴 Should be deployed separately from #3491.

---

This PR drops `User::scopeSqlSearch()` in favor of `User::search()` powered by Meilisearch. The weighting isn't exactly how it was before, but it's probably good enough.

**How to test this PR:**
```
sail artisan horizon
sail artisan scout:import "App\Models\User"
```

Now navigate to http://localhost:64000/messages/create and start searching for users.

Screenshot 2025-05-11 at 5 03 36 PM

Performance before: ~389ms
Performance after: ~135ms


app/Models/User.php
```diff
@@ -314,7 +314,10 @@ public function getActivitylogOptions(): LogOptions
     public function toSearchableArray(): array
     {
         return [
+            'banned_at' => $this->banned_at,
+            'deleted_at' => $this->Deleted,
```
Jamiras, line 318: Wouldn't it make more sense to have these drive the `shouldBeSearchable` property?

wescopeland: Yep, it would. Fixed in latest. To test, after pulling latest from the branch, start Horizon, and then:
```
sail artisan scout:flush "App\Models\User"
sail artisan scout:import "App\Models\User" 
```

PR #3509, opened 2025-05-11
perf(game): use Cache::flexible() for game recent players

wescopeland:
Similar to https://github.com/RetroAchievements/RAWeb/pull/3508, mitigates another slow query on some game pages.

Recent Players can be an expensive & high-volume fetch in some places, with an execution time of ~100ms. It is rare for up-to-the-moment data to be needed in this component, especially given the limited frequency of RP pings from clients. 

This PR puts the DB call behind a stale-while-revalidate (SWR) cache:
```php
$freshTtl = 3 * 60; // 3 minutes
$staleTtl = 7 * 60; // 7 minutes
$recentPlayerData = Cache::flexible(
    "game-recent-players:{$gameID}:10",
    [$freshTtl, $staleTtl],
    function () use ($gameID) {
        return getGameRecentPlayers($gameID, 10);
    }
);
```

Data is considered fresh for 3 minutes. During this time, the cached value will always be returned without calling the callback function. After the 3 minutes fresh TTL expires, the data is considered "stale", but will still be usable for up to 7 minutes total. When a new request comes in while data is stale, the stale data is immediately returned to the user, but in the background `getGameRecentPlayers()` is executed to refresh the cache.

With the combined changes of this PR and #3508, most game pages load in ~20ms or less on my local machine.


wescopeland:
@Jamiras

> Just spitballing here, but wouldn't it be better to cache the results for some ridiculous period of time (24 hours?) and have the rich presence ping expire the cache?
>
> That way, a game that's not being played can use the cached data for an extended period of time, and a game that is actively being played will always be up-to-date?

I think latest should strike a good balance between this idea and my original approach.

After pulling latest, be sure to spin up a Horizon instance to fully test the changes, as the changes now touch `ResumePlayerSessionAction`.

I've removed `Cache::flexible()` in latest. The new caching strategy in this branch tries to mitigate cache misses altogether.

In _gameInfo.blade.php_, we now have a `Cache::remember()` call with a 1 week TTL:
```php
/**
 * Cache data for up to 1 week.
 * The cache is automatically invalidated when a player 
 * starts the game or pings the server with an ongoing session.
 */
$recentPlayerData = Cache::remember(
    "game-recent-players:{$gameID}:10",
    7 * 24 * 60 * 60, // 1 week
    function () use ($gameID) {
        return getGameRecentPlayers($gameID, 10);
    }
);
```

However, to complement this, in `ResumePlayerSessionAction`, we're busting and regenerating the cache on every call via `ResumePlayerSessionAction::regenerateRecentPlayersCache()`. This has virtually no overhead when done in a queued job and effectively shaves off that 80-100ms of load time on popular game pages.

app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -118,8 +122,36 @@ public function execute(
...
+     * When a player starts/pings for a game session, we:
+     * 1. Remove the existing cached data.
+     * 2. Pre-warm the cache with fresh data that includes the current player.
+     *
+     * This ensures the game page shows current data without causing a
+     * cache miss for the next visitor.
+     */
+    private function regenerateRecentPlayersCache(int $gameId): void
+    {
+        $cacheKey = "game-recent-players:{$gameId}:10";
```
Jamiras, line 145: Consider putting this in `CacheKey` to ensure the same key is used in both places.

It would also invert the key to `game:{$gameId}:recent-players:10`.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -929,7 +929,19 @@ function resize() {
...
+             * The cache is automatically invalidated when a player 
+             * starts the game or pings the server with an ongoing session.
+             */
+            $recentPlayerData = Cache::remember(
+                "game-recent-players:{$gameID}:10",
+                7 * 24 * 60 * 60, // 1 week
+                function () use ($gameID) {
+                    return getGameRecentPlayers($gameID, 10);
+                }
+            );
```
Jamiras, line 943: There's a lot of duplicated stuff here. Would it make more sense to add the Cache layer inside `getGameRecentPlayers`?
```
function getGameRecentPlayers(int $gameID, int $maximum_results = 10): array
{
    if ($maximum_results === 10) {
        $cached = Cache::get($key);
        if ($cached) {
           return $cached;
        }
    }

    ...

    if ($maximum_results === 10) {
        Cache::put($key, 1week, $retval);
    }
    return $retval;
}
```

app/Platform/Actions/ResumePlayerSessionAction.php
wescopeland: Done in latest.

resources/views/pages-legacy/gameInfo.blade.php
wescopeland: I considered this as well, but I'm not a fan of this approach. I think the downside of mixing data fetching and caching concerns within the single function outweighs the upside of reducing the duplication of cache key generation and TTL.

With this approach, it may be non-obvious to any consumer of the function that if `$maximum_results` is 10, we may never actually be fetching data from the database. Conversely, if a consumer assumes stuff is going to be cache read, it may be non-obvious if the limit is changed to something like 5 that it'll now be a DB read 100% of the time.

Additionally, the current implementation provides explicit cache invalidation in `ResumePlayerSessionAction`, which gives precise control over when the cache should be refreshed based on actual player activity.

Jamiras (approved):

PR #3513, opened 2025-05-14
feat(multiset): add `?r=achievementsets`

wescopeland:
Usage:
```
http://localhost:64000/dorequest.php?u=WCopeland&t=...&r=achievementsets&m=1b1d9ac862c387367e904036114c4825
```

Accepts either an `m` or a `g` param. `m` is favored if both are given. If only `g` is given, the returned set type is always "core" (multiset cannot function without a given hash).


Jamiras:
> InjectPatchClientSupportLevelDataAction is probably broken if the v2 code path is taken
Yep. I passed in an outdated client and got this:
```
{"Success":false,"Error":"Undefined array key \"PatchData\"","Status":500}
```

app/Connect/Actions/BuildClientPatchDataV2Action.php
Jamiras, line 1: If I request data for a game with no achievements, the `Sets` array comes back empty when querying by hash
```
{"Success":true,"GameId":14134,"Title":"Zeliard","ImageIconUrl":"http:\/\/media.retroachievements.org\/Images\/071254.png","RichPresenceGameId":14134,"RichPresencePatch":null,"ConsoleId":47,"Sets":[]}
```

But non-empty when querying by ID:
```
{"Success":true,"GameId":14134,"Title":"Zeliard","ImageIconUrl":"http:\/\/media.retroachievements.org\/Images\/071254.png","RichPresenceGameId":14134,"RichPresencePatch":null,"ConsoleId":47,"Sets":[{"Title":null,"Type":"core","AchievementSetId":21845,"GameId":14134,"ImageIconUrl":"http:\/\/media.retroachievements.org\/Images\/071254.png","Achievements":[],"Leaderboards":[]}]}
```

I would expect it to be non-empty in both scenarios.

wescopeland:
As of latest, I've made the following changes:
* The `v` param has been removed. Instead, the routine can be called by `?r=achievementsets`.
* An empty set should always be returned regardless of whether `m` or `g` are given.
* `InjectPatchClientSupportLevelDataAction` should no longer explode, and now supports the new data format.

app/Connect/Actions/BuildClientPatchDataV2Action.php
```diff
@@ -0,0 +1,359 @@
...
+            ->first();
+        $achievementSetId = $coreAchievementSet ? $coreAchievementSet->achievement_set_id : 0;
+
+        return [
+            'Success' => true,
+            'GameId' => VirtualGameIdService::encodeVirtualGameId($game->id, $gameHashCompatibility),
+            'Title' => 'Unsupported Game Version',
+            'ImageIconUrl' => media_asset($game->ImageIcon),
+            'ConsoleId' => $game->ConsoleID,
+            'Sets' => [
```
Jamiras, line 310: The needs to be another level deeper (there's no array in the response):
```
..."Sets":{"Title":null,...
```

Should be
```
..."Sets":[{"Title":null,...
```

wescopeland: Made this change in latest.

app/Connect/Actions/BuildClientPatchDataV2Action.php
```diff
@@ -0,0 +1,361 @@
...
+                            title: 'Unsupported Game Version',
+                            description: match ($gameHashCompatibility) {
+                                GameHashCompatibility::Incompatible => "This version of the game is known to not work with the defined achievements. $seeSupportedGameFiles",
+                                GameHashCompatibility::Untested => "This version of the game has not been tested to see if it works with the defined achievements. $seeSupportedGameFiles",
+                                GameHashCompatibility::PatchRequired => "This version of the game requires a patch to support achievements. $seeSupportedGameFiles",
+                                default => $seeSupportedGameFiles,
+                            }),
+                    ],
+                ],
+                'Leaderboards' => [],
```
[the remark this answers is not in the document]
wescopeland: Fixed in latest.

app/Connect/Actions/BuildClientPatchDataV2Action.php
wescopeland: Not sure I'm following, this is what I see when I make an HTTP request:
Screenshot 2025-05-17 at 9 28 24 AM

Jamiras: Moving `leaderboards` fixed this. You can put it back if you want to try to reproduce.

I think that's the last of the obvious things I've encountered. You should be able to write the unit tests now.

wescopeland: Ah, that makes a lot of sense. Great, I'll get started on writing tests later this afternoon.

wescopeland:
Coverage added in latest. Moved the PR out of draft.

Jamiras (approved):

PR #3519, opened 2025-05-15
feat(api): cap API_GetClaims to 1000 results

wescopeland:
API_GetClaims was implemented without pagination, and is now returning over a 4MB response (and growing) by default in prod.

There is an existing comment that reads:
> returns information about 1000 max set claims

This PR enforces this cap, which reduces the payload size to roughly 390KB.

### DIFF
--- a/public/API/API_GetClaims.php
+++ b/public/API/API_GetClaims.php
@@ -2,12 +2,9 @@
 
 /*
  *  API_GetClaims - returns information about 1000 max set claims, sorted by latest `Created` date.
- *    o : offset - number of entries to skip (default: 0)
- *    c : count - number of entries to return (default: 100, max: 500)
  *    k : claim kind - 1 for completed, 2 for dropped, 3 for expired (default: 1)
- *  int        Count                number of claims returned in the response
- *  int        Total                number of claims that actually exist
- *  array      Results
+ *
+ *  array
  *   object     [value]
  *    int        ID                 unique ID of the claim
  *    string     User               user who made the claim
@@ -32,7 +29,6 @@
  */
 
 use App\Community\Enums\ClaimFilters;
-use App\Models\AchievementSetClaim;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Validation\Rule;
@@ -42,8 +38,6 @@ $input = Validator::validate(Arr::wrap(request()->query()), [
         'nullable',
         Rule::in(['1', '2', '3']),
     ],
-    'o' => 'nullable|integer|min:0',
-    'c' => 'nullable|integer|min:1|max:500',
 ], [
     'k.in' => 'k must be set to one of the following values: :values',
 ]);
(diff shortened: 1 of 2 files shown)


public/API/API_GetClaims.php
```diff
@@ -1,14 +1,13 @@
 <?php
 
-use App\Community\Enums\ClaimFilters;
-use Illuminate\Support\Facades\Validator;
-use Illuminate\Validation\Rule;
-
 /*
  *  API_GetClaims - returns information about 1000 max set claims, sorted by latest `Created` date.
+ *    o : offset - number of entries to skip (default: 0)
+ *    c : count - number of entries to return (default: 100, max: 500)
```
Jamiras, line 6: The comment on line 4 says this returns a maximum of 1000 claims (which I don't think it was actually enforcing)

public/API/API_GetClaims.php
```diff
@@ -1,14 +1,13 @@
...
  *    k : claim kind - 1 for completed, 2 for dropped, 3 for expired (default: 1)
- *
- *  array
+ *  int        Count                number of claims returned in the response
+ *  int        Total                number of claims that actually exist
+ *  array      Results
```
Jamiras, line 10: I realize this is necessary to return the additional fields, but it's a breaking change. Make sure it's well communicated in the release notes (possibly even beforehand).

public/API/API_GetClaims.php
wescopeland: I didn't even see this comment.

Given that it's already there, maybe a happy medium is to not support pagination, but just cap the response correctly to 1000 results.

I've gone in this direction in latest. Now the response shape is exactly the same, but ~10x smaller in size.

public/API/API_GetClaims.php
```diff
@@ -54,7 +55,7 @@
     $claimFilter = ClaimFilters::AllActiveClaims;
 }
 
-$claimsResults = getFilteredClaims(claimFilter: $claimFilter);
+$claimsResults = getFilteredClaims(claimFilter: $claimFilter)->take(1000);
```
Jamiras, line 58: Suggest passing `limit=1000` instead of fetching all records and then trimming here.

wescopeland: Updated in latest.

public/API/API_GetClaims.php
```diff
@@ -32,6 +28,11 @@
  *    int        MinutesLeft        time in minutes left until the claim expires
  */
 
+use App\Community\Enums\ClaimFilters;
+use Illuminate\Support\Arr;
```
Jamiras, line 32: I don't think this is needed.

Jamiras (approved):

public/API/API_GetClaims.php
wescopeland: If it's removed, my IDE starts throwing a red squiggly underneath `Arr`:
Screenshot 2025-05-21 at 6 13 25 PM

PR #3522, opened 2025-05-17
feat(game): add a new hub-derived metadata component to the sidebar

wescopeland:
This PR makes a significant change to how game metadata is displayed on the React game pages. Rather than metadata being split between chips in the main content area and hubs on the sidebar, the two are combined in a new sidebar component.

Screenshot 2025-05-17 at 6 00 36 PM

Clicking "see more" causes even more hub-derived metadata to display:
Screenshot 2025-05-17 at 6 01 11 PM

If a piece of hub-derived metadata appears in this list, then that hub will no longer appear in the "Hubs" component. For example, game 1454 (The Legend of Zelda, NES), now only displays:
Screenshot 2025-05-17 at 6 02 04 PM

I plan on doing a special treatment for the series hub in a future PR. It is the most important hub linked to game pages.

Here's what the totality of 1454's page appears as above the fold on desktop:
Screenshot 2025-05-17 at 6 03 10 PM

Additionally:
* Meta team hubs (ie: the noncompliant writing hub) only appear to users who can access `/manage/games`.
* Where the chips in the main area were is now completely devoid of content. This is intentional. I plan on filling this in during a subsequent PR.
* I tested all of the top 50 most popular game pages. Given that hubs are largely unstructured data, we'll probably always need to have a "Hubs" component in the sidebar, but now it should be significantly more limited in content.
* This design is inspired by PSNProfiles, which uses a similar design for their sidebar game metadata, visible [here](https://psnprofiles.com/trophies/29085-astro-bot).


resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,279 @@
...
+    () =>
+      buildMetaRowElements({
+        hubs: allGameHubs,
+        hubTitleIncludes: ['Developer -', 'Hacker -'],
+        primaryLabel: 'Developer',
+        altLabels: ['Hacker'],
+        fallbackValue: game.developer,
+      }),
+    [allGameHubs, game.developer],
+  );
```
Jamiras, line 20: Hacker isn't being stripped from the raw data:
image

image

resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,279 @@
...
+        hubTitleIncludes: [
+          'Meta -',
+          'Rollout Sets -',
+          'AotW -',
+          'RANews -',
+          'RA Awards -',
+          'Custom Awards -',
+          'Console Wars',
+          'DevJam',
+          'Challenge League',
```
Jamiras, line 164: While the "Meta - " hubs do seem to be "RA Features", the rest of these aren't. It feels weird to see "AotW" as a feature:
image

Maybe you could split most of these off into a "Featured in" row?

resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,279 @@
...
+      buildMetaRowElements({
+        hubs: allGameHubs,
+        hubTitleIncludes: [
+          'Meta -',
+          'Rollout Sets -',
+          'AotW -',
+          'RANews -',
+          'RA Awards -',
+          'Custom Awards -',
+          'Console Wars',
```
Jamiras, line 162: This doesn't seem to be matching "Console Wars I": http://localhost:64000/game2/24227

image

resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,279 @@
...
+    () =>
+      buildMetaRowElements({
+        hubs: allGameHubs,
+        hubTitleIncludes: ['Genre -', 'Subgenre -'],
+        primaryLabel: 'Genre',
+        altLabels: ['Subgenre'],
+        fallbackValue: game.genre,
+      }),
+    [allGameHubs, game.genre],
+  );
```
Jamiras, line 44: Would it make sense to group subgenres under their genre?
image

would become:
```
Genre:  Action (2D Platforming)
```

Of course, this would only apply when both are linked. I know there's a lot of games that only link the subgenre hub.

resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,279 @@
...
+          'Meta -',
+          'Rollout Sets -',
+          'AotW -',
+          'RANews -',
+          'RA Awards -',
+          'Custom Awards -',
+          'Console Wars',
+          'DevJam',
+          'Challenge League',
+        ],
```
Jamiras, line 165: This isn't matching `[DevQuest ` sets either:
image
image
image

Jamiras, line 165: Not sure how easy it would be to reclassify `[Misc - Junior Developer Graduate Sets]`
image
Or maybe the hub could be renamed to something that would get matched?

resources/js/common/components/PlayableHubsList/PlayableHubsList.tsx
```diff
@@ -44,3 +53,24 @@ export const PlayableHubsList: FC<PlayableHubsListProps> = ({ hubs }) => {
...
+    }
+
+    // There may be hub IDs marked for specific exclusion (such as those shown in game metadata).
+    if (excludeIds?.includes(hub.id)) {
+      return false;
+    }
+
+    return true;
+  });
+}
```
Jamiras, line 76: When all hubs have been promoted, an empty section is shown.
http://localhost:64000/game2/6850
image

resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,279 @@
...
+    () =>
+      buildMetaRowElements({
+        hubs: allGameHubs,
+        hubTitleIncludes: ['Publisher - ', 'Hacks -'],
+        primaryLabel: 'Publisher',
+        altLabels: ['Hacks'],
+        fallbackValue: game.publisher,
+      }),
+    [allGameHubs, game.publisher],
+  );
```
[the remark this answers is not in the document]
wescopeland: Updated in latest:
Screenshot 2025-05-18 at 4 56 38 PM

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: Both issues addressed in latest. Note I renamed "RA Features" to simply "RA". Still not 100% sold on that wording; I'm trying to shy away from reusing the word "feature" across multiple rows.

Screenshot 2025-05-18 at 4 57 36 PM

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: I decided to go with a dedicated row just for event hubs in latest:
Screenshot 2025-05-18 at 4 59 14 PM

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: These are now captured in the "Events" row: https://github.com/RetroAchievements/RAWeb/pull/3522#discussion_r2094625382

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: These should now be captured in the "Events" row: https://github.com/RetroAchievements/RAWeb/pull/3522#discussion_r2094625382

wescopeland: This one is tricky, but I think this should probably be renamed from a Misc. hub to something else.

The first thing that comes to mind is "Junior Developer - Graduate Sets".

resources/js/features/games/hooks/useAllMetaRowElements.ts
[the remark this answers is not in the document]
wescopeland: A handful of things to discuss here:

* In latest, publisher and genre values that aren't linked via a hub should now be loading correctly.
* I'm not quite sure what to do about Central - Homebrew or fangame hubs. Definitely open to ideas. For the Central - Homebrew, it almost seems like that could be linked to the tag somewhere, either in the game title itself or maybe a breadcrumb.
* The Meta|Team hubs only appear if you have the ability to manage games. I don't think these should be publicly visible on a game page for the general population.

resources/js/common/components/PlayableHubsList/PlayableHubsList.tsx
wescopeland: Fixed in latest.

resources/js/features/games/components/GameMetadata/GameMetadata.tsx
```diff
@@ -0,0 +1,136 @@
...
+                      'lg:transition-none lg:active:translate-y-0 lg:active:scale-100',
+                    )}
+                    onClick={() => setIsSeeMoreOpen(true)}
+                  >
+                    {t('See more')}
+                  </BaseButton>
+                </div>
+              </BaseTableCell>
+            </BaseTableRow>
+          ) : null}
```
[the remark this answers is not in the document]
wescopeland: Fixed in latest.

resources/js/features/games/components/GameMetadata/GameMetadata.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  ) as string,
+                },
+              ]}
+            />
+          ) : null}
+
+          <GameMetadataRow
+            rowHeading={t('metaLanguage', { count: languageRowElements.length })}
+            elements={languageRowElements}
+          />
```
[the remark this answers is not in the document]
wescopeland: > For this particular game, it would be easy enough to add the original language hub, but is that something we want to do for every game in the database, just those with supported patches (which will likely increase with the new compatibility testing process), or should there be some way to specify the default language of the game without a hub?

Yes, I'm inclined to believe this is a data issue exacerbated by limited management tooling.

My thoughts are:
* Every language and language patch associated with a game also should have the corresponding hub associated with it.
* We can probably make the language management tooling around this significantly better for both site cleanup (managing game pages) and developers (managing translation patches). These hubs are so routine to link/unlink, we should probably come up with some kind of "streamlined" way to do so. This has also been an explicit request made by the site cleanup team.

Cleaning up this data will not be a small undertaking. I did come across a handful of game pages during my testing where the only language displayed was "Portuguese*".

resources/js/features/games/hooks/useAllMetaRowElements.ts
Jamiras: I'm fine with that. Whatever the title ends up being, it feels like it should get grouped into the Events section with DevQuest sets.

Jamiras: Or maybe we should have a separate Collections sub-category for Site Events?

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: Changed to "Achievement Extras" in latest. I couldn't think of anything better.

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: I don't like the idea of a "Collections" sub-category - this seems like it may be fairly nebulous.

On prod, I've moved the JrDev grad sets hub to Dev Events: https://retroachievements.org/manage/hubs/7975
Additionally, it now has "[Central - Developer Events]" as its only related hub.

With these changes and the code in latest, Junior Developer Graduate Sets now shows up in the "Events" row.

resources/js/features/games/hooks/useAllMetaRowElements.ts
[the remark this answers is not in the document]
wescopeland: > Meta|Team filtering makes sense, but it might be good to have them visible for devs/art team members as they would be the people to make the changes to get them out of said hubs.

This behavior is driven by `can.manageGames`, so anyone with the artist role should be able to see these too. For reference, here's the policy:
```php
{
    return $user->hasAnyRole([
        Role::GAME_HASH_MANAGER,

        Role::DEVELOPER,
        Role::DEVELOPER_JUNIOR,

        Role::ARTIST,
    ]);
}
```

> For "Central - Homebrew" and "Fangames - Mario", it seems like they should just get bundled into the Misc. bucket. Other than the Series hubs, which you've specifically called out as a later thing, and the Meta|Team hubs, it seems like everything that doesn't fall into a nice bucket should just be thrown into Misc.

Haven't tackled this yet, but will do so in a subsequent commit (and mention it in this comments thread).

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: I think it's a good idea. I've spent some time on this, but it gets really tricky when there are edge cases involved (such as the one you've already mentioned). This is especially complicated if there are multiple genre and subgenre hubs linked to a single game.

Unless you feel strongly, let's push this to a future task for consideration.

resources/js/features/games/hooks/useAllMetaRowElements.ts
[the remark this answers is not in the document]
wescopeland: This is now addressed in latest.

With the exception of Meta|Team and Series hubs, ungrouped stuff is now bucketed into Misc.

wescopeland:
@Jamiras I think I've now covered all the feedback, but if there's anything I've missed, please let me know!

resources/js/features/games/hooks/useAllMetaRowElements.ts
```diff
@@ -0,0 +1,322 @@
...
+    regionalRowElements,
+    settingRowElements,
+    technicalRowElements,
+    themeRowElements,
+  ]);
+
+  const miscRowElements = useMemo(
+    () =>
+      buildMiscRowElements(allGameHubs, usedHubIdsFromOtherCategories, {
+        keepPrefixFor: ['Fangames'],
```
Jamiras, line 215: `Rollout Sets - Mega Duck` and `The Unwanted - Past Games` are appearing under both Misc and Events.
http://localhost:64000/game2/20221
image

If they should be under Misc, they should keep their prefix.

And `Freely Available in Forum Topic` could be an "Achievement Extra"?

lang/en_US.json
```diff
@@ -929,5 +927,33 @@
...
+    "metaMeta": "Meta",
+    "metaProtagonist_other": "Protagonists",
+    "metaProtagonist_one": "Protagonist",
+    "metaSetting_other": "Settings",
+    "metaSetting_one": "Setting",
+    "metaRegional": "Regional",
+    "metaHackOf": "Hack of",
+    "metaEvents": "Events",
+    "metaFormat": "Format",
+    "metaRaFeature": "Achievement Extras"
```
Jamiras, line 958: http://localhost:64000/game2/27576
`Set with DLC support` could be an Achievement Extra

resources/js/features/games/hooks/useAllMetaRowElements.ts
Jamiras, line 215: Consider keeping the "Clones" prefix:
http://localhost:64000/game2/13151
image

Jamiras: I wonder if it would be easier to have a list of prefixes _to_ strip: `['Misc.', 'Central']`.

resources/js/features/games/components/GameMetadata/GameMetadata.tsx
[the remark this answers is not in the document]
Jamiras: I assume this will get pushed to a separate PR too.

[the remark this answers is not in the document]
wescopeland: Yep.

lang/en_US.json
wescopeland: Agreed. Renaming the hub from "[Feature -" to "[Meta -" will cause this to automatically happen. I believe this rename is appropriate based on the current hub naming conventions of meta hubs vs feature hubs.

I've made this change on prod as well: https://retroachievements.org/manage/hubs/3484

resources/js/features/games/hooks/useAllMetaRowElements.ts
wescopeland: Fixed in latest so event hubs will no longer appear in Misc. They should appear in Events only.

I'm not sure "Freely Available in Forum Topic" is a good fit for achievement extra. To me, this is more _game-specific_ data -- an indicator that the game is freeware.

wescopeland: Updated in latest so the "Clones" prefix will be kept.

> I wonder if it would be easier to have a list of prefixes to strip: ['Misc.', 'Central'].

I had a similar thought. Though, after going through a bunch of the Misc. hubs, I think the level of effort would be roughly the same.

Jamiras: I guess "Freely Available" is fine where it is. I was just thrown off by the "Forum Topic" subclause and assumed it should go into the "RA Features" bucket, which has been renamed to "Achievement Extras".

Confirmed doubly linked events are now only in Events section.

Jamiras (approved):

PR #3523, opened 2025-05-17
perf(search): limit the scout queue to 1 process

wescopeland:
At least for the time being, the Scout queue is the lowest priority thing we have being provisioned by Horizon.

I don't see any reason for it to need more than a single process, but right now it can consume 2-4 depending on load. These search index synchronizations are not time-critical.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -200,7 +200,6 @@ return [
                 'player-metrics',
                 'player-points-stats',
                 'player-sessions',
-                'scout',
             ],
             'balance' => 'auto',
             'autoScalingStrategy' => 'size',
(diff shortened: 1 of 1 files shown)


config/horizon.php
```diff
@@ -230,6 +229,20 @@
...
+            ],
+            'balance' => 'simple',
+            'processes' => 1, // Fixed at exactly 1 process.
+            'maxTime' => 0,
+            'maxJobs' => 0,
+            'memory' => 128,
+            'tries' => 1,
+            'timeout' => 300, // NOTE timeout should always be at least several seconds shorter than the queue config's retry_after configuration value.
+            'nice' => 0,
+        ],
```
Jamiras, line 245: Is there something I need to do beyond `sail artisan queue:restart` to pick up these changes? It seems like my local instance doesn't want to allocate any processes to the new supervisor.
image

And there doesn't seem to be any tooltip or other way to see why it's got the red X.

wescopeland: Very strange, this is what I see locally:

Screenshot 2025-05-18 at 10 14 41 AM

The only thing I've done is `sail artisan horizon` - no other commands.

Something appears to be wrong with your local Horizon -- both supervisor-1 and supervisor-3 have "scout" queues. Try:
```
sail artisan horizon:terminate
sail artisan horizon:purge
sail artisan horizon
```

Jamiras: That did it.

Jamiras (approved):

PR #3524, opened 2025-05-18
feat(search): index comments, forum posts, and hubs

wescopeland:
This PR adds the `Searchable` trait to `Comment`, `ForumTopicComment`, and `GameSet`.

Additionally, the demo page has been enhanced so you can search for these things:
Screenshot 2025-05-18 at 12 00 56 PM

**How to test this PR:**
```bash
sail artisan horizon # be sure to restart your horizon instance if one is already running

sail artisan scout:import "App\Models\GameSet"
sail artisan scout:import "App\Models\ForumTopicComment"

# this will probably take anywhere from 15-30 minutes
sail artisan scout:import "App\Models\Comment"
```

**Things to note:**
* Comments left by banned users are not searchable. Forum posts are still searchable from banned users.
* When a user is banned, their comments are marked as unsearchable. There is no treatment for unbanning a user (should there be?).
* A few `->saveQuietly()` invocations have been added to `UpdateGameAchievementsMetricsAction`. One affects games, the other affects achievements. Achievements are not indexed yet, but they will be in a subsequent PR.

### DIFF
--- a/app/Models/Comment.php
+++ b/app/Models/Comment.php
@@ -86,6 +86,12 @@ class Comment extends BaseModel
             return false;
         }
 
+        // Don't index empty or extremely short comments (3 chars or less).
+        $trimmedPayload = trim($this->Payload);
+        if (empty($trimmedPayload) || mb_strlen($trimmedPayload) <= 3) {
+            return false;
+        }
+
         return true;
     }
 
--- a/app/Models/Comment.php
+++ b/app/Models/Comment.php
@@ -13,10 +13,12 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Laravel\Scout\Searchable;
 
 class Comment extends BaseModel
 {
     use SoftDeletes;
+    use Searchable;
     /** @use HasFactory<CommentFactory> */
     use HasFactory;
 
@@ -49,6 +51,44 @@ class Comment extends BaseModel
         return CommentFactory::new();
     }
 
+    // == search
+
+    public function toSearchableArray(): array
+    {
+        return [
+            'id' => $this->ID,
+            'user_id' => $this->user_id,
+            'ArticleType' => $this->ArticleType,
+            'ArticleID' => $this->ArticleID,
+            'commentable_type' => $this->commentable_type,
+            'commentable_id' => $this->commentable_id,
+            'body' => $this->Payload, // this is fine, as of 2025-05-18, 99.88% of comments are <1KB
+            'created_at' => $this->Submitted,
+        ];
+    }
+
+    public function shouldBeSearchable(): bool
+    {
+        // Don't index deleted comments.
+        if ($this->deleted_at) {
+            return false;
+        }
+
+        // Don't index automated system comments.
+        if ($this->user_id === self::SYSTEM_USER_ID) {
+            return false;
+        }
+
+        // Don't index comments from banned users.
+        $this->loadMissing('userWithTrashed');
+        $user = $this->userWithTrashed;
+        if ($user->banned_at !== null) {
+            return false;
+        }
+
+        return true;
+    }
+
     // == accessors
 
     public function getEditLinkAttribute(): string
@@ -72,8 +112,6 @@ class Comment extends BaseModel
     {
         throw new Exception('Use derived comment model class in the comments() morphTo() relationship instead of ' . Comment::class . '. Add link attribute getters to the derived class. Use A dedicated controller for it and use the prepared actions.');
     }
-
-    // == accessors
 
     public function getIsAutomatedAttribute(): bool
     {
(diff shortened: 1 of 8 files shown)


app/Models/Comment.php
```diff
@@ -49,6 +51,44 @@ protected static function newFactory(): CommentFactory
...
+        }
+
+        // Don't index comments from banned users.
+        $this->loadMissing('userWithTrashed');
+        $user = $this->userWithTrashed;
+        if ($user->banned_at !== null) {
+            return false;
+        }
+
+        return true;
```
Jamiras, line 89: I wonder if we should have a minimum length for indexing. We very little benefit to indexing all the comments that are just "Mastered!", "Thanks", "Obrigado pelo set" or "+1" 
```
MariaDB [retroachievements-web]> select Payload, count(*) from Comment where length(Payload) < 20 group by 1 order by 2 desc limit 20;
+---------------------+----------+
| Payload             | count(*) |
+---------------------+----------+
| mastered!           |     5306 |
| MASTERED            |      824 |
| manually awarded    |      573 |
| Mastered.           |      545 |
| achievements please |      389 |
| Obrigado pelo set.  |      372 |
| thanks              |      281 |
| "televandalist" .   |      266 |
| Thanks!             |      263 |
| Thank you!          |      261 |
| https://bitma.ru    |      252 |
| Hi                  |      251 |
| under               |      231 |
| need achievements   |      226 |
| Manually awarded.   |      221 |
| <3                  |      197 |
| yes                 |      197 |
| Fixed               |      194 |
| +1                  |      185 |
| f                   |      169 |
+---------------------+----------+
```
But we may want to be able to search on comments that are just single word slurs, so maybe we could have some sort of blacklist for the above?

```
MariaDB [retroachievements-web]> select length(Payload), count(*) from Comment group by 1 order by 1 limit 20;
+-----------------+----------+
| length(Payload) | count(*) |
+-----------------+----------+
|               0 |       39 |
|               1 |      429 |
|               2 |     2165 |
|               3 |     2320 |
|               4 |     3011 |
|               5 |     2628 |
|               6 |     2745 |
|               7 |     3558 |
|               8 |     3777 |
|               9 |     9018 |
|              10 |     4326 |
|              11 |     4042 |
|              12 |     4022 |
|              13 |     3690 |
|              14 |     3713 |
|              15 |     3607 |
|              16 |     4701 |
|              17 |     4740 |
|              18 |     4457 |
|              19 |     4344 |
+-----------------+----------+
```

app/Models/ForumTopicComment.php
```diff
@@ -39,6 +41,42 @@ protected static function newFactory(): ForumTopicCommentFactory
...
+
+    public function toSearchableArray(): array
+    {
+        $body = $this->body;
+
+        // As of 2025-05-18, ~0.23% of payload bodies are extremely large (>20KB).
+        // We'll truncate those.
+        if (mb_strlen($body) > 20000) {
+            $body = mb_substr($body, 0, 20000);
+        }
```
Jamiras, line 54: Would it be worth running this through the shortcode processor to strip out the non-visible stuff (like `[url=X]` or `[img=Y]`, `[b]`, etc)? The tradeoff is that the embedded records (games, achievements, etc) would get expanded to their titles.
image
image

config/scout.php
```diff
@@ -139,12 +142,41 @@
...
+            ],
+
+            ForumTopicComment::class => [
+                'filterableAttributes' => [
+                    'forum_topic_id',
+                    'author_id',
+                    'created_at',
+                ],
+                'sortableAttributes' => ['created_at'],
+                'searchableAttributes' => ['body'],
```
Jamiras, line 165: Consider putting `searchableAttributes` before `sortableAttributes` for consistency. (here and Comment)

Jamiras (commented):
Searching game sets and forum topics in the demo page seemed to work fine. Given that indexing the forum topics consumed 2% of my VM's hard drive and I only have another 7% free at this time, I did not index/test comments. I plan to rebuild my VM with a larger hard drive the next time I import a dump.

app/Models/Comment.php
wescopeland: Hard to say what the best approach is at the moment. I strong agree with having a minimum length for indexing. I've added that in latest:
```php
// Don't index empty or extremely short comments (3 chars or less).
$trimmedPayload = trim($this->Payload);
if (empty($trimmedPayload) || mb_strlen($trimmedPayload) <= 3) {
    return false;
}
```

app/Models/ForumTopicComment.php
wescopeland: I'm not sure at the moment we should do this -- the result models will still contain the non-transformed content, so the mismatch (unless handled gracefully by all consuming UIs) might be confusing or ambiguous.

config/scout.php
wescopeland: Updated in latest.

app/Models/ForumTopicComment.php
wescopeland: I think this applies to user IDs (`[user=1]`) as well, at least initially 🤔

If we use `stripAndClamp()` it implies anything referenced in the post will need to be reindexed when the content referenced in the post changes. This is complicated enough that I think we should probably hold for now on processing before indexing.

PR #3526, opened 2025-05-19
allow jr dev to overwrite empty note

Jamiras:
https://discord.com/channels/310192285306454017/1373734010965983404/1374038530636120065

### DIFF
--- a/app/Connect/Actions/SubmitCodeNoteAction.php
+++ b/app/Connect/Actions/SubmitCodeNoteAction.php
@@ -79,6 +79,10 @@ class SubmitCodeNoteAction extends BaseAuthenticatedApiAction
                 }
 
                 $memoryNote->restore();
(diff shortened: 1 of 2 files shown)


Jamiras:
I wonder if it would make more sense to update the data, rather than modify the code to support the old format?
```
MariaDB [retroachievements]> SELECT count(*) FROM memory_notes WHERE body='' AND deleted_at IS NULL;
+----------+
| count(*) |
+----------+
|   110058 |
+----------+
1 row in set (0.418 sec)
```
Those could be fixed with a simple query:
```
UPDATE memory_notes SET deleted_at=updated_at WHERE body='' AND deleted_at IS NULL;
```

wescopeland:
I like that idea more; I'm on board. It seems more practical than us continuing to support the legacy way of deletion.

PR #3529, opened 2025-05-21
add player:backfill-playtime-total command

Jamiras:
Queues up 1000 `player_games` to be processed every 15 minutes. With approximately 19 million `player_games` records, it will take around 200 days (7 months) to fully backfill the records. Depending on how much additional load this puts on the server, we can tweak it to be more frequent and/or do more/less records at a time. On my local test environment, it took about five minutes to drain the queue after running the command. Based on what we saw the server do on deployment of #3496, I expect it to be able to do it in under 1 minute.


app/Platform/Commands/EnqueueStaleGamePlayerGamesUpdates.php
Jamiras, line 1: This was added in #2530 and disabled shortly after deployment. As we haven't gone back to re-enable it yet, I took this opportunity to remove it.

wescopeland (approved):

PR #3534, opened 2025-05-23
feat(messages): use UserHeading component on index page

wescopeland:
This PR makes a small change to the messages index pages (`/messages` and `/messages/{user}`).

Now, `<UserHeading />` is used for the H1, so the user avatar displays for the currently-selected inbox:
Screenshot 2025-05-22 at 8 02 52 PM

This makes it a little more obvious which inbox is currently active for the user, assuming they have multiple available.


Jamiras (approved):
Should something similar be done on the create/edit/view message pages?

wescopeland:
Closing. I have an upcoming PR that takes a different approach which I think will be a bit better.

PR #3538, opened 2025-05-23
find parent game through game_achievement_sets

Jamiras:
addresses https://discord.com/channels/310192285306454017/888607146290343976/1375483841451524136

### DIFF
--- a/app/Helpers/database/game.php
+++ b/app/Helpers/database/game.php
@@ -30,18 +30,6 @@ function getGameData(int $gameID): ?array
     ]);
 }
 
-// If the game is a subset, identify its parent game.
-function getParentGameFromGameTitle(string $title, int $consoleId): ?Game
-{
-    if (mb_strpos($title, '[Subset') !== false) {
-        $foundGame = Game::where('Title', $title)->where('ConsoleID', $consoleId)->first();
-
-        return $foundGame->getParentGame() ?? null;
-    }
-
-    return null;
-}
-
 function getGameMetadata(
     int $gameID,
     ?User $user,
@@ -102,14 +90,7 @@ function getGameMetadata(
     }
 
     $metricsColumns = '';
-    $metricsJoin = '';
-    $metricsBindings = [];
     if ($metrics) {
-        $parentGame = getParentGameFromGameTitle($gameDataOut['Title'], $gameDataOut['ConsoleID']);
-        $numDistinctPlayersSelector = $parentGame?->players_total ?: getGameData($gameID)['NumDistinctPlayers'];
-        $gameDataOut['ParentGameID'] = $parentGame?->id;
-        $gameDataOut['NumDistinctPlayers'] = $numDistinctPlayersSelector ?? 0;
-
         $metricsColumns = 'ach.unlocks_total AS NumAwarded, ach.unlocks_hardcore_total AS NumAwardedHardcore,';
     }
 
@@ -131,14 +112,13 @@ function getGameMetadata(
         ach.type
     FROM Achievements AS ach
     LEFT JOIN UserAccounts AS ua ON ach.user_id = ua.ID
-    $metricsJoin
     WHERE ach.GameID = :gameId AND ach.Flags = :achievementFlag AND ach.deleted_at IS NULL
     $orderBy";
 
-    $achievementDataOut = legacyDbFetchAll($query, array_merge([
+    $achievementDataOut = legacyDbFetchAll($query, [
         'gameId' => $gameID,
         'achievementFlag' => $flag->value,
-    ], $metricsBindings))
+    ])
         ->keyBy('ID')
         ->toArray();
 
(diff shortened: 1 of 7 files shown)


app/Helpers/database/game.php
```diff
@@ -102,14 +90,7 @@ function getGameMetadata(
     }
 
     $metricsColumns = '';
-    $metricsJoin = '';
-    $metricsBindings = [];
     if ($metrics) {
-        $parentGame = getParentGameFromGameTitle($gameDataOut['Title'], $gameDataOut['ConsoleID']);
-        $numDistinctPlayersSelector = $parentGame?->players_total ?: getGameData($gameID)['NumDistinctPlayers'];
-        $gameDataOut['ParentGameID'] = $parentGame?->id;
-        $gameDataOut['NumDistinctPlayers'] = $numDistinctPlayersSelector ?? 0;
```
Jamiras, line 111: `NumDistinctPlayers` is already retrieved via `getGameData` on line 84. The update metrics action should ensure the subset has the same player count as the base set.

`ParentGameID` didn't seem to be used anywhere

app/Models/Game.php
```diff
@@ -322,27 +322,6 @@ public function shouldBeSearchable(): bool
         return true;
     }
 
-    // TODO refactor when game_achievement_sets is ready
```
Jamiras, line 325: I think it's ready. If not, this PR probably has to wait.

app/Models/Game.php
```diff
@@ -367,27 +346,29 @@ public function getImageIngameUrlAttribute(): string
         return media_asset($this->ImageIngame);
     }
 
-    public function getIsSubsetGameAttribute(): bool
+    public function getParentGameIdAttribute(): ?int
     {
-        if (str_contains($this->title, "[Subset")) {
-            return true;
-        }
```
Jamiras, line 374: I eliminated this check, but it could be used as a short-circuit for the `isSubsetGameAttribute` (still here, just lower in diff because the body got moved into another function). But as most games won't match the substring, you could argue it actually adds performance impact most of the time.

app/Models/Game.php
```diff
@@ -367,27 +346,29 @@ public function getImageIngameUrlAttribute(): string
...
         }
 
         // Check if any of these achievement sets are used in other games with non-core types.
-        $hasNonCoreUsages = GameAchievementSet::whereIn('achievement_set_id', $achievementSets)
+        $nonCoreUsage = GameAchievementSet::whereIn('achievement_set_id', $achievementSets)
             ->where('game_id', '!=', $this->id)
             ->where('type', '!=', AchievementSetType::Core)
-            ->exists();
+            ->select('game_id')
+            ->first();
```
Jamiras, line 364: Is it possible for a subset to be associated to multiple base games?

app/Models/Game.php
```diff
@@ -507,6 +488,11 @@ public function achievementSetClaims(): HasMany
         return $this->hasMany(AchievementSetClaim::class, 'game_id');
     }
 
+    public function parentGame(): ?Game
+    {
+        return $this->parentGameId ? Game::find($this->parentGameId) : null;
```
Jamiras, line 493: Correct me if I'm wrong, but `parentGameId` will only be evaluated once and stored in the model. Right?

app/Models/Game.php
wescopeland: Yes, such as for TLoZ: Oracle of Ages and TLoZ: Oracle of Seasons.

app/Models/Game.php
wescopeland: I'm 99% sure it'll query the DB every time it's called, even on the same model. This sort of caching only works for actual relationships.

I think you have two options:
* Convert it to a relationship, or
* Use Laravel's built-in `once()` helper, [docs](https://laravel.com/docs/11.x/helpers#method-once).

```php
// untested psuedocode
return once(fn () => $this->parentGameId ? Game::find().......
```

app/Models/Game.php
Jamiras: Interesting. And it doesn't appear like that can find the parent on prod because the title doesn't match:
image
image

This code also can't find it because neither game has it linked as a subset, which leads back to questioning if we're at a point where `when game_achievement_sets is ready`.

What sort of `orderBy` should I add to make the parent consistent?

app/Models/Game.php
Jamiras: Confirmed :(
```
before parentGame()
in parentGameIdAttribute
in parentGameIdAttribute
after parentGame()
```

wescopeland: This `return` should probably also be wrapped in `once()` (similar to the pseudocode example above).

app/Models/Game.php
wescopeland: > which leads back to questioning if we're at a point where when game_achievement_sets is ready.

I'm not sure how `game_achievement_sets` could handle this differently than it already is (assuming the subset is attached to both games properly).

I think in this case, we can't possibly have a breadcrumb that's correct because we're effectively fanning-in from two distinct games.

> What sort of orderBy should I add to make the parent consistent?

Good question - maybe `game_achievement_sets.created_at`? That's all that's coming to mind, though.

wescopeland (approved):

PR #3554, opened 2025-05-25
feat(hubs): add support for role-based access control

wescopeland:
This PR adds support for RBAC to hubs.

The ability to manage RBAC is restricted to admins only. [Navigating to a hub edit view](http://localhost:64000/manage/hubs/458/edit) in the management app as an admin now shows:
Screenshot 2025-05-25 at 11 35 54 AM

If RBAC is enabled on a hub, anyone with hub management abilities can see the hub's current RBAC settings.

If viewing the hub is restricted and the current user doesn't have the correct role to view the hub, game pages will not show the hub in the sidebar. Additionally, hub pages will not show the hub in the list of related hubs. When attempting to view the hub without permissions, the user will get an unauthorized error.

If editing the hub is restricted and the current user doesn't have the correct role to edit the hub, edit buttons are hidden and the user will get an unauthorized error if they try to manually navigate to the hub's edit page.


app/Policies/GameSetPolicy.php
```diff
@@ -32,6 +32,24 @@ public function view(?User $user, GameSet $gameSet): bool
...
+        if ($gameSet->has_view_role_requirement) {
+            // Admins can always view hubs.
+            if ($user->hasRole(Role::ADMINISTRATOR)) {
+                return true;
+            }
+
+            // If user is not authenticated, they cannot view the hub.
+            if (!$user) {
+                return false;
+            }
```
Jamiras, line 45: This needs to be above line 38.

app/Models/GameSet.php
```diff
@@ -61,6 +62,31 @@ public static function boot()
...
+                    ->withProperty('old', [$relationName => null])
+                    ->withProperty('attributes', [$relationName => $attachedRoles
+                        ->map(fn ($role) => [
+                            'id' => $role->id,
+                            'name' => $role->name,
+                            'permission' => $permission,
+                        ]),
+                    ])
+                    ->event('pivotAttached')
+                    ->log('pivotAttached');
```
Jamiras, line 87: Not sure if there's an easy way to consolidate these into a single activity. I added two create roles and one update role and it created three activity entries.
image

app/Filament/Resources/HubResource.php
Jamiras, line 1: Should a user without an update role still be able to associate/detach games to/from a restricted hub?

app/Models/GameSet.php
Jamiras: Indirectly related: My Add Role window always lists all available roles by ID, even if I add search text.
image

wescopeland: > Not sure if there's an easy way to consolidate these into a single activity. I added two create roles and one update role and it created three activity entries.

This is really tough unless we move all the logging to explicitly live inside `HubResource`. I'm not sure this trade-off is worth it.

> Indirectly related: My Add Role window always lists all available roles by ID, even if I add search text.

I was able to get this working right in latest for the "Attach role" dialog. I couldn't get the filtering to work right in the select controls in `HubResource` no matter what I tried.

app/Policies/GameSetPolicy.php
wescopeland: Good catch - updated in latest.

app/Filament/Resources/HubResource.php
Jamiras: Logged in as a Developer (no other roles), I still have an Add Games button and a Trash icon for the individual rows:
image

wescopeland: Resolved in latest.

Jamiras: The hub can still be associated/disassociated from the game page:
image

wescopeland: Resolved in latest:
Screenshot 2025-06-02 at 3 53 44 PM

Jamiras: Verified can no longer remove from game hubs page, but can still add: 
image
image

wescopeland: Good catch - should now be resolved in latest.

I've made a change such that non-editable hubs should no longer appear in search results. If the user manually enters a non-editable ID in the CSV field and submits, it'll be excluded from edit.

Jamiras: I wonder if you shouldn't just use the "excluded from edit" behavior for entering the hub name. As is, searching for the hub name makes it appear that the hub doesn't exist, even though the user does have permission to see it.

wescopeland: Updated in latest. Now the options appear, but if the user doesn't have permission to select them, they're disabled:
Screenshot 2025-06-07 at 4 50 10 PM

PR #3555, opened 2025-05-25
feat: add weighted points crawler

wescopeland:
This PR implements a weighted points crawler which is tasked with slowly updating all `User.TrueRAPoints` and `player_achievement_sets.weighted_points` values.

**How it works:**
The following has been added to _app/Platform/AppServiceProvider.php_:
```php
$schedule->command(CrawlPlayerWeightedPoints::class)->everyFiveMinutes();
```
Every five minutes, this command is invoked which updates data for a burst of 20 users. This processing is very low-overhead. The crawler is meant to be a set-and-forget kind of thing.

The crawler remembers its place between invocations, and only considers users who have a minimum of 1,250 RetroPoints.

You can invoke the crawler manually:
```bash
sail artisan ra:platform:player:crawl-weighted-points
```
```
Processing batch of 20 users...
Starting after user ID: 23012.
Batch completed. Processed up to user ID: 23201.
Updated 20 users' weighted points.
Updated 301 player achievement sets.
Approximately 57646 users remaining to process.
```

You can also target a specific user for testing purposes:
```bash
sail artisan ra:platform:player:crawl-weighted-points --user=1
```

According to my rough math, each full crawl of applicable users will take a little over two weeks.


app/Platform/AppServiceProvider.php
Jamiras, line 1: Why is this necessary? 

`player_achievement_sets.points_weighted` gets updated by [UpdatePlayerGameMetricsAction](https://github.com/RetroAchievements/RAWeb/blob/bb0107950061f524dde4d2b379f07eb2d3611ae4/app/Platform/Actions/UpdatePlayerGameMetricsAction.php#L124).

`Achievements.TrueRatio` gets updated by [UpdateGameAchievementMetricsAction](https://github.com/RetroAchievements/RAWeb/blob/bb0107950061f524dde4d2b379f07eb2d3611ae4/app/Platform/Actions/UpdateGameAchievementsMetricsAction.php#L70), which is invoked by [UpdateGameMetricsAction](https://github.com/RetroAchievements/RAWeb/blob/bb0107950061f524dde4d2b379f07eb2d3611ae4/app/Platform/Actions/UpdateGameMetricsAction.php#L117).

Shouldn't `UpdateGameAchievementsMetricsAction` just batch up `UpdatePlayerGameMetricsAction` calls for any users who have unlocked the achievement if its TrueRatio changes? Or maybe make direct queries:
```
$before = $set->points_weighted;
$set->points_weighted = $hardcoreUnlocks->sum('TruePoints');
$change = $set->pointsWeighted - $before;
if ($change != 0)
    User::whereIn('ID', PlayerAchievements::where('achievement_id', $achievementId)->whereNotNull('hardcore_unlocked_at')->pluck('user_id'))
        ->increment('TrueRAPoints', $change)
```

I guess we'd still need something to re-sync all the existing records.

wescopeland: It sounds like a good course of action might be to trim the crawler functionality to just cover `UserAccounts.TrueRAPoints`.

wescopeland: In latest, I've ripped out everything unrelated to `UserAccounts.TrueRAPoints`. This grants a significant amount of CPU breathing room: I've upped the batch size from 20 to 100.

app/Platform/Commands/CrawlPlayerWeightedPoints.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private function updateUserWeightedPoints(User $user): void
+    {
+        // Calculate the sum of weighted points for all hardcore achievements unlocked by the user.
+        // Use a direct query to avoid loading all achievements into memory.
+        $weightedPointsSum = $user->playerAchievements()
+            ->join('Achievements', 'player_achievements.achievement_id', '=', 'Achievements.ID')
+            ->whereNotNull('player_achievements.unlocked_hardcore_at')
+            ->sum('Achievements.TrueRatio');
```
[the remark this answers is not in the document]
wescopeland: I could be missing something in my understanding. This approach you're proposing seems unsafe only because of the scale we're dealing with. The Eloquent ORM query from the pseudocode above translates directly to:
```sql
SELECT DISTINCT
  `user_id`
FROM
  `player_achievements`
WHERE
  `achievement_id` IN ($pointsWeightedChangedAchievementIds)
  AND `unlocked_hardcore_at` IS NOT NULL
```

For the most popular sets, this would dispatch many thousands of jobs in rapid succession. For example:
* Achievement 9 on my local returns over 17,000 results.
* Achievement 342 returns over 35,000 results.

I agree the crawler isn't ideal architecturally (which is why I've held off on it for so long), but it does have the advantage of being predictable, in that we can tightly control its resource usage without worrying about nuanced supervisor provisioning config.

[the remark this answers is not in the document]
Jamiras: I'm fine if you want to go back to the crawler implementation, but it will have to update the `player_games` records in addition to the final total in the `User` record.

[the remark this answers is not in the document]
wescopeland: I've stuck with the crawler implementation in latest, but `player_games` are now also correctly being updated (in addition to the final total in the `User` record).

For the time being at least, I'm a fan of the crawler approach, as the load it produces is predictable, constant, and very easy to tune up or down. Additionally, to your point, we have a bunch of stale data that needs some mechanism to go through and correct it.

This certainly doesn't prevent us from moving to an event-driven approach in the future, especially after all the stale data has been corrected. Given the current implementation, I don't believe this would be too tough to achieve with a few hours of applied refactoring work.

app/Platform/Commands/CrawlPlayerWeightedPoints.php
```diff
@@ -0,0 +1,169 @@
...
+                INNER JOIN Achievements ach ON ach.ID = pa.achievement_id
+                WHERE
+                    pa.user_id = ?
+                    AND pa.unlocked_hardcore_at IS NOT NULL
+                GROUP BY pa.user_id, ach.GameID
+            ) AS calculated ON pg.user_id = calculated.user_id AND pg.game_id = calculated.game_id
+            SET pg.points_weighted = COALESCE(calculated.weighted_points, 0)
+            WHERE
+                pg.user_id = ?
+                AND pg.points_weighted != COALESCE(calculated.weighted_points, 0)
```
Jamiras: I'm not sure this is working as intended. I purposefully found a user what would get processed in the next batch and ran him first, and it still says 100 records updated:
```
jamiras@ubuntu24 pr/3555:~/source/RAWeb$ sail artisan ra:platform:player:crawl-weighted-points --user=1624
Updating user [1624:UNHchabo].
Current weighted points: 87303.
New weighted points: 75537.
Done.
jamiras@ubuntu24 pr/3555:~/source/RAWeb$ sail artisan ra:platform:player:crawl-weighted-points
Processing batch of 100 users...
Starting after user ID: 1608.
Batch completed. Processed up to user ID: 2372.
Updated 98 users' weighted points.
Updated 100 player_games records.
Approximately 64697 users remaining to process.
```

It's also weird that it always returns 100. I'd expect some users to have more than 1 game played.

Is this metric even important? Would this be any faster without the "unchanged" filter?

wescopeland: Confirmed with a local benchmark it's anywhere from ~20-50% faster without the "unchanged" filter.

Additionally, in latest, we just log (for observability purposes) how many players and `player_games` records were processed.

Jamiras (approved):

PR #3558, opened 2025-05-26
add action for submitgametitle connect api

Jamiras:


app/Connect/Actions/SubmitGameTitleAction.php
```diff
@@ -0,0 +1,148 @@
...
+        $this->gameId = request()->integer('g');
+
+        return null;
+    }
+
+    protected function process(): array
+    {
+        if (!$this->user->can('create', Game::class)) {
+            // TODO: Developers are currently allowed to create new titles,
+            //       but the policy doesn't allow it.
```
Jamiras, line 61: https://github.com/RetroAchievements/RAWeb/blob/bb0107950061f524dde4d2b379f07eb2d3611ae4/app/Policies/GamePolicy.php#L40-L46

wescopeland: I suspect this is done so the create action never surfaces for games in Filament.

This is no fault of your own - it's cruft from the primordial beginnings of setting up Filament and `spatie/laravel-permission`. However, I can't help but feel this seems like an anti-pattern:
```php
if (!$this->user->can('create', Game::class) {
  $permissions = ... stuff ...
  if ($permissions < Permissions::Developer) {
    // ... error ...
  }
}
```

I think a better course of action is to update the policy by uncommenting line 44 (where `Role::DEVELOPER` is) and then completely remove _app/Filament/Resources/GameResource/Pages/Create.php_ as well as the reference to it in _GameResource.php_.

Once this is done, we can just do a policy check and not worry about legacy permissions.

Given how many side effects are involved with game creation, as well as a game's dependency on having at least one hash for non-standalone games to even be viable, we probably don't want to support manual creation of games in Filament.

app/Connect/Actions/SubmitGameTitleAction.php
```diff
@@ -0,0 +1,148 @@
...
+            $gameHash = new GameHash([
+                'game_id' => $game->id,
+                'user_id' => $this->user->id,
+                'md5' => $this->hash,
+                'compatibility' => GameHashCompatibility::Compatible,
+            ]);
+            if (!empty($this->hashDescription)) {
+                $gameHash->name = $this->hashDescription;
+            }
+            $gameHash->save();
```
wescopeland, line 132: Very minor edge case: between checking if a hash exists and then creating it, another inbound request could create the same hash, leading to a duplicate key error.

This can probably be mitigated with `firstOrCreate`:
```php
$gameHash = GameHash::firstOrCreate(
  ['game_id' => $game->id, 'md5' => $this->hash],
  [
    'user_id' => $this->user->id,
    'compatibility' => GameHashCompatibility::Compatible,
  ]
);
```

wescopeland (approved):

PR #3564, opened 2025-05-28
ensure unranked_at is updated when user's Untracked value changes

Jamiras:
Implements double-writing suggested [here](https://discord.com/channels/310192285306454017/1305921595843285054/1376926752667996190).

Queries to correct existing data:
```
UPDATE UserAccounts SET unranked_at=Deleted WHERE Untracked=1 AND unranked_at IS NULL AND Deleted IS NOT NULL;

UPDATE UserAccounts SET unranked_at=banned_at WHERE Untracked=1 AND unranked_at IS NULL AND banned_at IS NOT NULL;

UPDATE UserAccounts ua INNER JOIN (
  SELECT c.ArticleID, MAX(c.Submitted) AS max_submitted FROM UserAccounts u
  JOIN Comment c ON c.ArticleID=u.ID WHERE c.ArticleType=9 
  AND c.Payload LIKE '% set status to Untracked%' 
  AND u.unranked_at IS NULL AND u.Untracked=1 GROUP BY u.ID
) AS temp ON temp.ArticleID = ua.ID
SET ua.unranked_at = temp.max_submitted;

UPDATE UserAccounts SET unranked_at=LastLogin WHERE Untracked=1 AND unranked_at IS NULL AND LastLogin IS NOT NULL;
```
The first query sets unranked_at to deletion time for all deleted accounts where it's not set
The second sets unranked_at to ban time for all banned accounts where it's not set
The third sets unranked_at to the timestamp of the audit log associated to unranking the user where it's not set
The fourth sets unranked_at to the last time the account was used for all other accounts where it's not set


app/Actions/ClearAccountDataAction.php
```diff
@@ -73,6 +74,7 @@ public function execute(User $user): void
                 'permissions' => min($user->Permissions, Permissions::Unregistered),
                 'userId' => $user->ID,
                 'now' => Carbon::now(),
+                'now2' => Carbon::now(),
```
Jamiras, line 77: This is the double-write for deletion

app/Helpers/database/player-rank.php
```diff
@@ -8,14 +8,13 @@
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Cache;
 
-function SetUserUntrackedStatus(User $user, int $isUntracked): void
+function SetUserUntrackedStatus(User $user, bool $isUntracked): void
 {
     $user->Untracked = $isUntracked;
+    $user->unranked_at = $isUntracked ? now() : null;
```
Jamiras, line 14: This is the double-write for manually unranking something. Banning is already handled by https://github.com/RetroAchievements/RAWeb/blob/e5eb5c086b6a1e5307f35aa22388b35d785b32b5/app/Helpers/database/user-permission.php#L168

app/Helpers/database/player-rank.php
```diff
@@ -8,14 +8,13 @@
...
     $user->save();
 
-    PlayerRankedStatusChanged::dispatch($user, (bool) $isUntracked);
-
-    // TODO update games that are affected by this user's library
```
Jamiras, line 18: The `PlayerRankedStatusChanged` event triggers `DispatchUpdateGameMetricsForGamesPlayedByUserJob` jobs.

public/request/user/update.php
```diff
@@ -27,6 +27,10 @@
 $foundSourceUser = User::whereName($user)->first();
 $foundTargetUser = User::whereName($targetUsername)->first();
 
+if (!$foundSourceUser || !$foundTargetUser) {
+    return back()->withErrors(__('legacy.error.error'));
+}
```
Jamiras, line 32: This prevents a possible null exception that could occur in the first couple branches below, and moves the logic from the other branches.

wescopeland (commented):
I believe it may also be possible to mutate `Untracked` from `UserResource`.

Jamiras:
> I believe it may also be possible to mutate `Untracked` from `UserResource`.

It was. Addressed.

PR #3565, opened 2025-05-29
write audit log when deleting official forum topic

Jamiras:
Should make tracking down what happened [here](https://discord.com/channels/476211979464343552/1002696640001478769/1377580705806028860) easier in the future.


app/Community/Controllers/Api/ForumTopicApiController.php
```diff
@@ -53,6 +54,14 @@ public function destroy(ForumTopic $topic): JsonResponse
 
         $topic->delete();
 
+        // if this was the official forum topic for a game, clear out the association
+        // so another forum topic can be created - soft deletes don't cascade update
+        // the field to null. this also creates an audit log entry on the game.
+        foreach (Game::where('ForumTopicID', $topic->id)->get() as $game) {
+            $game->ForumTopicID = null;
+            $game->save();
+        }
```
wescopeland, line 63: Possible to do this without a loop and using a single query?
```php
Game::where('ForumTopicID', $topic->id)->update(['ForumTopicID' => null]);
```

PR #3570, opened 2025-05-30
Allow querying users who have earned specific event badges

Jamiras:
Adds a link to the Award Tiers list of the event page that allows browsing through all users who have earned the badge, ordered by when they earned it.

This is the first step towards implementing https://discord.com/channels/310192285306454017/1330283616533614692/1376317309110718574.

The other request is the ability to make the columns sortable. It seems like that should be done using `DataTableColumnHeader` and the related components, but they're currently under the `game-list` component tree. If these should be reusable, they should be moved into the common components folder, and that should be done in a separate PR.

image

image


resources/views/layouts/demo.blade.php
wescopeland, line 1: nit: Assuming this is testing/debugging code, it's probably safe to undo.

resources/js/test/factories/createEvent.ts
wescopeland, line 1: You'll want to use `createRaEvent`. `Event` has special meaning in React which could cause ambiguity.

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
```diff
@@ -39,67 +48,69 @@ export const AwardTierItem: FC<AwardTierItemProps> = ({ event, eventAward, hasVi
...
-            alt={eventAward.label}
-            className={cn(
-              'size-12 rounded-sm transition',
-              eventAward.earnedAt
-                ? 'opacity-100 outline outline-2 outline-offset-1 outline-[gold]'
-                : 'opacity-50 group-hover:opacity-100',
-            )}
-          />
-        </div>
+      <a href={awardEarnersLink}>
```
wescopeland: This should use `<InertiaLink />`. Any link to an Inertia page should use `<InertiaLink />` rather than `<a />` to get an immediate performance boost on navigation.

resources/js/pages/event/[event]/award-earners.tsx
```diff
@@ -0,0 +1,34 @@
+import { useTranslation } from 'react-i18next';
+
+import { SEO } from '@/common/components/SEO';
+import { AppLayout } from '@/common/layouts/AppLayout';
+import type { AppPage } from '@/common/models';
+import { EventAwardEarnersMainRoot } from '@/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot';
```
Comment (wescopeland) [inline resources/js/pages/event/[event]/award-earners.tsx]: nit: This should import from an _index.ts_ file in the _EventAwardEarnersMainRoot_, eg:
```ts
import { EventAwardEarnersMainRoot } from '@/features/events/components/EventAwardEarnersMainRoot';
```

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
wescopeland: If the award has 0 earners, this should probably be a div rather than an anchor, ie:
```tsx
const isLinkable = event.badgeCount! > 0;
const Component = isLinkable ? InertiaLink : 'div';

<Component
  href={isLinkable ? route(...) : undefined}
  {/* stuff */}
/>
```

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
```diff
@@ -39,67 +48,69 @@ export const AwardTierItem: FC<AwardTierItemProps> = ({ event, eventAward, hasVi
...
-                </>
-              ) : null}
+          <div className="flex w-full items-center justify-between gap-2">
+            <div className="flex flex-col">
+              <div className="flex items-center gap-2">
+                {!hasVirtualTier ? (
+                  <>
+                    <p data-testid="award-tier-label" className="flex gap-2 text-xs font-medium">
+                      {cleanedAwardLabel}
+                    </p>
```
wescopeland: There are two subtle animation bugs here that occur if the user hasn't earned the award. Both occur on hover.

1. The hover zone of the `<a>` is deeper than the hover zone of the image opacity effect, resulting in this:

Screenshot 2025-06-02 at 3 12 31 PM

Notice the image opacity is 100%, but the text color is not `text-link-hover`.

2. The opacity change has a transition, but the text link color does not. When issue 1 above is fixed, this causes the opacity change to become disorienting.

Here's how to fix both issues:
```tsx
<p
  data-testid="award-tier-label"
  className={cn([
    'flex gap-2 text-xs font-medium',
    isLinkable ? 'transition group-hover:text-link-hover' : null,
  ])}
>
  {cleanedAwardLabel}
</p>
```

resources/js/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot.tsx
wescopeland, line 1: Screenshot 2025-06-02 at 3 20 14 PM

Recommend extracting `cleanEventAwardLabel` to _common/utils_ and using it on the page heading and breadcrumb label.

app/Platform/Services/AwardEarnersService.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function numEarners(): int
+    {
+        return $this->baseQuery()->count();
+    }
+
+    /**
+     * @return Builder<PlayerBadge>
+     */
+    public function allEarners(int $offset = 0, int $count = 10): Builder
```
wescopeland: These parameters and their corresponding `offset()` and `limit()` aren't actually used. `->paginate()` overrides these.

The implementation can be changed to:
```php
/**
 * @return Builder<PlayerBadge>
 */
public function allEarners(): Builder
{
    return $this->baseQuery()->with('user')
        ->orderBy('AwardDate');
}
```
This makes more sense for other contexts outside of the controller as well, as the function is already returning a `Builder`.

app/Platform/Services/AwardEarnersService.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    private function baseQuery(): Builder
+    {
+        return PlayerBadge::query()
+            ->where('AwardType', $this->awardType)
+            ->where('AwardData', $this->awardId)
+            ->where('AwardDataExtra', $this->awardExtra)
+            ->whereHas('user', function ($query) {
+                return $query->tracked();
+            });
```
wescopeland: I know `unranked_users` was implemented after this was written. Apologies for my delayed review on this PR. This filter may be a good candidate for using that table.

resources/views/layouts/demo.blade.php
Jamiras: gone.

resources/js/test/factories/createEvent.ts
Jamiras: We've had that discussion before. There's already a `createRaEvent`...

resources/js/pages/event/[event]/award-earners.tsx
Comment (Jamiras) [reply resources/js/pages/event/[event]/award-earners.tsx]: done

resources/js/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot.tsx
Jamiras: I think it's fine in `events/utils`. Updated code to use it.

app/Platform/Services/AwardEarnersService.php
Jamiras: done

app/Platform/Services/AwardEarnersService.php
Jamiras: I'm sure that's leftover cruft from whatever I copy/pasted to make this file. Removed.

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
Jamiras: I admit the page is very bare as-is if the list is empty, but only slightly more so than if there's only one earner.
http://localhost:64000/event/33-achievement-of-the-week-2019-top-players/award-earners?tier=2

The page does provide a larger view of the badge image, but that's probably not enough of a reason to keep the link.

I'm concerned about the inconsistency in the list if the link is conditional:
image

The yellow text next to the gold border implies a relationship to me having earned the badge more so than the fact that the link is clickable.

I'm going to leave the link present for all badges for now.

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
Jamiras: I'm not sure what effect you're reporting is problematic, but I've applied to prescribed changes.

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
wescopeland: I very much agree with you that having consistency in design is good, as inconsistent design can sometimes also be dishonest design. But I think _functional consistency_ is also important, which is what I'm advocating for here.

Many UIs, including our own, conditionally adjust elements based on data state. For example, we already conditionally show/hide elements throughout the site - disabled buttons, hidden menu items based on user context, etc. The design adapts to the user's or entity's context.

When users click through to a page which shows a redundant 0 earners label, they're met with essentially no content. This wastes their time and creates a frustrating dead-end experience. I believe it's good UX to differentiate between browsable content (awards with earners) and reference-only content (awards without earners). The addition of a link serves as a progressive disclosure.

As the event awards get earned by players, links would appear automatically, creating a nice sense of progression where new content becomes explorable as the community unlocks it.

resources/js/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot.tsx
wescopeland: This will likely inevitably wind up in _common/utils_ when it's used by another feature, such as search. It's okay to keep in _events/utils_ for now.

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
Jamiras: done

Jamiras:
I'm still not seeing the difference, but I've made the changes anyway
image

wescopeland:
Needs to be:
```tsx
    awardEarnersLink ? 'transition group-hover:text-link-hover' : null,
```
The logic right now is checking for `!== null`, and because the no-link state is an empty string, it's evaluating as truthy.

wescopeland:
Exact accuracy would be `awardEarnersLink !== ""`.

Jamiras:
Ah, right. I had to do that because it didn't like me assigning null to `href`.

wescopeland (approved):

PR #3572, opened 2025-05-30
perf(UpdateGameAchievementsMetricsAction): resolve N+1 writes, add dirty checks

wescopeland:
This PR:
* Adds dirty checking to `UpdateGameAchievementsMetricsAction`.
* Rounds values to 9 decimal places better support the dirty checking.
* Performs all achievement updates in a single atomic DB query.

### DIFF
--- a/app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
+++ b/app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
@@ -7,6 +7,7 @@ namespace App\Platform\Actions;
 use App\Models\Game;
 use App\Models\PlayerAchievement;
 use App\Platform\Services\SearchIndexingService;
+use Illuminate\Support\Facades\DB;
 
 class UpdateGameAchievementsMetricsAction
 {
@@ -18,6 +19,20 @@ class UpdateGameAchievementsMetricsAction
         $playersTotal = $game->players_total;
         $playersHardcore = $game->players_hardcore;
 
+        // If players_total is 0, calculate from actual unlocks
+        // This handles cases where metrics are updated before game player counts (ie: tests)
+        if ($playersTotal === 0) {
+            $playerCounts = PlayerAchievement::query()
+                ->whereIn('achievement_id', $game->achievements()->published()->pluck('ID'))
+                ->whereHas('user', function ($query) { $query->tracked(); })
+                ->selectRaw('COUNT(DISTINCT user_id) as total_players')
+                ->selectRaw('COUNT(DISTINCT CASE WHEN unlocked_hardcore_at IS NOT NULL THEN user_id END) as hardcore_players')
+                ->first();
+
+            $playersTotal = $playerCounts->total_players ?? 0;
+            $playersHardcore = $playerCounts->hardcore_players ?? 0;
+        }
+
         // force all unachieved to be 1
         $playersHardcoreCalc = $playersHardcore ?: 1;
         $pointsWeightedTotal = 0;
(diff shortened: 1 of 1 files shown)


app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
```diff
@@ -65,14 +86,45 @@ public function execute(Game $game): void
...
+            // Only update the achievement if values have actually changed.
+            $isDirty =
+                $achievement->unlocks_total !== $achievement->getOriginal('unlocks_total')
+                || $achievement->unlocks_hardcore_total !== $achievement->getOriginal('unlocks_hardcore_total')
+                || (float) $achievement->unlock_percentage !== (float) $achievement->getOriginal('unlock_percentage')
+                || (float) $achievement->unlock_hardcore_percentage !== (float) $achievement->getOriginal('unlock_hardcore_percentage')
+                || $achievement->TrueRatio !== $achievement->getOriginal('TrueRatio');
+
+            // If the achievement is truly dirty, add it to our list of batch updates.
+            if ($isDirty) {
```
Jamiras, line 109: A much simpler way to do this is to add casts to the AchievementModel:
```
    protected $casts = [
        ...
        'unlock_percentage' => 'double',
        'unlock_hardcore_percentage' => 'double',
    ];
```
And then use the `$achievement->isDirty()` method. I was doing that earlier when we were discussing the issue.

app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
```diff
@@ -82,4 +134,78 @@ public function execute(Game $game): void
...
+         *   unlock_hardcore_percentage = CASE ID
+         *     WHEN X THEN Y
+         *   END,
+         *   TrueRatio = CASE ID
+         *     WHEN X THEN Y
+         *   END,
+         *   Updated = "..."
+         * WHERE
+         *   ID IN ( ... );
+         */
```
Jamiras, line 169: I don't see how this could possibly be faster than doing a `save()` on each modified record. The query engine will have to process each case statement for every row.

app/Platform/Actions/UpdateGameAchievementsMetricsAction.php
```diff
@@ -18,6 +19,20 @@ public function execute(Game $game): void
...
+            $playerCounts = PlayerAchievement::query()
+                ->whereIn('achievement_id', $game->achievements()->published()->pluck('ID'))
+                ->whereHas('user', function ($query) { $query->tracked(); })
+                ->selectRaw('COUNT(DISTINCT user_id) as total_players')
+                ->selectRaw('COUNT(DISTINCT CASE WHEN unlocked_hardcore_at IS NOT NULL THEN user_id END) as hardcore_players')
+                ->first();
+
+            $playersTotal = $playerCounts->total_players ?? 0;
+            $playersHardcore = $playerCounts->hardcore_players ?? 0;
+        }
```
Jamiras, line 34: This doesn't handle the case where the player counts should be inherited from a parent set. The comment implies this is only a problem in tests, so it might not warrant the extra logic.

If I proceed to break out the player count updating logic into a separate job as we discussed earlier, this should just call into that action.

Jamiras (commented):
I have not actually tested any of this. I think you might be trying to over-optimize this part of the code at the cost of readability/maintainability. We should probably pursue the option of avoiding the queries when they're not needed (move player count and beaten updates into separate actions), rather than do them all the time and make them faster.

PR #3573, opened 2025-05-30
perf(UpsertGameCoreAchievementSetFromLegacyFlagsAction): batch sync

wescopeland:
This PR tackles another cause of N+1 writes in `UpdateGameMetricsAction`. Eloquent ORM's `sync()` kicks off an individual query for every entity to be synced.

Now, they're all rolled up into an `->upsert()`.


Jamiras (approved):
This does seem to be about three times faster than the old code, but it's the smallest of the bottlenecks that we identified.

PR #3581, opened 2025-05-31
perf(UpdateAchievementMetricsAction): bulk write

wescopeland:
Reimplements https://github.com/RetroAchievements/RAWeb/pull/3572.

* We're no longer using `DB::statement()` with bindings.
* We're no longer doing dirty checking with `syncOriginal()`.

The only intent here is to lift `$achievement->saveQuietly();` to a single bulk write to mitigate network overhead between Horizon and the DB.

### DIFF
--- a/app/Platform/Actions/UpdateAchievementMetricsAction.php
+++ b/app/Platform/Actions/UpdateAchievementMetricsAction.php
@@ -77,8 +77,10 @@ class UpdateAchievementMetricsAction
                 + $achievement->points * (($playersHardcoreCalc / $unlocksHardcoreCalc) * $weight)
             );
 
-            $unlockPercentage = $playersTotal ? $unlocksCount / $playersTotal : 0;
-            $unlockHardcorePercentage = $playersHardcore ? $unlocksHardcoreCount / $playersHardcore : 0;
+            // Round percentages to 9 decimal places to match the exact database column precision (decimal(10,9)).
+            // This prevents unnecessary updates due to precision differences in PHP.
+            $unlockPercentage = round($playersTotal ? $unlocksCount / $playersTotal : 0, 9);
+            $unlockHardcorePercentage = round($playersHardcore ? $unlocksHardcoreCount / $playersHardcore : 0, 9);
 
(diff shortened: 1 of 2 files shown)


app/Platform/Actions/UpdateAchievementMetricsAction.php
```diff
@@ -62,16 +77,25 @@ public function update(Game $game, Collection $achievements): void
...
-            $achievement->unlock_percentage = $playersTotal ? $unlocksCount / $playersTotal : 0;
-            $achievement->unlock_hardcore_percentage = $playersHardcore ? $unlocksHardcoreCount / $playersHardcore : 0;
+            $bulkUpdates[] = [
+                'ID' => $achievement->ID,
+                'unlocks_total' => $unlocksCount,
+                'unlocks_hardcore_total' => $unlocksHardcoreCount,
+                'unlock_percentage' => $playersTotal ? $unlocksCount / $playersTotal : 0,
+                'unlock_hardcore_percentage' => $playersHardcore ? $unlocksHardcoreCount / $playersHardcore : 0,
+                'TrueRatio' => $pointsWeighted,
+            ];
```
Jamiras, line 87: Did you not want to have the dirty checks? It's possible that this action could be called in such a way as no updates would occur.

wescopeland: Added dirty checks in latest.

app/Platform/Actions/UpdateAchievementMetricsAction.php
```diff
@@ -62,14 +77,34 @@ public function update(Game $game, Collection $achievements): void
                 + $achievement->points * (($playersHardcoreCalc / $unlocksHardcoreCalc) * $weight)
             );
 
+            $unlockPercentage = $playersTotal ? $unlocksCount / $playersTotal : 0;
+            $unlockHardcorePercentage = $playersHardcore ? $unlocksHardcoreCount / $playersHardcore : 0;
```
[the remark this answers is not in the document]
wescopeland: Updated in latest.

Jamiras (approved):

PR #3585, opened 2025-05-31
perf(UpdatePlayerPointsStats): low-hanging fruit optimizations

wescopeland:
Just a few quick wins that probably won't make a huge difference. A follow-up PR will have meatier changes, but I'd like to get this noise out of the way in isolation.

This PR:
* Prevents an `update()` call in `UpdatePlayerPointsStatsAction` if the existing value matches the new value.
* Removes `PlayerPointsStatsUpdated`. This is likely the more consequential change. These events are broadcast, and nothing is listening for them. At scale, it's a lot of I/O overhead: 30,000-60,000 dispatches per hour which instantiate objects that are immediately garbage collected.


app/Platform/Actions/UpdatePlayerPointsStatsAction.php
```diff
@@ -145,11 +144,14 @@ private function writePlayerPointsStat(
         $existingPlayerStat = PlayerStat::where($attributes)->first();
 
         if ($existingPlayerStat) {
+            // Skip the update if the value hasn't changed.
+            if ($existingPlayerStat->value === $points) {
+                return;
+            }
+
```
Jamiras: Just curious: Why is/was this using `update` instead of setting the value and calling `save`, which has a dirty check?

If the point of using `update` is to prevent the timestamp modifications, your new test is invalid because the `updated_at` would never change.

Jamiras (approved):

app/Platform/Actions/UpdatePlayerPointsStatsAction.php
wescopeland: I actually thought `save()` always wrote to the DB, even if stuff wasn't dirty. Given it has a built-in dirty check, that's much cleaner than using `update()`. Changed in latest.

I believe this means the new test case is still valid: the `updated_at` timestamp indeed should not change if the value is non-dirty, as we don't want the action to make any write to the DB.

PR #3586, opened 2025-05-31
perf: add player-points-stats-batch queue

wescopeland:
This PR dramatically improves the performance of hourly recurring player points stats syncs.

**How it works:**

This PR's optimization replaces the tens of thousands of individual queue jobs with a few dozen batch jobs (processing 200 users each). While the total database work remains the same, this is significantly faster in a Horizon context.

Each individual job has I/O overhead, and this approach eliminates 90%+ of the overhead. Additionally, bulk queries are leaned in to in order to mitigate network round trips as much as possible. Horizon workers spend less time polling the queue and more time doing actual work.

The implementation maintains the same logic while introducing:
- `UpdatePlayerPointsStatsBatchJob` for processing user batches.
- `UpdatePlayerPointsStatsBatchAction` for bulk operations.
- `CalculatesPlayerPointsStats` trait to share logic between the individual and batch processing actions.

**How to test this PR:**
You may need to clear existing points stats in your DB first:
```sql
DELETE FROM player_stats WHERE type LIKE 'points_%';
```
```bash
sail artisan horizon # make sure it's a fresh instance
```
Trigger a few manual unlocks for a few different users so you have some daily/weekly stuff.
```bash
sail artisan ra:platform:player:update-points-stats
```
You can verify daily/weekly points stats are created correctly in your DB (or if you're following the users you unlocked stuff for via your user profile sidebar component).

On my local, doing ~10,000 writes went from taking 2 minutes to taking about 15 seconds.


app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        if (empty($userIds)) {
+            return;
+        }
+
+        // Get users and handle untracked users.
+        $users = User::whereIn('id', $userIds)->get()->keyBy('id');
+
+        // Wipe stats for untracked users.
+        $untrackedUserIds = $users->where('Untracked', true)->pluck('id')->toArray();
```
Jamiras: Should this use `unranked_at`? Or possibly even a scope - I think we have a scope for tracked users.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
```diff
@@ -0,0 +1,127 @@
...
+     * @param array<int> $userIds
+     */
+    public function execute(array $userIds): void
+    {
+        if (empty($userIds)) {
+            return;
+        }
+
+        // Get users and handle untracked users.
+        $users = User::whereIn('id', $userIds)->get()->keyBy('id');
```
Jamiras: How much overhead is there to `keyBy`? It doesn't look like you're using the key for anything.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        $trackedUserIds = $trackedUsers->pluck('ID')->toArray();
+
+        // Fetch all player achievements for tracked users in the time window.
+        $allAchievements = PlayerAchievement::whereIn('player_achievements.user_id', $trackedUserIds)
+            ->whereBetween('player_achievements.unlocked_at', [Carbon::now()->subDays(8), Carbon::now()])
+            ->join('Achievements', 'player_achievements.achievement_id', '=', 'Achievements.ID')
+            ->join('GameData', 'Achievements.GameID', '=', 'GameData.ID')
+            ->where('GameData.ConsoleID', '!=', System::Events)
```
Jamiras: Suggest `whereNotIn('GameData.ConsoleID', System::getNonGameSystems())`
or `->whereHas('GameData.ConsoleID', fn ($q) => $q->gameSystems())`

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
```diff
@@ -0,0 +1,127 @@
...
+            ->select(
+                'player_achievements.user_id',
+                'player_achievements.unlocked_at',
+                'player_achievements.unlocked_hardcore_at',
+                'Achievements.Points as points',
+                'Achievements.TrueRatio as points_weighted'
+            )
+            ->get()
+            ->groupBy('user_id');
```
Jamiras, line 62: Should this be `orderBy`? You don't have any aggregate columns in your select.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
```diff
@@ -0,0 +1,127 @@
...
+                $hardcorePoints = $this->calculatePointsForInterval($hardcoreAchievements, $interval);
+                $softcorePoints = $this->calculatePointsForInterval($softcoreAchievements, $interval);
+
+                $periodStats = $this->buildStatsForPeriod($userId, $hardcorePoints, $softcorePoints, $period);
+
+                // Add timestamps for bulk insert.
+                foreach ($periodStats as $stat) {
+                    $bulkStats[] = array_merge($stat, [
+                        'created_at' => now(),
+                        'updated_at' => now(),
```
Jamiras: Not sure how much overhead `now()` has, but consider calculating it once outside the outer loop.

app/Platform/Jobs/UpdatePlayerPointsStatsBatchJob.php
```diff
@@ -0,0 +1,52 @@
...
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    /**
+     * @param array<int> $userIds
+     */
+    public function __construct(
+        private readonly array $userIds,
+        private readonly ?string $mockCurrentDate = null, // for testing only
```
Jamiras, line 29: If this just results in a call to `Carbon::setTestNow`, can't you just do that in the test instead of making it a parameter to the job?

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
wescopeland: We seem to have _two_ scopes for tracked users (see `User` and `ActsAsPlayer`). I'm not sure how Laravel reconciles them.

Unfortunately, we can't use a scope here because it's already a standard `Collection` at this point. But we should indeed be filtering by `unranked_at`. Updated in latest.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
wescopeland: This is leftover cruft from prior tinkering. Removed in latest.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
wescopeland: Good point. I like the `whereNotIn()` more, so I went with that in latest.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
wescopeland: I'm pretty sure `groupBy()` is correct here. The key (ambiguous) thing to observe is it's placed after `get()` -- it's not running within the query itself. It's being used as a `Collection` method.

The `groupBy()` is transforming the flat collection of achievements into a nested structure. This is necessary because we later grab the achievements for a given user ID:
```php
$userAchievements = $allAchievements->get($userId, collect());
```

Without the `groupBy()`, we've probably have to spend compute filtering through the entire collection for each user. With `groupBy()`, it's O(1) per each user.

app/Platform/Actions/UpdatePlayerPointsStatsBatchAction.php
wescopeland: Agreed. Updated in latest.

app/Platform/Jobs/UpdatePlayerPointsStatsBatchJob.php
wescopeland: Not necessarily. It remains possible to mock the date in the `UpdatePlayerPointsStats` command itself. This was used for manual testing purposes while originally writing the command.

Granted, I'm not sure how desirable it is to retain this behavior, and testing it did seem to cause weird things to happen in my Horizon instance.

Jamiras (approved):

PR #3592, opened 2025-06-01
ensure players_total gets intialized

Jamiras:
also initializes other nullable counts that were previously being set to 0 (I assume this has to be the result of #3558, but I can't find anything explicitly setting the values to 0 [there](https://github.com/RetroAchievements/RAWeb/pull/3558/files#diff-6689b0b689dabb410c1e2cd5db0c475fa05e5d538d56b13e66251f5064a214dcL693-L709)).

these should probably be made not nullable with a default of 0.

### DIFF
--- a/app/Connect/Actions/SubmitGameTitleAction.php
+++ b/app/Connect/Actions/SubmitGameTitleAction.php
@@ -98,7 +98,14 @@ class SubmitGameTitleAction extends BaseAuthenticatedApiAction
                 $game = $release->game;
             } else {
                 // no title match, it's a new game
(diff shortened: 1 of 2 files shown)


app/Connect/Actions/SubmitGameTitleAction.php
wescopeland, line 1: Anything we should cover in the tests? Seems like it may be easy to regress.

Jamiras: We could definitely add checks to ensure the values are 0, but if null is not a valid value, we really should just make the column not nullable.

wescopeland: That's a good point.

PR #3593, opened 2025-06-02
automatically expire/extend claims

Jamiras:
Extends the logic added in #3375 to automatically extend a claim that's In Review or with a Release Scheduled status instead of sending out the claim expiring warning email at 7 days prior to expiration ([decision reference](https://discord.com/channels/476211979464343552/1002694942826704936/1377651976938061845)).

Additionally, any overdue claims are now automatically dropped.


app/Platform/Actions/SendClaimExpirationWarningEmailsAction.php
Jamiras, line 1: This should probably be renamed now, but I wasn't sure if "ProcessExpiringClaimsAction" was too generic, or if the command (`ra:platform:claim:notify-expiry`) also needed to be changed.

wescopeland: Given the significant change in functionality, I agree this should be renamed.

This is doing much more now than sending emails. I think your idea of `ProcessExpiringClaimsAction` is appropriate. The command name could also change to `ra:platform:claim:process-expiring`.

wescopeland (approved):

app/Platform/Actions/SendClaimExpirationWarningEmailsAction.php
Jamiras: done

PR #3596, opened 2025-06-02
perf(UpdateDeveloperContributionYieldAction): use faster queries

wescopeland:
This query change yields the same results and runs ~20% faster on my local.

This job is now the largest bottleneck in Horizon, so trimming its runtime a bit should help.

**Before**
```sql
SELECT
  SUM(a.Points) AS author_points,
  COUNT(*) AS author_count
FROM
  Achievements a
  JOIN player_achievements pa ON pa.achievement_id = a.ID
  LEFT JOIN achievement_maintainer_unlocks amu ON amu.player_achievement_id = pa.id
WHERE
  a.user_id = -- user id here
  AND a.Flags = 3
  AND pa.user_id != -- user id here
  AND amu.id IS NULL;
```

**After**
```sql
SELECT
  SUM(a.Points) AS author_points,
  COUNT(*) AS author_count
FROM
  Achievements a
  JOIN player_achievements pa ON pa.achievement_id = a.ID
WHERE
  a.user_id = -- user id here
  AND a.Flags = 3
  AND pa.user_id != -- user id here
  AND NOT EXISTS (
    SELECT
      1
    FROM
      achievement_maintainer_unlocks amu
    WHERE
      amu.player_achievement_id = pa.id
  );
```

Also fixes a bug where maintainer unlocks weren't having their `Flags` filter set.

### DIFF
--- a/app/Platform/Actions/UpdateDeveloperContributionYieldAction.php
+++ b/app/Platform/Actions/UpdateDeveloperContributionYieldAction.php
@@ -39,11 +39,14 @@ class UpdateDeveloperContributionYieldAction
                 COUNT(*) as author_count
             FROM Achievements a
             JOIN player_achievements pa ON pa.achievement_id = a.ID
-            LEFT JOIN achievement_maintainer_unlocks amu ON amu.player_achievement_id = pa.id
             WHERE a.user_id = :user_id
                 AND a.Flags = :flags
                 AND pa.user_id != :user_id2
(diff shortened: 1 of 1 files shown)


wescopeland:
Note for future work: the more I think about it, the more I believe real-time dev stats calculation is undesirable. We could probably batch this stuff similar to how we do game/achievement Scout indexing and dramatically save on compute.

PR #3597, opened 2025-06-03
feat: launch search 2.0

wescopeland:
You'll need to resync your scout indexes and re-import hubs:
```bash
sail artisan horizon

sail artisan scout:import "App\Models\GameSet"
sail artisan scout:import "App\Models\Achievement"
```

---

This PR exposes Search 2.0 in the site navbar. The search bar has been converted to a "React island" -- a React component that is mounted from within Blade PHP.

### The Search Bar

Screenshot 2025-06-02 at 9 32 10 PM

You will immediately notice the meta key hint. This is a common pattern used to open a global search component. It's also used by [our docs site](https://docs.retroachievements.org/) as well as Discord.

The meta keybind is smart enough to detect the user's OS, and it will adjust accordingly:
Screenshot 2025-06-02 at 9 31 08 PM

At the XS breakpoint, the keybind is not shown:
Screenshot 2025-06-02 at 9 33 50 PM

Note the meta keybind is intentionally differentiated from the local search keybind: `/`.

Mirroring other global search experiences, the search bar is not an input field at all. It's a button cleverly-disguised as an input field. On click (or by using the global keybind), the Search 2.0 dialog appears.

### Search Dialog, Empty State

Screenshot 2025-06-02 at 9 37 34 PM

The search dialog fully supports point-and-click, full keyboard navigation, as well as touch. The design is heavily-inspired by our docs site's Algolia search component, an industry standard.

By default, the user's search query will try to match games, hubs, users, and achievements simultaneously. Alternatively, the user can click on one of those "scopes" and get more results for that specific scope.

Clicking the "Browse" link takes the user to the existing legacy _searchresults.php_ page, which still uses SQL searches.

### Search Dialog, Basic Searches

Screenshot 2025-06-02 at 9 41 07 PM

The component tries to sort the sections based on how relevant the back-end thinks they are to the user's query. This isn't perfect, but it's probably passable.

Under the "All" scope, each section is limited in the number of results it may contain.

Screenshot 2025-06-02 at 9 42 29 PM

Screenshot 2025-06-02 at 9 42 49 PM

If a user was recently active, we tack on a green dot over their avatar:
Screenshot 2025-06-02 at 9 43 06 PM

### Search Dialog, Scoped Searches

When selecting a specific scope, more results for that scope are displayed:
Screenshot 2025-06-02 at 9 43 48 PM

### Mobile

On mobile, the dialog is full screen and fully compatible with touch:
Screenshot 2025-06-02 at 9 44 33 PM

### Future Considerations
* We should probably also index and surface events as a search entity. I will open a follow-up PR to tackle this.
* **Test using the search component on both Inertia and Blade pages.**
* Loading states and animations could be greatly improved. This can probably be tackled in a follow-up PR.
* Be sure to **play around with keyboard navigation** - I find it to be a lot of fun to use.


app/Enums/UserPlatform.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * This is distinct from the `Platform` entities we hold in the
+ * DB. Those entities are associated with emulators. This enum
+ * is used for pattern-matching against the user's browser
+ * user agent so we can concretely trigger certain behaviors
+ * conditionally based on whatever we detect.
+ */
+#[TypeScript]
+enum UserPlatform: string
```
Jamiras: Would `ClientPlatform`, `ClientOS`, or `UserOS` work better?

resources/js/common/components/GlobalSearch/components/HelperFooter/HelperFooter.tsx
```diff
@@ -0,0 +1,43 @@
...
+              <kbd className={kbdClassNames}>{'↵'}</kbd>
+              {t('Select')}
+            </span>
+
+            <span className="flex items-center gap-1 text-neutral-500">
+              <kbd className={kbdClassNames}>{'esc'}</kbd>
+              {t('Close')}
+            </span>
+          </div>
+        </div>
```
Jamiras, line 37: Since you've gone through the trouble of adding keyboard navigation, should left/right switch between the filters?

resources/js/common/components/GlobalSearch/components/SearchResults/SearchResults.tsx
```diff
@@ -0,0 +1,194 @@
...
+    // Otherwise, use logical default ordering.
+    const defaultOrder: Record<string, number> = {
+      games: 1,
+      hubs: 2,
+      users: 3,
+      achievements: 4,
+    };
+
+    return (defaultOrder[a.key] as number) - (defaultOrder[b.key] as number);
+  });
```
Jamiras, line 171: The grouping for "All" is inconsistent. If I search on "2025", I get Hubs, Users, Games, Achievements. 

Searching on "crap" returns Users, Achievements and Games. 

If I search on "tele", I get Games, Hubs, Users, Achievements. If it was prioritizing based on match weight, I'd expect the "tele" user to be the first result, but Games and Hubs appear first.

resources/js/common/components/GlobalSearch/components/SearchResults/SearchResults.tsx
```diff
@@ -0,0 +1,194 @@
...
+      limit: 6,
+      icon: FaGamepad,
+
+      render: (game) => (
+        <BaseCommandItem
+          key={`game-${game.id}`}
+          className="group"
+          onSelect={() => {
+            window.location.assign(`/game/${game.id}`);
+            onClose();
```
Jamiras: There's no URL preview when hovering over a link. While most users probably don't care, it helps to explain why some records are being returned by ID when searching on numbers.
image
I search for "2025" a lot to get to the event pages. Here's the top 6 results I see:
image
The first two make sense, but the next four are all matches because of their IDs: 20255, 20259, 20251, and 20252.

resources/js/common/components/GlobalSearch/components/SearchResults/AchievementResultDisplay/AchievementResultDisplay.tsx
Jamiras, line 1: Another thing that's hard to reason is when an achievement description matches. Searching on "Jamiras" returns two users and three achievements. All three achievements are matched by description. Can we get card-style tooltips on the results?

resources/js/common/components/GlobalSearch/components/SearchModeSelector/SearchModeSelector.tsx
Jamiras, line 1: Searching on "Jamiras" returns 5 records (two users and three achievements). Switching from "All" to "Achievements" finds 10 achievements. 

If the expectation is to return 10 items, should the "All" list have included 8 achievements? 

If I search on "Quest", I get back 16 results. What is the expectation for returned items?

resources/js/common/components/GlobalSearch/components/SearchResults/SearchResults.tsx
Jamiras: There's no way to open a result in a new tab. Middle-click does nothing. Shift+click just opens it in the current tab, and there's no "Open in new tab/window" option on right click.

Jamiras (commented):
I really like the presentation, but I've noted several questions about the behavior. I have not reviewed the code.

app/Enums/UserPlatform.php
wescopeland: I like `UserOS` the most - renamed in latest.

resources/js/common/components/GlobalSearch/components/HelperFooter/HelperFooter.tsx
wescopeland: Good question. I don't think it's a bad idea. It may be an accessibility violation though, which is why I've held off for now.

The reason why is in the context of the dialog, the browser already prescribes native functionality to the left/right keys, namely when the input field has focus. Virtually always when you're using up/down keys, the input field will still have focus. If we add listeners for left/right, the user may not be able to update their cursor position in the input field.

resources/js/common/components/GlobalSearch/components/SearchResults/SearchResults.tsx
wescopeland: I'm not sure the best way to solve this problem, or if the problem is even solvable.

Though Meilisearch itself supports multi-index searches, Laravel Scout [does not](https://github.com/laravel/scout/issues/745#issuecomment-1586826674). We're really making four (now five, as of latest) searches and then having the back-end make a guess on the user's query intent. That guess is imperfect.

For example, it seems to make sense that if my search query is "tele", I'd see users as the first section in the UI. However, if we implement that weighting, we actually make a high number of queries less relevant. Searching for "zelda" is obviously a game or series hub search, but now the user with the username "zelda" would be weighted more heavily.

I still believe this behavior is better than what happens on prod currently. On prod, tele isn't a result at all:
Screenshot 2025-06-04 at 7 53 33 PM

A future improvement might be to explore if it's possible to tap in to Meilisearch's multi-indexing capability without the Scout abstraction. This would put Meilisearch in charge of figuring out the user's search query intent. However, after a cursory glance, implementing something like this appears to be non-trivial.

resources/js/common/components/GlobalSearch/components/SearchResults/SearchResults.tsx
wescopeland: Fixed in latest.

resources/js/common/components/GlobalSearch/components/SearchResults/AchievementResultDisplay/AchievementResultDisplay.tsx
wescopeland: Unless you feel strongly, I want to avoid the tooltip cards on these results. The reason why is they would technically be modals within modals.

I actually think it may be more desirable to remove `description` from the `Achievement` search index. I suspect after seeing your feedback on this that including `description` in the index is ultimately going to lead to less predictable and relevant achievement search results.

I've made this change in latest. As usual, it requires:
```bash
sail artisan horizon # fresh

sail artisan scout:import "App\Models\Achievement"
```

resources/js/common/components/GlobalSearch/components/SearchModeSelector/SearchModeSelector.tsx
wescopeland: In "All" mode, the UI adds a limit to the number of results returned.

A maximum of 6 games, 4 hubs, 4 events, 3 achievements, and 3 users: 20 results total.

When going out of "All" mode, up to 10 results of that kind of entity can be shown.

resources/js/common/components/GlobalSearch/components/SearchResults/SearchResults.tsx
wescopeland: Resolved in latest.

wescopeland:
As of latest, events are also now searchable.

Jamiras (approved):

PR #3603, opened 2025-06-05
use eloquent for API_GetGameInfoAndUserProgress response

Jamiras:
addresses https://github.com/RetroAchievements/RAWeb/pull/3538#discussion_r2106394875

### DIFF
--- a/app/Filament/Resources/LeaderboardResource.php
+++ b/app/Filament/Resources/LeaderboardResource.php
@@ -23,6 +23,7 @@ use Filament\Tables\Table;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Auth;
 
 class LeaderboardResource extends Resource
 {
@@ -114,6 +115,9 @@ class LeaderboardResource extends Resource
 
     public static function form(Form $form): Form
     {
+        /** @var User $user */
+        $user = Auth::user();
+
         return $form
             ->schema([
                 Forms\Components\Section::make('Primary Details')
@@ -123,15 +127,18 @@ class LeaderboardResource extends Resource
                         Forms\Components\TextInput::make('Title')
                             ->required()
                             ->minLength(2)
-                            ->maxLength(255),
+                            ->maxLength(255)
+                            ->disabled(!$user->can('updateField', [$form->model, 'Title'])),
 
                         Forms\Components\TextInput::make('Description')
-                            ->maxLength(255),
+                            ->maxLength(255)
+                            ->disabled(!$user->can('updateField', [$form->model, 'Description'])),
 
                         Forms\Components\TextInput::make('DisplayOrder')
                             ->numeric()
                             ->helperText("If set to less than 0, the leaderboard will be invisible to regular players.")
-                            ->required(),
+                            ->required()
+                            ->disabled(!$user->can('updateField', [$form->model, 'DisplayOrder'])),
                     ]),
 
                 Forms\Components\Section::make('Rules')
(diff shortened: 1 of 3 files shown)


public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -24,9 +24,10 @@
  *    int        ID                       unique identifier of the achievement
  *    string     Title                    title of the achievement
  *    string     Description              description of the achievement
- *    string     Points                   number of points the achievement is worth
- *    string     TrueRatio                number of RetroPoints ("white points") the achievement is worth
+ *    int        Points                   number of points the achievement is worth
+ *    int        TrueRatio                number of RetroPoints ("white points") the achievement is worth
```
Jamiras, line 28: These were already being returned as `int`s. Just updated the documentation.

public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -24,9 +24,10 @@
...
  *    string     BadgeName                unique identifier of the badge image for the achievement
+ *    string?    Type                     "progression", "win_condition", "missable" or null
```
Jamiras, line 30: This was being returned as `type` (lowercase) because we weren't explicitly excluding it. I've TitleCased it for consistency and documented it as officially exposed.

public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -55,83 +56,108 @@
...
+    'ImageIngame' => $game->ImageIngame,
+    'ImageBoxArt' => $game->ImageBoxArt,
+    'Publisher' => $game->Publisher,
+    'Developer' => $game->Developer,
+    'Genre' => $game->Genre,
+    'Released' => $game->released_at ? $game->released_at->format('Y-m-d') : null,
+    'ReleasedAtGranularity' => $game->released_at_granularity,
+    'IsFinal' => false,
+    'RichPresencePatch' => md5($game->RichPresencePatch),
+];
```
Jamiras, line 112: This explicitly exposes the data we want exposed without leaking anything else (like the aforementioned `type`). 

`sort_title` is also no longer exposed because of this change.

tests/Feature/Connect/PatchDataTest.php
```diff
@@ -383,6 +383,8 @@ public function testAchievementRarity(): void
 
     public function testUserAgent(): void
     {
+        Carbon::setTestNow(Carbon::now());
```
Jamiras, line 386: Unrelated to the actual issue, but addresses an intermittent failure when `now()` changes between initialization and assertions that I encountered when trying to push.

public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -55,83 +56,108 @@
...
-            $gameData['NumAwardedToUser']++;
-        }
-        if (isset($nextAch['DateEarnedHardcore'])) {
-            $gameData['NumAwardedToUserHardcore']++;
+        $achievements[$idStr]['DateEarned'] = $playerAchievement->unlocked_at->format('Y-m-d H:i:s');
+        if ($playerAchievement->unlocked_hardcore_at) {
+            $achievements[$idStr]['DateEarnedHardcore'] = $playerAchievement->unlocked_hardcore_at->format('Y-m-d H:i:s');
         }
     }
-}
```
Jamiras, line 112: These values are taken from the `player_game` aggregated metrics now.

public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -55,83 +56,108 @@
...
-unset($gameData['released_at_granularity']);
-unset($gameData['released_at']);
-unset($gameData['system']);
-unset($gameData['Updated']);
-
-$gameData['UserCompletion'] = '0.00%';
-$gameData['UserCompletionHardcore'] = '0.00%';
-if ($gameData['NumAchievements'] ?? false) {
-    $gameData['UserCompletion'] = sprintf("%01.2f%%", ($gameData['NumAwardedToUser'] / $gameData['NumAchievements']) * 100.0);
-    $gameData['UserCompletionHardcore'] = sprintf("%01.2f%%", ($gameData['NumAwardedToUserHardcore'] / $gameData['NumAchievements']) * 100.0);
```
Jamiras, line 129: Also taken from `player_game` aggregated metrics.

public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -55,83 +56,108 @@
...
+            'TrueRatio' => $achievement->TrueRatio,
+            'Type' => $achievement->type,
+            'BadgeName' => $achievement->BadgeName,
+            'NumAwarded' => $achievement->unlocks_total,
+            'NumAwardedHardcore' => $achievement->unlocks_hardcore_total,
+            'DisplayOrder' => $achievement->DisplayOrder,
+            'Author' => $achievement->developer->display_name,
+            'AuthorULID' => $achievement->developer->ulid,
+            'DateCreated' => $achievement->DateCreated->format('Y-m-d H:i:s'),
+            'DateModified' => $achievement->DateModified->format('Y-m-d H:i:s'),
```
Jamiras, line 140: format forced to match old format (which is the string format returned by the database).

wescopeland (approved):

PR #3604, opened 2025-06-05
perf(developer-metrics): introduce an increment/decrement action for simple unlocks and resets

wescopeland:
The majority of the server's compute time is now spent working on jobs in the `developer-metrics` queue:

Screenshot 2025-06-05 at 7 40 10 PM

Screenshot 2025-06-05 at 7 40 42 PM

This is slow because every achievement unlock recalculates the developer's entire `ContribYield` and `ContribPoints`, which is far more work than necessary. Achievement unlocks and resets can probably get by with naively incrementing or decrementing these stats. Full recalculations should only be necessary under rarer circumstances:

* An achievement's points value changes
* An achievement is promoted or demoted
* The maintainer of an achievement changes
* A player does a full reset of a game

This PR introduces said incremental updates for developer contribution yields. `IncrementDeveloperContributionYieldAction` performs O(1) increment/decrement operations. Simple achievement unlocks/resets now use this action rather than triggering a full calculation of dev stats.


app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
```diff
@@ -0,0 +1,101 @@
...
+        $pointsDelta = $achievement->Points * $delta;
+
+        // Store old values before incrementing.
+        $oldContribCount = $developer->ContribCount;
+        $oldContribYield = $developer->ContribYield;
+
+        // Update user stats incrementally.
+        $developer->ContribCount += $delta;
+        $developer->ContribYield += $pointsDelta;
+        $developer->saveQuietly();
```
Jamiras: Not doing a full recalculation introduces a race condition. Imagine a thousand players all playing the AotW. If even two or three of them unlock an achievement at the exact same time, the Contrib fields will be updated independently and only the last save will win.

You might be able to get away with a query of the old values, a direct update: `UPDATE UserAccounts SET ContribCount = ContribCount + $delta, ContribYield = ContribYield + $pointsDelta`, then another query of the updated values to see if a new badge is needed.

app/Platform/Actions/ResetPlayerProgressAction.php
```diff
@@ -83,10 +83,37 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
                 return;
             }
 
+            $achievement = $playerAchievement->achievement;
+
+            // Handle decrement for developer contribution credit before player_achievement deletion.
+            if ($achievement->isPublished && $achievement->Flags === AchievementFlag::OfficialCore->value) {
```
Jamiras: Isn't `isPublished` just a wrapper for `Flags == Core`?

app/Platform/Actions/ResetPlayerProgressAction.php
```diff
@@ -83,10 +83,37 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
...
+                    // Credit was given to the maintainer.
+                    $developer = User::find($maintainerUnlock->maintainer_id);
+                } else {
+                    // Credit was given to the original author.
+                    $developer = $achievement->developer;
+                }
+
+                if ($developer && $developer->id !== $user->id) {
+                    // Perform a quick incremental decrement.
+                    app(IncrementDeveloperContributionYieldAction::class)->execute(
```
Jamiras: What's the difference between
```
app(XXX:class)->execute()
```
and
```
(new XXX())->execute()
```
?

I see both in use in the application.

app/Platform/Actions/ResetPlayerProgressAction.php
```diff
@@ -136,14 +163,18 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
...
-            ->where(function ($query) use ($authorUsernames) {
-                $query->whereIn('User', $authorUsernames->unique())
-                    ->orWhereIn('display_name', $authorUsernames->unique());
-            })
-            ->get('ID');
-        foreach ($authors as $author) {
-            dispatch(new UpdateDeveloperContributionYieldJob($author->id));
+        // For game-wide or full resets, we need to do full recalculation of affected dev stats.
+        // For single achievement resets, we've already handled it incrementally above.
+        if ($achievementID === null) {
```
Jamiras, line 170: Would it make sense to do this infrequently? Maybe add an `|| rand(20) == 1` to do a full recalculation approximately 5% of the time?

app/Platform/Jobs/IncrementDeveloperContributionYieldJob.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public int $uniqueFor = 3600;
+
+    public function uniqueId(): string
+    {
+        return config('queue.default') === 'sync'
+            ? ''
+            : "{$this->developerId}-{$this->achievementId}-{$this->playerAchievementId}-{$this->isUnlock}";
+    }
```
Jamiras: Why does this need a unique ID? A player isn't likely going to ever unlock the same achievement twice. The only way I could see this generating a duplicate record is an earn immediately followed by a reset (or vice versa), in which case you'd want to do both the increment and the decrement.

app/Platform/Jobs/IncrementDeveloperContributionYieldJob.php
```diff
@@ -0,0 +1,78 @@
...
+        $playerAchievement = PlayerAchievement::find($this->playerAchievementId);
+        if (!$playerAchievement) {
+            return;
+        }
+
+        app()->make(IncrementDeveloperContributionYieldAction::class)->execute(
+            $developer,
+            $achievement,
+            $playerAchievement,
+            $this->isUnlock
```
Jamiras: Does this handle the case where a user upgrades a softcore unlock to hardcore? The contribution points will have already been assigned from the softcore unlock.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: I think having a healing mechanism like what you're proposing is a good idea, but I have mixed feelings about a `rand()`-style approach.

Ideally, healing shouldn't need to occur that much, and even at 5%, the amount of full calcs this would trigger seems disproportionate to the number of developers affected by these yield calculations.

It seems like it may be more efficient to have some kind of scheduled command, maybe that runs on a weekly basis at off-peak times, which triggers a full recalc for anyone with more than N published achievements. This would likely achieve the same healing effect but with far less and more predictable work.

I'm not sure how many unlocks we roughly process per day, but let's pretend it's 100,000. That'd be ~5,000 recalcs per day (average case).

```sql
SELECT COUNT(*) FROM UserAccounts WHERE ContribCount >= 10;
-- 839
```
With a scheduled command, we could achieve the same effect with ~97% less compute.

wescopeland:
```sql
SELECT
	COUNT(*) AS total_unlocks
FROM
	player_achievements
WHERE
	DATE(unlocked_at) = '2025-01-15'
	OR DATE(unlocked_hardcore_at) = '2025-01-15';
```
`total_unlocks` is 105,800. This is a sample size of 1, but I suspect the number is indeed close to 100,000 unlocks per day.

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
wescopeland: You're right. Thank you for the example scenario to help me better visualize.

In latest I've changed these to use direct `increment()` and `decrement()` calls, which should fix the race condition.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: Yep, removed in latest.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: Yea, it seems like in queues we use the `app(...)` syntax, and in other places we use the `(new ...)` syntax. In latest, I've changed this one over to the `(new ...)` syntax for consistency.

To the best of my knowledge, there's 1 subtle difference between the syntaxes. `(new XXX())->execute()` doesn't auto-inject any constructor dependencies (they have to be manually passed in), whereas the other syntax automagically does dependency injection.

app/Platform/Jobs/IncrementDeveloperContributionYieldJob.php
wescopeland: You're right. We should probably also remove `ShouldBeUniqueUntilProcessing`. I've removed that as well as the unique ID config in latest.

app/Platform/Jobs/IncrementDeveloperContributionYieldJob.php
wescopeland: Good catch, I forgot to consider this.

In latest, this is now handled properly. A new `$isHardcore` arg has been added to `IncrementDeveloperContributionYieldAction`, and we're able to see if it's an upgrade because the action already has a reference to the `PlayerAchievement` entity.

Additionally, I've added a few more test cases to verify this behavior.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: Weekly sync implemented in https://github.com/RetroAchievements/RAWeb/pull/3610. It uses `ContribCount > 0`, as the difference between 0 and 10 in the query and compute effort is negligible.

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
```diff
@@ -0,0 +1,115 @@
...
+                return;
+            }
+        }
+
+        $oldContribCount = $developer->ContribCount;
+        $oldContribYield = $developer->ContribYield;
+
+        if ($isUnlock) {
+            $developer->increment('ContribCount');
+            $developer->increment('ContribYield', $achievement->Points);
```
Jamiras: You might want to use [update with some raw math](https://stackoverflow.com/a/29816292) to avoid doing two queries.

app/Platform/Actions/ResetPlayerProgressAction.php
Jamiras: Did you mean to remove the `isPublished` check? Isn't it better to keep the helper function for readability?

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public function execute(
+        User $developer,
+        Achievement $achievement,
+        PlayerAchievement $playerAchievement,
+        bool $isUnlock = true,
+        bool $isHardcore = false
+    ): void {
+        // If we somehow made it here for an unofficial achievement, bail.
+        if ($achievement->Flags !== AchievementFlag::OfficialCore->value) {
```
Jamiras: This could check the `isPublished` helper too

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
```diff
@@ -0,0 +1,115 @@
...
+            return;
+        }
+
+        // Don't count the developer's own unlocks.
+        if ($playerAchievement->user_id === $developer->id) {
+            return;
+        }
```
Jamiras, line 36: Possible edge case: User resets an achievement they maintain but earned from the original author.

app/Platform/Actions/ResetPlayerProgressAction.php
```diff
@@ -83,10 +83,39 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
                 return;
             }
 
+            $achievement = $playerAchievement->achievement;
+
+            // Handle decrement for developer contribution credit before player_achievement deletion.
+            if ($achievement->Flags === AchievementFlag::OfficialCore->value) {
```
Jamiras: Another opportunity to use the `isPublished` helper

tests/Feature/Platform/Actions/IncrementDeveloperContributionYieldActionTest.php
```diff
@@ -0,0 +1,329 @@
...
+        // ... verify the badge was awarded ...
+        $badge = PlayerBadge::where('user_id', $developer->id)
+            ->where('AwardType', AwardType::AchievementPointsYield)
+            ->first();
+
+        $this->assertNotNull($badge);
+        $this->assertEquals(0, $badge->AwardData); // !! first tier (1000 points) is stored as 0
+    }
+
+    public function testItDoesNotDuplicateBadges(): void
```
Jamiras, line 188: May want to consider a case where the yield dips below the threshold via a reset and then crosses it again with an earn.

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
wescopeland: Good call, updated in latest.

I originally tried to use `incrementEach()` and `decrementEach()`, but this caused test failures that were difficult to reason about. The `DB` facade approach in your linked Stack Overflow thread worked.

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
wescopeland: Updated in latest.

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
wescopeland: I don't believe this edge case will occur with the way the action is currently being orchestrated.

`DispatchIncrementDeveloperContributionYieldJob` checks for an `AchievementMaintainerUnlock` record:
```php
$maintainerUnlock = AchievementMaintainerUnlock::query()
    ->where('player_achievement_id', $playerAchievement->id)
    ->first();
```
In this possible edge case scenario, since the maintainer earned the achievement before becoming a maintainer, they'd have no maintainer unlock record. This forces the code to fall into this block:
```php
$achievement->loadMissing('developer'); // Achievement.user_id (original author)
$developer = $achievement->developer;
```

Then, in the actual action, this conditional evaluates to false, so the `return` isn't invoked:
```php
if ($playerAchievement->user_id === $developer->id) {
    return;
}
```

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: Updated in latest.

tests/Feature/Platform/Actions/IncrementDeveloperContributionYieldActionTest.php
wescopeland: Additional test case to verify this scenario included in latest.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: I did (I'm most used to using the direct `Flags` comparison), but changed these all over to `isPublished` in latest.

Jamiras (approved):

PR #3609, opened 2025-06-06
fix(events): sanitize URLs properly for Plausible analytics

wescopeland:
Event URLs are not being sanitized properly (in their `u` param) when being sent to Plausible, making behavior analytics on event pages difficult to discern.

**Before**
Screenshot 2025-06-06 at 6 30 18 PM

**After**
Screenshot 2025-06-06 at 6 30 41 PM

### DIFF
--- a/app/Actions/ProcessPlausibleUrlAction.php
+++ b/app/Actions/ProcessPlausibleUrlAction.php
@@ -5,9 +5,11 @@ declare(strict_types=1);
 namespace App\Actions;
 
 use App\Models\Achievement;
+use App\Models\Event;
 use App\Models\Game;
 use App\Models\GameSet;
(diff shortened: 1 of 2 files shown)


Jamiras (approved):

wescopeland:
This is a good observation. Assuming I'm interpreting correctly, I believe this is normal/expected behavior.

One event is the pageview event, and the other is likely an engagement/behavior-tracking event. You may see a different result if you navigate directly to the award-earners page (via your URL bar) vs navigating there through the site UI.

Jamiras:
Is it still expected that the `u` parameter would be the same on the primary page and the subpage? Doesn't that skew the metrics for the primary page?

PR #3615, opened 2025-06-07
Show real beat/master times (when available)

Jamiras:
Surfaces some of the metrics calculated by #3496 on the player profile page and completion progress page.

If session duration has been calculated, the playtime will be reported as "Beaten in X"/"Mastered in X". If it hasn't been calculated yet, it will be reported as "Beaten over X". If the playtime is known and the "over X" is more than 50% larger than the playtime, then both will be shown: "Beaten in X over Y".

image
image

### DIFF
--- a/app/Platform/Services/PlayerProgressionService.php
+++ b/app/Platform/Services/PlayerProgressionService.php
@@ -6,6 +6,7 @@ namespace App\Platform\Services;
 
 use App\Community\Enums\AwardType;
 use App\Models\System;
+use App\Platform\Actions\GetAwardTimeTakenAction;
 use App\Platform\Enums\UnlockMode;
 
 class PlayerProgressionService
@@ -47,7 +48,7 @@ class PlayerProgressionService
         return $metrics;
     }
 
-    public function filterAndJoinGames(array $gamesList, array $siteAwards, bool $allowEvents = false): array
+    public function filterAndJoinGames(array $gamesList, array $siteAwards, int $userId, bool $allowEvents = false): array
     {
         /**
          * We need to append the most prestigious award kind+date to the game entities,
@@ -97,8 +98,10 @@ class PlayerProgressionService
         }
 
         $validConsoleIds = getValidConsoleIds();
+        $usedConsoleIds = [];
 
         // [B] Iterate once while appending the entities with constant time O(1).
+        $awardGames = [];
         $filteredAndJoined = [];
         foreach ($gamesList as &$game) {
             $canUseGame = (
@@ -108,6 +111,10 @@ class PlayerProgressionService
             );
 
             if ($canUseGame) {
+                if (!in_array($game['ConsoleID'], $usedConsoleIds)) {
+                    $usedConsoleIds[] = $game['ConsoleID'];
+                }
+
                 if (isset($awardsLookup[$game['GameID']])) {
                     $game['HighestAwardKind'] = $awardsLookup[$game['GameID']];
                     $game['HighestAwardDate'] = $awardsDateLookup[$game['GameID']];
@@ -121,6 +128,11 @@ class PlayerProgressionService
                     ) {
                         $game['HasNoAssociatedMasteryAward'] = true;
                     }
+
+                    if (!array_key_exists($game['HighestAwardKind'], $awardGames)) {
+                        $awardGames[$game['HighestAwardKind']] = [];
+                    }
+                    $awardGames[$game['HighestAwardKind']][] = $game['GameID'];
                 }
 
                 $filteredAndJoined[] = $game;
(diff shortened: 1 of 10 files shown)


resources/views/components/game-list-item/primary-meta.blade.php
```diff
@@ -51,14 +51,38 @@
...
 
-    $timeParts = [];
-    foreach ($timeUnits as $unit => $text) {
-        if ($datesDiff->$unit) {
-            $timeParts[] = "{$datesDiff->$unit} " . ($datesDiff->$unit === 1 ? $text : $text . 's');
+    if ($highestAwardTimeTaken) { // actual playtime (when available)
+        $datesDiff = CarbonInterval::seconds($highestAwardTimeTaken)->cascade();
+
+        $timeParts = [];
+        if ($datesDiff->totalHours > 0) { // don't report months/days for more than 24 hours of playtime
```
wescopeland: Wondering if there's potentially a logic error here, as the comment says checking for more than 24 hours, whereas the conditional has `->totalHours > 0`. If there isn't a logic error here, the comment may be ambiguous.

app/Platform/Services/PlayerProgressionService.php
```diff
@@ -151,10 +184,16 @@ public function filterAndJoinGames(array $gamesList, array $siteAwards, bool $al
                     && ($allowEvents ? true : $award['ConsoleID'] !== System::Events)
                     && in_array($award['ConsoleID'], $validConsoleIds)
                 ) {
+                    $system = $systems->where('id', $award['ConsoleID'])->first();
+                    if (!$system) {
+                        $system = System::find($award['ConsoleID']);
+                    }
```
wescopeland: When does this happen? We're in a loop here, so this could potentially be an N+1 query problem.

resources/views/components/game-list-item/primary-meta.blade.php
wescopeland, line 1: Looks good on desktop. I'm wondering if there's any way we can get it back to a single line on mobile though:
Screenshot 2025-06-10 at 3 19 12 PM

wescopeland, line 1: I'm going to see this sort of pattern on virtually every game I master:
Screenshot 2025-06-10 at 3 35 45 PM

I leave my mini PC's emulator running indefinitely because my settings get wiped every time I close a game.

I wonder if we can detect if playtime for a game exceeds 15 hours a day (or some threshold), and if so, switch to show only the calendar time.

resources/views/components/game-list-item/primary-meta.blade.php
Jamiras: I see how the comment could be misleading, but the behavior is correct. The idea is that anything that reports hours should _only_ report hours, not break it down into months/days/hours. I've updated the comment.

app/Platform/Services/PlayerProgressionService.php
Jamiras: I think the intent of loop [E] is to add beaten/mastery entries for games that have been reset.

The only way this line would get hit is if the award was for a game on a system where the user has no other awards. It probably isn't necessary at all, but I added it to ensure it didn't break the N+1 query problem being addressed here: https://github.com/RetroAchievements/RAWeb/pull/3615/files#diff-8e7403b0685f65e5dcb41463a81c886c5293d0f8211c3845b718983d915757b7L20-L25

I've pulled the logic out into a loop higher up to ensure we pre-fetch the data.

resources/views/components/game-list-item/primary-meta.blade.php
Jamiras: While we could potentially detect individual sessions exceeding 24 hours, that data isn't accessible anywhere along the current processing chain for these pages. The 404 hours is coming directly from your player_games record, and the 26 days is coming from subtracting the date of your first achievement earned for the set from the date on your SiteAward for the set.

26 days is 656 hours, which is just above the 1.5x threshold (404x1.5=606). Had it been under, it would have only reported playtime, which could have possibly been even more misleading.

I've reworked the heuristics for determining when to display "in" and "over" to better predict that when the game was left running overnight.

Now, there's a 36 hours threshold for using wall time. 
* Anything less than 36 hours of wall time will only report playtime (if playtime is available). 
  - Previously both were shown if wall time was more than 1.5x playtime so you could have things like "in 4 hours over 9 hours" - now it just says "in 4 hours".
* Anything more than 36 hours of wall time will only show wall time if playtime is more than half of wall time (this covers your scenario above because 404 > 328). If playtime is less than half of wall time, both will be shown.

Jamiras: I can't think of any good solutions. We'd either have to use shorter labels: "Mastered in 2h48m over 1y1m", or less labels: "Mastered in 3 hours over 13 months".

Even if we did either of those solutions, I imagine we'd run into issues with longer translations once these get internationalized.

wescopeland: I can't think of any good solutions either. Maybe we can revisit when this page finally gets converted to React.

wescopeland: This seems a lot better. Thanks 👍

PR #3616, opened 2025-06-07
chore: React warnings cleanup

wescopeland:
This PR resolves numerous warnings that are being logged:

```
stderr | resources/js/features/events/components/EventAwardTiers/AwardTierItem.test.tsx > Component: AwardTierItem > given no earn count, does not show a link to earners
An empty string ("") was passed to the href attribute. To fix this, either do not render the element at all or pass null to href instead of an empty string.
```
More empty string woes in `<AwardTierItem />`. Resolved by setting this to be `undefined` when there's no link and a type cast to `<InertiaLink />`'s `href` prop.

---

```
stderr | resources/js/features/game-list/components/SuggestionReasonCell/SuggestionReasonCell.test.tsx > Component: SuggestionReasonCell > given the suggestion reason is shared-author, renders the SharedAuthorReason component
Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.
```
Resolved by removing the `ref` prop from `SharedAuthorReason`. The `<Trans />` component gets really upset when this prop is used.

---

```
  52:5  warning  Object rest destructuring on a query will observe all changes to the query, leading to excessive re-renders  @tanstack/query/no-rest-destructuring
```
This has been resolved by using `Object.assign()` instead of a spread to have a stable object reference for React's context.

### DIFF
--- a/resources/js/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot.tsx
+++ b/resources/js/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot.tsx
@@ -36,7 +36,7 @@ export const EventAwardEarnersMainRoot: FC = memo(() => {
       />
       <PlayableHeader
         badgeUrl={eventAward.badgeUrl}
-        systemLabel={event.legacyGame?.title ?? 'Event'}
+        systemLabel={event.legacyGame!.title}
         systemIconUrl="/assets/images/system/events.png"
         title={cleanAwardLabel}
       >
--- a/resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
+++ b/resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
@@ -32,17 +32,9 @@ export const AwardTierItem: FC<AwardTierItemProps> = ({ event, eventAward, hasVi
     (ea) => ea.achievement?.points && ea.achievement.points === 1,
   );
 
-  const awardEarnersLink =
-    eventAward.badgeCount! < 1
-      ? ''
-      : eventAward.tierIndex > 0
-        ? route('event.award-earners.index', {
-            event: eventAward.eventId,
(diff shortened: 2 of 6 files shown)


resources/js/features/events/components/EventAwardEarnersMainRoot/EventAwardEarnersMainRoot.tsx
```diff
@@ -36,7 +36,7 @@ export const EventAwardEarnersMainRoot: FC = memo(() => {
       />
       <PlayableHeader
         badgeUrl={eventAward.badgeUrl}
-        systemLabel={event.legacyGame?.title ?? 'Event'}
+        systemLabel={event.legacyGame!.title}
```
wescopeland, line 39: The page will outright crash if it falls into the "else" clause. To avoid an uncovered branch warning, I used the `!` operator on `legacyGame` here.

resources/js/features/events/components/EventAwardTiers/AwardTierItem.tsx
```diff
@@ -106,7 +100,13 @@ export const AwardTierItem: FC<AwardTierItemProps> = ({ event, eventAward, hasVi
...
                   <div
                     data-testid="award-earned-checkmark"
-                    className="mr-1 flex size-6 items-center justify-center rounded-full bg-embed light:bg-neutral-200 light:text-neutral-700"
+                    className={cn([
+                      'mr-1 flex size-6 items-center justify-center rounded-full bg-embed',
+                      'light:bg-neutral-200 light:text-neutral-700',
+                      awardEarnersLink !== undefined
+                        ? 'transition group-hover:text-link-hover'
+                        : null,
+                    ])}
```
wescopeland, line 109: I discovered this check icon also needed the conditional styles applied while doing local testing of the changes.

Jamiras (approved):

PR #3618, opened 2025-06-08
fix(SearchApiController): make some minor group relevance tweaks

wescopeland:
This PR makes some minor adjustments to group relevance scores on multi-scoped ("All") searches.

Queries with multiple words now slant more heavily towards hubs and games. They also reduce the relevancy of the user results. Achievements and events are intentionally not touched by this change.

I suspect, though cannot prove, most users of global search are searching for games and hubs. I will add analytics in a subsequent PR so we can begin processing and understanding this user behavior data.

Without these changes, on my local, queries for "pokemon red" rank users as more relevant than games.

**Before**
Screenshot 2025-06-08 at 8 20 10 AM

**After**
Screenshot 2025-06-08 at 8 19 58 AM


Jamiras:
I wonder if it would make sense to get rid of "All" and just have the popup default to the "Games" filter?

Jamiras (approved):

PR #3634, opened 2025-06-14
fix(connect): tell Cloudflare that responses are immutable

wescopeland:
I've spent the last 6 hours trying to debug why `patch` responses aren't successfully resolving in RetroArch 1.20.0 and below.

In prod, I can reproduce the issue nearly 100% of the time with Super Mario World (89 achievements) on RetroArch 1.19.1.
In prod, I can almost never reproduce the issue on Marble Madness (GB, 6 achievements) on RetroArch 1.19.1.

When I point RetroArch to my local (by setting a `cheevos_custom_host` value in _retroarch.cfg_), I can never reproduce either issue, including when I introduce significant artificial delay on my local server.

After extensively testing the differences between the network calls of my local and prod, I discovered that for Super Mario World, my localhost:64000 server is sending a response that contains 20 chunks, with an average chunk size of 7,034 bytes. For the same achievement set, prod is sending 6 chunks, with an average chunk size of 23,491 bytes.

I dove through our nginx config and confirmed it's not nginx that's managing these chunk sizes. I'm not even sure if the chunk size is the problem, but it seems clear Cloudflare is mutating these responses in subtle ways.

I'd like to rule these mutations completely out. After doing some research, it seems you can set certain headers which inform Cloudflare that the response should be treated as immutable. This PR adds these headers to dorequest.php calls.


public/dorequest.php
```diff
@@ -840,13 +840,31 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
 
 $response['Success'] = (bool) $response['Success'];
 
+// Convert the response to a JSON string in order to calculate the exact Content-Length.
+// Cloudflare is manipulating the headers of dorequest.php responses, and some clients
+// are unable to gracefully handle this (ie: RetroArch 1.20.0 and below). By adding
+// explicit Content-Type, Content-Length, and Cache-Control headers, we inform Cloudflare
+// that these responses are immutable and should be passed straight through.
+$jsonContent = json_encode($response);
+$contentLength = (string) strlen($jsonContent);
```
Jamiras, line 847: I wonder if providing the `Content-Length` would be enough to fix the problem.

[RetroArch 1.21](https://github.com/libretro/RetroArch/pull/17460) increased the initial receive buffer size from 512 bytes to 16384. 

The socket read code reads enough to fill the buffer, then doubles the buffer size. So for a 320k response, the old code would have to read the socket at least 11 times: [512,1024,2048,4096,8192,16384,32768,65536,132072,262144,524288], whereas the new code would only have to read it at least 6 times (16384,...524288).  NOTE: I wouldn't expect the socket read function to return 128K at a time as the buffer gets larger, which is why I say "at least".

Unfortunately, even if a `Content-Length` is provided, I'm pretty sure that older RetroArch will still use the small-buffer doubling. That same PR linked above indicates that it now preallocates the exact amount specified by the `Content-Length`, so the buffer won't have to grow (avoids reallocations), and the reads can return much larger chunks at the beginning. So providing a `Content-Length` is probably for the better anyway.

I still don't know why the older RetroArch would be choking on the smaller reads - perhaps the socket is being flushed before everything can be pulled from it and the code just sits there waiting for the rest of the data that never arrives?

Or maybe the fact that it takes so many reads allows the server to disconnect before all the data has been received? I know we've been talking about the server keep-alive settings recently, but that's been _after_ this issue started occurring, right?

public/dorequest.php
Jamiras, line 1: I tried to reproduce on 1.19.1 with SMW on both my Windows machine and my Ubuntu VM and had 0/8 failures on both. I know the VS Windows build doesn't have SSL enabled, and I assume your localhost also doesn't use SSL. I'm not sure if that figures into the behavior at all.

Jamiras (approved):

public/dorequest.php
wescopeland: I can't reproduce the issue anymore either, no matter what I try. I'm really puzzled why the failure rate seems to be changing day by day.

public/dorequest.php
wescopeland: > I still don't know why the older RetroArch would be choking on the smaller reads - perhaps the socket is being flushed before everything can be pulled from it and the code just sits there waiting for the rest of the data that never arrives?
>
> Or maybe the fact that it takes so many reads allows the server to disconnect before all the data has been received? I know we've been talking about the server keep-alive settings recently, but that's been after this issue started occurring, right?

I'm still incredibly puzzled by the issue. It's not helping that I still can't repro it anymore :(

it seems to have permanently gone away for me after the recent outage recovery, though other users are still reporting it's happening for them. Let's see what happens post-deploy.

PR #3638, opened 2025-06-16
add action for achievementwondata connect api

Jamiras:
* Converted `getRecentUnlocksPlayersData` to eloquent.
* Added `Content-Length` header to response (as suggested in #3634).
* Changed the response for an unknown achievement ID to return 404 instead of success with no unlocks.

### DIFF
--- a/app/Connect/Support/BaseApiAction.php
+++ b/app/Connect/Support/BaseApiAction.php
@@ -26,16 +26,16 @@ abstract class BaseApiAction
 
     protected function buildResponse(array $result): JsonResponse
     {
-        if (array_key_exists('Status', $result)) {
-            $status = $result['Status'];
-            if ($status === 401) {
-                return response()->json($result, $status)->header('WWW-Authenticate', 'Bearer');
-            }
+        $status = $result['Status'] ?? 200;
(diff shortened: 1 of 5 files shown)


wescopeland (approved):

app/Connect/Support/BaseApiAction.php
```diff
@@ -26,16 +26,16 @@ public function handleRequest(Request $request): JsonResponse
...
-        if (array_key_exists('Status', $result)) {
-            $status = $result['Status'];
-            if ($status === 401) {
-                return response()->json($result, $status)->header('WWW-Authenticate', 'Bearer');
-            }
+        $status = $result['Status'] ?? 200;
+        $response = response()->json($result, $status);
 
-            return response()->json($result, $status);
+        $response->header('Content-Length', (string) strlen($response->getContent()));
```
wescopeland, line 32: https://retroachievements.org/forums/topic/31122?comment=294115#294115

Unfortunately, it appears adding the "Content-Length" header didn't remediate the issue.

Jamiras: I think including "Content-Length" is probably still in our best interests.

PR #3641, opened 2025-06-16
fix(leaderboard): remediate race condition on restored records

wescopeland:
https://discord.com/channels/310192285306454017/1383550440813101087/1384171929551900772

Set the `score` field _before_ calling `->restore()` to avoid a race condition with model observers.

### DIFF
--- a/app/Helpers/database/leaderboard.php
+++ b/app/Helpers/database/leaderboard.php
@@ -64,15 +64,16 @@ function SubmitLeaderboardEntry(
         } elseif ($existingLeaderboardEntry->trashed()
             || $leaderboard->isBetterScore($newEntry, $existingLeaderboardEntry->score)) {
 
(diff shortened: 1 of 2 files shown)


app/Helpers/database/leaderboard.php
```diff
@@ -64,15 +64,16 @@ function SubmitLeaderboardEntry(
...
             $existingLeaderboardEntry->score = $newEntry;
             $existingLeaderboardEntry->player_session_id = $playerSession->id;
             $existingLeaderboardEntry->updated_at = $timestamp;
-            $existingLeaderboardEntry->save();
+
+            if ($existingLeaderboardEntry->trashed()) {
+                $existingLeaderboardEntry->restore();
+            } else {
+                $existingLeaderboardEntry->save();
+            }
```
Jamiras, line 76: I didn't know `restore` also committed modified fields.

PR #3643, opened 2025-06-18
feat(game): add series hub component to React game pages

wescopeland:
This PR adds the dedicated "Series Hub" component to the React game page sidebar:

Screenshot 2025-06-17 at 9 09 56 PM

Implementation details:
* The component appears if the game is within either a series or subseries hub. If both, the subseries hub wins.
* "N games" represents the _total_ game count of the hub, even if the games have no associated achievement sets.
* The row of games that appears only includes games with achievement sets. The games are organized by earliest release date, and try to be in proximity of the current game's game (which is designated with the blue underline).
* Subset games are excluded from the totals counts.

Screenshot 2025-06-17 at 9 10 55 PM

Screenshot 2025-06-17 at 9 11 24 PM

Screenshot 2025-06-17 at 9 12 08 PM


.github/workflows/ci.yml
wescopeland, line 1: For whatever reason, the cached composer dependencies are now refusing to load in subsequent steps. I'm not sure if GitHub changed something on their end.

`php-setup` is only ~15s of overhead at worst, so getting rid of it is probably not a big deal.

app/Platform/Actions/BuildSeriesHubDataAction.php
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        // First, check for a subseries hub (these are more specific).
+        $subseriesHub = $hubs->first(fn ($hub) => str_contains($hub->title, 'Subseries -'));
+        if ($subseriesHub) {
+            return $subseriesHub;
+        }
+
+        // Then, check for a series hub.
+        $seriesHub = $hubs->first(fn ($hub) => str_contains($hub->title, 'Series -'));
```
Jamiras, line 66: Would it be desirable to add an `orderBy` to this? http://localhost:64000/game2/6531 has two series hubs:
image

Even if there isn't a way to specify the preferred series hub, an `orderBy` would at least provide consistency.

app/Platform/Actions/BuildSeriesHubDataAction.php
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        // Safety: if the current game wasn't found (shouldn't happen), just show the first 5.
+        if ($currentGameIndex === false) {
+            return $allGames->take($maxGamesToShow)->all();
+        }
+
+        // Try to center the current game in the list of 5.
+        // Ideal position is index 2 (middle of 0,1,2,3,4).
+        $idealStartIndex = $currentGameIndex - 2;
```
Jamiras: Logically, I understand the intent here, but visually it's confusing.
image

Why would Castlevania III be the first item in the Castlevania series? 

It's because there's 6 items before it that got discarded to center the selected item in the list. Would it make sense to split the "+31" into a "+6" on the left and a "+25" on the right?

It's particularly confusing when the game icons are all very similar.

app/Platform/Actions/BuildSeriesHubDataAction.php
```diff
@@ -0,0 +1,131 @@
...
+    {
+        $maxGamesToShow = 5;
+        $totalGames = $allGames->count();
+
+        // If we have 5 or fewer games total, show them all.
+        if ($totalGames <= $maxGamesToShow) {
+            return $allGames->all();
+        }
+
+        // Safety: if the current game wasn't found (shouldn't happen), just show the first 5.
```
Jamiras: This will actually happen somewhat frequently - if you go to a game that doesn't have achievements, it gets filtered out. Should you try to center around where it would be?

app/Platform/Actions/BuildSeriesHubDataAction.php
```diff
@@ -0,0 +1,131 @@
...
+
+        // Get all games sorted by release date.
+        $allGames = $seriesHub->games()
+            ->where('GameData.Title', 'not like', '%[Subset -%')
+            ->orderBy('GameData.released_at', 'asc')
+            ->orderBy('GameData.Title', 'asc') // Secondary sort for games with same release date.
+            ->get();
+
+        // Filter to only games with achievements for display (but keep all for counting).
+        $gamesWithAchievements = $allGames->filter(fn ($g) => $g->achievements_published > 0)->values();
```
Jamiras, line 36: If a hub only has a couple games with achievements, it still shows the "+N" instead of expanding the list:
image

Also, should the items in the series hub be removed from the Similar Games list? Or would that be more of a manual thing for QATeam?

app/Platform/Actions/BuildSeriesHubDataAction.php
```diff
@@ -0,0 +1,131 @@
...
+        $stats = $query->select([
+            DB::raw('COUNT(DISTINCT GameData.ID) as total_game_count'),
+            DB::raw('SUM(GameData.achievements_published) as total_achievements'),
+            DB::raw('SUM(GameData.points_total) as total_points'),
+        ])->first();
+
+        // Get all games sorted by release date.
+        $allGames = $seriesHub->games()
+            ->where('GameData.Title', 'not like', '%[Subset -%')
+            ->orderBy('GameData.released_at', 'asc')
```
Jamiras: Ordering by release date works well for simple series, but for things that are released on multiple systems, you end up with weird orderings where it's "original 3" "re-release 1" "original 4"
image

Jamiras:
You've provided some images, and I've provided several more. Maybe have a few people from the discord look at all the samples provided and ask how they would interact with the widget.

wescopeland:
I think I'm convinced at this point to remove the row altogether. Discord has overall been positive to the idea when I've previewed it, but this seems like mostly vibes.

This paragraph is what has me convinced:

> I think the biggest issue with the list is that except for FF and DQ that put the numerals in the badges, the icons within a series are all going to be very similar. As such, **any interaction with the list is going to be exploratory.** Player's aren't going to just click on the fourth icon knowing what they're getting into, they'll hover over it to see the tooltip, potentially click it, decide it's not what they were looking for, and go to the hub anyway.

I'll update this branch in a few minutes to remove the row & sorting/organizing algorithm.

app/Platform/Actions/BuildSeriesHubDataAction.php
wescopeland: Added an orderBy in latest. All of the orders I could think of have downsides. The "least bad" option is probably game count, but this isn't a strong opinion.

app/Platform/Actions/BuildSeriesHubDataAction.php
wescopeland: Removed the row altogether in latest.

app/Platform/Actions/BuildSeriesHubDataAction.php
wescopeland: Removed the row altogether in latest.

app/Platform/Actions/BuildSeriesHubDataAction.php
wescopeland: Removed the row altogether in latest.

app/Platform/Actions/BuildSeriesHubDataAction.php
wescopeland: Removed the row altogether in latest.

> should the items in the series hub be removed from the Similar Games list? Or would that be more of a manual thing for QATeam?

Now that the row has been removed, I'm not sure if we have to worry about this as much.

resources/js/features/games/components/SeriesHubDisplay/SeriesHubDisplay.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    components={{ 1: <span className="font-semibold" /> }}
+                  />
+                </span>
+              </div>
+
+              <p className="text-2xs">
+                {t('({{val, number}} with achievements)', {
+                  val: seriesHub.gamesWithAchievementsCount,
+                })}
+              </p>
```
Jamiras: Should this be hidden if all the games have achievements?
image

wescopeland: I wasn't too sure on this. UX research suggests it should be. Updated in latest.

PR #3645, opened 2025-06-18
feat(game): add 'Other Names' metadata row

wescopeland:
Adds a metadata row for unique non-canonical titles:

Screenshot 2025-06-18 at 7 12 33 PM

If there are 5 or more alt titles, only 3 are shown and the excess is buried in a tooltip. If there are exactly 4 alt titles, all are shown with no tooltip.

Screenshot 2025-06-18 at 7 32 07 PM

### DIFF
--- a/resources/js/features/games/components/GameMetadata/GameOtherNamesRow/GameOtherNamesRow.tsx
+++ b/resources/js/features/games/components/GameMetadata/GameOtherNamesRow/GameOtherNamesRow.tsx
@@ -1,0 +1,28 @@
+import type { FC } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import { BaseTableCell, BaseTableRow } from '@/common/components/+vendor/BaseTable';
+
+interface GameOtherNamesRowProps {
+  nonCanonicalTitles: string[];
+}
+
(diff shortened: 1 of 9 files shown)


resources/js/features/games/components/GameMetadata/GameOtherNamesRow/GameOtherNamesRow.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <BaseTableRow className="first:rounded-t-lg last:rounded-b-lg">
+      <BaseTableCell className="whitespace-nowrap text-right align-top">
+        {t('metaOtherName', { count: nonCanonicalTitles.length })}
+      </BaseTableCell>
+
+      <BaseTableCell>
+        <div className="flex flex-col">
+          {displayedTitles.map((title, titleIndex) => (
+            <span key={`title-${titleIndex}`}>{title}</span>
```
Jamiras, line 34: Should these titles be sanitized (or whatever it is we do to turn the psuedo-tags into tags)?
image

wescopeland: Those shouldn't be added to begin with. This is a data issue. I'll let the cleanup team know.

PR #3648, opened 2025-06-19
feat(game): add Want to Play/Develop toggle buttons to React game pages

wescopeland:
This PR adds the Want to Play / Want to Develop toggle buttons to React game pages.

The Want to Play toggle _always_ appears, even if the user is authenticated. This mirrors behavior on game datatables. If the user presses the button and is unauthenticated, they'll be redirected to the login page.

The Want to Develop toggle only appears if the user has `Role::DEVELOPER` attached to their account.

Screenshot 2025-06-19 at 6 42 28 PM


resources/js/features/games/components/GameHeaderSlotContent/GameHeaderSlotContent.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <BaseButton
+          onClick={() => toggleWantToDevelop()}
+          className="flex items-center gap-1 rounded-full !py-0 !text-xs"
+          size="sm"
+          aria-pressed={isOnWantToDevList}
+        >
+          {isOnWantToDevList ? <LuCheck className="size-4" /> : <LuPlus className="size-4" />}
+          {t('Want to Develop')}
+        </BaseButton>
+      ) : null}
```
Jamiras: Want to Develop feels out of place here - especially if the game already has a set. (Maybe the button should change to "Want to Revise" in that case?)

Jamiras: Where do you see the "Claim" button going? I think Want to Develop would make more sense near that. It  might also make sense to group those with the Set Requests component.

wescopeland: It may make the most sense to stick both "Want to Develop/Revise" and "Claim" over here instead:
Screenshot 2025-06-20 at 10 05 55 AM

That way we can have the main heading area only show stuff that isn't going to be conditional on a user-by-user basis.

Jamiras: A sidebar component would make sense. We could do something like this:

image

Where the last two would only be visible to developers. And the first two would disappear once a set exists.

The 3 marker is how many requests the set has, and the 4 marker would be how many requests the user has remaining.

wescopeland: Updated in latest. The Want to Develop button is now in the sidebar:
Screenshot 2025-06-21 at 8 36 51 AM

If the game already has a non-zero amount of achievements published, the label changes:
Screenshot 2025-06-21 at 8 39 02 AM

Jamiras (approved):

PR #3649, opened 2025-06-19
feat(api): add developer info to API_GetGameLeaderboards

wescopeland:
https://discord.com/channels/310192285306454017/645777658319208448/1385358230959423529

This PR adds two new fields to API_GetGameLeaderboards: `Author` and `AuthorULID`.


public/API/API_GetGameLeaderboards.php
```diff
@@ -77,6 +79,8 @@
         'Description' => $leaderboard->Description,
         'Format' => $leaderboard->Format,
         'TopEntry' => $topEntry,
+        'Developer' => $leaderboard->developer->display_name,
+        'DeveloperULID' => $leaderboard->developer->ulid,
```
Jamiras: API_GetAchievementUnlocks, API_GetAchievementOfTheWeek, and API_GetGameInfoAndUserProgress.php use `Author` and `AuthorULID`. API_GetTicketData uses `AchievementAuthor` and `AchievementAuthorULID`. It seems like we should use `Author` and `AuthorULID` here too.

wescopeland: Renamed these fields in latest.

Jamiras (approved):

PR #3650, opened 2025-06-19
perf: don't run scheduled commands in non-prod envs

wescopeland:
This PR ever-so-slightly reduces these batch sizes from 1,000 to 750.

Prod was unresponsive for 60s+ [earlier today](https://discord.com/channels/476211979464343552/1026273695133605888/1385286039744282708), with CPU briefly pinned at 100%.

There was a significant plateau overnight where CPU was also pinned:
Screenshot 2025-06-19 at 7 31 09 PM

I'm hoping lowering from 1,000 to 750 takes a little bit of the edge off.

### DIFF
--- a/app/Platform/AppServiceProvider.php
+++ b/app/Platform/AppServiceProvider.php
@@ -151,14 +151,15 @@ class AppServiceProvider extends ServiceProvider
             /** @var Schedule $schedule */
             $schedule = $this->app->make(Schedule::class);
 
+            $schedule->command(UpdateSearchIndexForQueuedEntities::class)->twiceDaily(1, 13); // 1AM and 1PM UTC
+            $schedule->command(DeleteStalePlayerPointsStatsEntries::class)->weekly();
+
             if (app()->environment() === 'production') {
                 $schedule->command(UpdateAwardsStaticData::class)->everyMinute();
                 $schedule->command(CrawlPlayerWeightedPoints::class)->everyFiveMinutes();
                 $schedule->command(BackfillPlaytimeTotal::class)->everyTenMinutes();
                 $schedule->command(UpdatePlayerPointsStats::class, ['--existing-only'])->hourly();
                 $schedule->command(ProcessExpiringClaims::class)->hourly();
-                $schedule->command(UpdateSearchIndexForQueuedEntities::class)->twiceDaily(1, 13); // 1AM and 1PM UTC
-                $schedule->command(DeleteStalePlayerPointsStatsEntries::class)->weekly();
                 $schedule->command(UpdateDeveloperContributionYield::class)->weeklyOn(2, '10:00'); // Tuesdays at 10AM UTC
             }
         });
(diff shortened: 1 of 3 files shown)


wescopeland:
I like that idea more. Set count back to 1000 in latest and added simple conditionals in `AppServiceProvider` files in latest.

app/Providers/AppServiceProvider.php
```diff
@@ -74,13 +74,15 @@ public function boot(): void
         $this->app->booted(function () {
             $schedule = $this->app->make(Schedule::class);
 
-            $schedule->command(LogUsersOnlineCount::class)->everyThirtyMinutes();
+            if (app()->environment() === 'production') {
+                $schedule->command(LogUsersOnlineCount::class)->everyThirtyMinutes();
```
Jamiras: We may want to keep this one since it drives the online players widget (which has no concept of when the counts were captured).

app/Platform/AppServiceProvider.php
```diff
@@ -151,14 +151,16 @@ public function boot(): void
...
-            $schedule->command(UpdateSearchIndexForQueuedEntities::class)->twiceDaily(1, 13); // 1AM and 1PM UTC
-            $schedule->command(DeleteStalePlayerPointsStatsEntries::class)->weekly();
-            $schedule->command(UpdateDeveloperContributionYield::class)->weeklyOn(2, '10:00'); // Tuesdays at 10AM UTC
+            if (app()->environment() === 'production') {
+                $schedule->command(UpdateAwardsStaticData::class)->everyMinute();
+                $schedule->command(CrawlPlayerWeightedPoints::class)->everyFiveMinutes();
+                $schedule->command(BackfillPlaytimeTotal::class)->everyTenMinutes();
+                $schedule->command(UpdatePlayerPointsStats::class, ['--existing-only'])->hourly();
+                $schedule->command(ProcessExpiringClaims::class)->hourly();
+                $schedule->command(UpdateSearchIndexForQueuedEntities::class)->twiceDaily(1, 13); // 1AM and 1PM UTC
```
Jamiras: What queues items here? Should this be kept?

app/Platform/AppServiceProvider.php
```diff
@@ -151,14 +151,16 @@ public function boot(): void
...
+                $schedule->command(DeleteStalePlayerPointsStatsEntries::class)->weekly();
```
Jamiras: Maybe this should be kept for similar reasons as the OnlineCount. It could make players appear recently active even if they weren't.

app/Providers/AppServiceProvider.php
wescopeland: Updated in latest.

app/Platform/AppServiceProvider.php
wescopeland: `SearchIndexingService` manages this. This is going to be so fast in non-prod environments there's probably no harm in removing it from this conditional.

app/Platform/AppServiceProvider.php
wescopeland: Updated in latest.

PR #3658, opened 2025-06-21
perf: add SyncAuthenticationCookie middleware

wescopeland:
https://discord.com/channels/476211979464343552/718091467893243937/1383440024846078133

This PR facilitates an experiment I'd like to try. If the experiment fails, the changes in this branch will be reverted.

~44% of our web traffic is unauthenticated. This traffic is still using a significant amount of server resources on every page hit, but it's largely requesting mostly-static or loosely-static content: the home page, game pages, leaderboard pages, and achievement pages.

Rather than serve this from the app server in Nuremberg, we can potentially save a tremendous amount of compute resources by having Cloudflare serve this stuff at the edge with a short TTL of like 5 minutes. I've tried this once before, but immediately had to disable the Cloudflare rule driving the behavior because it unfortunately served content at the edge for authenticated users as well.

This PR addresses that pitfall. Currently Cloudflare can't tell if a user is authenticated or a guest because every single hit to the site creates an encrypted `retroachievements_session` cookie by intentional design. The middleware added in this PR adds a simple accompanying `authenticated=1` cookie if the user is logged in. Now, Cloudflare's edge caching rule can be driven by the presence/absence of this cookie. If the cookie is absent and the user is requesting one of a whitelisted path of edge cached routes, the cached stuff can be served to the user instead.

The edge cached routes are effectively static HTML with whatever assets are needed sprinkled on top, served from the datacenter nearest to the guest's geolocation. I've confirmed this means `/game/228` loads in ~45ms for a guest located on the Harvard University campus as opposed to the ~700ms they see today, while Cloudflare only hits the app server once ~5 minutes on demand for the page to keep the cache fresh.


app/Http/Middleware/SyncAuthenticationCookie.php
```diff
@@ -0,0 +1,61 @@
...
+                cookie(
+                    'authenticated',
+                    '1',
+                    config('session.lifetime'),
+                    config('session.path'),
+                    config('session.domain'),
+                    config('session.secure'),
+                    true, // httpOnly
+                    false, // raw
+                    config('session.same_site')
```
Jamiras, line 47: I'm not sure what this does, but if we're not restricting this to same_site, we might want a better token name: `retroachievements_authenticated`?

Jamiras (approved):
Which pages is this going to affect? I imagine there's very little overlap between agents.

Players browsing unauthenticated will probably look at the home page, their profile, and/or whichever game they're currently playing.

Bots will crawl down whatever links they find, and won't likely revisit the pages they've hit before the edge cache expires them.

The only real benefit I can imagine from this is the forums. Both bots and multiple users might visit the same set of recent forum posts.

app/Http/Middleware/SyncAuthenticationCookie.php
wescopeland: This is currently restricted to same-site:
```php
// session.php
'same_site' => 'lax',
```
"lax" allows the cookie to be sent with same-site requests and cross-site top-level navigations (ie: when a user clicks on a link). It doesn't allow for cross-site subrequests though (ie: images, AJAX calls, iframes, etc).

Either way, I like the `retroachievements_authenticated` name more, so I've renamed the cookie to that in latest.

PR #3661, opened 2025-06-22
perf(UpdatePlayerGameMetricsJob): add lightweight fork for users with 0 unlocks

wescopeland:
https://discord.com/channels/@me/1065706550318084117/1386158764348145764

Currently, when an achievement set receives impactful changes, we dispatch a job for every player who has a `PlayerGame` record for that game. Each job recalculates all metrics - achievements, completion percentages, beat progress, badge eligibility, etc. This is expensive and unnecessary for players who have never unlocked any achievements for the game or its subsets.

For popular games, we're doing thousands of unnecessary & heavy computations.

This PR adds an early exit condition in `UpdatePlayerGameMetricsJob` that checks if the player has unlocked 0 achievements (both for core and for subsets). If so:
* The heavy metric recalculation stuff is skipped entirely.
* A lightweight `UpdatePlayerGamePlaytimeJob` is always dispatched to ensure playtime stays up to date.

For game 11285, ~63% of players have unlocked 0 achievements. This eliminates ~1,000 heavy jobs from the queue, replacing them with lightweight playtime-only jobs that do a fraction of the work.


app/Platform/Jobs/UpdatePlayerGameMetricsJob.php
```diff
@@ -78,6 +78,30 @@ public function handle(): void
...
+                    $query->where('GameID', $this->gameId);
+                })
+                ->exists();
+
+            // If they don't have any unlocks, we'll run a lightweight job just to calculate
+            // the player's playtime, and we'll skip processing all the heavy stuff.
+            if (!$hasAnyUnlocks) {
+                // Always update playtime for players with no achievements.
+                dispatch(new UpdatePlayerGamePlaytimeJob($this->userId, $this->gameId))
+                    ->onQueue('player-game-metrics-batch');
```
Jamiras: Why dispatch this? The much more expensive full call is executed without being dispatched on line 106 below. We're already in a dispatched job for the playerGame record at this point.

You could just inline the handle functionality and avoid another job and another query for the playerGame record:
```
if (!$hasAnyUnlocks) {
    $activityService = new PlayerGameActivityService();
    $activityService->initialize($playerGame->user, $playerGame->game);
    $summary = $activityService->summarize();

    // Only save if playtime actually changed.
    if ($playerGame->playtime_total !== $summary['totalPlaytime']) {
        $playerGame->playtime_total = $summary['totalPlaytime'];
        $playerGame->save();
    }
    return;
}
```

wescopeland: Yep, that's a lot better. Updated in latest.

PR #3578, opened 2025-05-31
Split UpdateGameMetricsAction into multiple actions

Jamiras:
* `UpdateGameMetricsAction` now only updates metadata about the game (number of achievements, number of points, hash of all published achievements).
  - It's called whenever an achievement is modified (promoted/demoted/points changed)
  - Calculating the hash is probably a lot less expensive than scanning millions of rows in the database, but it's still something that we don't need to be doing as frequently as we were.
  - Whenever the hash changes, the set is synced to the core achievement set for the game. Doing a sync (even if the only change was the unlock percentage) was the third most expensive part of calling `UpdateGameMetricsAction`. That was largely improved by #3573.
* `UpdateGamePlayerCountAction` updates the number of players for the game.
  - It's called when a player unlocks their first achievement for a game or resets their progress, when achievements are promoted/demoted, or when a player who has played the game is unranked/re-ranked.
  - This was previously the most expensive part of calling `UpdateGameMetricsAction`.
* `UpdateGameBeatenMetricsAction` updates the median time to beat metrics for a game.
  - It's called when a player earns or loses a SiteAward for mastering or beating a game (beaten awards are not shown on the profile page, but are tracked in the database).
  - This was the second most expensive part of calling `UpdateGameMetricsAction`.
* `UpdateAchievementMetricsAction` updates the unlock count, unlock percentage, and weighted points for an achievement.
  - It's called when an achievement is locked/unlocked or its points change.
  - It's also called for all achievements of a game when the game's player count changes.
  - This was the fourth most expensive part of calling `UpdateGameMetricsAction`.
  - Because I've separated this code from `UpdateGameAchievementsMetricsAction`, this will conflict with #3572, but it should be straightforward to port those changes to the new location.
  
This does not attempt to optimize any of the code/queries being used. It's trying to minimize the amount of code/queries that have to be executed in response to a player unlocking an achievement.

In the existing codebase, unlocking any achievement triggers the `UpdateGameMetricsAction` (presumably to update the player count for the set), and that does all of the things listed above every time it's called.

With this PR, unlocking an achievement should only call `UpdateAchievementMetricsAction` for the single achievement being unlocked.

If it's the player's first achievement for a set, it will also call `UpdateGamePlayerCountAction`, which will call `UpdateAchievementMetricsAction` for all achievements in the set. That will be the most expensive call for any unlock, and is still going to be faster than the previous implementation because it's not recalculating the beaten metrics or the achievement_set_version_hash.

[...]

PR #3662, opened 2025-06-22
fix alt event achievement unlocks being lost in migration

Jamiras:
Fixes https://discord.com/channels/310192285306454017/1312920636577808384/1386363974639026206

When an unlock was created for a main from an alt, it was almost immediately deleted by the cleanup code that was removing unlocks for any users who hadn't earned the source achievement. Since the main hadn't earned it, the converted unlock was being deleted.

I believe these commands are non-destructive, so running them again to reapply the unlocks for alts should be safe.


app/Platform/Commands/SyncEvents.php
```diff
@@ -930,7 +937,8 @@ protected function updateMetrics(Event $event): void
         $gameId = $event->legacyGame->id;
 
         // update metrics and sync to game_achievement_set
-        dispatch(new UpdateGameMetricsJob($gameId))->onQueue('game-metrics');
+        dispatch(new UpdateGamePlayerCountJob($gameId))->onQueue('game-player-count');
+        dispatch(new UpdateGameAchievementsMetricsJob($gameId))->onQueue('achievement-metrics');
```
Jamiras, line 941: Ensures the correct actions are invoked after #3578

wescopeland (approved):

PR #3663, opened 2025-06-23
add action for lbinfo connect api

Jamiras:


app/Helpers/database/leaderboard.php
```diff
@@ -149,116 +149,6 @@ function SubmitLeaderboardEntry(
...
-        $entry = getLeaderboardUserEntry($leaderboard, $user);
-        if ($entry) {
-            $retVal['Entries'][] = $entry;
-        }
-    }
-
-    return $retVal;
-}
-
-function getLeaderboardUserEntry(Leaderboard $leaderboard, User $user): ?array
```
Jamiras, line 235: was only called by `GetLeaderboardData`.

app/Platform/Actions/GetRankedLeaderboardEntriesAction.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Platform\Actions;
+
+use App\Models\Leaderboard;
+
+class GetRankedLeaderboardEntriesAction
```
Jamiras, line 9: This code is used in all three places that called `GetLeaderboardData`.

public API only uses this
lbinfo connect API also supports adjusting the offset to surround a user
leaderboard page also retrieves the entry for the logged in user

app/Platform/Actions/GetRankedLeaderboardEntriesAction.php
```diff
@@ -0,0 +1,40 @@
...
+
+        $entries = $leaderboard->sortedEntries()->with('user')->skip($offset)->take($count);
+        foreach ($entries->get() as $entry) {
+            if ($entry->score !== $rankScore) {
+                if ($rankScore === null) {
+                    // first entry, get its rank
+                    $rank = $leaderboard->getRank($entry->score);
+                } else {
+                    // non-first entry, just update rank to be the current index
+                    $rank = $index;
```
wescopeland, line 27: I'm pretty sure there's a logic error here, carried over from the original implementation. I'm not sure if it's even worth addressing, just seems like it's worth noting.

I believe if `$offset` is non-zero, setting `$rank = $index` gives incorrect/misleading ranks. `$index` is the position within the current result set, not the actual rank in the full leaderboard.

It's probably fine as a performance-vs-correctness trade-off, as I suspect resolving the issue requires executing a query to grab all unique scores from the result set.

Jamiras: Line 15:
```
$index = $offset + 1;
```
So `$index` will start at the offset (0-based) plus 1 (making it 1-based).

How is that not "the actual rank in the full leaderboard"?

Maybe you're missing the logic that fetches the rank of the first item (on line 24)? So if `$offset` is 20 and the first item has rank 15, `$rank` will start at 15, and at index 22, when the score changes, `$rank` will jump to 22.

wescopeland: Yep, you're right. Apologies for the confusion. I was thinking there was a possible edge case when certain distributions of tied scores existed in the leaderboard entries, but this doesn't actually occur. I just traced through the logic more carefully and the current algorithm does work correctly for all the data/offset combinations I tried.

PR #3666, opened 2025-06-25
feat(game): add achievement set credits display

wescopeland:
This PR adds achievement set credits to React game pages for non-XS displays. A follow-up PR will add credits for XS displays.

Additionally, some work that probably isn't even noticeable has been done to make the visible achievement set no longer collapsible, in favor of a future PR which will introduce some kind of "tabbed" display for multiset support.

## UX Walkthrough

Credit for the set appears underneath the set description.

Screenshot 2025-06-25 at 6 42 16 PM
*Game 1667*

Multiple credit chips are supported, with author being the most prominent. Progressive disclosure is used in the form of a tooltip display to show more details about the given credit:

Screenshot 2025-06-25 at 6 43 50 PM

### Authorship Credit

Sole authorship has already been shown above. If a set has 1-3 authors, all author avatars and display names are prominently shown, with authors sorted by how much they contributed to the set:

Screenshot 2025-06-25 at 6 45 10 PM
*Game 4*

If a set has 4-6 authors, the two most prominent authors have their display names shown, with the others going into an avatar stack:

Screenshot 2025-06-25 at 6 47 41 PM
*Game 3*

Edge case: if a set has 7+ authors, only an avatar stack is used:
Screenshot 2025-06-25 at 6 48 37 PM
Screenshot 2025-06-25 at 6 49 24 PM
*Game 20255*

### Artwork Credit

Game badge and achievement icon artwork is now properly credited on game pages. Only credit for the current badge and official achievements is shown. All artwork credit is shoved into a single artwork credit chip. For game badges, the date is shown in the tooltip. For achievement icons, the count of credited achievements is shown.

Screenshot 2025-06-25 at 6 51 16 PM
Screenshot 2025-06-25 at 6 51 40 PM
*Game 1*

Screenshot 2025-06-25 at 6 53 01 PM
*Game 1454*

### Maintainership and Trigger Code Credit

Active maintainers are now credited, as well as users who have contributed code to the set, but aren't necessarily authors. This sort of maintenance work that keeps sets healthy after the original authors have moved on deserves recognition.

Screenshot 2025-06-25 at 6 55 01 PM
Screenshot 2025-06-25 at 6 55 10 PM

### Design, Writing, and Playtesting Credit

These credit types are put into their own chip.

Screenshot 2025-06-25 at 6 56 29 PM
Screenshot 2025-06-25 at 6 56 36 PM
*Game 24453*

Screenshot 2025-06-25 at 6 57 47 PM
*Game 14739*

### More Examples

Selected previews from various game pages:

Screenshot 2025-06-25 at 6 58 35 PM
Screenshot 2025-06-25 at 6 59 02 PM
Screenshot 2025-06-25 at 6 59 24 PM
Screenshot 2025-06-25 at 6 59 46 PM
Screenshot 2025-06-25 at 7 00 08 PM
Screenshot 2025-06-25 at 7 00 41 PM


resources/js/features/games/components/AchievementSetCredits/ArtworkCreditsDisplay/ArtworkCreditsDisplay.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const artCreditUsers = [...badgeArtworkCredits, ...achievementArtworkCredits].filter(
+    (user, index, self) => index === self.findIndex((u) => u.displayName === user.displayName),
+  );
+
+  return (
+    <div
+      className={cn(
+        'flex items-center rounded-md bg-purple-900/20 py-1 pr-2 text-purple-400',
+        'light:border light:border-purple-200 light:bg-white light:text-purple-600',
+      )}
```
Jamiras, line 34: Are the secondary credit widgets intentionally smaller than the developer widget?
image

resources/js/features/games/components/AchievementSetCredits/AchievementAuthorsDisplay/AchievementAuthorsDisplay.tsx
```diff
@@ -0,0 +1,109 @@
...
+  authors: App.Platform.Data.UserCredits[];
+}
+
+export const AchievementAuthorsDisplay: FC<AchievementAuthorsDisplayProps> = ({ authors }) => {
+  const containerClassNames = cn(
+    'flex items-center rounded-md bg-amber-900/20 py-1 pr-2',
+    'light:bg-white light:border light:border-amber-300',
+  );
+
+  // 4-6 authors - show avatar images and name labels for top 2, stack for the rest
```
Jamiras: Rather than basing this on author count, could it be tied to percentage? Expand users who have contributed at least 33% of the achievements for the set, and collapse everyone else?

For example, on this game: http://localhost:64000/game2/4986 I have one achievement and the primary author has 35. I don't feel like I should be featured prominently for that set.

resources/js/features/games/components/AchievementSetCredits/AchievementSetCredits.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ...aggregateCredits.achievementsWriting,
+  ].filter(
+    (user, index, self) => index === self.findIndex((u) => u.displayName === user.displayName),
+  );
+
+  return (
+    <div
+      data-testid="set-credits"
+      className="hidden items-center justify-between text-neutral-300 light:text-neutral-700 sm:flex"
+    >
```
Jamiras, line 49: With where this it at now, it's very prominent and distracting for users who probably don't care about it. Maybe it's just all the extra colors that's grabbing my attention more than I feel is necessary.

Could we put it in a collapsed section, move it to the right side bar with the meta data, or just put the lists on the right side of the header widget? 

image

Maybe something even more discreet where it just shows icons for majority authors and everything else is in the tooltip?

image

Although I'm quickly realizing that without the "authors" label, the meaning is lost. Are those users being featured as first masters, or recent masters, or what does the trophy mean?

Jamiras: What about something like this?

image

The majority authors would have standard user tooltips and the "more" hover would show the full contribution list.

Jamiras: Have you considered where you're going to put other set metadata, like first released and revision dates or median mastery/completion times?

wescopeland: > Have you considered where you're going to put other set metadata, like first released and revision dates or median mastery/completion times?

It's on my TODO list. I'm kinda working as I go:
Screenshot 2025-06-26 at 11 13 48 AM

wescopeland: > Could we put it in a collapsed section, move it to the right side bar with the meta data, or just put the lists on the right side of the header widget?

I want to avoid putting stuff in the right-side of the header widget for now. This space could be used for the user's current progress on the set.

> Maybe something even more discreet where it just shows icons for majority authors and everything else is in the tooltip?

As much as possible, I want any burying of credit within a tooltip to be an edge case. It'll have to happen for mobile, but likely shouldn't be necessary for most sets on desktop.

> What about something like this? The majority authors would have standard user tooltips and the "more" hover would show the full contribution list.

Same issue - as much as we can get away with, I'd like to not bury credit within tooltips.

I want to focus on tackling the root concern:
> With where this it at now, it's very prominent and distracting for users who probably don't care about it. Maybe it's just all the extra colors that's grabbing my attention more than I feel is necessary.

The splash of color could be competing too strongly for your attention. Much of this info is already there on the PHP page - it's just noise among the rest of the clutter. In a subsequent commit on this branch, I'd like to try reducing the color splashes in the non-author chips.

resources/js/features/games/components/AchievementSetCredits/ArtworkCreditsDisplay/ArtworkCreditsDisplay.tsx
wescopeland: Definitely a bug - I didn't even notice. Updated in latest.

resources/js/features/games/components/AchievementSetCredits/AchievementSetCredits.tsx
wescopeland: In latest, I've made two changes:
* Avatar images are slightly smaller.
* With the exception of the trophy icon, the colors are now all neutral.

This gives the credits a much more subdued feeling that shouldn't compete so strongly for the viewer's attention:
Screenshot 2025-06-26 at 5 01 32 PM

resources/js/features/games/components/AchievementSetCredits/AchievementAuthorsDisplay/AchievementAuthorsDisplay.tsx
wescopeland: Great idea. I've made this change in latest.

Screenshot 2025-06-26 at 5 12 52 PM

resources/js/features/games/components/AchievementSetCredits/AchievementSetCredits.tsx
Jamiras: Much better. Thanks.

Jamiras (approved):

PR #3670, opened 2025-06-26
feat: make LG a fluid breakpoint

wescopeland:
This PR significantly improves the UX at the LG breakpoint by making it have a fully fluid main container width.

LG is typically targeting tablet resolutions, which are seeing a lot of dead space due to current fixed width constraints.

**Before**
Screenshot 2025-06-26 at 5 35 16 PM

**After**
Screenshot 2025-06-26 at 5 34 43 PM


Jamiras:
When transitioning from XL to LG (resizing window), the sidebar width changes, which causes a jarring change.

XL:
image

LG:
image

There are also jarring changes going down to MD and SM, but those also have more drastic changes to the layout, so it's less surprising.

This also occurs on master, but with the main container having a fixed width, it feels more like a size change. With the new changes, it feels a lot more unexpected.

PR #3672, opened 2025-06-26
fix(game): decode safe html entities in the meta description

wescopeland:
This PR resolves this encoding issue in PHP game page meta descriptions:
Screenshot 2025-06-26 at 6 51 03 PM

Tags are stripped to mitigate any risk of malicious content injection.


Jamiras:
Should we even be calling `sanitize_outputs` on `$gameTitle`? Looking at other places it's used, it's bound to several widgets, and further escaped into `$escapedGameTitle`, which is only used to populate the rename game field.

PR #3673, opened 2025-06-27
allow QA and set authors access to achievements for incompatible hash

Jamiras:
https://discord.com/channels/476211979464343552/1002695123030773790/1388105488473784471


app/Connect/Actions/BuildClientPatchDataAction.php
```diff
@@ -50,10 +51,24 @@ public function execute(
...
+
+            $canSeeIncompatibleSet = false;
+            if ($user) {
+                if ($user->id === $gameHash->compatibility_tester_id
+                    || $user->hasRole(Role::QUALITY_ASSURANCE)) {
+                    $canSeeIncompatibleSet = true;
+                } else {
+                    $canSeeIncompatibleSet = $game->achievements()->where('user_id', $user->id)->exists();
+                }
+            }
```
wescopeland: Recommend extracting this to a policy, maybe `GameHashPolicy`. This will likely have utility elsewhere in the future.

```php
// untested pseudocode
public function seeIncompatibleSet(User $user, GameHash $gameHash): bool
{
    // Compatibility testers can always access incompatible content.
    if ($user->is($gameHash->compatibilityTester)) {
        return true;
    }

    // QA members can always access incompatible content.
    if ($user->hasRole(Role::QUALITY_ASSURANCE)) {
        return true;
    }

    // Achievement authors can always access their own work.
    return $gameHash->game->achievements()->where('user_id', $user->id)->exists();
}
```

wescopeland (approved):

app/Connect/Actions/BuildClientPatchDataAction.php
Jamiras: done.

wescopeland (approved):

PR #3678, opened 2025-06-28
feat(game): add filter controls to React game pages

wescopeland:
This PR adds achievement filtering controls to React game pages. The sort menu and filter toggles live in a new "toolbar" underneath the set metadata:

Screenshot 2025-06-28 at 7 57 02 PM

There are some subtle details here worth noting:
1. This design fixes a UX issue from the PHP implementation where one checkbox was a "hide" action and the other was a "show" action. The filter labels are now much more simplified: "Locked Only" and "Missable Only".
2. The missable count is now planted directly on the filter itself rather than existing as copy elsewhere on the page.
3. The missable only filter now persists just like the locked only filter does (and existing cookies are re-used).


resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
```diff
@@ -0,0 +1,125 @@
...
+          'title',
+          '-title',
+          'type',
+          '-type',
+        ]}
+        buttonClassName="w-full sm:w-auto"
+      />
+
+      <div className="flex w-full gap-2 sm:w-auto">
+        {lockedAchievementsCount ? (
```
Jamiras: This should also hide the button if there aren't any unlocked achievements.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
```diff
@@ -0,0 +1,125 @@
...
+    isLockedOnlyFilterEnabledAtom,
+  );
+  const [isMissableOnlyFilterEnabled, setIsMissableOnlyFilterEnabled] = useAtom(
+    isMissableOnlyFilterEnabledAtom,
+  );
+
+  const handleToggleLockedOnlyFilter = (pressed: boolean) => {
+    setIsLockedOnlyFilterEnabled(pressed);
+    lockedOnlyCookie.toggleGameId(pressed);
+  };
```
Jamiras, line 53: I'm not sure if there's an easy way to change this, but would it be possible to only toggle the visibility on the affected items? As it is currently implemented, clicking the button hides everything, then reshows the items that should be visible. This creates a brief period where the list is blank:
image

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
```diff
@@ -0,0 +1,125 @@
...
+          '-wonBy',
+          'points',
+          '-points',
+          'title',
+          '-title',
+          'type',
+          '-type',
+        ]}
+        buttonClassName="w-full sm:w-auto"
+      />
```
Jamiras, line 78: This largely addresses everything in #1226 except:
> The one thing I would sometimes like to do is to see the locked/unlocked achievements in their original order (without grouping the unlocked first), as the display order often provides insight as to the order of achievements so I can figure out whereabouts in the game something I missed occurs.

Do you think that would be reasonable to add (probably as a separate PR?)

It should probably be a separate setting as you'd probably want to be able to use any of the sorts without grouping the unlocked achievements first.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
wescopeland: Fixed in latest.

Additionally, I've swapped the position of the two buttons so the "Locked Only" toggle placement will always be consistent, regardless of whether or not there are any missable achievements.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
wescopeland: I think this should be greatly improved in latest. There's still an animation, but it's much more subtle. I was going for just enough such that things don't pop in/out of existence in a single frame.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
wescopeland: Yep, it's something to consider. I'll add a note for this on my checklist.

PR #3679, opened 2025-06-29
feat(game): add content warnings to React game pages

wescopeland:
This PR adds content warnings (re-used from `hub.show`) to React game pages. A good ID to test with is 14107.

Screenshot 2025-06-29 at 9 42 03 AM


resources/js/features/games/components/+show/GameShowMainRoot.tsx
```diff
@@ -11,14 +12,16 @@ import { GameCommentList } from '../GameCommentList';
...
-  const { game } = usePageProps<App.Platform.Data.GameShowPageProps>();
+  const { game, hasMatureContent } = usePageProps<App.Platform.Data.GameShowPageProps>();
 
   if (!game.badgeUrl || !game.system?.iconUrl) {
     return null;
   }
 
   return (
     <div data-testid="game-show" className="flex flex-col gap-3">
+      {hasMatureContent ? <MatureContentWarningDialog /> : null}
```
Jamiras, line 23: Can we remember this in a cookie or with a URL parameter? If you hit "yes", but don't select "and don't ask again", any navigation away from the page causes the popup to be displayed again when returning to the page.

i.e:
* Open game page
* Press "Yes, I'm an Adult"
* Go to Supported Game Files or open any achievement
* Hit Back
  [Dialog appears again]

Jamiras, line 23: The way the dialog reads, it's unclear that selecting "and don't ask me again" applies to all pages. The old prompt was more explicit in that regard:
image

wescopeland: I like the URL parameter idea. In latest, we now add a "mature_content_accepted" param so the state is temporarily remembered in browser history.

wescopeland: This is tricky as we're somewhat cramped for space in the dialog. I've tried to improve the UX here in latest:

Screenshot 2025-07-15 at 7 52 39 PM

"No" is clear enough - it's answering the question being asked.

The two "yes" buttons are now subtly more closely positioned to each other to form more of a connection between them.

Neither of them explicitly says "Yes".

The most common (I assume) action - "Continue" is the right-most one.
"Always allow mature content" should be fairly explicit that the user will never see this warning again.

wescopeland:
> Would it be worthwhile to put some sort of indicator in the game header to inform the user about the mature content in case they have globally disabled the warning

In latest, I opted to go for the sidebar to keep the header area lean:

Screenshot 2025-07-15 at 7 56 22 PM

resources/js/features/games/components/+show/GameShowMainRoot.tsx
Jamiras: I'm not sure what the legal ramifications are, but this no longer requires the user lie about being an adult.

resources/js/features/games/components/MatureContentIndicator/MatureContentIndicator.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { useTranslation } from 'react-i18next';
+import { LuEyeOff } from 'react-icons/lu';
+
+export const MatureContentIndicator: FC = () => {
+  const { t } = useTranslation();
+
+  return (
+    <div className="flex justify-center rounded-lg bg-embed p-1 light:border light:border-neutral-200 light:bg-white">
+      <div className="flex w-full items-center justify-center gap-1.5 rounded-lg bg-neutral-800 p-1.5 text-neutral-300 light:text-neutral-700">
+        <LuEyeOff className="size-4" />
```
Jamiras: Is this the best icon? I would expect some sort of warning indicator. 

As it stands, I can't help but want to click on it to find the mature content.

resources/js/features/games/components/+show/GameShowMainRoot.tsx
wescopeland: I've had success using o3 for basic cross-jurisdictional legal advice. [Here's](https://chatgpt.com/share/6878129e-482c-8000-889b-5aefa7916a43) what it came up with.

I wasn't aware this meta tag was a thing:
```html
<meta name="rating" content="mature">
```
It may be worth adding later on pages that have this flag. I'll make a note of this.


In latest, I've changed the "Continue to this page" label back to "Yes, I'm an adult":
Screenshot 2025-07-16 at 5 04 13 PM

resources/js/features/games/components/MatureContentIndicator/MatureContentIndicator.tsx
wescopeland: Updated to a shield in latest:
Screenshot 2025-07-16 at 5 03 59 PM

resources/js/features/games/components/+show/GameShowMainRoot.tsx
Jamiras: Would it make sense to swap "Yes" and "Always"? It feels weird to have "Yes" so far from "No".

And why is "No" indented so far?

What about something like this?
image

Or maybe something more Windows-like?
image

wescopeland: There is a substantial amount of human-interface guidelines research which suggests the trailing (right-most) action in a dialog should be the acceptance state or action which is most probable to be used. 

I fear a checkbox is confusing from a UX POV because it's ambiguous whether the user can tick the box and still press "No". It's also unclear how to untick the box later (you can't).

Here's where I've landed in latest, which should hopefully be a happy medium. The left-most button has no outline, which makes it visually distinct from the primary accept and secondary cancel actions.
Screenshot 2025-07-17 at 6 32 43 PM

Jamiras: This highlight around "Always allow" could use a bit of horizontal padding.
image

wescopeland: Fixed in latest.

Screenshot 2025-07-18 at 5 17 23 PM

PR #3680, opened 2025-06-29
fix(PlayerGameActivityService): respect game resets

wescopeland:
This PR makes an adjustment to how `PlayerGameActivityService` calculates playtimes.

Currently, full game resets are not respected in the playtime calculation. If a player masters a game, resets the game, then masters the game again, both playthroughs will be calculated in their mastery time. This has negative consequences both for the user's profile and our aggregate playtime metrics.

This PR includes a migration to add a new table: `player_progress_resets`. Achievement and game resets are both tracked. `PlayerGameActivityService` now checks for if the player has reset the game, and if they have, filters included sessions in the calculation from the most recent game reset date.

There is nothing included in this branch to backfill data for `player_progress_resets`; I'm not even sure such a backfill is possible with what we currently store in the player metrics tables. Additionally, there is no mechanism included to "heal" `player_games` records currently affected by this issue. This _may_ be possible by checking if there are multiple dates in `beaten_dates`, `beaten_dates_hardcore`, `completion_dates`, or `completion_dates_hardcore` for a record, but that is beyond the scope of this PR.


app/Platform/Actions/ResetPlayerProgressAction.php
```diff
@@ -144,16 +153,25 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
...
             // Delete all maintainer unlock records related to these player_achievement entities.
             AchievementMaintainerUnlock::query()
                 ->whereIn('player_achievement_id', function ($query) use ($user) {
                     $query->select('id')->from('player_achievements')->where('user_id', $user->id);
                 })
                 ->delete();
 
-            // fulfill deletion request
             $user->playerGames()->forceDelete();
+            $user->playerSessions()->delete();
```
wescopeland: As best as I can tell, we seem to be holding on to player sessions for deleted users. I think we probably should not do this.

app/Platform/Actions/ResetPlayerProgressAction.php
```diff
@@ -132,6 +134,13 @@ public function execute(User $user, ?int $achievementID = null, ?int $gameID = n
                 }
             }
             $playerAchievement->delete();
+
+            // Track the achievement reset.
+            PlayerProgressReset::create([
+                'user_id' => $user->id,
+                'type' => PlayerProgressResetType::Achievement,
+                'type_id' => $achievementID,
+            ]);
```
Jamiras: If this is the player's last achievement for the game, it should also (or maybe only) mark the game as reset for purposes of calculating playtime. Should it not?

app/Platform/Actions/ResetPlayerProgressAction.php
Jamiras: Probably not, but that should happen in `ClearAccountDataAction`. Or maybe `$user->leaderboardEntries()->delete();` should be moved here.

app/Models/PlayerProgressReset.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        return $query->where('user_id', $user->id)
+            ->where(function ($q) use ($game) {
+                $q->where(function ($subQuery) use ($game) {
+                    $subQuery->where('type', PlayerProgressResetType::Game)
+                        ->where('type_id', $game->id);
+                });
+            })
+            ->orderByDesc('created_at');
+    }
```
Jamiras, line 111: This should also look for a full account reset. (I see a ResetType for account, but nothing generating one)

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -61,7 +62,18 @@ public function initialize(User $user, Game $game, bool $withSubsets = false): v
             $gameIds[] = $game->id;
         }
 
-        $playerSessions = $user->playerSessions()->with('gameHash')->whereIn('game_id', $gameIds)->orderBy('created_at')->get();
+        // Get the most recent reset for this user and game.
+        // If the player has a recent for the game, we only want to include
```
Jamiras: "recent" => "reset" or "recent reset"?

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -61,7 +62,18 @@ public function initialize(User $user, Game $game, bool $withSubsets = false): v
...
+        // sessions from after the most recent reset.
+        $lastReset = PlayerProgressReset::forUserAndGame($user, $game)->first();
```
Jamiras: Note that there may be more than one `gameId` if querying subset information too.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: It should. Updated in latest.

app/Platform/Actions/ResetPlayerProgressAction.php
wescopeland: I think I understand the difference in intent a little more clearly now.

`playerBadges->delete()` is in both, and I believe that implies `leaderboardEntries()->delete()` should also be in both. This block in `ResetPlayerProgressAction` could plausibly be used to wipe a user's progress while keeping their account intact.

That leads me to believe `$user->playerSessions()->delete()` is more appropriate to have in `ClearAccountDataAction` instead.

Updated in latest.

app/Models/PlayerProgressReset.php
wescopeland: Fixed in latest.

app/Platform/Services/PlayerGameActivityService.php
wescopeland: Yep, typo. Updated in latest.

app/Platform/Services/PlayerGameActivityService.php
wescopeland: Fixed in latest to read from multiple games if `$withSubsets` is truthy.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -61,7 +63,35 @@ public function initialize(User $user, Game $game, bool $withSubsets = false): v
...
+                        $subQuery->where('type', PlayerProgressResetType::Game)
+                            ->whereIn('type_id', $gameIds);
+                    })
+                    ->orWhere('type', PlayerProgressResetType::Account);
+                })
+                ->orderByDesc('created_at')
+                ->first();
+        } else {
+            $lastReset = PlayerProgressReset::forUserAndGame($user, $game)->first();
+        }
```
Jamiras, line 87: TODO: reset events should be captured in the timeline and shown on the game actiivity log for the user.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -61,7 +63,35 @@ public function initialize(User $user, Game $game, bool $withSubsets = false): v
...
+
+        $playerSessionsQuery = $user->playerSessions()
+            ->with('gameHash')
+            ->whereIn('game_id', $gameIds);
+        if ($lastReset) {
+            $playerSessionsQuery->where('created_at', '>=', $lastReset->created_at);
```
Jamiras: We don't want to do this here. It removes all history before the reset from the game activity log for the user. There should still be empty sessions shown.

i.e. go to http://localhost:64000/user/Jamiras/game/472/activity
Reset progress for the game
Refresh the page - all sessions are missing.

`$lastReset` should only be used in `getAchievementSetMetrics` as a more restrictive start date than `$achievementsPublishedAt`.

app/Platform/Services/PlayerGameActivityService.php
wescopeland: Great idea - I've added something in my backlog for this.

app/Platform/Services/PlayerGameActivityService.php
wescopeland: Updated in latest:
Screenshot 2025-07-16 at 6 01 10 PM

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -153,8 +152,13 @@ public function initialize(User $user, Game $game, bool $withSubsets = false): v
                 'Achievements.Points',
                 'Achievements.BadgeName',
                 'Achievements.type',
-            ])
-            ->get();
+            ]);
+
+        if ($this->lastResetCreatedAt) {
+            $playerAchievementsQuery->where('player_achievements.unlocked_at', '>=', $this->lastResetCreatedAt);
+        }
```
Jamiras: This should be redundant. If there are unlocks before the last reset, shouldn't we still process them?

wescopeland: Removed this block in latest.

PR #3683, opened 2025-06-30
fix(game): update Similar Games sort order on React game pages

wescopeland:
This PR updates the sort order for the Similar Games component on React game pages. Now, `sort_title` is the secondary sort instead of the primary sort. The primary sort order has been changed to whether the game has any achievements.

Screenshot 2025-06-30 at 6 49 06 PM


app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -76,7 +76,8 @@ public function execute(Game $game, ?User $user): GameShowPagePropsData
             ->filter(
                 fn ($game) => !str_contains($game->title, '[Subset')
             )
-            ->sortBy('sort_title');
+            ->sortBy('sort_title')
+            ->sortByDesc(fn ($game) => $game->achievements_published > 0);
```
Jamiras, line 81: I've verified the behavior, but this reads as if the sort title is still the primary sort. Aren't sorts processed in order?

Jamiras (approved):

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland: It's definitely counterintuitive. In Laravel collections, chaining multiple `->sortBy()` calls causes them to invoke in reverse order. [ref](https://github.com/laravel/ideas/issues/11#issuecomment-329866327)

PR #3688, opened 2025-07-02
perf: remove ziggy routes from Inertia props

wescopeland:
No functional changes.

This PR removes Ziggy routes from Inertia page props (a leftover from the `@routes` directive).

This cuts props size on all Inertia pages by ~45KB and speeds up client-side hydration runtime performance.


Jamiras (approved):
Not really sure what to test here, but I didn't notice anything while doing some basic smoketesting.

PR #3690, opened 2025-07-03
perf: compress user avatar URLs in Inertia page props

wescopeland:
No functional changes.

This PR removes more unnecessary data from Inertia page props by stripping user avatar base URLs (the media URL) from avatar URL fields.

On pages with lots of users loaded in memory, this can make a difference in client-side hydration performance.

**Before**
```
"https://media.retroachievements.org/UserPic/Scott.png"
```

**After**
```
"Scott.png"
```

### DIFF
--- a/resources/js/app.tsx
+++ b/resources/js/app.tsx
@@ -12,6 +12,7 @@ import { Ziggy } from './ziggy';
 
 // @ts-expect-error -- we're injecting this on purpose
 globalThis.Ziggy = Ziggy;
+globalThis.mediaUrl = import.meta.env.VITE_MEDIA_URL;
 
 const appName = import.meta.env.APP_NAME || 'RetroAchievements';
 
--- a/resources/js/common/components/UserAvatar/UserAvatar.tsx
+++ b/resources/js/common/components/UserAvatar/UserAvatar.tsx
@@ -3,6 +3,7 @@ import { route } from 'ziggy-js';
 
 import { useCardTooltip } from '@/common/hooks/useCardTooltip';
 import type { BaseAvatarProps } from '@/common/models';
+import { buildUserAvatarUrl } from '@/common/utils/buildUserAvatarUrl';
 import { cn } from '@/common/utils/cn';
 
 type UserAvatarProps = BaseAvatarProps &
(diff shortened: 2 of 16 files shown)


resources/js/common/components/UserAvatar/UserAvatar.tsx
```diff
@@ -40,7 +41,11 @@ export const UserAvatar: FC<UserAvatarProps> = ({
           decoding="async"
           width={size}
           height={size}
-          src={avatarUrl ?? 'https://media.retroachievements.org/UserPic/Server.png'}
+          src={
+            avatarUrl
+              ? buildUserAvatarUrl(avatarUrl)
+              : 'https://media.retroachievements.org/UserPic/Server.png'
```
Jamiras, line 47: Shouldn't this be `builderUserAvatarUrl('Server.png')`?

resources/js/common/utils/buildUserAvatarUrl.ts
```diff
@@ -0,0 +1,5 @@
+export function buildUserAvatarUrl(path: string) {
+  const baseUrl = globalThis.mediaUrl;
```
Jamiras, line 2: this seems to be `undefined` for me:
image
image
image

I am a bit confused how it's falling back to the server icon.

resources/js/app.tsx
```diff
@@ -12,6 +12,7 @@ import { Ziggy } from './ziggy';
 
 // @ts-expect-error -- we're injecting this on purpose
 globalThis.Ziggy = Ziggy;
+globalThis.mediaUrl = import.meta.env.VITE_MEDIA_URL;
```
Jamiras, line 15: should this just be `MEDIA_URL`?
https://github.com/RetroAchievements/RAWeb/blob/c00a8dc4a55c613c6722ea0b34fca8d1d7c2a284/config/app.php#L68
https://github.com/RetroAchievements/RAWeb/blob/c00a8dc4a55c613c6722ea0b34fca8d1d7c2a284/.env.example#L33

Mine is currently set to:
```
MEDIA_URL=https://media.retroachievements.org
```

Jamiras (commented):
I realize the rules are slightly different for the server, but weren't we moving away from constructing URLs so we'd have more control over changing them in the future?

PR #3694, opened 2025-07-03
perf(UserController): purge CDN-cached user avatars on upload or delete

wescopeland:
Currently, for users based in the Americas, uncached user avatars are taking 5x-10x longer to load than they should. This is because we're instructing Cloudflare explicitly never to cache these assets.

This makes sense. Given that we use stable usernames for user avatars (or could use `display_name` at some point), the file name is semi-stable, unlike game and achievement badge/icon images which use sequential numbers.

The solution implemented in this PR allows us to begin caching user avatars on Cloudflare's CDN, which means the assets will propagate to their datacenters worldwide, making that 1st hit much faster. The trick is rather than praying Cloudflare will detect the asset has changed, we explicitly tell Cloudflare via an HTTP API call something changed the user's avatar.

The API call immediately purges Cloudflare's cache for that image and begins propagating the new image to their datacenters worldwide. Optimistically, this should only take a few seconds.

Two new environment URLs have been added to _.env.example_. Real values can be pulled from prod's environment file in Forge. If the values aren't present in your _.env_ file, the private function to make the API call bails.

Caching will not be enabled for these assets in Cloudflare until this change is deployed to prod.


app/Http/Controllers/UserController.php
```diff
@@ -150,4 +153,24 @@ public function cancelAccountDeletion(Request $request): JsonResponse
...
+        $token = config('services.cloudflare.purge_token');
+        $zone = config('services.cloudflare.zone');
+        $mediaUrl = config('filesystems.disks.media.url');
+
+        if ($token && $zone && $mediaUrl) {
+            Http::withToken($token)
+                ->post("https://api.cloudflare.com/client/v4/zones/{$zone}/purge_cache", [
+                    // "https://media.retroachievements.org/UserPic/Scott.png"
+                    'files' => ["{$mediaUrl}/UserPic/{$user->username}.png"],
+                ]);
```
Jamiras, line 173: Do we want to try to capture when this errors? Not sure what types of errors it returns, or if the URL might change, or if there's just a generic network error.

app/Http/Controllers/UserController.php
```diff
@@ -122,6 +124,7 @@ public function deleteAvatar(Request $request): JsonResponse
         $this->authorize('updateAvatar', $user);
 
         removeAvatar($user->username);
+        $this->purgeUserAvatarFromCloudflare($user);
```
Jamiras, line 127: `removeAvatar` is also called when banning a user and clearing account data. Should this logic be moved into `media.php` and called from `removeAvatar` and `UploadAvatar`?

PR #3695, opened 2025-07-04
chore: create game_recent_players table, add backfill command

wescopeland:
Successor to #3509.

```
sail artisan horizon # start a new horizon instance

sail artisan migrate
sail artisan ra:platform:game:backfill-recent-players
```
---

`getGameRecentPlayers()` is the most expensive function/query on game pages, causing severe performance issues for high-traffic games due to expensive queries on the 50M+ row `player_sessions` table. 3509's caching attempt failed because synchronous cache invalidation in `ResumePlayerSessionAction` created cascading performance problems for popular games like 28275 (FFXI) which have millions of sessions.

The core performance issue is the current implementation uses a self-join pattern with `GROUP BY` to find `MAX(rich_presence_updated_at)` per user.  This O(n^2) operation becomes exponentially worse as the number of sessions we retain in the database grows over time.

This PR introduces a `game_recent_players` summary table that maintains only the most recent session per user per game. A subsequent PR will switch reads to use this table. The performance improvement comes from:

1. Eliminating the self-join. No more `GROUP BY` with `MAX()` across millions of rows.
2. Using a direct index lookup. We can use a simple `WHERE` on `game_id` with `ORDER BY` on an indexed column.

We'll be able to use an O(1) query regardless of session history size which we can update synchronously in `ResumePlayerSessionAction` without worrying about the complexity of managing some kind of cache.

I've tested the read performance of this change (not included in this PR) on game 28275. The recent players query goes from taking ~1000ms at best to ~20ms.

You can test the performance improvement by running the https://github.com/RetroAchievements/RAWeb/pull/3696's branch, which includes all the changes in this branch as well as the reads.


app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -91,6 +92,18 @@ public function execute(
...
                 $user->RichPresenceMsgDate = Carbon::now();
                 $user->saveQuietly();
+
+                // Update the player's game_recent_players table entry.
+                GameRecentPlayer::upsert(
+                    [
+                        'game_id' => $game->id,
+                        'user_id' => $user->id,
+                        'rich_presence' => $presence,
+                        'rich_presence_updated_at' => $timestamp,
```
Jamiras, line 102: Line 93 should probably also use `$timestamp`

app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -139,6 +152,18 @@ public function execute(
...
+        GameRecentPlayer::upsert(
+            [
+                'game_id' => $game->id,
+                'user_id' => $user->id,
+                'rich_presence' => $presence,
+                'rich_presence_updated_at' => $timestamp,
+            ],
+            ['game_id', 'user_id'],
+            ['rich_presence', 'rich_presence_updated_at']
+        );
```
Jamiras, line 165: Should this be wrapped in an `if ($presence)` check (like line 88)?

app/Platform/Actions/ResumePlayerSessionAction.php
wescopeland: Updated in latest.

app/Platform/Actions/ResumePlayerSessionAction.php
wescopeland: I don't think so, due to the conditional on line 130:
```php
// provide a default presence for the new session if none was provided
if (!$presence) {
    $presence = 'Playing ' . $game->title;
}
```
`$presence` will always be truthy on lines 162-165.

PR #3696, opened 2025-07-04
perf(game): read from game_recent_players

wescopeland:
Stacked on #3695.

Switches `getGameRecentPlayers()` to read from `game_recent_players` for a huge performance improvement. Should be deployed separately from #3695.

### DIFF
--- a/app/Helpers/database/player-game.php
+++ b/app/Helpers/database/player-game.php
@@ -4,13 +4,12 @@ use App\Enums\Permissions;
 use App\Models\Achievement;
 use App\Models\EventAchievement;
 use App\Models\Game;
+use App\Models\GameRecentPlayer;
 use App\Models\PlayerGame;
-use App\Models\PlayerSession;
 use App\Models\User;
 use App\Platform\Enums\AchievementFlag;
 use App\Platform\Services\GameTopAchieversService;
 use Illuminate\Database\Eloquent\Collection;
-use Illuminate\Support\Facades\DB;
 
 function getGameRankAndScore(int $gameID, User $user): array
 {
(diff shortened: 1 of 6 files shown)


app/Helpers/database/player-game.php
```diff
@@ -382,22 +381,14 @@ function getGameRecentPlayers(int $gameID, int $maximum_results = 10): array
...
-        ->groupBy('user_id')
-        ->select([DB::raw('MAX(rich_presence_updated_at) as rich_presence_updated_at'), 'user_id']);
-
-    // and use that information to get the recent session data for each user
-    $sessions = PlayerSession::where('game_id', $gameID)
-        ->join(DB::raw('(' . $subquery->toSql() . ') as subq'), function ($join) use ($subquery) {
-            $join->on('player_sessions.user_id', '=', 'subq.user_id')
-                 ->on('player_sessions.rich_presence_updated_at', '=', 'subq.rich_presence_updated_at')
-                 ->addBinding($subquery->getBindings());
+    $sessions = GameRecentPlayer::with(['user' => function ($query) {
```
Jamiras: Do you need the filter function on both the `with` and `whereHas` clauses?

wescopeland: Nope - it's redundant. Fixed in latest. Additionally, I've changed the filter to no longer rely on legacy permissions and instead check for truthiness of `email_verified_at`.

Jamiras: There are 50 users with Permissions > 0 and a null email_verified_at. I'm not sure what that means.

wescopeland: Those users are an edge case: https://discord.com/channels/476211979464343552/1002689485005406249/1202942157917388843

Ultimately, we should reach a point where we don't allow those users to interact with the system in any way until they have an email address set.

wescopeland: In latest, I've changed this to check nullishness of `banned_at`:
```php
$sessions = GameRecentPlayer::with('user')
    ->where('game_id', $gameID)
    ->whereHas('user', function ($query) {
        $query->whereNotNull('banned_at');
    })
    ->orderBy('rich_presence_updated_at', 'DESC');
```

This should hopefully get us the functionality we're looking for without needing to check legacy permissions.

app/Helpers/database/player-game.php
```diff
@@ -382,22 +381,12 @@ function getGameRecentPlayers(int $gameID, int $maximum_results = 10): array
...
-    // and use that information to get the recent session data for each user
-    $sessions = PlayerSession::where('game_id', $gameID)
-        ->join(DB::raw('(' . $subquery->toSql() . ') as subq'), function ($join) use ($subquery) {
-            $join->on('player_sessions.user_id', '=', 'subq.user_id')
-                 ->on('player_sessions.rich_presence_updated_at', '=', 'subq.rich_presence_updated_at')
-                 ->addBinding($subquery->getBindings());
+    $sessions = GameRecentPlayer::with('user')
+        ->where('game_id', $gameID)
+        ->whereHas('user', function ($query) {
+            $query->whereNotNull('banned_at');
```
Jamiras: Don't you want `whereNull('banned_at')`?

wescopeland: Yes, what a silly mistake. Updated in latest.

Jamiras (approved):

PR #3697, opened 2025-07-05
feat(game): add Recent Players to React game pages

wescopeland:
This PR adds the Recent Players component to React game pages. There are very few functional changes from the PHP implementation:

1. Table headers are visible to screen readers only.
2. Diff timestamps are now used.
3. Recent activity is given a green text color, and old activity is given a neutral text color.
4. RP strings are clamped to one line of text. The full string is available via a `title` attribute if the user wants to see it all through progressive disclosure.

**Before**
Screenshot 2025-07-05 at 4 36 25 PM

**After**
Screenshot 2025-07-05 at 4 36 35 PM


app/Platform/Data/GameRecentPlayerData.php
```diff
@@ -0,0 +1,63 @@
...
+    public function __construct(
+        public UserData $user,
+        public string $richPresence,
+        public Carbon $richPresenceUpdatedAt,
+        public int $achievementsUnlocked,
+        public int $achievementsUnlockedSoftcore,
+        public int $achievementsUnlockedHardcore,
+        public int $points,
+        public int $pointsHardcore,
+        public Lazy|PlayerBadgeData|null $highestAward,
```
wescopeland, line 32: The back-end logic to include `$highestAward` is present, but currently unused in the UI. I'll add the UI bits for this at some point in the future.

resources/js/features/games/components/GameRecentPlayers/GameRecentPlayersList/GameRecentPlayersList.tsx
```diff
@@ -0,0 +1,58 @@
...
+                  points: recentPlayer.points,
+                  pointsHardcore: recentPlayer.pointsHardcore,
+                }}
+                className="!py-0"
+              />
+            </div>
+          </div>
+
+          <p className="truncate text-2xs">
+            <RichPresenceMessage gameTitle={game.title} message={recentPlayer.richPresence} />
```
Jamiras: I understand the desire to limit this to a single line, but for some sets, it makes the RP impractical.
http://localhost:64000/game2/21700

image

vs
image

It's slightly more truncated in mobile mode (and I don't think there's any way to view the full value there):
image

Even the recent players widget shows more:
image

And while the RP itself looks broken ([who ]is exploring...), that's the way its coded:
```
?0xH004d8ae0>=1?is exploring 
```

Jamiras: Have we given any more thought to moving some of the data off the main game page into subpages? At least there the widget could stretch to full width.

wescopeland: > I understand the desire to limit this to a single line, but for some sets, it makes the RP impractical.

I think there's a spectrum of disclosure here and I'm trying my best to thread the needle to find the sweet spot while simultaneously steering dev behavior in a subtle way. On the other end of the spectrum lies showing everything, which I believe is equally impractical for numerous reasons including:
* It seems unlikely players are interested in seeing 10 paragraphs of RP text on a game page.
* If the player wants to see it, it's hoverable on desktop.
* This serves as a gentle design nudge to be more judicious with RP content length.

In latest, I've made two changes:
* The content is now tappable on mobile to see it all.
* If character length of the RP string is over 70, font size drops to 2xs to show more content.

> Have we given any more thought to moving some of the data off the main game page into subpages? At least there the widget could stretch to full width.

Yes, quite a bit. There are still several unknown unknowns with this that make me cautious, but I'm planning on trying my first stab at this with leaderboard listings.

Jamiras: I'm fully on board with encouraging shorter RP, and [the docs](https://docs.retroachievements.org/developer-docs/rich-presence.html#introduction) also support that.
> A good rich presence should inform other users how far into the game you are (level/town/dungeon), and give an overall sense of how the player is doing (score/remaining lives/character level). Additional details may be useful, depending on their context within the game. Try not to go overboard with the detail though. Other players don't care what you named your dog, or need to know how many bullets you have left, especially since it only updates every two minutes. Additionally, the more fluff you add to a rich presence display, the more diluted the important information becomes.

But even sufficiently terse text is sometimes longer than the space available:
image

wescopeland: Let's see what kind of feedback we get.

One thing I'm planning on doing is before we actually cut over from `game` to `game2`, leveraging `Role::BETA` for users to opt themselves in to the page. We're not quite ready for that yet, but I think this will be valuable for collecting feedback before we remove the PHP game page.

PR #3699, opened 2025-07-06
feat(game): set up back-end foundation for game.show multiset support

wescopeland:
This PR adds the back-end foundation for multiset support on React game pages:

Screenshot 2025-07-06 at 5 45 45 PM

This UI can be viewed by navigating to: http://localhost:64000/game2/668?set=8659. The URL should imply how this works.

If a `set` query param is present, that achievement set ID (and its legacy "backing game") are targeted for use by the game page. Several components are still hydrated from the "root game" (668 in this case), but several switch to being hydrated by the "backing game" instead:
* The achievement list and credits.
* Persistence with filter controls (they put the backing game ID in the cookie value).
* The ticket count.
* The "Want to Develop" / "Want to Revise" toggle.
* The "Want to Play" toggle.
* The "Compare Progress" component.
* The "Achievement Distribution" component.
* The comments section.
* The "Latest Masters" component.

There are several bits of the page that are intentionally _not_ using the backing game when a subset is in focus:
* The title and badge in the heading.
* The media.
* The metadata.
* The code notes URL.
* The series hub.
* The similar games.
* Non-metadata hubs.

I was less sure about what to do with these:
* The supported game files URL. We may want to base this on user multiset settings.
* The official forum topic - it's more about a game and less about a set.
* The "Find Related Games" URL - again, seems more game-driven than set-driven.

There is no ability to switch the set in the UI yet, but a prop has been prepared for it.

### DIFF
--- a/app/Platform/Commands/SyncEvents.php
+++ b/app/Platform/Commands/SyncEvents.php
@@ -849,7 +849,7 @@ class ConvertGame
                 $winners->where('unlocked_hardcore_at', '>=', $activeFrom);
             }
             if ($activeThrough) {
-                $winners->where('unlocked_hardcore_at', '<', $activeThrough);
+                $winners->where('unlocked_hardcore_at', '<', $activeThrough->clone()->addDays(1));
             }
 
             $winnerIds = [];


resources/js/pages/game/[game].tsx
Comment (Jamiras) [inline resources/js/pages/game/[game].tsx:1]: Should the set be added to the breadcrumb?

http://localhost:64000/game2/697?set=59
image

Old page:
image

resources/js/features/games/components/GameSidebarFullWidthButtons/GameSidebarFullWidthButtons.tsx
Jamiras, line 1: > I was less sure about what to do with these:
>   The supported game files URL. We may want to base this on user multiset settings.

Exclusive and Specialty subsets still require a unique hash (don't they?), and as such they need to have their own Supported Files shown.

For Bonus subsets, I don't have a good solution. We should display all supported hashes, especially for backwards compatibility. Maybe they should be hidden in a collapsed section like unsupported hashes?

Jamiras, line 1: > I was less sure about what to do with these:
> The official forum topic - it's more about a game and less about a set.

I think we should have both - at least until we can get per-game forums up and running.

image

Most of the subset official forum topics explain the rules of the subset, and/or provide tips and feedback specific to the subset, and I'd hate for that to get lost.

I see it as akin to the game comments, which you are currently showing for the subset.

Jamiras, line 1: > I was less sure about what to do with these:
> The "Find Related Games" URL - again, seems more game-driven than set-driven.

I agree. There might be someone looking for a similar type of subset (like rare drops), but the algorithm isn't going to return those.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -31,15 +33,43 @@ public function __construct(
...
+            $backingGame = Game::find($backingGameId);
+
+            // Load the visible comments relationship for the backing game.
+            $backingGame->load(['visibleComments' => function ($query) {
+                $query->latest('Submitted')
+                    ->limit(20)
+                    ->with(['user' => function ($userQuery) {
+                        $userQuery->withTrashed();
+                    }]);
+            }]);
```
Jamiras, line 71: Adding a comment to the subset page posts it on the main game page.

resources/js/pages/game/[game].tsx
Comment (Jamiras) [inline resources/js/pages/game/[game].tsx:1]: It's unclear that the "Want to Play" button is associated to the set. Should it be down in the set header, or somehow decorated to better indicate the association to the set?

image

Comment (Jamiras) [reply resources/js/pages/game/[game].tsx:1]: A similar concern could be raised for the sidebar buttons too: Tickets and Want to Revise (and anything else that would be subset specific) look no different than the ones that still point at the primary game (Memory. Manage, etc).

Comment (wescopeland) [reply resources/js/pages/game/[game].tsx:1]: It should. Updated in latest.

resources/js/features/games/components/GameSidebarFullWidthButtons/GameSidebarFullWidthButtons.tsx
wescopeland: > Exclusive and Specialty subsets still require a unique hash (don't they?)

Yep, fixed these counts in latest.

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland: Fixed in latest.

wescopeland:
Still in the process of working through feedback - tests are failing and coverage is missing.

resources/js/features/games/components/GameSidebarFullWidthButtons/GameSidebarFullWidthButtons.tsx
wescopeland: Addressed in latest:
Screenshot 2025-07-21 at 7 22 48 PM

resources/js/pages/game/[game].tsx
Comment (wescopeland) [reply resources/js/pages/game/[game].tsx:1]: Tried a bunch of things. Here's where I've landed.

Screenshot 2025-07-21 at 7 23 30 PM

The "Subset" indicator has associated `<LuLayers />` iconography. This iconography propagates out to the sidebar:
Screenshot 2025-07-21 at 7 24 09 PM

I'm content with this approach, but not happy. It seems "good enough", but I'll want to revisit this at some point.

Comment (Jamiras) [reply resources/js/pages/game/[game].tsx:1]: The "Subset Memory" link still goes to the memory page for the main game (which I believe is probably correct for most cases). Is the link wrong, or should the subset indicator be removed?

Comment (Jamiras) [reply resources/js/pages/game/[game].tsx:1]: The subset indicator is less important than the primary data type indicator. I think it should be after the page/ticket/plus icons, perhaps even smaller and inlayed after the text like in the header button (even without the text).

Comment (Jamiras) [reply resources/js/pages/game/[game].tsx:1]: For consistency, should the subset indicator be added to the Set Forum Topic button? Also for consistency, should that be Subset Forum Topic?

resources/js/features/games/components/GameSidebarFullWidthButtons/GameSidebarFullWidthButtons.tsx
Jamiras: I'm still not seeing any difference when viewing an Exclusive subset
http://localhost:64000/game2/1?set=9534

Should the Supported Files link to the subset's hashes? Or should the subset's hashes be included in the Supported Files for the main game under a subset hashes heading?

wescopeland: > I'm still not seeing any difference when viewing an Exclusive subset

Ah, I fixed the count but not the URL. The URL should also be fixed now in latest. I'm not sure if you were seeing an incorrect count, this is what my local looks like:

Screenshot 2025-07-22 at 3 34 41 PM

---

> Should the Supported Files link to the subset's hashes?

For the short-term, I think yes. 

> Or should the subset's hashes be included in the Supported Files for the main game under a subset hashes heading?

Long-term, I think this is a better idea. I've added something to my list in the lower-priority category.

resources/js/pages/game/[game].tsx
Comment (wescopeland) [reply resources/js/pages/game/[game].tsx:1]: > Is the link wrong, or should the subset indicator be removed [from the Memory link]?

I've removed the indicator for now on this specific link.

---

> The subset indicator is less important than the primary data type indicator. I think it should be after the page/ticket/plus icons, perhaps even smaller and inlayed after the text like in the header button (even without the text).

In latest, I've swapped its position with the link's primary icon. While I was iterating on this yesterday, I did attempt the smaller+inlayed idea, but this caused issues for button links that had counts associated with them (ie: open tickets).
Screenshot 2025-07-22 at 3 36 30 PM

---

> For consistency, should the subset indicator be added to the Set Forum Topic button?

IMO, it's redundant for that link, so likely detracts more than anything it might add.

> Also for consistency, should that be Subset Forum Topic?

Yes, updated in latest.

Screenshot 2025-07-22 at 3 38 24 PM

Jamiras (approved):

PR #3704, opened 2025-07-10
feat(game): improve credit handling for banned/deleted users

wescopeland:
This PR improves the credit handling UX for authors who have deleted their accounts (or are banned):

- Banned users are no longer clickable.
- Deleted users no longer have their delete date plumbed through page props.
- The strikethrough styles have been slightly modified and are now also seen in credit tooltips.

**Before**
Screenshot 2025-07-10 at 6 10 34 PM

Screenshot 2025-07-10 at 6 10 36 PM

---

**After**
Screenshot 2025-07-10 at 6 10 18 PM

Screenshot 2025-07-10 at 6 10 22 PM


resources/js/common/components/UserAvatar/UserAvatar.tsx
```diff
@@ -25,7 +26,7 @@ export const UserAvatar: FC<UserAvatarProps> = ({
 }) => {
   const { cardTooltipProps } = useCardTooltip({ dynamicType: 'user', dynamicId: displayName });
 
-  const canLinkToUser = displayName && !deletedAt;
+  const canLinkToUser = displayName && !deletedAt && !isGone;
```
Jamiras, line 29: Moderators can view the profile page of a banned user. It might be nice to leave the link for them? I'm not sure how often (if at all) it would be used.

Jamiras (approved):

resources/js/common/components/UserAvatar/UserAvatar.tsx
wescopeland: Messed around with this for about an hour. This is tricky for two reasons:
* `isGone` doesn't differentiate between deleted and banned (intentionally)
* `canManageUsers` would be a new global prop which needs to be read by `<UserAvatar />`

Not impossible by any means, but complex enough that I'm wondering if anyone cites this as being inconvenient. Going to punt to being a possible future task based on if we get feedback.

PR #3712, opened 2025-07-12
feat(game): show hash compatibility testing credit

wescopeland:
This PR adds hash compatibility testing credit to React game pages:

Screenshot 2025-07-12 at 10 51 07 AM

The credit is automatically derived from the game's connected hashes via this query:
```php
$compatibleHashes = $game->hashes()
    ->where('compatibility', GameHashCompatibility::Compatible)
    ->whereNotNull('compatibility_tester_id')
    ->whereColumn('compatibility_tester_id', '!=', 'user_id')
    ->with('compatibilityTester')
    ->get();
```
This:
* Grabs the compatible hashes that have testers associated with them.
* Excludes hashes where the tester is the same person who linked the hash.

Each of these hashes gets a credit entry displayed. Additionally, some new copy has been added to the hash management form in Filament:
Screenshot 2025-07-12 at 11 11 55 AM


app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -204,6 +205,24 @@ private function buildAggregateCredits(Game $game): AggregateAchievementSetCredi
             }
         }
 
+        // Process hash compatibility testing credits.
+        // Credit is given to users who successfully tested hash compatibility.
+        $compatibleHashes = $game->hashes()
+            ->where('compatibility', GameHashCompatibility::Compatible)
```
Jamiras, line 248: Would it make sense to also show credit to users who resulted in a hash being marked as incompatible? It's still work, but could only be a few minutes of work instead of having to play through the whole game.

Jamiras (approved):

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland: It's a good thing to consider. I lean towards no. To your point, there's likely a considerable effort disparity between marking something incompatible vs compatible, and I would be willing to bet incompatibility is identified within a few minutes more often than not, as opposed to having to play through and master the entire game.

There's also an incentive alignment here (subtle steering of user behavior). Everything on this page is going to steer behavior one way or another. I think showcasing compatible credits will more decisively encourage the hard work of thorough compatibility testing rather than the quick "this doesn't work" checks that anyone could do in a few minutes.

I think this also aligns with the design of how we're showing credit on the page: it reflects meaningful contributions that required notable work and/or expertise of some kind.

PR #3722, opened 2025-07-18
fix(game): remediate sorting bug on React game pages

wescopeland:
Resolves a regression on React game pages where client-side sorting is no longer working correctly.


resources/js/common/utils/sortAchievements.ts
```diff
@@ -13,6 +13,9 @@ export function sortAchievements(
   sortOrder: AchievementSortOrder,
   eventAchievements?: App.Platform.Data.EventAchievement[],
 ): App.Platform.Data.Achievement[] {
+  // Create a copy to avoid mutating the original array.
+  const sortedAchievements = [...achievements];
```
Jamiras, line 17: Nothing describes the problem being fixed, so I'm not sure how to test. If this is the only change, the list order may be affected, but not it's contents, so what was the problem?

wescopeland: As best as I can tell, sorting isn't working at all right now:
Screenshot 2025-07-19 at 2 02 26 PM

This is being caused by an issue with how React is diffing the DOM due to recent animation changes in the achievement list. Cloning the array with `[...achievements]` forces JavaScript to use a new memory address for the sorted values, which React can pick up on.

PR #3731, opened 2025-07-20
feat: add Meta+Enter keyboard hotkey to forum, message, and comment forms

wescopeland:
This PR adds a keyboard shortcut to message, forum post create/edit/reply forms, and React comment forms (ie: comments on game2). When the user presses Meta+Enter, (Cmd+Enter on macOS, Ctrl+Enter otherwise), the form will attempt to submit.

This is a conventional shortcut that exists on many other websites, such as GitHub.

https://github.com/user-attachments/assets/66d97947-e03e-47c7-b216-bfb1e000e36a


Jamiras (approved):
Verified functionality on message, forum page, and game2 comments.

It does not work on legacy comments (game/achievements/leaderboards), which I suspect was intentional.

PR #3747, opened 2025-07-25
fix: stop misattributing forum recent posts payload text

wescopeland:
This PR fixes a bug on a few pages such as the home page where forum recent post payloads that contain quotes might have misattributed text (the author of the post didn't write the quoted text, but it's being attributed to them).

**Before**
Screenshot 2025-07-25 at 6 41 22 PM

**After**
Screenshot 2025-07-25 at 6 56 21 PM

Quoted text to repro:
```
[quote]I agree with you however I also had the amusing thought that someones "back in the day" may have been the PS3, playing Sonic for the first ever via Sonic's Ultimate Genesis Collection. A collection in which you can save and load state anytime, anywhere! Wonder what happens if/when things like that get sets, since like the Konami code, save/load state is now also an official tool in the game.[/quote]

When we made the set for, we disallowed use of the compilation's save state functionality. Even if we did allow use of that feature, it would not work the way players expect, as we can't store achievement state in the game's save states like we can with emulator save states.
```

**Root Cause**
The post payload truncation is currently happening in the SQL query, which is too far upstream. It needs to happen in `stripAndClamp`. The bug is happening because the closing `[/quote]` tag is part of the truncated payload content.


app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
Jamiras, line 1: Still happens in the Forum Activity sidebar on http://localhost:64000/forum.php

wescopeland: Thanks, fixed in latest:
Screenshot 2025-07-27 at 1 46 41 PM

Jamiras (approved):

PR #3749, opened 2025-07-26
add API_GetGameProgression

Jamiras:
Meant to replace the APIs used by the RATools tool for determining how long it takes to reach various achievements in a set.

image

There's three main changes:
1) Uses actual session data instead of estimates from unlock timestamps (accounts for time leading up to first achievement).
2) Only looks at the last 100 active players of a set to better account for revisions, achievement fixes, and comments revealing ways to "cheese" the set.
3) All data can be fetched with a single API call instead of fetching data for all users individually and collating them on the client.

```
     {
         "ID":360110,
         "Title":"A Harder Journey",
         "Description":"Complete Story Mode on 4 Star difficulty or higher.",
         "Points":10,
         "TrueRatio":17,
         "Type":null,
         "BadgeName":"405469",
         "NumAwarded":29,
         "NumAwardedHardcore":26,
         "TimesUsedInUnlockMedian":16,
         "TimesUsedInHardcoreUnlockMedian":13,
         "MedianTimeToUnlock":2729,
         "MedianTimeToUnlockHardcore":2799
      },
```

Event managers use this tool to pick achievements with a rough target playtime:
* AotW = 2 hours = ~7200 MedianTimeToUnlockHardcore
* AotM = 8 hours = ~28800 MediumTimeToUnlockHardcore

### DIFF
--- a/app/Platform/Actions/ResumePlayerSessionAction.php
+++ b/app/Platform/Actions/ResumePlayerSessionAction.php
@@ -65,6 +65,11 @@ class ResumePlayerSessionAction
             $user->saveQuietly();
         }
 
+        // if the timestamp is within the session, just return it (backdated unlocks)
+        if ($playerSession && $timestamp->between($playerSession->created_at, $playerSession->rich_presence_updated_at)) {
+            return $playerSession;
+        }
+
         // if the session is less than 10 minutes old, resume session
         if ($playerSession && ($timestamp->diffInMinutes($playerSession->rich_presence_updated_at, true) < 10)) {
             $newDuration = max(1, (int) $timestamp->diffInMinutes($playerSession->created_at, true));
--- a/app/Platform/Actions/UpdatePlayerGameMetricsAction.php
+++ b/app/Platform/Actions/UpdatePlayerGameMetricsAction.php
@@ -14,6 +14,7 @@ use App\Platform\Enums\AchievementFlag;
 use App\Platform\Enums\AchievementSetType;
 use App\Platform\Enums\AchievementType;
 use App\Platform\Events\PlayerGameMetricsUpdated;
+use App\Platform\Jobs\UpdateGameBeatenMetricsJob;
 use App\Platform\Jobs\UpdateGamePlayerCountJob;
 use App\Platform\Services\PlayerGameActivityService;
 use ErrorException;
@@ -179,11 +180,16 @@ class UpdatePlayerGameMetricsAction
         $beatSummary = $activityService->getBeatProgressMetrics($coreAchievementSet, $playerGame);
         $playerGame->time_to_beat = $beatSummary['beatPlaytimeSoftcore'];
         $playerGame->time_to_beat_hardcore = $beatSummary['beatPlaytimeHardcore'];
+        $beatenChanged = $playerGame->isDirty(['time_to_beat', 'time_to_beat_hardcore']);
 
         $playerGame->save();
 
         if (!$silent) {
             PlayerGameMetricsUpdated::dispatch($user, $game);
+        }
+
+        if ($beatenChanged) {
+            dispatch(new UpdateGameBeatenMetricsJob($game->ID))->onQueue('game-beaten-metrics');
         }
 
         foreach ($possiblePlayerCountChangeGameIds as $gameId) {
--- a/public/API/API_GetGameProgression.php
+++ b/public/API/API_GetGameProgression.php
@@ -1,0 +1,177 @@
+<?php
+
+/*
+ *  API_GetGameProgression - returns information about the average time to unlock achievements in a game
+ *    i : game id
+ *    h : 1=ignore users with no hardcore unlocks
+ *
+ *  int        ID                                unique identifier of the game
+ *  string     Title                             name of the game
+ *  int        ConsoleID                         unique identifier of the console associated to the game
+ *  string     ConsoleName                       name of the console associated to the game
+ *  string     ImageIcon                         site-relative path to the game's icon image
+ *  int        NumDistinctPlayers                number of unique players who have earned achievements for the game
+ *  int        TimesUsedInBeatMedian             number of beats analyzed for the MedianTimeToBeat metric
+ *  int        TimesUsedInHardcoreBeatMedian     number of hardcore beats analyzed for the MedianTimeToBeatHardcore metric
+ *  int?       MedianTimeToBeat                  median number of seconds required to beat the game
+ *  int?       MedianTimeToBeatHardcore          median number of seconds required to beat the game in hardcore
+ *  int        TimesUsedInCompletionMedian       number of completions analyzed for the MedianTimeToComplete metric
+ *  int        TimesUsedInMasteryMedian          number of masteries analyzed for the MedianTimeToMaster metric
+ *  int?       MedianTimeToComplete              median number of seconds required to complete the game
+ *  int?       MedianTimeToMaster                median number of seconds required to master the game
+ *  int        NumAchievements                   count of core achievements associated to the game
+ *  array      Achievements
+ *   int        ID                               unique identifier of the achievement
+ *   string     Title                            title of the achievement
+ *   string     Description                      description of the achievement
+ *   int        Points                           number of points the achievement is worth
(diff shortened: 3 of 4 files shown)


app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -65,6 +65,11 @@ public function execute(
             $user->saveQuietly();
         }
 
+        // if the timestamp is within the session, just return it (backdated unlocks)
+        if ($playerSession && $timestamp->between($playerSession->created_at, $playerSession->rich_presence_updated_at)) {
+            return $playerSession;
+        }
```
Jamiras, line 75: The unit tests were creating extra sessions at "now" when trying to create unlocks at specific times in the past. I'm not sure if this issue would occur in production. We do allow backdating unlocks when retrying unlocks while the client can't reach the server.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -179,13 +180,18 @@ public function execute(PlayerGame $playerGame, bool $silent = false): void
...
 
         $playerGame->save();
 
         if (!$silent) {
             PlayerGameMetricsUpdated::dispatch($user, $game);
         }
 
+        if ($beatenChanged) {
+            dispatch(new UpdateGameBeatenMetricsJob($game->ID))->onQueue('game-beaten-metrics');
+        }
```
Jamiras, line 193: `UpdateGameBeatenMetrics` is currently only being called when badges were being awarded/revoked. Since we don't award badges for beating a game (only for mastering it), the beaten metrics weren't being updated quite as frequently as they should have been.

tests/Feature/Api/V1/GameProgressionTest.php
```diff
@@ -0,0 +1,250 @@
...
+
+class GameProgressionTest extends TestCase
+{
+    use RefreshDatabase;
+    use BootstrapsApiV1;
+    use TestsPlayerAchievements;
+
+    public function testUnknownGame(): void
+    {
+        $this->get($this->apiUrl('GetGameProgress', ['i' => 999999]))
```
wescopeland: Typo: this should be `GetGameProgression`.

public/API/API_GetGameProgression.php
```diff
@@ -0,0 +1,177 @@
...
+    ->where('type_id', $game->ID)
+    ->whereIn('user_id', $recentPlayerIds)
+    ->pluck('created_at', 'user_id');
+
+foreach ($recentPlayerIds as $playerId) {
+    $unlocks = PlayerAchievement::query()
+        ->where('user_id', $playerId)
+        ->whereIn('achievement_id', $achievementIds)
+        ->whereNull('unlocker_id')
+        ->get();
```
wescopeland: Performance: this executes separate queries for each player in the loop. For 100 players, that's potentially 200+ queries (`PlayerAchievement` + `PlayerSession` for each).

https://github.com/user-attachments/assets/80845c13-2045-44f8-b2f7-fab9d4dd3430

We should probably lift up loading all the unlocks & sessions in a couple of batch queries outside this loop.

public/API/API_GetGameProgression.php
```diff
@@ -0,0 +1,177 @@
...
+        'NumAwarded' => $achievement->unlocks_total,
+        'NumAwardedHardcore' => $achievement->unlocks_hardcore_total,
+        'TimesUsedInUnlockMedian' => count($unlock_times[$achievement->ID]),
+        'TimesUsedInHardcoreUnlockMedian' => count($unlock_hardcore_times[$achievement->ID]),
+        'MedianTimeToUnlock' => $get_median($unlock_times[$achievement->ID]),
+        'MedianTimeToUnlockHardcore' => $get_median($unlock_hardcore_times[$achievement->ID]),
+    ];
+}
+
+usort($response['Achievements'], fn ($a, $b) => $a['MedianTimeToUnlockHardcore'] - $b['MedianTimeToUnlockHardcore']);
```
wescopeland, line 226: Potential logic error: should the field used for `$a` and `$b` respect the `$preferHardcore` flag?

public/API/API_GetGameProgression.php
wescopeland: Performance: this endpoint takes ~2s on my machine to resolve. That may be dramatically improved by resolving the N+1 issues noted above. If it isn't, it may be worth considering if caching should be used on this endpoint just to mitigate someone aggressively slamming the same ID over and over.

public/API/API_GetGameProgression.php
```diff
@@ -0,0 +1,177 @@
...
+$unlock_hardcore_times = [];
+foreach ($achievementIds as $achievementId) {
+    $unlock_times[$achievementId] = [];
+    $unlock_hardcore_times[$achievementId] = [];
+}
+
+// ===== process the 100 most recent players to earn at least half of the achievements in the set =====
+$recentPlayerIds = PlayerGame::query()
+    ->where('game_id', $game->ID)
+    ->where($preferHardcore ? 'achievements_unlocked_hardcore' : 'achievements_unlocked', '>=', $game->achievements_published / 2)
```
wescopeland: Intellectual curiosity: why limit this to only players who have earned at least half the achievements? Could this skew statistics for difficult games where most players earn <50%?

Jamiras: To increase the number of usable rows. If we queried all users, the limiting of the 100 most recent rows would get a lot of players with 0-a few unlocks due to players just starting the game, or leaving before really investing any time in it.

I've modified the logic to use more reasonable thresholds for sets with fewer players.

public/API/API_GetGameProgression.php
Jamiras: I expected this feedback. I was concerned about the amount of data being loaded into memory and the possible per-user filtering caused by resets. I think I've come up with a reasonable alternative to address both.

public/API/API_GetGameProgression.php
Jamiras: `$preferHardcore` only drives the behavior of the threshold query (for determining who the last 100 players were). The results still contain data for both hardcore and softcore.

tests/Feature/Api/V1/GameProgressionTest.php
Jamiras: fixed

public/API/API_GetGameProgression.php
```diff
@@ -0,0 +1,229 @@
...
+                break;
+            }
+
+            $unlock_times[$unlock['achievement_id']][] = $elapsed +
+                $unlock['unlocked_at']->diffInSeconds($sessionStart, true);
+        }
+
+        // remove processed element so we don't try to process it again.
+        // this also allows us to avoid processing the user in the future once
+        // all of their acheivements have been processed.
```
wescopeland: Typo: "ach**ei**vements"

wescopeland (approved):

public/API/API_GetGameProgression.php
wescopeland: This is significantly improved from the previous iteration. Processing time for game ID 228 decreased from ~3s to ~380ms on my local machine.

While playing around with the latest code, I came up with another optimization which may be worth considering:
```php
foreach ($allSessions->get() as $session) {
    // Check if we have enough samples for meaningful medians.
    $totalSamples = 0;
    foreach ($unlock_times as $times) {
        $totalSamples += count($times);
    }
    // If we have enough samples, stop processing.
    if ($totalSamples > 1000) {
        break;
    }
```

On my local, this shaves off another ~50-80ms.

Jamiras: I wouldn't feel comfortable using any sort of hard-coded number here. There are already 9 sets with more than 500 achievements. 1000 captured unlocks might not provide data for many achievements.

Conversely, if we're querying the last 100 players, that's an average of 10 unlocks per player. Over 90% of sets have more than 10 achievements.

public/API/API_GetGameProgression.php
Jamiras: fixed.

PR #3751, opened 2025-07-27
perf(BackfillPlaytimeTotal): skip cron jobs on sundays

wescopeland:
Should hopefully reduce CPU pressure on Sundays.


Jamiras (approved):
Given that this code should go away in a few months, I think this is fine. If this were a more permanent solution, I'd suggest using the `->days(range(1,6))` helper as suggested [here](https://laracasts.com/discuss/channels/laravel/task-scheduler-exclude-sunday).

PR #3754, opened 2025-07-29
feat(game): add playtime stats to React game pages

wescopeland:
This PR adds playtime statistics to React game pages.

A migration is included in this PR which adds two indexes to `player_games`. On my local machine, execution time took 30s.

At the moment, award counts are fetched via the query in `GameTopAchieversService`, as it excludes untracked players and leverages the top-achievers cache.

Game 1454:
Screenshot 2025-07-28 at 9 00 04 PM

Subsets don't show the beaten row. Here's game 668, set 8659:
Screenshot 2025-07-28 at 9 00 33 PM

Game 5803:
Screenshot 2025-07-28 at 9 01 20 PM

Screenshot 2025-07-28 at 9 01 24 PM


app/Platform/Services/GameTopAchieversService.php
```diff
@@ -130,13 +157,13 @@ public function getRank(PlayerGame $playerGame): int
...
     {
-        $cacheKey = "game:{$gameId}:top-achievers:v2";
+        $cacheKey = "game:{$gameId}:top-achievers:v3";
         Cache::forget($cacheKey);
     }
 
     public function getTopAchieversComponentData(): array
     {
-        $cacheKey = "game:{$this->gameId}:top-achievers:v2";
+        $cacheKey = "game:{$this->gameId}:top-achievers:v4";
```
Jamiras: This is `v4`, but expire (line 160) is `v3`.

app/Platform/Data/GameData.php
```diff
@@ -88,7 +96,14 @@ public static function fromGame(Game $game): self
             )->all()),
 
             gameAchievementSets: Lazy::create(fn () => $game->gameAchievementSets->map(
-                fn ($gameAchievementSet) => GameAchievementSetData::from($gameAchievementSet)
+                function ($gameAchievementSet) {
+                    $gameAchievementSet->achievementSet->median_time_to_complete =
+                        $gameAchievementSet->achievementSet->median_time_to_complete ?? 0;
+                    $gameAchievementSet->achievementSet->median_time_to_complete_hardcore =
+                        $gameAchievementSet->achievementSet->median_time_to_complete_hardcore ?? 0;
```
Jamiras: Simplify to `x ??= 0`?

app/Platform/Services/GameTopAchieversService.php
```diff
@@ -38,6 +38,33 @@ private function baseQuery(): Builder
...
+            ->whereNull('beaten_hardcore_at')
+            ->count();
+    }
+
+    public function numBeaten(): int
+    {
+        return $this->baseQuery()
+            ->whereNotNull('beaten_hardcore_at')
+            ->count();
+    }
```
Jamiras, line 54: I know we're still in the midst of backfilling the data, but shouldn't these just be the `times_beaten` and `times_beaten_hardcore` values?

resources/js/features/games/components/PlaytimeStatistics/PlaytimeStatistics.tsx
```diff
@@ -0,0 +1,111 @@
...
+            rowSeconds={
+              currentMode === 'hardcore' ? game.medianTimeToBeatHardcore : game.medianTimeToBeat
+            }
+            totalPlayers={totalPlayers}
+            totalSamples={currentMode === 'hardcore' ? game.timesBeatenHardcore : game.timesBeaten}
+          />
+        ) : null}
+
+        <PlaytimeRow
+          headingLabel={t('Mastered')}
```
Jamiras: Should this change to "completed" for Softcore mode?

Image

Perhaps the icon should go from gold to silver too?

resources/js/features/games/components/PlaytimeStatistics/PlaytimeStatistics.tsx
```diff
@@ -0,0 +1,111 @@
...
+              : achievementSet.medianTimeToComplete
+          }
+          totalPlayers={totalPlayers}
+          totalSamples={
+            currentMode === 'hardcore'
+              ? achievementSet.timesCompletedHardcore
+              : achievementSet.timesCompleted
+          }
+        />
+      </motion.div>
```
Jamiras, line 115: Would it make sense to link the Latest Masters page from here?

Or does that have some implication that the Beaten list should be accessible too?

resources/js/features/games/components/PlaytimeStatistics/PlaytimeStatistics.tsx
```diff
@@ -0,0 +1,111 @@
...
+            iconClassName="text-neutral-400 light:text-neutral-500"
+            iconContainerClassName="bg-neutral-500/30 light:bg-neutral-500/20"
+            rowPlayers={currentMode === 'hardcore' ? numBeaten : numBeatenSoftcore}
+            rowSeconds={
+              currentMode === 'hardcore' ? game.medianTimeToBeatHardcore : game.medianTimeToBeat
+            }
+            totalPlayers={totalPlayers}
+            totalSamples={currentMode === 'hardcore' ? game.timesBeatenHardcore : game.timesBeaten}
+          />
+        ) : null}
```
Jamiras, line 89: The softcore numbers use players who beat the game in softcore out of all players for the set, which leads to the percentage values in softcore being low.

Image

I'm not sure if there's any easy way to derive the number of softcore players. `players_total - players_hardcore` could work, but that captures the number of people who have no hardcore unlocks. A softcore beat may still have hardcore unlocks.

app/Platform/Services/GameTopAchieversService.php
wescopeland: Is it desirable to include untracked players in the count? (I have no opinion)

resources/js/features/games/components/PlaytimeStatistics/PlaytimeStatistics.tsx
wescopeland: I think it would indeed imply that we would have a beaten list. At the moment, I don't think this component is a good fit for the link, but maybe we could do some additional treatment in the future.

app/Platform/Services/GameTopAchieversService.php
wescopeland: Fixed in latest.

app/Platform/Data/GameData.php
wescopeland: Updated in latest.

resources/js/features/games/components/PlaytimeStatistics/PlaytimeStatistics.tsx
wescopeland: Yep, this is a bug. Fixed in latest:
Screenshot 2025-07-30 at 8 50 32 PM

resources/js/features/games/components/PlaytimeStatistics/PlaytimeStatistics.tsx
wescopeland: Unfortunately, we probably don't have a great way to get the truly accurate number for softcore players at the moment. I went with the `players_total - players_hardcore` approach in latest, which at least is likely closer than what the previous approach was.

app/Platform/Services/GameTopAchieversService.php
Jamiras: If we want to do the filtering, we should probably do it at the denormalization stage.

I don't expect a handful of untracked timestamps to skew the median by any significant amount and eliminates the need for a filter.

For the service, it's probably best to be consistent with the other properties.

resources/js/features/games/components/PlaytimeStatistics/PlaytimeRow/PlaytimeRow.tsx
```diff
@@ -0,0 +1,114 @@
...
+                      minimumFractionDigits: 1,
+                      maximumFractionDigits: 1,
+                    }),
+                  })}
+                </span>
+              ) : (
+                <span>{t('{{val, number}} players', { count: rowPlayers, val: rowPlayers })}</span>
+              )}
+            </p>
+          ) : null}
```
Jamiras: Would it be better to explicitly state 0 players?
http://localhost:64000/game2/32138

Image

wescopeland: I actually think this is a bug.

In latest, I've made an update so this component will no longer render if there are 0 players.

app/Platform/Services/GameTopAchieversService.php
wescopeland: Agreed we should switch to the denormalized values. The only thing I'm getting hung up on is until the backfill is done, the counts for awards in the component are going to be wildly off, which may be difficult to explain to users.

I propose we cut over to the denormalized counts in this component once the backfill has completed.

resources/js/features/games/components/PlaytimeStatistics/PlaytimeRow/PlaytimeRow.tsx
Jamiras: The game I linked has players, they're just all hardcore. The widget should still be visible (at least the hardcore half)

wescopeland: Ah, I get what you're saying now. Made this change in latest:
Screenshot 2025-07-31 at 8 51 56 PM

Both tabs are still always visible, as long as there's at least 1 total player for the game.

PR #3758, opened 2025-07-30
feat: structure the Api namespace, prepare an internal-only service-to-service API

wescopeland:
This PR establishes the foundation for an internal service-to-service API with comprehensive logging and authentication capabilities.

**Why?**
The Internal API is ultimately meant to facilitate stuff like https://github.com/RetroAchievements/RABot-Next/issues/27, where one of our services (RABot) needs to talk to RAWeb in some secure way via HTTP. Ideally, it's used sparingly.

**Changes:**
* Adds an internal API namespace with a healthcheck endpoint at `/api/internal/health`.
* Adds header-based token authentication (`X-API-Key`) with service account restrictions.
* Creates a comprehensive API request logging system that can later be expanded to the V1 and V2 web APIs. **A migration is included with this PR** which creates an `api_logs` table.
* Adds a `PruneApiLogs` command to enforce a data retention policy for the request logging system.

The `Api` namespace has been restructured:
```
Api/
├── Internal/    # brand new, meant for service-to-service communications
├── Middleware/  # potentially shared by all HTTP APIs
├── V1/          # legacy, should ultimately house the existing web API
└── V2/          # should ultimately house the V2 JSON:API spec'd web API
```

Header token authentication is restricted to specific service accounts. Service accounts are configured via a new `INTERNAL_API_ALLOWED_USER_IDS` environment variable.

---

**How to test this PR:**
```bash
sail artisan migrate
```

API logs are recorded in the DB (new `api_logs` table) with a 14-day retention policy (enforced by new `PruneApiLogs` command).

In your _.env_, you need to add this env variable:
```env
INTERNAL_API_ALLOWED_USER_IDS=83906
```
83906 is RABot's user ID. You can use any user ID here, and the variable supports a comma-separated list of values.

There is one endpoint available on the internal API:
```
GET http://localhost:64000/api/internal/health
```
The legacy `?y=` parameter for the web API token **is not supported**. This is very insecure and an anti-pattern. I fully anticipate based on https://github.com/RetroAchievements/RABot-Next/issues/27 that the internal service API will not just be reading data, but making POST calls too.

Instead, you need to authenticate with a header. The header value should be the allowed user's web API key. Here's a sample `curl` using RABot's web API key:
```bash
curl --request GET \
  --url http://localhost:64000/api/internal/health \
  --header 'X-API-Key: 7t5...redacted...PdZ'
```
You should get back a simple response and see a log written to the `api_logs` table.
```json
{
	"status": "ok",
	"timestamp": "2025-07-30T02:29:30+00:00"
}
```


app/Api/Middleware/ServiceAccountOnly.php
```diff
@@ -0,0 +1,47 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+class ServiceAccountOnly
+{
+    public function handle(Request $request, Closure $next): Response
+    {
+        $user = $request->user();
+
+        if (!$user) {
+            return response()->json(['error' => 'Unauthorized'], 401);
```
Jamiras: I'm not sure where it's happening, but if I pass in a token that's not in the database (or no token header), I get back status 401 with the following content:
```
{
  "message":"Unauthenticated.",
  "errors":[{"status":419,"code":"unauthorized","title":"Unauthenticated."}]
}
```

I don't know how to make it reach this line.

One thing I think we want to proceed with in the new API is returning an error code in addition to an error message, that way we can tweak (and/or translate) the error message and still have a constant for the client to look for.

Given that it's an internal client, that may not be important, but we may open it up more in the future, or leverage whatever is done here in a more public API.

app/Api/Middleware/ServiceAccountOnly.php
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        $allowedUserIds = $this->getAllowedUserIds();
+
+        if (!in_array((int) $user->id, $allowedUserIds, true)) {
+            return response()->json(['error' => 'Access denied'], 403);
```
Jamiras: Since the logs are routinely purged, should we try to capture authentication failures?

app/Api/Middleware/LogApiRequest.php
```diff
@@ -0,0 +1,79 @@
...
+    public function __construct(
+        private readonly string $apiVersion = 'v2'
+    ) {
+    }
+
+    public function handle(Request $request, Closure $next, ?string $apiVersion = null): Response
+    {
+        $startTime = microtime(true);
+
+        $version = $apiVersion ?? $this->apiVersion;
```
Jamiras, line 22: `api_version` is being logged as "internal". Shouldn't it be "2" or something similar?

app/Api/Middleware/ServiceAccountOnly.php
```diff
@@ -0,0 +1,47 @@
...
+
+class ServiceAccountOnly
+{
+    public function handle(Request $request, Closure $next): Response
+    {
+        $user = $request->user();
+
+        if (!$user) {
+            return response()->json(['error' => 'Unauthorized'], 401);
+        }
```
Jamiras: Similarly, if I hit an endpoint that doesn't exist, I get back a 405 status and the following content:
```
{
  "message":"Method not allowed",
  "errors":[{"status":405,"code":"method_not_allowed","title":"Method not allowed"}]
}
```

database/migrations/2025_07_30_000000_create_api_logs_table.php
```diff
@@ -0,0 +1,40 @@
...
+            $table->unsignedBigInteger('user_id')->nullable();
+            $table->string('endpoint', 255);
+            $table->string('method', 10);
+            $table->integer('response_code');
+            $table->integer('response_time_ms')->nullable();
+            $table->string('ip_address', 45);
+            $table->text('user_agent')->nullable();
+            $table->json('request_data')->nullable();
+            $table->string('error_message')->nullable();
+            $table->timestamp('created_at')->useCurrent();
```
Jamiras, line 25: Would it be worth also capturing the response length?

app/Api/Middleware/LogApiRequest.php
```diff
@@ -0,0 +1,79 @@
...
+            $request->method(),
+            $response->getStatusCode(),
+            $this->calculateResponseTime($startTime),
+            $request->ip(),
+            $request->userAgent(),
+            $this->sanitizeRequestData($request),
+            $this->getErrorMessage($response)
+        );
+
+        return $response;
```
Jamiras, line 40: Consider injecting the `Content-Length` header in the response. See https://github.com/RetroAchievements/RAWeb/pull/3634#discussion_r2148064509

app/Api/Middleware/ServiceAccountOnly.php
wescopeland: > I'm not sure where it's happening, but if I pass in a token that's not in the database (or no token header), I get back status 401 with the following content:

Those calls were being intercepted by the `api-token-header` middleware, which happens to be throwing an error that looks like a JSON:API spec'd error response.

I've removed the `if (!$user) { ... }` conditional in latest.

> One thing I think we want to proceed with in the new API is returning an error code in addition to an error message, that way we can tweak (and/or translate) the error message and still have a constant for the client to look for.
> 
> Given that it's an internal client, that may not be important, but we may open it up more in the future, or leverage whatever is done here in a more public API.

I agree. The V2 API is probably going to be using the JSON:API spec, and it seems that's already what Laravel prefers anyway. In latest, I've installed and configured the Laravel JSON:API package. The forbidden error below the one you pointed out now throws a standardized exception:
```php
if (!in_array((int) $user->id, $allowedUserIds, true)) {
    throw JsonApiException::error([
        'status' => 403,
        'title' => 'Unauthorized',
        'code' => 'forbidden',
    ]);
}
```

This results in a response of:
```json
{
	"message": "Unauthorized",
	"errors": [
		{
			"code": "forbidden",
			"status": "403",
			"title": "Unauthorized"
		}
	]
}
```

The only thing weird here is the `status` value is a string. However, this seems to be compliant with [the JSON:API specification](https://jsonapi.org/format/#error-objects):
> status: the HTTP status code applicable to this problem, expressed as a string value. This SHOULD be provided.

app/Api/Middleware/ServiceAccountOnly.php
wescopeland: Updated in latest to capture these:
Screenshot 2025-07-31 at 7 56 42 PM

app/Api/Middleware/LogApiRequest.php
wescopeland: In this case, "internal" is correct. There are effectively 3 APIs in the `Api` namespace: V1, V2, and Internal.

Ideally, if someone calls API_GetUserSummary, the `api_version` would be "1".
If someone calls some hypothetical V2 JSON:API endpoint, the `api_version` would be "2".
If a service we control calls an internal endpoint, the `api_version` would be "internal".

app/Api/Middleware/ServiceAccountOnly.php
wescopeland: This one is handled automagically by Laravel, and I believe for this instance is probably correct.

database/migrations/2025_07_30_000000_create_api_logs_table.php
wescopeland: Yeah, I think so. Updated the migration & logging stuff in latest.

Screenshot 2025-07-31 at 8 14 27 PM

app/Api/Middleware/LogApiRequest.php
wescopeland: Wrote a small middleware to handle this for us automatically. Success responses now contain a `Content-Length`:
Screenshot 2025-07-31 at 8 15 02 PM

The value mirrors what's stored in the new `response_size_bytes` column in `api_logs`.

wescopeland:
Most of the diff in latest is vendor and lockfile noise.

app/Api/Middleware/ServiceAccountOnly.php
Jamiras: I'm still getting back 419 errors for invalid or no API key:
```
jamiras@ubuntu24 pr/3758:~/source/RAWeb$ curl "localhost:64000/api/internal/health" && echo
{"message":"Unauthenticated.","errors":[{"status":419,"code":"unauthorized","title":"Unauthenticated."}]}
jamiras@ubuntu24 pr/3758:~/source/RAWeb$ curl --header "X-API-Key: bogus" "localhost:64000/api/internal/health" && echo
{"message":"Unauthenticated.","errors":[{"status":419,"code":"unauthorized","title":"Unauthenticated."}]}
```

419 seems to be a [Laravel extension](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#Nonstandard_codes), but shouldn't it match the HTTP response code? Also note it's not a string.
```
jamiras@ubuntu24 pr/3758:~/source/RAWeb$ curl "localhost:64000/api/internal/health" -v
...
< HTTP/1.1 401 Unauthorized
...
```

I do get back the 403 error when using an unauthorized user's API key.

database/migrations/2025_07_30_000000_create_api_logs_table.php
Jamiras: Minor: `response_size_bytes` is NULL for laravel errors. You can see it the second row of your image. The 403 error from using a valid token for an unauthorized user is captured properly.

wescopeland: This should now be resolved in latest:
Screenshot 2025-08-14 at 8 07 37 PM

app/Api/Middleware/ServiceAccountOnly.php
wescopeland: Resolved in latest - these are now returned as string "401" statuses:
Screenshot 2025-08-14 at 8 08 08 PM

Jamiras (approved):

PR #3762, opened 2025-08-01
fix(PlayerGameProgressBar): resolve warning flooding the sentry logs

wescopeland:
Resolves https://retroachievementsorg.sentry.io/issues/53025161/?environment=production&query=is%3Aunresolved&referrer=issue-stream.

> Tooltip is changing from controlled to uncontrolled. Components should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled value for the lifetime of the component.

This warning is thrown when the `open` prop changes from a boolean value to `undefined` during a rerender.

To reproduce this warning, go to any game datatable and apply a Progress filter. If all logging is visible in the browser devtools (you may need to enable verbose logging), you should see the warning.

Now, a pure CSS solution is used instead to disable the tooltip. Tooltips are only disabled for this component on mobile.


Jamiras (approved):
I didn't see the original error, but I don't see any error with the change.

PR #3764, opened 2025-08-01
fix: resolve mobile Select warnings showing up in Sentry

wescopeland:
Related https://github.com/RetroAchievements/RAWeb/pull/3762.

Resolves https://retroachievementsorg.sentry.io/issues/53027333/?environment=production&query=is%3Aunresolved&referrer=issue-stream.

To reproduce the issue, change either the game type or set type select option on mobile. You'll see the warning appear when verbose logging is enabled in the browser devtools.

**Root Cause**
`value` should never changed from `undefined` (uncontrolled) to a truthy (controlled) value, or vice versa. A string "null" for these fields is considered the default state.


wescopeland:
> Blocked aria-hidden on an element because its descendant retained focus.

As best as I can tell, this appears to be a bug with the underlying drawer library (vaul). It's hosting an `aria-hidden` attribute to this div in `AppLayoutBase`:
```tsx
<div className="container lg:max-w-none xl:max-w-screen-xl">
```

A fix seems to have been merged three weeks ago that we'll inherit when I do a round of package updates: https://github.com/emilkowalski/vaul/pull/580

PR #3766, opened 2025-08-02
fix(manage): handle Livewire temporary image errors

wescopeland:
Resolves these recurring errors in the server logs:
Screenshot 2025-08-02 at 9 57 54 AM

**Root Cause**
Users are pressing submit after the image they've uploaded in Filament has been deleted from temporary storage. This results in multiple server errors.

Now, the user is still able to save the form (we can't easily prevent this, and they may have modified other fields), but if the image asset was purged, we'll tell them via a Filament notification.

Additionally, the expiration time has been doubled from 5 minutes to 10 minutes.


app/Filament/Resources/HubResource/Pages/Edit.php
```diff
@@ -25,19 +25,13 @@ protected function getHeaderActions(): array
...
-        } else {
-            // If no new image was uploaded, retain the existing image.
-            $data['image_asset_path'] = $existingImage;
+        (new ApplyUploadedImageToDataAction())->execute($data, 'image_asset_path', ImageUploadType::HubBadge);
+
+        // If no new image was uploaded, retain the existing image.
+        if (!isset($data['image_asset_path'])) {
+            /** @var GameSet $record */
+            $record = $this->record;
+            $data['image_asset_path'] = $record->image_asset_path;
```
Jamiras: I _think_ this could be removed. In all other cases, ensuring we don't provide any data for the field prevents the underlying code from trying to change it.

app/Filament/Resources/NewsResource/Pages/Edit.php
```diff
@@ -24,19 +24,13 @@ protected function getHeaderActions(): array
...
-        } else {
-            // If no new image was uploaded, retain the existing image.
-            $data['image_asset_path'] = $existingImage;
+        (new ApplyUploadedImageToDataAction())->execute($data, 'image_asset_path', ImageUploadType::News);
+
+        // If no new image was uploaded, retain the existing image.
+        if (!isset($data['image_asset_path'])) {
+            /** @var News $record */
+            $record = $this->record;
+            $data['image_asset_path'] = $record->image_asset_path;
```
Jamiras: Same comment. Maybe I'm wrong.

app/Filament/Actions/ApplyUploadedImageToDataAction.php
```diff
@@ -0,0 +1,48 @@
...
+     * If the image processes successfully, the field is updated with the new path.
+     * If processing fails (ie: the temp file on disk expired), the field is unset (preserving the existing value).
+     *
+     * @param array $data the form data array (modified in place)
+     * @param string $field the field name to process
+     * @param ImageUploadType $uploadType the type of image being uploaded
+     */
+    public function execute(array &$data, string $field, ImageUploadType $uploadType): void
+    {
+        // If we don't have a value, bail.
```
Jamiras: I think this comment needs clarification. A blank value means a file wasn't uploaded. Since we want to keep the previous value instead of changing it to blank, we explcitly remove the blank value sent in from the form.

Jamiras (commented):
Verified functionality for uploading in-game image. Did not test other scenarios.

Left a comment about why we're unsetting the field, and some suggestions to potentially leverage that functionality instead of doing additional work.

app/Filament/Resources/HubResource/Pages/Edit.php
wescopeland: Removed in latest.

app/Filament/Resources/NewsResource/Pages/Edit.php
wescopeland: Removed in latest.

app/Filament/Actions/ApplyUploadedImageToDataAction.php
wescopeland: Updated in latest to:
```php
// If no file was uploaded, unset the field to preserve the existing database value.
// Filament sends blank values for file inputs when nothing is uploaded.
if (!isset($data[$field])) {
```

Jamiras (approved):

PR #3770, opened 2025-08-02
feat(game): add support for showing unpublished achievements to React game pages

wescopeland:
This PR adds support for the `?f=5` (legacy) and `?unpublished=true` query params to React game pages. The query params both do the same thing, with the `unpublished` one being the preferred param.

The list of unpublished achievements can be accessed via a new link in the sidebar, visible only to developers:
Screenshot 2025-08-02 at 2 04 51 PM

The link is visible even if there are no unpublished achievements.

When viewing unpublished achievements, several sections of the page are set to not render:
* Comments
* Achievement distribution
* Recent players
* Top players
* Compare progress

Additionally, authors are visible on achievement items:
Screenshot 2025-08-02 at 2 07 02 PM


app/Platform/Controllers/GameController.php
```diff
@@ -110,8 +116,8 @@ public function show(
             }
         }
 
-        $game = $loadGameWithRelationsAction->execute($game, AchievementFlag::OfficialCore, $targetAchievementSet);
-        $props = $buildGameShowPagePropsAction->execute($game, $user, $targetAchievementSet);
+        $game = $loadGameWithRelationsAction->execute($game, $targetAchievementFlag, $targetAchievementSet);
+        $props = $buildGameShowPagePropsAction->execute($game, $user, $targetAchievementFlag, $targetAchievementSet);
```
Jamiras, line 129: I'm not sure what happened, but I demoted an achievement with unlocks and got this error:
```
Undefined array key 43 at /var/www/html/app/Helpers/render/game.php:421)
[stacktrace]
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(256): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError()
#1 /var/www/html/app/Helpers/render/game.php(421): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->Illuminate\\Foundation\\Bootstrap\\{closure}()
#2 /var/www/html/app/Platform/Actions/BuildGameAchievementDistributionAction.php(43): handleAllAchievementsCase()
#3 /var/www/html/app/Platform/Actions/BuildGameShowPagePropsAction.php(253): App\\Platform\\Actions\\BuildGameAchievementDistributionAction->execute()
#4 /var/www/html/app/Platform/Controllers/GameController.php(120): App\\Platform\\Actions\\BuildGameShowPagePropsAction->execute()
```
After repromoting it and demoting it again, the error didn't occur. The error also doesn't seem to occur when demoting other achievements or just [going to a page](http://localhost:64000/game2/31?unpublished=true) where there are already unofficial achievements with unlocks.

I assume it was some sort of race condition as the number of published achievements changed from 43 to 42, so the fix would simply be to add `?? 0` to the affected line (and the following line). But on a more relevant note: If we're not showing the achievement distribution, do we even need to build it?

resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.tsx
```diff
@@ -24,15 +26,42 @@ export const SidebarDevelopmentSection: FC = () => {
...
+
+      <PlayableSidebarButton
+        IconComponent={isViewingPublishedAchievements ? LuFolderLock : LuFolder}
+        href={buildToggleHref()}
+        showSubsetIndicator={game.id !== backingGame.id}
+      >
+        {isViewingPublishedAchievements
+          ? t('View Unpublished Achievements')
+          : t('View Published Achievements')}
+      </PlayableSidebarButton>
```
Jamiras: Would it be desirable to include the unpublished achievement count (like we do for supported game files)?

app/Platform/Controllers/GameController.php
```diff
@@ -93,9 +93,15 @@ public function show(
...
 
-        // Get the target achievement set ID from query parameter.
+        // Get the target achievement set ID from query params.
         $targetAchievementSetId = $request->query('set') ? (int) $request->query('set') : null;
 
+        // Get whether to show published or unpublished achievements from query params.
+        $targetAchievementFlag =
+            $request->query('unpublished') === 'true' || $request->query('f') === '5'
+                ? AchievementFlag::Unofficial
+                : AchievementFlag::OfficialCore;
```
Jamiras, line 112: Since the button is only visible to developers, shouldn't the page also be similarly restricted? It's currently visible when not logged in.

app/Platform/Controllers/GameController.php
```diff
@@ -93,9 +93,15 @@ public function show(
         /** @var ?User $user */
         $user = $request->user();
 
-        // Get the target achievement set ID from query parameter.
+        // Get the target achievement set ID from query params.
         $targetAchievementSetId = $request->query('set') ? (int) $request->query('set') : null;
 
+        // Get whether to show published or unpublished achievements from query params.
+        $targetAchievementFlag =
+            $request->query('unpublished') === 'true' || $request->query('f') === '5'
```
Jamiras: Do you see any reason to keep the `f=5` functionality? That seems like unnecessarily exposing an implementation detail.

resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.tsx
```diff
@@ -24,15 +26,42 @@ export const SidebarDevelopmentSection: FC = () => {
...
-    <PlayableSidebarButton
-      aria-pressed={isOnWantToDevList}
-      IconComponent={isOnWantToDevList ? LuCheck : LuPlus}
-      onClick={() => toggleWantToDevelop()}
-      showSubsetIndicator={game.id !== backingGame.id}
-    >
-      {backingGame.achievementsPublished ? t('Want to Revise') : t('Want to Develop')}
-    </PlayableSidebarButton>
+    <>
+      {/* TODO hide this to jr devs when more buttons are added to this section */}
```
Jamiras, line 46: The development sidebar section is currently completely missing for Junior Developers.

app/Platform/Controllers/GameController.php
wescopeland: Good find - there's indeed a race condition here. The function is trying to access array indices that may not exist anymore. You're also right about not needing to calculate this for the React page if we're not going to show the component anyway.

In latest I've fixed both things: we now fall back to zero (the function isn't necessarily tightly coupled to the UI), and we don't build achievement distribution stats for the page if we're viewing unofficial achievements.

resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.tsx
wescopeland: I think so. Done in latest. Additionally, when viewing published achievements, the link button no longer appears if there are 0 unpublished achievements.

app/Platform/Controllers/GameController.php
wescopeland: I don't have a strong opinion on this, but this same thought also occurred to me. The reason it's unrestricted is to mirror existing unrestricted functionality in the PHP/Blade UI. I'm happy to change it if you'd like me to.

app/Platform/Controllers/GameController.php
wescopeland: People manually navigate to the game page for sets they've requested with that query param out of habit in order to check progress.

Rather than treating both `?f=5` and `?unpublished=true` as the same, I've updated latest so `?f=5` redirects to `?unpublished=true`.

resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.tsx
wescopeland: I could be mistaken, but I believe this is resolved in https://github.com/RetroAchievements/RAWeb/pull/3788.

app/Platform/Controllers/GameController.php
Jamiras: I can think of a handful of reasons why a developer would want to share the unpublished achievements before publishing them. It's probably best to leave it as-is.

Jamiras (approved):

resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.tsx
Jamiras: It doesn't appear to be.
image

wescopeland: No problem. I'll tackle this issue as a separate PR from both this and 3788.

PR #3772, opened 2025-08-03
feat(tickets): add 'Unable to Debug' close reason

wescopeland:
This PR adds a new close reason to the tickets feature.

If the user's selected emulator has `can_debug_triggers` equal to `false`, the developer can now choose that they are unable to debug the ticket:
Screenshot 2025-08-03 at 8 50 53 AM


redwizard42:
I have a small worry that 'unable to debug' may get misinterpreted as something similar to 'unable to reproduce'. I don't have a better suggestion at the moment though.

redwizard42:
"Unable to debug (no dev tool support)" maybe?

redwizard42:
Those updates look great. Thanks!

public/request/ticket/update.php
```diff
@@ -70,6 +70,13 @@
         }
         break;
 
+    case TicketAction::UnableToDebug:
+        if ($permissions >= Permissions::Developer) {
+            $ticketState = TicketState::Closed;
+            $reason = "Unable to debug";
```
Jamiras: The menu option was changed to "Unable to debug due to no toolkit support". Should this be updated?

wescopeland: It should - good catch. Updated in latest.

Jamiras (approved):

PR #3773, opened 2025-08-03
feat(downloads): add ticket warning popover to emulators that don't support the toolkit

wescopeland:
Related: https://github.com/RetroAchievements/RAWeb/pull/3772.

This PR adds a popover to emulator cards on the Downloads page where `can_debug_triggers` is `false`.

A popover is used rather than a tooltip so it will also be accessible to mobile users who are unable to hover.

Screenshot 2025-08-03 at 9 01 15 AM


resources/js/features/downloads/components/DownloadableClientCard/DownloadableClientCard.tsx
```diff
@@ -52,7 +60,32 @@ export const DownloadableClientCard: FC<DownloadableClientCardProps> = ({ emulat
...
+              </BasePopoverTrigger>
+
+              <BasePopoverContent
+                onMouseEnter={() => setIsPopoverOpen(true)}
+                onMouseLeave={() => setIsPopoverOpen(false)}
+                onPointerDownOutside={(e) => e.preventDefault()}
+                onInteractOutside={(e) => e.preventDefault()}
+              >
+                {t(
+                  'Developers may not be able to easily resolve tickets submitted from this emulator.',
```
Jamiras: Tickets aren't submitted _from_ this emulator, maybe "for this emulator" or "against this emulator"?

The tickets aren't even about the emulator. Is there some way to say "for issues when using this emulator"?

resources/js/features/downloads/components/DownloadableClientCard/DownloadableClientCard.tsx
```diff
@@ -52,7 +60,32 @@ export const DownloadableClientCard: FC<DownloadableClientCardProps> = ({ emulat
       className="flex h-full flex-col light:bg-white light:shadow-sm"
     >
       <BaseCardHeader>
-        <BaseCardTitle className="text-xl">{emulator.name}</BaseCardTitle>
+        <BaseCardTitle className="flex items-center gap-2 text-xl">
+          <span>{emulator.name}</span>
+
+          {!emulator.canDebugTriggers ? (
+            <BasePopover open={isPopoverOpen}>
```
Jamiras, line 67: The cursor changes to a pointer finger, making me think I can click on the indicator. It should probably be a question mark pointer like the system chips.

resources/js/features/downloads/components/DownloadableClientCard/DownloadableClientCard.tsx
wescopeland: I went with:

> Developers may not be able to easily resolve tickets for issues encountered while using this emulator.

Happy to tweak this further if it still feels not quite right.

resources/js/features/downloads/components/DownloadableClientCard/DownloadableClientCard.tsx
wescopeland: Fixed in latest - also updated the text color so it's consistent with the other tooltips on the page.

Jamiras (approved):

PR #3775, opened 2025-08-04
fix(ResumePlayerSessionAction): gracefully re-attempt deadlocks

wescopeland:
This PR resolves a recurring error in the server logs. I believe this error is nearly impossible to reproduce locally, as it depends on prod's volume to occur.

Screenshot 2025-08-04 at 5 39 38 PM

About 1 in every 500,000 executions of `ResumePlayerSessionAction`, a deadlock occurs because two workers are simultaneously trying to update the same row.

With this PR's changes, surgical retry logic is added to the upserts to gracefully handle these deadlocks. We don't want to retry the whole job - just the thing that's locking.

### DIFF
--- a/app/Platform/Actions/ResumePlayerSessionAction.php
+++ b/app/Platform/Actions/ResumePlayerSessionAction.php
@@ -17,6 +17,7 @@ use App\Platform\Events\PlayerSessionResumed;
 use App\Platform\Events\PlayerSessionStarted;
 use App\Platform\Jobs\UpdatePlayerGameMetricsJob;
 use Carbon\Carbon;
+use Illuminate\Database\QueryException;
 
 class ResumePlayerSessionAction
 {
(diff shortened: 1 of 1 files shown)


app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -221,4 +222,43 @@ private function extendPlayTime(PlayerGame $playerGame, PlayerSession $playerSes
...
+    /**
+     * Upsert to game_recent_players with retry logic for MySQL deadlocks.
+     *
+     * With concurrent workers processing player-sessions queue (49M jobs/month),
+     * deadlocks can occur when multiple workers try to update the same user+game row.
+     * MySQL error 1213 "Deadlock found" is expected but rare (every ~3 days).
+     * We retry up to 3 times with random delay to resolve it automatically.
+     *
+     * This is a bit more surgical than putting a retry on the job.
+     * We don't want to retry everything on the job.
```
Jamiras, line 235: If two jobs are trying to do the update simultaneously, I would expect either:
1) both to generate the same results, in which case there's no reason to retry
2) both to generate unique results, in which case we should run the job again to ensure the correct state is being written.

wescopeland: Given the recent server upgrade, this bug/error may no longer even be of concern. I'll be on the lookout for if it surfaces again. Until then, I'll keep this PR closed.

PR #3788, opened 2025-08-13
feat(game): add claims system to React game pages

wescopeland:
This PR adds full support for the claims system to React game pages. A link to the Claim History page has been intentionally omitted for the time being - this should be migrated to Filament.

This PR covers the following functionality:
* Primary and collaboration claim creation.
* Claim extension (when expiring within 7 days).
* Claim dropping (with In Review restrictions in place).
* Claim completion (with various warnings and validations).
* Toggling In Review statuses.
* Applying special treatment for collaboration claims.
* Applying special treatment for revision claims.

Behavior change: given the user has more than 2 unresolved tickets, they cannot create a new claim.

The existing API layer in `AchievementSetClaimController` is re-used, however, additional security policy validations have been put in place, especially on the `update()` method.

### Claim Creation
Given the user has the ability to make a claim, they can create a new claim from two places on the game page: the empty state and the sidebar. The empty state will not always be visible (eg: revisions), so the sidebar is used as a fallback.

Screenshot 2025-08-12 at 8 16 08 PM

Screenshot 2025-08-12 at 8 17 00 PM

Upon clicking the button, a dialog appears:
Screenshot 2025-08-12 at 8 17 42 PM

Various warnings are supported in the code:
```tsx
{wouldBeRevisionClaim && !isSoleAuthor ? <RevisionPlanWarning /> : null}
{wouldBeSubsetClaim && !wouldBeRevisionClaim ? <SubsetApprovalWarning /> : null}
{claimData?.numUnresolvedTickets ? <UnresolvedTicketsWarning /> : null}
{!game.forumTopicId ? <ForumTopicNotice /> : null}
```

Upon confirming, the API call is made to the back-end to create the claim.

### Dropping the Claim
Similar to creation, users can drop claims from both the empty state and the sidebar.
Screenshot 2025-08-12 at 8 19 22 PM

Screenshot 2025-08-12 at 8 19 24 PM

On click, a confirm dialog appears:
Screenshot 2025-08-12 at 8 19 45 PM

Given the claim is in review, the drop claim button is not visible in the sidebar (in review revisions are an edge case) and the empty state button appears as disabled with a tooltip:
Screenshot 2025-08-12 at 8 21 51 PM

### Extending the Claim
Given the claim expires within 7 days, the empty state and sidebar again shift their UX:
Screenshot 2025-08-12 at 8 23 38 PM

Screenshot 2025-08-12 at 8 23 42 PM

On click, the user sees a confirm dialog:
Screenshot 2025-08-12 at 8 23 59 PM

### Creating Revision Claims
Given the user wishes to create a revision claim, the only create claim button that's accessible is within the sidebar:
Screenshot 2025-08-12 at 8 25 36 PM

On click, the user sees this confirm dialog:
Screenshot 2025-08-12 at 8 25 53 PM

### In Review Status
Users with either a moderator or code reviewer role can toggle a Junior Developer's claim to In Review status:
Screenshot 2025-08-12 at 8 27 16 PM

Screenshot 2025-08-12 at 8 27 32 PM

### DIFF
--- a/resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.test.tsx
+++ b/resources/js/features/games/components/GameSidebarFullWidthButtons/SidebarDevelopmentSection/SidebarDevelopmentSection.test.tsx
@@ -218,11 +218,13 @@ describe('Component: SidebarDevelopmentSection', () => {
 
   it('given the user is viewing unpublished achievements, shows a link to view published achievements', () => {
     // ARRANGE
-    const game = createGame({ id: 1 });
+    const game = createGame({ id: 1, gameAchievementSets: [] });
     const backingGame = createGame({ id: 2 });
     const pageProps = {
       backingGame,
       game,
+      achievementSetClaims: [],
+      can: {},
       isOnWantToDevList: false,
       isViewingPublishedAchievements: false, // !!
       ziggy: createZiggyProps(),
@@ -236,7 +238,7 @@ describe('Component: SidebarDevelopmentSection', () => {
 
   it('given the user is viewing published achievements and there are unpublished achievements, shows the count of unpublished achievements', () => {
     // ARRANGE
-    const game = createGame({ id: 1 });
+    const game = createGame({ id: 1, gameAchievementSets: [] });
     const backingGame = createGame({
       id: 2,
       achievementsPublished: 50,
@@ -245,6 +247,8 @@ describe('Component: SidebarDevelopmentSection', () => {
     const pageProps = {
       backingGame,
       game,
+      achievementSetClaims: [],
+      can: {},
       isOnWantToDevList: false,
       isViewingPublishedAchievements: true, // !!
       ziggy: createZiggyProps(),
@@ -260,7 +264,7 @@ describe('Component: SidebarDevelopmentSection', () => {
 
   it('given the user is viewing unpublished achievements, shows the count of published achievements', () => {
     // ARRANGE
-    const game = createGame({ id: 1 });
+    const game = createGame({ id: 1, gameAchievementSets: [] });
     const backingGame = createGame({
       id: 2,
       achievementsPublished: 75, // !!
@@ -269,6 +273,8 @@ describe('Component: SidebarDevelopmentSection', () => {
     const pageProps = {
       backingGame,
       game,
+      achievementSetClaims: [],
+      can: {},
       isOnWantToDevList: false,
       isViewingPublishedAchievements: false, // !!
       ziggy: createZiggyProps(),
@@ -284,7 +290,7 @@ describe('Component: SidebarDevelopmentSection', () => {
 
   it('given the user is viewing published achievements and there are no unpublished achievements, hides the link button to view unpublished achievements', () => {
     // ARRANGE
-    const game = createGame({ id: 1 });
+    const game = createGame({ id: 1, gameAchievementSets: [] });
     const backingGame = createGame({
       id: 2,
       achievementsPublished: 50,
@@ -293,6 +299,8 @@ describe('Component: SidebarDevelopmentSection', () => {
     const pageProps = {
       backingGame,
       game,
+      achievementSetClaims: [],
+      can: {},
       isOnWantToDevList: false,
       isViewingPublishedAchievements: true, // !!
       ziggy: createZiggyProps(),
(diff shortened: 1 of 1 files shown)


drisc:
Is there a way to create or link a forum topic that doesn't involve claiming the game on this version of the page?

Code Reviewers usually create the topic so that a claim can be made so unless a Jr is able to create the topic this would mean the CR would show up as the initial claimant in the history.

wescopeland:
https://github.com/RetroAchievements/RAWeb/blob/master/resources/js/features/games/components/GameSidebarFullWidthButtons/GameSidebarFullWidthButtons.tsx#L104-L106

app/Policies/AchievementSetClaimPolicy.php
Jamiras, line 1: `create` should validate that the user still has available claims.

User with 3 claims opens two games. Claim button is enabled on both. The user can then click button on both and ends up with extra claims. (I have not tested this in the existing pages - it may or may not exist there too).

Image

resources/js/features/games/components/AchievementSetEmptyState/ClaimActionButton/ClaimActionButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  if (!claimData?.numClaimsRemaining && !claimData?.isSoleAuthor) {
+    return (
+      <BaseTooltip>
+        <BaseTooltipTrigger asChild>
+          <DisabledButton />
+        </BaseTooltipTrigger>
+
+        <BaseTooltipContent>{t("You've used all your achievement set claims.")}</BaseTooltipContent>
```
Jamiras, line 76: I can't see this tooltip. When the button is disabled, the cursor changes to a NO sign, but there's no tooltip.

Jamiras: Similar comment around other tooltips for the button being disabled.

app/Policies/AchievementSetClaimPolicy.php
wescopeland: This should now be enforced in the policy in latest.

resources/js/features/games/components/AchievementSetEmptyState/ClaimActionButton/ClaimActionButton.tsx
wescopeland: Fixed in latest.

app/Policies/AchievementSetClaimPolicy.php
```diff
@@ -31,22 +34,93 @@ public function view(?User $user, AchievementSetClaim $achievementSetClaim): boo
...
+        // First, check the basic role requirement.
+        if (!$user->hasAnyRole([Role::DEVELOPER, Role::DEVELOPER_JUNIOR])) {
+            return false;
+        }
+
+        // If no game is provided, we'll just be happy the user passes role verification.
+        if (!$game) {
+            return true;
+        }
+
```
Jamiras, line 48: Would it make sense to disallow claiming games for unsupported systems?
http://localhost:64000/game2/7775

```
        $game->loadMissing('system');
        if (!$game->system->active) {
            return false;
        }
```

resources/js/features/games/components/AchievementSetEmptyState/ClaimActionButton/ClaimActionButton.tsx
Jamiras: The button is no longer visible if the user is out of claims, so there's no way to get to this tooltip.

The tooltip for Jr Dev missing forum topic does work.

app/Policies/AchievementSetClaimPolicy.php
wescopeland: To an extent, I think it would. The only thing I'm not sure about is how to handle systems that are currently undergoing a rollout. Aren't they still considered unsupported? It almost seems like we may need some kind of `are_achievement_sets_claimable` flag on the system model.

Jamiras: Marking it as active would allow publishing of achievements, so it's probably best not to restrict it at this time.

Instead of adding a new flag, maybe we could expand the meaning of active (see #1065)

resources/js/features/games/components/AchievementSetEmptyState/ClaimActionButton/ClaimActionButton.tsx
wescopeland: Should now be fixed in latest:
Screenshot 2025-08-16 at 1 22 22 PM

Jamiras (approved):
Unrelated to this PR, but the Set Request information is still visible when not logged in:

Image

wescopeland:
> Unrelated to this PR, but the Set Request information is still visible when not logged in:

Should we disable this? It seems non-risky to expose to guests. I'm almost wondering if someone is a guest whether we should enable the "Request Set" button and have that direct to the login page.

PR #3793, opened 2025-08-15
don't double-escape title for ticket page

Jamiras:
http://localhost:64000/ticket/90528
image


wescopeland (approved):
Verified locally with an attempted XSS.

PR #3795, opened 2025-08-16
feat: Added sort by award date functionality

ludvigrosenqvist:
This PR adds a minor feature to the "Reorder Site Awards" page that I have personally been wanting for a while, and hopefully other users will find it useful as well.

The feature provides users with an easy way to sort their awards in ascending or descending order by date of acquisition. For some users like me, looking back at awards in order of completion serves as a nice trip down memory lane. But like many other users, I often want to re-sort my awards by game series or other preferences. However, I felt that an option for easily resetting awards to the "default" order was missing.

I've tested the functionality locally and it works for all three award tables: game awards, event awards, and site awards. Users still need to press the "Save All Changes" button for the sort to take effect, as the help text suggests.

bild

bild

### DIFF
--- a/app/Helpers/render/site-award.php
+++ b/app/Helpers/render/site-award.php
@@ -317,6 +317,15 @@ function RenderAwardOrderTable(
 
     echo "<div class='flex w-full items-center justify-between'>";
     echo "<h4>$title</h4>";
+    echo <<<HTML
+        <div class="flex items-center gap-x-1 ml-auto mr-4">
+            <p class="text-3xs">
+                Sort by Award Date:
+            </p>
+            <button onclick='reorderSiteAwards.handleSortByAwardDate("asc", "$humanReadableAwardKind")' class="btn text-3xs py-0.5">↑</button>
+            <button onclick='reorderSiteAwards.handleSortByAwardDate("desc", "$humanReadableAwardKind")' class="btn text-3xs py-0.5">↓</button>
+        </div>
+    HTML;
     echo "<select data-award-kind='$humanReadableAwardKind'>";
     for ($i = 1; $i <= $renderedSectionCount; $i++) {
         if ($initialSectionOrder === $i) {
@@ -344,6 +353,7 @@ function RenderAwardOrderTable(
         $awardType = $award['AwardType'];
         $awardData = $award['AwardData'];
         $awardDataExtra = $award['AwardDataExtra'];
+        $awardDate = $award['AwardedAt'];
         $awardTitle = $award['Title'];
         $awardDisplayOrder = $award['DisplayOrder'];
 
@@ -380,6 +390,7 @@ function RenderAwardOrderTable(
             <tr
                 data-row-index='$awardCounter'
                 data-award-kind='$humanReadableAwardKind'
+                data-award-date='$awardDate'
                 draggable='$isDraggable'
                 class='$rowClassNames'
                 ondragstart='reorderSiteAwards.handleRowDragStart(event)'
--- a/resources/js/tall-stack/dynamic/reorderSiteAwards.ts
+++ b/resources/js/tall-stack/dynamic/reorderSiteAwards.ts
@@ -1,5 +1,6 @@
 export * from './reorderSiteAwards/handleRowDragOver';
 export * from './reorderSiteAwards/handleShowSavedHiddenRowsChange';
+export * from './reorderSiteAwards/handleSortByAwardDate';
 export * from './reorderSiteAwards/index';
 export * from './reorderSiteAwards/refreshVisibilityAfterSave';
 export { reorderSiteAwardsStore as store } from './reorderSiteAwards/reorderSiteAwardsStore';
--- a/resources/js/tall-stack/dynamic/reorderSiteAwards/handleSortByAwardDate.ts
+++ b/resources/js/tall-stack/dynamic/reorderSiteAwards/handleSortByAwardDate.ts
@@ -1,0 +1,38 @@
+export function handleSortByAwardDate(
+    direction: "asc" | "desc",
+    awardKind: string
+): void {
+    // Determine the correct table by awardKind
+    const tableId = `${awardKind}-reorder-table`;
+    const table = document.getElementById(tableId) as HTMLTableElement | null;
+    const tbody = table ? table.querySelector("tbody") : null;
+    if (!tbody) return;
+
+    // Only select rows for this awardKind in this table
+    const rows = Array.from(
+        tbody.querySelectorAll<HTMLTableRowElement>(".award-table-row")
+    ).filter(
+        (row) => row.getAttribute("data-award-kind") === awardKind
+    );
+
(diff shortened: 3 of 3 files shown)


resources/views/pages-legacy/reorderSiteAwards.blade.php
```diff
@@ -108,6 +82,9 @@ function postAllAwardsDisplayOrder(awards) {
 
             <div class="flex w-full items-center justify-between mt-3 mb-6">
                 <div class="flex items-center gap-x-1">
+                    <div class="ml-auto mr-4">
```
wescopeland: CSS: I think we should remove this div and handle element positioning with flexbox. It seems like what you're trying to create is a 1rem gap between the button and the checkbox. This is more maintainable when lifted up into a `gap` class, eg:

```html
<div class="flex items-center gap-4">
    <button onclick="...">Reset Order</button>
    <label class="flex...">
        <input ... />
    </label>
</div>
```

resources/views/pages-legacy/reorderSiteAwards.blade.php
wescopeland, line 1: CSS: Mobile on this page isn't particularly spectacular, but the mobile UX for this control might need a small touchup.

Screenshot 2025-08-16 at 7 01 27 PM

---

Initial ideas:
* Remove "text-base" from the "Save All Changes" button.
* At the XS breakpoint, put the button and checkbox in a single flex-col wrapper, ie: `flex flex-col sm:flex-row sm:items-center gap-4`.

resources/views/pages-legacy/reorderSiteAwards.blade.php
```diff
@@ -108,6 +82,9 @@ function postAllAwardsDisplayOrder(awards) {
...
+                         <button onclick='reorderSiteAwards.handleResetOrder()' class="btn text-3xs py-0.6">Reset Order</button>
```
wescopeland: a11y: It only now occurs to me, in the button's present form, it's a destructive action.

We may want to add a "btn-danger" class to this element.

It may also be desirable to swap its position with the checkbox in case anyone reaches for the checkbox out of muscle memory.

resources/js/tall-stack/dynamic/reorderSiteAwards/handleResetOrder.ts
```diff
@@ -0,0 +1,52 @@
...
+import { computeDisplayOrderValues } from './index';
+
+declare global {
+  interface Window {
+    postAllAwardsDisplayOrder: (awards: Partial<MappedTableRow>[]) => void;
+  }
+}
+
+export function handleResetOrder(): void {
+  if (!confirm('This will reset the order of all awards. Are you sure?')) {
```
wescopeland: Ambiguity: this will indeed reset the order, but could be misinterpreted to mean "it will undo the changes I've made in my current session". I recommend a more precise confirmation:

> This will resort all your awards by the date they were earned (oldest first). You can preview the changes before saving.

The "preview" bit is related to another comment I'll leave in the review.

resources/js/tall-stack/dynamic/reorderSiteAwards/handleResetOrder.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  for (const row of rows) {
+    const awardKind = row.getAttribute('data-award-kind');
+    document
+      .querySelector<HTMLTableElement>(`#${awardKind}-reorder-table`)
+      ?.querySelector('tbody')
+      ?.appendChild(row);
+  }
+
+  const mappedTableRows = collectMappedTableRows();
```
wescopeland: Performance: `collectMappedTableRows` requeries all rows.

For most users this is not really a big deal. However, some users have 1,000+ rows rendered to the page.

I believe `rows` is virtually identical to `collectMappedTableRows::awardTableRowEls`. Can it be passed as an arg?

resources/js/tall-stack/dynamic/reorderSiteAwards/handleResetOrder.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  window.postAllAwardsDisplayOrder(computeDisplayOrderValues(mappedTableRows));
```
wescopeland: UX: I recommend we remove this line.

The page has a "Save All Changes" button which users expect to use for saving. Having this button auto-save while other operations don't might be considered a dishonest design.

After seeing their awards resorted, users might decide they don't like what they see. By requiring the click to Save All Changes, users can discard the change if they decide they don't want to go through with it.

resources/js/tall-stack/dynamic/reorderSiteAwards/handleResetOrder.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    if (!isNaN(Number(dateA)) && !isNaN(Number(dateB))) {
+      numA = parseInt(dateA, 10);
+      numB = parseInt(dateB, 10);
+    } else {
+      numA = Date.parse(dateA) || 0;
+      numB = Date.parse(dateB) || 0;
+    }
+
+    return numA - numB;
```
wescopeland: Minor: I believe `numA` and `numB` are unnecessarily mutable.

```sql
-- production db query

SELECT * FROM SiteAwards WHERE AwardDate IS NULL;
-- 0 rows returned
```

I _think_ this means `data-award-date` should always have a value. Just to be extra safe, we can fall back to "0":
```ts
const dateA = a.getAttribute('data-award-date') ?? '0';
const dateB = b.getAttribute('data-award-date') ?? '0';

const numA = parseInt(dateA, 10);
const numB = parseInt(dateB, 10);

return numA - numB;
```

resources/views/pages-legacy/reorderSiteAwards.blade.php
ludvigrosenqvist: Decided to give the reset button  `self-start w-fit` since I felt it looked a bit nicer than without it:

With: 
bild

Without:
bild

resources/js/tall-stack/dynamic/reorderSiteAwards/handleResetOrder.ts
ludvigrosenqvist: I removed the saving functionality from the reset function (which you're totally right about from an UX-perspective), thus this line is no longer needed. However, I kept the collectMappedTableRows function as a separate TypeScript file, as I believe that aligns with the project's direction.

wescopeland (approved):
Thanks!

resources/js/tall-stack/dynamic/reorderSiteAwards/handleResetOrder.ts
```diff
@@ -0,0 +1,34 @@
+export function handleResetOrder(): void {
+  if (
+    !confirm(
+      'This will resort all your awards by the date they were earned (oldest first). You can preview the changes before saving.',
```
Jamiras, line 4: I wonder if we should mention the fact that it will unhide any hidden awards too.

PR #3796, opened 2025-08-16
feat(forums): allow users to create topics, reply, and edit posts on behalf of team accounts

wescopeland:
This PR allows users with access to team accounts to post new threads, post replies, and edit replies on behalf of that team account. This is similar to what we currently support with accessible team DM inboxes, but is a bit more robust due to the additional functionality required to bring this feature fully to life.

**A migration is included in this PR.**

**How it works**
As a user belonging to several teams that have dedicated team accounts, when I want to start a new forum topic, the form now has a new visible field: "Post as". This field lets me post as either myself (the default option) or any of those teams:
Screenshot 2025-08-16 at 12 33 28 PM

When I select a team, the submit button label changes to also show that team account's avatar and display name. This enhancement has also been ported to the team inboxes feature:
Screenshot 2025-08-16 at 12 34 28 PM

Similar to team messages, a "Posted by" indicator appears on the post:
Screenshot 2025-08-16 at 12 35 07 PM

You'll also notice the "Edit" button is visible. Team members can continue editing the post. The migration adds two columns to `forum_topic_comments`: `sent_by_id` and `edited_by_id`. This means we don't have to overwrite the `sent_by_id` metadata when someone decides to edit the post (very common with RAEvents).

When a team member edits the post, an additional indicator appears. Obviously, **these posted by / edited by indicators are only visible to team members**, and the props values are set to null unless the current auth'd user has permission to see them:

Screenshot 2025-08-16 at 12 37 50 PM

Users can also reply on behalf of their team accounts via the quick reply form at the bottom of the page. If the user belongs to any team accounts, two behavior changes will kick in:
* A "Post as" select control will always appear.
* The user's avatar will _always_ appear within the Submit button. This is an intentional safeguard to prevent accidental submits from the wrong account.

Screenshot 2025-08-16 at 12 39 08 PM

Screenshot 2025-08-16 at 12 39 37 PM

Screenshot 2025-08-16 at 12 39 44 PM


config/teams.php
wescopeland, line 1: There's a lot of noise in this PR. Want to raise attention to this file specifically as it contains a TODO regarding future functionality.

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
```diff
@@ -56,6 +56,34 @@ export const ForumPostCardMeta: FC<ForumPostCardMetaProps> = ({ comment, topic }
       ) : null}
 
       <ForumPostCardTimestamps comment={comment} />
+
+      {comment.sentBy ? (
+        <>
+          <span className="text-neutral-400/80 light:text-neutral-700">{'·'}</span>
+
+          <span className="flex items-center gap-1.5 text-2xs text-neutral-400/80 light:text-neutral-700">
```
Jamiras: The vertical alignment of this seems off.

Image

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
```diff
@@ -56,6 +56,34 @@ export const ForumPostCardMeta: FC<ForumPostCardMetaProps> = ({ comment, topic }
...
+        </>
+      ) : null}
+
+      {comment.editedBy && comment.editedBy.displayName !== comment.sentBy?.displayName ? (
+        <>
+          <span className="text-neutral-400/80 light:text-neutral-700">{'·'}</span>
+
+          <span className="flex items-center gap-1.5 text-2xs text-neutral-400/80 light:text-neutral-700">
+            <Trans
+              i18nKey="Edited by <1>{{username}}</1>"
```
Jamiras: Should this be "Last edited by"?

app/Community/Controllers/Api/ForumTopicCommentApiController.php
```diff
@@ -46,6 +71,14 @@ public function update(
         $newPayload = Shortcode::convertLegacyGameHubShortcodesToHubShortcodes($newPayload);
 
         $comment->body = $newPayload;
+
+        // If this is a team account post being edited by a team member,
+        // update edited_by_id to track who made the edit.
+        $accessibleTeamIds = (new ForumTopicCommentPolicy())->getAccessibleTeamIds($requestingUser);
+        if (in_array($comment->author_id, $accessibleTeamIds, true)) {
+            $comment->edited_by_id = $requestingUser->id;
+        }
```
Jamiras, line 78: Edits made by a moderator aren't captured. I could see that as intended behavior, and also unlikely.

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
```diff
@@ -56,6 +56,34 @@ export const ForumPostCardMeta: FC<ForumPostCardMetaProps> = ({ comment, topic }
...
+            />
+          </span>
+        </>
+      ) : null}
+
+      {comment.editedBy && comment.editedBy.displayName !== comment.sentBy?.displayName ? (
+        <>
+          <span className="text-neutral-400/80 light:text-neutral-700">{'·'}</span>
+
+          <span className="flex items-center gap-1.5 text-2xs text-neutral-400/80 light:text-neutral-700">
```
Jamiras: It steps up again (though a much smaller step) for the edited by label:

Image

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
wescopeland: Looks like this was caused by a discrepancy in font sizes (at some point, I hope we can get rid of that "smalltext" class - it doesn't play nicely with Tailwind). Here's what latest looks like:

Screenshot 2025-08-25 at 7 17 06 PM

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
wescopeland: I think that's probably more accurate - updated in latest.

app/Community/Controllers/Api/ForumTopicCommentApiController.php
wescopeland: I hadn't even considered that use case. It's probably valuable to capture and display. I've made some adjustments in latest so this will happen.

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
wescopeland: Fixed in latest.

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
Jamiras: Edited by seems to be aligned with Posted by, but Posted by still appears to be misaligned.

image

app/Community/Actions/BuildShowForumTopicPagePropsAction.php
```diff
@@ -57,18 +60,46 @@ public function execute(
...
+        $forumTopicComments = $paginatedForumTopicComments->getCollection()->map(
+            function ($comment, $index) use ($updatedBodies, $user, $accessibleTeamIds) {
+                $comment->body = $updatedBodies[$index];
+
+                $includes = ['user.createdAt', 'user.deletedAt', 'user.visibleRole'];
+
+                /**
+                 * Include the sentBy and editedBy values if:
+                 * A. The user is viewing a team account post they have access to, OR
+                 * B. The user can manage forum posts (they can moderate the forum).
```
Jamiras, line 84: This enables the edited by functionality for non-team account posts, and there doesn't appear to be any way to specify that the edit occured from a team account.

i.e. As a moderator, I edit a generic user's post to remove a rom link:

Image

I think this is a good change, but there should be some way to say the edit was done by RAdmin or something similar.

wescopeland: Doesn't doing so obscure who made the edit, though? "Last edited by RAdmin" doesn't seem very helpful if this is supposed to provide some kind of trace of who is editing things.

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
wescopeland: As best as I can tell, the text is now perfectly centered along the Y axis of the container:
Screenshot 2025-08-26 at 12 34 24 PM

There could be a slight optical illusion happening because the user avatars are what the text is centered against (as they're the tallest element in the container).

app/Community/Actions/BuildShowForumTopicPagePropsAction.php
Jamiras: I guess it depends on who it's visible to. If it's only visible to moderators, then the behavior is correct.

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
Jamiras: image

app/Community/Actions/BuildShowForumTopicPagePropsAction.php
wescopeland: That's definitely what's intended:
```php
/**
 * Include the sentBy and editedBy values if:
 * A. The user is viewing a team account post they have access to, OR
 * B. The user can manage forum posts (they can moderate the forum).
 * If we always naively include it, Inertia will leak the value into the DOM.
 */
$shouldIncludeSentByEditedBy = $user && (
    ($comment->sent_by_id !== null && in_array($comment->author_id, $accessibleTeamIds, true))
    || ($comment->edited_by_id !== null && (new ForumTopicCommentPolicy())->manage($user))
);
```

resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardMeta.tsx
wescopeland: Latest shifts the "Posted by" and "Last edited by" spans down by 1 pixel.

PR #3798, opened 2025-08-16
fix(notifications): respect user preferences for achievement comment emails

### LINKED ISSUE #430: (Un)Subscribing to a game's tickets/comments doesn't work
Under the Dev tools on [Pokemon Pinball](https://retroachievements.org/game/725) it says "Unsubscribe from Tickets" implying I am subscribed.
However, I just found [a ticket](https://retroachievements.org/ticketmanager.php?i=25056) that was opened a few days ago on [an achievement](https://retroachievements.org/achievement/29870) for that game made by another dev. I never received an email despite being subscribed.

On the other side, I tried to unsubscribe from achievement comments on [Tomba](https://retroachievements.org/game/11276) since another dev agreed to take over my set when adding regional support.
However, my email is filled with hundreds of "New Achievement Comment" emails from them editing the achievements.
NewAchievementComment

wescopeland:
Resolves #430.

Screenshot 2025-08-16 at 6 45 08 PM

This PR fixes issues preventing users from unsubscribing from achievement comment email notifications.

**Root Cause**
`getSubscribersOfArticle()` has two problems:
1. The function hardcodes `EmailOn_ActivityComment` rather than using the passed `$reqWebsitePrefs` (already identified as an issue by a prior comment left in the function).
2. The `UNION` query that adds achievement authors wasn't checking the author's actual email preferences.

**Behavior Changes**
This PR makes two targeted changes to resolve both of these problems, resulting in two related issues being resolved:
1. Achievement authors can now unsubscribe from their own achievements on the _achievementInfo.blade.php_ page. This is done by pressing the subscribe toggle button enough times to force an unsubscribe state (we'll clean up this UX in a React conversion later). Previously, they received email notifications regardless of their record in the `subscriptions` table.
2. Achievement authors can now uncheck "Comments on an achievement I created" on the `/settings` page and it will work correctly.


app/Helpers/database/subscription.php
```diff
@@ -160,7 +180,7 @@ function getSubscribersOfArticle(
         SELECT _ua.*
         FROM UserAccounts as _ua
         WHERE (_ua.User = :subjectAuthor OR _ua.display_name = :subjectAuthor)
-              $websitePrefsFilter
+              AND (_ua.websitePrefs & :websitePrefs) != 0
```
Jamiras: This change seems possibly wrong. Previously, if `$noExplicitSubscriptions` was false, this subclause would be blank.

If it is desirable to stil have it, then `$websitePrefsFilter` should be eliminated.

wescopeland: Confirmed. Latest sets this line back to using `$websitePrefsFilter`.

PR #3799, opened 2025-08-17
feat: add Patreon Supporters page

wescopeland:
https://discord.com/channels/476211979464343552/1002696640001478769/1405510186633793548

This PR adds a dedicated page to showcase our list of Patreon supporters.

From a technical POV, this is also an experiment with Inertia deferred props, a concept I probably want to try on the React game pages for some of the data. There are too many users to fit into page props (1000+), so we load the first 100 initially, then use `Inertia::defer()` in the controller to automatically fetch the remaining during client-side hydration.

Screenshot 2025-08-17 at 10 37 20 AM

Screenshot 2025-08-17 at 10 37 38 AM

cc @ScottFromDerby @SporyTike


wescopeland:
In https://github.com/RetroAchievements/RAWeb/pull/3799/commits/32e56d3ed2bfc79aaedfa0019a03952791545d74, the Patreon badge now links to the new page.

For the footer - I think we should probably keep that as-is for now.

Jamiras (approved):

PR #3801, opened 2025-08-17
feat: enhance Discord team inbox webhooks

wescopeland:
This PR completely refactors how messages are forwarded to Discord team inboxes, ultimately making two key enhancements.

**A migration is included in this PR.**

**1. Messages longer than 2000 characters are now chunked in the same thread.**
Users are no longer left with a truncated message which requires someone to manually go into the inbox and paste the rest into the thread. The message is now chunked, and subsequent webhook calls are made for each chunk so the entire message content appears in a single thread:
Screenshot 2025-08-17 at 4 23 03 PM

**2. Replies made by the sender no longer create new threads.**
The migration included in this PR creates a mapping table between our message thread IDs and Discord thread IDs. Replies can then be appended to Discord threads rather than us needing to create new ones.

Screenshot 2025-08-17 at 4 24 13 PM

https://discord.com/channels/476211979464343552/1406734407943065660

### DIFF
--- a/app/Community/Actions/ForwardMessageToDiscordAction.php
+++ b/app/Community/Actions/ForwardMessageToDiscordAction.php
@@ -1,0 +1,435 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Community\Actions;
+
+use App\Community\Data\ProcessedDiscordMessageData;
+use App\Models\DiscordMessageThreadMapping;
+use App\Models\Message;
+use App\Models\MessageThread;
+use App\Models\User;
+use App\Support\Shortcode\Shortcode;
+use GuzzleHttp\Client;
+
+class ForwardMessageToDiscordAction
+{
+    /** Discord API limits */
+    private const DISCORD_EMBED_DESCRIPTION_LIMIT = 2000;
+    private const DISCORD_THREAD_NAME_LIMIT = 100;
+
+    /** Message processing limits */
+    private const MESSAGE_BODY_MAX_LENGTH = 10_000;
+
+    /** Timing constants */
+    private const CHUNK_SEND_DELAY_MICROSECONDS = 100_000;
+
+    /** Embed colors */
+    private const COLOR_DEFAULT = 0x0066CC;
+    private const COLOR_VERIFICATION = 0x00CC66;
+
+    private Client $client;
+
+    public function __construct(?Client $client = null)
+    {
+        // Optionally inject a mock client in tests to avoid real Discord API calls.
+        $this->client = $client ?? new Client();
+    }
+
+    /**
+     * Forward a message to Discord via webhook.
+     */
+    public function execute(
+        User $userFrom,
+        User $userTo,
+        MessageThread $messageThread,
+        Message $message
+    ): void {
+        $fullBody = Shortcode::stripAndClamp($message->body, self::MESSAGE_BODY_MAX_LENGTH, preserveWhitespace: true);
+
+        $inboxConfig = config('services.discord.inbox_webhook.' . $userTo->username);
+        if ($inboxConfig === null || empty($inboxConfig['url'] ?? null)) {
+            return;
+        }
+
+        $webhookUrl = $inboxConfig['url']; // each inbox has its own dedicated webhook url
+
+        if (empty($messageThread->title) || empty($fullBody)) {
+            return;
+        }
+
+        $color = self::COLOR_DEFAULT;
+        $isForum = $inboxConfig['is_forum'] ?? false;
+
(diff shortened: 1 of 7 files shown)


database/migrations/2025_08_17_000000_create_discord_message_thread_mappings_table.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up(): void
+    {
+        Schema::create('discord_message_thread_mappings', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedBigInteger('message_thread_id');
+            $table->unsignedBigInteger('recipient_id');
```
Jamiras: Would it make sense to not filter on `recipient_id`? Shouldn't all messages in the RA thread be forwarded to the discord discussion?

app/Community/Actions/ForwardMessageToDiscordAction.php
```diff
@@ -0,0 +1,435 @@
...
+    /**
+     * Forward a message to Discord via webhook.
+     */
+    public function execute(
+        User $userFrom,
+        User $userTo,
+        MessageThread $messageThread,
+        Message $message
+    ): void {
+        $fullBody = Shortcode::stripAndClamp($message->body, self::MESSAGE_BODY_MAX_LENGTH, preserveWhitespace: true);
```
Jamiras: suggest moving this down after the check for the presence of a hook.

app/Community/Actions/ForwardMessageToDiscordAction.php
```diff
@@ -0,0 +1,435 @@
...
+        $webhookUrl = $inboxConfig['url']; // each inbox has its own dedicated webhook url
+
+        if (empty($messageThread->title) || empty($fullBody)) {
+            return;
+        }
+
+        $color = self::COLOR_DEFAULT;
+        $isForum = $inboxConfig['is_forum'] ?? false;
+
+        $isNewThread = $isForum ? !$this->getExistingDiscordThreadId($messageThread, $userTo) : true;
```
Jamiras: Suggest capturing the `extistingThreadId` instead of looking it up again in `sendDiscordWebhooks()`

app/Community/Actions/ForwardMessageToDiscordAction.php
```diff
@@ -0,0 +1,435 @@
...
+    ): array {
+        $embed = [
+            'description' => $description,
+            'color' => $color,
+        ];
+
+        if ($includeAuthor) {
+            $embed['author'] = [
+                'name' => $userFrom->display_name,
+                // TODO 'url' => route('user.show', $userFrom), -- we've lost context on why this TODO was ever added.
```
Jamiras: I think it's just a request to use a route instead of hard-coding a path. I'm not sure why it wasn't possible at the time the comment would have been added.

config/services.php
```diff
@@ -75,7 +75,6 @@
                     env('DISCORD_ROLE_ADMIN'),
                     env('DISCORD_ROLE_MODERATOR'),
                 ],
-                'manual_unlock_url' => env('DISCORD_WEBHOOK_MOD_MANUAL_UNLOCK'),
```
Jamiras: This is still present in the server `.env`. Is it not used?

database/migrations/2025_08_17_000000_create_discord_message_thread_mappings_table.php
wescopeland: Yea, this is probably overengineered. Removed this column from the migration & implementation in latest.

app/Community/Actions/ForwardMessageToDiscordAction.php
wescopeland: Done in latest

app/Community/Actions/ForwardMessageToDiscordAction.php
wescopeland: Good point. Made this optimization in latest.

app/Community/Actions/ForwardMessageToDiscordAction.php
wescopeland: I've verified this works in latest, so I've swapped to using `route()` and removed the TODO comment.

config/services.php
wescopeland: Unused - I've ripped it out of the server `.env`.

wescopeland: Correction: it turns out this is still used. I need to restore it in this branch. I'll take care of that later today.

wescopeland: Manual unlock message handling has been restored in latest (as well as in the server env) and covered with a new test case.

PR #3802, opened 2025-08-17
feat(api): add an achievement PATCH endpoint to the service API

wescopeland:
RABot is ultimately going to need the ability to demote achievements on behalf of a team as part of some workflow behavior. There will be several Lego pieces required for RABot to fully execute a workflow (send DMs, etc). This PR adds one of those Lego pieces.

A new endpoint has been added which follows the JSON:API specification:

```bash
curl --request PATCH \
  --url https://retroachievements.org/api/internal/achievements/42 \
  --header 'Content-Type: application/json' \
  --header 'X-API-Key: redacted' \
  --data '{
	"data": {
		"type": "achievement",
		"id": "67",
		"attributes": {
			"published": false,
			"title": "This is a new title!"
		},
		"meta": {
			"actingUser": "DevCompliance"
		}
	}
}'
```


app/Api/Internal/Controllers/AchievementController.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $wasTitleUpdated = false;
+
+        DB::transaction(function () use ($achievement, $user, $newTitle, &$wasTitleUpdated) {
+            updateAchievementFlag($achievement->id, AchievementFlag::Unofficial);
+
+            if ($newTitle !== null && $newTitle !== $achievement->title) {
+                $achievement->title = $newTitle;
+                $achievement->save();
+                $wasTitleUpdated = true;
```
Jamiras: I realize both Flags and Title are being logged in the management audit log, but only the demotion is being logged in the public audit comments. Previously a title change would also be logged in the public audit comments.

app/Api/Internal/Controllers/AchievementController.php
```diff
@@ -0,0 +1,102 @@
...
+     * }
+     * ```
+     */
+    public function demote(DemoteAchievementRequest $request): JsonResponse
+    {
+        $achievementId = $request->getAchievementId();
+        $username = $request->getUsername();
+        $newTitle = $request->getTitle();
+
+        $user = User::whereName($username)->first();
```
Jamiras, line 78: I don't remember - is the acting user (associated to the API key) going to be the bot, or the user initiating the command through the bot? Should there be some sort of Role check to ensure the acting user has permission to demote the achievement?

There does need to be more restrictions on the `username` parameter. I can currently call this method with any user - including non-devs, and banned users. If I pass a deleted user, it throws an exception.

app/Api/Internal/Controllers/AchievementController.php
```diff
@@ -0,0 +1,102 @@
...
+        $achievement = Achievement::find($achievementId);
+
+        $wasTitleUpdated = false;
+
+        DB::transaction(function () use ($achievement, $user, $newTitle, &$wasTitleUpdated) {
+            updateAchievementFlag($achievement->id, AchievementFlag::Unofficial);
```
Jamiras: This function doesn't return whether or not anything changed. If you try to demote an already demoted achievement, it still generates a "{user} demoted this achievement to Unofficial" message and the response suggests the record was just demoted by the requested user.

```
{"data":{"type":"achievement-demotion","id":"525276","attributes":{"achievementId":525276,"status":"demoted",
"demotedAt":"2025-08-25T21:45:00+00:00","demotedBy":"Jamiras","wasTitleUpdated":false}}}
```

```
{"data":{"type":"achievement-demotion","id":"525276","attributes":{"achievementId":525276,"status":"demoted",
"demotedAt":"2025-08-25T21:46:49+00:00","demotedBy":"Jamiras","wasTitleUpdated":false}}}
```

app/Api/Internal/Requests/DemoteAchievementRequest.php
```diff
@@ -0,0 +1,53 @@
...
+class DemoteAchievementRequest extends BaseJsonApiRequest
+{
+    public function rules(): array
+    {
+        return [
+            'data' => 'required|array',
+            'data.type' => 'required|in:achievement-demotion',
+            'data.attributes' => 'required|array',
+            'data.attributes.achievementId' => 'required|integer|exists:Achievements,ID',
+            'data.attributes.username' => 'required|string|exists:UserAccounts,User',
```
Jamiras: This doesn't allow custom display_name matches. Should it?

Maybe it should use the ULID?

app/Api/Internal/Controllers/AchievementController.php
```diff
@@ -0,0 +1,102 @@
...
+     * ```
+     *
+     * @example Success response (200):
+     * ```json
+     * {
+     *   "data": {
+     *     "type": "achievement-demotion",
+     *     "id": "123",
+     *     "attributes": {
+     *       "achievementId": 123,
```
Jamiras: It feels redundant to have the achievement ID at both the top level and here. Suggest removing the top level value to be consistant with the input.

app/Api/Internal/Controllers/AchievementController.php
```diff
@@ -0,0 +1,102 @@
...
+     *       "title": "DEMOTED AS UNWELCOME CONCEPT - Original Title" // optional
+     *     }
+     *   }
+     * }
+     * ```
+     *
+     * @example Success response (200):
+     * ```json
+     * {
+     *   "data": {
```
Jamiras, line 45: Should there be an explicit `"success":"true"` at the top level?

app/Api/Internal/Controllers/AchievementController.php
```diff
@@ -0,0 +1,102 @@
...
+     *     "type": "achievement-demotion",
+     *     "attributes": {
+     *       "achievementId": 123,
+     *       "username": "DevCompliance",
+     *       "title": "DEMOTED AS UNWELCOME CONCEPT - Original Title" // optional
+     *     }
+     *   }
+     * }
+     * ```
```
Jamiras, line 40: If this is supposed to be [JSON-API compliant](https://jsonapi.org/format/#crud-updating), I think it needs to be:
```
"data": {
  "type": "achievement",
  "id": 123,
  "attributes": {
    "flags": 5,
    "actingUser": "DevCompliance",
    "title": "DEMOTED AS UNWELCOME CONCEPT - Original Title" // optional
  }
}
```
And the URL should be `http://localhost:64000/api/internal/achievements/123`.

I don't see anything about indirect mutations in the JSON-API document, but maybe we could treat it as a relationship? `http://localhost:64000/api/internal/achievements/123/relationships/demote`

I guess it just depends on how strictly we want to follow the JSON-API spec.

app/Api/Internal/Requests/BaseJsonApiRequest.php
```diff
@@ -0,0 +1,66 @@
...
+abstract class BaseJsonApiRequest extends FormRequest
+{
+    /**
+     * Handle a failed validation attempt for JSON:API requests.
+     * This matches the error structure used in Handler.php for JsonApiException.
+     */
+    protected function failedValidation(Validator $validator): void
+    {
+        throw JsonApiException::error([
+            'status' => '422',
```
Jamiras, line 20: Should we update the connect API to return 422 for invalid/missing parameter errors? It's currently returning 400. This seems more appropriate, and I don't think anything actually relies on the http status code - rcheevos doesn't even look at it unless the body is empty.

app/Api/Internal/Controllers/AchievementController.php
wescopeland: Updated in latest so both comments are recorded.

app/Api/Internal/Controllers/AchievementController.php
wescopeland: Acting user (associated to the API key) for the foreseeable future is going to be RABot. I don't think a permission check for the acting user is needed - everything in the service API is already guarded by the `ServiceAccountOnly` middleware which enforces a strict whitelist of which accounts are allowed to call these endpoints.

Agreed on restrictions for the `username` parameter. This is really is an action meant to be done by the DevCompliance account. In latest, I've added:
```php
if (!$user->hasRole(Role::TEAM_ACCOUNT)) {
    throw JsonApiException::error([
        'status' => 403,
        'title' => 'Forbidden',
        'code' => 'insufficient_role',
        'detail' => 'User does not have the required team account role.',
    ]);
}
```

app/Api/Internal/Controllers/AchievementController.php
wescopeland: In latest, the endpoint will omit "published" from "updatedFields" if the `Flags` value is not updated (see later comments about request/response payload structure changes).

If there's nothing to change, a 422 is returned.

app/Api/Internal/Requests/DemoteAchievementRequest.php
wescopeland: I think either `User` or `ulid` are better for this specific case, only because they're stable values and due to the internal nature of this API, we'll always have access to peek at those values. I've stuck with `User` (username) in latest, but happy to change it over to `ulid` if that's preferable.

app/Api/Internal/Controllers/AchievementController.php
wescopeland: Done in latest.

app/Api/Internal/Controllers/AchievementController.php
wescopeland: Per https://github.com/RetroAchievements/RAWeb/pull/3802#discussion_r2299240224, I don't think so. Ideally, a 200 response should be taken for granted that it was successful (many APIs don't do this, including some of our own probably).

app/Api/Internal/Controllers/AchievementController.php
wescopeland: You're right. This is a standard resource update that just happens to only support two fields. I think that means this should actually be a `PATCH`, and the first URL example you gave is the correct one. I haven't implemented a `PATCH` in quite some time.

After reading over that doc, the request payload needs to change too. Additionally, the returned `id` "MUST" be a string.

In latest:
* The route has been changed from `POST /achievements/demote` to `PATCH /achievements/{achievement}`.
* `AchievementController::demote` has been renamed to `AchievementController::update`, and it now accepts an `Achievement` model through route model binding.

The request format is now:
```json
{
	"data": {
		"type": "achievement",
		"id": "66",
		"attributes": {
			"published": false,
			"title": "This is a new title!"
		},
		"meta": {
			"actingUser": "DevCompliance"
		}
	}
}
```

The response format is now:
```json
{
	"data": {
		"type": "achievement",
		"id": "66",
		"attributes": {
			"title": "FooBarBaz",
			"description": "Invade the Imperial Spacecraft and destroy its core on any difficulty",
			"points": 5,
			"published": false,
			"gameId": 7
		},
		"meta": {
			"updatedAt": "2025-09-01T19:31:51+00:00",
			"updatedBy": "DevCompliance",
			"updatedFields": [
				"published",
				"title"
			]
		}
	}
}
```

Here's a sample curl:
```bash
curl --request PATCH \
  --url http://localhost:64000/api/internal/achievements/67 \
  --header 'Content-Type: application/json' \
  --header 'X-API-Key: redacted' \
  --data '{
	"data": {
		"type": "achievement",
		"id": "67",
		"attributes": {
			"published": false,
			"title": "This is a new title!"
		},
		"meta": {
			"actingUser": "DevCompliance"
		}
	}
}'
```

app/Api/Internal/Requests/BaseJsonApiRequest.php
wescopeland: I agree that a 422 seems more appropriate. I haven't studied what it would take to have the Connect API return 422 responses for validation errors. Is it a single place that'd need to be updated, or would each routine need to have a surgical modification?

app/Api/RouteServiceProvider.php
```diff
@@ -12,15 +13,31 @@
...
 use Illuminate\Support\Facades\Route;
+use LaravelJsonApi\Core\Exceptions\JsonApiException;
 
 class RouteServiceProvider extends ServiceProvider
 {
     use HandlesPublicFileRequests;
 
     public function boot(): void
     {
+        Route::bind('achievement', function ($value) {
```
Jamiras, line 27: I think this broke the achievements page:

http://localhost:64000/achievement/12345

> App\Providers\RouteServiceProvider::handlePageRequest(): Argument #2 ($resourceId) must be of type ?string, App\Models\Achievement given, called in /var/www/html/app/Providers/RouteServiceProvider.php on line 58

app/Api/Internal/Controllers/AchievementController.php
Jamiras: Now the type and id are redundant with the URL, but it seems like that's intended by the JSON-API spec. My only question is should "type" be "acheivements" (plural) to match the URL path as show in the JSON-API spec examples?

app/Api/Internal/Requests/BaseJsonApiRequest.php
Jamiras: There's a couple places in BaseApiAction that would need to be updated. I'll address it the next time I convert an action.

app/Api/Internal/Controllers/AchievementController.php
wescopeland: [ref](https://jsonapi.org/format/#document-resource-object-identification)
> Note: This spec is agnostic about inflection rules, so the value of type can be either plural or singular. However, the same value should be used consistently throughout an implementation.

Most examples in the spec are plural, so maybe it makes the most sense for us to use plural too. Updated in latest.

app/Api/RouteServiceProvider.php
wescopeland: Unexpected - I think I have a clear understanding now of how _pages-legacy_ route binding works. It'll be good to finally rip all that out. In the meantime, I've fixed this by making a tweak to how the _pages-legacy_ page gets its `achievement` binding so it won't conflict with the `Api` namespace. I've also verified the achievement tickets Folio page still works.

PR #3810, opened 2025-08-20
perf(api): optimize API_GetGameList

wescopeland:
Resolves https://retroachievementsorg.sentry.io/issues/45318267/?environment=production&query=is%3Aunresolved&referrer=issue-stream.

This PR significantly optimizes _API_GetGameList.php_. There is still a slight bottleneck for the server to actually build the gigantic payload, but the internal code is now fairly well optimized.

With 10 iterations for the internal code with `i` set to 21 and `h` set to 1:
* Base query: went from ~850ms to ~24ms.
* Hash lookup: went from ~15,000ms to ~92ms.

Data is now pre-aggregated in subqueries before joining it with the base query. Nested loops for hash lookups have been replaced with indexed lookups.


public/API/API_GetGameList.php
```diff
@@ -91,15 +116,19 @@
...
 if ($withHashes) {
+    $responseIndex = [];
+    foreach ($response as $index => $entry) {
+        $responseIndex[$entry['ID']] = $index;
+    }
+
     $hashes = GameHash::compatible()
         ->select('game_id', 'md5')
-        ->whereIn('game_id', array_column($response, 'ID'));
+        ->whereIn('game_id', array_column($response, 'ID'))
```
Jamiras: Could use `array_keys($responseIndex)` here. Not sure if that would provide any additional performance avoiding the extra lookup.

public/API/API_GetGameList.php
```diff
@@ -91,15 +116,19 @@
...
+        ->orderBy('game_id');
+
     foreach ($hashes->get() as $hash) {
-        foreach ($response as &$entry) {
-            if ($entry['ID'] == $hash->game_id) {
-                $entry['Hashes'][] = $hash->md5;
-                break;
-            }
+        if (isset($responseIndex[$hash->game_id])) {
```
Jamiras: This `isset` shouldn't be necessary as the query is limited to records in the array.

Jamiras (approved):

public/API/API_GetGameList.php
wescopeland: At the very least, I don't think it could hurt. Updated in latest.

public/API/API_GetGameList.php
wescopeland: You're right - removed that conditional in latest.

Jamiras (approved):

Issue #3812, opened 2025-08-20
New game page: don't count contributors more than once


wescopeland:
### Discussed in https://github.com/RetroAchievements/RAWeb/discussions/3811

<div type='discussions-op-text'>

<sup>Originally posted by **suspect15** August 20, 2025</sup>
On the game2 pages, contributors are counted for each category of contribution separate from being the set author even though they are the same person.  This leads to a seemingly inaccurate number of contributors listed since the author is also counted again as a contributor.

As I see the below example, I think 3 people worked on the set, but that is not true.

Example implying 3 total contributors, suspect15 +2 more:
![Screenshot_20250820_150222_Chrome](https://github.com/user-attachments/assets/55853652-5085-48bb-8533-966e1d85a405)

Upon closer inspection, there are only 2 contributors, suspect15 and lethauntic.
![Screenshot_20250820_150232_Chrome](https://github.com/user-attachments/assets/11018515-fe4a-4873-a60f-59ff43d83659)
</div>

PR #3674, opened 2025-06-27
feat(game): introduce server-side deduping of credits

wescopeland:
https://discord.com/channels/@me/1065706550318084117/1388158735951728831

This PR moves all set credit deduplication logic to the server.

**Before**
Screenshot 2025-06-27 at 5 37 17 PM

**After**
Screenshot 2025-06-27 at 5 46 57 PM

Additionally, makes a minor tweak to the 33% authorship threshold, lowering it to 30%. This fixes an issue where only one author label was appearing for Super Mario 64, even though another user had contributed something like 31% of the set's achievements.

PR #3815, opened 2025-08-20
fix(game): don't double-count authors as contributors

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3812.

http://localhost:64000/game2/3127

The set author is being double-counted as a contributor.

**Before**
Screenshot 2025-08-20 at 5 25 41 PM

Screenshot 2025-08-20 at 5 25 44 PM

---

**After**
Screenshot 2025-08-20 at 5 31 06 PM


Jamiras (commented):
Verified this fixes the count in smaller window sizes, but I'm reminded of #3674. Did we decide that's undesirable behavior?

Jamiras (approved):

wescopeland:
> I'm reminded of https://github.com/RetroAchievements/RAWeb/pull/3674. Did we decide that's undesirable behavior?

I'm going to wait until the beta is running and then do a sweep of how much sorting/filtering in Node.js we can lift up to PHP. Some if it may still need to live in the Node.js layer for one reason or another (filters from computed values, etc), but whatever we can lift up will likely be beneficial for hydration performance.

PR #3822, opened 2025-08-22
fix(PlaytimeStatistics): don't animate on mount

wescopeland:
This PR fixes a very minor bug in the game2 playtime statistics component. Currently, the component is firing its animation on mount (see video). With this PR's changes, it will only animate after user interaction.

This bug can only be observed when SSR is enabled.

https://github.com/user-attachments/assets/541ed00e-b863-4067-80c6-c815b1e28395


Jamiras (approved):
I'm not seeing it, probably because I have to scroll to the component, and refreshing "jumps" to the component, eliminating the ability to capture the initial animation.

PR #3824, opened 2025-08-23
feat(game): add multiset tab buttons

wescopeland:
This PR adds multiset tab buttons to React game pages. Set type for selectable achievement sets is no longer considered.

Given a game has multiple sets attached to it, they appear as selectable tab buttons. On hover, the destination is prefetched, and on click, a client-side navigation occurs to the `?set=X` route.

Screenshot 2025-08-23 at 5 06 23 PM

Screenshot 2025-08-23 at 5 06 43 PM

https://github.com/user-attachments/assets/c0949d8b-9ad6-4578-8c3e-4647f2334f4c


resources/js/features/games/components/GameAchievementSetsContainer/SetSelectionTabs/SetSelectionTabs.tsx
```diff
@@ -0,0 +1,84 @@
...
+                  )}
+                >
+                  <img
+                    src={gas.achievementSet.imageAssetPathUrl}
+                    alt={gas.title ?? t('Base Set')}
+                    className="size-8 select-none rounded-sm"
+                  />
+                </div>
+              </InertiaLink>
+            </BaseTooltipTrigger>
```
Jamiras, line 76: Accessing subset from http://localhost:64000/game2/3443 shows a blank page. Other games with empty subsets work, but do we want to show them?

```
select gas.game_id from game_achievement_sets gas join achievement_sets a
on a.id=gas.achievement_set_id where gas.type != 'core' and a.achievements_published = 0;
```

wescopeland: Ah, that blank page was actually a major bug. The page was crashing because that subset didn't have a forum topic. I've fixed that issue in latest.

(Loosely-held opinion) I do think we should probably show the empty subsets on the list. That way players can at least request the set, add it to their "want to play" list, or otherwise interact with it.

SalbugVR:
I noticed that when switching between multiset tabs, the `Want to Play` button's state can affect buttons in other sets. For example, a set that is not intended to be played may appear as `Want to Play`, and the reverse may also occur sometimes, but not consistently.

HhbSrcVNW6

SalbugVR:
> I don't understand.

For example, I have Super Mario Sunshine marked as `Want to Play`. When I switch to any of the other subsets, they also appear marked as `Want to Play`, even though none of them should actually be marked as such. The previous state of the `Want to Play` button is persisting across subsets even in situations when it should not.

VDpUJqZTyF

SalbugVR:
Hmm, I wasn't sure if I should've made this as a separate issue tbh since it's related to a WIP feature and it's sorta tied to this PR.

PR #3829, opened 2025-08-24
feat(home): add milestone points box

wescopeland:
https://discord.com/channels/476211979464343552/1002689485005406249/1409256182446555247


resources/js/features/home/components/+sidebar/GlobalStatistics/MilestonePointsBox.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+          <div className="!text-[20px] leading-7 text-neutral-300 light:text-neutral-950">
+            <span>
+              <span
+                className={cn(
+                  'inline-block text-amber-400 motion-safe:animate-pulse-glow',
+                  'light:text-amber-500',
+                )}
+              >
+                {firstChar}
```
Jamiras: I'm not very fond of making the first number a different color and animating it (I only even noticed the animation because I was staring at the oddly colored number trying to decide if it was intentional). 

I think the animated background is sufficient. 

Recommend either making the entire text gold or leaving it white.

wescopeland: Updated in latest.

PR #3849, opened 2025-08-30
fix(connect): correctly resolve root game id when two games are siblings that share the same subset

wescopeland:
https://retroachievements.org/ticket/95239

Screenshot 2025-08-30 at 7 12 23 PM

`ResolveRootGameIdFromGameIdAction` is not able to correctly handle when two games are siblings that share the same subset. When this occurs, the action believes one game is actually the parent of the other.

This is causing RetroArch to report 0 achievement unlocks for some users who are currently booting Pokemon Blue in production.


app/Connect/Actions/ResolveRootGameIdFromGameIdAction.php
```diff
@@ -16,14 +16,21 @@ class ResolveRootGameIdFromGameIdAction
...
-        // Get the game's achievement sets.
-        $sets = GameAchievementSet::whereGameId($gameId)->get();
-        if ($sets->isEmpty()) {
+        // Get ONLY the game's CORE achievement sets.
+        // We only check core sets to determine parent relationships.
+        // This prevents sibling games that share bonus sets from being
+        // incorrectly identified as parent/child relationships.
+        $coreSets = GameAchievementSet::whereGameId($gameId)
+            ->where('type', AchievementSetType::Core)
+            ->get();
```
Jamiras: This is clearly the same change made in #3848. Why can't this code just call `$game->parentGameId`? Or have that call into this action? It seems like this action could be eliminated entirely in favor of juse using the attribute extension.

I do notice that the two codes aren't quite the same. `$game->parentGameId` returns the first set using the game as a non-core subset, then falls back to a substring match of removing the `[Subset-` from the game title. This code returns any set using the game as a non-core subset (no orderby), and has no fallback. It seems like `$game->parentGameId` is more robust.

wescopeland: Good point.

In latest, I've removed `ResolveRootGameIdFromGameIdAction` altogether. Now, `ping` and `unlocks` are both using `$game->parentGameId ?? $game->id`.

PR #3852, opened 2025-08-31
feat(ticket): use Laravel mail, respect notification preferences

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3851.

This PR:
* Converts the two ticket emails to use Laravel mailables with Blade/HTML templates.
* Fixes an issue mentioned in #3851 where inactive devs are getting ticket notifications regarding achievements they authored sometimes in the distant past.

Screenshot 2025-08-31 at 11 09 52 AM

Screenshot 2025-08-31 at 11 12 28 AM

PR #3853, opened 2025-08-31
feat(mail): add unsubscribe links to all HTML mailables

wescopeland:
Stacked on https://github.com/RetroAchievements/RAWeb/pull/3852.

---

This PR adds various "unsubscribe" links to all Laravel/HTML mailables. Here are a few examples:

Screenshot 2025-08-31 at 7 41 39 PM

Screenshot 2025-08-31 at 7 42 07 PM

Screenshot 2025-08-31 at 7 42 13 PM

---

The unsubscribe links themselves are Laravel [temporary signed URLs](https://laravel.com/docs/11.x/urls#signed-urls), eg: 

http://localhost:64000/unsubscribe/eyJwcmVmZXJlbmNlIjozLCJ1c2VyX2lkIjo4MzI1MCwidHlwZSI6ImNhdGVnb3J5In0=?expires=1764455571&signature=aa9425e83b43ad78a7edd5505be29658575f8c228276117db0e37e6058bffcb2

These URLs/tokens are **not stored anywhere**. They're generated on-demand and validated cryptographically. Additionally, the links are **not authenticated**. Users who are unsubscribing may not even remember their credentials - requiring auth on these pages is a dark pattern.

Here's what an unsubscribe page looks like:
Screenshot 2025-08-31 at 7 46 31 PM

There is no confirm on this page (confirmation is also a dark pattern), but the page does give the user the option to undo the unsubscribe if they got here via a misclick.

From a technical POV, nearly all logic is self-contained within an `UnsubscribeService` (covered by exhaustive tests). This should hopefully not interfere with any sort of capability/service for listing a user's implicit or explicit subscriptions.


app/Mail/Services/UnsubscribeService.php
```diff
@@ -0,0 +1,283 @@
...
+
+        $user = User::find($data->userId);
+        if (!$user) {
+            return ['success' => false, 'errorCode' => 'user_not_found'];
+        }
+
+        if ($data instanceof GranularUnsubscribeData) {
+            // Remove the explicit unsubscribe record to restore the subscription.
+            // This will revert to implicit subscription if the user has commented,
+            // or no subscription if they haven't.
```
Jamiras: What if they had an explicit subscription before?
1) Subscribe to game wall X
2) Get email for game wall X
3) Unsubscribe from game wall X
4) Undo

The user is no longer subscribed at this point.

app/Mail/Services/UnsubscribeService.php
Jamiras, line 1: > This should hopefully not interfere with any sort of capability/service for listing a user's implicit or explicit subscriptions.

The main thing I would be concerned about would be duplication of logic to deal with implicit subscriptions. I don't see any of that here.

app/Mail/Services/UnsubscribeService.php
wescopeland: Fixed in latest. Also added another test case to verify.

PR #3854, opened 2025-09-01
feat(game): add player progress label to set header

wescopeland:
This PR is one of three PRs for showing player progress on React game pages:
1. A basic current progress label.
2. (upcoming) Beaten / Mastered award labels and progress bars.
3. (upcoming) The ability to reset progress on the achievement set.

The label UI has four possible states:

**State 1. Either no progress OR a completion/mastery with _no mixed progress_.**
The component doesn't render at all if the user:
* Has no progress whatsoever.
* Has a mastery and no softcore progress.
* Has a completion and no hardcore progress.

Supplemental treatments for those bottom two bullet points will be handled in a separate PR.

**State 2. Only partial hardcore progress.**
Screenshot 2025-09-01 at 1 14 33 PM

**State 3. Only partial softcore progress.**
Screenshot 2025-09-01 at 1 14 44 PM

**State 4. Mixed progress of any kind.**
Screenshot 2025-09-01 at 1 14 14 PM

Mixed progress looks kinda funky at first, but this treatment feels like an elegant way to handle what's really an edge case. In `player_games` where `achievements_unlocked` > 0:
* ~51.57% of rows are hardcore only progress.
* ~45.67% of rows are softcore only progress.
* ~2.76% of rows are mixed progress.

PR #3857, opened 2025-09-04
feat(game): add player progress indicators to set header

wescopeland:
This PR is one of three PRs for showing player progress on React game pages:

1. A basic current progress label. https://github.com/RetroAchievements/RAWeb/pull/3854
2. (this PR) Beaten / Mastered award labels and progress bars.
3. (upcoming) The ability to reset progress on the achievement set.

When a user is authenticated and navigates to an achievement set, they now see two indicators in the set heading:

Screenshot 2025-09-03 at 9 02 31 PM

These indicators are mostly intended to replace what currently appears in the PHP/Blade sidebar for progress. This design is specifically intended to address two directly competing concerns:
1. There is a now a hefty chunk of players who only care about beaten awards. The current PHP/Blade component always shows them as falling short, even though they're successfully playing to their preferred playstyle.
2. There will always be a hefty chunk of players who only care about masteries. They always want to be able to access a progress bar that goes from 0% to 100%.

These concerns are addressed by:
1. The "Beaten" indicator is the rightmost indicator. This gives a subtle nudge that when a game is beaten, something isn't abruptly still missing.
2. The "Beaten" indicator gets its own progress bar which takes cues from the beaten credit dialog.
3. The "Mastered" indicator always has a progress bar which goes from 0-100%.

On desktop and mobile, clicking/tapping the beaten indicator opens the beaten game credit dialog.
On mobile, tapping the mastered indicator still displays the tooltip (requires mobile user agent).

Additionally, the rare "Unbeaten" state is now more elegantly handled (by not needing to actually be handled at all).

If the user prefers softcore mode, the label for the Beaten indicator does not change. The tooltip content changes instead, and the "Mastered" indicator changes to "Completed".

Screenshot 2025-09-03 at 9 09 04 PM

Screenshot 2025-09-03 at 9 09 22 PM

Screenshot 2025-09-03 at 9 10 13 PM

Screenshot 2025-09-03 at 9 10 30 PM

Screenshot 2025-09-03 at 9 11 12 PM

Screenshot 2025-09-03 at 9 11 37 PM

Screenshot 2025-09-03 at 9 12 08 PM


resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/MasteredProgressIndicator/MasteredProgressIndicator.tsx
```diff
@@ -0,0 +1,169 @@
...
+    return null;
+  }
+
+  const unlockedHardcoreCount = achievements.filter((ach) => ach.unlockedHardcoreAt).length;
+  const unlockedSoftcoreCount = achievements.filter(
+    (ach) => ach.unlockedAt && !ach.unlockedHardcoreAt,
+  ).length;
+
+  const isMastered = playerGameProgressionAwards?.mastered;
+  const isCompleted = playerGameProgressionAwards?.completed;
```
Jamiras, line 50: Logically, it makes sense to have "Mastered" greyed out if the user hasn't mastered the set, but I can't help but feel it's disingenuous to say mastered if it's not mastered. Maybe we could change it to say "Not Mastered" or just show the current percentage.

Instead of:
Image

Maybe
Image

Or
Image

(Similar comment for Beaten)

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/GameAchievementSetProgress.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  return (
+    <div className="flex items-center gap-4">
+      <div className="flex h-full items-center">
+        <MasteredProgressIndicator achievements={achievements} />
+
+        {backingGame.id === game.id ? (
+          <BeatenProgressIndicator achievements={achievements} />
+        ) : null}
```
Jamiras, line 28: The Beaten indicator is hidden for subsets, but not the separator.

Image

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/MasteredProgressIndicator/MasteredProgressIndicator.tsx
```diff
@@ -0,0 +1,169 @@
...
+      <BaseTooltipTrigger>
+        <div
+          className={cn(
+            'flex items-center gap-0.5 border-r border-neutral-700 pr-4',
+            indicatorColorClassName,
+          )}
+        >
+          <LuAward className="size-5" />
+          <p className="font-medium">
+            {unlockedSoftcoreCount <= unlockedHardcoreCount ? t('Mastered') : t('Completed')}
```
Jamiras: This doesn't seem to be working correctly.

Image

It's not yellow, but it still says Mastered

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/MasteredProgressIndicator/MasteredProgressIndicator.tsx
wescopeland: I understand where you're coming from - the greyed out indicators might seem counterintuitive at first glance.

This current design is intentional, specifically aiming to address competing concerns from two distinct player groups:

1. Beaten-only players: a growing segment of players consider "Beaten" as their goal and don't pursue masteries.
2. Mastery-focused players: this is our traditional group that always shoots for 100%.

Showing "Not Mastered" or an always-visible percentage inadvertently tells beaten-only players they've fallen short, even when they've achieved their preferred playstyle goal. This is the exact pain point I'm trying to solve.

The greyed-out indicators serve as neutral, non-judgmental states. They simply show status without implying failure or incompletion. Combined with the active/hover state progress bars, both playstyles are equally validated.

From another point of view, the greyed-out state is very similar to a disabled button that has an action label like "Submit", but the user isn't able to submit yet due to some validation error. The action is possible, but it's not currently active. The visual distinction on these indicators (hopefully) make it immediately clear whether beaten/mastery has been achieved or not, while the labels themselves remain consistent to avoid negative framing for players who aren't pursuing it.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/GameAchievementSetProgress.tsx
wescopeland: Thanks, fixed in latest.
Screenshot 2025-09-04 at 7 43 02 PM

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/MasteredProgressIndicator/MasteredProgressIndicator.tsx
wescopeland: Updated this logic in latest.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetProgress/MasteredProgressIndicator/MasteredProgressIndicator.tsx
Jamiras: I understand the comparison to a disabled button. 

My qualm is that if I go to some page for a game that I've never played, it looks like I might have already beaten and mastered it:
image

And it will continue to look like that until I do beat it.

The only indicators that I'm even playing the set are the fact that some badges are not greyscale, and the unlock count in the summary:

Image

The old page had a progress bar right in the widget:
image

Jamiras: I could maybe see "not mastered" as subconciously shaming players who choose not to master everything, but I'm not convinced that reporting the unlock percentage would have any impact on players only going for the beat. It would provide feedback to everyone as they work on a set - and _everyone_ loves to see "number go up".

There's already a progress/mastery bar pretty much everywhere leading to the game page (in addition to the one in the old widget). Why shouldn't something visually similar be on the game page itself?

image

.
image

I'm not saying you should eliminate the tooltip. I kind of like them - the beaten one in particular is nice as I can look at it when I want without having to click open the dialog and scroll through the list to see how close I am. The new dialog also shows the summary progress, but previously required that I find a progression achievement to click on to open it.

And something else I rarely consider: How do these behave on mobile? Using the browser faker thing, it looks like the mastery tooltip shows up on press, but press on the beaten immediately shows the dialog.

Jamiras: The more I play around with this and summarize my thoughts, the more I'm driven towards just having something like this instead of the "Mastered | Beaten" widget:
image

It shows the progress, with a potential hardcore/softcore breakdown, and the beaten/mastery state. Plus, it's UI that the players are already familiar with.

Jamiras: Having a single dynamic field (like the current widget) would also work (and potentially address your concerns about beat-only players):

`Unstarted` -> `Started` -> `Beaten` -> `Mastered`.

The first state would have no tooltip. The second would show the Beaten tooltip. The third would show the Overall Set Progress tooltip, and the last would have no tooltip.

Personally, I'd still like to see the percentage outside of the tooltip (either raw or as a progress bar), but that may just be me.

wescopeland: > Unstarted -> Started -> Beaten -> Mastered.

I (originally) decided to go down this proposed path, and I implemented the whole UX locally.

I'm not happy with it because players who don't care for beaten awards aren't able to see a progress bar of their overall completion until they beat the game.

As a natural next step, I put both progress bars in a single tooltip, but then I was bothered by other issues such as the indicator having different clickable states based on the current progression status. In the end, I think this flow has subtle but significant UX issues.

> I could maybe see "not mastered" as subconciously shaming players who choose not to master everything, but I'm not convinced that reporting the unlock percentage would have any impact on players only going for the beat. It would provide feedback to everyone as they work on a set - and everyone loves to see "number go up".

I agree with your sentiment that everyone loves to see number go up. I think replacing the Mastered/Completed indicator label with the % is the next-best compromise, and I've warmed up to this idea.

Screenshot 2025-09-05 at 6 05 11 PM

---

> How do these behave on mobile? Using the browser faker thing, it looks like the mastery tooltip shows up on press, but press on the beaten immediately shows the dialog.

Yep, that's what I was going for, and it's purely based on vibes. I don't have a terribly strong opinion on this - on mobile these are quite hidden interactions to begin with.

---

I think the last remaining contention is the label for "Beaten" when the game itself isn't actually beaten. I'm pretty happy with it as-is. How strongly do you feel it should change?

Jamiras: > I think the last remaining contention is the label for "Beaten" when the game itself isn't actually beaten. I'm pretty happy with it as-is. How strongly do you feel it should change?

I'm fine with leaving the wording as-is, but can we make the icon change so it's more obvious when something is or isn't beaten? The color shift from dark grey to light grey doesn't feel strong enough. Whereas the color shift from dark grey to gold is very prominent.

image

.
image
.
image

How would you feel about using the Progression icon for unbeaten and the Win Condition icon for beaten?

Jamiras: Light mode messes with my brain. The disabled state is closer to the background color than the enabled state, but it feels like the darker color should be the disabled state. 
image

.
image
.
image

wescopeland: Not opposed to an icon change. Two things I don't like about using the Progression/Win icons:
1. The win icon is a ribbon, so it might visually compete with the ribbon icon for mastery.
2. The current icons are visually consistent with the icons used in the Playtime Stats component.

I went with a different icon in latest which might work equally well:
Screenshot 2025-09-05 at 6 53 09 PM

Additionally, in latest I've made updates to light mode's contrast.

PR #3859, opened 2025-09-04
move subscription deduction logic into service

Jamiras:
Replaces the handful of helper functions with a streamlined interface:
```
SubscriptionService
  getSubscribers(SubscriptionSubjectType, int): Collection<User>
  isSubscribed(User, SubscriptionSubjectType, int): bool
  updateSubscription(User, SubscriptionSubjectType, int, bool): Subscription
```

Handles both explicit and implicit subscriptions. Everything is constructed using fluent queries.

`SubscriptionSubjectType` is now an actual enum.

Future work (for subscription management UI):
```
  getSubscriptionCount(User, array $subjectTypes): int
  getSubscriptions(User, array $subjectTypes, int offset, int count): Collection<Subscription>
```


app/Community/Services/SubscriptionService.php
```diff
@@ -0,0 +1,260 @@
...
+                    } else {
+                        $includeMaintainer = !$ignoreUserIds || !in_array($maintainer->id, $ignoreUserIds);
+                    }
+                }
+
+                if ($includeMaintainer) {
+                    /** @var Builder<Model> $query4 */
+                    $query4 = Ticket::query()
+                        ->select([
+                            DB::raw(strval($maintainer->id) . ' as user_id'),
```
wescopeland: nit: Recommend `intval()` instead of `strval()`. `intval()` always forces an integer, which may be slightly safer within a `DB::raw()`.

app/Community/Services/SubscriptionService.php
```diff
@@ -0,0 +1,260 @@
...
+
+        $implicitSubscriptionsQuery = $this->getImplicitSubscriptionsQuery($subjectType, $subjectId, ignoreUserIds: $explicitSubcriberIds);
+
+        // discard explicitly unsubscribed
+        $subscriberIds = $subscribers->filter(fn ($s) => $s->state === true)->pluck('user_id');
+
+        // merge in implicit subscriptions
+        foreach ($implicitSubscriptionsQuery->get() as $implicitSubscription) {
+            $subscriberIds[] = $implicitSubscription->user_id;
+        }
```
wescopeland: nit: likely slightly more efficient:
```php
// discard explicitly unsubscribed
$subscriberIds = $subscribers->filter(fn ($s) => $s->state === true)->pluck('user_id')->toArray(); // !! toArray

// merge in implicit subscriptions
$implicitIds = $implicitSubscriptionsQuery->pluck('user_id')->toArray();
$subscriberIds = array_merge($subscriberIds, $implicitIds);
```

app/Community/Services/SubscriptionService.php
Jamiras: The `strval` was added because I didn't think it was possible to directly concatenate an integer and a string. It seems to work if I leave it off, so I'll go that route unless you have strong opinions.

app/Community/Services/SubscriptionService.php
Jamiras: I frequently forget that `pluck()` still returns a collection. I had intended `$subscriberIds` be an array.

I'm not certain that `array_merge` is more efficient than multiple appends, as I would assume`pluck` is doing the same multiple appends internally. But I'll take your word for it.

PR #3866, opened 2025-09-06
fix(manage): set deleted_at when deleting a code note

wescopeland:
On the PHP code notes page, when deleting a note, the note body is set to an empty string and the `deleted_at` field is set to now.

In the management app, ie http://localhost:64000/manage/games/1?activeRelationManager=4, when deleting a note, only the note body is set to an empty string. Currently, the `deleted_at` field is **not** set, resulting in inconsistent behavior.

This PR also has the management app update `deleted_at`.


app/Filament/Resources/GameResource/RelationManagers/MemoryNotesRelationManager.php
```diff
@@ -149,6 +149,7 @@ public function table(Table $table): Table
 
                             $memoryNote->user_id = $user->id;
                             $memoryNote->forceFill(['body' => '']);
+                            $memoryNote->deleted_at = now();
```
Jamiras: The non-manage page doesn't actually blank out the note when setting the `deleted_at` field. We probably shouldn't here either.

https://github.com/RetroAchievements/RAWeb/blob/6c1922f05be87899082aa2c10984edb273d0be81/app/Connect/Actions/SubmitCodeNoteAction.php#L54-L62

Jamiras: I'd suggest using the `SubmitCodeNoteAction` class, but it would take the note properties as input and have to look up the note again just to do the permissions check and delete call.

wescopeland: Updated this in latest to use `DeleteAction` instead of a generic `Action` and removed the action method.

`DeleteAction` automagically knows to soft delete the record.

PR #3871, opened 2025-09-06
feat(multiset): exclude subset game IDs from average completion percentage

wescopeland:
This PR makes a slight adjustment to the user profile's average completion percentage calculation. Subset game IDs are now excluded from the calculation.

As we know, this stat in particular steers user behavior. The motivation behind this change is we don't want players to disable multiset as a means to further manipulate this stat on their profile. They can safely leave multiset globally enabled and not feel "guilt" for engaging with additional content.

I believe this subtle change will be beneficial both for players (engaging with more stuff) and devs (more motivation to create interesting subsets).

**Before**
Screenshot 2025-09-06 at 7 30 00 PM

**After**
Screenshot 2025-09-06 at 7 29 50 PM


resources/views/pages-legacy/userInfo.blade.php
```diff
@@ -54,6 +56,23 @@
...
+        ->whereExists(function($query) {
+            $query->select('*')
+                ->from('game_achievement_sets as gas2')
+                ->whereColumn('gas2.achievement_set_id', 'game_achievement_sets.achievement_set_id')
+                ->whereColumn('gas2.game_id', '!=', 'game_achievement_sets.game_id')
+                ->where('gas2.type', '!=', AchievementSetType::Core);
+        })
+        ->pluck('game_id')
+        ->toArray();
+}
```
Jamiras, line 74: Is this really necessary? Can't we just exclude any game_id that is used as a non-core set?
```
    $subsetGameIds = GameAchievementSet::whereIn('game_id', $gameIds)
        ->where('type', '!=', AchievementSetType::Core)
        ->pluck('game_id')
        ->toArray();
```

Jamiras (approved):

resources/views/pages-legacy/userInfo.blade.php
wescopeland: Good point - this often trips me up. Updated in latest.

Jamiras: Nope, I was wrong. The new query finds any games that _have_ subsets. Not games that _are_ subsets.

wescopeland: No worries. Reverted back to the original complex query in latest.

PR #3873, opened 2025-09-07
feat(game): add 'Reset all progress' button to React game pages

wescopeland:
This PR adds a "Reset all progress" button to React game pages.

Similar to the Blade/PHP game page, there are multiple user interactions required to trigger a reset, and we don't really broadcast where the button is on the page.

Once the user confirms the reset, the page makes a client-side reload of page props.

https://github.com/user-attachments/assets/90b4f83f-95a7-4467-af7f-6e2caf3e72ba


resources/js/features/games/components/ResetAllProgressAlertDialog/ResetAllProgressAlertDialog.tsx
Jamiras, line 1: Resetting a beaten/mastered game doesn't change the color of the master/beaten indicators.

Image

.
Image

wescopeland: Fixed in latest. This is because there's a delay on how long it takes to update `playerGame` - the page refreshes faster than the back-end can process the update. We have to figure out the current value some other way (ie: achievement unlock records).

PR #3874, opened 2025-09-07
Add functions to query player subscriptions

Jamiras:


app/Community/Services/SubscriptionService.php
```diff
@@ -57,58 +62,205 @@ public function isSubscribed(User $user, SubscriptionSubjectType $subjectType, i
...
+            SubscriptionSubjectType::GameAchievements => new GameAchievementsSubscriptionHandler(),
+            SubscriptionSubjectType::GameTickets => new GameTicketsSubscriptionHandler(),
+            SubscriptionSubjectType::GameWall => new GameWallSubscriptionHandler(),
+            SubscriptionSubjectType::Leaderboard => new LeaderboardWallSubscriptionHandler(),
+            SubscriptionSubjectType::UserWall => new UserWallSubscriptionHandler(),
         };
     }
+}
 
+abstract class BaseSubscriptionHandler
```
Jamiras, line 215: I wanted these to be nested classes, but PHP doesn't support that. If they should be moved to a subfolder or something else, let me know.

They just group similar functions together by `SubscriptionSubjectType` instead of having multiple `match` statements.

app/Community/Services/SubscriptionService.php
```diff
@@ -57,58 +62,205 @@ public function isSubscribed(User $user, SubscriptionSubjectType $subjectType, i
...
+        }
+
+        return $results;
+    }
+
+    private function getSubscriptionSubjectIds(User $user, SubscriptionSubjectType $subjectType): array
+    {
+        $ttl = 12 * 60 * 60; // 12 hours
+
+        return Cache::remember("users.{$user->id}.subscriptions.{$subjectType->value}", $ttl, function () use ($user, $subjectType) {
```
Jamiras, line 170: I'm assuming the UI will need to query the total number of results in addition to multiple slices of results as the user goes from page to page. This is an attempt to only do the implicit subscription calculations once for all the little interactions. It might be premature optimization, and could potentially be removed once the UI is in place.

app/Community/Services/SubscriptionService.php
```diff
@@ -57,58 +62,205 @@ public function isSubscribed(User $user, SubscriptionSubjectType $subjectType, i
...
+            $subscriptions = Subscription::query()
+                ->where('user_id', $user->id)
+                ->where('subject_type', $subjectType)
+                ->get();
+
+            $explicitSubscriptionIds = $subscriptions->filter(fn ($s) => $s->state === true)->pluck('subject_id')->toArray();
+
+            $implicitSubscriptionIds = $handler->getImplicitSubscriptionQuery(null, $user->id, null)
+                ->when(!empty($subscriptions), function ($query) use ($subscriptions) {
+                    $query->whereNotIn('subject_id', $subscriptions->pluck('subject_id'));
```
wescopeland: Trying to demo this out locally to get a good feel for it.

Possible bug here?
```php
$gameWallSubs = $service->getSubscriptions(
  $user,
  [SubscriptionSubjectType::GameWall],
  0,
  10
);
```
> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'subject_id' in 'WHERE' (Connection: mysql, SQL: select distinct `user_id`, ArticleId as subject_id from `Comment` where `ArticleType` = 1 and `user_id` != 14188 and `user_id` = 117089 and `subject_id` not in (3110, 5527, 11270, 12139, 26257, 26258, 20694, 32909, 28237, 31952, 32565, 668, 14403) and `Comment`.`deleted_at` is null)

Jamiras: It's an aliased column: `ArticleId as subject_id`. And while that does seem to work in the unit tests, I do get the same error trying to use it naturally. Apparently, it's not standard SQL:

> Standard SQL doesn't allow you to refer to a column alias in a WHERE clause. This restriction is imposed because when the WHERE code is executed, the column value may not yet be determined.

There's two potential solutions:
1) Change it to a `having` (which requires raw construction because eloquent doesn't provide a havingNotIn):
```
$query->havingRaw('subject_id NOT IN (' . implode(',', $subscriptions->pluck('subject_id')->toArray()) . ')');                    
```
2) Add a helper to the handler to get the source column:
```
$query->whereNotIn($handler->getSubjectColumnName(), $subscriptions->pluck('subject_id'));
```

Both seem to take about the same amount of time to execute, so I'm leaning more towards the first as it's more lenient in case the `subject_id` isn't just an alias.

Jamiras: Sadly, it looks like sqllite doesn't like the `having` implementation:
```
SQLSTATE[HY000]: General error: 1 HAVING clause on a non-aggregate query (...
```

wescopeland: Yea, I suspect adding `getSubjectColumnName()` to the abstract class is likely going to be the cleanest solution.

Jamiras: There are a couple cases where the filter has to be applied to subqueries, so I've added another parameter to pass to each implementation and the filter is applied internally.

app/Community/Services/SubscriptionService.php
wescopeland, line 1: How hard do you think it would be for `getSubscriptions()` to return something like a `subscribed_at` field for each entity? I only had this thought after actually playing with the response. We could use this for sorting, or we could use it to let users know which of their subscriptions are ancient / recent.

wescopeland (approved):

app/Community/Services/SubscriptionService.php
Jamiras: I'm not sure we know that information.

For explicit subscriptions, we could assume the `updated_at` value is when they subscribed (as `state` is the only mutable field and we can't rely on `created_at` because the user may have toggled their subscription since then).

For implicit subscriptions, we'd have to find the earliest comment from the user, which makes the queries much more intensive - especially if we intend to sort and paginate on that value.

PR #3875, opened 2025-09-07
add PostActivityAction

Jamiras:


app/Connect/Support/BaseApiAction.php
```diff
@@ -52,7 +52,7 @@ protected function invalidParameter(string $message): array
     {
         return [
             'Success' => false,
-            'Status' => 400,
+            'Status' => 422,
```
Jamiras, line 55: Addresses https://github.com/RetroAchievements/RAWeb/pull/3802#discussion_r2316369423

app/Connect/Concerns/HeartbeatRequests.php
```diff
@@ -13,42 +12,6 @@
...
-    /**
-     * Used by RAIntegration
-     * Called on
-     * - game load -> StartedPlaying
-     *
-     * @throws Exception
-     *
-     * @since 1.0
-     */
-    protected function postactivityMethod(Request $request): array
```
Jamiras, line 25: It looks like luchaos may have already started some sort of Connect replacement. This is used by ConnectApiController, but I'm not sure how that's used. As best as I can tell, ConnectApiController is looking for some sort of new domain to access the APIs.

https://github.com/RetroAchievements/RAWeb/blob/644058faef1193cf6aca14c05f11bfc82f24565d/app/Connect/RouteServiceProvider.php#L41

I'm not sure how this plays into the new API stuff you're adding (i.e. #3802). Maybe we can remove it entirely?

app/Connect/Actions/PostActivityAction.php
```diff
@@ -0,0 +1,62 @@
...
+use App\Connect\Support\BaseAuthenticatedApiAction;
+use App\Models\Game;
+use App\Platform\Events\PlayerSessionHeartbeat;
+use App\Platform\Services\VirtualGameIdService;
+use Illuminate\Http\Request;
+
+// NOTE: This function originally allowed posting activity directly to the global feed. That
+//       was removed in 2019. This function is still used to start sessions on older clients.
+// DEPRECATED: clients should be using startsession instead [added 2023 Jul 11], which serves
+//             as a combined request for postactivity and unlocks.
```
wescopeland: Recommend converting this to PHPDoc and using `@deprecated`:
```php
/**
 * NOTE: This function ...
 *
 * @deprecated Clients should be using ...
 */
```
The benefit of doing this is IDEs loudly point out when code is marked as deprecated, even when not peeking into the implementation details of the code:
Screenshot 2025-09-07 at 5 29 38 PM

On hover, the deprecation details can be read.

app/Connect/Concerns/HeartbeatRequests.php
wescopeland: I believe it's leftover V2 code, and we've gone in a different direction. We should plan to remove it entirely.

app/Connect/Actions/PostActivityAction.php
wescopeland, line 1: Intellectual curiosity: do we have any kind of usage metrics / visibility into how often this gets called? I wasn't even aware this routine was a thing.

I guess what I'm wondering is when will we reach a point where this can be removed entirely.

Jamiras: Not directly. The cutover in rcheevos was 11.0, so any RetroArch prior to 1.17.0 would still be using it. And anyone using DLLs older than 1.3 would still be using it. We'd have to query user agents to derive the usage metric.

rcheevos 11.0 is also the version that introduced `rc_client`, so most integrations should be using it. There's only a handful of clients that integrated directly with rcheevos before `rc_client` was available. Most older integrations interfaced directly with the DLL.

app/Connect/Actions/PostActivityAction.php
Jamiras: It's deprecated from the client perspective, not really from the webserver code perspective. Marking it with the doc comment such that the code editor flags it as deprecated would imply to the user viewing the dorequest file that the functionality needs to be updated. The Action code itself isn't deprecated. The API that it's supporting is. 

The webserver will need to continue to support it until the deprecated clients are no longer supported (which basically means forever as we effectively allow very old clients to run in softcore). The deprecation comment is more of a documentation detail than a true deprecation.

If you feel this is still an appropriate change to make, I will. Or if the comment needs to be more explicit about what is actually deprecated, I can attempt to clarify that too.

wescopeland: I see - based on this context I don't believe using the `@deprecated` would be appropriate as it probably sends the wrong signal.

Based on https://github.com/RetroAchievements/RAWeb/pull/3875#discussion_r2330489047, how would you feel about a comment like this? (I'm taking creative liberties):

```php
/**
 * Legacy function for client session initialization.
 * New clients should use ?r=startsession instead (available since rcheevos 11.0) which
 * serves as a combined request for postactivity and unlocks.
 * 
 * This originally supported posting activity directly to the global feed (removed in 2019).
 * Now it only processes StartedPlaying (activity type 3) so as to maintain sessions for older clients.
 * 
 * This endpoint must be maintained indefinitely for backwards compatibility with:
 * - RetroArch versions prior to 1.17.0.
 * - DLL integrations older than 1.3.
 * - Other legacy clients that haven't migrated to rc_client.
 */
```

Jamiras: I made a few minor changes, but mostly used what you provided.

PR #3879, opened 2025-09-08
feat(connect): add ?r=submitrichpresence

wescopeland:
```
curl -X POST "localhost:64000/dorequest.php?r=submitrichpresence" \
    --header "User-Agent: RALibretro/1.8.1 (WindowsNT 10.0)" \
    --header "Content-Type: application/x-www-form-urlencoded" \
    -d "u=SomeDeveloper&t=redacted&g=1&d=Display%3A%0D%0ADoing+some+testing."
```
This is a POST request due to concerns about query param length.

### DIFF
--- a/tests/Feature/Connect/SubmitRichPresenceTest.php
+++ b/tests/Feature/Connect/SubmitRichPresenceTest.php
@@ -323,10 +323,10 @@ class SubmitRichPresenceTest extends TestCase
             $this->apiUrl('submitrichpresence'),
             ['g' => $game->id]
         )
-            ->assertStatus(400)
+            ->assertStatus(422)
             ->assertExactJson([
                 'Success' => false,
-                'Status' => 400,
+                'Status' => 422,
                 'Code' => 'missing_parameter',
                 'Error' => 'One or more required parameters is missing.',
             ]);
@@ -370,10 +370,10 @@ class SubmitRichPresenceTest extends TestCase
             $this->apiUrl('submitrichpresence'),
             ['d' => 'Display:\nMissing Game ID']
         )
-            ->assertStatus(400)
+            ->assertStatus(422)
             ->assertExactJson([
                 'Success' => false,
-                'Status' => 400,
+                'Status' => 422,
                 'Code' => 'missing_parameter',
                 'Error' => 'One or more required parameters is missing.',
             ]);


app/Connect/Actions/SubmitRichPresenceAction.php
```diff
@@ -0,0 +1,70 @@
...
+        $this->richPresence = request()->post('d', '');
+
+        return null;
+    }
+
+    protected function process(): array
+    {
+        if (VirtualGameIdService::isVirtualGameId($this->gameId)) {
+            [$this->gameId, $compatibility] = VirtualGameIdService::decodeVirtualGameId($this->gameId);
+        }
```
Jamiras: This shouldn't be necessary. Once the client has received the patch data, it should use the correct game ID from there on.

It doesn't hurt to leave it here. `isVirtualGameId` will just always return false. And if we didn't do it, and someone did send a virtual game ID, it would error on `gameNotFound`.

app/Connect/Actions/SubmitRichPresenceAction.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $game = Game::find($this->gameId);
+        if (!$game) {
+            return $this->gameNotFound();
+        }
+
+        // Check if user has permission to update this game.
+        if (!$this->user->can('update', $game)) {
+            return $this->accessDenied();
+        }
```
Jamiras, line 51: Should this be checking for "can update field"? 

Related: https://github.com/RetroAchievements/RAWeb/pull/3891#discussion_r2341088914

app/Connect/Actions/SubmitRichPresenceAction.php
```diff
@@ -0,0 +1,70 @@
...
+
+        // modifyGameRichPresence handles versioning automatically.
+        $success = modifyGameRichPresence($this->user, $this->gameId, $this->richPresence);
+
+        if (!$success) {
+            return [
+                'Success' => false,
+                'Error' => 'Failed to update rich presence.',
+            ];
```
Jamiras, line 56: An error response should have four fields. You're missing "Status" and "Code" (see BaseApiAction for examples). "Status" will bubble out as the HTTP response code, and "Code" provides an untranslated string in case we want to reword (or translate) the actual error message.

Also (at least currently), the only way `modifyGameRichPresence` can fail is if a bad game ID is provided, which we already check on line 46.

You might even consider eliminating `modifyGameRichPresence` in favor of using this action. The only other thing that calls it is the handler for editing it on the game page (and your new use on the manage page). That's why the action pattern has an `execute` method - for internal use.

app/Connect/Actions/SubmitRichPresenceAction.php
wescopeland: If it's unnecessary, probably just best to remove it. Removed in latest.

app/Connect/Actions/SubmitRichPresenceAction.php
wescopeland: Yep, updated in latest (and also set #3891 temporarily to draft).

app/Connect/Actions/SubmitRichPresenceAction.php
wescopeland: Decided to remove `modifyGameRichPresence` in favor of using this action. Also removed the error response.

PR #3888, opened 2025-09-09
fix(multiset): improve subset game select field UX in attach wizard

wescopeland:
http://localhost:64000/manage/games/2606?activeRelationManager=1

When clicking "Attach Subset", the Subset game field is currently limited to N results, even though it's fetching everything. This makes it almost appear as if games are missing from the list unless the user explicitly types for them.

Now, all results are shown.
Screenshot 2025-09-09 at 7 36 24 PM

### DIFF
--- a/app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
+++ b/app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
@@ -151,7 +151,6 @@ class AchievementSetsRelationManager extends RelationManager
                                 Select::make('game')
                                     ->label('Subset game')
                                     ->helperText('Find the current subset game in the database. For example: "Mega Man 2 [Subset - Bonus]".')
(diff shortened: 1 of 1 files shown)


app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
Jamiras, line 1: I think I have more results on `master` than this branch (note the scrollbar):

master:
Image

this branch:
Image

Also note that the ability to filter the results is now missing. Although if you're looking for a specific game, it should be trivial to scroll to it.

I wonder how scalable this solution is. Shouldn't it be a pure search field, like we use for attaching hubs?

wescopeland: I can't explain there being more results - the query should be identical. I'm going to close this for now and think of a better way to attack this problem.

PR #3891, opened 2025-09-10
feat(manage): add support for viewing and editing game RP scripts

wescopeland:
http://localhost:64000/manage/games/1/edit

Tabs are reserved for relation managers and subresources. This is a simple textarea on the resource in both the infolist and form.

Screenshot 2025-09-09 at 9 21 05 PM

Edits do not appear in the audit log. They should eventually be exposed in some kind of versioning table - I haven't totally thought through that yet.


app/Filament/Resources/GameResource.php
```diff
@@ -304,6 +314,19 @@ public static function form(Form $form): Form
...
+                    ->icon('heroicon-s-chat-bubble-left-right')
+                    ->schema([
+                        Forms\Components\Textarea::make('RichPresencePatch')
+                            ->label('Rich Presence Script')
+                            ->maxLength(60000)
+                            ->rows(10)
+                            ->helperText(new HtmlString('<a href="https://docs.retroachievements.org/developer-docs/rich-presence.html" target="_blank" class="underline">Learn more about Rich Presence</a>'))
+                            ->placeholder("Format:Number\nFormatType=VALUE")
+                            ->extraInputAttributes(['class' => 'font-mono'])
+                            ->disabled(!$user->can('updateField', [$form->model, 'RichPresencePatch'])),
```
Jamiras, line 328: Junior developers are currently allowed to update Rich Presence for games where they're the sole developer and have a claim:
https://github.com/RetroAchievements/RAWeb/blob/8fccfc170deef900e1ce854849be0399f72dd208/resources/views/pages-legacy/gameInfo.blade.php#L565-L571

`RichPresencePatch` needs to be added to the `$roleFieldPermissions` for `DEVELOPER_JUNIOR`, but I think that would allow them to edit it even if they weren't the sole developer. I'm not sure how important that clause is.

wescopeland: The changes from #3879 are merged into this branch in latest, so this shouldn't be possible (and is covered by a test: `SubmitRichPresenceTest::testJuniorDeveloperWithMixedAuthorshipCannotSubmitRichPresence`).

wescopeland:
@Jamiras Given https://github.com/RetroAchievements/RAIntegration/pull/1229, do you think it'd make sense to either set this to being a read-only field (or even not merge this at all)?

drisc:
Being able to edit an RP script without booting a game is a good QoL feature in my opinion, it's not something that requires rigorous validation like an achievements or leaderboard and it allows quick fixes without the need of a computer running an emulator.

wescopeland:
Sounds good to me. Going to merge as-is.

PR #3897, opened 2025-09-10
feat(game): enable React game pages beta

wescopeland:
I think we're probably at a point now where it's reasonable to begin a public beta of the new game pages. This PR facilitates that.

**How it works:**
This PR includes a migration. The migration adds a new boolean field to `UserAccounts`: `enable_beta_features`. Additionally, `Role::BETA` has been removed. This SQL should be executed in prod:
```sql
DELETE FROM `retroachievements`.`auth_roles` WHERE `id` = 19;
```

On the user settings page, there is a new preferences toggle:
Screenshot 2025-09-10 at 7 25 19 PM

If this toggle is enabled, `game.show` will now always route to the Inertia page instead of the legacy gameInfo PHP page.

We can have various callouts and whatnot to this toggle later - I just wanted a dedicated PR to isolate the functionality. Another subsequent PR will also add some kind of feedback component so we can gather reception to the new page.


app/Community/Controllers/UserSettingsController.php
```diff
@@ -220,4 +220,18 @@ public function resetConnectApiKey(ResetConnectApiKeyRequest $request): JsonResp
 
         return response()->json(['success' => true]);
     }
+
+    public function toggleBetaFeatures(Request $request): JsonResponse
+    {
+        /** @var User $user */
+        $user = $request->user();
+
+        $user->enable_beta_features = !$user->enable_beta_features;
```
Jamiras: Curious why you chose to create a separate field for this instead of adding it to `websitePrefs` (like SuppressionMatureContent and OptOutOfSubsets).

app/Providers/RouteServiceProvider.php
```diff
@@ -56,10 +61,35 @@ protected function mapWebRoutes(): void
...
+                    );
+                }
+
+                // For non-beta users, we need to bypass Inertia and serve the legacy page.
+                request()->merge(['game' => $game]);
+
+                // Return a non-Inertia response that will bypass Inertia middleware.
+                return response()->view('pages-legacy.gameInfo');
+            })->name('game.show');
+        });
```
Jamiras, line 95: Recommend providing a `/game1/` alias to get back to the original page (for comparison) without having to disable the feature.

app/Providers/RouteServiceProvider.php
Jamiras, line 1: Similar Games widget links directly to `game2` pages:

Image

As does the subset redirect: http://localhost:64000/game/20026 and changing the selected subset from within the page.

app/Community/Controllers/UserSettingsController.php
wescopeland: I figured at some point we may want to have the ability to set this value in the DB on behalf of users as we go in between different betas (turning it off).

app/Providers/RouteServiceProvider.php
wescopeland: This is fixable, but I'm cautious about fixing it. I think it may be safer to leave until the real cutover happens.

If the user takes these steps:
1. Have beta features disabled.
2. Navigate to game2 directly.
3. Click one of these links.

It results in Inertia throwing a client-side error and popping the correct page in a modal. This is because `<InertiaLink />` can't work correctly if it's linking to a Blade/PHP page. Replacing `<InertiaLink />` with `<a />` is undesirable.

app/Providers/RouteServiceProvider.php
wescopeland: Added in latest.

app/Community/Controllers/UserSettingsController.php
Jamiras: Bitflags do not prevent doing that.

This will clear the OptOutOfAllSubsets flag for all users:
```
UPDATE UserAccounts SET websitePrefs = websitePrefs & ~262144;
```

app/Providers/RouteServiceProvider.php
Jamiras: That makes sense for changing the selected tab, but the redirect doesn't happen if not opted in to the feature. And it's a wash as to whether or not it's important that the Similar Games links stay within the new ecosystem.

app/Providers/RouteServiceProvider.php
```diff
@@ -56,8 +61,37 @@ protected function mapWebRoutes(): void
...
+            Route::get('game1/{game}{slug?}', fn ($game) => $this->handlePageRequest('gameInfo', $game))->name('game1.show');
+        });
+
+        // Intelligently serves different pages based whether the user has enabled beta features.
+        Route::middleware(['web', 'csp', 'inertia'])->group(function () {
+            Route::get('game/{game}{slug?}', function ($game) {
+                /** @var User $user */
+                $user = Auth::user();
+
+                if ($user->enable_beta_features) {
```
Jamiras: This throws an error if not logged in.

app/Providers/RouteServiceProvider.php
wescopeland: I'm having trouble following.

My position is I don't want to change these to anchor tags just to change them back to InertiaLink tags once we're done with the beta. They should be InertiaLink tags while we're running the beta - this causes the page the user is navigating to to render faster.

I can't make them InertiaLink tags if they're pointing to the PHP page - it's going to throw an error. The only way around this is to create a temporary component whose purpose is to conditionally render either an anchor tag or an InertiaLink that we're going to delete anyway.

Jamiras: > I'm having trouble following.

* Switching between sets makes sense to hardcode as there is no game1 equivalent for the subset tab.
* The game page for a subset redirecting to a subset tab is something that should only happen if the beta functionality is enabled, and therefore it should be reasonable to go directly to `/game/` instead of `/game2/`.
* Similar games links are relatively insignificant, but could link to the `/game/` page regardless of whether or not the beta functionality is enabled. If it is enabled, they'd load the game2 page. Otherwise, they'll load the game1 page.

I don't caim to understand the differences between `<InertiaLink>` and `<a>`. If it's a hassle to have it just point to `/game/` (regardless of whether that points to game1 or game2), it can wait.

I'm just seeing links explicitly pointing at game2 that I imagine should be able to point to the standard `/game/` route and let that determine whether or not to go to game1 or game2.

wescopeland: In latest, I've done my best to gracefully handle this so an error won't be thrown.

app/Providers/RouteServiceProvider.php
wescopeland: Thanks, fixed in latest.

app/Community/Controllers/UserSettingsController.php
wescopeland: Good to know - I've removed the migration in latest and moved this to using websitePrefs.

resources/js/features/games/components/GamePageLink/GamePageLink.tsx
Jamiras, line 1: I didn't realize this would require a new widget. Feel free to go back to using the direct `/game2/` links.

I had just assumed it would be a simple `route('game2.show'` => `route('game.show'` sort of change.

wescopeland: Unfortunately it isn't. Sorry, I probably didn't explain why well enough.

We want React to React page navigations to use `InertiaLink`. These navigations are client-side transitions. For these navigations, instead of remounting the entire DOM, the browser just makes an API call to grab the props of the page being navigated to. Once that API call returns a 200, the page seamlessly transitions. The layout, scripts, assets, etc aren't remounted - InertiaLink effectively transforms the site into a single-page app. Inertia gives us the best of both worlds: the first hit to the site is rendered on the server (SSR), and subsequent navigations are rendered on the client.

This effect can be observed most profoundly when paginating on the forum. Under best conditions, it almost feels like using desktop software.

You may notice a subtle loading bar that appears on the top of the page when a client-side navigation is running:

https://github.com/user-attachments/assets/3ddeb3fd-732b-457e-8cf9-fd774c9b092d




Anchor tags (`<a>`) are a full reload on navigation - the entire DOM is unmounted, then remounted for the next page.

`InertiaLink` throws an error if the link being navigated to isn't an Inertia route, because it's trying to make an API call for a client-side navigation to its destination, can't, and then explodes.

wescopeland: In latest, I've set these to use game2 links again.

app/Community/Controllers/UserSettingsController.php
Jamiras: How does this work? I see `getEnableBetaFeaturesAttribute` which pulls it out of the bitfield, but I don't see any helper to shove the new value back in.

Jamiras (approved):

app/Community/Controllers/UserSettingsController.php
wescopeland: Whoops, I forgot to delete all this. This is all dead code and not even used by the UI anymore. I've removed this method and its route definition in latest.

wescopeland:
I think this is probably good to merge after #3931 and #3933 are in.

PR #3900, opened 2025-09-13
feat(game): add beta feedback form

wescopeland:
This PR provides a way for users to submit feedback on the new React game pages. **A migration is included in this PR.**

A user may submit beta feedback for the new game page as long as:
* They are not muted.
* Their email is verified.
* They've been a member for at least 2 weeks.
* They have 3,000 points in either mode.
* They've been using the new game page for at least 2 days.
* They've visited the new game page at least 10 times.

We'll probably get some feedback that isn't very helpful, but these conditions should weed out a lot of noise.

If all the conditions are met, a new button appears in the header:
Screenshot 2025-09-13 at 2 49 06 PM

The new button is hard to miss, but also not overly-loud. On click, a form dialog appears:
Screenshot 2025-09-13 at 2 49 55 PM

The two textareas are optional, but we (deliberately) don't point this out. Feedback is saved to the database in a new table called `user_beta_feedback_submissions`:
Screenshot 2025-09-13 at 2 53 40 PM


Jamiras:
I'm not seeing the new button. How do you determine "They've been using the new game page for at least 2 days."?

wescopeland:
This is handled by two new functions in `BuildGameShowPagePropsAction`: `trackBetaVisit` and `getCanSubmitBetaFeedback`.

To force the button to appear, you can make this tweak:
```php
private function getCanSubmitBetaFeedback(?User $user, string $betaName): bool
{
    return true;
```

app/Community/Requests/StoreBetaFeedbackRequest.php
```diff
@@ -0,0 +1,21 @@
...
+
+class StoreBetaFeedbackRequest extends FormRequest
+{
+    public function rules(): array
+    {
+        return [
+            'betaName' => 'required|string|max:100',
+            'rating' => ['required', 'integer', Rule::in([1, 2, 3, 4, 5])],
+            'positiveFeedback' => 'sometimes|string',
+            'negativeFeedback' => 'sometimes|string',
```
Jamiras, line 18: This doesn't appear to be working right. If I only click a face and hit submit, the page generates a generic error:
Image

Digging into the response, I see:
```
{
    "message": "The positive feedback must be a string. (and 1 more error)",
    "errors": {
        "positiveFeedback": [
            "The positive feedback must be a string."
        ],
        "negativeFeedback": [
            "The negative feedback must be a string."
        ]
    }
}
```

database/migrations/2025_09_13_000000_create_user_beta_feedback_submissions_table.php
```diff
@@ -0,0 +1,36 @@
...
+        Schema::create('user_beta_feedback_submissions', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedBigInteger('user_id');
+            $table->string('beta_name', 100);
+            $table->tinyInteger('rating'); // 1-5
+            $table->text('positive_feedback');
+            $table->text('negative_feedback');
+
+            $table->string('page_url', 500)->nullable();
+            $table->text('user_agent')->nullable();
```
Jamiras, line 21: this is a `varchar(255)` in the `player_sessions` table.

database/migrations/2025_09_13_000000_create_user_beta_feedback_submissions_table.php
```diff
@@ -0,0 +1,36 @@
...
+            $table->unsignedInteger('visit_count')->nullable();
+            $table->timestamp('first_visited_at')->nullable();
+            $table->timestamp('last_visited_at')->nullable();
+
+            $table->timestamps();
```
Jamiras, line 26: This data is immutable, does it need an `updated_at` column?

app/Policies/UserBetaFeedbackSubmissionPolicy.php
```diff
@@ -0,0 +1,72 @@
...
+class UserBetaFeedbackSubmissionPolicy
+{
+    use HandlesAuthorization;
+
+    public function manage(User $user): bool
+    {
+        return $user->hasAnyRole([
+            Role::ROOT,
+            Role::ADMINISTRATOR,
+        ]);
```
Jamiras, line 21: Where are these records exposed? I'm guessing a future PR. Or are we just going to use metabase?

How do you plan to aggregate the data? It's easy enough to chart the rating distribution, but I suspect that the positive/negative feedback will have to collated by hand.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -673,4 +680,58 @@ private function getLeaderboardsCount(Game $game): int
...
+        // Get existing data, or initialize with new data.
+        $data = Cache::get($cacheKey, [
+            'visit_count' => 0,
+            'first_visited_at' => null,
+            'last_visited_at' => null,
+        ]);
+
+        // Update visit data.
+        $data['visit_count']++;
+        $data['first_visited_at'] = $data['first_visited_at'] ?? now()->timestamp;
```
Jamiras, line 704: Wouldn't it be easier to just put this in the initial values on line 698?

app/Community/Requests/StoreBetaFeedbackRequest.php
wescopeland: Yep, it's blowing up on empty strings. Fixed in latest.

database/migrations/2025_09_13_000000_create_user_beta_feedback_submissions_table.php
wescopeland: Fixed in latest.

database/migrations/2025_09_13_000000_create_user_beta_feedback_submissions_table.php
wescopeland: Removed `updated_at` in latest (and updated the model accordingly).

app/Policies/UserBetaFeedbackSubmissionPolicy.php
wescopeland: I wasn't planning on exposing any of this data. Metabase probably makes the most sense. I was planning on querying for it and modeling stuff by hand.

I've updated this policy to always return false.

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland: Updated in latest.

PR #3902, opened 2025-09-13
feat(game): show user current leaderboard position

wescopeland:
This PR enhances the leaderboards display on React game pages to show the user's current leaderboard positions (if they have any):

Screenshot 2025-09-13 at 5 31 32 PM

### DIFF
--- a/resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
+++ b/resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
@@ -1,8 +1,10 @@
 import * as motion from 'motion/react-m';
 import type { FC } from 'react';
+import { useTranslation } from 'react-i18next';
 import { LuChartBar, LuCrown } from 'react-icons/lu';
 
 import { UserAvatar } from '@/common/components/UserAvatar';
+import { useFormatPercentage } from '@/common/hooks/useFormatPercentage';
 
 interface LeaderboardsListItemProps {
   index: number;
@@ -15,6 +17,10 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
   isLargeList,
   leaderboard,
 }) => {
+  const { t } = useTranslation();
+
+  const { formatPercentage } = useFormatPercentage();
+
   return (
     <motion.li
       className="flex w-full gap-x-3.5 px-2 py-3 odd:bg-[rgba(50,50,50,0.4)] light:odd:bg-neutral-100 md:gap-x-3 md:py-1"
@@ -38,8 +44,8 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
         </a>
       </div>
 
-      <div className="grid w-full gap-x-5 gap-y-1.5 pb-2.5 leading-4 md:grid-cols-6">
-        <div className="md:col-span-4 md:mt-1">
+      <div className="grid w-full gap-x-5 gap-y-1.5 pb-2.5 leading-4 sm:grid-cols-6">
+        <div className="sm:col-span-4 md:mt-1">
           {/* Title */}
           <div className="mb-0.5 md:mt-0">
             <span className="mr-2">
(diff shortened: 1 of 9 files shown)


resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
```diff
@@ -67,6 +73,34 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
...
+        {leaderboard.userEntry ? (
+          <div className="flex flex-col gap-1 sm:col-span-2 sm:mt-1 sm:items-end sm:justify-center">
+            <div className="flex items-center justify-between sm:flex-col sm:items-end sm:justify-normal sm:gap-0.5">
+              <p className="text-neutral-400 light:text-neutral-700">{t('Your Position')}</p>
+
+              <p className="sm:text-lg">
+                <span className="text-neutral-300 light:text-neutral-700">
+                  {t('#{{rank, number}}', { rank: leaderboard.userEntry.rank })}
+                </span>
+                {' / '}
```
Jamiras: I'm not sure this is the best separator to use. As it stands I'm expecting to see "my rank / all ranks", not "my rank / my score". 

Maybe use a colon or an arrow/greater than?
* my rank: my score
* my rank -> my score

Or parentheses:
* my score (my rank)

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
```diff
@@ -67,6 +73,34 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
...
+                {leaderboard.userEntry.formattedScore}
+              </p>
+            </div>
+
+            {leaderboard.userEntry.percentageOfTopScore ? (
+              <p className="mt-0.5 hidden text-xs text-neutral-500 sm:block">
+                {t('{{percentage}} of #1', {
+                  percentage: formatPercentage(leaderboard.userEntry.percentageOfTopScore / 100, {
```
Jamiras: I'm really having trouble wrapping my head around this calculation for speedrun leaderboards:

Image

I think this is saying that #1 is 25% faster than me, but it took me several minutes to make that assumption.

For "lower is better", we should change the wording to "(100-N)% more than" instead of "N% of". I might even go so far as to suggest "(100-N)% less than" for "higher is better" leaderboards.

Jamiras: Or maybe just list the # 1 value and call it good:

Image

Which I just realized is shown on the left side

Image

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
```diff
@@ -67,6 +73,34 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
             ) : null}
           </div>
         </div>
+
+        {/* User entry */}
+        {leaderboard.userEntry ? (
+          <div className="flex flex-col gap-1 sm:col-span-2 sm:mt-1 sm:items-end sm:justify-center">
+            <div className="flex items-center justify-between sm:flex-col sm:items-end sm:justify-normal sm:gap-0.5">
+              <p className="text-neutral-400 light:text-neutral-700">{t('Your Position')}</p>
```
Jamiras: I'd posit that "Your score" or maybe "Your entry" would work better here.

Or maybe even just "You: " in front of the rank/score.

Jamiras: It feels redundant to have this text on every leaderboard.

Image

Maybe put it in a tooltip?

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
wescopeland: Agreed. I've changed this to use an interpunct in latest. Happy to adjust it further:
Screenshot 2025-09-14 at 3 38 56 PM

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
wescopeland: After looking at this with a fresh set of eyes, I think this current design for showing the user's rank+score is a bit too busy.

In latest, I've removed everything other than those two values. "Your Position" and "N% of #M" have both been removed.

PR #3904, opened 2025-09-13
feat: remove all Discord roles when a user is muted or banned

wescopeland:
Verified as working from a volunteer in the main Discord.

This PR contains code to strip a user's Discord roles when the user is either muted or banned, assuming they're in our Discord server.


app/Observers/UserObserver.php
```diff
@@ -0,0 +1,59 @@
...
+use Carbon\Carbon;
+
+class UserObserver
+{
+    public function updating(User $user): void
+    {
+        // Bail as soon as we can if it's not an update we care to observe.
+        if (!$user->isDirty('muted_until') && !$user->isDirty('banned_at')) {
+            return;
+        }
```
Jamiras: Recommend eliminating this check. It's an early-out, which makes it harder to add additional logic at a later time, and the `isBeingMuted` and `isBeingBanned` functions both perform these exact checks making it redundant.

app/Observers/UserObserver.php
Jamiras, line 1: I understand why we'd want to strip roles from a banned user, but why from a muted user? Doesn't discord have it's own way to mute users? If so, I would expect that to also be exposed to the API.

Otherwise, the user won't be able to follow any discussions they were already involved in (particularly the events channels). And when the user's mute period expires, all of their roles would have to be manually restored.

wescopeland: Discord only allows a maximum timeout of 7 days, and I'm not sure if it's easily-manageable with their web API.

There have been several instances now where we've muted someone and they've gone on to make a scene in public channels. I'm fine with a little inconvenience if we can avoid situations such as this.

app/Observers/UserObserver.php
wescopeland: I'm waffling back and forth on this and want to share my perspective on why we may want to keep this early-out.

`UserObserver::updating` is going to fire on every `->save()` call for the user, which may happen frequently for peripheral fields. With the early-out, we can at least immediately bail for the great majority of user updates. Without it, we'll be calling two methods on most user updates just to have them return false.

Jamiras: > UserObserver::updating is going to fire on every ->save() call for the user, which may happen frequently for peripheral fields.

Yep. And the fields that affect the logic aren't going to change often.

> Without it, we'll be calling two methods on most user updates just to have them return false.

`isDirty` is a function call that we're calling twice on every user update and up to four times if there's a match. Using the early returns in the `isBeing` functions adds very little additional overhead.

> With the early-out, we can at least immediately bail for the great majority of user updates. 

My bigger concern is when we decide to watch some other field, we have to put that logic before the early return (and early returns in the middle of a function are generally frowned upon). It would be better to invert the early return:
```
    public function updating(User $user): void
    {
        if ($user->isDirty('muted_until') || !$user->isDirty('banned_at')) {
            if ($this->isBeingMuted($user) || $this->isBeingBanned($user)) {
                (new RemoveDiscordRolesAction())->execute($user);
            }
        }

        // other logic can still go here
    }
```
And hopefully that makes it more clear why I feel the double-check is redundant.

Jamiras (approved):
This code is functional. I don't feel strongly enough about eliminating the early return to block it.

app/Observers/UserObserver.php
wescopeland: I get it now - removed the double check in latest:
```php
public function updating(User $user): void
{
    if ($this->isBeingMuted($user) || $this->isBeingBanned($user)) {
        (new RemoveDiscordRolesAction())->execute($user);
    }
}
```

PR #3906, opened 2025-09-13
fix(getCodeNoteCounts): exclude soft-deleted rows

wescopeland:
https://discord.com/channels/310192285306454017/962817919698501702/1416556810952904734

### DIFF
--- a/app/Helpers/database/code-note.php
+++ b/app/Helpers/database/code-note.php
@@ -45,7 +45,8 @@ function getCodeNoteCounts(User $user): array
               LEFT JOIN GameData AS gd ON gd.ID = mn.game_id
               LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
               WHERE LENGTH(body) > 0
-              AND gd.ID IN (SELECT game_id from memory_notes WHERE user_id = $userId GROUP BY game_id)
+              AND mn.deleted_at IS NULL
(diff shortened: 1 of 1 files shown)


app/Helpers/database/code-note.php
```diff
@@ -45,7 +45,8 @@ function getCodeNoteCounts(User $user): array
               LEFT JOIN GameData AS gd ON gd.ID = mn.game_id
               LEFT JOIN Console AS c ON c.ID = gd.ConsoleID
               WHERE LENGTH(body) > 0
-              AND gd.ID IN (SELECT game_id from memory_notes WHERE user_id = $userId GROUP BY game_id)
+              AND mn.deleted_at IS NULL
+              AND gd.ID IN (SELECT game_id from memory_notes WHERE user_id = $userId AND deleted_at IS NULL GROUP BY game_id)
```
Jamiras: Would adding a `DISTINCT` to this subquery help performance at all?

wescopeland: Unfortunately, using `DISTINCT` doesn't seem to help performance at all.

After a little bit of digging, it seems the MariaDB internal optimizer is treating them identically when they're used on a single column. However, I think `DISTINCT` is probably more understandable, so I've made the change in latest.

PR #3907, opened 2025-09-14
chore: remove more dead code

wescopeland:
Another dead code elimination PR. Most of this code is to sync tables or support Livewire UI components. No functional changes.


Jamiras (approved):
Did some basic smoketesting in areas around where I though the components might be used. Didn't encounter any issues.

Did not dig into the removed commands.

If anything turns out to be needed, it shouldn't be too hard to restore from this PR.

PR #3909, opened 2025-09-14
perf(game): optimize hub loading

wescopeland:
http://localhost:64000/game/668

This PR reduces `GameSet` model loads by ~76% on game ID 668 (84 -> 20) and eliminates an 8ms query caused by expensive dependent subqueries.

In `master`, hubs are currently loaded twice on game2. Once in `LoadGameWithRelationsAction`, and again in `BuildSeriesHubDataAction`. All child hubs are loaded just to check if a hub is an event hub, and `withCount('games')` is an expensive subquery that runs for each hub linked.

Now, hubs are loaded only once, without children or counts. We use a single `GROUP BY` for the games count, and we cache event hub IDs behind an SWR (`flexible`) cache. This removes a bunch of unnecessary child hub model loads.

From a UI POV, there should be no functional changes as part of this PR.


app/Platform/Services/EventHubIdCacheService.php
```diff
@@ -0,0 +1,44 @@
...
+                ->whereIn('parent_game_set_id', [
+                    GameSet::CommunityEventsHubId,
+                    GameSet::DeveloperEventsHubId,
+                ])
+                ->pluck('child_game_set_id');
+
+            $titleEventHubIds = GameSet::where('title', 'like', '%Events -%')
+                ->pluck('id');
+
+            return $childEventHubIds->merge($titleEventHubIds)->unique();
```
Jamiras: Is there a benefit to storing this as a collection? I assume an array would be more performant in both memory allocation and scanning.

wescopeland: I didn't think of that - swapped to an array in latest.

PR #3910, opened 2025-09-14
fix(game): don't show empty progress tooltips

wescopeland:
image-1

Empty Recent Players progress tooltips appear for players who haven't unlocked any achievements yet for the game. This PR resolves the issue.


app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland, line 1: These props are unused - removing them from the SSR payload.

resources/js/common/components/PlayerGameProgressBar/PlayerGameProgressBar.tsx
```diff
@@ -115,7 +115,7 @@ export const PlayerGameProgressBar: FC<PlayerGameProgressBarProps> = ({
   const isTooltipDisabled = achievementsUnlocked === 0 || !isTooltipEnabled;
 
   return (
-    <BaseTooltip>
+    <BaseTooltip open={!achievementsUnlocked ? false : undefined}>
```
Jamiras, line 125: In addition to removing the tooltip, this also removes the link.

Also, the link appears to be broken. It should go to the game compare page for the user. Instead it just goes to the game1 page for the game.

wescopeland: I figured removing the link was desirable given there's nothing to compare to (the bar is effectively in an empty state).

In latest, I've fixed the `href` - the link is still removed. I don't feel overly strongly about this though, if you'd like to see the link present, I can try to make it always appear for these bars.

Jamiras: It's a first step towards getting to the "View User Game Activity" page where we can look at the emulator version information for the user.

wescopeland: Not a problem. The hrefs should now always render in latest.

Jamiras (approved):

PR #3913, opened 2025-09-14
feat(game): allow sorting the achievement list by orderColumn

wescopeland:
https://github.com/RetroAchievements/RAWeb/pull/3678#discussion_r2207445298
https://github.com/RetroAchievements/RAWeb/discussions/1226
https://discord.com/channels/310192285306454017/453242743292952578/1406274490148589703

This PR allows users to sort game page achievement lists by the original display order when they have achievements unlocked in the achievement set.

When a user is unauthenticated or has no achievements unlocked, the default sort order remains as "Display order (first)".

When the user has some achievements unlocked, two new sort orders appear: "Unlocked first" and "Locked first". The "Unlocked first" order becomes the default order.

Once all achievements are unlocked, these two options are removed and "Display order (first)" once again becomes the default.

**Before**
Screenshot 2025-09-14 at 5 56 33 PM

**After**
Screenshot 2025-09-14 at 5 56 03 PM

Screenshot 2025-09-14 at 5 56 13 PM


app/Platform/Enums/GamePageListSort.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace App\Platform\Enums;
+
+use Spatie\TypeScriptTransformer\Attributes\TypeScript;
+
+#[TypeScript]
+enum GamePageListSort: string
+{
+    case Normal = 'normal';
+    case NormalReverse = '-normal';
```
Jamiras: I'm not sure if `NormalReverse` should be changed or just removed. In addition to grouping the unlocked achievements first, it also reverses the sort on display order.

Normal:

Image

NormalReverse:

Image

This puts the last locked achievement as the first to show in the list (which is technically correct for a reverse sort). If I were to use that functionality, I'd want the locked achievements to still appear in display order so I can see which achievements are the most likely for me to unlock next. But that functionality is already available via the Normal sort and toggling the Locked Only filter.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
Jamiras, line 1: The Missable and Locked Only filters are persisted via cookie (see #3678). The selected sort doesn't seem to be persisted. After selecting a sort and refreshing it goes back to Normal.

wescopeland: Let's do selected sort persistence in a subsequent PR. It's in my Tier 2 list and this PR sets the stage for it quite nicely with the `currentPlayableListSortAtom` now being hydrated from a page prop in _[game].tsx_:
```tsx
useHydrateAtoms([
  [currentListViewAtom, initialView],
  [currentPlayableListSortAtom, initialSort], // <------------
  [isLockedOnlyFilterEnabledAtom, isLockedOnlyFilterEnabled],
  [isMissableOnlyFilterEnabledAtom, isMissableOnlyFilterEnabled],
  //
]);
```
For anyone reading who may not know, atoms are basically global `useState()` hooks. They're good for managing client-side state that changes frequently and is also shared across several different components.

app/Platform/Enums/GamePageListSort.php
wescopeland: I like the idea of erring on the side of simplicity and removing `NormalReverse`. Removed in latest.

PR #3914, opened 2025-09-14
fix(SeriesHubDisplay): rename bad prop

wescopeland:
Typo fix. Renames a prop that's causing i18next to throw a "missing key" warning in the browser devtools console:
Screenshot 2025-09-14 at 6 15 56 PM


Jamiras (approved):
I don't actually see the error, but I can confirm that the "points" label is not being translated on master and is being translated on this branch.

PR #3917, opened 2025-09-17
feat(game): add new mobile-only header component

wescopeland:
This PR adds a new **mobile-only** header component to the React game page. This is the first of several PRs that will attempt to streamline the page's mobile UX.

**Before**
Screenshot 2025-09-16 at 7 21 02 PM

The current UX has several problems:
1. The height is variable based on the image sizes.
2. A design goal of the new page is to lift the achievements list as high on the screen as possible. Because we're space-constrained, the current mobile UX doesn't do a very good job at this.
3. Extra-tall screenshots (ie: Nintendo DS) wind up behaving bizarrely in the media carousel, and this likely cannot be fixed.
4. To compound pushing the achievements list down, the UX wastes a lot of space on the right-hand side of the screen while doing so.

---

**After**
Screenshot 2025-09-16 at 7 21 09 PM

This UX solves virtually all of the noted challenges above:
1. The height of the component is always fixed and predictable.
2. The achievement list is pulled much further up the page.
3. The in-game screenshot and H1 are blended into a seamless single component with much less wasted space for mobile displays.

There are a few measured trade-offs with this approach:
1. The in-game screenshots weren't really uploaded with this kind of UX in mind. The component itself applies a _lot_ of CSS filtering to make things as workable as is reasonably possible. At some point, I think it'll be desirable to allow devs/game editors to upload higher quality banner images for games. We have something of a chicken-and-egg problem here.
2. Title images are no longer visible on mobile. I'm not convinced this is actually much of a loss. The in-game screenshot was chosen over the title image because the title image is going to be oftentimes duplicative with the H1 itself.

Additionally, the events mobile UI is unaffected by this change.

---

**How to test this PR:**
The new component only appears when a mobile user agent is active, so you will need to use the device simulator in the browser devtools.

### DIFF
--- a/.github/workflows/crowdin-download.yml
+++ b/.github/workflows/crowdin-download.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: 9
+          version: 10.17
       
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
--- a/.github/workflows/crowdin-upload.yml
+++ b/.github/workflows/crowdin-upload.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: 9
+          version: 10.17
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile


wescopeland:
Going to wait until https://github.com/RetroAchievements/RAWeb/pull/3900 is approved before merging.

PR #3921, opened 2025-09-18
chore: fix a bunch of PHPStan 2.0 errors

wescopeland:
https://discord.com/channels/476211979464343552/1026595325038833725/1417999536923611266

No functional changes.

PHPStan 2.0 is throwing hundreds of errors. It'd be nice to upgrade before moving to Laravel 12. This PR fixes a handful of these errors, all in the same error class (unnecessary safe navigation with `?->`).


app/Community/Commands/MigrateTicketCommentMetadata.php
```diff
@@ -127,7 +127,7 @@ private function syncMetadataForTicket(Ticket $ticket): bool
 
         $userAgentService = new UserAgentService();
         $userAgents = PlayerSession::where('user_id', $ticket->reporter_id)
-            ->where('game_id', $ticket->achievement?->game?->id ?? 0)
+            ->where('game_id', $ticket->achievement?->game->id ?? 0)
```
Jamiras, line 130: Why is the `?->` safe on achievement, but not game?

app/Community/Actions/LoadThinActivePlayersListAction.php
```diff
@@ -55,7 +55,7 @@ function () use ($lookbackMinutes, $minimumPermissions) {
                         'username' => $player->username,
                         'display_name' => $player->display_name,
                         'rich_presence' => $player->rich_presence,
-                        'game_title' => $player->game?->Title ?? '',
+                        'game_title' => $player->game->Title ?? '',
```
Jamiras, line 58: In general, what error are these changes trying to address?

There's a `with(['game'])` clause on the query, but what if that returns a null game? Isn't `?->` safer?

app/Community/Controllers/MessageThreadController.php
```diff
@@ -107,8 +107,8 @@ public function create(Request $request, ?User $user = null): InertiaResponse
             templateKind: $request->input('templateKind')
                 ? MessageThreadTemplateKind::tryFrom($request->input('templateKind'))
                 : null,
-            senderUserAvatarUrl: $teamAccount?->avatar_url ?? null,
-            senderUserDisplayName: $teamAccount?->display_name ?? $user->display_name,
+            senderUserAvatarUrl: $teamAccount->avatar_url ?? null,
+            senderUserDisplayName: $teamAccount->display_name ?? $user->display_name,
```
Jamiras, line 111: `$teamAccount = $user`, and `?User $user = null`. Why is this wrong?

Does the `??` somehow prevent the exception when processing the earlier `a->b` chain?

wescopeland:
I think there are a sufficient number of false positives thrown by PHPStan 2 we should disable the rule that's causing this problem. The rule this PR tried to make happy is [identifier: nullsafe.neverNull](https://phpstan.org/error-identifiers/nullsafe.neverNull).

I'm going to close this PR in favor of https://github.com/RetroAchievements/RAWeb/pull/3924, which is likely a more sustainable approach to upgrading our version of PHPStan.

PR #3923, opened 2025-09-19
feat(manage): update emulator releases and user agents relation managers

wescopeland:
This PR tries to make managing emulator releases and user agents more intuitive for someone who doesn't necessarily have DB access or programming knowledge.

**Releases Table Before**
Screenshot 2025-09-19 at 10 13 56 AM

**Releases Table After**
Screenshot 2025-09-19 at 10 14 00 AM

**Releases Form Before**
Screenshot 2025-09-19 at 10 14 06 AM

**Releases Form After**
Screenshot 2025-09-19 at 10 14 11 AM

**User Agents Table Before**
Screenshot 2025-09-19 at 10 14 19 AM

**User Agents Table After**
Screenshot 2025-09-19 at 10 14 25 AM

**User Agents Form Before**
Screenshot 2025-09-19 at 10 14 31 AM

**User Agents Form After**

Screenshot 2025-09-19 at 10 14 38 AM


app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
```diff
@@ -30,46 +30,73 @@ public function form(Form $form): Form
...
-                    ->date(),
+                    ->date()
+                    ->required(),
 
-                Forms\Components\Toggle::make('stable'),
+                Forms\Components\Select::make('stable')
+                    ->label('Release Type')
+                    ->options([
+                        '1' => 'Stable',
+                        '0' => 'Beta',
```
Jamiras: This is really any sort of nightly that we'd want to distribute/track for one reason or another. It doesn't necessarily indicate a beta release. Maybe "Pre-release", "Custom", or "One-off"?

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
```diff
@@ -30,46 +30,73 @@ public function form(Form $form): Form
...
+                    ])
+                    ->default('1')
+                    ->required()
+                    ->helperText('Stable releases can be set as minimum and are reported via the Connect API. Beta releases are experimental.'),
 
-                Forms\Components\Toggle::make('minimum'),
+                Forms\Components\Toggle::make('minimum')
+                    ->label('Set as recommended minimum version')
+                    ->helperText('Sets the minimum version reported via the Connect API. Note: This does NOT block connections - use the User Agents tab for actual blocking.'),
```
Jamiras: This is technically misleading. If the user tries to log in using a version older than the minimum version, they'll get this dialog:

Image

While it doesn't block connections, it does block hardcore. It's redundant with the user agent's minimum hardcore version (and should probably be sychronized with it). But it's enforced by the DLL. This is how we did minimum version enforcement before we had the user agent functionality. It's also probably a better UX than just relying on the warning achievement.

The user agent's minimum allowed version is the only thing that actually blocks a client.

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
```diff
@@ -30,46 +30,73 @@ public function form(Form $form): Form
...
             ]);
     }
 
     public function table(Table $table): Table
     {
         return $table
             ->recordTitleAttribute('title')
+            ->description('Manage emulator version information reported to clients. The minimum version setting here is informational only and does NOT block connections. To actually block old versions, use the User Agents tab.')
```
Jamiras: Might be worth noting that this information is purely informational for any clients not using the DLL.

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
wescopeland: Updated to "Pre-release" in latest.

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
wescopeland: Updated this copy in latest to:
> Sets the minimum version for hardcore mode (enforced by the toolkit DLL). Users with older versions can still connect but cannot use hardcore. To block connections entirely, use the User Agents tab.

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorReleasesRelationManager.php
wescopeland: Updated this copy in latest to:
> Manage emulator version information reported to clients. The minimum version blocks hardcore mode for DLL-based clients only (purely informational for other clients). To block connections entirely, use the User Agents tab.

PR #3927, opened 2025-09-20
refactor: extract most mutation hooks into separate files

wescopeland:
No functional changes. This is a cleanup PR.

This PR extracts most `useMutation` hooks into their own separate files, neatly isolating the mutation code from actual logic code. All tests continue to pass.

**How to test this PR**
Since the React integration tests are passing and coverage is at 100%, I expect little to no issues to arise from this PR. The page with the most mutations is the user settings page. Try changing a few settings as a smoke test to verify that things still work.


Jamiras (approved):
Did some spot-testing in settings, forums, and tickets.

PR #3933, opened 2025-09-21
feat(game): add mobile tabs

wescopeland:
This PR adds tabs to the React game page mobile view:

Screenshot 2025-09-21 at 4 21 57 PM

Screenshot 2025-09-21 at 4 25 16 PM

---

There are four tabs:
* Achievements - shows the achievement list and controls which manipulate the list.
* Info - shows game meta, resources, hubs, and similar games.
* Stats - shows playtime stats, compare progress, achievement distribution, and top achievers.
* Community - shows recent players and comments.

---

Additionally, on the achievement list, a new mobile-only button has been added which appears if the user starts scrolling up after scrolling ~1000px down the page:

https://github.com/user-attachments/assets/0d36d3e8-a713-451d-ba7c-7e15dc033cd1


resources/js/features/games/components/+show-mobile/GameShowMobileRoot.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+            <BaseTabsTrigger value="stats" variant="underlined">
+              {t('Stats')}
+            </BaseTabsTrigger>
+
+            {isViewingPublishedAchievements ? (
+              <BaseTabsTrigger value="community" variant="underlined">
+                {t('Community')}
+              </BaseTabsTrigger>
+            ) : null}
```
Jamiras, line 96: When viewing a subset, the tab content changes for the subsets (which is good). Should the subset selector be available on all tabs? Or at least some indicator that you're viewing subset data?

resources/js/features/games/components/+show-mobile/GameShowMobileRoot.tsx
```diff
@@ -0,0 +1,166 @@
...
+  return (
+    <div data-testid="game-mobile" className="flex flex-col gap-3">
+      {currentTab === 'achievements' ? <ScrollToTopButton /> : null}
+
+      {hasMatureContent ? <MatureContentWarningDialog /> : null}
+      {allPageAchievements.length ? <ResetAllProgressAlertDialog /> : null}
+
+      <GameMobileHeader />
+
+      <BaseTabs value={currentTab} onValueChange={setCurrentTab}>
```
Jamiras, line 74: This could potentially be a problem with translations.

Image

It does wrap if there's whitespace.

Image

But goes offscreen if there isn't.

Image

resources/js/features/games/components/+show-mobile/GameShowMobileRoot.tsx
```diff
@@ -0,0 +1,166 @@
...
+              {t('Achievements')}
+            </BaseTabsTrigger>
+
+            <BaseTabsTrigger value="info" variant="underlined">
+              {t('Info')}
+            </BaseTabsTrigger>
+
+            <BaseTabsTrigger value="stats" variant="underlined">
+              {t('Stats')}
+            </BaseTabsTrigger>
```
Jamiras: This page is blank for games with no players:

Image

resources/js/features/games/components/+show-mobile/GameShowMobileRoot.tsx
wescopeland: Addressed in latest.

Going off-screen is fine (and likely unpreventable/unpredictable), but when that occurs the div should be scrollable. In latest, it should be.

resources/js/features/games/components/+show-mobile/GameShowMobileRoot.tsx
wescopeland: Fixed in latest - that tab will no longer be shown under this circumstance.

resources/js/features/games/components/+show-mobile/GameShowMobileRoot.tsx
wescopeland: This is a really tricky problem.

The data hierarchy is such that the set selection is driving the page, but the selector lives within a tab. We could lift the selector out of the tab and make it always-visible, but then we have persistent tabs within tabs, which I'm not a fan of.

I think to ultimately fix this, we may need a different metaphor for selecting the set on mobile that isn't tabs. I have an idea, but I want to spend more time thinking about it.

In the interim, what I've done is make it so the badge in the header now reflects the selected set for mobile instead of just the game:
Screenshot 2025-09-22 at 5 40 27 PM

PR #3934, opened 2025-09-22
show tooltips for addresses in the middle of a 'N bytes' note

Jamiras:
Some examples:
http://localhost:64000/achievement/117890
http://localhost:64000/achievement/106667
http://localhost:64000/achievement/187964

image

### DIFF
--- a/app/Platform/Services/TriggerDecoderService.php
+++ b/app/Platform/Services/TriggerDecoderService.php
@@ -359,7 +359,10 @@ class TriggerDecoderService
         }
 
         $codeNotes = MemoryNote::where('game_id', $gameId)
-            ->whereIn('address', $memoryReferences)
+            ->where(function ($q) use ($memoryReferences) {
+                $q->whereIn('address', $memoryReferences);
+                $q->orWhere('body', 'like', '%bytes%');
+            })
             ->get()
             ->mapWithKeys(function ($row, $key) {
                 return [$row['address'] => $row['body']];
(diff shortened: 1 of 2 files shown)


app/Platform/Services/TriggerDecoderService.php
```diff
@@ -496,6 +510,41 @@ private function getIndirectNote(string $parentNote, int $offset): string
         return '';
     }
 
+    private function findArrayNote(int $address, array $codeNotes): ?int
+    {
+        $lastAddress = null;
```
wescopeland: nit: This variable is unused.

Jamiras:
> I noticed it's possible to load N models into memory (see http://localhost:64000/achievement/482020), but after doing some benchmarking, I found the memory footprint is actually really small - to the point we probably shouldn't prematurely optimize.

That's mostly from 4-byte notes:
```
Medal Data - Noneffective Artist [4 Bytes]
```
I'm not sure how to smartly discard those. The preferred way to annotate them would be "[32-bit]".

Also, the models aren't really necessary. The query is using `mapWithKeys` to extract the `address` and `body`. Maybe you have a way to circumvent creating the models?

app/Platform/Services/TriggerDecoderService.php
Jamiras: removed.

wescopeland:
> Also, the models aren't really necessary. The query is using mapWithKeys to create a dictionary of address => body. Maybe you have a way to circumvent creating the models?

The only way I can think of is to use the `DB` facade - but it's probably not worth it for this and feels slightly error-prone (ie: we have to explicitly filter out soft-deleted records).

PR #3935, opened 2025-09-22
add GetLatestClientVersionAction

Jamiras:


app/Connect/Actions/GetLatestClientVersionAction.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $emulator = Emulator::find($this->emulatorId);
+        if ($emulator === null || !$emulator->active || !$emulator->latestRelease) {
+            return $this->resourceNotFound('client');
+        }
+
+        // turn relative URLs into full URLs
+        $format_url = function (?string $url): ?string {
+            return (!$url || str_starts_with($url, 'http')) ? $url : config('app.url') . '/' . $url;
+        };
```
wescopeland: nit: I know it's copied from _dorequest.php_, but I wonder if this may be better-suited as a private function of the class, just for consistency.

The comment could probably be eliminated if the function name better described what the function was doing, eg: `$this->makeAbsoluteUrl()` or `$this->toFullUrl()`.

wescopeland (approved):

app/Connect/Actions/GetLatestClientVersionAction.php
Jamiras: done.

wescopeland (approved):

PR #3940, opened 2025-09-24
fix(game): don't show empty 'Development' section in sidebar

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1420202141183901716.

**Root Cause**
The sections already have logic to not mount if `children` isn't truthy. However, when the Development section has nothing to render, it returns an empty fragment (`<></>`), which is not being properly handled by this logic.


Jamiras (approved):
There's still an HR coming from somewhere:

Image

PR #3952, opened 2025-09-24
fix(game): show unlock percentages correctly on mobile

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1420440601123487827.

Unlock percentages are not currently displaying on achievement items on mobile.

**Before**
Screenshot 2025-09-24 at 5 24 57 PM

**After**

Screenshot 2025-09-24 at 5 24 20 PM

### DIFF
--- a/resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
+++ b/resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
@@ -172,7 +172,11 @@ export const AchievementsListItem: FC<AchievementsListItemProps> = ({
             </p>
 
             <p className="mb-0.5 flex gap-x-1 text-2xs md:mb-0 md:justify-center md:text-center">
-              <ProgressBarMetaText achievement={achievement} playersTotal={playersTotal} />
+              <ProgressBarMetaText
+                achievement={achievement}
+                playersTotal={playersTotal}
+                unlockPercentage={unlockPercentage}
+              />
             </p>
 
             <BaseProgress
--- a/resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
+++ b/resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
@@ -7,23 +7,24 @@ import { formatPercentage } from '@/common/utils/l10n/formatPercentage';
 interface ProgressBarMetaTextProps {
   achievement: App.Platform.Data.Achievement;
   playersTotal: number;
+  unlockPercentage: number;
 }
 
 export const ProgressBarMetaText: FC<ProgressBarMetaTextProps> = ({
   achievement,
   playersTotal,
+  unlockPercentage,
 }) => {
   const { t } = useTranslation();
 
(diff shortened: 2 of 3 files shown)


resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
```diff
@@ -172,7 +172,11 @@ export const AchievementsListItem: FC<AchievementsListItemProps> = ({
             </p>
 
             <p className="mb-0.5 flex gap-x-1 text-2xs md:mb-0 md:justify-center md:text-center">
-              <ProgressBarMetaText achievement={achievement} playersTotal={playersTotal} />
+              <ProgressBarMetaText
+                achievement={achievement}
+                playersTotal={playersTotal}
+                unlockPercentage={unlockPercentage}
```
Jamiras: I see this fixing the issue, but I don't understand why.
```
const unlockPercentage = achievement.unlockPercentage ? Number(achievement.unlockPercentage) : 0;
```

Is it just the changes from `unlockHardcorePercentage` to `unlockPercentage`?
```
const unlockHardcorePercentage = achievement.unlockHardcorePercentage ?? 0;
```
If so, wouldn't that be the easier fix?

resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
```diff
@@ -7,23 +7,24 @@ import { formatPercentage } from '@/common/utils/l10n/formatPercentage';
...
   const unlocksHardcoreTotal = achievement.unlocksHardcoreTotal ?? 0;
   const unlocksTotal = achievement.unlocksTotal ?? 0;
-  const unlockHardcorePercentage = achievement.unlockHardcorePercentage ?? 0;
 
   return (
     <Trans
       i18nKey="<1>{{totalUnlocks, number}}</1> <2>({{totalHardcoreUnlocks, number}})</2> of <3>{{totalPlayers, number}}</3> <4>- {{unlockHardcorePercentage}}</4> <5>unlock rate</5>"
       values={{
-        unlockHardcorePercentage: formatPercentage(Number(unlockHardcorePercentage), {
+        unlockHardcorePercentage: formatPercentage(unlockPercentage, {
```
Jamiras, line 26: should the `i18nKey` also be updated to reflect the non-hardcore-ness of the value?

Jamiras (approved):

resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
wescopeland: Yep, no need for this to be passed down as a prop. Updated in latest.

resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
wescopeland: Yeah, it probably should be. Let's do it as a separate PR so we don't bust the current translation for the hotfix.

I'll open a (non-urgent) PR immediately after merging this one which includes the change.

PR #3954, opened 2025-09-24
chore(game): change ProgressBarMetaText i18n key

wescopeland:
Addresses https://github.com/RetroAchievements/RAWeb/pull/3952#discussion_r2377143781.

The hardcore stat is no longer used for this variable.


lang/en_US.json
Jamiras, line 1: Would manually updating the other language files (it's just a placeholder) interfere with crowdin?

wescopeland: Crowdin won't be smart enough to resync the other languages, unfortunately. It treats the base en_US.json as the only source of record.

PR #3959, opened 2025-09-25
fix(game): on Playtime Stats, always show 'Not enough data' if there's no data

wescopeland:
On newly-released sets, the "Not enough data" indicator+tooltip doesn't currently display for the "Mastered" row due to a logic error:

Screenshot 2025-09-25 at 6 15 21 PM


resources/js/features/games/components/PlaytimeStatistics/PlaytimeRow/PlaytimeRow.tsx
```diff
@@ -91,7 +91,7 @@ export const PlaytimeRow: FC<PlaytimeRowProps> = ({
         </div>
       ) : null}
 
-      {totalSamples !== undefined && totalSamples < MIN_SAMPLES_FOR_MEDIAN ? (
+      {totalSamples === undefined || totalSamples < MIN_SAMPLES_FOR_MEDIAN ? (
```
Jamiras: Not sure what is exactly leading to these outcomes, but clicking on recently released sets, I see the following:

Image

Image

In both, "Not enough data" is showing for unlocked 1 achievement, and "Not enough data" isn't showing for mastery if there aren't any masteries.

"Not enough data" also appears for unlocked 1 achievement on all sets:

Image

Image

wescopeland: Should all be fixed up now as of latest.

Jamiras (approved):

PR #3983, opened 2025-09-28
feat: initial conversion of setRequestors page into Inertia component

Chew:
This PR converts the `setRequestors.php` page into Inertia! It uses styling and logic similar to the Patreon page to load 100 users and then defers the rest.

The page is available at `/game/{gameId}/requests`.

`getSetRequestorsList` was removed from `set-request.php` because it was only used on `setRequestors.php` and was kinda iffy since it only returned a response made for that page. I removed it and just moved the logic to the corresponding `GameController::setRequests()` function since it was never re-used.

Example screenshot is provided below.

CleanShot 2025-09-28 at 17 10 42@2x

I believe a redirector from old links also needs to be added.


app/Platform/Controllers/GameController.php
```diff
@@ -278,4 +282,36 @@ private function checkSubsetGameRedirect(Request $request, Game $game): ?Redirec
 
         return redirect()->route('game2.show', array_merge(['game' => $backingGameSet->game_id], $queryParams));
     }
+
+    /**
+     * Shows the set requestors for a given game
+     */
+    public function setRequests(Request $request, Game $game): InertiaResponse
```
wescopeland: nit: Recommend placing this public function above the private function so all public functions are above private functions in the class.

app/Platform/Controllers/GameController.php
```diff
@@ -278,4 +282,36 @@ private function checkSubsetGameRedirect(Request $request, Game $game): ?Redirec
...
+    {
+        $allRequestors = UserGameListEntry::where('GameID', $game->id)
+            ->where('type', UserGameListType::AchievementSetRequest)
+            ->with('user')->get();
```
wescopeland: nit: style consistency.
```suggestion
            ->with('user')
            ->get();
```

app/Platform/RouteServiceProvider.php
```diff
@@ -76,6 +76,7 @@ protected function mapWebRoutes(): void
                 Route::get('game/{game}/hashes', [GameHashController::class, 'index'])->name('game.hashes.index');
                 Route::get('game/{game}/top-achievers', [GameTopAchieversController::class, 'index'])->name('game.top-achievers.index');
                 Route::get('game/{game}/suggestions', [GameController::class, 'suggestSimilar'])->name('game.suggestions.similar');
+                Route::get('game/{game}/requests', [GameController::class, 'setRequests'])->name('game.requests.index');
```
wescopeland: Behavior change: The existing page is behind auth (see https://retroachievements.org/setRequestors.php?g=20649 while unauthenticated).

For the time being, let's put this migrated page behind auth too. To achieve this, I think all you'll need to do is move this route definition to the group defined near the bottom of the file.

resources/js/features/game-list/utils/column-definitions/buildNumRequestsColumnDef.tsx
```diff
@@ -52,7 +52,7 @@ export function buildNumRequestsColumnDef<TEntry extends App.Platform.Data.GameL
       }
 
       return (
-        <a href={`/setRequestors.php?g=${gameId}`} className="transition hover:text-link">
+        <a href={`/game/${gameId}/requests`} className="transition hover:text-link">
```
wescopeland: Performance: now that the page is using Inertia, we should use an `<InertiaLink />` component here to ensure a client-side navigation.

wescopeland: Consistency: We'll want to use `route()` from `ziggy-js` here. When migrating stuff away from PHP, we have the added benefit of being able to leverage strongly-typed routes.

The route definitions can be regenerated by executing the `composer types` command in your terminal.

resources/js/features/games/components/AchievementSetEmptyState/AchievementSetEmptyState.tsx
```diff
@@ -44,7 +44,7 @@ export const AchievementSetEmptyState: FC = () => {
               }}
               components={{
                 // eslint-disable-next-line jsx-a11y/anchor-has-content -- this is passed in by the consumer
-                1: auth?.user ? <a href={`/setRequestors.php?g=${backingGame.id}`} /> : <span />,
+                1: auth?.user ? <a href={`/game/${backingGame.id}/requests`} /> : <span />,
```
wescopeland: Performance: now that the page is using Inertia, we should use an `<InertiaLink />` component here to ensure a client-side navigation.

wescopeland: Consistency: We'll want to use `route()` from `ziggy-js` here. When migrating stuff away from PHP, we have the added benefit of being able to leverage strongly-typed routes.

The route definitions can be regenerated by executing the `composer types` command in your terminal.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland, line 1: Testing: we'll definitely want a test suite to cover this component. We shoot for 100% functional coverage.

_PatreonSupportersRoot.test.tsx_ should serve as a good example.

wescopeland, line 1: nit, personal preference: Let's move this from:
* _games/components/GameSetRequests/GameSetRequests.tsx_ 

to:
* _games/components/+requests/GameSetRequestsRoot.tsx_ 

(and rename the component accordingly to `GameSetRequestsRoot`).

resources/js/features/games/components/GameSetRequests/GameSetRequests.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return requestors;
+  }, [initialRequestors, deferredRequestors]);
+
+  return (
+    <div className="flex flex-col gap-5">
+      <h1>{'List of Set Requests'}</h1>
+
+      <div>
+        <GameAvatar {...game} size={96} />
+      </div>
```
wescopeland: Consistency: let's leverage `<GameBreadcrumbs />` and `<GameHeading />` here.

_GameModificationCommentsMainRoot.tsx_ should have a good example.

resources/js/features/games/components/GameSetRequests/GameSetRequests.tsx
```diff
@@ -0,0 +1,50 @@
...
+    if (deferredRequestors) {
+      requestors.push(...deferredRequestors);
+    }
+
+    return requestors;
+  }, [initialRequestors, deferredRequestors]);
+
+  return (
+    <div className="flex flex-col gap-5">
+      <h1>{'List of Set Requests'}</h1>
```
wescopeland: i18n: Careful, this string is untranslated.

For untranslated strings, you'll only need to worry about updating _lang/en_US.json_ (as you've already done in the PR). Once the PR is merged, those source strings are fed out to the translation team automagically (see [here](https://discord.com/channels/310192285306454017/1299083037908406402/1305281296615866378)).

resources/js/features/games/components/GameSetRequests/GameSetRequestsRoot.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <div className="flex flex-col gap-5">
+      <h1>{'List of Set Requests'}</h1>
+
+      <div>
+        <GameAvatar {...game} size={96} />
+      </div>
+
+      <div>{'A set for this game has been requested by the following users:'}</div>
```
wescopeland: i18n: Careful, this string is untranslated.

For untranslated strings, you'll only need to worry about updating _lang/en_US.json_ (as you've already done in the PR). Once the PR is merged, those source strings are fed out to the translation team automagically (see [here](https://discord.com/channels/310192285306454017/1299083037908406402/1305281296615866378)).

resources/js/features/games/components/GameSetRequests/index.js
wescopeland, line 1: Consistency: Let's rename this to _index.ts_. Additionally, it shouldn't export a JS file. The contents should be:
```ts
export * from './GameSetRequests';
```

resources/js/pages/game/[game]/requests.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { AppLayout } from '@/common/layouts/AppLayout';
+import type { AppPage } from '@/common/models';
+import { GameSetRequests } from '@/features/games/components/GameSetRequests';
+
+const SetRequestorsPage: AppPage = () => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <SEO title={t('List of Set Requests')} description="A list of set requests for this game." />
```
Comment (wescopeland) [inline resources/js/pages/game/[game]/requests.tsx]: Consistency: take a peek at _resources/js/pages/game/[game]/modification-comments.tsx_.

All child pages of the game page follow a similar `title` prop format: "Thing - {{gameTitle}}". Let's do something similar here, too.

resources/views/components/game/set-requests.blade.php
```diff
@@ -69,7 +69,7 @@ function submitSetRequest(user, gameID) {
 
 <div>
     <h2 class='text-h4'>Set Requests</h2>
-    <div class='gameRequestsLabel'>Set Requests: <a href='/setRequestors.php?g={{ $gameId }}'>{{ $gameRequests }}</a></div>
+    <div class='gameRequestsLabel'>Set Requests: <a href='/game/{{ $gameId }}/requests'>{{ $gameRequests }}</a></div>
```
wescopeland: We should be able to use `route()` here (albeit, not from Ziggy this time 🙂).

resources/views/components/game/link-buttons/index.blade.php
```diff
@@ -104,7 +104,7 @@
         @can('viewAny', App\Models\Game::class)
             <x-game.link-buttons.game-link-button
                 icon="📜"
-                href="{{ '/setRequestors.php?g=' . $game->id }}"
+                href="{{ '/game/' . $game->id . '/requests' }}"
```
wescopeland: We should be able to use `route()` here (albeit, not from Ziggy this time 🙂).

resources/js/features/games/components/GameSetRequests/GameSetRequestsRoot.tsx
wescopeland: We'll need to figure out different copy to use here due to the upcoming release of the multiset feature. This is going to be a long-winded explanation:

"Multiset" is an upcoming planned feature where emulators will be able to load multiple achievement sets simultaneously. You can see traces of how this is exhibited in the new Inertia game page: https://retroachievements.org/game2/10003

Screenshot 2025-09-28 at 8 41 39 PM  



The legacy site assumes all games have 1 achievement set. In the legacy world, "subsets" are just additional games with funny looking names, like https://retroachievements.org/setRequestors.php?g=34500. In the modern world we're building, sets are simply things that get attached to games, and a game can have N sets attached to it. Hopefully that makes sense.

The link above is another achievement set people can request for TLoZ: Majora's Mask, but it isn't Majora's Mask itself (which already has a set!). That's the nuance.

Maybe a simple solution for now would be for the copy to say something like "This achievement set has been requested by the following users:".

wescopeland:
> I believe a redirector from old links also needs to be added.

Yes - this is handled in the code rather than at the nginx level. You should be able to set this up in _missing-page-redirector.php_.

resources/js/features/games/components/AchievementSetEmptyState/AchievementSetEmptyState.tsx
```diff
@@ -43,8 +43,7 @@ export const AchievementSetEmptyState: FC = () => {
                 val: setRequestData.totalRequests,
               }}
               components={{
-                // eslint-disable-next-line jsx-a11y/anchor-has-content -- this is passed in by the consumer
-                1: auth?.user ? <a href={`/setRequestors.php?g=${backingGame.id}`} /> : <span />,
+                1: auth?.user ? <InertiaLink href={route('game.request.index')} /> : <span />,
```
Jamiras: This doesn't match the route definition:
```
                Route::get('game/{game}/requests', [GameController::class, 'setRequests'])->name('game.requests.index');
```
Additionally, I think you need to pass the game (or at least the game id).

In any case, the link is pointing to `/games/requests`, not the request list for the current game.

app/Platform/Controllers/GameController.php
```diff
@@ -247,6 +251,39 @@ public function suggestSimilar(GameListRequest $request, Game $game): InertiaRes
...
 
+    /**
+     * Shows the set requestors for a given game
+     */
+    public function setRequests(Request $request, Game $game): InertiaResponse
+    {
+        $allRequestors = UserGameListEntry::where('GameID', $game->id)
+            ->where('type', UserGameListType::AchievementSetRequest)
+            ->with('user')
+            ->get();
```
Jamiras: Recommend adding a sort to this list. I am aware that the old page didn't sort either.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
```diff
@@ -0,0 +1,50 @@
...
+    if (deferredRequestors) {
+      requestors.push(...deferredRequestors);
+    }
+
+    return requestors;
+  }, [initialRequestors, deferredRequestors]);
+
+  return (
+    <div className="flex flex-col gap-5">
+      <GameBreadcrumbs game={game} system={game.system} t_currentPageLabel={t('Set Requests')} />
```
Jamiras, line 28: The breadcrumb for the system is blank. Did it not get fetched or passed through the props?

Image

resources/js/pages/game/[game]/requests.tsx
```diff
@@ -0,0 +1,29 @@
...
+import type { AppPage } from '@/common/models';
+import { GameSetRequestsRoot } from '@/features/games/components/GameSetRequests';
+
+const SetRequestorsPage: AppPage<App.Community.Data.GameClaimsCommentsPageProps> = ({ game }) => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <SEO
+        title={t('List of Set Requests - {{gameTitle}}', { gameTitle: game.title })}
```
Comment (Jamiras) [inline resources/js/pages/game/[game]/requests.tsx]: The page title should reflect the title in the page and the breadcrumb. This should be `'Set Requests - {{gameTitle}}'`.

resources/js/features/game-list/utils/column-definitions/buildNumRequestsColumnDef.tsx
```diff
@@ -52,9 +53,9 @@ export function buildNumRequestsColumnDef<TEntry extends App.Platform.Data.GameL
       }
 
       return (
-        <a href={`/setRequestors.php?g=${gameId}`} className="transition hover:text-link">
+        <InertiaLink href={`/game/${gameId}/requests`} className="transition hover:text-link">
```
Jamiras: This should use `route()`.

resources/js/features/games/components/GameSetRequests/GameSetRequestsRoot.tsx
Jamiras: > Maybe a simple solution for now would be for the copy to say something like "This achievement set has been requested by the following users:".

Agreed. And it keeps the wording in line with the page tiltle "Set Requests".

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+      <div>{'A set for this game has been requested by the following users:'}</div>
+
+      <div>
+        <h2 className="border-b-0 text-center text-sm font-semibold sm:text-left sm:text-xl">
+          {t('All requestors ({{val, number}})', { val: totalCount })}
+        </h2>
+        <div className="grid grid-cols-1 gap-3 sm:grid-cols-2 md:grid-cols-3 xl:grid-cols-4">
+          {allRequestors.map((supporter) => (
+            <PatreonSupporterItem key={supporter.id} supporter={supporter} />
```
Jamiras: This should probably be renamed to `UserButtonItem` or something similar (@wescopeland probably has better ideas on how to reuse the component).

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <GameHeading game={game} wrapperClassName="!mb-1">
+        {t('Set Requests')}
+      </GameHeading>
+
+      <div>{'A set for this game has been requested by the following users:'}</div>
+
+      <div>
+        <h2 className="border-b-0 text-center text-sm font-semibold sm:text-left sm:text-xl">
+          {t('All requestors ({{val, number}})', { val: totalCount })}
+        </h2>
```
Jamiras: I'm not sure we want an "All requestors" subheading if there are no other subheadings. Perhaps the count can be merged into the "...by the following users:" string.

Conversely, maybe it would be interesting to have a "Lastest requestors" list, but I don't think that would provide any real benefit - and would be pointless of most games that only have a handful of requestors (or less).

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <div className="flex flex-col gap-5">
+      <GameBreadcrumbs game={game} system={game.system} t_currentPageLabel={t('Set Requests')} />
+      <GameHeading game={game} wrapperClassName="!mb-1">
+        {t('Set Requests')}
+      </GameHeading>
+
+      <div>{'A set for this game has been requested by the following users:'}</div>
+
+      <div>
```
Jamiras, line 41: Recommend adding an empty list special case:

Image

resources/views/components/game/link-buttons/index.blade.php
```diff
@@ -104,7 +104,7 @@
         @can('viewAny', App\Models\Game::class)
             <x-game.link-buttons.game-link-button
                 icon="📜"
-                href="{{ '/setRequestors.php?g=' . $game->id }}"
+                href="route('game.requests.index', ['game' => $game])"
```
Jamiras: I realize this isn't a new link, but I don't know where this button would appear. I don't see it on the desktop sidebar or the mobile info tab.

resources/views/components/game/set-requests.blade.php
```diff
@@ -39,7 +39,7 @@ function getSetRequestInformation(user, gameID) {
             var gameTotal = parseInt(results.gameRequests);
             var thisGame = results.requestedThisGame;
 
-            $('.gameRequestsLabel').html('Set Requests: <a href=\'/setRequestors.php?g=' + gameID + '\'>' + gameTotal + '</a>');
+            $('.gameRequestsLabel').html('Set Requests: <a href=\'/game/' + gameID + '/requests\'>' + gameTotal + '</a>');
```
Jamiras: Another opportunity to use `route()`.

app/Platform/Controllers/GameController.php
Chew: I think it is by default (and luck maybe) sorting sequentially, so it's in order of requests anyway. Do we want to maintain that, or sort by something else?

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
Chew: Yeah this is just a leftover from the "All supporters", I do like seeing the full count so if it can be worded to include that, or just put it somewhere, it can be removed.

And yeah, I didn't see a point of latest requestors. Latest are just at the bottom if we end up sorting in order or request anyway.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
Chew: Maybe changing the text to something like "This game currently has no requests. Could you be the first?"

resources/js/features/games/components/AchievementSetEmptyState/AchievementSetEmptyState.tsx
wescopeland: Careful, `game.request.index` leads to http://localhost:64000/games/requests.

See _app/Community/RouteServiceProvider.php_. I wonder if this newly-converted Inertia page should be `game.request.show` and also live in that RouteServiceProvider file instead of the platform one.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland: Let's also rename the containing folder from _GameSetRequests_ to _+requests_.

resources/js/features/games/components/AchievementSetEmptyState/AchievementSetEmptyState.tsx
wescopeland: @Chew 👋  Some more context here: https://github.com/RetroAchievements/RAWeb/pull/3983#discussion_r2389389735

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland: @Chew 👋 
In `GameController`, you may need to include `'system'` on line 278 (it isn't passed down by default due to being `Lazy`).

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland: Indeed, re-use is good, but features should not depend on other features. When features depend on other features, that may inhibit our future ability to code split by feature module.

`PatreonSupporterItem` should be renamed to something like `UserListLinkItem` and lifted to _resources/js/common/components_.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
Chew: Yep, forgot this one. Updated!

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland: We have an `<EmptyState />` component that could also be leveraged (see _ActivePlayerFeed.tsx_).

resources/views/components/game/link-buttons/index.blade.php
wescopeland: It appears on the legacy page when the game doesn't have a set: https://retroachievements.org/game1/3143

Screenshot 2025-09-29 at 6 22 38 PM

For game2, it's intentionally kept out of the sidebar.

wescopeland:
Hi @Chew! I haven't checked yet, but I'm wondering if the issues you're facing might be related to: https://github.com/RetroAchievements/RAWeb/pull/3983#discussion_r2389389735.

app/Platform/RouteServiceProvider.php
Chew: I will move this down (we could also add `->middleware('auth')` if we want to as well, but that might get messy). It gives me an error when I do this (something about "[app] not found", for some reason, so I haven't done it yet since I need to figure it out.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
Chew: Added the `system` tag and it shows up now. Thanks for that!

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
Chew: I went with `UserGridLinkItem` since the point is more that it's a grid more-so than a simple list (which to me implies just one column), but I can always go with `UserListLinkItem` if desired :)

app/Platform/Controllers/GameController.php
Jamiras: I think an alphabetical sort `->orderBy('user.display_name')` would make the most sense.

app/Community/RouteServiceProvider.php
```diff
@@ -131,6 +132,8 @@ protected function mapWebRoutes(): void
                         Route::get('forums/{category}/{forum}/create', [ForumTopicController::class, 'create'])->name('forum-topic.create');
                         Route::get('forums/post/{comment}/edit', [ForumTopicCommentController::class, 'edit'])->name('forum-topic-comment.edit');
 
+                        Route::get('game/{game}/requests', [GameController::class, 'setRequests'])->name('game.requests.index');
```
wescopeland: DX: Let's rename this to `game.requests.show` to reduce ambiguity between it and `game.request.index`.

resources/js/features/games/components/+requests/GameSetRequestsRoot.test.tsx
wescopeland, line 1: Consistency: This should also use AAA comments like our other front-end test suites. [ref](https://github.com/goldbergyoni/javascript-testing-best-practices?tab=readme-ov-file#-%EF%B8%8F-12-structure-tests-by-the-aaa-pattern)

resources/js/features/games/components/AchievementSetEmptyState/AchievementSetEmptyState.tsx
wescopeland: We should probably change this route name to `game.request.show`.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland: `UserGridLinkItem` works for me 👍 

The props interface likely needs to be renamed though to match:
`UserGridItemProps` -> `UserGridLinkItemProps`

app/Platform/Controllers/GameController.php
Chew: I sorted the resulting collection since I wasn't quite sure how to sort it in the DB query. It does work, though, so let me know if that's alright.

resources/js/features/games/components/+requests/GameSetRequestsRoot.test.tsx
wescopeland, line 1: nit: The `t` import can be removed. In tests, we mock i18next to always dump the en_US stuff.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
wescopeland, line 1: l10n: Looks like there's a pluralization bug on the page.

Screenshot 2025-10-16 at 9 45 27 PM

To fix this, you'll need two keys in _en_US.json_. Something like:
```json
"gameRequestsCount_other": "This achievement set has been requested by the following {{val, number}} users:",
"gameRequestsCount_one": "This achievement has been requested by the following user:"
```
Some locales such as Polish have up to 4-5 plural forms, so doing pluralization through i18next allows those locales to properly accommodate the string.

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
```diff
@@ -0,0 +1,56 @@
...
+        {t('Set Requests')}
+      </GameHeading>
+
+      <div>
+        {allRequestors.length > 0 ? (
+          t('This achievement set has been requested by the following {{val, number}} users:', {
+            val: totalCount,
+          })
+        ) : (
+          <EmptyState>{t('There are currently no active requests.')}</EmptyState>
```
wescopeland, line 37: UX, nit: I recommend wrapping this in a div like:
```tsx
<div className="rounded-lg bg-embed">
  <EmptyState>...</EmptyState>
</div>
```
It's subtle, but makes the empty state feel a bit more polished.

Screenshot 2025-10-16 at 9 49 57 PM

app/Platform/Controllers/GameController.php
```diff
@@ -252,6 +256,41 @@ public function suggestSimilar(GameListRequest $request, Game $game): InertiaRes
...
+     * Shows the set requestors for a given game
+     */
+    public function setRequests(Request $request, Game $game): InertiaResponse
+    {
+        $allRequestors = UserGameListEntry::where('GameID', $game->id)
+            ->where('type', UserGameListType::AchievementSetRequest)
+            ->with('user')
+            ->get()
+            // sort by user display name
+            ->sortBy(fn ($request) => $request->user->username);
```
wescopeland: Bug: This should sort by `display_name`, otherwise users who have customized their display names will be sorted out of order (see top-right user in this screenshot).

Screenshot 2025-10-16 at 9 52 15 PM

resources/js/features/games/components/AchievementSetEmptyState/AchievementSetEmptyState.test.tsx
```diff
@@ -1,3 +1,5 @@
+import { expect } from 'vitest';
```
wescopeland: nit: This is safe to remove - we have it configured as a global in the project.

resources/js/pages/game/[game]/requests.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { GameSetRequestsRoot } from '@/features/games/components/+requests';
+
+const SetRequestorsPage: AppPage<App.Community.Data.GameClaimsCommentsPageProps> = ({ game }) => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <SEO
+        title={t('Set Requests - {{gameTitle}}', { gameTitle: game.title })}
+        description={t('A list of set requests for {{gameTitle}}.', { gameTitle: game.title })}
```
Comment (wescopeland) [inline resources/js/pages/game/[game]/requests.tsx]: nit: No need to translate this specific string. Descriptions are surfaced to search engines, where they'll always need to be in English.

app/Platform/Controllers/GameController.php
Chew: I asked about [this elsewhere](https://github.com/RetroAchievements/RAWeb/pull/3983#discussion_r2422249269) but I wasn't sure how to do this since I don't get the user yet.

wescopeland: Sorry, I missed that. I can confirm this fixes the bug:
```php
->sortBy(fn ($request) => $request->user->display_name);
```

resources/js/pages/game/[game]/requests.tsx
Comment (Chew) [reply resources/js/pages/game/[game]/requests.tsx]: Will this reach search engines if it's behind a login page?

Comment (wescopeland) [reply resources/js/pages/game/[game]/requests.tsx]: No, it won't. I forgot this was behind auth. Either way, we're good to drop the `t()` and en_US values here so the string doesn't get sent to the translation team.

Comment (Chew) [reply resources/js/pages/game/[game]/requests.tsx]: Yeah I went ahead and dropped it anyway, I kept the title one though

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
```diff
@@ -30,15 +30,18 @@ export const GameSetRequestsRoot: FC = () => {
...
-          t('This achievement set has been requested by the following {{val, number}} users:', {
-            val: totalCount,
-          })
-        ) : (
+      {allRequestors.length > 0 ? (
+        <Trans
+          i18nKey="gameRequestsCount"
+          count={totalCount}
+          values={{ val: totalCount }}
+          components={{ 1: <span className="font-bold" /> }}
```
wescopeland: nit: This line is unused and can be removed.

---

The `Trans` component's `components` prop is useful when we need to interpolate components into translations. For example:
```tsx
This achievement was unlocked <span className="font-bold">100</span> times
```
The "100" can appear anywhere in the sentence after it's translated, but we want to ensure it always remains bold in the UI. For example, in Korean, the number appears in a totally different position:
```tsx
이 업적이 <span className="font-bold">100</span>번 달성되었습니다
```
i18next gives us a powerful way to handle this requirement. In the translation string, we use numbered tags:
```json
"unlockCount_other": "This achievement was unlocked <1>{{val, number}}</1> times"
```
The `<1>` tag corresponds to the component at index 1 in the `components` prop. This allows the bold styling to follow the number regardless of where it appears in the translated sentence.

wescopeland (approved):
Thanks for working through my feedback!

resources/js/features/games/components/+requests/GameSetRequestsRoot.tsx
Chew: This does seem useful! Thanks for the heads up and explanation 😄 Updated the component params

lang/en_US.json
```diff
@@ -1060,6 +1060,11 @@
     "There are no achievements for this set yet.": "There are no achievements for this set yet.",
     "No achievements yet": "No achievements yet",
     "Set requests help developers decide what games to work on next.": "Set requests help developers decide what games to work on next.",
+    "All requestors ({{val, number}})": "All requestors ({{val, number}})",
```
Jamiras: This is no longer used.

Jamiras (approved):
Other than a leftover string, I think this is probably good-to-go.

lang/en_US.json
Chew: Good catch, thanks

PR #3984, opened 2025-09-29
feat: add app_version column to beta feedback submissions table

wescopeland:
Just a little bit of additional helpful context when running reports.


wescopeland:
It does. I want to track sentiment trend shifts on certain bits of functionality as releases occur. I'm doing this manually already, but it's a lot easier if we have this field mixed in the data.

PR #3991, opened 2025-10-01
don't double escape SEO metadata

Jamiras:
https://discord.com/channels/310192285306454017/991400303482122300/1421829187626467348
https://discord.com/channels/310192285306454017/453242743292952578/1421236306263933030


resources/js/common/components/SEO/SEO.tsx
Jamiras, line 1: This file reverts #3374, which doesn't seem to be necessary any more.

resources/views/components/game/devbox-claim-management.blade.php
```diff
@@ -84,7 +84,7 @@
 
 <script>
 function makeClaim() {
-    const gameTitle = "{!! html_entity_decode($gameTitle) !!}";
+    const gameTitle = "{!! addslashes($gameTitle) !!}";
```
Jamiras, line 87: Since `gameTitle` is no longer sanitized, we have to do an alternate escaping. In reality, we only need to escape double quotes here.

resources/views/pages-legacy/gameInfo.blade.php
```diff
@@ -323,10 +323,7 @@
 }
 
 sanitize_outputs(
-    $gameTitle,
-    $consoleName,
     $richPresenceData,
-    $user,
```
Jamiras, line 329: All of these (except `richPresenceData`) are passed to other widgets, which are responsible for their own escaping when necessary.

PR #3994, opened 2025-10-01
fix(hubs): add fallback for some event hub breadcrumbs

wescopeland:
http://localhost:64000/hub/18379

**Before**
Screenshot 2025-10-01 at 5 07 14 PM

**After**
Screenshot 2025-10-01 at 5 06 48 PM

### DIFF
--- a/app/Platform/Actions/BuildHubBreadcrumbsAction.php
+++ b/app/Platform/Actions/BuildHubBreadcrumbsAction.php
@@ -7,6 +7,7 @@ namespace App\Platform\Actions;
 use App\Models\GameSet;
 use App\Platform\Data\GameSetData;
 use App\Platform\Enums\GameSetType;
+use App\Support\Cache\CacheKey;
 use Carbon\Carbon;
 use Illuminate\Support\Facades\Cache;
 use Spatie\LaravelData\Lazy;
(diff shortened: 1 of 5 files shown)


app/Platform/Actions/BuildHubBreadcrumbsAction.php
```diff
@@ -447,6 +448,20 @@ private function findSameTypeParent(GameSet $gameSet, string $type, array $visit
...
+
+            // As a fallback, look for ANY parent hub that is an Events hub.
+            // This handles cases where naming is inconsistent (eg: "[RA Roulette 2022 - X]" with parent "[Events - RA Roulette (RAWR)]").
+            $anyEventParent = $gameSet->parents()
+                ->select('game_sets.*')
+                ->where('game_sets.type', GameSetType::Hub)
+                ->where('game_sets.title', 'like', '[Events - %')
+                ->whereNotIn('game_sets.id', $visited)
+                ->whereNull('game_sets.deleted_at')
+                ->first();
```
Jamiras, line 460: Should the events themselves have a similar breadcrumb?
http://localhost:64000/event/80-ra-roulette-2022

Jamiras (approved):

app/Platform/Actions/BuildHubBreadcrumbsAction.php
wescopeland: Yes. This is nontrivial enough that I'll address it in a follow-up PR (already have a stash prepared).

PR #3996, opened 2025-10-02
ci: add two custom ESLint rules

wescopeland:
This PR adds two custom ESLint rules to the project:
* `@retroachievements/enforce-typescript-in-app-code`
* `@retroachievements/no-cross-boundary-imports`

These both came up while reviewing #3983. These rules are mostly derived from custom rules I've built for several other projects I maintain - they're already fairly well battle-tested.

`@retroachievements/enforce-typescript-in-app-code` ensures *.js and *.jsx files are invalid in Inertia app code. Anything that's user-facing should be a strongly-typed TypeScript file.

`@retroachievements/no-cross-boundary-imports` automates some documentation that was previously in the project (which I've now removed as part of this branch):
* Feature modules should never import from other feature modules.
* Common/core/shared modules should never import from feature modules. NOTE: We don't have the core or shared modules yet. We eventually will (common will be split into these two) - this is just future-proofing.
* Core should never import from shared.

Screenshot 2025-10-02 at 8 30 33 AM

Screenshot 2025-10-02 at 8 31 10 AM

Screenshot 2025-10-02 at 8 31 32 AM


Jamiras (approved):
untested. seems reasonable.

PR #4000, opened 2025-10-04
feat(game): update authorization policy

wescopeland:
https://discord.com/channels/476211979464343552/1002689485005406249/1424151964127858890


app/Policies/GameCommentPolicy.php
```diff
@@ -75,6 +84,15 @@ public function update(User $user, GameComment $comment): bool
...
     public function delete(User $user, GameComment $comment): bool
     {
+        // Handle if comments are locked for this game.
+        if ($comment->commentable && $comment->commentable->comments_locked_at) {
+            // Only moderators and admins can delete when locked.
+            return $user->hasAnyRole([
+                Role::MODERATOR,
+                Role::ADMINISTRATOR,
+            ]);
+        }
```
Jamiras, line 94: This doesn't seem to be getting called on either the `/game1/` or `/game2/` pages. A user without any of the specified roles is still able to delete their own comments.

I can introduce an exception here and neither page is affected.

wescopeland: 👍  We should fix this in a follow-up. I'll go ahead and deploy the branch for now.

PR #4008, opened 2025-10-09
feat(game): show set rarity

wescopeland:
This PR adds a small rarity indicator to the set header on React game pages. This functionality has been requested numerous times in user beta feedback submissions.

**Before**
Screenshot 2025-10-09 at 7 37 41 PM

**After**
Screenshot 2025-10-09 at 7 37 30 PM

Additionally, tooltip contents have been updated:
Screenshot 2025-10-09 at 7 07 32 PM

Screenshot 2025-10-09 at 7 07 37 PM


wescopeland:
> I think I need to understand why people want this particular statistic to determine how prominent we should make it.

Looking over the latest feedback, my perception is users want this primarily as a quick difficulty gauge / effort estimation when deciding what to play. Sample feedback:
> I like seeing the retro ratio somewhere on the game page, it may be flawed but it still gives me a quick metric for "this is a damn hard set" or "everyone can finish this set easy"

Users are expressing frustration in their feedback at having to either:
1. Do mental math.
2. Navigate away to the list view.
3. Hover over tooltips.

I agree with your concerns about visual prominence - the white text contrast + gem icon might command too much attention, potentially more than warranted given the metric's inaccuracies.

How would you feel about any of these compromises?:
* **Option A.** We could remove the gem icon and keep the "xN.NN" format in its current location. This addresses the "commands attention" concern while still reducing the cognitive load that users are giving feedback about.
* **Option B.** We could move the rarity to a tooltip only. This means users will still need to hover (which some feedback has explicitly mentioned as friction).
* **Option C.** Go with Option A, but also use a darker shade of gray for the multiplier. This reduces visual prominence while keeping the number scannable.

Do you have a preference? I'd be happy to iterate on this implementation to find something we're both comfortable with.

Jamiras:
> Do you have a preference?

My preference would be to eliminate it entirely (or at least move it to the sidebar). There's a lengthy discussion around how inaccurate and misleading white points are in #1209.

I think if it stays, I'd definitely get rid of the gem. Having it there makes it feel like some sort of prize, rather than a rough gauge of difficulty. 

I haven't tried this in the page directly, but even just moving the white points out of the middle of the sentence feels less distracting as it feels like a separate element instead of something within the large element that's more important.

image
image
image
image
image

Given these mock-ups, I'd probably go with the all white text, no gem, end of sentence.

Jamiras:
No matter how you present it, the metric is flawed. Your sample feedback even says as much:

> it may be flawed but it still gives me a quick metric for "this is a damn hard set" or "everyone can finish this set easy"

I feel like this information is a more accurate gauge of difficulty / effort than RetroRatio:

image

The fact that only 22% of players finished the game let's me know how hard the game is. The time taken to master the game lets me know how much time I would have to invest to beat it.

The fact that only 2% of players mastered the game gives me a better feeling of how exceptional the challenge achievements are, and how much additional time I would have to invest to turn the beat into a mastery.

Regarding reducing cognitive load: There's an implicit relationship between time and completion percentage. The higher the time required, the more likely players will move on to another game before finishing it, which will lower the rate at which people reach the end of the game. Maybe we could have another stat that's something like "30% of players played this game for less than an hour".

When I look at something like this:
image

I'm much more apprehensive of the 105 achievements/800 points than the 7x rarity, because I know the rarity value is basically garbage.

Rarity is a very hand-wavy metric. If I asked you to tell me a game with a 4x rarity, would you even have a guess? But if I asked you to tell me a game that takes 10 hours to beat, you'd probably have a few ideas.

A much more meaningful metric on difficulty would be something like "50% of players earned 30% of the achievements/points in this game" or "10% of the achievements in this game were earned by less than 20% of players", but those are already kind of handled by the Achievement Distribution graph:
image

As RetroRatio is largely deteremined by number of unlocks, perhaps it could live somewhere near the achievement distribution graph?

Jamiras:
> Would making these two changes work for you?

I think so.

PR #4010, opened 2025-10-09
fix(game): remediate issue with subset meta description / Discord embeds

wescopeland:
This PR resolves a bug with game2 meta description that appears on Google and Discord embeds. Currently, it always points to the base set, even if it's a subset URL.

Now, the numbers are correct based on the target set.


resources/js/pages/game/[game].tsx
```diff
@@ -53,8 +54,8 @@ const GameShow: AppPage = () => {
     <>
       <SEO
         title={title as TranslatedString}
-        description={buildGameMetaDescription(game)}
-        ogImage={game!.badgeUrl}
+        description={buildGameMetaDescription(game, backingGame)}
+        ogImage={backingGame!.badgeUrl}
```
Comment (Jamiras) [inline resources/js/pages/game/[game].tsx:58]: The title is still for the base game (as is the page title). Do we want to change the meta title and/or the page title to include the subset name?

Comment (wescopeland) [reply resources/js/pages/game/[game].tsx:58]: Yes - fixed in latest.

Issue #4013, opened 2025-10-10
Some Firefox users forced to clear cache after every deployment


wescopeland:
**Describe the bug**
After each deployment, some Firefox users encounter site-wide JS errors and broken UI until they manually clear their browser cache.

Chrome/Edge users don't experience this issue.

I suspect there are numerous subtle root causes:
1. URLs are being reused for different content. We deploy to a static `/assets/build` path, so new deploys overwrite the old chunk URLs.
2. Old builds immediately 404. The static host serves assets via a single symlink to the current release, so all previous chunk URLs break after deploy.
3. Firefox caches aggressively. When Firefox caches a module graph and then encounters 404s or HTML responses for JS files (served from nginx), it keeps serving the broken state until the cache is cleared.
4. HTML fallback poisoning. The `@cache` fallback in nginx can return HTML for missing assets, which Firefox caches as "valid" JS.

To resolve this problem, we need to make asset URLs immutable and keep old static builds available. I believe we should:

1. Version the build directory, eg: `/assets/build/${VERSION}` before each build.
2. Persist builds in storage. We should probably hang on to the last 5 versions.
3. Flip the asset symlink direction. Make `public/assets` point to persistent storage (not the reverse).
4. Remove nginx HTML fallbacks. We should change `try_files $uri @cache =404` to `try_files $uri =404` for asset routes.

**To Reproduce**
1. Use Firefox.
2. Be unlucky.

PR #4018, opened 2025-10-12
build: add persistent versioning to compiled JavaScript assets

wescopeland:
Probably resolves https://github.com/RetroAchievements/RAWeb/issues/4013.

Also requires deployment script modifications, which are now present (commented out for now) in the stage file:
```bash
# ensure storage assets directory exists as a real directory (remove old symlink if present)
# rm -f "${BASEDIR}/storage/app/static/assets"
# mkdir -p "${BASEDIR}/storage/app/static/assets/build"

$FORGE_PHP artisan ra:storage:link --force

# ...

# version the Node.js build by setting APP_VERSION
# sed -i "s/APP_VERSION=.*/APP_VERSION=${VERSION}/g" "${BASEDIR}/.env"

pnpm build

# copy built Node.js assets to persistent storage
# mkdir -p "${BASEDIR}/storage/app/static/assets/build/${VERSION}"
# rsync -a "${BASEDIR}/${RELEASE_DIR}/public/assets/build/${VERSION}/" "${BASEDIR}/storage/app/static/assets/build/${VERSION}/"

# keep only the last 5 built asset versions
# cd "${BASEDIR}/storage/app/static/assets/build" && ls -1t | tail -n +6 | xargs -r rm -rf --
```

All this stuff is commented for now, and I've backed up the existing deploy script elsewhere. I'll uncomment before merging this branch.

I'm not actually sure if this will work without doing a stage deploy.

### DIFF
--- a/app/Helpers/database/player-game.php
+++ b/app/Helpers/database/player-game.php
@@ -210,6 +210,11 @@ function getUserProgress(User $user, array $gameIDs, int $numRecentAchievements 
 
         if ($numRecentAchievements === 0) {
             usort($lockedAchievements, function ($a, $b) {
+                if ($a['Achievement']['DisplayOrder'] === $b['Achievement']['DisplayOrder']) {
+                    //  DisplayOrders haven't been setup correctly; fallback to IDs for consistency
+                    return $a['Achievement']['ID'] <=> $b['Achievement']['ID'];
+                }
+
                 return $a['Achievement']['DisplayOrder'] <=> $b['Achievement']['DisplayOrder'];
             });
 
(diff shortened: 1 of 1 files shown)


config/filesystems.php
```diff
@@ -98,13 +98,13 @@
     */
 
     'links' => [
+        public_path('assets') => storage_path('app/static/assets'),
```
Jamiras, line 101: If I can decode the proposed changes here, it's trying to replace the `public/assets` folder with a symlink to the `app/static/assets`. `public/assets` is an existing folder in the tree, so it shouldn't be possible to replace it with a symlink.

.env.example
```diff
@@ -127,7 +127,7 @@ PUSHER_PORT=443
 PUSHER_SCHEME=https
 PUSHER_APP_CLUSTER=mt1
 
-VITE_BUILD_PATH=assets/build
+VITE_BUILD_PATH=assets/build/${APP_VERSION}
```
Jamiras, line 130: This change seems like it would be sufficient by itself. What is the other change trying to do?

wescopeland: Without the filesystems change, old versioned assets will 404 immediately after deploy, as they aren't being copied to persistent storage.

config/filesystems.php
wescopeland: You're right to flag this - the order of operations matters here. _public/assets_ currently exists as a real directory, so we can't just drop a symlink on top of it.

The deploy script sequence is:
1. Remove the old reverse symlink at _storage/app/static/assets_ if it exists.
2. Create _storage/app/static/assets/build_ as a real directory.
3. Run `storage:link --force`, which creates the new symlink: _public/assets_ -> _storage/app/static/assets_.
4. `pnpm build` the assets, then copy them into _storage/app/static/assets/build/${VERSION}/_.

The `--force` flag should handle replacing the existing _public/assets_ folder with the symlink. However, if that doesn't work, the deploy script may need to add `rm -rf "${BASEDIR}/public/assets"` before step 3.

PR #4027, opened 2025-10-16
convert login/login2 connect APIs to action

Jamiras:


public/login_app.php
```diff
@@ -1,10 +0,0 @@
-<?php
-
-// Auto login from app uses token. Standard login from app uses password.
-$user = request()->post('u');
-$pass = request()->post('p');
-$token = request()->post('t');
-
-$response = authenticateForConnect($user, $pass, $token);
```
Jamiras, line 8: This file was used by the DLL prior to version 0.79. As the minimum supported version is 1.x, and I didn't see any recent references to the URL in the server logs, I believe this is safe to remove.

app/Connect/Actions/LoginAction.php
```diff
@@ -0,0 +1,170 @@
...
+    private function authenticateFromPassword(): array
+    {
+        $user = User::whereName($this->username)->first();
+        if (!$user) {
+            return $this->invalidCredentials();
+        }
+
+        // don't let Banned or Spam users log in - treat as if the account was not found
+        $permissions = (int) $user->getAttribute('Permissions');
+        if ($permissions < Permissions::Unregistered) {
```
wescopeland: Suggestion: it may be worthwhile to also check `|| $user->isBanned()` here. At some point, we probably won't have these legacy `Permissions` around anymore, and while the check may redundant now, that would give this area of the code a clear cutover.

app/Connect/Actions/LoginAction.php
```diff
@@ -0,0 +1,170 @@
...
+    protected function initialize(Request $request): ?array
+    {
+        if (!$request->has(['u'])) {
+            return $this->missingParameters();
+        }
+        if (!$request->has(['p']) && !$request->has(['t'])) {
+            return $this->missingParameters();
+        }
+
+        $this->username = request()->input('u') ?? '';
```
wescopeland: nit: This `?? ''` seems unnecessary given the conditional on line 31.

app/Connect/Actions/LoginAction.php
Jamiras: Right now, they seem to be in sync, so I've switched over to just checking `isBanned()`.

```
> select count(*) from UserAccounts where permissions < 0 and banned_at is null;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
```

app/Connect/Actions/LoginAction.php
Jamiras: The class property is not nullable. `input()` returns a nullable string. I assumed the coersion was necessary to satisfy phpstan. It appears not. Removed.

PR #4035, opened 2025-10-18
fix(game): fetch leaderboards when loading the page from bfcache

### LINKED ISSUE #4030: Going back on your browser in some cases will make the leaderboard button not do anything (game2 beta page)
**Describe the bug**
If a game has leaderboards, such as [Super Mario Sunshine](https://retroachievements.org/game/6049), you can click the leaderboard button to see leaderboards. This button no longer works if you click on any of the following (that I could find) and you have to reload the page to make the leaderboards show up again when clicked.

For the below, this assumes you click these and press the browser's "back" button.

Where leaderboards no longer show up:
- Achievements
- Memory
- Tickets
- Profiles

Where leaderboards do keep showing up:
- Hubs
- Related Games
- Supported Game Files
- Official Forum Topic

I have a feeling this is due to how these pages are linked. Perhaps in some cases it doesn't re-render the page and some it does? 

**To Reproduce**
1. Go to a set with leaderboards
2. Go to any of the "no longer show up" items
3. Press the browser's back button

**Expected behavior**
I expected leaderboards to still show up.

- OS: macOS

List of tested browsers:
- Arc (Chromium), 1.114.1 (140.0.7339.214)
- Firefox, 144.0 (aarch64)

**Additional context**

wescopeland:
Resolves #4030.

To easily reproduce the issue:
1. On `master`, navigate to http://localhost:64000/game2/6049.
2. Click on the first achievement on the list to open its achievement details page.
3. Click the back button on your browser to return to the game page.
4. Click the leaderboards tab.
5. Observe that leaderboards fail to appear.

**Root Cause**
Leaderboards are a deferred prop to keep the page size down. They lazily-load client-side when the browser is idle. However, Inertia appears to have a bug where if the page loads from the browser's back-forward cache, deferred prop loading never fires. This is precisely what happens when going back from a Blade page to an Inertia page.

To work around this bug, we now have a custom hook to detect if the page is loaded from the bfcache and if leaderboards need to be lazy-loaded. If so, we assume Inertia will fail to load them and we optimistically fetch them instead.


Jamiras:
Not sure if it's caused by this PR, but:
1. Go to http://localhost:64000/game/10247
2. Click on Leaderboards
3. Click on Achievements
4. Click on Game Details
5. Back

page is blank.
image

Jamiras (approved):
tested the specific steps provided (on a different set), and did some spot checking on other sets/paths.

PR #4036, opened 2025-10-18
add game-checklist page

Jamiras:
Primarily intended to provide an easy way for the events team to validate users have beaten games:

http://localhost:64000/user/Jamiras/game-checklist?list=1

But can be expanded using the same syntax as the achievement checklist for more complex validation. Here's a page for the currently ongoing Halloween event: http://localhost:64000/user/Jamiras/game-checklist?list=Primary%3A2374%2C8368%2C16274%2C3541%2C13440%2C13367%2C19128%2C11656%2C1314%2C837%2C6130%2C1317%2C7118%2C19259%2C19111%2C7896%2C3007%2C4325%2C3504%2C14507%2C1275%2C18261%2C4667%2C19266%2C11252%2C8492%7CWildcard%3A11265%2C8742%2C11277%2C3529%2C8307


resources/js/common/components/GamesListItem/GamesListItem.tsx
Jamiras, line 1: This needs a great amount of styling work. The columns are all over the place and the beaten indicator labels should be on a single line. I was hoping to get some assistance for this.

wescopeland: Will do a full review tomorrow morning. In the meantime, I'm wondering if this diff gets you closer to where you want: https://github.com/wescopeland/RAWeb/commit/18f87ddcb4994722e3c851289ead955d112ddcad

Jamiras: Yes, thank you. I'd still like to get the second column centered vertically. And Im open to any further changes you think would be beneficial.

wescopeland: Looks like you may have missed a change on line 25, this should do it:
```tsx
<div className="flex flex-col justify-center sm:col-span-2">
```

resources/js/pages/user/[user]/game-checklist.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  return (
+    <>
+      <AppLayout.Main>
+        <div>
+          <UserBreadcrumbs t_currentPageLabel={t('Game Completion Checklist')} user={player} />
+          <UserHeading user={player}>{t('Game Completion Checklist')}</UserHeading>
+
+          <div className="form-grid-container mb-4">
+            <label>{t('Examine another user:')}</label>
```
Comment (wescopeland) [inline resources/js/pages/user/[user]/game-checklist.tsx]: a11y: I didn't catch this in _achievement-checklist.tsx_ either. This label is currently orphaned (browsers aren't natively associating it with the select control).

I think we have two options:
* Easy: change it from a `label` to a `span`. This retains the current behavior.
* Better, but slightly more difficult: have `BaseSelectAsync` accept an `id` prop so we can connect the two elements. If you'd like to do this:

1. In _BaseSelectAsync.tsx_, add `id` to the interface around line 58:
```tsx
/** ID for the trigger button for accessibility. */
id?: string;
```

2. Accept it in the function signature around line 70:
```tsx
value,
onChange,
id, // <----- added this line
```

3. Apply it to the `BaseButton` around line 113:
```tsx
<BaseButton
  id={id}
  variant="outline"
```

4. Finally, in _game-checklist.tsx_, update the label and `BaseSelectAsync` with new props to connect them to each other:
```tsx
<label htmlFor="user-select">...</label>

<BaseSelectAsync<App.Data.User>
  id="user-select"
  // ...
/>
```

resources/js/pages/user/[user]/game-checklist.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+            <BaseSelectAsync<App.Data.User>
```
Comment (wescopeland) [inline resources/js/pages/user/[user]/game-checklist.tsx:42]: Wondering if I may be doing something wrong. When I select another user, nothing seems to happen on the page. Do you see the same behavior? For me, I'm just searching for "Scott" and clicking on him when he appears in the user search results.

app/Community/Controllers/UserGameChecklistController.php
```diff
@@ -0,0 +1,33 @@
...
+use App\Data\UserData;
+use App\Http\Controller;
+use App\Models\User;
+use Illuminate\Http\Request;
+use Inertia\Inertia;
+use Inertia\Response as InertiaResponse;
+
+class UserGameChecklistController extends Controller
+{
+    public function index(Request $request, User $user): InertiaResponse
```
wescopeland: nit: This probably should be `show` because it's focused on one user rather than an index of users.

resources/js/common/components/GamesListItem/GamesListItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import { PlayerGameProgressBar } from '@/common/components/PlayerGameProgressBar';
+import { formatDate } from '@/common/utils/l10n/formatDate';
+
+import { AwardIndicator } from '../AwardIndicator';
+import { GameAvatar } from '../GameAvatar';
+
+interface GamesListItemProps {
+  game: App.Platform.Data.Game;
+  index: number;
```
wescopeland: nit: Safe to remove this unused prop.

resources/js/pages/user/[user]/game-checklist.tsx
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  return (
+    <>
+      <AppLayout.Main>
+        <div>
+          <UserBreadcrumbs t_currentPageLabel={t('Game Completion Checklist')} user={player} />
+          <UserHeading user={player}>{t('Game Completion Checklist')}</UserHeading>
+
+          <div className="form-grid-container mb-4">
```
Comment (wescopeland) [inline resources/js/pages/user/[user]/game-checklist.tsx]: Minor CSS suggestion (also applies to _achievement-list.tsx_):
```tsx
<div className="mb-4 flex items-center gap-2">
```
This centers the label and field and adds just a bit of whitespace between them.

app/Community/Controllers/UserGameChecklistController.php
Jamiras: changed.

resources/js/pages/user/[user]/game-checklist.tsx
Comment (Jamiras) [reply resources/js/pages/user/[user]/game-checklist.tsx:42]: I admit to not actually testing this as I generally keep meilisearch disabled to free resources.

The console is saying:
```
Uncaught Error: Ziggy error: route 'user.game-checklist' is not in the route list.
```

But the route exists:
```
$ sail artisan route:list | grep checklist
  GET|HEAD  user/{user}/achievement-checklist user.achievement-checklist › Ap…
  GET|HEAD  user/{user}/game-checklist user.game-checklist › App\Community\Co…
```

The route seems to work if I use it in the blade menu. I'm not sure why it's not working here.

resources/js/common/components/GamesListItem/GamesListItem.tsx
Jamiras: done.

resources/js/pages/user/[user]/game-checklist.tsx
Comment (Jamiras) [reply resources/js/pages/user/[user]/game-checklist.tsx]: done.

resources/js/pages/user/[user]/game-checklist.tsx
Comment (Jamiras) [reply resources/js/pages/user/[user]/game-checklist.tsx]: done. thank you for the walkthrough.

resources/js/pages/user/[user]/game-checklist.tsx
Comment (wescopeland) [reply resources/js/pages/user/[user]/game-checklist.tsx:42]: Ziggy maintains a [route manifest](https://github.com/RetroAchievements/RAWeb/blob/master/resources/js/ziggy.js) used by the React app (otherwise the React app has no idea what routes exist). However, this route manifest needs to be updated when new routes are added to the front-end.

`composer types` will magically refresh the routes manifest.

`localhost` references being present in the route manifest are expected. In the build script on the server, we run the command to regenerate the manifest on every deploy.

Comment (Jamiras) [reply resources/js/pages/user/[user]/game-checklist.tsx:42]: I'm not sure how the output order is determined, but when I run `composer types` (and `php artisan typescript:generate`), I end up with large diffs for small changes because things get reordered. Should I commit the full diffs, or continue to try to surgically extract the important changes?

Comment (wescopeland) [reply resources/js/pages/user/[user]/game-checklist.tsx:42]: I'm not sure either. The full diff doesn't bother me, no worries.

wescopeland (approved):

PR #4041, opened 2025-10-19
prevent exception passing blank username

Jamiras:
https://github.com/RetroAchievements/RAWeb/pull/4027#issuecomment-3419683460


wescopeland (approved):
Verified locally.

PR #4048, opened 2025-10-21
add GetBadgeIdRangeAction

Jamiras:


app/Helpers/util/media.php
```diff
@@ -112,17 +113,28 @@ function UploadBadgeImage(array $file): string
...
+        $badgeIterator = str_pad((string) $badgeRange['NextBadge'], 5, '0', STR_PAD_LEFT);
+
+        $imagePath = 'Badge/' . $badgeIterator . '.png';
+        $imagePathLocked = 'Badge/' . $badgeIterator . '_lock.png';
+
+        $localImagePath = storage_path('app/media/' . $imagePath);
+        $localImagePathLocked = storage_path('app/media/' . $imagePathLocked);
+        if (!file_exists($imagePath)) {
+            break;
+        }
```
Jamiras, line 127: Uploading a badge is a separate API call from updating the achievement, so `GetBadgeIdRangeAction` will not immediately detect the new badge. The client will use the response from the upload badge call to associate the new badge to the achievement and then upload the achievement changes separately.

This double checks against the file system to ensure images that have been uploaded but aren't yet associated to achievements don't get replaced. It also serves as a check against a potential race condition where multiple uploads occur at the same time.

app/Support/Media/FilenameIterator.php
Jamiras, line 1: `BadgeIter.txt` is no longer needed after this PR.

app/Connect/Actions/GetBadgeIdRangeAction.php
```diff
@@ -0,0 +1,48 @@
...
+
+    protected function process(): array
+    {
+        $data = Achievement::query()
+            ->select([
+                DB::raw("MIN(CAST(BadgeName AS UNSIGNED)) AS first_badge"),
+                DB::raw("MAX(CAST(BadgeName AS UNSIGNED)) AS last_badge"),
+            ])
+            ->where('BadgeName', '>', '00002')
+            ->first();
```
Jamiras, line 32: This could be cached, which would also allow the upload functionality to reserve values.

It could also be stored in the StaticData table, but that has a comment specifically suggesting that this sort of data should be queried every time.

app/Helpers/util/media.php
wescopeland: Understood, but I'm not 100% sure this is working as intended based on my local simulation of the logic:
Screenshot 2025-10-21 at 6 10 49 PM

When badge 99999 already exists on disk, `file_exists($imagePath)` returns false because it seems to be checking a relative path from the project root where no _Badge/_ directory exists.

Assuming I didn't mess my simulation in Tinkerwell up, I think this means the protection against overwriting recently-uploaded badges that aren't yet in the DB doesn't work. The loop will always break immediately and use whatever the database returns as `MAX + 1`.

I'm wondering if we should use `file_exists($localImagePath)` in the conditional instead.

Jamiras: > I'm wondering if we should use `file_exists($localImagePath)` in the conditional instead.

Yep. I must have missed that when I renamed stuff.

Jamiras: FWIW, I tested this using RALibretro. I changed a badge to ensure I had a local file, copied the local file to fill the next ID, then changed the badge from the client again, resulting in a third local file. I just don't remember what exact state the code was in when I did the testing :(

Updated/fixed.

wescopeland: No worries at all. I wanted to bounce this off you because truthfully I wasn't 100% confident in what I was seeing in my local simulation :)

wescopeland (approved):

PR #4051, opened 2025-10-22
fix(requests): update user sort order

wescopeland:
https://discord.com/channels/476211979464343552/1026595325038833725/1430651222062272594

cc @Chew


Chew:
Tested locally and this works for me too!

Issue #4026, opened 2025-10-16
Inconsistency with order of locked achievements on profile page


GenericMadScientist:
**Describe the bug**
The locked achievements on a user's profile page can deviate from the intended order, and can be inconsistent between users.

**To Reproduce**
(As of writing)
1. Go to my profile, https://retroachievements.org/user/GenericMadScientist
2. Look at the achievements for Pokémon Diamond Version [Subset: Professor Oak Challenge] and observe the Coal Badge achievement is after the achievement for Floatzel.
3. Go to ExponiK's profile showing 'more...' games, https://retroachievements.org/user/ExponiK?g=15
4. Look at the achievements for Pokémon Diamond Version [Subset: Professor Oak Challenge] and observe the Coal Badge achievement is before the achievement for Floatzel.

**Expected behavior**
The Coal Badge achievement on my profile should show in the same place it does for ExponiK.

**Screenshots**
First screenshot is from my profile at time of writing, second is from ExponiK's.

<img width="868" height="311" alt="Image" src="https://github.com/user-attachments/assets/0ce82bfa-b02f-409f-a9bc-334fa7787a49" />

<img width="880" height="283" alt="Image" src="https://github.com/user-attachments/assets/d11e3372-6e80-4c71-80f6-8fa2779d238a" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Firefox
 - Version 143.0.4

Also checked on
 - Browser Ungoogled Chromium
 - Version 141.0.7390.54

**Additional context**
The order on my profile was initially correct, it was only some time after unlocking the Magikarp and Staravia achievements that I noticed the problem.

Soulsderfuchs:
I have noticed this happen too, as an example. [](url)

<img width="990" height="409" alt="Image" src="https://github.com/user-attachments/assets/b4320c86-6e8c-466e-a955-1b563623fa06" />

<img width="872" height="254" alt="Image" src="https://github.com/user-attachments/assets/18601658-0944-4e5a-b7ff-eea911de35b8" />

In my profile the achivement order shows "Ramblings of a Mad Man" as the last achievement but the intented order should be "Speaking with Ghosts"

PR #4056, opened 2025-10-23
fix(user): remediate inconsistency with locked achievement order on profile

ScottFromDerby:
on profile page (#4026) whitespace fix

If DisplayOrder hasn't been set properly, the rendered order falls to whatever order php has decided to shuffle the array in. Falling back to ID should provide consistency

### DIFF
--- a/resources/js/pages/user/[user]/achievement-checklist.tsx
+++ b/resources/js/pages/user/[user]/achievement-checklist.tsx
@@ -14,11 +14,11 @@ import type { AppPage } from '@/common/models';
 import { EventAchievementSection } from '@/features/events/components/EventAchievementSetContainer/EventAchievementSet/EventAchievementSection';
 
 const UserAchievementChecklist: AppPage = () => {
-  const { player, groups } = usePageProps<App.Community.Data.AchievementChecklistPageProps>();
+  const { player, groups, auth } = usePageProps<App.Community.Data.AchievementChecklistPageProps>();
 
   const { t } = useTranslation();
 
-  const query = useUserSearchQuery({ initialSearchTerm: player.displayName });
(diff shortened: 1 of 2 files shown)


app/Helpers/database/player-game.php
```diff
@@ -210,7 +210,12 @@ function getUserProgress(User $user, array $gameIDs, int $numRecentAchievements
 
         if ($numRecentAchievements === 0) {
             usort($lockedAchievements, function ($a, $b) {
-                return $a['Achievement']['DisplayOrder'] <=> $b['Achievement']['DisplayOrder'];
+                if ($a['Achievement']['DisplayOrder'] == $b['Achievement']['DisplayOrder']) {
+                    //  DisplayOrders haven't been setup correctly; fallback to IDs for consistency
+                    return $a['Achievement']['ID'] <=> $b['Achievement']['ID'];
+                } else {
```
wescopeland: nit: Since both branches of the conditional return, I recommend removing this `else`:
```php
if (...) {
  return ...;
}

return ...;
```

app/Helpers/database/player-game.php
```diff
@@ -210,7 +210,12 @@ function getUserProgress(User $user, array $gameIDs, int $numRecentAchievements
 
         if ($numRecentAchievements === 0) {
             usort($lockedAchievements, function ($a, $b) {
-                return $a['Achievement']['DisplayOrder'] <=> $b['Achievement']['DisplayOrder'];
+                if ($a['Achievement']['DisplayOrder'] == $b['Achievement']['DisplayOrder']) {
```
wescopeland: nit: Recommend changing the `==` to `===`. Since both fields are the same type, it should be safe to rely on a strict comparison.

wescopeland (approved):
Thanks!

PR #4057, opened 2025-10-23
fix(game): adjust hubs 'See More' button logic

wescopeland:
Makes a slight tweak to the game metadata's "See more" button for hubs.

Now, the 'See more' section is auto-expanded if there are only 1 or 2 rows of 'See more' content as opposed to just one.

The 'See more' button will only appear if there are at least 3 rows to put behind the button click.


wescopeland:
> Is there even any reason to keep the Show More?

I think so, start reading here: https://discord.com/channels/310192285306454017/888607146290343976/1430936494813614100

We may make one final tweak to the metadata/hubs display.

PR #4060, opened 2025-10-23
feat(auth): block disposable email services

wescopeland:
**Before deployment, see https://github.com/RetroAchievements/RAWeb/pull/4060#discussion_r2463019143.**

---

Pulls in [erag/laravel-disposable-email](https://github.com/eramitgupta/laravel-disposable-email) to block disposable email services being used for account registration. These are most commonly used for spam and abuse.

To test this, try creating an account with an email address like `me@10minutemail.com`.


app/Community/Requests/UpdateEmailRequest.php
```diff
@@ -17,7 +17,7 @@ public function authorize(): bool
     public function rules(): array
     {
         return [
-            'newEmail' => 'required|email',
+            'newEmail' => 'required|email|disposable_email',
```
Jamiras: Would you be opposed to changing this to `not_disposable_email`? It reads like the email must be disposable to be allowed.

app/Community/Requests/UpdateEmailRequest.php
Jamiras, line 1: Creating a new account gives a verbose error message explicitly saying the email provider is not allowed, but trying to change the email address to a disposable address just generates a "something went wrong" error.

config/disposable-email.php
```diff
@@ -0,0 +1,50 @@
...
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Local Domain Blacklist File
+    |--------------------------------------------------------------------------
+    |
+    | This file will store the list of disposable email domains. You can update
+    | it manually or via the `erag:sync-disposable-email-list` artisan command.
```
Jamiras, line 11: Don't forget to mark this PR as deployment/command. Took me a bit to figure out I needed to do this.

wescopeland: We should probably bake this into the deployment script itself.

app/Community/Requests/UpdateEmailRequest.php
wescopeland: I agree 100%. `disposable_email` is what's exported by the library itself. The only idea I have is to make an alias for the rule name, which I've done in latest.

app/Community/Requests/UpdateEmailRequest.php
wescopeland: Fixed in latest.

PR #4061, opened 2025-10-23
feat: add MetaKeySubmitTooltip component

wescopeland:
Very small QoL improvement to React social forms. We support the Meta+Enter hotkey to submit, but it's not discoverable. This simply adds a tooltip. The tooltip doesn't instantly appear, and it's not loud. It also doesn't appear for users on Android devices.

Screenshot 2025-10-23 at 6 05 29 PM

```tsx
// before
<BaseButton>Submit</BaseButton>

// after
<MetaKeySubmitTooltip>
  <BaseButton>Submit</BaseButton>
</MetaKeySubmitTooltip>
```


resources/js/common/components/MetaKeySubmitTooltip/MetaKeySubmitTooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+  if (!metaKey) {
+    return children;
+  }
+
+  return (
+    <BaseTooltip {...rest}>
+      <BaseTooltipTrigger asChild>{children}</BaseTooltipTrigger>
+
+      {/* Doesn't need to be localized. The 'Enter' key is universal. */}
+      <BaseTooltipContent>{`${metaKey} + Enter`}</BaseTooltipContent>
```
Jamiras: Recommend including the "Submit" verb in the tooltip.

Seeing a tooltip that just says "Ctrl + Enter" doesn't tell me why that tooltip is there. 

Image

vs

Image

Gives me a much clearer indication that Ctrl+Enter is helper text, not a description of what the button does (as would be typical for a tooltip).

And if Android doesn't have a meta key: 
```
        $metaKey = match ($userOS) {
            UserOS::MacOS, UserOS::IOS => '⌘',
            UserOS::Android => '',
            default => 'Ctrl',
        };
```
And only "Enter" is shown, it becomes a verb in the tooltip, rather than a keystroke.

Jamiras: Also recommend removing the spaces on either side of the plus. "Ctrl+Enter" ties the chord together better than "Ctrl + Enter".

Google search AI says:
> Most style guides recommend omitting extra spaces around the plus sign for keyboard shortcuts like Ctrl+S. The compact format is cleaner and used consistently by major developers like Microsoft and Apple.

wescopeland: Android won't show a tooltip at all:
```ts
if (!metaKey) {
  return children;
}
```

Updated in latest:
```tsx
<BaseTooltipContent>
  {t('Submit: {{metaKey}}+Enter', {
    metaKey,
    nsSeparator: null,
  })}
</BaseTooltipContent>
```

wescopeland: Changed to "Submit: MetaKey+Enter" in latest.

Jamiras: More feedback from Google search AI:
> Parentheses are typically used to explicitly state a standard keyboard shortcut, like Ctrl + C for copy. This method is common in menus and user guides to display the shortcut next to its associated command, but not within the interactive label itself. 
> * **Usage in menus:** Copy (Ctrl+C)
> * **Best practice:** When a shortcut is listed with a menu item or control, placing it in parentheses is a clear way to distinguish it from the command name.
> * **Example:** In a web or desktop application, a "Save" menu option might display as Save (Ctrl+S). 

Recommend using parentheses instead of colon.

wescopeland: Claude agrees with Gemini (search AI):
> Yes, **both recommendations are correct**:
>
> **No spaces around `+`**: "Ctrl+Enter" is the standard format across all major platforms (Windows, macOS, Linux). The compact form is cleaner and universally recognized.
> 
> **Parentheses over colon**: "Submit (MetaKey+Enter)" follows standard UI convention. The colon format can be ambiguous and isn't commonly used in professional interfaces.
> 
> Both changes improve consistency with industry standards and reduce cognitive load.

Updated in latest.

PR #4065, opened 2025-10-24
fix(game-list): disable 'Surprise me' button always when client-side fetch is running

wescopeland:
This PR makes a minor adjustment to the 'Surprise me' button on game list pages:
Screenshot 2025-10-24 at 7 18 28 PM

We have a recurring error that appears in the server logs:
Screenshot 2025-10-24 at 7 19 00 PM

This is caused by a race condition where a user has a slow network connection, changes their filter options, then hovers over the 'Surprise me' button which triggers a prefetch that has 0 results.

Now, because the button is disabled while the client-side fetch is running, the race condition can no longer occur.

To see the changes, just start changing your filter options while viewing a game list. You should see the button auto-disable while the client-side API call is running.

### DIFF
--- a/app/Community/Actions/GenerateAnnualRecapAction.php
+++ b/app/Community/Actions/GenerateAnnualRecapAction.php
@@ -57,11 +57,14 @@ class GenerateAnnualRecapAction
 
         $this->extractDevelopmentTime($recapData, $user, $gameData, $startDate, $endDate);
 
+        $subsetGameIds = $this->identifyAndMergeSubSets($gameData);
+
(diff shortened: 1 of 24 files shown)


resources/js/features/game-list/components/GamesDataTableContainer/DataTableToolbar/DataTableToolbar.test.tsx
wescopeland, line 1: Unrelated to this change, I added some test coverage here.

PR #4066, opened 2025-10-25
feat(middleware): convert POST requests from Cloudflare challenges to GET

wescopeland:
Probably resolves https://discord.com/channels/476211979464343552/1002693810037477377/1431401553629155480.

Based on the .har file shared in this discussion, we can see how Cloudflare managed challenge redirects are hitting the server and causing the "405 Method Not Allowed" error.

Cloudflare seems to have made a change on their end to send these redirects as POST requests rather than GET requests. [It's almost certainly a bug on their end](https://community.cloudflare.com/t/js-challenge-changes-get-request-to-post/300387/15) somewhere.

Either way, this band-aids the issue for us. We now detect incoming POST redirects from Cloudflare and invisibly convert them to GET calls via a new middleware.

I have no way of testing this beyond writing a test suite to actually verify the code does what it's supposed to. The best way to test it will be to deploy to stage and make sure the "Method not allowed" error stops appearing.

### DIFF
--- a/app/Connect/Controllers/ConnectApiController.php
+++ b/app/Connect/Controllers/ConnectApiController.php
@@ -5,14 +5,10 @@ declare(strict_types=1);
 namespace App\Connect\Controllers;
 
 use App\Connect\Concerns\AchievementRequests;
-use App\Connect\Concerns\AuthRequests;
 use App\Connect\Concerns\BootstrapRequests;
-use App\Connect\Concerns\DevelopmentRequests;
-use App\Connect\Concerns\GameRequests;
 use App\Connect\Concerns\HeartbeatRequests;
 use App\Connect\Concerns\LeaderboardRequests;
 use App\Connect\Concerns\LegacyCompatProxyRequests;
-use App\Connect\Concerns\TicketRequests;
 use App\Http\Controller;
 use Exception;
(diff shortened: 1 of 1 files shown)


app/Http/Middleware/HandleCloudflareChallenge.php
```diff
@@ -0,0 +1,66 @@
...
+     * Cloudflare's managed challenges submit as POST requests with challenge tokens.
+     * Laravel routes expect GET requests for page views, causing users to see 405 errors.
+     * This middleware detects Cloudflare challenge completions and converts them to GET requests.
+     */
+    public function handle(Request $request, Closure $next): Response
+    {
+        if ($request->isMethod('POST') && $this->isCloudflareChallenge($request)) {
+            // Convert the POST request to GET and clear POST data.
+            $request->setMethod('GET');
+            $request->request->replace([]);
```
Jamiras, line 23: What happens to a challenged GET request with parameters? If the user has a subset page bookmarked, there's a `?set=X` GET parameter. Wouldn't this clear that too?

wescopeland: It would - this is broken. I've made an update in latest & added another test case to verify.

Jamiras (approved):

PR #4074, opened 2025-10-27
don't create new sessions if backdated timestamp predates newest session start date

Jamiras:
Fixes an issue where a backdated unlock or leaderboard submission might create a new session record instead of updating an existing one.

The problem is the session recycling logic is only looking at the latest session, so if a newer session was created before the backdated requests get processed, the logic will see that the timestamp is more than 10 minutes away from the latest session being updated and create a new session for the backdated logic.

Additionally, as the new session for the backdated timestamp has a created_at time of when the session was created, the session has an unintentional negative duration (rich_presence_updated_at - created_at is negative). This messes with the playtime stats, particularly if the backdating period is large.

There are ~28k records in the player_sessions table where the rich_presence_updated_at is more than 10 minutes prior to the created_at timestamp. Of those ~7k records have durations > 1m. I looked at a handful of these, and don't feel like the effort to fix this data is worth it.

### DIFF
--- a/app/Http/Actions/DetectUserOSAction.php
+++ b/app/Http/Actions/DetectUserOSAction.php
@@ -11,20 +11,22 @@ class DetectUserOSAction
 {
     public function execute(): ?UserOS
     {
-        $agent = new Agent();
+        return once(function () {
+            $agent = new Agent();
 
-        if ($agent->is('Windows') || $agent->is('Windows NT')) {
-            return UserOS::Windows;
-        } elseif ($agent->is('OS X') && !$agent->isMobile()) {
-            return UserOS::MacOS;
-        } elseif ($agent->is('AndroidOS')) {
-            return UserOS::Android;
-        } elseif ($agent->is('Debian') || $agent->is('Ubuntu') || $agent->is('OpenBSD') || $agent->is('Linux')) {
-            return UserOS::Linux;
-        } elseif ($agent->is('iOS')) {
-            return UserOS::IOS;
-        }
+            if ($agent->is('Windows') || $agent->is('Windows NT')) {
+                return UserOS::Windows;
+            } elseif ($agent->is('OS X') && !$agent->isMobile()) {
+                return UserOS::MacOS;
+            } elseif ($agent->is('AndroidOS')) {
+                return UserOS::Android;
+            } elseif ($agent->is('Debian') || $agent->is('Ubuntu') || $agent->is('OpenBSD') || $agent->is('Linux')) {
+                return UserOS::Linux;
+            } elseif ($agent->is('iOS')) {
+                return UserOS::IOS;
+            }
 
-        return null;
+            return null;
+        });
     }
 }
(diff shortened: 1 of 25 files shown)


tests/Feature/Connect/StartSessionTest.php
```diff
@@ -730,7 +730,7 @@ public function testStartSessionDelegatedByName(): void
             'game_id' => $bonusGameOne->id,
         ])->first();
         $this->assertModelExists($playerSession);
-        $this->assertEquals(1, $playerSession->duration);
+        $this->assertEquals(30, $playerSession->duration);
```
Jamiras, line 733: This is the result of how the unlocks are populated above. Because all of the unlocks are considered backdated, the first session gets extended for the other backdated unlocks, instead of creating separate sessions for each of them.

app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -29,34 +29,52 @@ public function execute(
...
-            })
-            ->where(function ($query) use ($userAgent) {
-                if ($userAgent) {
-                    $query->where('user_agent', $userAgent)
-                        ->orWhereNull('user_agent');
-                }
-            })
-            ->orderByDesc('id')
-            ->first();
+        // if a historical session was found, adjust the duration so it includes the timestamp and return it
```
wescopeland: nit: recommend changing "if a historical session was found" to "if a historical session **_is_** found". Unless I'm misunderstanding, when we arrive to this comment, we won't have found anything yet - we're about to search for it.

app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -29,34 +29,52 @@ public function execute(
         ?string $userAgent = null,
         ?string $ipAddress = null,
     ): PlayerSession {
+        $sessionKeepAliveTimeInMinutes = 10;
+        $isBackdated = ($timestamp && $timestamp->diffInMinutes(Carbon::now(), true) > $sessionKeepAliveTimeInMinutes);
+
         $timestamp ??= Carbon::now();
```
wescopeland, line 35: I wonder if it's worth having some kind of validation here for a maximum backdate timespan threshold. At the moment, I think we accept timestamps from days, weeks, or even months ago.

Maybe we should limit backdating to some arbitrary threshold (72 hours?), and if it exceeds the threshold, set the timestamp to `now()` and `$isBackdated` to false.

app/Platform/Actions/ResumePlayerSessionAction.php
Jamiras: This comment actually doesn't apply here at all. It's from a previous iteration of the implementation.

I had some short-circuiting code for backdated logic that got removed because some of the tests were still expecting other updates to the backdated sessions.

app/Platform/Actions/ResumePlayerSessionAction.php
Jamiras: It's 14 days:
https://github.com/RetroAchievements/RAWeb/blob/54cacec6bec6f5011385cf71c78039f51e26680f/public/dorequest.php#L306-L308

Putting a fixed upper limit in this function will break unit tests which have hard-coded "now" values that are already more than 14 days in the past.

PR #4076, opened 2025-10-28
feat: convert redirect page to inertia

Chew:
Very basic for now. Doesn't change anything, but open to ideas for improvements.

Only notable change is instead of a `<a href>` it uses `window.location.replace`, which will make pressing "Back" on the browser go to the original page instead of this redirect page.

Also, of course, a test and i18n keys.

Redirect's route was moved into an Inertia middleware so it could work, it could be moved elsewhere if needed.

Preview, if relevant.

CleanShot 2025-10-27 at 22 03 32@2x


resources/js/pages/redirect.tsx
wescopeland, line 1: Security, major: We definitely want to make this `noindex` and probably `nofollow`. Without it, the page is subject to an [SEO poisoning attack](https://www.vectra.ai/topics/seo-poisoning).

Someone on the forum could plant a link that ultimately causes Google and/or LLMs to learn about `https://retroachievements.org/redirect?url=https://evil-link.com`. This would appear in search results with authority from our trusted domain, but ultimately lands a user somewhere malicious. The end result of this attack is Google either removing a lot of our SEO juice, or removing us from search results entirely.

wescopeland (commented):
Will do an exhaustive review either later this evening or tomorrow once checks are passing, but wanted to shoot along one quick comment.

resources/js/pages/redirect.tsx
Chew: This was on the original page, but is there a way to trigger that with the <SEO> somehow? I think this was the only page using noindex.

wescopeland: For some reason I thought that component already supported both `noindex` and `nofollow` props. We should add them 😅

I think what we'll want is two optional boolean props called `noindex` and `nofollow`.

If only `noindex` is truthy, this meta tag should be added:
```tsx
<meta name="robots" content="noindex" />
```

If only `nofollow` is truthy, this meta tag should be added:
```tsx
<meta name="robots" content="nofollow" />
```

If both props are truthy, this **single** meta tag should be added:
```tsx
<meta name="robots" content="noindex,nofollow" />
```

It'll also be important to cover these scenarios in the SEO component's test suite.

resources/js/common/components/SEO/SEO.tsx
```diff
@@ -128,6 +128,16 @@ interface SEOProps {
...
 
+  /**
+   * Adds "noindex" flag for robots
+   */
+  noIndex?: boolean;
+
+  /**
+   * Adds "nofollow" flag for robots
+   */
+  noFollow?: boolean;
```
wescopeland: DX: Let's go with the direct HTML attribute names for the props: `noindex` and `nofollow`.

resources/js/common/components/SEO/SEO.test.tsx
wescopeland, line 1: Testing: Let's cover this exhaustively, as mistakes will be costly. I think we need four test cases:
* When only `noindex` is used.
* When only `nofollow` is used.
* When both are used.
* When neither is used.

wescopeland:
Still looks like we're having some trouble with the PHPUnit test suite: https://github.com/RetroAchievements/RAWeb/actions/runs/19050202919/job/54408281395?pr=4076

Want me to try to poke around?

wescopeland:
They all pass for me too. I wonder if this might fix it, would you mind trying this solution?
```php
// config/inertia.php

// line 50
            resource_path('js/pages'),
```
"Pages" -> "pages".
I'm wondering if it's a file system / sensitivity issue with CI specifically.

Jamiras:
> They all pass for me too. I wonder if this might fix it, would you mind trying this solution?
> 
> ```
> // config/inertia.php
> 
> // line 50
>             resource_path('js/pages'),
> ```
> 
> "Pages" -> "pages". I'm wondering if it's a file system / sensitivity issue with CI specifically.

I do see the errors on my VM and this prescribed change does make them go away.

```
...F.....F.F.FF...F..F.......................................  671 / 1127 ( 59%)
```

```
.............................................................  671 / 1127 ( 59%)
```

Chew:
I think it might be a case sensitivity issue since I use Mac and it does not have case sensitive folders like Linux (and presumably Windows).

resources/js/ziggy.js
wescopeland, line 1: I wonder if there's any way we can make this compatible both with raweb.test and localhost:64000. For a moment, I thought something was wrong with my local env 😅

wescopeland: Maybe in one or more git hooks we can auto-run `composer types`? Though I suppose this may also produce undesirable noise in commits later.

resources/js/features/redirect/components/+root/RedirectRoot.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+          <div className="flex w-full justify-center">
+            <button
+              onClick={() => {
+                window.location.replace(url);
+              }}
+              className="btn"
+            >
+              {t('Continue to external site')}
+            </button>
```
wescopeland: a11y: This needs a `role="link"` and should still support right click > open link in new tab.

I understand why the `.replace()` is desirable though. I think we can get the best of both worlds with an `onClick` handler attached to an anchor tag:
```tsx
<a
  href={url}
  onClick={(e) => {
    e.preventDefault();
    window.location.replace(url);
  }}
  className="btn"
  rel="noreferrer" // <--------- recommend adding this too
>
  {t('Continue to external site')}
</a>
```

resources/js/features/redirect/components/+root/RedirectRoot.tsx
```diff
@@ -0,0 +1,53 @@
...
+            </h1>
+          </div>
+        </div>
+      </div>
+
+      <div className="rounded border border-embed-highlight bg-embed p-4 md:text-center">
+        <div className="flex flex-col gap-y-4">
+          <p>
+            <Trans
+              i18nKey={'URLisExternalWarning'}
```
wescopeland: nit: Recommend changing this key to be identical to the value.

Also, given this is a string, no need to wrap the key in curly braces.

resources/js/features/redirect/components/+root/RedirectRoot.tsx
```diff
@@ -0,0 +1,53 @@
+import type { FC } from 'react';
+import { Trans, useTranslation } from 'react-i18next';
+
+import { usePageProps } from '@/common/hooks/usePageProps';
+
+export const RedirectRoot: FC = () => {
+  const { url } = usePageProps<{ url: string }>();
```
wescopeland: DX, consistency: the page props type should come from an auto-generated value from some `laravel-data` DTO.

resources/js/pages/redirect.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { AppLayout } from '@/common/layouts/AppLayout';
+import type { AppPage } from '@/common/models';
+import { RedirectRoot } from '@/features/redirect/components/+root';
+
+const Redirect: AppPage = () => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <SEO title={t('Redirect')} description="Redirect page" />
```
wescopeland: Security: this should be
```tsx
<SEO title={t('Redirect')} description="Redirect page" noindex={true} nofollow={true} />
```

resources/js/features/redirect/components/+root/RedirectRoot.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { expect } from 'vitest';
```
wescopeland: nit, consistency: We have these vitest imports configured as globals. This line is safe to remove.

resources/js/features/redirect/components/+root/RedirectRoot.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    render(<RedirectRoot />, {
+      pageProps: {
+        url: testUrl,
+      },
+    });
+
+    // ACT
+    const button = screen.getByRole('button', { name: 'Continue to external site' });
+    button.click();
```
wescopeland: This feedback will likely be redundant from the `button` -> `a` comment I left earlier, but when clicking on things in tests, always best to use `await userEvent.click(button)`. This simulates a real user asynchronously clicking on stuff rather than firing a browser event.

resources/js/ziggy.js
Chew: I'm not super familiar with this ziggy.js file, does it need to be committed at all? If we remove it, then ignore it, and keep an ignored file, what happens?

wescopeland: Good question. Every time I pull this branch, it breaks my local (and I'm sure you experience something similar for our branches).

The ziggy.js file is glue between the Inertia app and Laravel's strongly-typed routes. Without it, all routing in the Inertia app breaks.

Our deployment script automatically rebuilds the file on every deploy. I wonder if it would indeed be best to gitignore it and automatically execute `php artisan ziggy:generate --types` in whatever hooks we need (ie: _post-merge_).

resources/js/features/redirect/components/+root/RedirectRoot.test.tsx
wescopeland: Still need to use a `userEvent.click()` here.

resources/js/ziggy.js
wescopeland: I wonder if it's possible to kick this off when either `pnpm dev` or `pnpm build` run. That may universally solve the problem.

wescopeland: This seems promising: https://github.com/orgs/pnpm/discussions/5758#discussioncomment-4329215

wescopeland (approved):

resources/js/ziggy.js
wescopeland: Resolved in https://github.com/RetroAchievements/RAWeb/pull/4165.

PR #4078, opened 2025-10-29
feat(game): allow sorting leaderboards by your rank

Chew:
Very simple but seemingly useful QoL sort option for leaderboard sorting: sorting by your rank in those leaderboards.

Leaderboards where you are not ranked are not sorted.

Below are examples:

CleanShot 2025-10-28 at 21 15 40@2x
CleanShot 2025-10-28 at 21 15 34@2x


resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
wescopeland, line 1: Bug (unrelated to this file): when refreshing the page, the new rank sort orders aren't retained.

resources/js/common/utils/sortLeaderboards.ts
wescopeland, line 1: Bug: `-rank` doesn't seem to be working correctly. It probably needs a test case.
Screenshot 2025-10-29 at 8 40 18 AM

wescopeland (commented):
Hi @Chew! I agree this feature is desirable. Just did a quick UX check.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
Chew: Had to hunt this down but I got it!

resources/js/common/utils/sortLeaderboards.ts
Chew: Yeah I forgot to handle the case of 1st and 3rd leaderboard instead of 1st and 2nd. I rewrote the sorting a bit but I still think it's a bit of a workaround... not sure the best way to forcefully sort out the unranked leaderboards efficiently.

resources/js/common/utils/sortLeaderboards.test.ts
```diff
@@ -106,6 +106,27 @@ describe('Util: sortLeaderboards', () => {
     expect(result.map((l) => l.id)).toEqual([1, 3, 2]);
   });
 
+  it('given rank sort, sorta by the user rank', () => {
```
wescopeland: Typo:
```suggestion
  it('given rank sort, sorts by the user rank', () => {
```

resources/js/common/utils/sortLeaderboards.test.ts
wescopeland, line 1: Testing: Let's also include a test case for `-rank`.

resources/js/common/components/PlayableListSortButton/PlayableListSortButton.tsx
```diff
@@ -68,6 +69,9 @@ export const PlayableListSortButton: FC<PlayableListSortButtonProps> = ({
 
     '-wonBy': { label: t('Won by (least)'), icon: LuUsers },
     wonBy: { label: t('Won by (most)'), icon: LuUsers },
+
+    '-rank': { label: t('Rank (lowest)'), icon: LuHash },
+    rank: { label: t('Rank (highest)'), icon: LuHash },
```
wescopeland: UX: I'm wondering if we should reconsider these labels. When we sort by "Rank (highest)", we're technically sorting by the lowest number 😅 

Screenshot 2025-10-30 at 6 12 06 PM

Possible alternatives:
* Best / Worst
* Top / Bottom

Chew: Yeah I couldn't think of a better term. "Highest rank" was what I had in mind. I think best / worst is probably the better option here, and might translate better into other languages when that happens.

Jamiras: It's just splitting hairs at this point, but `99/100` is arguably "worse" than `100/200`.

Having said that, I do agree it's probably the best choice.

resources/js/common/components/PlayableListSortButton/PlayableListSortButton.tsx
```diff
@@ -68,6 +69,9 @@ export const PlayableListSortButton: FC<PlayableListSortButtonProps> = ({
 
     '-wonBy': { label: t('Won by (least)'), icon: LuUsers },
     wonBy: { label: t('Won by (most)'), icon: LuUsers },
+
+    '-rank': { label: t('Rank (lowest)'), icon: LuHash },
```
Jamiras: @wescopeland - I was going to sugget using `LuArrowDownNarrowWide` and `LuArrowDownWideNarrow` for more consistency with other items in the menu, but then I realized the sort menu for achievements has items without arrows, so I'm not sure if it would be better. Do you have a preference?

wescopeland: Yeah, unfortunately not all Lucide icons have complementary arrow companions.

No hard opinion on this from me. Happy to go with alternative icons.

Chew: Yeah I just went with whatever I could find from the autocomplete, happy to tru anything else. Do wish there was an arrow up/down with a hash

resources/js/common/utils/sortLeaderboards.test.ts
Chew: Added a test for this and also added a null rank entry in the middle to make sure that gets sorted out. I don't think that needs 2 separate tests (to not factor that in) but it is important to check that

resources/js/common/utils/sortLeaderboards.ts
```diff
@@ -37,6 +37,20 @@ export function sortLeaderboards(
       });
     }
 
+    case 'rank':
+    case '-rank': {
+      const reversed = sortOrder === 'rank';
```
wescopeland: DX, nit: `reversed` as a variable name might be confusing for a future maintainer.

When `sortOrder === 'rank'`, you want best-first (ascending), not reversed. I suggest renaming the variable to either `isAscending` or `isBestFirst`.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameAchievementSetToolbar.tsx
```diff
@@ -116,7 +116,7 @@ export const GameAchievementSetToolbar: FC<GameAchievementSetToolbarProps> = ({
                   'type',
                   '-type',
                 ]
-              : ['displayOrder', '-displayOrder', 'title', '-title']
+              : ['displayOrder', '-displayOrder', 'title', '-title', 'rank', '-rank']
```
wescopeland: Sorry I didn't notice this one when testing earlier.

Bug/UX, minor: The new sort orders appear when the user isn't logged in. This should work, will just need a test case:
```ts
: [
    'displayOrder',
    '-displayOrder',
    'title',
    '-title',
    ...(auth?.user ? ['rank', '-rank'] : []),
  ]
```

resources/js/common/utils/sortLeaderboards.ts
```diff
@@ -37,6 +37,20 @@ export function sortLeaderboards(
...
+
+      const sorted = sortedLeaderboards.sort((a, b) => {
+        const aRank = a.userEntry?.rank || (!reversed ? 0 : 9999999); // this isn't ideal
+        const bRank = b.userEntry?.rank || (!reversed ? 0 : 9999999);
```
wescopeland: nit: Rather than `9999999`, we can avoid reaching for magic numbers by using `Number.MAX_SAFE_INTEGER`.

wescopeland (approved):
Thanks!

Jamiras (approved):

PR #4079, opened 2025-10-29
feat(forum): add support for [game=X set=Y] shortcodes

wescopeland:
This PR adds support for game URLs with set parameters in the forum.

On submit, these are converted to `[game={backingGameId}]`.

Screenshot 2025-10-29 at 6 30 44 PM


app/Support/Shortcode/Shortcode.php
```diff
@@ -128,6 +129,45 @@ public static function convertUserShortcodesToUseIds(string $input): string
         }, $input);
     }
 
+    public static function convertGameSetShortcodesToBackingGame(string $input): string
+    {
+        // Extract all [game=X?set=Y] or [game=X set=Y] patterns.
```
Jamiras, line 134: If the pattern is detected, the game portion is ignored.
```
[game=9999 set=9534]
```
still maps to the sonic subset.

Should the game association be validated?

Wouldn't it be easier to just have `[set=9534]`?

app/Helpers/shortcode.php
```diff
@@ -96,6 +96,17 @@ function normalize_shortcodes(string $input): string
 function normalize_targeted_shortcodes(string $input, string $kind, ?string $tagName = null): string
 {
     // Find any URL variants of user links and transform them into shortcode tags.
+    // First, handle URLs with a ?set= query param (these are games).
+    if ($kind === 'game') {
+        $findWithSet = [
+            "~https?://(?:[\w\-]+\.)?retroachievements\.org/game/([\w]{1,20})(?:-[^\s\"'<>]*)?(?:/)?\\?set=(\d+)~si",
```
Jamiras: Should this also match `/game2/`?

resources/js/common/utils/shortcodes/preProcessShortcodesInBody.ts
```diff
@@ -7,31 +7,52 @@ const shortcodeTypes = [
...
+      'gi',
+    ),
+  ];
+
+  // For games specifically, also handle URLs with ?set= query parameter.
+  if (type === 'game') {
+    patterns.unshift(
+      // Production URLs with a ?set= parameter.
+      new RegExp(
+        `https?://(?:[\\w-]+\\.)?retroachievements\\.org/game/(\\w{1,20})(?:-[^\\s"'<>]*)?(?:/)?\\?set=(\\d+)`,
```
Jamiras: Same question about `/game2/`.

app/Support/Shortcode/Shortcode.php
wescopeland: Logically, I think a `[set=]` shortcode is easier, but I'm afraid it may be more challenging from a UX POV:
1. It's one more button.
2. We don't expose the set IDs of base sets, which could make the functionality ambiguous or confusing.

Granted, `[game=X?set=Y]` doesn't feel good from a UX POV either.

app/Helpers/shortcode.php
wescopeland: I don't think so. It'd probably be best to merge this change when we wrap up the beta.

app/Support/Shortcode/Shortcode.php
wescopeland: The more I think on this, the more I'm wondering if it's worth simplifying this implementation significantly. Rather than messing with shortcodes, what if we just detect inline game set URLs (whether it's game or game2), and have this functionality only work on those things. That may be an easier UX to explain to folks.

wescopeland: Turns out it's really hard to rip out some kind of `?set=Y]` shortcode handling and still have the Rube Goldberg machine work correctly.

In latest, I've resigned to us keeping the `[game=X set=Y]` handling, but we also support auto-processing of game2 URLs as suggested in other feedback. I'm hoping from a UX POV we can keep the "set=" shortcode stuff as an implementation detail, and not something we advertise to users.

Jamiras: If I include a `[game=X set=Y]` link in a post, all the other subset links point to set Y, regardless of what they're set to.
```
http://localhost:64000/game2/697?set=59

http://localhost:64000/game/697?set=59

http://localhost:64000/game/20026

[game=697 set=59]

[game=9999 set=59]

[game=697 set=9999]
```

The first five should all point at the same subset.

image

Oddly, removing the last line causes all the broken items to disappear.

image

wescopeland: Tricky, but should be addressed in latest.

wescopeland:
I think we should obscure `[game=X set=Y]` as an implementation detail and instead reverse [game={backingGameId}] to `route('game.show', ['game' => $game, 'set' => $set]`. In other words, show the user the absolute URL when they go to edit the post.

This is sufficiently complicated enough that I'm going to merge this and tackle the reversal logic in a subsequent PR.

PR #4084, opened 2025-10-30
(refactor) add PingAction

Jamiras:


app/Connect/Actions/PingAction.php
```diff
@@ -0,0 +1,87 @@
...
+            return $result;
+        }
+
+        $this->richPresenceMessage = utf8_sanitize(request()->post('m', ''));
+        $this->gameHashMd5 = request()->input('x');
+
+        $hardcore = request()->integer('h');
+        if ($hardcore !== null) {
+            $this->hardcore = ($hardcore === 1);
+        }
```
wescopeland, line 59: Intellectual curiosity: is this covered by any test? If not, should it be?

Jamiras: The `h` parameter was added to the API at the same time the `x` parameter was added. Nothing is currently using the `h` parameter, therefore there's nothing to test against.

PR #4090, opened 2025-10-31
feat: add a shortcode to markdown converter, hook it up to Discord forwarding

wescopeland:
Adds a method which converts all the shortcodes we support, except `[link]` (only 15 usages in the DB), to markdown.

Also hooks this up to our Discord forwarding, as demo'd here: https://discord.com/channels/476211979464343552/1433922449292001481

Screenshot 2025-10-31 at 7 00 50 PM

Screenshot 2025-10-31 at 7 02 00 PM


tests/Feature/Community/ShortcodeTest.php
```diff
@@ -334,4 +341,485 @@ public static function youtubeUrlProvider(): array
...
+        );
+    }
+
+    public function testConvertToMarkdownItalic(): void
+    {
+        $this->assertSame(
+            '*Hello*',
+            Shortcode::convertToMarkdown('[i]Hello[/i]')
+        );
+    }
```
Jamiras, line 359: Suggest test for both italics and bolt:
```
[b][i]Hello[/i][/b]
```

tests/Feature/Community/ShortcodeTest.php
```diff
@@ -334,4 +341,485 @@ public static function youtubeUrlProvider(): array
...
+        $this->assertSame(
+            "> Line 1\n> Line 2\n> Line 3",
+            Shortcode::convertToMarkdown("[quote]Line 1\nLine 2\nLine 3[/quote]", 10000, preserveWhitespace: true)
+        );
+    }
+
+    public function testConvertToMarkdownNestedQuotes(): void
+    {
+        $this->assertSame(
+            '> > Inner quote Outer quote', // nested quotes get prefixed together
```
Jamiras, line 412: This seems wrong. I would expect:
```
> > Inner quote
> Outer quote
```

But that might be because there isn't a newline between the inner and outer quote.

Additionally, Discord doesn't seem to do nested quotes, so it's probably good enough.

tests/Feature/Community/ShortcodeTest.php
wescopeland: Added this test case in latest.

PR #4092, opened 2025-11-01
fix(game): sort unlocked achievements by timestamp correctly

wescopeland:
Resolves https://discord.com/channels/476211979464343552/1026595325038833725/1434044496215277650.

**Before**
Screenshot 2025-11-01 at 7 56 35 AM

**After**
Screenshot 2025-11-01 at 7 56 25 AM


Jamiras:
I believe this behavior is intentional. This sorts by display order and groups by unlocked. It's the same order that is the default on the game1 page. Maybe it should be renamed to "Display Order (Unlocked first)"?

PR #4099, opened 2025-11-03
feat(user): add 'Muted' role on Discord when users are muted

wescopeland:
Implements No.1 from https://github.com/RetroAchievements/RAWeb/discussions/4067:
> 1. When someone gets muted on the site, they should receive a ''Muted'' role on Discord automatically.

When someone is muted, they will receive the new "Muted" role rather than having all their roles removed.
When they are manually unmuted or their mute expires, the Muted role will automatically be removed.

Additionally, we now automatically clear `muted_until` when users are unmuted or their mute expires.

`DISCORD_ROLE_MUTED` has been configured in prod's environment file.

This code is currently untested. I believe it should work, as it mostly leverages stuff we already have that does work. I'm not sure how to actually test it though short of getting it out to prod.


app/Community/Actions/AddUserDiscordRoleAction.php
```diff
@@ -0,0 +1,62 @@
...
+                return;
+            }
+
+            $userId = $member['user']['id'];
+            $currentRoles = $member['roles'] ?? [];
+
+            // Add the role if they don't already have it.
+            if (!in_array($roleId, $currentRoles, true)) {
+                $currentRoles[] = $roleId;
+            }
```
Jamiras: Would it be better to avoid the discard API call if the user already has the role. Is there any reason to update the user to their current state?

It looks like the remove action is doing things that way.

app/Observers/UserObserver.php
```diff
@@ -12,13 +14,26 @@ class UserObserver
...
+        }
+
+        // Handle manual unmuting - remove the "Muted" role if they're a Discord member.
+        if ($this->isBeingUnmuted($user) && $mutedRoleId) {
+            (new RemoveUserDiscordRoleAction())->execute($user, $mutedRoleId);
+        }
+
+        // Handle banning - remove all roles if they're a Discord member.
+        if ($this->isBeingBanned($user)) {
             (new RemoveDiscordRolesAction())->execute($user);
```
Jamiras: Should this be renamed to `RemoveUserDiscordRolesAction`?

Jamiras (commented):
Looked through the code. Did not do any testing due to the integration requirement.

app/Community/Actions/AddUserDiscordRoleAction.php
wescopeland: Agreed - in latest the API call will now be avoided if `$rolesToAdd` is empty.

app/Observers/UserObserver.php
wescopeland: I came up with a different approach in latest I think I like more. I consolidated both actions into a single `RemoveUserDiscordRolesAction` that accepts an array `$roleIds` as the 2nd arg. If the array is empty, all roles are removed, otherwise the given role IDs are removed.

Additionally, `AddUserDiscordRolesAction` (now plural) similarly accepts an array as its 2nd arg for consistency.

PR #4103, opened 2025-11-05
feat: add a Report button to forum posts and incoming DMs

wescopeland:
This PR gives users the ability to report forum posts and incoming DMs to the moderation team.

**Forum Posts**
Screenshot 2025-11-05 at 5 37 17 PM

**Incoming DMs**
Screenshot 2025-11-05 at 5 36 31 PM

Upon clicking the "Report" button, the user is taken to a special variant of the message thread composition page:
Screenshot 2025-11-05 at 5 37 30 PM

Upon submitting the report, a new Discord forum thread is automatically created:
Screenshot 2025-11-05 at 5 39 18 PM

If a user reports an incoming DM, the entire DM is included in the report rather than a small excerpt.

Comments and user profiles can't be reported yet. That'll be handled in a subsequent PR.


app/Policies/UserPolicy.php
```diff
@@ -285,6 +285,16 @@ public function clearUserWall(User $user, User $model): bool
         return $this->requireAdministrativePrivileges($user, $model);
     }
 
+    public function createModerationReports(User $user): bool
+    {
+        // Unverified users and muted users cannot create reports.
+        if (!$user->isEmailVerified() || $user->isMuted()) {
+            return false;
+        }
```
Jamiras, line 294: Recommend also requiring `!isFreshAccount()` to minimize abuse.

app/Community/Actions/ForwardMessageToDiscordAction.php
```diff
@@ -83,17 +94,36 @@ public function execute(
...
+        } elseif ($isForum) {
+            $existingThreadId = $this->getExistingDiscordThreadId($messageThread);
+        }
+
         $isNewThread = $isForum ? !$existingThreadId : true;
 
+        // Prepend context for NEW report messages only (the OP).
+        // For deduplicated reports, context is already in the OP.
+        if ($reportableType && $reportableId && !$isExistingReport) {
+            $fullBody = $this->prependReportContext($fullBody, $reportableType, $reportableId);
```
Jamiras: Is there a reason this functionality is restricted to the discord message? There's no context _at all_ in the message left in the inbox.

Shouldn't the actual message stored in the Inbox at least include a link to the offending content?

wescopeland: Not sure I am 100% following, can you explain this a different way?

For example, with [this report](https://discord.com/channels/476211979464343552/1435760300338057227), we do include a link to the reported content as well as an excerpt for context.

For [this DM report](https://discord.com/channels/476211979464343552/1435418307988557894), we can't link directly to the DM, so we include the full message and some context about when it was sent and the message author.

Jamiras: I'm referring to the Message created in the website.

image

image

There's nothing in these providing any context beyond the text entered by the user to describe what is being reported.

Jamiras: I would expect a link to the forum post (or the direct message - as the author would still have access to that), possibly with a snippet of the content.

wescopeland: Thanks for that explanation. Updated in latest:

Screenshot 2025-11-07 at 7 05 33 PM

app/Policies/UserPolicy.php
wescopeland: Done in latest, though I did add some conditional logic for the forum verification check.

app/Community/Actions/ForwardMessageToDiscordAction.php
Jamiras: Looks good for forum posts, but I still feel like context is missing for the direct messages.

Having the user and a date helps to find the reported message, but would it be possible/practical to link to the message thread (if not the actual message)? I realize that will be inacessible for moderators, but it still provides context for the reporter.

app/Policies/UserPolicy.php
Jamiras: Why would forum verification matter? `isFreshAccount` requires all four checks to fail to say the account is new. Skipping the forum verification would actually prevent someone who is forum verified but doesn't match the other requirements from reporting a post.

wescopeland: I don't know what I was thinking here. Reverted that change in latest.

app/Community/Actions/ForwardMessageToDiscordAction.php
wescopeland: Added in latest
Screenshot 2025-11-09 at 1 49 47 PM

Also confirmed it doesn't arrive in the message forwarding to Discord: https://discord.com/channels/476211979464343552/1437152140585079039

Jamiras (approved):

PR #4100, opened 2025-11-04
chore(connect): remove dead code

wescopeland:
Fixes https://github.com/RetroAchievements/RAWeb/pull/4099#issuecomment-3485746093 and updates the MVC documentation for the service provider.

PR #4105, opened 2025-11-06
(refactor) move connect token management into trait

Jamiras:
addresses https://github.com/RetroAchievements/RAWeb/pull/4100#discussion_r2495377984 and https://github.com/RetroAchievements/RAWeb/pull/4100#discussion_r2495373921

moves all the connect token management stuff into a new trait: `HasConnectToken`.


app/Platform/Concerns/HasConnectToken.php
wescopeland, line 1: nit: Multiline comments in this file are inconsistently formatted. IDEs generally require this format for IntelliSense to work correctly:
```php
/**   # one slash and two stars to open
 * words begin here
 */   # one star and a slash to close
```

app/Platform/Concerns/HasConnectToken.php
```diff
@@ -0,0 +1,65 @@
...
+     * token to persist indefinitely as long as the user remains active.
+     */
+    private const CONNECT_TOKEN_EXPIRY_DAYS = 14;
+
+    /**
+     * Gets whether the Connect API Key is valid.
+     */
+    public function isConnectTokenValid(): bool
+    {
+        return !empty($this->appToken) && strlen($this->appToken) === self::CONNECT_TOKEN_LENGTH;
```
wescopeland, line 33: nit: The old code used `mb_strlen` rather than `strlen` for multibyte safety.

This probably isn't that much of an issue because `Str::random()` is only going to generate ASCII characters, but I wonder if this should err on the side of caution and remain as `mb_strlen`.

app/Platform/Concerns/HasConnectToken.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * NOTE: does not commit the new token. caller must ->save() or ->saveQuietly().
+     */
+    public function generateNewConnectToken(): void
+    {
+        do {
+            $this->appToken = Str::random(self::CONNECT_TOKEN_LENGTH);
+        } while ($this->where('appToken', $this->appToken)->exists());
+
+        $this->extendConnectTokenExpiry();
```
wescopeland, line 55: Intellectual curiosity: I can't help but wonder if it's better to remove this line.

Currently, when generating a new token, this function calls `extendConnectTokenExpiry()` to set the new expiry. Then `completeLogin()` immediately calls `extendConnectTokenExpiry()` again before saving. This double extension is harmless, albeit redundant.

app/Platform/Concerns/HasConnectToken.php
Jamiras: If something is passed in that is 16 unicode characters, it can't possibly match 16 randomly generated ASCII characters. This also enshrines the fact that a valid connect token doesn't contain multibyte characters.

`mb_strlen` has significant overhead as it has to process each character to determine how many bytes it takes. `strlen` can just scan for a terminator and subtract the two pointers. It's much more efficient.

app/Platform/Concerns/HasConnectToken.php
Jamiras: `generatedNewConnectToken` is also called when a user changes their password or logs out of all emulators. In all cases, when a new token is generated, we want to also set the expiration time.

The redundancy you've highlighted only happens when new tokens are generated by the login process (which should only occur if the user doesn't have a previous token, or the token has expired). These are infrequent events, so I feel comfortable doing the double update. Also, because the function doesn't do the commit, there's still only one write to the DB.

app/Platform/Concerns/HasConnectToken.php
Jamiras: done.

PR #4107, opened 2025-11-06
(refactor) store password reset tokens in password_reset_tokens table

Jamiras:
Moves them from the UserAccounts table, and periodically destroys tokens older than 5 days.

Instead of migrating the 300k+ tokens already in the database, I've chosen to make a clean break on deploying this feature. That means any tokens generated before deployment (even seconds before deployment) will be invalid and the user will have to request another password change to get a new token.

As this PR removes the PasswordResetToken column from the UserAccounts table, I've taken the liberty of removing a few other unused columns too - more details in the migration step.


app/Models/User.php
```diff
@@ -87,9 +87,8 @@ class User extends Authenticatable implements CommunityMember, Developer, HasLoc
     use CollectsBadges;
 
     // TODO rename UserAccounts table to users
-    // TODO drop cookie, fbUser, fbPrefs, LastActivityID, LastGameID, PasswordResetToken, UnreadMessageCount
```
Jamiras, line 92: LastGameID and UnreadMessageCount are still TODO'd a couple lines down. The rest of these have been removed.

database/migrations/2025_11_06_000000_update_password_reset_tokens_table.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::dropIfExists('password_reset_tokens');
```
Jamiras, line 12: This table is completely recreated. It was previously unused, and mapped an email address to a token. Since resets are requested for a Username, it doesn't make sense to capture the email address the token was sent to. Now the token is mapped to a user_id, and I've added an additional column to help with https://discord.com/channels/476211979464343552/1002689485005406249/1426984613548658820

database/migrations/2025_11_06_000001_update_user_accounts_table.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::table('UserAccounts', function (Blueprint $table) {
+            $table->dropIndex('users_last_activity_id_index');
```
Jamiras, line 13: This freed up 200MB on my local copy of the database.

database/migrations/2025_11_06_000001_update_user_accounts_table.php
```diff
@@ -0,0 +1,37 @@
...
+
+            $table->dropColumn([
+                'PasswordResetToken',
+                'fbUser',
+                'fbPrefs',
+                'cookie',
+                'LastActivityID',
+            ]);
```
Jamiras, line 21: * `PasswordResetToken` is being migrated to the `password_reset_tokens` table and now supports multiple tokens per user_id in case emails take an unusual amount of time to get delivered and the user makes more requests. Tokens now expire after 5 days, and there's still a rate limited in effect for requesting tokens.
* `fbUser` and `fbPrefs` are remnants from long ago. It's hard to tell when the feature was eliminated, but the most recently created account that has it set was made in 2018.
* I'm not exactly sure how `cookie` was used, and it's very hard to grep for. I think I got all the references.
* `LastActivityID` points to a table that no longer exists (#2074)

public/request/auth/send-password-reset-email.php
```diff
@@ -22,7 +25,15 @@
 $targetUser = User::whereName($input['username'])->first();
 
 if ($targetUser && !$targetUser->isBanned()) {
-    RequestPasswordReset($targetUser);
+    $newToken = Str::random(20);
+
+    PasswordResetToken::create([
+        'user_id' => $targetUser->id,
+        'token' => $newToken,
```
wescopeland: Security, minor: I [recommend](https://security.stackexchange.com/a/86915) we always hash these tokens:
```php
$newToken = Str::random(20);
$hashedToken = Hash::make($newToken);
```

The `token` column itself will likely need to be expanded from 32 characters to support the hashed string length.

public/request/auth/reset-password.php
```diff
@@ -17,7 +18,15 @@
...
 
-if (!$targetUser || $targetUser->isBanned() || !isValidPasswordResetToken($targetUser->username, $passResetToken)) {
+if (!$targetUser || $targetUser->isBanned()) {
+    return back()->withErrors(__('legacy.error.token'));
+}
+
+$token = PasswordResetToken::query()
+    ->where('user_id', $targetUser->id)
+    ->where('token', $passResetToken)
+    ->first();
```
wescopeland: Security, major: this query should also check for token expiration. 60 minutes is probably good enough.
```php
$token = PasswordResetToken::query()
    ->where('user_id', $targetUser->id)
    ->where('token', Hash::check($passResetToken))
    ->where('created_at', '>', now()->subMinutes(60))
    ->first();
```

public/request/auth/send-password-reset-email.php
wescopeland, line 1: If the user has any existing password reset tokens and requests a new one, we should immediately invalidate (delete) the existing token.

wescopeland, line 1: nit: My local VSCode is unhappy about this file missing a `Str` import with the other `use` statements.

app/Models/PasswordResetToken.php
```diff
@@ -0,0 +1,39 @@
...
+    protected $table = 'password_reset_tokens';
+
+    protected $fillable = [
+        'user_id',
+        'token',
+        'ip_address',
+        'created_at',
+    ];
+
+    public $timestamps = false;
```
wescopeland, line 24: This means we'll also need:
```php
protected $casts = [
    'created_at' => 'datetime',
];
```

database/migrations/2025_11_06_000001_update_user_accounts_table.php
wescopeland: > now supports multiple tokens per user_id in case emails take an unusual amount of time to get delivered and the user makes more requests

This is well-intentioned, but in my view undesirable. I don't believe we should support this.

Laravel itself has a hard restriction on only storing a single token per user with its built-in auth system. This is a standard approach because:
1. Email delivery is typically fast.
2. If an email doesn't arrive, the user can just request another token.
3. By the time a delayed email arrives minutes/hours later, they've probably already reset their password using a newer email.
4. Supporting multiple concurrent tokens has a larger potential surface area for things to go wrong (security, user error) than only supporting one for a user.

public/request/auth/send-password-reset-email.php
Jamiras: The next post on that link says:
> Typically, password reset tokens are time-limited. 
> ...
> If you do this, there's no absolute need to hash the password reset tokens when they're stored in your database. Let's remember the threat model that hashing is designed to defend against: it's designed to mitigate one-time database breaches.

How long should the column be? The Password field on the UserAccounts table is 255 characters. 

And do we really want to include something like `$argon2id$v=19$m=65536,t=4,p=1$....` in the email link?

public/request/auth/reset-password.php
Jamiras: The current model doesn't expire tokens at all. That's a major securty issue. This PR will purge tokens older than 5 days, which means a token could live up to a few seconds less than 6 days. I could also validate the expiration here, but assumed the purge was sufficient.

public/request/auth/send-password-reset-email.php
wescopeland: No, the email sent to the user should contain the plain unhashed token. The hashed token only lives in the database. When the user uses the token, we do a hash check on it, similar to passwords. Password reset tokens are effectively passwords with an expiry.

Regarding the quoted comment - that advice applies when tokens expire in 15-60 minutes with enforced expiry validation. The current implementation has a 5-day expiry, no expiry check in the validation code, and tokens can be days old between cleanup crons.

[Laravel itself hashes password reset tokens](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L124).

Imagine this scenario:
* We store unhashed plain-text tokens in the DB.
* Someone finds a SQL injection vulnerability in the code.
* They request a password reset for any admin account.
* They use the vulnerability to get the latest record in the table, which has the plain-text reset token.
* They reset the password and do unfathomable damage.

public/request/auth/reset-password.php
wescopeland: I agree that the current implementation in `master` is flawed. I'm grateful we're visiting it in this PR.

I think the nuance here is the purge cleanup is garbage collection, whereas a `created_at` check is validation. This is also something [enforced by Laravel's own implementation](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L141).

Expiry validation is generally considered a must-have:
* [NIST SP 800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html): for out-of-band verifiers "In all cases, the authentication SHALL be considered invalid if not completed within 10 minutes."
* [OWASP ASVS 2.7.3](https://owasp-aasvs4.readthedocs.io/en/latest/2.7.3.html): this requirement explicitly states the validation must happen for the authentication request itself.
* [OWASP Forgot Password Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#url-tokens): "additional validation steps" should occur when the user presents the token.

app/Models/PasswordResetToken.php
Jamiras: done.

public/request/auth/reset-password.php
Jamiras: added a check to ensure the token is less than 60 minutes old.

public/request/auth/send-password-reset-email.php
Jamiras: I've chosen to null out the previous tokens. This keeps the extra data around indicating further requests had been made.

Jamiras: added.

public/request/auth/send-password-reset-email.php
Jamiras: That makes more sense. I hadn't considered sending the unhashed value to the client.

Done.

PR #4111, opened 2025-11-06
feat: elevate Muted/Banned status social visibility

wescopeland:
This PR implements the remainder of section 4.1 from this discussion: https://github.com/RetroAchievements/RAWeb/discussions/4067

> ... nobody is aware of when we mute users, so we could possibly add a small sign on a user profile and/or under their profile picture in forum posts that they are muted. Similar to how it shows Junior Developer/Developer/Moderator under the profile picture, but then instead shows ''Muted'', like it's a role.

Now, a user's muted status is publicly visible in their authored forum posts and their user profile:

Screenshot 2025-11-06 at 6 25 59 PM

Screenshot 2025-11-06 at 6 26 28 PM

---

Additionally, a user's banned status is now publicly visible on their authored forum posts, since we are exposing that as well (https://github.com/RetroAchievements/RAWeb/pull/4110):

Screenshot 2025-11-06 at 6 25 48 PM


resources/views/components/user/profile/primary-meta.blade.php
Jamiras, line 1: Should these changes also be propogated to the user card?

Image

Image

Image

wescopeland: We should. Updated in latest.

Additionally, in latest, the React `UserAvatar` component will no longer hyperlink banned users. This seems desirable, as it's otherwise leading people to a 404.

Jamiras (approved):

PR #4120, opened 2025-11-09
feat: add a Report button to most comment walls

wescopeland:
This PR gives users the ability to report most wall comments to the moderation team. Both the legacy Blade comment items and React comment items are supported.

Assuming the user can create a report, the new link button is always visible on mobile and visible on comment hover for desktop users.

**Blade/PHP UI**
_Mobile_
Screenshot 2025-11-09 at 10 43 42 AM

_Desktop_
Screenshot 2025-11-09 at 10 44 13 AM

**React UI**
_Mobile_
Screenshot 2025-11-09 at 10 45 26 AM

_Desktop_
Screenshot 2025-11-09 at 10 45 44 AM

---

Similar to other reports, when a comment is reported a new forum thread for the mod team is created.

Screenshot 2025-11-09 at 10 46 41 AM

### DIFF
--- a/app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
+++ b/app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
@@ -8,6 +8,7 @@ use App\Data\ForumTopicData;
 use App\Data\PaginatedData;
 use App\Enums\Permissions;
 use App\Models\ForumTopic;
+use App\Support\Shortcode\Shortcode;
 use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Facades\DB;
 
@@ -26,8 +27,17 @@ class BuildAggregateRecentForumPostsDataAction
     ): PaginatedData|array {
         $topics = $this->getRecentForumTopics($page, $permissions, $limit);
 
+        $shortcodeIds = [];
+        foreach ($topics as $topic) {
+            $postShortcodeIds = Shortcode::extractShortcodeIds($topic['ShortMsg']);
+            foreach ($postShortcodeIds as $key => $ids) {
+                $shortcodeIds[$key] = array_merge($shortcodeIds[$key] ?? [], $ids);
+            }
+        }
+        $shortcodeRecords = Shortcode::fetchRecords($shortcodeIds);
+
         $transformedTopics = array_map(
-            fn ($topic) => ForumTopicData::fromRecentlyActiveTopic($topic)->include(
+            fn ($topic) => ForumTopicData::fromRecentlyActiveTopic($topic, $shortcodeRecords)->include(
                 'commentCount24h',
                 'oldestComment24hId',
                 'commentCount7d',
--- a/app/Community/Actions/BuildThinRecentForumPostsDataAction.php
+++ b/app/Community/Actions/BuildThinRecentForumPostsDataAction.php
@@ -6,6 +6,7 @@ namespace App\Community\Actions;
 
 use App\Data\ForumTopicData;
 use App\Enums\Permissions;
+use App\Support\Shortcode\Shortcode;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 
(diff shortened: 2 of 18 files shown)


resources/views/components/comment/item.blade.php
```diff
@@ -40,26 +67,62 @@
...
+                            href="{{ route('message-thread.create', [
+                                'to' => 'RAdmin',
+                                'subject' => 'Report: ' . $commentTypeLabel . ' by ' . ($author->display_name ?? 'Deleted User'),
+                                'rType' => 'Comment',
+                                'rId' => $commentId
+                            ]) }}"
+                            aria-label="Report comment"
+                            title="Report comment"
+                        >
+                            <x-fas-flag class="text-yellow-600 size-3 sm:hidden" />
```
Jamiras, line 96: I wonder if there's a better icon for this? Outlook uses a flag to allow a user to mark important things. Since there's no label, and there's a slight delay for the tooltip, I wonder how many users will use it incorrectly.

Although, since it takes you to an input form and doesn't just submit a report, it should be self-correcting without any additional overhead to us.

app/Community/Actions/BuildReportContextAction.php
```diff
@@ -98,4 +108,39 @@ public function execute(
...
+        if (!$comment->ArticleID) {
+            return null;
+        }
+
+        $anchor = '#comment_' . $comment->ID;
+
+        return match ($comment->ArticleType) {
+            ArticleType::Game => ($game = Game::find($comment->ArticleID))
+                ? route('game.show', $game) . $anchor
+                : null,
```
Jamiras, line 127: This link doesn't take me to the comment.

http://localhost:64000/game/20816#comment_4891773

For reference, this is the email notification link, which doesn't work either.

http://localhost:64000/game/20816?tab=community#comment_5700588

resources/views/components/comment/item.blade.php
```diff
@@ -40,26 +67,62 @@
          $author && $author->banned_at && !request()->user()?->can('manage', $author))
     {{-- banned user comments are only visible to moderators --}}
 @else
-    <tr class="comment" @if ($commentId > 0) id="comment_{{ $commentId }}_highlight" @endif>
+    <tr class="comment group" @if ($commentId > 0) id="comment_{{ $commentId }}_highlight" @endif>
         <td class="align-top py-2">
             @if ($commentId > 0)
                 <div class="relative">
                     <div class="absolute h-px w-px left-0" style="top: -74px;" id="comment_{{ $commentId }}"></div>
```
Jamiras, line 75: The error points me to this line, but viewing a game or achievement without being logged in fails.

> Call to a member function can() on null

resources/views/components/comment/item.blade.php
wescopeland: I'm not sure we should change this icon - a flag is fairly conventional for social reporting.

Twitter:
Screenshot 2025-11-17 at 3 11 10 PM

YouTube:
Screenshot 2025-11-17 at 3 12 03 PM

Reddit:
Screenshot 2025-11-17 at 3 12 21 PM

There are a few counterexamples where a different icon is used such as Facebook and Instagram, but the norm does indeed seem to be a flag.

app/Community/Actions/BuildReportContextAction.php
wescopeland: This will only work with SSR enabled, eg:
https://retroachievements.org/game/20816#comment_4891773

This happens locally because without SSR enabled, the UI isn't mounted on the initial page render, so there isn't anything for the browser to snap to.

resources/views/components/comment/item.blade.php
wescopeland: Thanks - fixed in latest.

resources/views/components/comment/item.blade.php
Chew: As a user I typically associate a flag with reporting (especially if it's red), and I think most others will too. I'm not sure what the ambiguity would be otherwise, Outlook is "flagging" it for you to review later, this is just flagging a comment for review by a mod.

app/Community/Actions/BuildReportContextAction.php
Jamiras: Do we care about the presence/lack of `tab=community`?

wescopeland: Yep, added in latest.

PR #4121, opened 2025-11-09
refactor(game): convert the achievement set tooltip cards to use BaseHoverCard

wescopeland:
This PR migrates this component from `BaseTooltip` to `BaseHoverCard`:
Screenshot 2025-11-09 at 11 21 20 AM

There shouldn't be much, if any, of a discernible difference (with the exception of you can now move your mouse into the card content).

Hover cards have different accessibility implications than tooltips: they aren't announced to screen readers and they don't pop for mobile devices. They also are more responsive when stacked next to each other horizontally. All in all, I think they're a better fit for these components.


Jamiras:
> they don't pop for mobile devices.

I'm not sure if it's because I'm emulating mobile, but they do seem to pop sometimes - presumably dependent on how quickly the tooltip responds vs. the page turn.

https://github.com/user-attachments/assets/811c1c3d-558f-4048-8da7-1d12ef70cc2d

wescopeland:
> they do seem to pop sometimes

Good catch. It took a lot of clicking, but I managed to repro locally. Should be fixed now in latest.

PR #4122, opened 2025-11-09
feat: add a new 'prioritize hardcore stats' preference

wescopeland:
This PR implements a longstanding user feature request. Related: https://github.com/RetroAchievements/RAWeb/pull/2787

A new (disabled by default) setting has been added:
Screenshot 2025-11-09 at 12 39 51 PM

When this setting is enabled, on React game pages, hardcore unlock rates are prioritized and softcore chunks are hidden in the achievement list item progress bars. Other components on the page remain unaffected:
Screenshot 2025-11-09 at 12 40 58 PM

### DIFF
--- a/app/Enums/UserPreference.php
+++ b/app/Enums/UserPreference.php
@@ -48,4 +48,6 @@ abstract class UserPreference
     public const Game_OptOutOfAllSubsets = 18;
 
     public const User_EnableBetaFeatures = 19;
+
+    public const Game_PrioritizeHardcoreStats = 20;
 }
--- a/resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
+++ b/resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
@@ -1,6 +1,7 @@
 import type { FC } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 
+import { calculateUnlockPercentage } from '@/common/utils/calculateUnlockPercentage';
 import { cn } from '@/common/utils/cn';
 import { formatPercentage } from '@/common/utils/l10n/formatPercentage';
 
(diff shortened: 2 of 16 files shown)


resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
Jamiras, line 1: This feature only affects the achievement items. Are there plans to hide softcore in other widgets?

Image

app/Enums/UserPreference.php
```diff
@@ -48,4 +48,6 @@ abstract class UserPreference
     public const Game_OptOutOfAllSubsets = 18;
 
     public const User_EnableBetaFeatures = 19;
+
+    public const Game_PrioritizeHardcoreStats = 20;
```
Jamiras, line 52: I'm still strongly against doing this (see https://github.com/RetroAchievements/RAWeb/discussions/2782#discussioncomment-11075304), but having a toggle allows us to determine how many users have adopted this functionality after some period of time, but also makes it harder to eliminate the feature if it's not popular.

I think just hiding the gray bars is disingenuous. Without having a separate count of pure hardcore players, the hardcore percentage (and therefore the _perceived_ difficulty) is still based off players from both modes.

Oh my gosh! Look how hard this achievement is:

Image

resources/js/common/components/AchievementsListItem/ProgressBarMetaText/ProgressBarMetaText.tsx
Jamiras, line 1: I'd be much happier with a solution that left the gray bar, but updated the text, as I agree that "unlock_count (hardcore_count) of total_count" is not intuitive.

Image

And the feature would just change the text:

Image

Jamiras: And even that seems unnecessarily verbose. Maybe we could just show percentages and have the raw numbers in a tooltip.

Image

.

Image

I might even argue that the progress bar is really the only thing that should be prominently shown, and all of the data should be moved into a tooltip.

The only raw data I'm really interested in is the total number of players for the set (which could be displayed somewhere else). For individual achievements, I'm more interested in the data represented by the progress bar. The full bar (yellow + gray) gives a rough estimate of engagement (total unlock rate vs. total player count), and the ratio of yellow to gray gives a rough estimate of difficulty.

wescopeland: This feedback lands on what the branch is actually trying to solve.

I'm leaning towards closing this PR and opening a new one that goes with your screenshot here:
513449359-8d2c516e-b4be-47a5-b9ab-9af2d4319371

No setting, just a global change.

Jamiras: Sounds good.

Having read through the associated github and forum discussions a couple times now, it seems like the desire is to have a "hardcore-only" view that basically purges _all_ softcore data from view. Hardcore players only want to compare themselves with other hardcore players and the additional data provided by softcore players taints their experience. I don't think there's any way to truly reach that goal without a much broader set of changes - first and foremost being tracking hardcore/softcore player counts separately.

wescopeland: > I don't think there's any way to truly reach that goal without a much broader set of changes

I don't think I'd be in favor of that to be honest, at least not at this time. It's a lot of complexity.

80/20 rule applies here: my gut feeling is the label change on these achievement list items would probably satisfy what most people bothered by the softcore stuff are looking for, with a small minority still indefinitely opposed.

Kaiserlucas: My concern with the proposed change is how useful of a metric the "47.27% hardcore earn rate will actually be in practice. It's going to fluctuate depending on how many softcore players the set has in total, which could lead to it not really solving the issue some people are having with the current system.

If there are 50% softcore players, that number will be near 50% for an early game achievement. If only 25% of players on that set use softcore, then it will be 75% instead. In essence, it could end up being to be more of a reflection of the hardcore versus softcore split on a set than showing the actual difficulty of the achievement for hardcore players only, which is what I assume people actually desire.

I understand that actually differentiating between hardcore players and softcore players is a big increase in complexity, and I don't think it's currently all that necessary to introduce it. In my mind, the current statistics are perfectly serviceable as they are. But for a truly useful hardcore unlock metric to be added, I feel like having a way to separate between hardcore players and softcore players first would be a prerequisite.

I hope this doesn't come across as too negative. I just wanted to give my perspective in case it helps. :prayo:

wescopeland: @Kaiserlucas It's a valid concern. We also track the number of unique hardcore players on a set, so we can use that to derive the hardcore unlock %.

PR #4123, opened 2025-11-10
feat(home): add a new 'Latest Site Updates' sidebar button

wescopeland:
**A migration is included in this PR.** About 50% of this PR's diff is lockfile noise.

---

This PR adds several primitives (markdown rendering, view tracking) which will be useful for future features that, for now, all come together to form a new "Latest Site Updates" button on the homepage sidebar.

We don't have a great way to reach non-Discord users to inform them of what's changing on the site other than using the home page news section. Without a news index, it's tough to see a quick summary of what the recent site updates have been, and many helpful changes we're making are probably flying under the radar. This PR aims to address this problem.

### UX Walkthrough

**[1] An engineer publishes a release notes blurb via [Filament](http://localhost:64000/manage/site-release-notes), written in Markdown.**
This can be a carbon copy of what usually goes out for Discord users. In a future state, we may be able to have RABot automatically populate this from the Discord community news posts, or vice versa. In other words, these announcements should probably be reserved for major releases rather than minor/hotfix releases.
Screenshot 2025-11-10 at 3 40 47 PM

**[2] On save, authenticated users who haven't read the release notes post will see an unread indicator on the home page.**
The indicator only appears if the post is less than or equal to 48 hours old, the user is authenticated, and the user has not clicked the button.
Screenshot 2025-11-10 at 4 19 15 PM

**[3] On click, the indicator disappears, and it does not reappear for the user.**
The release notes are rendered in the React app via Markdown. The user can scroll through a partial history of recent release notes.
Screenshot 2025-11-10 at 4 19 44 PM

---

**Technical Notes:**
View tracking is powered by a `viewables` table and a `HasViews` trait:
```php
class News extends BaseModel {
    use HasViews;
}

$news->markAsViewedBy($user);
if ($news->wasViewedBy($user)) {
    // ... user has viewed this news item ...
}
```

Virtually every element of GitHub-supported markdown is also supported here.
```md
This is another announcement.

![Alt text](https://media.retroachievements.org/Images/087461.png)

**Bold text**. *Italic text.* ~~Strikethru text.~~ [This](https://google.com) is a link.

### h3

#### h4

> This is a blockquote!

1. Foo
2. Bar
3. Baz

* Foo
* Bar
* Baz

Hope this all renders okay in the React app.
```
Screenshot 2025-11-10 at 4 20 11 PM


app/Http/Controllers/HomeController.php
Jamiras, line 1: > [2] On save, authenticated users who haven't read the release notes post will see an unread indicator on the home page.
> The indicator only appears if the post is less than or equal to 48 hours old, the user is authenticated, and the user has not clicked the button.

I'm getting writes to the `viewables` table just loading the home page, so an immediate refresh (without every having clicked the button) causes the indicator to go away.

app/Community/Concerns/HasViews.php
```diff
@@ -0,0 +1,62 @@
...
+    public function markAsViewedBy(User $user): void
+    {
+        $this->views()->firstOrCreate(
+            [
+                'user_id' => $user->id,
+            ],
+            [
+                'viewed_at' => now(),
+            ]
+        );
```
Jamiras, line 52: For news items (especially the site updates), can we just track the most recently viewed news item? Either by updating the viewable_id in the row, or deleting the older rows?

We get roughly 1 news post per day. If 10,000 Users hit the home page every day, that's going to be ~10,000 viewable records created per day, or 300,000 per month just for the news posts.

database/migrations/2025_11_10_000000_create_viewables_table.php
```diff
@@ -0,0 +1,30 @@
...
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::create('viewables', function (Blueprint $table) {
+            $table->id();
+            $table->string('viewable_type');
+            $table->unsignedBigInteger('viewable_id');
+            $table->unsignedBigInteger('user_id');
+            $table->timestamp('viewed_at');
+            $table->timestamps();
```
Jamiras: Does this really need `viewed_at`, `created_at` and `updated_at`? As best as I can tell, the only mutable field is `viewed_at`, which would implicitly also update `updated_at`, and I'm not sure how much merit this is in updating when a user last viewed something.

app/Community/Concerns/HasViews.php
Jamiras: FWIW, it only seems to insert rows for the most recent post. If I create three new site updates, then go to the home page and click the button, only the most recent post gets a row:
```
+-------------+---------------------+
| viewable_id | viewed_at           |
+-------------+---------------------+
|        2136 | 2025-11-15 14:01:05 |
|        2137 | 2025-11-15 14:03:30 |
|        2138 | 2025-11-15 14:10:13 |
|        2141 | 2025-11-15 14:21:06 |
+-------------+---------------------+
```

app/Http/Actions/BuildSiteReleaseNotesAction.php
```diff
@@ -0,0 +1,29 @@
...
+use Illuminate\Support\Collection;
+
+class BuildSiteReleaseNotesAction
+{
+    /**
+     * Fetch the site release notes to display in the Latest Site Updates dialog.
+     *
+     * @return Collection<int, NewsData>
+     */
+    public function execute(int $limit = 5): Collection
```
Jamiras, line 19: Do you plan to have some way to view more than the five most recent updates?

I realize this limitation also exists for the nine most recent posts on the home page, which will cycle out faster.

app/Http/Controllers/HomeController.php
wescopeland: Yep, this is a bug I didn't catch. Fixed in latest.

app/Community/Concerns/HasViews.php
wescopeland: Updated the trait in latest to support this behavior. In the _News.php_ model file, we now have:
```php
public function shouldTrackLatestViewOnly(): bool
{
    return $this->category === NewsCategory::SiteReleaseNotes;
}
```

This mode defaults to `false`. When enabled, on tracking a view, all previous rows for the user+entity will be deleted.

database/migrations/2025_11_10_000000_create_viewables_table.php
wescopeland: Removed `$table->timestamps()` in latest.

app/Http/Actions/BuildSiteReleaseNotesAction.php
wescopeland: It's plausible at some point we'd want to allow users to see more the latest 5. It's not on my near-term radar though to develop this feature.

app/Community/Concerns/HasViews.php
```diff
@@ -0,0 +1,92 @@
...
+    public function markLatestAsViewedBy(User $user): void
+    {
+        Viewable::where('user_id', $user->id)
+            ->where('viewable_type', $this->getMorphClass())
+            ->delete();
+
+        $this->views()->create([
+            'user_id' => $user->id,
+            'viewed_at' => now(),
+        ]);
```
Jamiras: Is there some way to do a createOrUpdate here to recycle the row?

app/Filament/Resources/SiteReleaseNotesResource/Pages/Create.php
Jamiras, line 1: If I Save a new record, the editor remains on the edit page. If I try to cancel out of that to return to the list, it goes back to the create page, and hitting Create there tells me some of the fields aren't set.

Cancel from a saved record should go back to the list.

Potentially, just saving a record should take me back to the list.

app/Community/Concerns/HasViews.php
wescopeland: I think it's probably impractical.

I agree it's desirable to recycle the row, but as best as I can tell, `updateOrCreate` can't handle a situation like "find by X and update Y where Y is part of [given search criteria]".

app/Filament/Resources/SiteReleaseNotesResource/Pages/Create.php
wescopeland: Fixed in latest.

Saving a new record and editing an existing record will now return you to the list.

app/Community/Concerns/HasViews.php
Jamiras: How is Y part of the given search criteria? 

From the [docs](https://laravel.com/docs/12.x/eloquent#updates), it looks like this should work:
```
        Viewable::updateOrCreate(
            [
                'viewable_type' => $this->getMorphClass(),
                'viewable_id' => $this->id,
                'user_id' => $user->id,
            ],
            [
                'viewed_at' => now(),
            ]
        );
```

It doesn't - it creates duplicate rows, but maybe I didn't do it quite right?

wescopeland: Got it to work in latest:
```php
Viewable::updateOrCreate(
    [
        'viewable_type' => $this->getMorphClass(),
        'user_id' => $user->id,
    ],
    [
        'viewable_id' => $this->id,
        'viewed_at' => now(),
    ]
);
```

This finds the user's existing record for the morph type and updates it to point to the new item.

Jamiras: Ah, that makes sense.

PR #4125, opened 2025-11-10
refactor: change how moderation reports are modeled/stored in the DB

wescopeland:
This PR changes how user moderation reports are stored in the database in order to facilitate future functionality.

Currently, the only references to reports in the DB are stored in `discord_message_thread_mappings` via `reportable_type` and `reportable_id` columns.

This PR drops those two columns and creates a `user_moderation_reports` table instead. The two columns are then replaced with a foreign key to that new table.

The focus of this branch's changes is purely centered on shifting where the data is stored so reports are a first-class citizen in the database. There should be no functional changes as a result of this refactor.


database/migrations/2025_11_10_000001_create_user_moderation_reports_table.php
```diff
@@ -0,0 +1,55 @@
...
+            $table->foreign('reporter_user_id')->references('ID')->on('UserAccounts')->onDelete('cascade');
+            $table->foreign('reported_user_id')->references('ID')->on('UserAccounts')->onDelete('set null');
+            $table->foreign('message_thread_id')->references('id')->on('message_threads')->onDelete('cascade');
+        });
+
+        Schema::table('discord_message_thread_mappings', function (Blueprint $table) {
+            $table->dropIndex('idx_reportable_type_id');
+            $table->dropColumn(['reportable_type', 'reportable_id']);
+
+            $table->unsignedBigInteger('moderation_report_id')->nullable()->after('discord_thread_id');
```
Jamiras: This is effectively denormalized data. Do we want it?

discord_message_thread_mappings.moderation_report_id = user_moderation_reports.id where user_moderation_reports.message_thread_id=discord_message_thread_mappings.message_thread_id

Jamiras: As best as I can tell, it looks like it's only used by the `moderationReport` helper function, and that doesn't seem to be used.

app/Models/DiscordMessageThreadMapping.php
```diff
@@ -68,30 +71,31 @@ public static function storeMapping(int $messageThreadId, string $discordThreadI
...
+        return $report?->discordThreadMapping;
     }
 
     /**
-     * Store a new Discord thread mapping for a reportable item.
+     * Store a new Discord thread mapping for a moderation report.
      */
     public static function storeReportMapping(
-        DiscordReportableType $reportableType,
-        int $reportableId,
```
Jamiras, line 83: `moderationReportId` is going to be unique, so this should never actually do an update.

Was the old implementation trying to maintain a single discord thread regardless of how many unique users reported it?

app/Models/DiscordMessageThreadMapping.php
```diff
@@ -68,30 +71,31 @@ public static function storeMapping(int $messageThreadId, string $discordThreadI
...
      * Find an existing Discord thread mapping for a reportable item.
      */
-    public static function findReportMapping(DiscordReportableType $reportableType, int $reportableId): ?self
+    public static function findReportMapping(ModerationReportableType $reportableType, int $reportableId): ?self
     {
-        return self::where('reportable_type', $reportableType->value)
+        $report = UserModerationReport::where('reportable_type', $reportableType->value)
             ->where('reportable_id', $reportableId)
+            ->whereHas('discordThreadMapping')
             ->first();
```
Jamiras: What if there's more than one?

database/migrations/2025_11_10_000001_create_user_moderation_reports_table.php
wescopeland: Removed in latest.

app/Models/DiscordMessageThreadMapping.php
wescopeland: Yes, we should indeed maintain [a single Discord thread per reportable item](https://discord.com/channels/476211979464343552/1441189751364522139).

However, in latest this method is no longer needed and has been removed.

app/Models/DiscordMessageThreadMapping.php
wescopeland: Made this more robust in latest.

PR #4130, opened 2025-11-13
(perf) batch fetch shortcode embedded resources

Jamiras:
Addresses:
* https://retroachievementsorg.sentry.io/issues/44908979 (N+1 query viewing recent posts)
* https://retroachievementsorg.sentry.io/issues/48961784 (N+1 query viewing recent posts)
* https://retroachievementsorg.sentry.io/issues/45361091 (N+1 query viewing recent posts of user)
* https://retroachievementsorg.sentry.io/issues/72309844 (N+1 query viewing recent posts of user)

Also addresses the same issue for recent posts on the main forum page and home page, but I couldn't find specific sentry issues for either.

### DIFF
--- a/app/Connect/Actions/BuildClientPatchDataAction.php
+++ b/app/Connect/Actions/BuildClientPatchDataAction.php
@@ -63,18 +63,20 @@ class BuildClientPatchDataAction
         }
 
         // Use the game from the hash for legacy clients.
-        return $this->buildPatchData($gameHash->game, $user, $flag);
+        return $this->buildPatchData($gameHash->game, $user, $flag, $gameHash->compatibility);
     }
 
     /**
      * @param Game $game The game to build root-level data for
      * @param User|null $user The current user requesting the patch data (for player count calculations)
      * @param AchievementFlag|null $flag Optional flag to filter the achievements by (eg: only official achievements)
+     * @param GameHashCompatibility $compatibility Indicates the compatibility of the hash being loaded (affects game title)
      */
     private function buildPatchData(
         Game $game,
         ?User $user,
         ?AchievementFlag $flag,
+        GameHashCompatibility $compatibility = GameHashCompatibility::Compatible,
     ): array {
         $gamePlayerCount = $this->calculateGamePlayerCount($game, $user);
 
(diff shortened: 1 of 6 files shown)


app/Community/Controllers/ForumTopicController.php
```diff
@@ -59,7 +60,7 @@ public function create(ForumCategory $category, Forum $forum, Request $request):
     public function show(ForumTopic $topic, ShowForumTopicRequest $request): InertiaResponse|RedirectResponse
     {
         $this->authorize('view', $topic);
-        abort_if($topic->comments->isEmpty(), 404);
+        abort_if(!ForumTopicComment::where('forum_topic_id', $topic->id)->exists(), 404);
```
Jamiras, line 63: This is indirectly related. The "comments" helper function queries all comments with their associated user data just to determine if records exist.

Image

app/Support/Shortcode/Shortcode.php
```diff
@@ -204,18 +215,96 @@ public static function extractShortcodeIds(string $input): array
...
+                        });
+                    User::withTrashed()
+                        ->where(function ($query) use ($ids) {
+                            $query->whereIn('User', $ids)
+                                ->orWhereIn('display_name', $ids);
+                        })
+                        ->get()->map(function ($user) use ($key) {
+                            $results[$key][$user->display_name] = $user;
+                            $results[$key][$user->User] = $user;
+                        });
```
wescopeland: I'm wondering if there may be a subtle bug here. I think `map()` is creating a new collection and not modifying `$results[$key]` in place.

Screenshot 2025-11-15 at 9 52 16 AM

I think we may actually want:
```php
->get()->each(function ($user) use ($key, &$results) {
    $results[$key][$user->display_name] = $user;
    $results[$key][$user->User] = $user;
});
```

Jamiras: Good catch. I've made the prescribed change.

Although, I'm not certain how pertinent this is. Almost all `[user=x]` links in the database are normalized to IDs. Those that aren't probably predate the ability to change display names.

I've also modified the logic to prefer an ID match over a name match (for example [user=3636] should match the user with ID 3636, not the user with name 3636).

PR #4131, opened 2025-11-13
(feat) update game title for compatibility testing

Jamiras:
addresses https://discord.com/channels/476211979464343552/1002695123030773790/1436505387460333648

QATesters are automatically opted in to "testing" mode for unsupported hashes. Previously, there was no indication that the tester was loading an unsupported hash, leading to time lost trying to earn achievements on unsupported hashes.

This updates the game title returned with an unsupported hash to clearly indicate the hash is unsupported:
image
(Previously just returned the normal game title)

The title is also updated for non-testers, but they only receive the warning achievement:
image
(Previously just returned the "Unsupported Game Version")


app/Connect/Actions/BuildClientPatchDataAction.php
```diff
@@ -63,18 +63,20 @@ public function execute(
...
         // Use the game from the hash for legacy clients.
-        return $this->buildPatchData($gameHash->game, $user, $flag);
+        return $this->buildPatchData($gameHash->game, $user, $flag, $gameHash->compatibility);
     }
 
     /**
      * @param Game $game The game to build root-level data for
      * @param User|null $user The current user requesting the patch data (for player count calculations)
      * @param AchievementFlag|null $flag Optional flag to filter the achievements by (eg: only official achievements)
+     * @param GameHashCompatibility $compatibility Optional flag to modify the title to reflect the hash compatibility
```
wescopeland: Docs, nit: "Optional" may be potentially misleading because there is a default value that is always used.

Potential alternative wording (I don't feel too strongly about this, if you want to change it I'm sure you'll land on something better):
> The compatibility status of the hash, used to modify the game title for non-compatible hashes (defaults to Compatible)

wescopeland (approved):

app/Connect/Actions/BuildClientPatchDataAction.php
Jamiras: I've changed it to:

> Indicates the compatibility of the hash being loaded (affects game title)

PR #4135, opened 2025-11-14
feat(mail): add support for RFC 8058 one-click unsubscribes

wescopeland:
Google and Yahoo are displeased with us because we don't support [RFC 8058 headers](https://datatracker.ietf.org/doc/html/rfc8058) in several of our email comms.

These headers are what facilitate the one-click "Unsubscribe" buttons to appear in email clients:
Screenshot 2025-11-13 at 7 46 39 PM

Email clients need to be able to send a POST request to your unsubscribe route with some kind of metadata directing the server what is being unsubscribed from. All that metadata is defined via headers attached to the email.


app/Community/Controllers/UnsubscribeController.php
```diff
@@ -18,8 +19,33 @@ public function __construct(
...
     }
 
-    public function show(Request $request, string $token): InertiaResponse
+    public function show(Request $request, string $token): InertiaResponse|Response
     {
+        // Invalid signatures are handled automatically by the 'signed' middleware (returns 403).
+        // If we reach this point, the signature is valid.
+
+        // Handle POST requests for RFC 8058 one-click unsubscribe from email clients.
+        if ($request->isMethod('post')) {
```
Jamiras, line 28: I can verify that the URL in the Unsubscribe header (as exposed by mailpit) does Unsubscribe the user, but I'm having trouble simulating the POST to actually exercise this bit of code. I'm going to assume it's working based on the unit tests.

PR #4139, opened 2025-11-17
feat(manage): improve game set editing ux

wescopeland:
This PR improves the editing UX of attached game sets. Set title and type are now editable inline. Previously it required a detach and reattach.

Additionally, the title label now reflects how it shows in the player-facing UI.

**Before**
Screenshot 2025-11-17 at 4 42 48 PM

**After**

Screenshot 2025-11-17 at 4 41 47 PM

Screenshot 2025-11-17 at 4 42 07 PM


app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
Jamiras, line 1: The modified title 

Image

Image

Isn't reflected in the [Edit Subset Game Details] page.

Image

Unrelated: the permalink for the subset still points to a game page, not the base game page with a set parameter.

Image

wescopeland: > Curious why you opted to have this be inline editing instead of a popup like we do for Releases (and you still have a popup for changing the type).

I figured given the Display Order was already editable inline, it'd be consistent with the existing editing UX for the rows. I originally had type editable inline without a dialog too, but decided that editing the type could actually be a destructive action, so it needed to be behind (at the very least) a confirm.

> The modified title Isn't reflected in the [Edit Subset Game Details] page.

I'm not sure the best way to handle this. They're technically two different titles used in two different places of our API - the set title and the backing game title. We can certainly update the backing game title automatically, but something about it doesn't seem right. For example, it also implies updating the backing game title should update the set title (two-way binding), but that may be fragile if the set is attached to multiple games 😵‍💫  I'm almost afraid right now to introduce this two-way binding layer between set title and backing game title.

Jamiras: > I figured given the Display Order was already editable inline, it'd be consistent with the existing editing UX for the rows.

I think everything should be moved to a popup diaog. I don't there's anywhere else in the management UI that allows inline editing currently. I'm also fond of having the ability to cancel out if I have second thoughts while making the changes.

Jamiras: >They're technically two different titles used in two different places of our API

Searching still uses the legacy game title. We don't index the set names.

Additionally, subset links get normalized to `[game=XXXX]` shortcode, and that expands using the legacy game title.

wescopeland: Made a bunch of adjustments in latest:
* Editing the title and type are now both in a dialog.
* There's now two-way binding between set titles and backing game titles.
* If the user can manage the `AchievementSets` resource, a link to access the set details page appears. This should work in tandem with https://github.com/RetroAchievements/RAWeb/pull/4142's changes.

PR #4142, opened 2025-11-17
chore(multiset): restrict the 'Sets' resource in Filament

wescopeland:
Accompanies https://github.com/RetroAchievements/RAWeb/pull/4139.

Locks down the "Sets" resource (in the sidebar) in Filament to just users who have `Role::GAME_HASH_MANAGER`.

I've thought this over, and I don't think we have enough data to make this tool generally available to developers yet. If it isn't used wisely, it can cause more harm than good, and using it should really be an edge case.

If updating the hash compatibility matrix for a game's sets becomes a common ask and we want to make it generally available to full devs, I think we should come up with an easier UX. Otherwise, for now, we can restrict it to a small subset of users.


app/Policies/AchievementSetPolicy.php
```diff
@@ -37,32 +34,27 @@ public function create(User $user): bool
...
-            Role::DEVELOPER,
-            Role::DEVELOPER_JUNIOR,
         ]);
     }
 
     public function update(User $user): bool
     {
         return $user->hasAnyRole([
             Role::GAME_HASH_MANAGER,
-            Role::DEVELOPER,
```
Jamiras, line 49: As a developer without hash manager permissions, I can still click on a row in the Game's Set list, and I get a forbidden error. I expect the page should open readonly as `view` returns true. But if it's not supposed to open, then the click shouldn't do anything.

wescopeland: Disabled the row links in latest. The title labels still have an underline on hover which Filament doesn't seem to give me any way to get rid of, but this will naturally go away as a result of #4139.

PR #4143, opened 2025-11-17
fix(multiset): add a game_id,title unique constraint

wescopeland:
**Pre-deploy SQL as a sanity check:**
```sql
-- should return 0 records in prod
SELECT t1.*
FROM game_achievement_sets t1
INNER JOIN game_achievement_sets t2
WHERE t1.id > t2.id
  AND t1.game_id = t2.game_id
  AND t1.title = t2.title
  AND t1.title IS NOT NULL;

```

---

This PR adds a unique constraint to `game_achievement_sets` between `game_id` and `title`. This will plausibly solve a potential downstream issues with clients/emulators as they try to show multiple sets per game.

### DIFF
--- a/app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
+++ b/app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
@@ -8,6 +8,7 @@ use App\Data\ForumTopicData;
 use App\Data\PaginatedData;
 use App\Enums\Permissions;
 use App\Models\ForumTopic;
+use App\Support\Shortcode\Shortcode;
 use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Facades\DB;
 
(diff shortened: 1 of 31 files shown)


app/Helpers/database/player-game.php
```diff
@@ -258,6 +258,7 @@ function getUserAchievementUnlocksForGame(User|string $user, int $gameID, Achiev
         ->join('game_achievement_sets', 'game_achievement_sets.achievement_set_id', '=', 'achievement_sets.id')
         ->where('game_achievement_sets.game_id', $gameID)
         ->where('Flags', $flag->value)
+        ->orderBy('player_achievements.achievement_id')
```
wescopeland, line 261: Adding the constraint seems to have modified the SQL execution plan for this query (ultimately causing a test to fail). I've added this `orderBy` to fix the test and make the returned achievement IDs deterministic.

The query is already fast, though interestingly in my local benchmarks adding this `orderBy` seems to cause the query to execute even faster.

PR #4147, opened 2025-11-18
fix(game): persist 'Locked Only' and 'Missable Only' filters on ?set= pages too

wescopeland:
This PR fixes a bug on React game pages where the "Locked Only" and "Missable Only" filters are not persisting during a refresh when viewing a subset.

This is easy to repro by setting any achievement in a subset to be missable, enabling "Missable Only" on the game page when viewing the set (with a `?set=` param present), then refreshing the game page.


app/Platform/Actions/BuildGameShowPagePropsAction.php
Jamiras, line 1: There's a weird sort of interaction between sets on a game.

1. Open game page
2. Mark Missable filter
3. Switch to subset (see missable filter still selected)
4. Refresh (missable filter cleared)
5. Switch back to core set (see missable filter cleared)
6. Refresh (missable filter restored)

wescopeland: Adjusted which game ID we're persisting in latest - I think it should resolve this issue.

Jamiras: The setting is now consistent across all subsets within a game (changing on any set affects all sets). That's different than the previous behavior where each set could remember separately.

PR #4148, opened 2025-11-19
feat(GameHeaderSlotContent): make Want to Play press interaction consistent with mobile

wescopeland:
This PR makes the "Want to Play" button on React game pages (desktop) consistent with mobile.

On mobile, the check has a green icon and rotates in/out from the plus.

https://github.com/user-attachments/assets/e45d22b3-bbf6-42aa-a250-0def447b91d0


wescopeland:
> Would you want to apply this change to the "Want to Develop/Revise" button?

I don't think so. The feedback driving this change is a user reported it was barely perceptible on desktop whether a game is on their list or not.

Agreed the green does stick out a little bit. I think this might be more subdued once we start making some tweaks to the header UX up here for banner images.

PR #4153, opened 2025-11-19
add separate low-priority queue for scheduled emails

Jamiras:
So the sudden influx of annual recap emails don't starve other queues.

Could also be leveraged for daily digest emails in the future.


config/horizon.php
```diff
@@ -316,6 +316,25 @@
...
+            ],
+            'balance' => 'simple',
+            'maxProcesses' => 2, // low priority - don't starve other processes
+            'maxTime' => 0,
+            'maxJobs' => 0,
+            'memory' => 128,
+            'tries' => 1,
+            'timeout' => 300, // NOTE timeout should always be at least several seconds shorter than the queue config's retry_after configuration value.
+            'nice' => 5, // low priority
+        ],
```
Jamiras, line 341: This is basically a copy of `supervisor-5` with a `maxProcesses` of 2 and a `nice` of 5.

config/horizon.php
wescopeland, line 1: Docs: we should also update the block comment from lines 190-203.

Line 198 (total workers) needs to be updated, and a new line needs to be added under 203 for supervisor-6.

config/horizon.php
```diff
@@ -316,6 +316,25 @@
...
+         * Daily email supervisor - handles mass email distribution.
+         * Isolated to allow non-time-critial emails to be sent without affecting other queues.
+         */
+        'supervisor-6' => [
+            'connection' => 'redis',
+            'queue' => [
+                'summary-emails',
+            ],
+            'balance' => 'simple',
+            'maxProcesses' => 2, // low priority - don't starve other processes
```
wescopeland: Perf: after doing some experimentation, I can confirm with `balance` set to "simple", `maxProcesses` also serves as a "min processes". In other words, when no work is being done, 2 processes will always sit and constantly poll the queue.

I think we can get the desired scaling effect by doing:
```php
'balance' => 'auto',
'autoScalingStrategy' => 'size',
'maxProcesses' => 2,
'balanceMaxShift' => 1,
'balanceCooldown' => 3,
```

With this change, processes will scale down to 1 minimum when idle and scale up to 2 based on queue size.

config/horizon.php
Jamiras: updated

config/horizon.php
Jamiras: applied suggested changes.

PR #4159, opened 2025-11-22
fix(SetRarityLabel): remediate tooltip word wrapping bug

Chew:
Fixes the word wrapping in the rarity label tooltip. If there's a better way, please let me know!

Before:
CleanShot 2025-11-21 at 19 40 35@2x

After:
CleanShot 2025-11-21 at 19 40 46@2x

### DIFF
--- a/app/Community/Actions/AddToMessageThreadAction.php
+++ b/app/Community/Actions/AddToMessageThreadAction.php
@@ -4,11 +4,12 @@ declare(strict_types=1);
 
 namespace App\Community\Actions;
 
-use App\Community\Enums\DiscordReportableType;
+use App\Community\Enums\ModerationReportableType;
 use App\Community\Events\MessageCreated;
 use App\Models\Message;
 use App\Models\MessageThread;
 use App\Models\User;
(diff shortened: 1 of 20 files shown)


resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/SetRarityLabel/SetRarityLabel.tsx
```diff
@@ -33,13 +33,15 @@ export const SetRarityLabel: FC<SetRarityLabelProps> = ({ pointsTotal, pointsWei
         <p className="font-bold">{t('Rarity')}</p>
 
         <div className="flex flex-col gap-2">
```
wescopeland: Verified this change works. Idea: can we lift `whitespace-normal` up to this div?

Chew: Yep, that seems to work as well!

PR #4161, opened 2025-11-23
feat(game): support uploadable hi-res banner images

wescopeland:
This PR adds support for a new type of game media: banner images. In so doing, it also adds support for `spatie/laravel-medialibrary`, which was originally planned in V2.

The custom banner images only appear on mobile currently. The desktop UI/UX is complex enough that it justifies its own PR.

> [!IMPORTANT]  
> You'll need to follow the "How to test this PR" section closely. These images are _only_ stored in S3, which requires MinIO to simulate.

## How it works
A new file input is now available on game page management:
Screenshot 2025-11-23 at 2 08 03 PM

We really want people to upload 32:9 banners here. We gracefully allow for a little bit of tolerance - the back-end will do some transformations to make stuff within tolerance workable.

Here's a sample image you can try with:
[ff7_hero.png.zip](https://github.com/user-attachments/files/23698062/ff7_hero.png.zip)

On upload, the file is stored in the temporary disk until the user clicks "Save changes":
Screenshot 2025-11-23 at 2 10 55 PM

Upon clicking "Save changes", the file is uploaded to S3 and numerous transformations are done on it. These transformations are defined in _Game.php_. They are:
* We scale the image to 5 pre-defined sizes: `mobile-sm`, `mobile-md`, `desktop-md`, `desktop-lg`, and `desktop-xl`.
* We convert all of these scalings to [WebP](https://web.dev/learn/images/webp) and [AVIF](https://web.dev/learn/images/avif).
* We create mobile and desktop blur-up placeholder images that are extremely small and load fast.

All this stuff gets dumped into a folder on S3 that looks something like this:
Screenshot 2025-11-23 at 2 14 26 PM

> [!NOTE]  
> Unlike our other assets, we don't write any of these files to local disk on save. We only have a few hundred GB of space to work with on the local disk. S3 storage is orders of magnitude less expensive.

Now, as a mobile user, you can navigate to the game page and see your new custom banner image:
Screenshot 2025-11-23 at 2 20 25 PM

You will notice there's one minor difference - if the game has a custom banner image, we now use a placeholder blur-in:

https://github.com/user-attachments/assets/610477e6-359c-4517-86d2-e5a925391ceb

Because these images are going to be higher quality than in-game screenshots, **placeholder blur-in is highly desirable to maintain good [LCP](https://web.dev/articles/lcp) scores and healthy Core Web Vitals.** It happens on every refresh. We don't have any way of knowing if the user has the high-quality image cached yet. This sort of placeholder blur-in for large LCPs is normal, see: https://psnprofiles.com/trophies/32271-age-of-mythology-retold

## How to test this PR
I'm assuming you're running Sail.

```
# be sure horizon is running, restart it if necessary
# image processing happens in background jobs
sail artisan horizon
```

In your _.env_, uncomment these four lines:
```env
AWS_BUCKET=local
AWS_DEFAULT_REGION=eu-west-1
```
Add this line:
```env
AWS_URL=http://localhost:${FORWARD_MINIO_PORT}/local
```

You'll need to set up a bucket in MinIO to simulate our AWS S3 bucket:
1. Navigate to http://localhost:64041/login.
2. The username and password should both be "retroachievements". Log in.
3. Click the "Create a Bucket" link.
4. **Use "local" (without quotes) as the bucket name** and click the "Create Bucket" button.
5. You should now see a giant list item for the local bucket. Click on it to open it.
6. In the Summary section, click the little pencil icon next to "Access Policy". Set it to "Public".

You're now good to upload banners for games in Filament. Remember: you have to click the "Save Changes" button for them to persist.

It's best to test the React UI while SSR is enabled.
```
pnpm build && sail artisan inertia:start-ssr
```

## Future work
We'll need at least two more PRs from here to finish the feature.
1. Desktop UI/UX.
2. Allow granting banner art credit in Filament (https://github.com/wescopeland/RAWeb/tree/banner-credit - going to wait on this PR first because I'd like to add some validation to this credit that depends on the new `banner` field).

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2039,6 +2039,43 @@
             "time": "2025-12-05T14:53:55+00:00"
         },
         {
+            "name": "filament/spatie-laravel-media-library-plugin",
+            "version": "v4.3.1",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/filamentphp/spatie-laravel-media-library-plugin.git",
+                "reference": "73748df28a9c2e8c34d2c02f9314c330602e1830"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/filamentphp/spatie-laravel-media-library-plugin/zipball/73748df28a9c2e8c34d2c02f9314c330602e1830",
+                "reference": "73748df28a9c2e8c34d2c02f9314c330602e1830",
+                "shasum": ""
+            },
+            "require": {
+                "filament/support": "self.version",
+                "php": "^8.2",
+                "spatie/laravel-medialibrary": "^11.0"
+            },
+            "type": "library",
+            "autoload": {
+                "psr-4": {
+                    "Filament\\": "src"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "description": "Filament support for `spatie/laravel-medialibrary`.",
+            "homepage": "https://github.com/filamentphp/filament",
+            "support": {
+                "issues": "https://github.com/filamentphp/filament/issues",
+                "source": "https://github.com/filamentphp/filament"
+            },
+            "time": "2025-12-09T09:54:02+00:00"
+        },
+        {
             "name": "filament/support",
             "version": "v4.3.1",
             "source": {


app/Filament/Resources/GameResource/Pages/Edit.php
```diff
@@ -46,6 +49,38 @@ protected function mutateFormDataBeforeSave(array $data): array
...
+    protected function afterSave(): void
+    {
+        /** @var Game $game */
+        $game = $this->record;
+        $banner = $game->getFirstMedia('banner');
+
+        // Extract and store edge colors.
+        if ($banner && !$banner->getCustomProperty('left_edge_color')) {
+            try {
+                $action = new ExtractBannerEdgeColorsAction();
```
wescopeland, line 84: This is for future desktop functionality.

MediaLibrary allows us to save any custom properties we want to an uploaded image (you'll find all this stuff in the `media` table). Here's an example:
Screenshot 2025-11-23 at 2 41 23 PM

For desktop, the screen width is unbounded. Users on ultrawide displays will see the edges of the banner. The make those edges look better, we'll add a subtle gradient whose colors are dynamically generated/calculated from the banner itself.

PSNProfiles does something similar:
Screenshot 2025-11-23 at 2 45 18 PM

In this PR, these two color values are stored, but unused. They'll only be used in the desktop UI/UX.

config/media.php
```diff
@@ -27,19 +27,83 @@
...
+     * - game banners remain on singleFile() for S3 disk efficiency
+     * - user profile banners are stable independent snapshots
+     */
+        // 'profile_banner' => [
+        //     'mobile-sm' => [...], // same sizes as game.banner
+        //     'mobile-md' => [...],
+        //     'desktop-md' => [...],
+        //     'desktop-lg' => [...],
+        //     'desktop-xl' => [...],
+        // ],
```
wescopeland: This is documenting an idea for future functionality.

I think it would be neat if, once user profiles are migrated to React, users could (optionally) set their profile banner image to the same banner image of a game they've beaten. This has an added benefit of further incentivizing beaten awards and making profiles a bit more aesthetically pleasing.

We can figure out the technical details later - I wanted to dump all the thoughts from my brain in here before I lost them.

resources/js/pages/game/[game].tsx
```diff
@@ -58,6 +60,14 @@ const GameShow: AppPage = () => {
         ogImage={backingGame!.badgeUrl}
       />
 
+      {/* TODO desktop preload image */}
+      {ziggy.device === 'mobile' && (game.banner?.mobileSmAvif || game.imageIngameUrl) ? (
+        <SEOPreloadImage
+          src={game.banner?.mobileSmAvif ?? (game.imageIngameUrl as string)}
+          type={game.banner?.mobileSmAvif ? 'image/avif' : 'image/png'}
+        />
```
Comment (wescopeland) [inline resources/js/pages/game/[game].tsx:71]: To load the banner image as fast as possible, we communicate all sorts of hints to the browser that it needs to prioritize it.

app/Models/Game.php
```diff
@@ -299,6 +300,65 @@ public function registerMediaCollections(): void
...
+                    $height = config("media.game.banner.{$size}.height");
+
+                    // Use Fit::Crop for all sizes to avoid letterboxing.
+                    // Uploaded images may vary slightly in aspect ratio (SteamGridDB, etc).
+                    // Cropping ensures no white bars are added to the image while maintaining our target dimensions.
+                    /**
+                     * WebP conversion.
+                     * @see https://web.dev/learn/images/webp
+                     */
+                    $this->addMediaConversion("{$size}-webp")
```
wescopeland, line 321: This causes the extension to also be within the filename, eg:
```
file-webp.webp
```
For MediaLibrary stuff, this seems to be the common convention.

config/media.php
```diff
@@ -27,19 +27,83 @@
...
-     * user for games
+    /**
+     * used for game-related media
      * TODO: aspect ratio should be defined by system
      */
     'game' => [
+        /**
+         * banner images
+         * game editors upload 32:9 source images
+         * mobile uses 16:9 crops, desktop uses full 32:9
```
Jamiras, line 47: I'm really having trouble picturing how this would look on desktop. With the page "fixed" at 840px wide, the banner would be 234px tall. That's roughly this:

Image

Would the header components be moved around? What about pages without a custom banner? Would that still allocate a 234px header and just leave it blank? Or would the in-game image be scaled up like it is on mobile?

config/media.php
Jamiras: It doesn't look like we're deleting any files when changing the banner image. This is consistent with other image assets. In that case, we just need some way to capture the image path for the user instead of copying the entire file.

resources/js/features/games/components/GameMobileHeader/GameMobileBannerImage/GameMobileBannerImage.tsx
```diff
@@ -0,0 +1,113 @@
...
+                }}
+                onLoad={() => setIsImageLoaded(true)}
+                fetchPriority="high"
+                loading="eager"
+                decoding="async"
+              />
+            </picture>
+          </>
+        ) : (
+          // If the game has no custom banner set, fall back to ImageIngame.
```
Jamiras, line 78: I imagine most games won't have a custom banner as they won't have a big enough fandom to warrant the additional art. Are we expecting a lot of games just to provide upscaled in-game images or boxart crops for their backgrounds?

In that case, I would expect this upscaling behavior to be duplicated for desktop. I know you've said before that you don't want to.

app/Filament/Resources/GameResource/Pages/Edit.php
Jamiras: > For desktop, the screen width is unbounded.

It seems like the screen width is still capped at something like 1280px. At XL, I have unused space on both sides of my game page.

Jamiras (commented):
I'm really struggling to do much with this without the desktop support. My mobile simulator only runs in XS, so it's impractical to validate the image at other sizes.

app/Filament/Resources/GameResource/Pages/Edit.php
wescopeland: Yep, it's pretty much meant to handle this kind of situation (notice the dynamic colors on the sides of the banner, they stretch infinitely):
Screenshot 2025-12-05 at 5 48 57 PM

My screen resolution is 3840x2160. The page's container width is capped at ~1280px, but the actual screen width is unbounded.

Jamiras: Yea, as we briefly discussed in DMs - I wasn't expecting the image to extend beyond the container.

config/media.php
wescopeland: Based on our most recent discussion, I don't think I'm prepared to say concretely yet what this will look like on desktop. Let's tackle that in a separate PR. We may have to try a few things until we land somewhere we're both happy with.

resources/js/features/games/components/GameMobileHeader/GameMobileBannerImage/GameMobileBannerImage.tsx
wescopeland: I would imagine we're going to have a weird in-between period where most images aren't upscaled for quite some time. However, my sense is the new game page is going to incentivize people to upscale stuff, and a next logical step will be to support higher-quality media using medialibrary. This PR serves as a bit of precursor to that just to prove out we can handle medialibrary functionality at all.

config/media.php
wescopeland: Removed this whole blurb in latest.

resources/js/features/games/components/GameMobileHeader/GameMobileBannerImage/GameMobileBannerImage.tsx
Jamiras: > we're going to have a weird in-between period where most images aren't upscaled

> my sense is the new game page is going to incentivize people to upscale stuff

It's something we can discuss more in the desktop support PR, but my concern is that we should still provide a fallback on desktop. I know you don't want to do upscale the in-game image, so what would the fallback be?

Jamiras (approved):
Approved as a service for uploading an images and storing it at multiple sizes. Still have my concerns about how it's going to be leveraged on desktop.

resources/js/features/games/components/GameMobileHeader/GameMobileBannerImage/GameMobileBannerImage.tsx
wescopeland: That will probably have to be the fallback. We'll just have to be clever in how we do it to make sure the upscaling isn't overly-obvious.

wescopeland: > We'll just have to be clever in how we do it to make sure the upscaling isn't overly-obvious.

For example: in my original idea for desktop banners, the way I achieved this was making the banner significantly less prominent if it was an upscaled fallback.

PR #4164, opened 2025-11-24
feat: add the foundation for OAuth2 with Laravel Passport

m1guelpf:
[How to test this PR](https://github.com/RetroAchievements/RAWeb/pull/4164#issuecomment-3578059516)

Adds OAuth2 authentication using Laravel Passport!

Current state of things:
- The end-to-end flow works!
- The end-to-end flow for device authentication works!
- There is no UI to create clients, use `php artisan passport:client` and `php artisan passport:client --device` for testing.
- There is no UI to revoke authorizations
- All the routes are disabled in production

_edited by @wescopeland to update the current state of things and directly link to my comment_


wescopeland:
I collaborated heavily on this PR.

**How to test this PR**
```bash
sail artisan migrate
sail artisan passport:keys # this is already done on stage and prod
```

There are two authorization flows you can test-drive. The first is a standard client authorization flow where some app wants to log in on behalf of a user:
```
# Create a test OAuth client
# name: "SampleApp"
# redirect: [default] "http://localhost:64000/auth/callback"
# device auth? "no"
sail artisan passport:client
```
You'll receive a Client ID and Client Secret.

Navigate to:
http://localhost:64000/oauth/authorize?client_id={CLIENT_ID}&redirect_uri=http://localhost:64000/auth/callback&response_type=code

Screenshot 2025-11-25 at 6 36 42 PM

In the real world, clicking Authorize and Reject will direct back to the consumer app. In our demo, we have an `/auth/callback` dummy route set up to simulate an end-to-end authorization flow.

The other supported authorization flow is the device flow:
```
# This time, on the device auth question, answer "yes"
sail artisan passport:client

# Now, initiate the device flow.
curl -X POST http://localhost:64000/oauth/device/code \
  -d "client_id={CLIENT_ID}"
```
In the response, you'll see an 8-character `user_code`. Copy that value, and navigate to http://localhost:64000/oauth/device.

Screenshot 2025-11-25 at 6 39 13 PM

Here, you can enter the 8-character code. Upon doing that, you'll see an Approve and Reject button again. Click either, and you'll see a confirmation. In the real-world, the user's app is going to be polling for the approve/reject status.

Screenshot 2025-11-25 at 6 41 03 PM

Screenshot 2025-11-25 at 6 41 08 PM

database/migrations/2025_11_24_000001_create_oauth_tables.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Get the migration connection name.
+     * Auto-generated by `php artisan passport:install`.
+     */
+    public function getConnection(): ?string
+    {
+        return $this->connection ?? config('passport.connection');
+    }
```
Jamiras: Why is this necessary? None of the other migration steps have it.

resources/js/features/auth/components/+enter-device-code/EnterDeviceCodeRoot.tsx
Jamiras, line 1: If not logged in, submitting this form redirects to a login page as a child window. Entering an invalid password returns to this page (as a child window). Submitting the form again opens another level of child window, and submitting from there seems to cause everything to give up.

Image

resources/js/features/auth/components/+authorize/AuthorizeRoot.tsx
Jamiras, line 1: Using the longform `http://localhost:64000/oauth/authorize?client_id=[client_id]&redirect_uri=http%3A%2F%2Flocalhost%3A64000%2Fauth%2Fcallback&response_type=code` while not logged in goes to the login page. Entering the correct credentials works, but entering the wrong credentials, then the correct ones takes you to the main page.

resources/js/features/auth/components/+enter-device-code/EnterDeviceCodeRoot.tsx
Jamiras, line 1: Entering the correct credentials opens the form in a popup window where the code has to be entered again.

app/Providers/RouteServiceProvider.php
```diff
@@ -152,6 +152,33 @@ protected function mapWebRoutes(): void
...
+                            'state' => request()->get('state'),
+                        ]);
+                    }
+
+                    return response()->json([
+                        'message' => 'OAuth authorization successful',
+                        'code' => request()->get('code'),
+                        'state' => request()->get('state'),
+                    ]);
+                })->name('auth.callback.test');
```
Jamiras, line 146: General question: What does the app do with this response? Is the `code` the key for future API requests? It differs from the `id` in the `oauth_auth_codes` table. How does it get mapped back to a client_id/user_id?

Similarly, as there's no response from the device path (it just updates the `user_id` in the `oauth_device_codes` table), what key would those API requests use and how would it map back to the client_id/user_id?

app/Providers/RouteServiceProvider.php
```diff
@@ -152,6 +152,33 @@ protected function mapWebRoutes(): void
...
+                            'error' => $error,
+                            'error_description' => request()->get('error_description'),
+                            'state' => request()->get('state'),
+                        ]);
+                    }
+
+                    return response()->json([
+                        'message' => 'OAuth authorization successful',
+                        'code' => request()->get('code'),
+                        'state' => request()->get('state'),
```
Jamiras, line 144: is `state` supposed to be null?

Image

resources/js/features/auth/components/+authorize/AuthorizeRoot.tsx
m1guelpf: @wescopeland is authentication done manually? if so we should reflash the session on validation error to make sure the redirect url is preserved

app/Providers/RouteServiceProvider.php
m1guelpf: `state` is an optional parameter according to the OAuth2 spec. The idea is the authenticating app will store this on the session, then compare it with the returned state to make sure it matches.

app/Providers/RouteServiceProvider.php
m1guelpf: You exchange the code for an API token using the /oauth/token endpoint. You can see a detailed description of the flow on the [Laravel Passport](https://laravel.com/docs/12.x/passport#requesting-tokens-converting-authorization-codes-to-access-tokens) docs.

database/migrations/2025_11_24_000001_create_oauth_tables.php
wescopeland: I have no idea. It's what Laravel automatically spits out when doing a Passport install in a fresh project.

It looks like some Laravel apps use a microservice for authentication, which is what this might be for. Seems like an edge case. Removed this function from the migration in latest.

resources/js/features/auth/components/+enter-device-code/EnterDeviceCodeRoot.tsx
wescopeland: Fixed in latest. The device code page now requires auth. Upon login, the user should be redirected back to the device code page automatically.

resources/js/features/auth/components/+authorize/AuthorizeRoot.tsx
wescopeland: I think this was caused by a bug in _FortifyServiceProvider.php_:
```php
if (!session()->has('url.intended')) { // it's incorrectly "intended_url" in master
    session()->put('url.intended', url()->previous());
}
```
I could've sworn I fixed this at some point previously.

Should now be working in latest.

resources/js/features/auth/components/+enter-device-code/EnterDeviceCodeRoot.tsx
wescopeland: Check in latest if you're still seeing this - I wasn't while authenticated.

app/Providers/RouteServiceProvider.php
wescopeland: Yep, this is all normal.

For the auth code flow, the app receives `code` at the callback URL. The app can then exchange `code` for an access token via `POST /oauth/token`:

```bash
curl --request POST \
  --url http://localhost:64000/oauth/token \
  --header 'Content-Type: application/json' \
  --data '{
	"code": "... insert code here ...",
	"grant_type": "authorization_code",
	"client_id": "019af19d-9a07-7367-a20b-d50b78584b7d",
	"client_secret": "V121GM7w4bci6a55NbKnle6OXAbULBBh3ZA8Yy2Z",
	"redirect_uri": "http://localhost:64000/auth/callback"
}'
```
Screenshot 2025-12-05 at 10 07 27 PM

This returns both an `access_token` and a `refresh_token`, allowing the app to make API calls on the user's behalf (when we hook up the corresponding middleware).

The device flow is very similar. Some app requests a device code with `POST /oauth/device/code`. They then tell the user "go to this URL and enter this code" (which is why the device auth route should be authenticated, fixed in latest). The user enters the code while the app is polling `POST /oauth/token` with a few special params. Once the user enters the code, one of the polled responses returns an access token.

The test callback route in our code is just to prove returning a code works. In the real world, we'll delete this route, and `code` will be attached by retroachievements.org as a query param to the app's `redirect_uri` when the user approves the auth request.

resources/js/features/auth/components/+enter-device-code/EnterDeviceCodeRoot.tsx
Jamiras: I'm not. And if it wasn't clear - this was hitting the page in an incognito mode, which would require logging in _after_ entering the code. Then it would prompt to enter the code again after logging in. Putting the login before the code addresses the issue.

Jamiras: Appears to be fixed.

Similar to other comment: "Entering the correct credentials" implies doing this while not authenticated.

PR #4171, opened 2025-11-28
feat(api): add V2 JSON:API endpoints for the `System` resource

wescopeland:
This PR builds some more foundational work for the V2 API by adding endpoints for the `System` resource. I chose this resource because it's something of a "leaf node" in our back-end in that it has no child resource dependencies (ie: games depends on systems).

The endpoints are built using `laravel-json-api`'s features and are JSON:API spec compliant.

**How to test this PR**
```bash
# Get a paginated list of systems (defaults to 50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/systems \
  --header 'Accept: application/vnd.api+json' \

# Get a specific system
curl --request GET \
  --url http://localhost:64000/api/v2/systems/1 \
  --header 'Accept: application/vnd.api+json' \

# Custom pagination (page 2, 10 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/systems?page[number]=2&page[size]=10 \
  --header 'Accept: application/vnd.api+json' \

# Sort by name (ascending)
curl --request GET \
  --url http://localhost:64000/api/v2/systems?sort=name \
  --header 'Accept: application/vnd.api+json' \

# Filter by active, sort by name, 25 per page
curl --request GET \
  --url http://localhost:64000/api/v2/systems?filter[active]=1&sort=name&page[size]=25 \
  --header 'Accept: application/vnd.api+json' \
```


app/Api/RouteServiceProvider.php
Jamiras, line 1: I'm not seeing the `Content-Length` header in the response, but I am seeing the size logged in `api_logs`.

app/Http/Middleware/JsonResponse.php
```diff
@@ -11,7 +11,12 @@ class JsonResponse
 {
     public function handle(Request $request, Closure $next): mixed
     {
-        $request->headers->set('Accept', 'application/json');
+        // Don't override the Accept header if it's already set to the media type
+        // required by JSON:API responses.
+        $currentAccept = $request->header('Accept');
+        if ($currentAccept !== 'application/vnd.api+json') {
+            $request->headers->set('Accept', 'application/json');
+        }
```
Jamiras, line 18: Should the response echo the accept type? The response is still sending `Content-Type: application/json`, regardless of whether or not I send `Accept: application/vnd.api+json`.

Jamiras, line 18: Also, what is this block doing? The comment suggests it forces `application/json` if `application/vnd.api+json` is not provided, but if I pass in something like `application/image` (or even `application/json`), I get back an error stating it's not a valid type for the request.

tests/Feature/Api/V2/SystemsTest.php
```diff
@@ -0,0 +1,133 @@
...
+        // Arrange
+        $user = User::factory()->create(['APIKey' => 'test-key']);
+        $active = System::factory()->create(['active' => true]); // active system
+        $inactive = System::factory()->create(['active' => false]); // inactive system
+
+        // Act
+        $response = $this->jsonApi('v2')
+            ->expects('systems')
+            ->withHeader('X-API-Key', 'test-key')
+            ->get('/api/v2/systems?filter[active]=1');
```
Jamiras: Why `filter[active]=1` when the response contains `"active":true`? Shouldn't it be `filter[active]=true`?

app/Api/V2/Systems/SystemSchema.php
Jamiras, line 1: When querying for a specific system, `links[self]` appears at both the top level and data level.

Jamiras, line 1: Additionally, `links[self]` uses the tagged ID, which can't be queried:
```
$ curl --header 'X-API-Key: XXXX' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/systems/27-arcade" && echo
{"message":"API endpoint not found","errors":[{"status":"404","code":"not_found","title":"API endpoint not found"}]}
```

The tagged ID is also returned in the data. I assume the behavior is connected.
```
"id":"27-arcade",
```

wescopeland: This is expected JSON:API behavior per the spec.

The top-level `links[self]` represents "the link that generated the current response document". [ref](https://jsonapi.org/format/#document-top-level)

The data-level `links[self]` is "a self link that identifies the resource represented by the resource object". [ref](https://jsonapi.org/format/#document-resource-object-links)

They'll be identical in cases where a resource is directly fetched, but will often differ in other responses which the V2 API doesn't support yet, but will soon.

app/Api/RouteServiceProvider.php
wescopeland: Will now be returned in response headers in latest.

app/Http/Middleware/JsonResponse.php
wescopeland: I think I originally did this to make tests happy, but looking at it now, it seems a bit too zealous. I've adjusted this in latest to:
```php
if (!$request->headers->has('Accept')) {
    $request->headers->set('Accept', 'application/json');
}
```
Coincidentally, this means calls can be made without needing to manually set an `Accept: application/vnd.api+json` header.

wescopeland: I don't think it should echo it verbatim. Most clients are going to naively send `Accept: */*`.

Also see: https://github.com/RetroAchievements/RAWeb/pull/4171#discussion_r2594410240

tests/Feature/Api/V2/SystemsTest.php
wescopeland: Both are valid and work, although I've changed this to `true` in latest.

Pretty sure this is related to PHP type coercion.

app/Api/V2/Systems/SystemSchema.php
wescopeland: "Fixed" in latest - `id` is still a string, per the spec, but we don't return the slugified ID anymore.

This also implies we can't query by it via the path param.

app/Http/Middleware/JsonResponse.php
wescopeland: OK, I got a quick reminder of why this was here:
Screenshot 2025-12-05 at 9 29 00 PM

Adjusted in latest to:
```php
// Don't override the Accept header for JSON:API routes.
// JSON:API requires Accept: application/vnd.api+json
if (!str_starts_with($request->path(), 'api/v2/')) {
    $request->headers->set('Accept', 'application/json');
}
```

We need `$request->headers->set('Accept', 'application/json')` so our legacy code is happy, but we need it to not run on V2 API routes. Unfortunately, it's global in _Kernel.php_, so we need this conditional in the middleware to bail out.

Jamiras: I misread the "!==" and thought this was turning "vnd.api+json" into "json".

Jamiras: I'm not seeing these changes anywhere in this PR (only match to `str_starts_with` is the comment here). Did you not push?

wescopeland: It's surely there: https://github.com/RetroAchievements/RAWeb/pull/4171/commits/026218263a6aeeccdacfaff06bc241948569d844

Jamiras: Not sure what happened. Maybe I was looking at an individual commit.

PR #4177, opened 2025-11-30
feat(mail): add more implicit subscriber filtering for forum notifications

wescopeland:
This PR adds a few more community email notification filters to bring email rates down another notch.

~~Now, for forum threads with at least 6 pages (90 comments), if the user is implicitly subscribed, hasn't replied in 90+ days, and isn't the OP, they no longer receive an automatic email.~~

For threads with 200+ subscribers, implicit subscribers who haven't posted in the last 45 days will no longer receive email notifications. This targets threads where each new comment triggers hundreds of emails. Most normal threads are completely unaffected.

### DIFF
--- a/app/Helpers/database/forum.php
+++ b/app/Helpers/database/forum.php
@@ -236,14 +236,12 @@ function notifyUsersAboutForumActivity(ForumTopic $topic, User $author, ForumTop
     }
 
     /**
-     * For larger threads (90+ posts), we'll filter out implicit subscribers who
-     * haven't posted in the thread for 90+ days, unless they're the OP.
-     *
-     * Explicit subscribers and implicit subscribers in smaller threads always receive email notifications.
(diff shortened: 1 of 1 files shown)


Jamiras (approved):
I've verified the functionality, but I'm not certain I agree with the change. 

There are several forum topics that I've participated in where I would want to be notified if they've been necro'd (I don't know how many of them are 6+ pages). I guess I could explicitly subscribe to those, but that's putting the onus on the user for functionality that "just worked" before.

If I get too much spam from a message thread, I just unsubscribe from it. I think the problem we're trying to solve here is exacerbated by the fact that these emails usually go to user's junk mail folders, so it's pointless to be sending them anyway.

I think a better solution would be to move to the daily digest for implicit subscriptions, but I realize that's a bigger project.

PR #4180, opened 2025-11-30
perf(user): fix very slow `player_sessions` query

wescopeland:
Resolves https://retroachievementsorg.sentry.io/issues/44906574/?notification_uuid=dfa95762-57c8-479b-a8c2-dca698ffef08&project=4508694685417552&referrer=weekly_report.

The "Last Seen In" component's `$mostRecentSession` query is filtering on 50M+ rows without an index. We can get the same data from `game_recent_players` at about 25x the speed.

There's a similar issue with a mod-only component on game pages (when the claimed game was last played by the claimant). I fixed it in the React game page and ignored it in the legacy game page.

### DIFF
--- a/tests/Feature/Api/V1/UserSummaryTest.php
+++ b/tests/Feature/Api/V1/UserSummaryTest.php
@@ -113,6 +113,10 @@ class UserSummaryTest extends TestCase
 
     public function testGetUserSummary(): void
     {
+        // Freeze time.
+        $now = Carbon::now();
+        Carbon::setTestNow($now);
+
         $game = $this->seedGame(achievements: 3);
         $game->fill([
             'ForumTopicID' => 222334,


resources/views/components/user/profile/last-seen-in.blade.php
```diff
@@ -10,9 +10,9 @@
 ])
 
 <?php
-$mostRecentSession = PlayerSession::where('user_id', $userMassData['ID'])
+$mostRecentSession = GameRecentPlayer::where('user_id', $userMassData['ID'])
     ->with('game')
-    ->orderBy('created_at', 'desc')
+    ->orderByDesc('rich_presence_updated_at')
     ->first();
```
Jamiras: `PruneGameRecentPlayers` will truncate the table daily. For someting like the AotW, this may cause user's entries to disappear before they play another game.

I was going to suggest falling back to the slower query if the first one fails, but there is likely going to be _some_ game where the user is still a recent player.

wescopeland: I forgot about the prune 😞 

Updated in latest to just use the denormalized stuff on the user record. Also updated a few TODOs in the model. I also changed the `AchievementSetClaim` query back to `player_sessions`, but it's now ordering by `rich_presence_updated_at` (which has an index).

Jamiras:
It looks like the `expireRecentlyPlayedGames` call just above that TODO might be dead code? I don't see anything else using `buildUserRecentGamesCacheKey`.

wescopeland:
Removed the TODO and the dead code.

PR #4182, opened 2025-11-30
feat(game): add mobile comments preview card to achievements list

wescopeland:
This PR adds a comments preview card to mobile game pages, appearing at the bottom of the achievement list. On tap, it takes users directly to comments in the Community tab. The card shows:
- An avatar stack of recent commenters.
- A preview of the most recent comment (truncated to 3 lines).
- The total comment count.

Screenshot 2025-11-30 at 2 03 37 PM

---

**Why?**
For performance reasons, we don't want to render the comment list anywhere other than the Community tab on mobile. However, we currently have an information architecture mismatch problem. The desktop layout has trained users that comments = scroll down, but mobile breaks that mental model by hiding them in a tab. This new component meets users where they expect to find the content.

### DIFF
--- a/app/Helpers/database/user-activity.php
+++ b/app/Helpers/database/user-activity.php
@@ -6,8 +6,6 @@ use App\Models\Comment;
 use App\Models\PlayerGame;
 use App\Models\System;
 use App\Models\User;
-use App\Support\Cache\CacheKey;
-use Illuminate\Support\Facades\Cache;
 
 function getIsCommentDoublePost(int $userID, array|int $articleID, string $commentPayload): bool
 {
@@ -61,12 +59,6 @@ function addArticleComment(
     }
 
     return true;
-}
-
-function expireRecentlyPlayedGames(string $user): void
-{
-    $userRecentGamesCacheKey = CacheKey::buildUserRecentGamesCacheKey($user);
-    Cache::forget($userRecentGamesCacheKey);
 }
 
 function getRecentlyPlayedGames(User $user, int $offset, int $count, array &$dataOut): int
--- a/app/Models/AchievementSetClaim.php
+++ b/app/Models/AchievementSetClaim.php
@@ -109,10 +109,10 @@ class AchievementSetClaim extends BaseModel
         $playerSession = PlayerSession::query()
             ->where('game_id', $this->game_id)
             ->where('user_id', $this->user_id)
-            ->orderByDesc('updated_at')
-            ->first(['updated_at']);
+            ->orderByDesc('rich_presence_updated_at')
+            ->first(['rich_presence_updated_at']);
 
-        return $playerSession?->updated_at;
+        return $playerSession?->rich_presence_updated_at;
     }
 
     // == mutators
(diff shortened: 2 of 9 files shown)


resources/js/features/games/components/CommentsPreviewCard/CommentsPreviewCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+            at={previewComment.createdAt}
+            asAbsoluteDate={auth?.user?.preferences?.prefersAbsoluteDates ?? false}
+            enableTooltip={false}
+          />
+        </div>
+      </div>
+
+      {/* CTA that looks like a link */}
+      <div className="flex items-center gap-1 text-xs text-link">
+        <span>{t('View all comments')}</span>
```
Jamiras: This is misleading as it takes you to "Recent Comments", where you can click on the "All X" to actually get to "All Comments".

Also, would it be reasonable/feasible to nagivate to the most recent comment instead of the top of the recent comments list? There's some context being lost when you click on a comment and get taken back to the start of the conversation.

resources/js/features/games/components/CommentsPreviewCard/CommentsPreviewCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (!seenDisplayNames.has(comment.user.displayName)) {
+      seenDisplayNames.add(comment.user.displayName);
+      uniqueUsers.push(comment.user);
+    }
+  }
+
+  const previewComment = recentFirstComments[0];
+
+  const handleClick = () => {
+    setCurrentTab('community');
```
Jamiras: I realize the tab selection control behaves the same way, but if I "swipe back" after clicking the "View All Comments" button, it takes me to the last page I viewed before going to the game instead of back to the end of the achievements list. I feel this may be important due to the fact that the entire widget can be clicked to navigate to the comments, and users may misclick while trying to scroll.

resources/js/features/games/components/CommentsPreviewCard/CommentsPreviewCard.tsx
wescopeland: > This is misleading as it takes you to "Recent Comments", where you can click on the "All X" to actually get to "All Comments".

Good point. I've changed this label to "View recent comments".

> Also, would it be reasonable/feasible to nagivate to the most recent comment instead of the top of the recent comments list? There's some context being lost when you click on a comment and get taken back to the start of the conversation.

Good idea. This should now be done in latest.

resources/js/features/games/components/CommentsPreviewCard/CommentsPreviewCard.tsx
wescopeland: I agree. In latest I've added an optional 2nd arg to `setCurrentTab` which allows it to push to history instead of replacing history.

Jamiras (approved):
I'm still marginally concerned that the entire widget is clickable, but otherwise this seems to be working.

wescopeland:
In latest, I've made an update so only the "View recent comments" text is clickable.

PR #4189, opened 2025-12-02
perf: use a materialized player_stat_rankings table for the Beaten Games leaderboard

wescopeland:
This PR improves performance for the beaten games leaderboard by pre-computing rankings rather than calculating them on page load.

Previously, the page's associated service would aggregate `player_stats` data and compute `RANK()` and `ROW_NUMBER()` window functions in real-time for most requests. On a cache miss, this is very slow. Now, a scheduled job computes and stores these rankings in a dedicated `player_stat_rankings` table, and the service simply reads from it.

This change is a precursor to migrating the global beaten games and global points ranking pages to React.

**How to test this PR**
```bash
sail artisan migrate
sail artisan horizon
sail artisan ra:platform:player:update-beaten-games-leaderboard
```
Navigate to http://localhost:64000/ranking/beaten-games.

On deploy, there will be a brief moment where there are no rankings shown in prod. I think this brief downtime of the leaderboard is likely acceptable given the page is (relatively) low traffic.


app/Platform/AppServiceProvider.php
```diff
@@ -124,6 +126,7 @@ public function boot(): void
             if (app()->environment() === 'production') {
                 $schedule->command(UpdateAwardsStaticData::class)->everyMinute();
                 $schedule->command(CrawlPlayerWeightedPoints::class)->everyFiveMinutes();
+                $schedule->command(UpdateBeatenGamesLeaderboard::class)->everyThirtyMinutes();
```
Jamiras: Shouldn't this just occur in `UpdateGameBeatenMetricsAction`?

wescopeland: We could, but I'm not sure we should. It seems like it'd be awfully computationally expensive - we'd be recalculating the whole leaderboard every time someone beats a game.

If small delays in leaderboard accuracy are acceptable, what we're doing right now is much more efficient.

Jamiras: You could still dispatch the job from there. 

```
UpdateBeatenGamesLeaderboardJob::dispatch($game->id, PlayerStatRankingKind::AllBeaten)->onQueue('game-beaten-metrics');
UpdateBeatenGamesLeaderboardJob::dispatch($game->id, $kind)->onQueue('game-beaten-metrics');
```

Is it not more efficient to spawn a couple jobs every few minutes than to spawn 216 jobs (54 consoles x 4 game types) every half hour (equivalent to 1 job every 8 seconds) - especially given that most of them are no-ops?

The last 100 beats in my dump took place over ~77 minutes, so that's roughly one job every 46 seconds. And those 100 beats only occurred on 23 different consoles. When a singular game is beaten, you only need to queue two jobs: one for that game's system with the all_beaten kind, and one with the appropriate retail/homebrew/hack kind.

wescopeland: That makes sense. I think we actually need 4 jobs dispatched per beat, not 2, otherwise the "All systems" view will go stale:
1. System + AllBeaten
2. System + specific kind (retail, or homebrew, or hack)
3. Overall + AllBeaten
4. Overall + specific kind (retail, or homebrew, or hack)

This is still more efficient than the scheduled job. I've made this change in latest.

app/Platform/Actions/UpdateGameBeatenMetricsAction.php
```diff
@@ -49,6 +52,44 @@ public function execute(Game $game): void
...
+
+        if ($tags->contains('Hack')) {
+            return PlayerStatRankingKind::HacksBeaten;
+        }
+
+        if ($tags->contains('Homebrew') || System::isHomebrewSystem($systemId)) {
+            return PlayerStatRankingKind::HomebrewBeaten;
+        }
+
+        return PlayerStatRankingKind::RetailBeaten;
```
Jamiras, line 104: I notice this is slightly different than the logic in `UpdatePlayerBeatenGamesStatsAction`, which also categorizes demo, prototype, and unlicensed. How do those games get rolled into the beaten_games_leaderboard?

Also, the `isHomebrewSystem` check in `UpdaetPlayerBeatenGamesStatsAction::determineGameKind` should be outside the loop.

wescopeland: You're right on both points.

In latest, I've updated `determineRankingKindForGame` to also check for Demo, Prototype, and Unlicensed tags. Demo and prototype games now correctly return `null` since they only should contribute to `AllBeaten` rankings.

I also moved the `isHomebrewSystem` check to be outside the loop.

Jamiras (approved):

PR #4199, opened 2025-12-05
chore: add aggressive email filtering for some events

wescopeland:
https://discord.com/channels/310192285306454017/1017568867574362283/1446588363258724354


wescopeland:
cc @Jamiras I'll deploy this in a little while so we can beat the weekend rush.

app/Helpers/database/forum.php
```diff
@@ -227,6 +227,21 @@ function submitTopicComment(
...
+        12108, // Peak Streak
+        29289, // AOTW 2025
+        29592, // RA Roulette 2025
+        32219, // Bounty Hunters Villains
+        33001, // RA-TALITY
+    ];
+    if (in_array($topic->id, $blockedTopicIds, strict: true)) {
+        return;
+    }
+    // ENDTODO
```
Jamiras, line 236: Would it make sense to just check `topic->forum_id==25`?

wescopeland: I doubt it'd hurt. Updated to that in latest.

PR #4202, opened 2025-12-06
build: use php 8.4

wescopeland:
This PR migrates the project from PHP 8.2 to PHP 8.4. Most of the legwork for this migration happened during the Laravel 12 upgrade.

There should be no functional changes as part of this PR.

The deployment process should be:
1. Before merging this PR into `master`, install PHP 8.4 on the server through Forge's PHP version manager.
2. Merge this branch, which will deploy to stage while 8.2 is running in that env. The app should continue running normally.
3. Switch stage to 8.4. Make sure that doesn't blow anything up. This will probably cause a ~5 second outage in stage, which should be monitored to approximate the production impact of the version change.
4. Deploy this branch to prod in isolation and then switch prod to 8.4 in the middle of the night.


Jamiras:
> I repro'd the issue you're seeing locally and think I've fixed it with a config change in the container. Try:
> 
> ```
> sail down
> sail build --no-cache # this will take a minute
> sail up
> ```
> 
> The issue was related to something with Swoole and Octane, which are enabled in the container but are features we don't actually use at the moment.

That seems to have worked. Thanks.

PR #4203, opened 2025-12-07
(feature) add daily digest email for implicit forum post subscriptions

Jamiras:
https://discord.com/channels/476211979464343552/1440724197952192654/1440725252790550678

When a post is made to a forum topic, any users who are implicitly subscribed and haven't been active in the last 7 days will have their notification delayed. Once a day, delayed notifications will be processed and users will receive a single email containing all delayed notifications.

For topics that have been around for a while with many pages of posts, a new post can generate hundreds of mails. If several posts are made, this gets compounded. Right now, notifications caused by people posting to the event forum pages constitute roughly 50% of all of our email volume. Most participants don't care about other user's progress, so sending these emails is wasted. Now, if a user posts to one of the mega-threads, instead of sending out 100+ emails per post, we'll send out 20 or so to the users who are explicitly subscribed, and who have been recently active in the discussion. The other 80 will be queued and sent the next day. If a second post is made, we'll send out the 20 or so immediate emails, and the other 80 will remain queued, saving us 80 emails.

image

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -121,6 +121,10 @@ class Handler extends ExceptionHandler
             return $e->redirect;
         }
 
+        if ($request->is('api/v2/*')) {
+            return $this->renderWebApiV2Exception($e);
+        }
+
         if ($request->expectsJson()) {
             if ($e instanceof JsonApiException) {
                 $errors = $e->getErrors()->toArray();
@@ -242,5 +246,38 @@ class Handler extends ExceptionHandler
         }
 
         return parent::render($request, $e);
+    }
+
+    private function renderWebApiV2Exception(Throwable $e): Response
+    {
+        if ($e instanceof JsonApiException) {
+            $errors = $e->getErrors()->toArray();
+
+            return response()->json([
+                'message' => $errors[0]['title'] ?? 'JSON:API error',
+                'errors' => $errors,
+            ], $e->getStatusCode());
+        }
+
+        $status = match (true) {
+            $e instanceof AuthenticationException => 401,
+            $e instanceof AuthorizationException => 403,
+            $e instanceof HttpExceptionInterface => $e->getStatusCode(),
+            default => 500,
+        };
+
+        $statusText = Response::$statusTexts[$status] ?? 'Error';
+        $title = $e->getMessage() ?: $statusText;
+
+        return response()->json([
+            'message' => $title,
+            'errors' => [
+                [
+                    'status' => (string) $status,
+                    'code' => Str::snake($statusText),
+                    'title' => $title,
+                ],
+            ],
+        ], $status);
     }
 }
--- a/app/Models/Achievement.php
+++ b/app/Models/Achievement.php
@@ -12,6 +12,7 @@ use App\Platform\Enums\AchievementAuthorTask;
 use App\Platform\Enums\AchievementFlag;
 use App\Platform\Enums\AchievementType;
 use App\Platform\Events\AchievementCreated;
+use App\Platform\Events\AchievementDeleted;
 use App\Platform\Events\AchievementMoved;
 use App\Platform\Events\AchievementPointsChanged;
 use App\Platform\Events\AchievementPublished;
@@ -180,6 +181,12 @@ class Achievement extends BaseModel implements HasVersionedTrigger
         // When restoring an achievement, restore its tickets.
         static::restoring(function (Achievement $achievement) {
             $achievement->tickets()->restore();
+        });
+
+        // When an achievement is deleted, dispatch an event so game metrics can be recalculated.
+        // Otherwise, the denormalized unpublished achievement count will be wrong.
+        static::deleted(function (Achievement $achievement) {
+            AchievementDeleted::dispatch($achievement);
         });
     }
 
--- a/app/Platform/EventServiceProvider.php
+++ b/app/Platform/EventServiceProvider.php
@@ -6,6 +6,7 @@ namespace App\Platform;
 
 use App\Events\UserDeleted;
 use App\Platform\Events\AchievementCreated;
+use App\Platform\Events\AchievementDeleted;
 use App\Platform\Events\AchievementMoved;
 use App\Platform\Events\AchievementPointsChanged;
 use App\Platform\Events\AchievementPublished;
@@ -47,6 +48,9 @@ class EventServiceProvider extends ServiceProvider
 {
     protected $listen = [
         AchievementCreated::class => [
+            DispatchUpdateGameMetricsJob::class, // dispatches GameMetricsUpdated
+        ],
+        AchievementDeleted::class => [
             DispatchUpdateGameMetricsJob::class, // dispatches GameMetricsUpdated
         ],
         AchievementMoved::class => [
--- a/app/Platform/Events/AchievementDeleted.php
+++ b/app/Platform/Events/AchievementDeleted.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Platform\Events;
+
+use App\Models\Achievement;
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Broadcasting\PrivateChannel;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+
+class AchievementDeleted
+{
+    use Dispatchable;
+    use InteractsWithSockets;
+    use SerializesModels;
+
+    public function __construct(
+        public Achievement $achievement,
+    ) {
+    }
+
+    public function broadcastOn(): PrivateChannel
+    {
+        return new PrivateChannel('channel-name');
+    }
+}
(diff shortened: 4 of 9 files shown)


app/Community/Services/SubscriptionService.php
```diff
@@ -18,16 +18,56 @@
...
+     *   'explicitlySubscribed': Users who are explicitly subscribed to the content.
+     *   'implicitlySubscribedNotifyNow': Users who are implicitly subscribed to the content and should be notified immediately.
+     *   'implicitlySubscribedNotifyLater': Users who are implicitly subscribed to the content and should be notified when convienent.
+     */
+    public function getSegmentedSubscriberIds(SubscriptionSubjectType $subjectType, int $subjectId, ?int $authorUserId): array
+    {
+        [$explicitSubscriberIds, $implicitSubscriberIds] = $this->getSubscriberIds($subjectType, $subjectId);
+
+        // split implicit subscriptions into recent and older
+        $recentActivityCutoff = now()->subDays(7);
```
Jamiras, line 60: Realistically, this could be 2 or 3 days. It's just a hook to keep recently active participants engaged. I wouldn't go below 48 hours though, as that gives people getting the daily digest a chance to respond before the recently active participant is considered no longer active.

app/Community/Actions/SendDailyDigestAction.php
```diff
@@ -0,0 +1,119 @@
...
+            ->where('author_id', '!=', $delayedSubscription->user_id)
+            ->where('is_authorized', true)
+            ->count();
+    }
+
+    public function getLink(int $subjectId, int $firstUpdateId): string
+    {
+        return route('forum-topic.show', ['topic' => $subjectId, 'comment' => $firstUpdateId]) . '#' . $firstUpdateId;
+    }
+}
```
wescopeland, line 123: nit: Unless `BaseDelayedSubscriptionHandler` and `ForumTopicDelayedSubscriptionHandler` are going to stay private to the action forever, we should probably break them both out to their own individual separate files so we're compliant with [the PSR-4 autoloading standard](https://www.php-fig.org/psr/psr-4/).

Right now, both of these classes ignore `namespace App\Community\Actions;`, effectively making them globally-namespaced.

app/Community/Actions/SendDailyDigestAction.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private function getHandler(SubscriptionSubjectType $subjectType): BaseDelayedSubscriptionHandler
+    {
+        return match ($subjectType) {
+            SubscriptionSubjectType::ForumTopic => new ForumTopicDelayedSubscriptionHandler(),
+
+            // this is wrong, but we have to return something to satisfy phpstan.
+            // other cases will be filled in as the calls are updated.
+            default => new ForumTopicDelayedSubscriptionHandler(),
```
wescopeland: Maybe it'd be better to throw an exception here?
```php
default => throw new InvalidArgumentException("Supported subject type: {$subjectType->value}"),
```

app/Helpers/database/forum.php
```diff
@@ -227,71 +231,31 @@ function submitTopicComment(
...
-        sendActivityEmail(
-            $subscriber,
-            $topic->id,
-            $author,
-            ArticleType::Forum,
-            $topic->title ?? '',
-            $urlTarget,
-            payload: $payload
-        );
+    $subscribers = $subscriptionService->getSegmentedSubscriberIds(SubscriptionSubjectType::ForumTopic, $topic->id, $topic->author_id);
```
wescopeland, line 235: Intellectual curiosity: should the 3rd arg be `$topic->author_id` or `$author->id`? I think I'm getting confused on this one implementation detail.

database/migrations/2025_12_06_000000_create_user_delayed_subscriptions.php
wescopeland, line 1: Missing a `created_at` column.

wescopeland, line 1: We should probably add a unique constraint on `user_id` + `subject_type` + `subject_id`.

It's already enforced as a business rule in the implementation code, adding it as a unique constraint would also automatically give us an efficient lookup index.

app/Community/Actions/SendDailyDigestAction.php
```diff
@@ -0,0 +1,119 @@
...
+                    'count' => $count,
+                ];
+            }
+        }
+
+        // send the mail
+        Mail::to($user->EmailAddress)->queue(
+            new DailyDigestMail($user, $notificationItems)
+        );
+    }
```
wescopeland, line 81: I could be misinterpreting, but I _think_ this logic means it's theoretically possible for `$notificationItems` to be empty and, thus, our email body to virtually be empty.

Even if that's implausible, we could save ourselves a lot of trouble in the long run with overzealous email providers by:
```php
if (empty($notificationItems)) {
    return;
}

// send the mail
Mail::to(...);
```

app/Community/Services/SubscriptionService.php
```diff
@@ -36,19 +76,20 @@ public function getSubscribers(SubscriptionSubjectType $subjectType, int $subjec
             ->get();
         $explicitSubcriberIds = $subscribers->pluck('user_id')->toArray();
```
wescopeland: nit: I noticed a typo on this variable name.

app/Community/Actions/SendDailyDigestAction.php
Jamiras: Possible, but unlikely. It would mean every delayed subscription the user was queued up for got deleted before the email was sent. I've added the return anyway.

app/Community/Actions/SendDailyDigestAction.php
Jamiras: done.

app/Community/Actions/SendDailyDigestAction.php
Jamiras: I'd make them private members of the class if I could. This mimics the implementation in `SubscriptionService` - https://github.com/RetroAchievements/RAWeb/pull/3874#discussion_r2328755421

app/Community/Services/SubscriptionService.php
Jamiras: fixed.

app/Helpers/database/forum.php
Jamiras: In this case, `$topic->author_id` is correct. It ensures the original author gets an email by forcing it into the `notifyNow` bucket if they only have an implicit subscription.

I've renamed it from `authorUserId` to `subjectAuthorUserId` and added doc comments.

database/migrations/2025_12_06_000000_create_user_delayed_subscriptions.php
Jamiras: Intentional. These records are ephemeral and shouldn't last more than 24 hours.

Jamiras: There's two queries against this table currently:
1) All records for a user_id
2) All users for a subject type/id pair.

Having an index on all three properties doesn't satisfy either of those criteria.

Jamiras: I'm worried that a unique constraint may actually be detrimental. There's currently a small race condition where two users can post at the exact same time. If both posts get past the check for an existing row before the other creates one, they will both create a new row. In that case, I feel safer writing both rows than throwing a constraint error. In this rare edge case, the user would get two links to the same updated content. We could deduplicate them in the processing logic if it's actually a concern.

app/Community/Actions/SendDailyDigestAction.php
wescopeland: You make a fair point; I just can't help but wonder if it'd be best to not accumulate this pattern.

I wouldn't want to block the PR on this, because you're right, there's already precedent for it. But I'm wondering if it may be worth breaking these classes out into a dedicated namespace like `App\Community\Actions\DailyDigest` so they remain grouped but also PSR-4 compliant.

database/migrations/2025_12_06_000000_create_user_delayed_subscriptions.php
wescopeland: I assumed otherwise due to this line in `UserDelayedSubscription`:
```php
public const UPDATED_AT = null; // We only track created_at.
```
I think this means in the model we should probably remove this line and set `$timestamps` to false.

wescopeland: I agree with your reasoning, thanks for that explanation. Marking this as resolved.

Jamiras: I guess I cloned the wrong class to start with. Removed `UPDATED_AT` and added `$timestamps = false`.

wescopeland (approved):
LGTM.

Here's a screenshot of the latest stats in SES:
Screenshot 2025-12-09 at 7 38 49 AM

To be fair, I'm not actually convinced anymore these stats are accurate, because they don't match up with other stats I'm seeing in our billing dashboard.

We may want to cover comment sections with digests in the near future. It'll be interesting to see what the chart says a few days after this change goes out.

PR #4204, opened 2025-12-07
(feature) add indicator for demoted achievements used in events

Jamiras:
https://discord.com/channels/310192285306454017/1017568867574362283/1441512635525894255

This is pretty much the least invasive solution. An active event achievement associated to an unofficial achievement will get a error indicator in the list. Hopefully, this is sufficient for a user to contact the events team so they can sort it out.

Ideally, there would be some sort of master list that the events team could look at showing all affected achievements.

Example from http://localhost:64000/event/14:

image


resources/js/common/components/AchievementsListItem/AchievementDateMeta/AchievementDateMeta.tsx
```diff
@@ -70,6 +73,18 @@ export const AchievementDateMeta: FC<AchievementDateMetaProps> = ({
         </p>
       ) : null}
 
+      {isDemoted ? (
+        <span
+          className="flex cursor-help items-center"
+          title={t(
+            'This achievement cannot currently be earned because the source achievement has been demoted.',
+          )}
```
wescopeland: UX: Recommend using BaseTooltip for this.

resources/js/common/components/AchievementsListItem/AchievementDateMeta/AchievementDateMeta.tsx
```diff
@@ -70,6 +73,18 @@ export const AchievementDateMeta: FC<AchievementDateMetaProps> = ({
         </p>
       ) : null}
 
+      {isDemoted ? (
+        <span
+          className="flex cursor-help items-center"
```
wescopeland: UX: Recommend also adding `gap-1` or `gap-1.5` to add some space between the icon and the label, whichever of the two looks better.

Jamiras: That's what the `ml-0.5` was supposed to be doing. Replaced with `gap-1`.

resources/js/common/components/AchievementsListItem/AchievementDateMeta/AchievementDateMeta.tsx
Jamiras: changed.

Jamiras: Image

PR #4205, opened 2025-12-07
(refactor) add SubmitLeaderboardAction

Jamiras:
This has an intentional change that new leaderboards can no longer be added to games without an active claim. This most likely only affects the LotM, and I have already communicated this change to drisc, who is working on a Clone tool for the administration UI: https://discord.com/channels/476211979464343552/1159595979746582602/1446606651305103412


app/Connect/Actions/SubmitLeaderboardAction.php
```diff
@@ -0,0 +1,242 @@
...
+        $this->valueDefinition = request()->input('l') ?? '';
+        $this->lowerIsBetter = request()->boolean('w', false);
+        $this->format = request()->input('f') ?? '';
+
+        $checksum = request()->input('h') ?? '';
+        if (!$this->checksumMatches($checksum, $this->user->display_name)) {
+            if ($this->user->User === $this->user->display_name || !$this->checksumMatches($checksum, $this->user->User)) {
+                return $this->accessDenied('Invalid checksum.');
+            }
+        }
```
Jamiras, line 91: I've enabled the checksum validation in this PR. We've been sending it basically forever.

app/Connect/Actions/SubmitLeaderboardAction.php
```diff
@@ -0,0 +1,242 @@
...
+                    $newMem,
+                    versioned: true, // we don't currently support unpublished leaderboards
+                    user: $this->user,
+                );
+            }
+
+            $editString = implode(', ', $fields);
+            if (!empty($editString)) {
+                $this->addLegacyAuditComment(ArticleType::Leaderboard, $leaderboard->id,
+                    "{$this->user->display_name} edited this leaderboard's $editString."
```
Jamiras, line 176: The legacy audit string now specifies which fields were modified.

app/Connect/Support/GeneratesLegacyAuditComment.php
```diff
@@ -0,0 +1,22 @@
...
+trait GeneratesLegacyAuditComment
+{
+    protected function addLegacyAuditComment(int $articleType, int $articleId, string $comment): void
+    {
+        $comment = Comment::create([
+            'ArticleType' => $articleType,
+            'ArticleID' => $articleId,
+            'user_id' => Comment::SYSTEM_USER_ID,
+            'Payload' => $comment,
+        ]);
```
wescopeland, line 18: DX: `string $comment` is shadowed due to it being immediately overwritten by this created `Comment` model. It works, but it's kinda confusing.

I recommend renaming the `string $comment` arg to something like `$payload`.

tests/Feature/Connect/UploadLeaderboardTest.php
wescopeland, line 1: Testing: Recommend adding a case for an invalid checksum.

app/Connect/Actions/SubmitLeaderboardAction.php
wescopeland, line 1: Intellectual curiosity: the old code had guards for `isValidConsoleId()`. It makes sense now that these are no longer present: creating a leaderboard requires a claim and we're assuming non-system "games" can't be claimed.

However, I can't help but wonder if we should also bake this prevention logic into `AchievementSetClaimPolicy::create()`.

wescopeland (approved):
Non-blocking feedback.

app/Connect/Support/GeneratesLegacyAuditComment.php
Jamiras: I could see myself being confused by that too. Updated.

tests/Feature/Connect/UploadLeaderboardTest.php
Jamiras: I thought i had. I was definitely planning to. Done.

app/Connect/Actions/SubmitLeaderboardAction.php
Jamiras: `isValidConsoleId` actually validates the console is active, so you couldn't add leaderboards to DOS games for example.

I thought claims were already restricted to valid consoles, but it doesn't look like that's true. Although, it would be pretty obvious is someone claims a DOS game before we start the DOS rollout.

PR #4206, opened 2025-12-07
feat: migrate searchresults.php to React/Inertia

wescopeland:
This PR migrates the legacy search results page to React.

**How to test this PR:**
```bash
sail artisan horizon # restart it

sail artisan scout:flush "App\Models\ForumTopicComment"
sail artisan scout:flush "App\Models\Comment"

sail artisan scout:import "App\Models\ForumTopicComment"
sail artisan scout:import "App\Models\Comment"

pnpm build && sail artisan inertia:start-ssr
```
The page is accessible at http://localhost:64000/search.

Screenshot 2025-12-07 at 12 47 17 PM

Functionally, this should be nearly identical to the global search box, with a few notable exceptions:
* Most forum posts are now searchable.
* Most social comments are now searchable.
* Pagination is now supported on filtered results (when "All" isn't picked).
* Scout/Meilisearch totally take the wheel on what's returned, meaning we don't show partial results that look like "... my query ..." anymore.

The page does not have 1:1 feature parity with the legacy page. There are a few omissions:
* Internal/moderation comments are no longer searchable.
* Safeguards have been added to prevent invisible comments from being searched (nothing from banned users, deleted users, users whose walls are disabled, etc).

One more implementation detail to note is **initial results are not sent down via page props.** They are always fetched client-side. This means a URL like http://localhost:64000/search?query=super will briefly flash a loading state. We can enhance this later, but it's a lot of complexity.

Given `searchresults.php` is still one of our most-visited pages, and its queries are among our slowest, this change should hopefully take a bit of pressure off the database.

Screenshot 2025-12-07 at 12 51 05 PM

Screenshot 2025-12-07 at 12 51 11 PM

### DIFF
--- a/resources/js/features/games/components/GameAchievementSetsContainer/SetSelectionControl/SubsetConfigurationButton/SubsetConfigurationButton.test.tsx
+++ b/resources/js/features/games/components/GameAchievementSetsContainer/SetSelectionControl/SubsetConfigurationButton/SubsetConfigurationButton.test.tsx
@@ -145,7 +145,7 @@ describe('Component: SubsetConfigurationButton', () => {
     });
 
     // ACT
-    await userEvent.click(screen.getByRole('button', { name: /subset configuration/i }));
+    await userEvent.click(await screen.findByRole('button', { name: /subset configuration/i }));
 
     // ASSERT
     await waitFor(() => {


resources/js/features/search/components/CommentResultDisplay/CommentResultDisplay.tsx
Jamiras, line 1: Comment results seem to return oldest first

Image

resources/js/features/search/components/+root/SearchMainRoot.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <div className="flex flex-col gap-6">
+      <h1 className="text-h3 font-semibold">{t('Search')}</h1>
+
+      <SearchInput query={query} isLoading={isLoading} onSearch={handleSearch} />
+
+      <ScopeSelector scope={scope} onScopeChange={handleScopeChange} />
+
+      <div className="flex min-h-[50vh] flex-col gap-4">
+        <SearchResultsContainer searchResults={searchResults} isLoading={isLoading} query={query} />
```
Jamiras, line 70: I wonder if it would be possible to make the results stretch to fill?

Image

Image

resources/js/features/search/components/CommentResultDisplay/CommentResultDisplay.tsx
wescopeland: Looks like the index was entirely missing `rankingRules`. Added in latest, though to see the result it does require:
```bash
sail artisan horizon # restart
sail artisan scout:flush "App\Models\Comment"
sail artisan scout:import "App\Models\Comment"
```

resources/js/features/search/components/+root/SearchMainRoot.tsx
wescopeland: Yea, messed something up before the initial commit. Fixed in latest.

app/Models/Comment.php
```diff
@@ -124,6 +145,34 @@ public function getIsAutomatedAttribute(): bool
         return $this->user_id === self::SYSTEM_USER_ID;
     }
 
+    public function getUrlAttribute(): ?string
+    {
+        switch ($this->ArticleType) {
+            case ArticleType::Game:
+                return route('game.show', ['game' => $this->ArticleID, 'tab' => 'community']) . "#comment_{$this->ID}";
```
Jamiras: If the comment is no longer in the recent comments, the link is ineffective and the result is just the game page.

Search on "primehack".

The first result just loads the game page without even jumping to the comments section.

http://localhost:64000/game/3154?tab=community#comment_3966491

The actual comment is on page 2:

http://localhost:64000/game/3154/comments?page=2#comment_3966491

Jamiras: I assume this applies to the other comment types below.

app/Models/ForumTopicComment.php
Jamiras, line 1: I'm not sure if it's a bug in the data, but `forum_topic` 33178 has been deleted, but it's `forum_topic_comments` have not. As a result, you can search on 'layanan' to find the comment (and see it's contents), but clicking on it will give you a 404.

app/Models/ForumTopicComment.php
```diff
@@ -76,6 +76,12 @@ public function shouldBeSearchable(): bool
             return false;
         }
 
+        // Don't index comments from topics that require permissions.
+        $this->loadMissing('forumTopic');
+        if ($this->forumTopic && $this->forumTopic->required_permissions !== 0) {
+            return false;
+        }
```
Jamiras, line 83: As best as I can tell, changing the required permissions on a forum topic doesn't remove it from the index or add it back.

A reasonable topic to test with is http://localhost:64000/forums/topic/33180

Jamiras: Similarly, authorizing a post seems to trigger a `RemoveFromSearch` job for the UserModeration comment (which shouldn't have been added in the first place), but did not add the `forum_topic_comment` to the index.

resources/js/features/search/components/ForumCommentResultDisplay/ForumCommentResultDisplay.tsx
```diff
@@ -0,0 +1,65 @@
...
+      ) : (
+        <div className="flex size-10 items-center justify-center rounded bg-neutral-700 light:bg-neutral-200">
+          <LuMessageSquare className="size-5 text-neutral-400" />
+        </div>
+      )}
+
+      <div className="flex min-w-0 flex-1 flex-col gap-0.5">
+        <div className="flex items-center gap-2">
+          {forumComment.user ? (
+            <span className="font-medium text-link">{forumComment.user.displayName}</span>
```
Jamiras: Making this appear as a link is misleading, particularly for banned users

Search on "goldstein", first result.

app/Models/Comment.php
wescopeland: Addressing this is complex enough to go into its own PR: https://github.com/RetroAchievements/RAWeb/pull/4247

wescopeland: Merged latest `master` into this branch, which should address this feedback.

app/Models/ForumTopicComment.php
wescopeland: It's a bug in the search index. Should be fixed in latest.

app/Models/ForumTopicComment.php
wescopeland: The superfluous `RemoveFromSearch` job is a quirk of Laravel Scout - it's trying to remove something that isn't indexed. Probably not worth chasing down.

In latest, changing the required perms of a forum topic shouldn't update the replies that are indexed. Authorizing a post should also update what's stored in the index.

resources/js/features/search/components/ForumCommentResultDisplay/ForumCommentResultDisplay.tsx
wescopeland: Fixed to use `text-text` in latest.

app/Models/ForumTopicComment.php
Jamiras: > In latest, changing the required perms of a forum topic shouldn't update the replies that are indexed. 

If I change a forum topic to be Moderator only, the results are still returned as search matches when not logged in, but clicking on them generates an access denied error. Restricted ForumTopicComments should not be exposed through the search engine.

Changing a forum topic from Moderated to Unregistered does index it (33180 illustrates this behavior), but changing it back to Moderated does not remove it from the index.

A recently authorized post is now being indexed correctly, so it's just restricting access that still needs to be addressed.

wescopeland: This should be fixed now - I also refactored the user model logic to move it into the observer.

PR #4207, opened 2025-12-07
feat(user): on account delete, close all reporter's unresolved tickets

wescopeland:
Deleted users can't collaborate with developers on ticket resolutions. When someone's account deletion goes through, close all their unresolved tickets.

### DIFF
--- a/app/Actions/ClearAccountDataAction.php
+++ b/app/Actions/ClearAccountDataAction.php
@@ -5,9 +5,13 @@ declare(strict_types=1);
 namespace App\Actions;
 
 use App\Community\Actions\DeleteMessageThreadAction;
+use App\Community\Enums\ArticleType;
+use App\Community\Enums\TicketState;
 use App\Enums\Permissions;
 use App\Events\UserDeleted;
+use App\Models\Comment;
 use App\Models\Leaderboard;
+use App\Models\Ticket;
 use App\Models\UnrankedUser;
 use App\Models\User;
 use App\Platform\Actions\RecalculateLeaderboardTopEntryAction;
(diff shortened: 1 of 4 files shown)


app/Actions/ClearAccountDataAction.php
```diff
@@ -97,4 +103,37 @@ public function execute(User $user): void
...
+            return;
+        }
+
+        $now = Carbon::now();
+        foreach ($unresolvedTickets as $ticket) {
+            $ticket->update([
+                'ReportState' => TicketState::Closed,
+                'ResolvedAt' => $now,
+                'resolver_id' => null,
+            ]);
```
Jamiras, line 125:
```
  Add fillable property [ReportState, ResolvedAt, resolver_id] to allow mass assignment on [App\Models\Ticket].
```

televandalist:
> In general, I disagree with this change. Just because a user isn't around any more doesn't mean the ticket isn't valid.
> 
> What we should do is change any tickets in Request state back to Open, and remove the ability to "Transfer to reporter" if the reporter has been deleted.
> 
> Developers still have the option to close stuff as not reproducible if the reporter doesn't respond. I see a deleted reporter account as just someone guaranteed to not respond.

It's very possible that it would cause legitimate issues to be put on the back burner, but as opposed to dealing with possible wild goose chases on a problem that might not even be real and not being able to get clarification about anything from a dead account... I think it's a better option. If they don't get closed automatically, then I think at the very least, someone deleting their account should be used as a deal breaker when deciding whether or not to close a ticket they opened. I have drafts for an entire rubric for this sort of thing...

Between the many disconnects between players and devs, RA having over 180k achievements in core without an active dev/maintainer, having no control over what cores actually load achievements, disc-based games being easily modified without changing the RA hash, softcore being the wild west, cheaters, and countless other issues... there needs to be a serious crackdown on what's actually considered a legitimate issue. It's bad enough that roughly a third of tickets weren't even necessary, but the weight they have on development as a whole is way too extreme. They're becoming more of a ball & chain for devs than anything else and I totally understand why some devs get fed up and quit or just disappear entirely.

We're also running out of ways to incentivize handling outstanding tickets. We're nearly a month into this year's Devember and have barely made a dent...

PR #4209, opened 2025-12-07
feat(moderation): add user_moderation_actions table

wescopeland:
This PR adds a new table, `user_moderation_actions`, that will be extremely helpful for facilitating the remaining work in https://github.com/RetroAchievements/RAWeb/discussions/4067.

The table is basically a queryable log of all moderation actions taken on RA. We track six actions:
* Unranks (untracks).
* Reranks.
* Bans.
* Unbans.
* Mutes.
* Manual unmutes.

`UserObserver` has been augmented so moderation actions get recorded. A backfill command that queries activitylog is included. I could query the legacy server comments too, but I'm not sure it's worth the additional complexity.

**How to test this PR:**
```bash
sail artisan migrate
sail artisan ra:community:backfill-moderation-actions
```

### DIFF
--- a/app/Community/Commands/BackfillModerationActions.php
+++ b/app/Community/Commands/BackfillModerationActions.php
@@ -233,7 +233,7 @@ class BackfillModerationActions extends Command
                 ->where('actioned_by_id', $note->user_id)
                 ->whereBetween('created_at', [
                     $noteTimestamp->copy()->subMinutes(5),
-                    $noteTimestamp->copy()->addMinutes(5),
+                    $noteTimestamp->copy()->addMinutes(30),
                 ])
                 ->whereNull('reason')
                 ->first();


app/Models/UserModerationAction.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public const UPDATED_AT = null;
+
+    protected $fillable = [
+        'user_id',
+        'actioned_by_id',
+        'action',
+        'reason',
+        'expires_at',
+    ];
```
Jamiras, line 23: The second comments widget in the User profile page under the Moderation dropdown was meant to be used for this sort of thing. It already captures when users are banned. It was intended that a moderator would add a comment explaining the ban or linking to the discord discussion that led to the ban. That rarely happens.

There are 1.6M comments `where ArticleType=9`, but 95% of those are "Server set account to registered" and "Server set account type to Unregistered". Another 3% is logs for authorizing forum posts. And the remaining 2% are notes about email addresses being changed and account deletion requests/cancelations. There are only ~700 actual moderation notes (`where user_id != 14188`).

It's probably going to be difficult, but it would be nice to roll that data into the new table.

wescopeland: Gave it a fair shot in latest with a new `correlateModeratorNotes` function in the backfill command.

It's not able to find too much, but probably better than nothing:
```
❯ sail artisan ra:community:backfill-moderation-actions
Backfilling moderation actions from activity log...
Found 140699 activity log entries to process.
 140699/140699 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
Backfill complete. Created 1579 moderation action records.
Correlating moderator notes from Comment table...
Correlated 79 moderator notes with action records.
```

app/Community/Commands/BackfillModerationActions.php
```diff
@@ -204,4 +208,42 @@ private function determineUnrankAction(array $old, array $attributes): ?Moderati
...
+        $numCorrelated = 0;
+        foreach ($notes as $note) {
+            /** @var Carbon $noteTimestamp */
+            $noteTimestamp = $note->Submitted;
+
+            $action = UserModerationAction::where('user_id', $note->ArticleID)
+                ->where('actioned_by_id', $note->user_id)
+                ->whereBetween('created_at', [
+                    $noteTimestamp->copy()->subMinutes(5),
+                    $noteTimestamp->copy()->addMinutes(5),
```
Jamiras: Consider expanding this window. I think 5 minutes before is fine, as the user would post the comment, then immediately make the change. But on the other side of things, the user might take some time composing the comment after applying the change. Just bumping this to 15 minutes found an additional 12 correlations over the base query. Bumping it to 30 found 4 more. And 1 more was found at 60 minutes.

PR #4210, opened 2025-12-08
feat(leaderboards): add active/disabled/unpublished and clone functionality

drisc:
This PR adds the ability to clone and disable leaderboards from the Filament management application.

In order to facilitate this Event Managers have been added to the create and update policy and a clone and disable policy have been created that includes Developers and Event Managers. Jr Devs have not been included in these permissions.

Event Managers are limited to being able to update the Title, Description and Display Order of leaderboards, they require the create permission to make a new board but the policy also prevents them from creating a completely new board via the Filament interface as it does for all users.

Leaderboards now have a `state` field with three possible values: active, disabled, and unpublished. Active leaderboards function normally. Disabled leaderboards are excluded from the Connect API patch/achievementsets data, preventing them from activating in emulators while still appearing on the site. Unpublished leaderboards are additionally hidden from normal page listings. Promote and demote actions allow changing state individually or in bulk via Filament.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -100,4 +100,4 @@ Route keys for route model binding should be validated in the respective RouteSe
 
 ## Translating the Application
 
-We welcome contributions to translate the RetroAchievements website into different languages. To get started, please refer to our [translations guide](TRANSLATIONS.md) for detailed instructions on how to add or update translations.
+We welcome contributions to translate the RetroAchievements website into different languages. To get started, please refer to our [translations guide](https://github.com/RetroAchievements/RAWeb/blob/master/docs/TRANSLATIONS.md) for detailed instructions on how to add or update translations.


app/Filament/Actions/DisableLeaderboardAction.php
wescopeland, line 1: I have very mixed feelings about this. Based on the `LeaderboardPolicy` change, this has implications well beyond just event management: we'll be baking a band-aid into our code.

drisc: I'm not a fan of it either but without the ability to disable boards from Filament LotM will have to halt unless another events team member with full dev perms can take over it. The upcoming change that requires a game be claimed to alter leaderboards will mean that both Xionx (no dev perms) and I (jr dev perms) can't manage it any more. That full dev would also just be doing the same thing this code is but from the dev tools in RALib.

Once the ability to mark leaderboards as 'Unofficial' is added I will rip this out. As implemented in this PR, it is basically the same as what we have done via RALib to manage the event since it started, prior event hosts have always added an impossible condition to remove the board from use, this is just a web accessible version of that.

drisc: I can remove the Developer role from the disable policy and only allow Event Managers to see it if that helps. It's pretty much only there for event leaderboards that need to be disabled anyway.

wescopeland: I completely respect & understand the position you're in with LotM. I have to set a firm boundary: I can't possibly merge & deploy this PR in its current form.

My specific objection is the `Mem` field hack. Having a dedicated action class to inject 0=1 codifies a workaround into our codebase. "I'll remove it later" is how technical debt becomes permanent. I'd be happy to accept either:
1. This PR without the disable functionality (the clone stuff is fine).
2. A proper implementation with a `disabled_at` timestamp or some such `status` field, though I realize this work will have tentacles all over the place.

If option 2 is what you need for LotM, I can assist with this.

drisc: Option 2 is where I'd like to go but from my understanding emulators would still try to activate the leaderboard when the conditions are met as they don't have a concept of a 'disabled' leaderboard currently and wouldn't know to look for a flag indicating such.

I would still like to continue hosting LotM so any assistance with Option 2 would be greatly appreciated.

wescopeland: Option 2 definitely seems like the way to go. I think we can do it without touching anything outside of RAWeb - we can modify Connect API responses so the leaderboards aren't sent in the payload if they're disabled.

app/Policies/LeaderboardPolicy.php
```diff
@@ -140,4 +148,21 @@ public function resetAllEntries(User $user): bool
             Role::DEVELOPER,
         ]);
     }
+
+    public function clone(User $user): bool
+    {
+        return $user->hasAnyRole([
+            Role::DEVELOPER,
+            Role::EVENT_MANAGER,
+        ]);
```
Jamiras: Does a developer really need to be able to do this from filament? They should use the tools available to them (i.e. the DLL).

app/Filament/Actions/DisableLeaderboardAction.php
Jamiras: Filtering the disabled leaderboards out of the server response is a reasonable solution - and actually better than the current one where we send them even though they'll never activate.

Demoting them to unofficial would effectively do the same thing, but it would be a filter processed on the client (which is why I said that would be a much more extensive change).

When a LotM leaderboard is disabled, should it still be showed on the game page? I think that decision will drive the implementation.

app/Policies/LeaderboardPolicy.php
```diff
@@ -140,4 +148,21 @@ public function resetAllEntries(User $user): bool
...
+    }
+
+    // Only required until we are able to set leaderboards to 'unofficial'
```
Jamiras: Setting a leaderboard to "unofficial" would make it hidden on the game page too. Having the abillity to promote/demote leaderboards is on my radar, but it's not something I would want to try to push through any time soon.

app/Filament/Actions/DisableLeaderboardAction.php
wescopeland: I think what may be ideal is to add a `state` field/column to leaderboards that supports multiple possible values. I'm just shooting from the hip here, but here's a few ideas:
* "active" - probably the default value, at least initially?
* "disabled" - accepts no new entries, but still visible?
* "unofficial" - accepts no new entries, does not appear on the game page unless `?unpublished=true`?

app/Policies/LeaderboardPolicy.php
drisc: I added it as a QoL thing, a Developer could duplicate the board in Filament and make minor changes (for example adding a leaderboard for a level that they forgot without needing to open the game on a computer by changing the stage ID in the site editor).

If this is undesirable then I can remove Developers and only allow Event Mangers to access the functionality.

app/Filament/Actions/DisableLeaderboardAction.php
drisc: I have active and disabled actions coded locally and the `state` column accepts strings rather than being a bool so we can add actions for unofficial or other states at a later date.

Ideally the UX for each state would be:

- "active" - shows as normal on the game page and in emulator
- "disabled - shows with a visual distinction on the game page to show that it can't be submitted to anymore and doesn't show in the emulator
- "unofficial" - doesn't show on the game page unless viewing unpublished items and doesn't show in the emulator


I would like the LotM boards to still be visible to players after the fact for historical reasons.

I'm currently working through removing disabled boars from the patch and achievementsets responses, it already works for individual boards via `lbinfo` and returns a 404 if you try to fetch a disabled board.

app/Policies/LeaderboardPolicy.php
drisc: I'll remove this since we are going a different route for leaderboard state.

app/Filament/Actions/DisableLeaderboardAction.php
wescopeland: That sounds good to me.

For the field itself, let's use a native PHP enum.
For the game page UI/UX, feel free to ignore _gameInfo.blade.php_. We don't need to backport this functionality. We'll need to update the React game page. I recommend:
* Never show a disabled leaderboard in the sidebar (featured leaderboards).
* Only show disabled leaderboards in the leaderboard toggle view, and put them at the bottom of the list.
* Only add unofficial leaderboards into page props when `?unpublished=true` is set.
* Writing tests for all the behavior above - both sides of each conditional.

_leaderboardInfo.blade.php_ will also likely need a minor update.

app/Policies/LeaderboardPolicy.php
Jamiras: I thought we got rid of the site editor? Where can a user directly modify logic (outside of Rich Presence) on the site?

drisc: It looks like I missed that you can't actually edit stuff from the old Dev dropdown box anymore. I'll remove the role as it has no use.

app/Filament/Actions/DisableLeaderboardAction.php
drisc: All of these points have been implemented in latest.

I skipped the old `gameinfo.blade.php` view since we are cutting over to the new pages tomorrow.

 `leaderboardInfo.blade.php` and its accompanying sidebar list have been updated to follow the same display logic as the react game pages main list. There is also a disabled badge directly next to the title on the leaderboard information page.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameListViewSelectToggleGroup/GameListViewSelectToggleGroup.tsx
```diff
@@ -62,7 +62,7 @@ export const GameListViewSelectToggleGroup: FC = () => {
           </BaseTooltipTrigger>
         </BaseToggleGroupItem>
 
-        <BaseTooltipContent>{t('Leaderboards')}</BaseTooltipContent>
+        <BaseTooltipContent>{t('Active Leaderboards')}</BaseTooltipContent>
```
Jamiras: I think it's still appropriate to leave this button labelled as "Leaderboards".

resources/views/components/game/leaderboards-listing/index.blade.php
```diff
@@ -11,6 +11,17 @@
...
                 ->withTopEntry()
+                ->whereIn('state', [
+                    \App\Platform\Enums\LeaderboardState::Active->value,
+                    \App\Platform\Enums\LeaderboardState::Disabled->value,
+                ])
+                ->orderByRaw("
+                    CASE state
+                        WHEN '" . \App\Platform\Enums\LeaderboardState::Active->value . "' THEN 0
+                        WHEN '" . \App\Platform\Enums\LeaderboardState::Disabled->value . "' THEN 1
+                        ElSE 2
```
Jamiras: odd casing of `else`

tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php
```diff
@@ -237,6 +237,12 @@ public function testItIncludesLeaderboardData(): void
             'DisplayOrder' => -1, // !! hidden
             'Format' => 'VALUE',
         ]);
+        $leaderboard3 = Leaderboard::factory()->create([
+            'GameID' => $game->id,
+            'DisplayOrder' => 1,
+            'Format' => 'VALUE',
+            'state' => 'disabled',
```
Jamiras: Suggest using the enum here

tests/Feature/Connect/Actions/BuildClientPatchDataV2ActionTest.php
```diff
@@ -258,6 +258,12 @@ public function testItIncludesLeaderboardData(): void
             'DisplayOrder' => -1, // !! hidden
             'Format' => 'VALUE',
         ]);
+        $leaderboard3 = Leaderboard::factory()->create([
+            'GameID' => $game->id,
+            'DisplayOrder' => 1,
+            'Format' => 'VALUE',
+            'state' => 'disabled',
```
Jamiras: suggest using the enum here

tests/Feature/Connect/LeaderboardInfoTest.php
```diff
@@ -109,6 +109,19 @@ public function testLeaderboardInfo(): void
             ]);
     }
 
+    public function testDisabledLeaderboardInfo(): void
+    {
+        $game = $this->seedGame();
+        /** @var Leaderboard $leaderboard */
+        $leaderboard = Leaderboard::factory()->create([
+            'GameID' => $game->id,
+            'state' => 'disabled',
```
Jamiras: suggest using the enum here

tests/Feature/Connect/LeaderboardInfoTest.php
```diff
@@ -109,6 +109,19 @@ public function testLeaderboardInfo(): void
...
+        ]);
+
+        $this->get($this->apiUrl('lbinfo', ['i' => $leaderboard->id]))
+            ->assertStatus(404);
```
Jamiras: While connect shouldn't be asking for a disabled leaderboard due to it not being exposed by the patch endpoint. i don't see any reason why it couldn't. In fact, there's a small window immediately after the leaderboard is disabled where the client will know about it and may ask for the related data.

app/Models/Game.php
```diff
@@ -739,7 +739,8 @@ public function lastAchievementUpdate(): HasOne
      */
     public function leaderboards(): HasMany
     {
-        return $this->hasMany(Leaderboard::class, 'GameID', 'ID');
+        return $this->hasMany(Leaderboard::class, 'GameID', 'ID')
+        ->with('topEntry.user'); // eager load required to filter active leaderboards on game detail page
```
Jamiras: this should be indented.

why is this necessary? the filter should be on the leaderboard, not the topentry.user.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -781,7 +837,9 @@ private function getLeaderboardsCount(Game $game): int
             return 0;
         }
 
-        return $game->leaderboards->count();
+        return $game->leaderboards
+        ->where('state', LeaderboardState::Active)
+        ->count();
```
Jamiras: these should be indented.

app/Filament/Resources/LeaderboardResource.php
```diff
@@ -261,6 +270,7 @@ public static function table(Table $table): Table
             ->recordActions([
                 Actions\ActionGroup::make([
                     Actions\ActionGroup::make([
+                        CloneLeaderboardAction::make('clone_leaderboard'),
```
Jamiras, line 287: This also needs to be added to LeaderboardsRelationManager. I thought something was wrong with the Policy until I realized that I wasn't looking at the list where this menu applies.

I assume it would be easier for most users to find the leaderboard as a child of the game.

Image

app/Filament/Actions/CloneLeaderboardAction.php
```diff
@@ -0,0 +1,64 @@
...
+                ];
+            })
+            ->schema([
+                Forms\Components\TextInput::make('title')
+                    ->label('Title')
+                    ->required()
+                    ->maxLength(255),
+                Forms\Components\TextInput::make('description')
+                    ->label('Description')
+                    ->maxLength(1000),
```
Jamiras: description maxlength is also 255.

https://github.com/RetroAchievements/RAWeb/blob/e843acffac940afdc1ffd2b2946b1a909b3aad58/app/Filament/Resources/LeaderboardResource.php#L136-L137

app/Filament/Resources/LeaderboardResource.php
```diff
@@ -219,6 +220,14 @@ public static function table(Table $table): Table
...
                     ->sortable()
                     ->toggleable(),
+                Tables\Columns\SelectColumn::make('state')
+                    ->label('State')
+                    ->options([
+                        'active' => 'Active',
+                        'disabled' => 'Disabled',
+                        'unofficial' => 'Unofficial',
+                    ])
+                    ->selectablePlaceholder(false),
```
Jamiras: I don't think this should be inline editable. I would expect it to be a field in the edit page of the leaderboard (similar to achievements where it's in the sidebar):

Image

That would also solve the discrepency where it's not available on the nested leaderboards list on the game page.

Format, Lower is Better, and Display Order could be moved into a sidebar too for consistency. But for simplicity, I'd just put the State field between Description and Display Order.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameListViewSelectToggleGroup/GameListViewSelectToggleGroup.tsx
Jamiras, line 1: While the list is filtered for Unofficial, the count isn't updated for Unofficial.

Image

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -725,13 +726,68 @@ private function buildSetRequestData(Game $backingGame, ?User $user): ?GameSetRe
...
+            $leaderboards = $leaderboards->filter(function ($leaderboard) {
+                return $leaderboard->state === LeaderboardState::Active;
+            })->values();
+        } else {
+            // Filter based on showUnofficial flag
+            $leaderboards = $leaderboards->filter(function ($leaderboard) use ($showUnofficial) {
+                if ($showUnofficial) {
+                    // On the unpublished page: ONLY show Unofficial and Disabled
+                    return $leaderboard->state === LeaderboardState::Unofficial
+                        || $leaderboard->state === LeaderboardState::Disabled;
```
Jamiras: I wouldn't expect to see Disabled leaderboards on both pages. They're still "official", just not active.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
```diff
@@ -50,6 +55,20 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
                 {leaderboard.title}
               </a>
             </span>
+
+            {/* State badge */}
+            <BaseTooltip>
+              {leaderboard.state === 'disabled' && (
+                <BaseTooltipTrigger asChild>
+                  <span className="cursor-pointer rounded-full bg-embed px-2.5 py-0.5 text-xs text-neutral-500 opacity-50">
```
Jamiras: The border around the Disabled text isn't visible in Light mode

Image

app/Filament/Resources/LeaderboardResource.php
```diff
@@ -219,6 +220,14 @@ public static function table(Table $table): Table
                     ->label('Display Order')
                     ->sortable()
                     ->toggleable(),
+                Tables\Columns\SelectColumn::make('state')
```
Jamiras: Junior developers are currently able to change this field. They should not be able to.

app/Models/Leaderboard.php
```diff
@@ -68,12 +69,16 @@ class Leaderboard extends BaseModel implements HasVersionedTrigger
         'LowerIsBetter',
         'DisplayOrder',
         'trigger_id',
-
+        'state',
```
Jamiras, line 72: `state` should be added to the list of loggable fields (~line 112).

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
Jamiras: This should be `cursor-help`. `cursor-pointer` makes it look clickable.

app/Models/Game.php
drisc: This is no longer necessary, at one point during the features development I was getting lazy load errors due to a relation not being loaded when trying to filter the list. I've since worked around that that this is a leftover from that workaround. Removed in latest.

tests/Feature/Connect/Actions/BuildClientPatchDataActionTest.php
drisc: Changed in latest.

tests/Feature/Connect/Actions/BuildClientPatchDataV2ActionTest.php
drisc: Changed in latest.

tests/Feature/Connect/LeaderboardInfoTest.php
drisc: Changed in latest.

resources/views/components/game/leaderboards-listing/index.blade.php
drisc: Typo fixed in latest.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
drisc: Border has been added in latest.

app/Models/Leaderboard.php
drisc: Added in latest.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
drisc: Changed to cursor-help in latest.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Updated in latest.

app/Filament/Actions/CloneLeaderboardAction.php
drisc: Updated in latest.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameListViewSelectToggleGroup/GameListViewSelectToggleGroup.tsx
drisc: Updated in latest in the props action file, count now reflects the current views amount.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Disabled no longer appears in Unofficial in latest.

app/Filament/Resources/LeaderboardResource.php
drisc: This has been removed and turned into a text label in latest. The state edit has been moved to the leaderboard edit form.

app/Filament/Resources/LeaderboardResource.php
drisc: This is a now a text label in latest. Jr Devs can no longer edit it.

app/Filament/Resources/LeaderboardResource.php
drisc: The clone action has been added to the RelationManager in latest. This required adding Event Mangers to the Game `manage` policy so that they can see it in the sidebar. I have added a comment for this in the file.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSetToolbar/GameListViewSelectToggleGroup/GameListViewSelectToggleGroup.tsx
drisc: This has been reverted in latest, I removed the added translation string as well.

tests/Feature/Connect/LeaderboardInfoTest.php
drisc: I can remove the `where` in the query that filters to only `active` boards, thenadd a field to the LeaderboardData e.g. `LBState` which can then be used by clients to enable/disable items.

The `active` filter would stay in the patch data but this means that if the boards is disabled after the player loads the list the client can still access the data.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
```diff
@@ -50,6 +55,20 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
...
+                <BaseTooltipTrigger asChild>
+                  <span className="cursor-info rounded-full bg-embed px-2.5 py-0.5 text-xs text-neutral-500 opacity-50 light:border light:border-neutral-500 light:bg-white">
+                    {t(`Disabled`)}
+                    <BaseTooltipContent>
+                      {t('This leaderboard is currently disabled and not accepting new entries.')}
+                    </BaseTooltipContent>
+                  </span>
+                </BaseTooltipTrigger>
+              )}
+            </BaseTooltip>
```
wescopeland: UX: Something about this just isn't working for me, especially on mobile:
Screenshot 2025-12-09 at 7 47 03 AM

I think we should modify this instead:
Screenshot 2025-12-09 at 7 50 11 AM

It can get the tooltip and have the text-neutral-500.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
wescopeland, line 1: UX: Disabled leaderboards are appearing as interleaved within enabled leaderboards. I think they should always be put at the bottom of the list. Bonus points if we can somehow separate the two sections using a `<BaseSeparator />` in an elegant & subdued way.

Screenshot 2025-12-09 at 7 49 10 AM

app/Platform/Enums/LeaderboardState.php
```diff
@@ -0,0 +1,20 @@
...
+enum LeaderboardState: string
+{
+    /** The leaderboard is currently active and accepting entries. */
+    case Active = "active";
+
+    /** The leaderboard is disabled and not accepting new entries. */
+    case Disabled = "disabled";
+
+    /** The leaderboard is unofficial and not counted in official stats. */
+    case Unofficial = "unofficial";
```
wescopeland: Now that I'm grokking this for real and clicking on things, I think we should change this to:
```php
case Unpublished = "unpublished";
```
This aligns with the current direction of the website copy.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -725,13 +726,67 @@ private function buildSetRequestData(Game $backingGame, ?User $user): ?GameSetRe
     /**
      * @return Collection<int, LeaderboardData>
      */
-    private function buildLeaderboards(Game $game, ?User $user = null, ?int $limit = null): Collection
+    private function buildFeaturedLeaderboards(Game $game, ?User $user = null, ?int $limit = null): Collection
+    {
+        return $this->buildLeaderboards($game, $user, $limit, activeOnly: true, showUnofficial: false);
+    }
```
wescopeland: Given `buildFeaturedLeaderboards()` is only one line, it seems like an unnecessary abstraction around `buildLeaderboards()`.

app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
wescopeland, line 1: Similar to achievements, I should be able to bulk and individually modify the `state` of leaderboards.

Screenshot 2025-12-09 at 7 53 40 AM
Screenshot 2025-12-09 at 7 53 42 AM

app/Platform/Enums/LeaderboardState.php
drisc: Changed to this in latest.

*Note*: If you have any "unofficial" enums in the state column you will need to manually change them or the app will error.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Removed in latest, I've added the params to the featureLeaderboards definition instead.

tests/Feature/Connect/LeaderboardInfoTest.php
drisc: I have implemented this in latest, lbinfo now includes `LBState` which will reflect the state of the leaderboard for client use.

Test has also been updated to include the new item in the other tests Asserts and local testing shows the new return.

app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
Jamiras: That will be more important for bulk promote/demote. I'm not sure we want bulk disable.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
drisc: I have redesigned this in latest to place an X icon in the box and moved the tooltip to appear there as well.

It also has full test coverage.
CleanShot 2025-12-09 at 9  02 23

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
drisc: Changes included in latest. I have added the `<BaseSeparator />` and ensured that all disabled items are now placed below that separation.

CleanShot 2025-12-09 at 9  05 07

This also has full test coverage for each state scenario.

app/Connect/Actions/BuildClientPatchDataV2Action.php
wescopeland, line 1: Functionality: I can't help but feel if the `$flag` arg of `execute()` is given a value of `Unofficial`, we should still include unpublished leaderboards in the response. Same feedback applies to _BuildClientPatchDataAction.php_.

cc @Jamiras in case you have any objections or additions to this feedback

app/Connect/Actions/GetLeaderboardEntriesAction.php
```diff
@@ -114,6 +114,7 @@ protected function process(): array
                 'LBAuthor' => $leaderboard->developer?->User,
                 'LBCreated' => $leaderboard->Created?->format('Y-m-d H:i:s'),
                 'LBUpdated' => $leaderboard->Updated?->format('Y-m-d H:i:s'),
+                'LBState' => $leaderboard->state->value,
```
wescopeland, line 117: Testing: this should also be covered in `LeaderboardEntriesTest`.

app/Filament/Actions/CloneLeaderboardAction.php
```diff
@@ -0,0 +1,64 @@
...
+                    ->required()
+                    ->maxLength(255),
+                Forms\Components\TextInput::make('description')
+                    ->label('Description')
+                    ->maxLength(255),
+            ])
+            ->action(function (Leaderboard $leaderboard, array $data) use ($user) {
+                if (!$user->can('create', Leaderboard::class)) {
+                    return;
+                }
```
wescopeland: DX: This seems incorrect. We're already creating a `LeaderboardPolicy::clone()`. We shouldn't need an additional check on `LeaderboardPolicy::create()`. It should be implied.

app/Filament/Actions/CloneLeaderboardAction.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        parent::setUp();
+
+        /** @var User $user */
+        $user = Auth::user();
+
+        $this->label('Clone leaderboard')
+            ->icon('heroicon-s-clipboard-document-list')
+            ->color('success')
+            ->modalDescription('Cloning a leaderboard will create a new leaderboard with the same settings as the original, but with a new title and description. It will also be placed at the bottom of the display order. You can modify these fields below. Entries will not be cloned.')
```
wescopeland: UX: I think we can tighten this copy while preserving all the important information:
> This creates a duplicate of this leaderboard with the same settings, placed at the bottom of the display order. Entries will not be copied to the new leaderboard.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -781,7 +828,15 @@ private function getLeaderboardsCount(Game $game): int
...
-        return $game->leaderboards->count();
+        if (request()->boolean('unpublished')) {
+            return $game->leaderboards
+                ->where('state', LeaderboardState::Unpublished)
+                ->count();
+        } else {
+            return $game->leaderboards
+                ->where('state', LeaderboardState::Active)
+                ->count();
+    }
```
wescopeland: nit: this should be indented one more block over.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -781,7 +828,15 @@ private function getLeaderboardsCount(Game $game): int
...
 
-        return $game->leaderboards->count();
+        if (request()->boolean('unpublished')) {
+            return $game->leaderboards
+                ->where('state', LeaderboardState::Unpublished)
+                ->count();
+        } else {
+            return $game->leaderboards
+                ->where('state', LeaderboardState::Active)
+                ->count();
```
wescopeland: DX: We can remove this conditional.
```php
return $game->leaderboards
    ->where('state', request()->boolean('unpublished') ? LeaderboardState::Unpublished : LeaderboardState::Active)
    ->count();
```

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -781,7 +828,15 @@ private function getLeaderboardsCount(Game $game): int
             return 0;
         }
 
-        return $game->leaderboards->count();
+        if (request()->boolean('unpublished')) {
```
wescopeland: DX: Rather than `request()->boolean()`, we should leverage `$targetAchievementFlag`. This has the added benefit of reducing side effects in the action.

app/Filament/Actions/CloneLeaderboardAction.php
```diff
@@ -0,0 +1,64 @@
...
+                $clonedLeaderboard = $leaderboard->replicate([
+                    'entries_count', // Excludes computed attribute
+                    'top_entry_id',  // Also excludes this since it's specific to the original leaderboard
+                ]);
+                $clonedLeaderboard->Title = $data['title'];
+                $clonedLeaderboard->Description = $data['description'];
+                $clonedLeaderboard->author_id = $user->id;
+
+                // Set DisplayOrder to be last
+                $maxDisplayOrder = Leaderboard::max('DisplayOrder') ?? 0;
```
wescopeland: Bug: This calculates the max across all leaderboards in the entire database rather than just for this specific game.

database/migrations/2025_12_08_145911_update_leaderboarddef_status.php
wescopeland, line 1: Performance: given we're filtering by `state` in multiple queries, we should probably also add an index to the column.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
```diff
@@ -165,14 +166,62 @@ export const GameAchievementSet: FC<GameAchievementSetProps> = ({
...
               <>
-                {sortedLeaderboards.map((leaderboard, index) => (
-                  <LeaderboardsListItem
-                    key={`lbd-${leaderboard.id}`}
-                    index={index}
-                    isLargeList={isLargeLeaderboardsList}
-                    leaderboard={leaderboard}
-                  />
-                ))}
+                {(() => {
```
wescopeland: DX: Putting an IIFE in JSX tends to be an anti-pattern. It's effectively a component masquerading as a function. Claude writes React code like this; do not naively trust the React code that Claude writes.

We'll need a different solution here, one that probably involves creating a new component.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
```diff
@@ -165,14 +166,62 @@ export const GameAchievementSet: FC<GameAchievementSetProps> = ({
...
+                      leaderboard.state === 'active' || leaderboard.state === 'unpublished',
+                  );
+                  const disabledLeaderboards = sortedLeaderboards.filter(
+                    (leaderboard) => leaderboard.state === 'disabled',
+                  );
+
+                  // Check if the user is viewing unpublished leaderboards
+                  const isUnpublishedView = sortedLeaderboards.some(
+                    (leaderboard) => leaderboard.state === 'unpublished',
+                  );
```
wescopeland: DX: This isn't a good way to know if we're viewing unpublished stuff. There should be a page prop that can be leveraged instead.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
```diff
@@ -165,14 +166,62 @@ export const GameAchievementSet: FC<GameAchievementSetProps> = ({
...
-                    index={index}
-                    isLargeList={isLargeLeaderboardsList}
-                    leaderboard={leaderboard}
-                  />
-                ))}
+                {(() => {
+                  const activeStateLeaderboards = sortedLeaderboards.filter(
+                    (leaderboard) =>
+                      leaderboard.state === 'active' || leaderboard.state === 'unpublished',
+                  );
```
wescopeland: DX: `activeStateLeaderboards` is a misnomer. This includes "active" and "unpublished".

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.test.tsx
```diff
@@ -127,4 +129,26 @@ describe('Component: LeaderboardsListItem', () => {
     });
     expect(screen.getByText(/200/i)).toBeVisible();
   });
+
+  it('given the leaderboard is disabled, shows a tooltip', async () => {
+    // ARRANGE
+    const user = userEvent.setup();
```
wescopeland: Testing: There are a few other test cases where we do this, but `const user = userEvent.setup()` is typically unnecessary due to the customizations in our `render()` wrapper.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -725,13 +726,59 @@ private function buildSetRequestData(Game $backingGame, ?User $user): ?GameSetRe
...
+
+                    return match ($state) {
+                        LeaderboardState::Active => 0,
+                        LeaderboardState::Unpublished => 0,  // Same priority as Active
+                        LeaderboardState::Disabled => 1,
+                    };
+                },
+                fn ($active, $notActive) => $active->DisplayOrder <=> $notActive->DisplayOrder,
+            ])->values();
+        }
```
wescopeland: DX: I believe this can be significantly simplified down to only three statements.

```php
// untested pseudocode

$allowedLeaderboardStates = match (true) {
    $activeOnly => [LeaderboardState::Active],
    $unpublishedOnly => [LeaderboardState::Unpublished],
    default => [LeaderboardState::Active, LeaderboardState::Disabled],
};

$leaderboards = $game->leaderboards
    ->filter(fn ($lb) => in_array($lb->state, $allowedStates, true))
    ->values();

if (!$activeOnly) {
    // Active/Unpublished first, Disabled last, then by DisplayOrder.
    $leaderboards = $leaderboards->sortBy([
        fn ($lb) => $lb->state === LeaderboardState::Disabled ? 1 : 0,
        fn ($a, $b) => $a->DisplayOrder <=> $b->DisplayOrder,
    ])->values();
}
```

app/Connect/Actions/BuildClientPatchDataV2Action.php
Jamiras: AFAIK, the `f` flag is not actually used. And if it is, the caller is likely asking only for core achievements. I don't think there's any reason a client would only ever ask for just the Unofficial data.

The problem comes in the fact that not providing an `f` _should_ return both Published and Unofficial data, but the client doesn't know how to handle Unofficial leaderboards.

So we have three choices:
1) return Active leaderboards when f=0 or 3, and Unpublished leaderboards when f=5
2) return Active leaderboards when f=0 or 3, and nothing when f=5
3) always return only Active leaderboards.

I believe 3 is what's currently implemented. 1 is probably the best we can do for now. I need to figure out a good way to get Unofficial leaderboards in the response without forcing them onto clients that aren't ready for them. It will probably be a new flag where the client can indicate they're ready for unofficial leaderboards.

drisc: I’ll work towards implementing option 1. As it’s stands option 3 is currently implemented as the query is only looking for active boards.

app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
drisc: Added in latest.

app/Filament/Actions/CloneLeaderboardAction.php
drisc: This now calculates the max displayorder for the leaderboards connected GameID in latest.

database/migrations/2025_12_08_145911_update_leaderboarddef_status.php
drisc: Index added to migration in latest.

app/Filament/Actions/CloneLeaderboardAction.php
drisc: You're right, this is a leftover from when I was experimenting with the policy, it has been removed in latest.

app/Filament/Actions/CloneLeaderboardAction.php
drisc: Copy updated in latest

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Changed in latest.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: The new code now uses `$targetAchievementFlag` to check if the page is for active or unpublished items in latest.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Changed in latest.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
drisc: Lifted this out of the IIFE and formatted it the same as the achievements component within the page in latest.

I'll open a new PR to split both out to discreet components as an exercise in React once I'm done with this PR if that is ok?

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
drisc: Reworded to `visibleLeaderboards` in latest. Disabled leaderboards are visible as well but they require a new `const` so that they can be separated if present.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.test.tsx
drisc: This has been removed and the test now uses `userEvent` directly in latest.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: The pseudocode is correct and I have applied the simplification in latest.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
drisc: This has been changed to use the `isViewingPublishedAchievements` from the GameShowPageProps, the same as the BuildAction does in latest.

app/Connect/Actions/GetLeaderboardEntriesAction.php
drisc: The test for this Connect action is `LeaderboardInfoTest`. I did a quick search and found that the test you mentioned is related to the V1 `API_GetLeaderboardEntries` which has a json response of `Count`, `Total` and `Results`.

It did not return the state of the leaderboard so I have added `state` in latest.

app/Filament/Resources/LeaderboardResource.php
```diff
@@ -137,6 +144,18 @@ public static function form(Schema $schema): Schema
                             ->maxLength(255)
                             ->disabled(!$user->can('updateField', [$schema->model, 'Description'])),
 
+                        Forms\Components\Select::make('state')
+                            ->label('State')
+                            ->selectablePlaceholder(false)
+                            ->helperText('If set to Disabled, the leaderboard be prevented from activating. If set to Unofficial, the leaderboard will additionally be removed from public listings.')
```
wescopeland: There's a typo in this copy: "the leaderboard be prevented".

I'm not sure I agree with this sentence: "If set to Unofficial, the leaderboard will additionally be removed from public listings.". 'Unofficial' isn't an option, and when a leaderboard is unpublished, it's still viewable by others.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
```diff
@@ -35,9 +40,22 @@ export const LeaderboardsListItem: FC<LeaderboardsListItemProps> = ({
...
+            <BaseTooltip>
+              <BaseTooltipTrigger asChild data-testid="disabled-tooltip-trigger">
+                <LuX className="size-6 text-neutral-500 transition-colors group-hover:text-white light:text-neutral-400 light:group-hover:text-black" />
+              </BaseTooltipTrigger>
+              <BaseTooltipContent>
+                <div className="max-w-xs items-center text-center">
+                  {t('This leaderboard is currently disabled and not accepting new entries.')}
+                </div>
+              </BaseTooltipContent>
+            </BaseTooltip>
```
wescopeland, line 59: UX: Only wrapping the icon creates an interaction bug where only the icon is tooltipped but the whole container is tappable.

The whole container should be tooltipped.

https://github.com/user-attachments/assets/8a000fbf-4f22-4e06-8765-f67381ab41f3

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
wescopeland, line 1: Bug: When viewing unpublished stuff, the leaderboard claims it's disabled. It should appear as a normal leaderboard in this state that just happens to be in the unpublished asset list.

Image

app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
wescopeland: How did you test this? Can you share a screenshot of what you see when you click the "Bulk promote or demote" button?

drisc: Attached is a video of what I see when logged in as Event Manager. The bulk action and individual promote/demote action are only visible to those that have the ability to update the `state` column as defined in `$roleFieldPermissions` in `LeaderboardPolicy`.

I did forget to add the individual actions and they have been included in latest.

https://github.com/user-attachments/assets/24693053-759c-43bb-b157-4332080ed25b

wescopeland: This is what I saw yesterday, though I now realize it was fixed in https://github.com/RetroAchievements/RAWeb/pull/4210/changes/2f13302af506313316c1700546028b9eb5e80ceb.

Screenshot 2025-12-17 at 5 11 45 PM

The iconography is inconsistent with the achievements relation manager:
Screenshot 2025-12-18 at 7 28 56 PM

drisc: Ah ok, I'll fix that in my next round of commits. Getting to the other feedback now.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
drisc: Fixed in latest.

app/Filament/Resources/LeaderboardResource.php
drisc: Copy updated in latest.

resources/js/features/games/components/LeaderboardsListItem/LeaderboardsListItem.tsx
drisc: Fixed in latest. I used `span` due to `a` causing a hydration error and `div` causing a JSX warning about parent elements.

app/Connect/Actions/BuildClientPatchDataV2Action.php
drisc: It looks like `$flag` is specifically tied to `AchievementFlag` so to be able to individually call for both we would need a new input such as `l` for the patch data.

They use different types of backing enum currently as well, I could alter `LeaderboardState` to use 3 and 5 for Active and Unpublished but there isn't currently a numeric id for something that is Disabled so I don't know how that would interact with `AchievementFlag`.

wescopeland: Maybe it'd be best to punt on this piece for now until we've figured out how clients should handle this data.

drisc:
DB: Looks like JrDevs get an error when viewing the leaderboard list in Filament.

CleanShot 2025-12-19 at 3  00 18@2x

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland: nit: It's indented one block too far over. Current:
```php
return $game->leaderboards
        ->where('state', $isViewingPublishedAchievements ? LeaderboardState::Active : LeaderboardState::Unpublished)
        ->count();
```
Should be:
```php
return $game->leaderboards
    ->where('state', $isViewingPublishedAchievements ? LeaderboardState::Active : LeaderboardState::Unpublished)
    ->count();
```

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
```diff
@@ -243,14 +258,38 @@ export const GameAchievementSet: FC<GameAchievementSetProps> = ({
...
+                    initial={{ opacity: 0 }}
+                    animate={{ opacity: 1 }}
+                    transition={{ delay: visibleLeaderboards.length * 0.015 }}
+                  >
+                    <BaseSeparator data-testid="disabled-separator" />
+                  </motion.li>
+                )}
+
+                {/* Disabled Leaderboards */}
+                {isViewingPublishedAchievements &&
```
wescopeland: Recommend using ternaries for conditional rendering vs short-circuit `&&`. This should be a linting rule. I will add a note to configure one.

There are two benefits to conditionally rendering with a ternary:
* Short-circuit conditional rendering is prone to having React display "0" to the page. (search [this page](https://react.dev/learn/conditional-rendering) for "if the left side is" to see an example).
* Having a `: null` forces coverage of the else branch.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.test.tsx
wescopeland, line 1: Curious what happened here. The same three new test cases are duplicated four times each.

public/API/API_GetLeaderboardEntries.php
```diff
@@ -58,4 +59,5 @@
     'Count' => count($results),
     'Total' => $leaderboard->entries()->count(),
     'Results' => $results,
+    'state' => $leaderboard->state,
```
wescopeland: This feels misplaced. No other metadata about the leaderboard is returned on this endpoint - it's a collection endpoint.

This probably instead belongs on `API_GetGameLeaderboards`.

Additionally, we should probably name the field "State" instead of "state" for consistency.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -727,13 +728,41 @@ private function buildSetRequestData(Game $backingGame, ?User $user): ?GameSetRe
...
+
+        $leaderboards = $game->leaderboards
+            ->filter(fn ($leaderboard) => in_array($leaderboard->state, $allowedLeaderboardStates))
+            ->values();
+
+        if (!$activeOnly) {
+            // Sort: Active/Unpublished first, Disabled last, then by DisplayOrder.
+            $leaderboards = $leaderboards->sortBy([
+                fn ($leaderboard) => $leaderboard->state === LeaderboardState::Disabled ? 1 : 0,
+                fn ($active, $notActive) => $active->DisplayOrder <=> $notActive->DisplayOrder,
```
wescopeland: `$active` and `$notActive` are misleading parameter names. `sortBy` passes two elements for comparison, but they're not necessarily representative of the given element's active status. `$a` and `$b` are conventional parameter names for this operation.

app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
wescopeland, line 1: UX: Recommend moving the clone action up one space so the destructive actions are grouped together.

Image

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Fixed in latest.

database/migrations/2025_12_08_145911_update_leaderboarddef_status.php
```diff
@@ -0,0 +1,24 @@
...
+    {
+        Schema::table('LeaderboardDef', function (Blueprint $table) {
+            $table->string('state')->default('active')->after('DisplayOrder')->index();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('LeaderboardDef', function (Blueprint $table) {
+            $table->dropIndex('state');
```
wescopeland: This down migration fails.

The index was created without a name, but `dropIndex('state')` is looking for an index named exactly "state".

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.test.tsx
drisc: That is weird, I copied them out to simplify doing a merge on GitHub and then copied them back in again. The extras will be removed in the next round of commits.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.tsx
drisc: Separator and Disabled now use ternaries in latest.

resources/js/features/games/components/GameAchievementSetsContainer/GameAchievementSet/GameAchievementSet.test.tsx
drisc: Duplicates removed in latest.

app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
drisc: This is now placed above promote/demote in the list for those that can see it.

database/migrations/2025_12_08_145911_update_leaderboarddef_status.php
drisc: Migration has been fixed and now properly drops the `state` column on down action.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Variable names have been changed in latest.

public/API/API_GetLeaderboardEntries.php
drisc: I have removed these in latest, they will be added to `GetGameLeaderboards` shortly with the capitalised field name for consistency.

drisc: Added to `GetGameLeaderboards` endpoint and test updated in latest.

app/Filament/Actions/CloneLeaderboardAction.php
```diff
@@ -0,0 +1,61 @@
...
+                    ->maxLength(255),
+                Forms\Components\TextInput::make('description')
+                    ->label('Description')
+                    ->maxLength(255),
+            ])
+            ->action(function (Leaderboard $leaderboard, array $data) use ($user) {
+                $clonedLeaderboard = $leaderboard->replicate([
+                    'entries_count', // Excludes computed attribute
+                    'top_entry_id',  // Also excludes this since it's specific to the original leaderboard
+                ]);
```
wescopeland, line 48: Bug: `trigger_id` is also replicated.

Image


This means the cloned leaderboard shares the same versioned trigger - an edit to one may indirectly affect both (undesirable).

I think the latest version of the trigger needs to also be cloned with a new ID, beginning at version 1, and the cloned leaderboard's `trigger_id` needs to point to that version.

app/Filament/Actions/CloneLeaderboardAction.php
wescopeland, line 1: UX, minor: This submit button is inconsistent with other submit buttons in relation managers attached to the game resource.

This relation manager:

Image

Other relation managers:

Image

Image

drisc: Fixed in latest, it pulls the colour from the action so I had to add a new line for the modal submit action to recolour it like the default submit buttons on other modals.

Promote, Demote and Delete inherit their colours from their actions as well but they confirmation modals so it makes sense.

app/Filament/Actions/CloneLeaderboardAction.php
drisc: Fixed in latest, the action now uses the correct leaderboard as the `trigger_id` and has it's own versioning.

wescopeland (approved):
Approving on my end. I'd like to get a final pass / smoke test from @Jamiras before we merge.

I also want to leave a quick note for future PRs. The pivot from the `Mem` field hack to a proper `state` field was absolutely the right call, and, as expected, it added significant complexity. The feature touches Filament, the Connect API, React code, policies, and migrations - lots of different areas of the code that require different skills to work on.

That said, there was quite a bit of avoidable churn on this PR. I am far from perfect myself, but I want to re-emphasize a few things alluded to in our [contributor guidelines](https://github.com/RetroAchievements/RAWeb/blob/master/docs/CONTRIBUTING.md):
- **Try to remember to always manually test your changes.** A few issues (ie: the migration rollback failing) would've surfaced before I found them.
- **Try to remember to always self-review before pushing a commit.** I like to give every file's diff a quick individual scan in VSCode and stage files one at a time vs `git add .`. This probably would've caught some things like indentation inconsistency or test case duplication.

This isn't meant at all to discourage - just looking to make the process smoother for both of us. Thanks for sticking with this one.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -727,13 +728,41 @@ private function buildSetRequestData(Game $backingGame, ?User $user): ?GameSetRe
...
         }
 
+        $allowedLeaderboardStates = match (true) {
+            $activeOnly => [LeaderboardState::Active],
+            $showUnpublished => [LeaderboardState::Unpublished],
+            default => [LeaderboardState::Active, LeaderboardState::Disabled],
+        };
+
+        $leaderboards = $game->leaderboards
+            ->filter(fn ($leaderboard) => in_array($leaderboard->state, $allowedLeaderboardStates))
```
Jamiras: Can we use `whereIn` here?

app/Platform/Actions/BuildGameShowPagePropsAction.php
Jamiras, line 1: I don't think it's blocking for this PR, but if a game has unpublished leaderboards, but no unpublished achievements, there isn't a link to the unpublished version of the game page.

We might need to change the label too, as it specifically says "Unpublished Achievements" now.

Image

app/Policies/LeaderboardPolicy.php
```diff
@@ -73,9 +80,10 @@ public function update(User $user, Leaderboard $leaderboard): bool
     public function updateField(User $user, ?Leaderboard $leaderboard, string $fieldName): bool
     {
         $roleFieldPermissions = [
-            Role::DEVELOPER_JUNIOR => ['Title', 'Description', 'Format', 'LowerIsBetter', 'DisplayOrder'],
-            Role::DEVELOPER => ['Title', 'Description', 'Format', 'LowerIsBetter', 'DisplayOrder'],
+            Role::DEVELOPER_JUNIOR => ['Title', 'Description', 'Format', 'LowerIsBetter', 'DisplayOrder', 'state'],
+            Role::DEVELOPER => ['Title', 'Description', 'Format', 'LowerIsBetter', 'DisplayOrder', 'state'],
             Role::WRITER => ['Title', 'Description'],
+            Role::EVENT_MANAGER => ['Title', 'Description', 'DisplayOrder', 'state'],
```
Jamiras, line 86: I'm not sure how difficult it would be to implement, but I think EventManagers should be limited to toggling between Published and Demoted. I'm not sure they should be able to toggle between Published and Unpublished.

It's probably fine for now. I don't expect the feature to be abused.

Jamiras (approved):
Noted a couple of minor things. I don't think any of it is blocking.

app/Policies/LeaderboardPolicy.php
drisc: I’d say that Event Managers need to be able to unpublish as well as disable. If a board doesn’t need to be visible on a game at all, such as if we swap to a different board due to issues for LotM then being able to fully hide it from casual view is desirable. State changes are visible in the audit log as well so if someone is abusing it we will have a record.

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland: I agree this is going to be an issue; let's punt on this in this branch for now.

There are broader UX concerns that will need to be considered for how we go about fixing this. Right now devs use the unpublished achievement count (1 in this screenshot) as a visual cue, and that will likely need to somehow change as well.

drisc: We can most likely reuse the indication component from the main body of the page to show achievement and leaderboard counts but without interaction.

wescopeland: If we show both counts, we still require some way to discern between both counts while keeping things accessible. I'm not even sure showing both counts is desirable. We should probably let this new functionality bake for a little while and observe how people respond to it first.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: Indeed we can, updated in latest.

app/Platform/Actions/BuildGameShowPagePropsAction.php
drisc: What I meant by reusing a component was adding this to the "Unpublished" link in place of the current single number: 

CleanShot 2025-12-27 at 12  04 20


If we remove "Achievements" from the end and just keep "Unpublished" then it accurately indicates what is behind the link since leaderboards will now be included. It should also give more space for translations even with the new icons/counters.

Though I also agree that we should let the new functionality sit for a while to gauge the usage.

Issue #4212, opened 2025-12-08
Soft deleted achievements still show as "Unpublished" on game2 pages


suspect15:
**Describe the bug**
If an achievement is soft deleted by a developer, it is still reported as an "Unpublished" achievement.

**To Reproduce**
1. Go to https://retroachievements.org/game/16881
2. Click on "Contribute" and notice it lists 1 unpublished achievement
3. Check the Unpublished achievement list and there are none since I deleted the one that was uploaded

**Expected behavior**
Unpublished achievements should only be undeleted ones

PR #4213, opened 2025-12-08
fix(achievement): update game metrics on soft delete

wescopeland:
Prevents #4212 from recurring.

When an achievement is soft deleted, game metrics currently aren't being recalculated, so the denormalized count of unpublished achievements doesn't change.

This PR causes game metrics to be recalculated on achievement soft deletes.


Jamiras (commented):
Untested. Looks correct.

PR #4217, opened 2025-12-09
(feature) automatically email RACheats for improbable times

Jamiras:
https://discord.com/channels/476211979464343552/1002687577280757770/1447813476339679314

image

### DIFF
--- a/tests/Feature/Platform/Actions/UpdatePlayerGameMetricsActionTest.php
+++ b/tests/Feature/Platform/Actions/UpdatePlayerGameMetricsActionTest.php
@@ -286,7 +286,7 @@ class UpdatePlayerGameMetricsActionTest extends TestCase
 
         $thread = MessageThread::first();
         $this->assertNotNull($thread);
-        $this->assertEquals("Suspicious User: $user->display_name", $thread->title);
+        $this->assertEquals("Suspicious Beat Time for User: $user->display_name", $thread->title);
 
         $participant = MessageThreadParticipant::first();
         $this->assertNotNull($participant);


app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -344,4 +363,41 @@ public function beatProgressMetrics(PlayerGame $playerGame, AchievementSet $core
...
+            return;
+        }
+
+        $thread = null;
+
+        $moderationReport = UserModerationReport::query()
+            ->where('reportable_type', ModerationReportableType::PlayerBeatTime)
+            ->where('reported_user_id', $user->id)
+            ->where('reporter_user_id', $systemUser->id)
+            ->first();
```
Jamiras, line 383: I'm using `UserModerationReport` to tie multiple reports against a single user together. I'm not sure if that will cause duplicate messages (one to RACheats, one to mod-forum), or if it will cause any other unintended side-effects.

If that's not desirable, I could try to match the message subject, but there's a small opportunity for that to have been created outside of the automatic process, and we wouldn't want to merge in with those.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -344,4 +363,41 @@ public function beatProgressMetrics(PlayerGame $playerGame, AchievementSet $core
...
+
+    private function createPotentialCheaterMessageThread(User $user, PlayerGame $playerGame, string $message): void
+    {
+        $systemUser = User::find(Comment::SYSTEM_USER_ID);
+        $raCheats = User::whereName('RACheats')->first();
+        if (!$systemUser || !$raCheats) {
+            Log::info($message);
+
+            return;
+        }
```
wescopeland, line 375: I'm worried this might flood their inbox.

Can we direct these either to another inbox (another user) or to RACheats with a different webhook? I can help if needed.

Jamiras: That's the primary reason I wanted these to be a single thread per user. A single user may trigger this several times, but I doubt there's going to be many unique users triggering this in any given period of time.

The logic is loose enough that it shouldn't have many false positives. If it is too annoying, we can tune the threshold. I'm actually concerned that it might be too loose and a lot of things will slip through. I'm relying on the user doing this across multiple games, hoping at least one of them triggers this.

wescopeland: Right. I guess my main concern is these are feed-related messages that we're going to have flow into an inbox of user-submitted stuff. I think we should separate feed stuff from user-submitted stuff.

Jamiras: We could certainly eliminate the "email" entirely and just funnel these straight to discord. If you want to pursue that direction, the logic is really just [this chunk of code](https://github.com/RetroAchievements/RAWeb/pull/4217/files#diff-f2a72a17ab7df5784276ac764ab96daa02dfbd3f530f60ada884acc86a0d31faR201-R209). Everything else is for managing the mail.

wescopeland: Spent quite a while thinking on this. Per [our recent chat](https://discord.com/channels/@me/1065706550318084117/1447341675109421076), I'd like to propose an alternative design: https://github.com/RetroAchievements/RAWeb/compare/master...wescopeland:RAWeb:alert-system

I'm trying to solve for a few things with this design:
* It's probably desirable for us to have numerous different kinds of alerts.
* The alert logic should probably not be deeply embedded into actions where stuff is being processed - it's more of a fire-and-forget side effect rather than core business logic.
* Different alert types should be easily routable to different Discord channels via simple config.
* Adding new alert types should be a predictable & consistent pattern, with very minimal cognitive load. Just write the detection logic and message format with as minimal boilerplate as possible.

With the proposed design, creating new alerts is pretty straightforward:
1. Create a new alert class like _app/Support/Alerts/SomeNewAlert.php_. The class will have `extends Alert` - so they all have a predictable convention-based interface.
2. Update _config/services.php_ to add a new webhook URL within the new "alerts_webook" list. The key is auto-derived from the class name, ie: `SomeNewAlert` -> `some_new`.
3. Create a trigger listener like _app/Support/Alerts/Listeners/TriggerSomeNewAlert.php_. This contains the detection logic (the "chunk of code" mentioned in your previous message).
5. Register the trigger listener in `EventServiceProvider`.
6. Add the webhook env var to the environment.

And that's it - we can now add alerts for all sorts of things.

Jamiras: I like the idea of discrete listeners doing the logic and forwarding the message. 

I'm not sure I like adding the beatenChanged flag to the PlayerMetricsUpdated event. This should probably trigger off the PlayerGameBeaten event.

wescopeland: That makes a lot of sense and was pretty easy to change. Check out this commit for the diff: https://github.com/RetroAchievements/RAWeb/commit/540b986dfddd4ccd0edda9ba405ffc68debb1912

Jamiras:
Closing in favor of #4261

PR #4223, opened 2025-12-10
(feature) add game metadata to seeders

Jamiras:
Developer/Publisher/Genre/Release Dates/Series.

image

image

image

image

image

Indirectly populates the [Central - Genres] and [Central - Series] pages as a result.


database/seeders/GameReleasesSeeder.php
```diff
@@ -0,0 +1,155 @@
...
+                    $range = [2004, 2014];
+                    break;
+                default:
+                    // unhandled. weight towards the newest date
+                    $year = (int) sqrt(random_int(1976 * 1976, 2012 * 2012));
+                    break;
+            }
+            if ($year === 0) {
+                // weight ranged releases towards the initial date
+                $year = $range[1] - (int) sqrt(random_int(0, pow($range[1] - $range[0] + 1, 2)));
```
wescopeland: This can technically produce years outside the valid range. For example, NES is [1983, 1995]:
* `$range[1] - $range[0] + 1 = 13`
* `pow(13, 2) = 169`
* `sqrt(169) = 13`
* `1995 - 13 = 1982 // ??`
I'm not sure if this is actually important.

database/seeders/GameReleasesSeeder.php
```diff
@@ -0,0 +1,155 @@
...
+                // 95% (105/110) chance of only being year for 1976
+                $release->released_at_granularity = ReleasedAtGranularity::Year;
+            } elseif ($year < 1988 && random_int(0, (1990 - $year) * (1991 - $year)) > 9) {
+                // 25% (3/12) chance of month granularity for 1987
+                // 70% (21/30) chance of month granularity for 1985
+                // 87% (61/70) chance of month granularity for 1982 (after 25% chance of it not being year only)
+                $release->released_at = $release->released_at->addMonths(rand(0, 11));
+                $release->released_at_granularity = ReleasedAtGranularity::Month;
+            } else {
+                $release->released_at = $release->released_at->addDays(rand(0, 364));
```
wescopeland, line 103: Consistency: some places use `rand()`, others use `random_int()`. I'm not sure if their behavior is different or if it's worth conforming the whole seeder to use just one.

database/seeders/GameSetsSeeder.php
```diff
@@ -31,5 +37,152 @@ public function run(): void
...
+                    'child_game_set_id' => GameSet::GenreSubgenreHubId,
+                ]);
+
+                $genreHubIds[] = [$subGenre->id, $subGenreTitle];
+            }
+        }
+
+        $developerHubIds = [];
+        $developerHub = GameSet::create(['title' => "[Central - Developer]", 'type' => GameSetType::Hub]);
+        $num_to_create = sqrt(Game::count()) + random_int(0, 10) + random_int(0, 2) + random_int(0, 2) + random_int(0, 2);
```
wescopeland, line 97: Intellectual curiosity: why not just do `sqrt(Game::count()) + random_int(0, 16)`?

database/seeders/GameReleasesSeeder.php
Jamiras: It's an off by one error. 168 would result in 1983. The chances of that happening are less than 1% though (1/169). I'll correct for it anyway.

database/seeders/GameReleasesSeeder.php
Jamiras: According to google, `rand()` is psuedo-random and `random_int()` is cryptographically secure. For our purposes, we don't care about the uniqueness of the results, so the faster `rand()` should probably be used.

From my perspective, I assumed `rand` returned a double, so I used `random_int` any time I wanted a whole number.

database/seeders/GameSetsSeeder.php
Jamiras: To weight the curve. It's the difference between rolling a d20 or 3d6. With a d20, you're just as likely to get a 1 or a 20 as a 12. With 3d6, you're more likely to end up with a number in the middle.

wescopeland: Thanks, I didn't think of it that way.

PR #4225, opened 2025-12-10
(refactor) create GetUserProgressForConsoleAction

Jamiras:
also fixes
> GetAllUserProgress(): Argument #1 ($user) must be of type App\Models\User, null given, called in /home/forge/retroachievements.org/releases/2025-12-10T120953-2025.12.10-77f658f5/public/dorequest.php on line 224


public/dorequest.php
```diff
@@ -219,11 +221,6 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
     /*
      * Global, no permissions required
      */
-    case "allprogress":
```
Jamiras, line 222: Technically, this does require authentication. `$user` only gets set by processing the token. I've made that more clear with this refactor by inheriting from `BaseAuthenticatedApiAction`.

And I can't think of any reason we'd want to query all progress for a console for some other user (particuarly through the Connect API).

wescopeland (approved):

PR #4232, opened 2025-12-10
fix(game): use correct code notes link in sidebar

wescopeland:
https://discord.com/channels/310192285306454017/310195377993416714/1448445288011403266


wescopeland:
Maybe this change should only be scoped to "exclusive" typed subsets.

Jamiras:
Another long-term problem is going to be the trigger decoder. It's currently reading the notes for the game associated to the achievement/leaderboard. If the client is only updating the core game notes, the trigger decoder won't be able to find them when viewing the subset achievement definition.

wescopeland:
Yea. It's almost like code notes should be associated with _hashes_ rather than game or set IDs.

But I don't know what that means in respect for either the toolkit or the game page URLs.

Jamiras:
I think we need some way to point a GameData record at another game_id for purposes of code notes. Then a subset/hack can choose to just use the code notes from the base game, or point at their own unique code notes. Then the UI just has to use that to determine where the link goes. The client would still only use the set associated to the identified hash, regardless of which additional subsets were retrieved.

PR #4234, opened 2025-12-11
feat(manage/achievements): add 'Move to top', 'Move to bottom', and manual display order editing

wescopeland:
This PR adds two new features to the management app that are shown when the user is managing a game's achievements. These features are based on real user feedback.

Additionally, thanks to the upgrade to Filament 4, we can now bump the max page size from 150 to 400. Filament sends significantly smaller Livewire payloads now (though this will not be reflected in a local environment).

**Quick Move Buttons**
There are now buttons to instantly move a given achievement to the top or bottom of the list.
Screenshot 2025-12-10 at 6 58 34 PM

**Manual display order editing**
A user can now decide whether they want to drag+drop to reorder, or manually edit display orders. Both cannot be active at the same time.

Screenshot 2025-12-10 at 6 59 45 PM

Screenshot 2025-12-10 at 6 59 59 PM

### DIFF
--- a/app/Connect/Actions/GetUserProgressForConsoleAction.php
+++ b/app/Connect/Actions/GetUserProgressForConsoleAction.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Connect\Actions;
+
+use App\Connect\Support\BaseAuthenticatedApiAction;
+use App\Models\Game;
+use App\Models\PlayerGame;
+use App\Models\User;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Http\Request;
+
+class GetUserProgressForConsoleAction extends BaseAuthenticatedApiAction
+{
+    protected int $consoleId;
+
+    public function execute(User $user, int $consoleId): array
+    {
+        $this->user = $user;
+        $this->consoleId = $consoleId;
+
+        return $this->process();
+    }
+
+    protected function initialize(Request $request): ?array
+    {
+        if (!$request->has(['c'])) {
+            return $this->missingParameters();
+        }
+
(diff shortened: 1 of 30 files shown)


app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -177,10 +183,35 @@ public function table(Table $table): Table
...
+                        ->label('Move to Top')
+                        ->icon('heroicon-o-arrow-up')
+                        ->action(fn (Achievement $record) => $this->moveAchievementToPosition($record, 'top'))
+                        ->visible(fn () => $this->canReorderAchievements() && !$this->isEditingDisplayOrders),
+
+                    Actions\Action::make('move-to-bottom')
+                        ->label('Move to Bottom')
+                        ->icon('heroicon-o-arrow-down')
+                        ->action(fn (Achievement $record) => $this->moveAchievementToPosition($record, 'bottom'))
+                        ->visible(fn () => $this->canReorderAchievements() && !$this->isEditingDisplayOrders),
```
Jamiras, line 213: Is there a reason these are only visible when not reordering (they don't appear when drag reordering or when editing order values).

Jamiras (approved):

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
wescopeland: Yep. Filament unmounts the menu completely during drag reordering (internal framework implementation detail), and when manual editing it gets really complex to keep the in-memory table state in sync with the table's actual values.

It's probably still desirable to support this during manual order editing, but seems quite complex. I may open a separate PR for this in the future.

PR #4236, opened 2025-12-11
feat: wasm hash matcher

m1guelpf:
Adds a file upload button to the game hashes page, which will calculate the hash for a ROM and see whether it matches any of the known hashes.

- [x] Add hashing logic
- [ ] UI


wescopeland:
Hi @m1guelpf 👋 

Thanks for putting up this draft - the approach makes sense technically.

We've, unfortunately, had to make some changes recently to this page in response to a legal/compliance request. For now, we need to keep the hashes page as an informational list only.

Even given the hashing is done entirely client-side, adding a `type="file"` input is likely to upset [insert megacorp here], which is something that may be perilous to support on RAWeb at the moment.

PR #4246, opened 2025-12-13
feat(manage): add 'View Legacy Comments' links to games and hubs

wescopeland:
This PR implements two suggestions from https://github.com/RetroAchievements/RAWeb/discussions/4109.

Similar to the hashes resource, games and hubs now have a link to view legacy comments.

It might be a good idea to consider preventing users from manually writing new comments of these kinds, but that's probably a separate discussion.

Screenshot 2025-12-12 at 6 58 14 PM

Screenshot 2025-12-12 at 6 58 22 PM

### DIFF
--- a/app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
+++ b/app/Filament/Resources/GameResource/RelationManagers/LeaderboardsRelationManager.php
@@ -13,6 +13,7 @@ use App\Platform\Enums\ValueFormat;
 use BackedEnum;
 use Filament\Actions\Action;
 use Filament\Actions\ActionGroup;
+use Filament\Notifications\Notification;
 use Filament\Resources\RelationManagers\RelationManager;
 use Filament\Schemas\Schema;
 use Filament\Tables;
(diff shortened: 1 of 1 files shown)


app/Filament/Resources/GameResource.php
```diff
@@ -87,6 +89,26 @@ public static function infolist(Schema $schema): Schema
...
+                            ->url(function (Game $record) {
+                                return route('game.modification-comment.index', ['game' => $record->id]);
+                            })
+                            ->openUrlInNewTab()
+                            ->visible(function (Game $record) {
+                                return Comment::where('ArticleType', ArticleType::GameModification)
+                                    ->where('ArticleID', $record->id)
+                                    ->exists();
+                            }),
+                    ])
```
Jamiras: Since this is specifically for viewing GameModification comments, shouldn't the button be on the Audit Log page? Is that even practical?

If not, it should be renamed. I see "view legacy comments" and I don't expect to be shown audit tracking information.

Jamiras: Same comment applies to the hub button.

Jamiras: Hashes is probably fine where its at since it's on a subtab already, but the label could still be improved.

wescopeland: Tried to incorporate all this in latest:
Screenshot 2025-12-13 at 7 40 12 AM

Additionally, I changed the hashes buttonlink label to "View Legacy Hash Maintenance Comments". It's kinda verbose, but I couldn't come up with anything better. Also, that buttonlink no longer appears if the count is 0.

Jamiras (approved):

PR #4249, opened 2025-12-13
feat(manage): improve game hash change logging, apply audit log UX improvements

wescopeland:
This PR adds a `LogGameHashActivityAction`, which is modeled after `LogGameReleaseActivityAction` and serves a similar purpose: when hashes are linked, unlinked, or modified, an activity log record will now propagate to the game's audit log.

Additionally, three UX improvements have been made to audit log pages:
1. The default page size has been increased from 5 to 50.
2. Rows where "Old" and "New" values are identical will no longer be shown.
3. Records where no rows are shown (eg: declaring some action took place like an achievements reorder) will no longer have table headings shown.

The new events for game hashes are shown in these before and after images, which demonstrate the audit log UX changes.

**Before**
Screenshot 2025-12-13 at 9 43 55 AM

**After**
Screenshot 2025-12-13 at 9 43 19 AM


app/Filament/Resources/GameResource/Pages/Hashes.php
```diff
@@ -176,38 +176,17 @@ public function table(Table $table): Table
...
-                                ],
-                                'old' => [
-                                    'name' => $name,
-                                    'md5' => $md5,
-                                    'labels' => $labels,
-                                ],
-                            ])
-                            ->event('unlinkedHash')
-                            ->log('Unlinked hash');
+                        $gameHash->forceDelete();
```
Jamiras, line 206: Since this is being force deleted, can we capture the title and/or description of the hash in the audit log?

wescopeland: We can. Done in latest:
Screenshot 2025-12-16 at 5 03 40 PM

Jamiras (approved):

PR #4251, opened 2025-12-13
feat(hubs): add support for `sort_title` values

wescopeland:
This PR adds a nullable `sort_title` to the `game_sets` table which functions very similarly to the one in `GameData`. This is a request from https://github.com/RetroAchievements/RAWeb/discussions/4109.

```bash
sail artisan migrate
sail artisan ra:platform:write-game-set-sort-titles
```

Only users with either an admin role or game editor role are allowed to edit hub sort titles.

Additionally, the following changes have been made:
* Hub pages now use the new `sort_title` field to sort related hubs.
* The "Sort Title" field is now completely hidden from display for both games and hubs in view & edit modes if the user doesn't have permission to update it.
* When editing a resource title in Filament, the sort title now automatically refreshes if the value was recomputed.
* Games are now sortable by title in Filament (using `sort_title`).

Screenshot 2025-12-13 at 11 01 27 AM


tests/Feature/Platform/Actions/WriteGameSetSortTitleActionTest.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        // Arrange
+        $gameSet = GameSet::factory()->create(['title' => '[Series - The Legend of Zelda]', 'sort_title' => null]);
+
+        // Act
+        (new WriteGameSetSortTitleAction())->execute($gameSet, $gameSet->title);
+        $gameSet->refresh();
+
+        // Assert
+        $this->assertEquals('series - the legend of zelda', $gameSet->sort_title);
```
Jamiras: Isn't the whole point of this feature to eliminate the "the"?

I guess the user could remove it themselves.

wescopeland: It'd probably be best if we added a thin layer of preprocessing to these. I've done that in latest and also updated this test case.

PR #4253, opened 2025-12-13
feat(SubmitGameTitleAction): auto-attach new 'subset games' to their parent game

wescopeland:
This PR makes an enhancement to `SubmitGameTitleAction`.

When a developer adds a new "subset game" to the database (a game with a title like 'Some Game [Subset - Bonus]'), the new game's achievement set will automatically be attached to the parent game (if found) with a type of `Exclusive`.

We intentionally make no attempt whatsoever to try to discern what the set type is, even if the given title for the subset is "Bonus". This is purely meant to make navigation for players on the website less confusing (currently, they're seeing the same game across two game pages because the subset game is commonly orphaned).


wescopeland:
Merging this to address https://discord.com/channels/310192285306454017/1469892331309236437.

PR #4254, opened 2025-12-13
(feat) add preview text for single-post topics

Jamiras:
https://discord.com/channels/310192285306454017/310195377993416714/1449412219585695744

When building the daily summary email, if a notification only has one new post, inline the first 200 characters of the post.

image

### DIFF
--- a/app/Connect/Actions/GetUserProgressForConsoleAction.php
+++ b/app/Connect/Actions/GetUserProgressForConsoleAction.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Connect\Actions;
+
+use App\Connect\Support\BaseAuthenticatedApiAction;
+use App\Models\Game;
+use App\Models\PlayerGame;
+use App\Models\User;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Http\Request;
+
+class GetUserProgressForConsoleAction extends BaseAuthenticatedApiAction
+{
+    protected int $consoleId;
+
+    public function execute(User $user, int $consoleId): array
+    {
+        $this->user = $user;
+        $this->consoleId = $consoleId;
+
+        return $this->process();
+    }
+
+    protected function initialize(Request $request): ?array
+    {
+        if (!$request->has(['c'])) {
+            return $this->missingParameters();
+        }
+
+        $this->consoleId = $request->integer('c', 0);
+
+        return null;
+    }
+
+    protected function process(): array
+    {
+        $games = Game::where('ConsoleID', $this->consoleId)
+            ->where('achievements_published', '>', 0)
+            ->get();
+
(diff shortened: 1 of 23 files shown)


app/Community/Actions/SendDailyDigestAction.php
```diff
@@ -67,13 +69,57 @@ public function execute(User $user): void
...
+                $post = $posts[$delayedSubscription->subject_type->value][$delayedSubscription->subject_id] ?? null;
+                if ($post) {
+                    $summary = match ($type) {
+                        SubscriptionSubjectType::ForumTopic->value => nl2br(Shortcode::stripAndClamp($post->latestComment->body, previewLength: 200, preserveWhitespace: true)),
+                        default => null,
+                    };
+                    $displayName = match ($type) {
+                        SubscriptionSubjectType::ForumTopic->value => $post->latestComment->user->display_name,
+                        default => null,
+                    };
```
wescopeland, line 114: There may be a latent bug here, but I think my analysis could use another pair of eyes. I think it's only not surfacing because we're just matching for forum topics and nothing else at the moment.

I believe in the `foreach ($singleItems as $singleItem) { ... }` block, the value for `$type` is stale. I'm envisioning this scenario:
1. The user has a `ForumTopic` subscription with 1 new post.
2. The user also has a `FooBar` subscription with 1 new comment.
3. `$ids` ends up as `['ForumTopic' => [123], 'FooBar' => [456]]`.
4. After the first foreach, `$type = 'FooBar'`.
5. When processing the forum topic stuff in the second foreach, `$post` is correctly fetched, but `match ($type)` hits `default => null` because `$type = 'FooBar'`.
6. This results in no preview being added for the forum reply even though it should have one.

I think replacing `$type` with `$delayedSubscription->subject_type->value` on both match expressions probably resolves the issue.

resources/views/mail/community/daily-digest.blade.php
```diff
@@ -4,17 +4,29 @@
...
 @foreach ($notificationItems as $notificationItem)
-@switch ($notificationItem['type'])
+@php
+    switch ($notificationItem['type']) {
+        case 'ForumTopic':
+            $clause = "in [${notificationItem['title']}](${notificationItem['link']}) (forum topic)";
+            $postType = 'post';
+            break;
+        default:
+            $lowerType = tolower($notificationItem['type']);
```
wescopeland: typo: `strtolower`

app/Community/Actions/SendDailyDigestAction.php
Jamiras: You're right. That's leftover from copy/pasting the above match statement. Fixed.

resources/views/mail/community/daily-digest.blade.php
Jamiras: fixed.

PR #4259, opened 2025-12-14
(feature) also delay notifications for comments

Jamiras:
adds logic to delay implicit comment subscriptions to the daily email.

image


app/Community/Actions/SendDailyDigestAction.php
```diff
@@ -106,11 +123,11 @@ public function execute(User $user): void
                 if ($post) {
                     $summary = match ($type) {
                         SubscriptionSubjectType::ForumTopic->value => nl2br(Shortcode::stripAndClamp($post->latestComment->body, previewLength: 200, preserveWhitespace: true)),
-                        default => null,
+                        default => nl2br(mb_strlen($post->Payload) > 200 ? mb_substr($post->Payload, 200) . '...' : $post->Payload),
```
wescopeland: Bug: this slices the string rather than showing the first 200 chars.

It probably should be `mb_substr($post->Payload, 0, 200)`.

Jamiras: Ah, yes. Good eye. I got sloppy transposing this from the Shortcode code.

Fixed.

PR #4260, opened 2025-12-14
feat: add support for achievement groups

wescopeland:
This PR adds support for grouping achievements within a single achievement set. It is mostly leveraging functionality from the existing events page and how event achievements can be grouped.

This functionality is meant to later support a small piece of the back-end for [Quests](https://github.com/RetroAchievements/RAWeb/discussions/4194#discussioncomment-15241609). Also: https://github.com/RetroAchievements/RAWeb/discussions/745.

https://github.com/user-attachments/assets/83983b3e-e42c-4770-bca4-b0a8a973bb41

_A migration is included in this PR._

**How it works:**
After running `sail artisan migrate`, you can create groups for achievements in the management app, assuming you have one of these roles: `ADMINISTRATOR`, `DEV_COMPLIANCE`, `GAME_EDITOR`, or `QUALITY_ASSURANCE`.

Screenshot 2025-12-14 at 10 15 19 AM

Groups are very much like folders. Clicking "Manage groups" allows you to define what groups are available to use for the achievement set:

Screenshot 2025-12-14 at 10 16 00 AM

The achievement set I'm going to work on for this demo is [Final Fantasy I & II: Dawn of Souls](http://localhost:64000/manage/games/762), a compilation title featuring the first two FF games in a single achievement set. I'll create groups for both games in the collection:

Screenshot 2025-12-14 at 10 18 44 AM

Now, it's just a matter of selecting achievements and moving them to groups (notice the "Assign to group" button that appears - this is a bulk action):
Screenshot 2025-12-14 at 10 20 09 AM
Screenshot 2025-12-14 at 10 20 42 AM

Once everything is assigned, I see the groups on the achievement set page:
Screenshot 2025-12-14 at 10 21 20 AM

To add some visual flair, an icon is automatically picked for each group. It goes in this priority:
- The last Win Condition achievement in the group, or
- The last Progression achievement in the group, or
- The highest value (by points) achievement in the group, or
- The first achievement in the group (reaching this point is probably an edge case).

Managing a set's groups should not be the wild west. Initially, this should be handled by the 4 roles mentioned above, and probably only done at first for game collections. If we're to have groups beyond grouping stuff within game collections, we should come up with **a standardized set of group titles.**


app/Platform/Data/AchievementSetGroupData.php
```diff
@@ -0,0 +1,33 @@
...
+    ) {
+    }
+
+    public static function fromAchievementGroup(AchievementGroup $group): self
+    {
+        return new self(
+            id: $group->id,
+            label: $group->label,
+            orderColumn: $group->order_column,
+            achievementCount: $group->achievements_count ?? 0,
```
Jamiras, line 29: This isn't filtering by flags. If I associate an achievement to a group and demote it, it still gets counted on the core page. And all achievements in the group get counted on the Unofficial page:

Image

Jamiras: And if you go to the Unofficial page without demoting anything, the group appears, but is empty.

Image

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
Jamiras, line 1: Can "Assign to Group" be added to the row menu? Or is single assigning not expected to be frequent enough to warrant it?

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -448,6 +569,64 @@ public function table(Table $table): Table
...
+                                $record->id,
+                                ['achievement_group_id' => $groupId]
+                            );
+                        });
+
+                        Notification::make()
+                            ->title('Success')
+                            ->body('Successfully assigned achievements to group.')
+                            ->success()
+                            ->send();
```
Jamiras, line 754: Consider de-selecting the items after applying the change (should be done for assign type too). Since I have to scroll back to the top to press the "Assign to Group" button, I may forget that items offscreen are selected.

Alternately (or possibly additionally), is there some way to lock the action bar to the top of the screen when scrolling?

app/Platform/Actions/LoadGameWithRelationsAction.php
```diff
@@ -65,13 +68,89 @@ public function execute(Game $game, AchievementFlag $flag, ?GameAchievementSet $
...
+        // Priority 3: Highest value achievement (by points).
+        $maxPoints = $sortedAchievements->max(fn (Achievement $a) => $a->points);
+        $highestPointAchievements = $sortedAchievements->filter(fn (Achievement $a) => $a->points === $maxPoints);
+
+        if ($highestPointAchievements->count() === 1) {
+            return $highestPointAchievements->first();
+        }
+
+        // Priority 4: If they're all same points (or multiple with the highest), return the first achievement in the list.
+        // This is very likely an edge case.
```
Jamiras: I would suggest using the last achievement in the group as the fallback - assuming the group is sorted in the order the player would earn them, the last achievement would be the "completion condition" for the group.

Although that might be adequately covered by higest value achievement?

And this may not be as uncommon as you'd like. If we allow things to be group into "base game", "post game", and "DLC", neither "post game" nor "DLC" would have any progression markings.

app/Models/AchievementSetAchievement.php
Jamiras, line 1: Changing the achievement group doesn't create a log message.

app/Platform/Data/AchievementSetGroupData.php
wescopeland: Fixed in latest by adding a filter to the back-end query.

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
wescopeland: I think it's a good idea to add. Done in latest.

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
wescopeland: > Consider de-selecting the items after applying the change (should be done for assign type too). Since I have to scroll back to the top to press the "Assign to Group" button, I may forget that items offscreen are selected.

Done in latest.

> Alternately (or possibly additionally), is there some way to lock the action bar to the top of the screen when scrolling?

This is very complex without CSS sorcery, as it's difficult to isolate the sticky header styles to one or more tables without affecting all of them. I think it's a good idea, but probably complex enough to be its own PR.

app/Platform/Actions/LoadGameWithRelationsAction.php
wescopeland: Updated in latest so priority 4 is now the last achievement in the group.

app/Models/AchievementSetAchievement.php
wescopeland: Added a `LogAchievementGroupActivityAction` class in latest which now logs group-related stuff similar to our game releases logging pattern.

app/Platform/Actions/LoadGameWithRelationsAction.php
```diff
@@ -65,13 +68,90 @@ public function execute(Game $game, AchievementFlag $flag, ?GameAchievementSet $
...
     }
+
+    /**
+     * Compute and attach representative badge URLs for each achievement group.
+     *
+     * Priority rules:
+     * 1. The last win condition achievement in the list.
+     * 2. (if no win conditions) The last progression achievement in the list.
+     * 3. (if no progression achievements) The highest value achievement (by points) in the list.
+     * 4. (if all the same points) The first achievement in the list.
```
Jamiras: This comment needs to be updated to reflect the changes.

PR #4261, opened 2025-12-14
feat: add feeds alert system

wescopeland:
Follow-up from #4217.

https://github.com/RetroAchievements/RAWeb/pull/4217#discussion_r2612183412

This PR creates a reusable alerts system that we can use to build feed alerts for all sorts of things. The first alert is for suspicious beat times.

Creating new alerts with this system is pretty straightforward:
1. Create a new alert class like _app/Support/Alerts/SomeNewAlert.php_. The class will have `extends Alert` - so they all have a predictable convention-based interface.
2. Update _config/services.php_ to add a new webhook URL within the new "alerts_webook" list. The key is auto-derived from the class name, ie: `SomeNewAlert` -> `some_new`.
3. Create a trigger listener like _app/Support/Alerts/Listeners/TriggerSomeNewAlert.php_. This contains the detection logic (the "chunk of code" mentioned in your previous message).
5. Register the trigger listener in `EventServiceProvider`.
6. Add the webhook env var to the environment.

### DIFF
--- a/tests/Feature/Support/Alerts/Jobs/SendAlertWebhookJobTest.php
+++ b/tests/Feature/Support/Alerts/Jobs/SendAlertWebhookJobTest.php
@@ -47,7 +47,7 @@ class SendAlertWebhookJobTest extends TestCase
 
         $system = System::factory()->create();
         $user = User::factory()->create(['display_name' => 'Scott']);
-        $game = Game::factory()->create(['Title' => 'Sonic', 'ConsoleID' => $system->id]);
+        $game = Game::factory()->create(['title' => 'Sonic', 'system_id' => $system->id]);
 
         $alert = new SuspiciousBeatTimeAlert(
             user: $user,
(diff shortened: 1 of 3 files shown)


app/Support/Alerts/Listeners/TriggerSuspiciousBeatTimeAlert.php
```diff
@@ -0,0 +1,60 @@
...
+use App\Support\Alerts\SuspiciousBeatTimeAlert;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+/**
+ * Flag beaten games when players beat the game in
+ * less than 5% of the median time.
+ */
+class TriggerSuspiciousBeatTimeAlert implements ShouldQueue
+{
+    public string $queue = 'alerts';
```
Jamiras, line 18: What is this used for?

wescopeland: This is similar to `$queue` in `ResumePlayerSession`:
```php
class ResumePlayerSession implements ShouldQueue
{
    public string $queue = 'player-sessions';
```

It forces these things to stay off the `default` queue so they don't siphon off processes used for other things.

PR #4268, opened 2025-12-16
feat(ticket): don't allow transferring to deleted reporter

Jamiras:


resources/views/pages/ticket/[Ticket].blade.php
```diff
@@ -319,7 +319,7 @@ function AwardManually(hardcore) {
                                     }
                                 }
                             @endphp
-                            @if ($lastComment != null && ($lastComment['User'] === $user->User || $lastComment['User'] === $ticket->achievement->developer->display_name))
+                            @if ($lastComment != null && ($lastComment['User'] === $user->User || $lastComment['User'] === $ticket->achievement->developer->display_name) && !$ticket->reporter->trashed())
```
Comment (wescopeland) [inline resources/views/pages/ticket/[Ticket].blade.php:322]: Technically, this is an authorization check that only lives in the front-end/view layer. In theory, someone could still manually call `/request/ticket/update.php` and the back-end would accept the request with this disallowed value for `action`.

Long-term, it may be better for this conditional logic to be lifted up into `TicketPolicy`. That would make the logic a bit more resilient for when the view is eventually migrated out of Blade.

Comment (Jamiras) [reply resources/views/pages/ticket/[Ticket].blade.php:322]: I've added redundant logic to `request/ticket/update.php`. It's already doing permissions validation, so I didn't feel comfortable (or confident) in making the switch over to policy-based validation.

Comment (wescopeland) [reply resources/views/pages/ticket/[Ticket].blade.php:322]: No worries, this is a good happy medium.

wescopeland (approved):

PR #4270, opened 2025-12-16
Add TotalPlaytime to API_GetGameInfoAndUserProgress

m1guelpf:
Requested in https://github.com/RetroAchievements/RAWeb/discussions/2081#discussioncomment-15273352


public/API/API_GetGameInfoAndUserProgress.php
```diff
@@ -19,6 +19,7 @@
  *  int        NumAwardedToUserHardcore   number of achievements earned by the user in hardcore
  *  string     UserCompletion             percentage of achievements earned by the user
  *  string     UserCompletionHardcore     percentage of achievements earned by the user in hardcore
+ *  int        TotalPlaytime              total time the user has spent playing the game, in seconds
```
Jamiras: Recommend using `UserTotalPlaytime` for consistency with `UserCompletion` and `UserCompletionHardcore` to better stress that this metric is user specific.

Although, `HighestAwardKind` and `HighestAwardDate` don't do that. 

@wescopeland, opinions?

wescopeland: `UserTotalPlaytime` feels contextually appropriate for this endpoint.

m1guelpf: renamed

PR #4271, opened 2025-12-17
feat(api): add V2 JSON:API endpoints for the `User` resource

wescopeland:
This PR adds V2 API endpoints for the `User` resource. There are a few complexities with this resource compared to `Game` and `System` that should be noted:

* ULID is the JSON:API `id` field for users. Unlike other resources that use stringified numeric IDs, users return their ULID as the id.
* Similar to the V1 API, users have a flexible identifier lookup. `username`, `display_name`, and `ulid` are all supported.
* Banned users and unverified users (no `email_verified_at`) are excluded from both endpoints.
* Index results are sorted by points descending when no sort is specified.

**How to test this PR**
```bash
# Get a paginated list of users (defaults to 50 per page, sorted by points)
curl --request GET \
  --url http://localhost:64000/api/v2/users \

# Get a specific user by ULID
curl --request GET \
  --url http://localhost:64000/api/v2/users/0178KDFZ3GDQSVWG8VX5XGQAJ7 \

# Get a user by display name
curl --request GET \
  --url http://localhost:64000/api/v2/users/Scott \

# Filter users by displayable role
curl --request GET \
  --url 'http://localhost:64000/api/v2/users?filter[role]=developer' \

# Sort users by weighted points (descending)
curl --request GET \
  --url 'http://localhost:64000/api/v2/users?sort=-pointsWeighted' \

# Apply custom pagination (page 2, 25 per page)
curl --request GET \
  --url 'http://localhost:64000/api/v2/users?page[number]=2&page[size]=25' \
```


app/Models/User.php
```diff
@@ -538,6 +538,6 @@ public function scopeHasAnyPoints(Builder $query): Builder
      */
     public function scopeVerified(Builder $query): Builder
     {
-        return $query->where('Permissions', '>', 0);
+        return $query->whereNotNull('email_verified_at');
```
wescopeland, line 541: I've resolved some data integrity issues in the prod DB to ensure this is now accurate:

```sql
SELECT
	COUNT(*)
FROM
	UserAccounts
WHERE
	Permissions > 0
	AND email_verified_at IS NULL
	AND Deleted IS NULL;
-- 0 rows returned
```

app/Api/V2/Users/UserResource.php
```diff
@@ -0,0 +1,95 @@
...
+            'pointsWeighted' => $this->resource->TrueRAPoints,
+
+            'yieldUnlocks' => $this->resource->ContribCount,
+            'yieldPoints' => $this->resource->ContribYield,
+
+            'joinedAt' => $this->resource->Created,
+            'lastActivityAt' => $this->resource->LastLogin,
+
+            'isUnranked' => $this->resource->unranked_at !== null,
+            'isUserWallActive' => (bool) $this->resource->UserWallActive,
```
Jamiras, line 49: Is `isUserWallActive` something clients really need/want to know?

app/Api/V2/Users/UserResource.php
```diff
@@ -0,0 +1,95 @@
...
+            'pointsSoftcore' => $this->resource->RASoftcorePoints,
+            'pointsWeighted' => $this->resource->TrueRAPoints,
+
+            'yieldUnlocks' => $this->resource->ContribCount,
+            'yieldPoints' => $this->resource->ContribYield,
+
+            'joinedAt' => $this->resource->Created,
+            'lastActivityAt' => $this->resource->LastLogin,
+
+            'isUnranked' => $this->resource->unranked_at !== null,
```
Jamiras, line 48: I imagine a client would be more interested in the user's current rank (or null if unranked), but that's not currently denormalized, and we'd potentially have to calculate both rank and softcore rank.

app/Api/V2/Users/UserResource.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * @param Request|null $request
+     */
+    public function attributes($request): iterable
+    {
+        return [
+            'displayName' => $this->resource->display_name,
+            'username' => $this->resource->User,
```
Jamiras: Do we really want to expose this? Isn't it supposed to basically not exist if a user changes their display name?

app/Api/V2/Users/UserResource.php
```diff
@@ -0,0 +1,95 @@
...
+
+            'avatarUrl' => $this->resource->avatarUrl,
+            'motto' => $this->resource->Motto,
+
+            'points' => $this->resource->RAPoints,
+            'pointsSoftcore' => $this->resource->RASoftcorePoints,
```
Jamiras: Consider exposing these as `pointsHardcore` and `points`.  Everything exposed by the `game` endpoint is `x` and `xHardcore`.

app/Api/V2/Users/UserSchema.php
```diff
@@ -0,0 +1,140 @@
...
+            ID::make('ulid')->matchAs('.+'),
+
+            Str::make('displayName', 'display_name'),
+            Str::make('username', 'User'),
+
+            Str::make('avatarUrl')->readOnly(),
+            Str::make('motto', 'Motto'),
+
+            Number::make('points', 'RAPoints')->sortable(),
+            Number::make('pointsSoftcore', 'RASoftcorePoints'),
```
Jamiras: Why wouldn't we allow sorting by softcore points?

app/Api/V2/Users/UserSchema.php
```diff
@@ -0,0 +1,140 @@
...
+            Number::make('yieldPoints', 'ContribYield'),
+
+            DateTime::make('joinedAt', 'Created')->sortable(),
+            DateTime::make('lastActivityAt', 'LastLogin')->sortable(),
+
+            Boolean::make('isUnranked')->readOnly(),
+            Boolean::make('isUserWallActive', 'UserWallActive'),
+
+            Str::make('richPresenceMessage', 'RichPresenceMsg'),
+            DateTime::make('richPresenceUpdatedAt', 'RichPresenceMsgDate'),
```
Jamiras: I imagine these should be readonly too? What about `isUserWallActive`? Or `joinedAt`,`lastActivityAt`, or any of the points fields?

app/Api/V2/Users/UserSchema.php
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Get the resource filters.
+     */
+    public function filters(): array
+    {
+        return [
+            Where::make('displayName', 'display_name'),
+            Where::make('username', 'User'),
```
Jamiras: What are the points of these filters? They seem to be exact matches, in which case, you'd just call the endpoint with the username/displayname directly.

app/Api/V2/Users/UserResource.php
wescopeland: I think we should probably leave it - it's an indicator of whether or not comments may be accessible, which is a relationship we'll need to expose in order to reach parity with the V1 API.

app/Api/V2/Users/UserResource.php
wescopeland: If we're going to go down this route, I feel strongly we need to update some TODOs and attributes in `User`:
```php
// TODO rename RAPoints column to points
// TODO rename TrueRAPoints column to points_weighted
// TODO rename RASoftcorePoints column to points_softcore
```
```php
public function getPointsAttribute(): int
{
    return (int) $this->getAttribute('RAPoints');
}

public function getPointsSoftcoreAttribute(): int
{
    return (int) $this->getAttribute('RASoftcorePoints');
}

public function getPointsWeightedAttribute(): int
{
    return (int) $this->getAttribute('TrueRAPoints');
}
```
Changing these attribute definitions likely has far-reaching impacts outside of the currently-touched files in the branch.

I think, as much as it makes sense, the attribute names we expose should mirror the camelCased model field names we're moving towards. There have been one or two places already I've taken a creative liberty (ie: `isUserWallActive` which has an ambiguous migration path in its TODO), but I think we should try to mirror the model fields as closely as possible.

app/Api/V2/Users/UserResource.php
wescopeland: I agree - this seems like a long-term goal but not something we'll be able to squeeze in for a first pass of these endpoints.

app/Api/V2/Users/UserResource.php
wescopeland: Removed this field from the response in latest.

app/Api/V2/Users/UserSchema.php
wescopeland: Good question, we should also allow sorting by the yield stuff too. Added in latest.

app/Api/V2/Users/UserSchema.php
wescopeland: Set everything except the ID (which is implied) to readonly in latest.

app/Api/V2/Users/UserSchema.php
wescopeland: There's only a marginal benefit in that using a filter gives you an empty array instead of a 404 when the user doesn't exist, but that's a weak justification IMO. Removed in latest.

However, I've also made the displayable role filterable in latest, eg:
```
curl --request GET \
  --url 'http://localhost:64000/api/v2/users?filter[role]=developer' \
```

app/Api/V2/Users/UserResource.php
Jamiras: I'm more concerned that once we "publish" this, we won't be able to fix it. While it would be nice to have a 1:1 mapping to the underlying models, if we're able to provide a mapping, we should. I'd much rather have a consistent public-facing API than a 1:1 mapping to the underlying models.

And, as you point out, we want to fix the underlying models. To do that, we'd need mappings to not break the API. As such, I think we should prioritize the mapping now and give the public-facing properties the best names we can.

wescopeland: Fair enough - I've updated these field names accordingly and left the `User` model as-is.

Jamiras (approved):

PR #4274, opened 2025-12-18
feat(manage): improve the Achievement resource

wescopeland:
_achievementInfo.blade.php_ is going to be the next major page I migrate to React. The first part of doing that is making sure Filament is in good shape.

This PR improves the UX of `AchievementResource` and rips everything out of the legacy page's devbox except the memory-related stuff.

**Summary of changes:**
* The resource now has panel-based sections similar to `GameResource` (Primary Details, Classification, Media, Maintainer).
* "Flags" has been renamed to "Published Status" and moved to a dropdown in the Classification section.
* "Badge" has been renamed to "Icon" and "Icon (Locked)".
* The "Type" field now displays humanized labels instead of raw enum values, shows "none" as a placeholder when empty, and includes helper text explaining how types affect beaten status.
* The display order field has been removed altogether from the form. It's still visible on the infolist.
* Author now links to the user's profile page.
* `AssocVideo` is now deprecated and only appears if it has an existing value. When show, it can only be cleared or saved as-is, not changed to a different URL. A deprecation notice with a link to the guides RFC is also displayed.
* The legacy devbox in _achievmementInfo.blade.php_ has been simplified to only show the Mem code and trigger decoder. All form fields have been removed and will only be managed in Filament going forward.

**Before**
Screenshot 2025-12-17 at 7 01 56 PM

**After**
Screenshot 2025-12-17 at 7 01 47 PM

### DIFF
--- a/app/Community/Actions/SendDailyDigestAction.php
+++ b/app/Community/Actions/SendDailyDigestAction.php
@@ -122,8 +122,8 @@ class SendDailyDigestAction
                 $post = $posts[$type][$delayedSubscription->subject_id] ?? null;
                 if ($post) {
                     $summary = match ($type) {
-                        SubscriptionSubjectType::ForumTopic->value => nl2br(Shortcode::stripAndClamp($post->latestComment->body, previewLength: 200, preserveWhitespace: true)),
-                        default => nl2br(mb_strlen($post->Payload) > 200 ? mb_substr($post->Payload, 0, 200) . '...' : $post->Payload),
+                        SubscriptionSubjectType::ForumTopic->value => Shortcode::stripAndClamp($post->latestComment->body, previewLength: 200, preserveWhitespace: true),
+                        default => mb_strlen($post->Payload) > 200 ? mb_substr($post->Payload, 0, 200) . '...' : $post->Payload,
                     };
                     $displayName = match ($type) {
                         SubscriptionSubjectType::ForumTopic->value => $post->latestComment->user->display_name,


app/Filament/Resources/AchievementResource.php
```diff
@@ -213,100 +214,87 @@ public static function form(Schema $schema): Schema
...
+                            ->default(0)
+                            ->options(
+                                collect(AchievementPoints::cases())
+                                    ->mapWithKeys(fn ($value) => [$value => $value])
+                            )
+                            ->disabled(!$user->can('updateField', [$schema->model, 'Points'])),
+
+                        Forms\Components\Select::make('type')
+                            ->label('Type')
+                            ->placeholder('none')
```
Jamiras: Is `none` intentionally lowercase?

Image

app/Filament/Resources/AchievementResource.php
```diff
@@ -73,94 +75,53 @@ public static function infolist(Schema $schema): Schema
...
-                                        ->visible(fn (): bool => $user->can('assignMaintainer', [Achievement::class])),
-                                ]),
-                        ]),
+                    Infolists\Components\ImageEntry::make('badge_url')
+                        ->label('Icon')
+                        ->size(config('media.icon.lg.width')),
+
+                    Infolists\Components\ImageEntry::make('badge_locked_url')
+                        ->label('Icon (Locked)')
+                        ->size(config('media.icon.lg.width')),
```
Jamiras: These images are stored at 64x64, but displayed at 96x96, causing them to be a little fuzzy.

wescopeland: Ah, my bad. I always get that mixed up with these. Updated to use `.md.width` in latest.

app/Filament/Resources/AchievementResource.php
wescopeland: It was, but I don't have a good justification for why (it's to match the "none" placeholder text in the infolist). For the select, it probably makes more sense to use "None".

I think the placeholder label in the infolist should remain lowercased.

PR #4275, opened 2025-12-18
ci: switch from PHPUnit to Pest

wescopeland:
Most of this PR's diff is lockfile noise. This PR swaps PHPUnit for [Pest 4](https://pestphp.com/docs/pest-v4-is-here-now-with-browser-testing), which wraps around PHPUnit and supports sharding in CI.

I've enabled sharding for PHP tests in CI in this PR.

This is an ideal prereq for adding test coverage to `GameController`.

GitHub is showing there's "1 pending check" because that check no longer exists. I'll fix that after this branch is merged.

### DIFF
--- a/app/Models/PlayerStatRanking.php
+++ b/app/Models/PlayerStatRanking.php
@@ -4,34 +4,40 @@ declare(strict_types=1);
 
 namespace App\Models;
 
+use App\Platform\Enums\PlayerStatRankingKind;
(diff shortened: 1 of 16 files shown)


Jamiras:
How does this impact local testing? `composer paratest` and `php artisan test` both seem to run exactly like they used to.

wescopeland:
There should be no impact locally. Those commands should continue working as-is because Pest wraps PHPUnit.

PR #4276, opened 2025-12-18
update ubuntu vm installation instructions to not require sudo for sail

Jamiras:


docs/UBUNTU24_DESKTOP_VM_SETUP.md
```diff
@@ -105,6 +110,8 @@ $ git config --global user.name username
 #### Install docker
 ```
 $ sudo groupadd docker
+$ sudo usermod -aG docker $(whoami)
+$ newgrp docker
```
Jamiras, line 114: This is the "secret sauce" that was missing previously. Without it, `sail` can't find `docker` as the current user, so it has to be run as `root`, causing all sorts of permissions errors.

I still occasionally get permissions errors writing to the log file, so I've left the fix for that below.

Most of the other changes are just updating the process for new versions of VirtualBox and PHP.

docs/UBUNTU24_DESKTOP_VM_SETUP.md
```diff
@@ -184,16 +197,11 @@ $ composer update
 $ composer install
 $ composer setup
 $ sail up -d
-$ sail root-shell
- # cd ..
- # chown -R sail:sail html
- # exit
 $ sail artisan ra:storage:link
-$ php composer.phar install
+$ rm public/fi-build
```
Jamiras, line 201: `sail artisan ra:storage:link` creates this folder, but with different permissions than whatever uses it needs. By removing the folder, the other process can create and populate is appropriately.

docs/UBUNTU24_DESKTOP_VM_SETUP.md
```diff
@@ -184,16 +197,11 @@ $ composer update
...
 $ pnpm install
 $ pnpm build
-$ sudo chmod 777 bootstrap
-$ sudo chmod 777 bootstrap/cache
+$ sail artisan passport:keys --force
```
Jamiras, line 204: This is required for the oauth unit tests

wescopeland (approved):

PR #4277, opened 2025-12-18
refactor(profile) do point sums in database

Jamiras:
Greatly reduces the number of models fetched for calculating the player's recent points earned for very active players.

before:
image

after:
image

Unfortunately, it has minimal impact on memory consumption
image
image

Note that these queries are for data that's hidden by default.
image


app/Community/Components/UserProfileMeta.php
```diff
@@ -49,8 +50,6 @@ public function render(): View
             $developerStats = $this->buildDeveloperStats($this->user, $this->userMassData);
         }
 
-        $this->calculateRecentPointsEarned($this->user);
```
Jamiras, line 52: Return value ignored. Wasteful.

PR #4278, opened 2025-12-18
refactor(profile): fix n+1 query fetching event award information

Jamiras:


app/Helpers/render/site-award.php
```diff
@@ -226,7 +227,7 @@ function RenderAward(array $award, int $imageSize, string $ownerUsername, bool $
     }
 
     if ($awardType == AwardType::Event) {
-        $event = Event::find($awardData);
+        $event = $eventData->find($awardData);
```
Jamiras, line 230: This is the repetetive query being eliminated.

PR #4279, opened 2025-12-18
perf(user): reduce memory usage on profiles with many touched games

wescopeland:
https://discord.com/channels/310192285306454017/1451313537304756336

Drops http://localhost:64000/user/Kinglink's memory consumption from 140MB to 18MB (~87%) per hit on my local machine.

We achieve this by making two optimizations:
1. `getUsersCompletedGamesAndMax()` now accepts an optional `$limit` arg. We pass `limit: 200` for the completion progress stuff in the sidebar, so we can avoid loading nearly 10,000 rows for power users. We can't have an unbounded list of things here anymore.
2. A new `GetUserProgressionStatusCountsAction` uses SQL aggregation instead of PHP iteration to build up Progression Status component data. Previously, all player games were loaded into memory and we looped through them many times to calculate per-system counts. Now a single query with `GROUP BY ConsoleID` returns rows with pre-computed values.

"Subset games" are still excluded from unfinished counts and the average completion % stat.

### DIFF
--- a/app/Community/Components/UserProfileMeta.php
+++ b/app/Community/Components/UserProfileMeta.php
@@ -7,6 +7,7 @@ namespace App\Community\Components;
 use App\Community\Enums\Rank;
 use App\Community\Enums\RankType;
 use App\Enums\Permissions;
+use App\Models\Achievement;
 use App\Models\PlayerStat;
 use App\Models\User;
 use App\Platform\Enums\PlayerStatType;
@@ -48,8 +49,6 @@ class UserProfileMeta extends Component
         if ($this->userMassData['ContribCount'] > 0 || $this->user->getAttribute('Permissions') >= Permissions::JuniorDeveloper) {
             $developerStats = $this->buildDeveloperStats($this->user, $this->userMassData);
         }
-
-        $this->calculateRecentPointsEarned($this->user);
 
         return view('components.user.profile-meta', [
             'developerStats' => $developerStats,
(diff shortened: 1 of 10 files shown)


app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -0,0 +1,163 @@
...
+namespace App\Platform\Actions;
+
+use App\Models\PlayerGame;
+use App\Models\System;
+use App\Models\User;
+use App\Platform\Enums\AchievementSetType;
+
+class GetUserProgressionStatusCountsAction
+{
+    private const MIN_ACHIEVEMENTS = 5;
```
Jamiras: I realize this is duplicated from the other function, but it's misleading. It should be:
```
private const MIN_ACHIEVEMENTS_FOR_COMPLETION = 6;
```

And the query should use `>=` instead of `>`.

Jamiras: It could also just use `PlayerBadge::MINIMUM_ACHIEVEMENTS_COUNT_FOR_MASTERY`, which is I think what the logic is trying to capture.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -0,0 +1,163 @@
...
+            ")
+            ->join('GameData', 'GameData.ID', '=', 'player_games.game_id')
+            ->join('Console', 'Console.ID', '=', 'GameData.ConsoleID')
+            ->where('player_games.user_id', $user->id)
+            ->where('player_games.achievements_unlocked', '>', 0)
+            ->whereRaw('Console.active = true')
+            ->whereRaw('GameData.achievements_published > ?', [self::MIN_ACHIEVEMENTS])
+            ->whereNotIn('GameData.ConsoleID', System::getNonGameSystems())
+            ->groupBy('GameData.ConsoleID')
+            ->get();
```
Jamiras, line 89: Something about this query isn't exactly the same.

Before:
Image

After:
Image

There 2 additional unbeaten games, 2 additional beaten softcore games, 11 additional beaten games, 2 less completed games, and 15 less mastered games.

Focusing just on SCD because it has low numbers and showed up in the screenshot, the missing mastery appears to have been revised and is being counted as Beaten now:

Image

That could potentially explain the bulk of the discrepancy. 2 completed->2 beaten softcore, 15 mastered->11 beaten. But where did the other 4 masteries go? I'm guessing two are no longer beaten and got counted as the additional unbeaten games, and the other two were either reset or the set was demoted so it has less than 6 achievements, so they aren't being counted anywhere.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
wescopeland: Updated in latest:
```php
->whereRaw('GameData.achievements_published >= ?', [PlayerBadge::MINIMUM_ACHIEVEMENTS_COUNT_FOR_MASTERY])
```

`whereRaw` is used to keep PHPStan happy due to the joins.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
wescopeland: I think this is now fixed in latest. I had to shift the query over to primarily being driven by awards rather than denormalized data.

resources/views/pages-legacy/userInfo.blade.php
Jamiras, line 1: Started Games Beaten is now 100% for all users.

Image

wescopeland: Should be fixed now.

Jamiras (approved):

PR #4281, opened 2025-12-19
fix(multiset): report base unlocks when resolving a specialty hash

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1451398633005252739/1451398633005252739.

### DIFF
--- a/app/Connect/Actions/StartSessionAction.php
+++ b/app/Connect/Actions/StartSessionAction.php
@@ -80,10 +80,8 @@ class StartSessionAction extends BaseAuthenticatedApiAction
         $resolvedSets = collect();
         if (config('feature.enable_multiset') && $gameHash) {
             $resolvedSets = (new ResolveAchievementSetsAction())->execute($gameHash, $this->user);
-        }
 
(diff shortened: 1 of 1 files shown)


app/Connect/Actions/StartSessionAction.php
```diff
@@ -74,7 +74,15 @@ protected function process(): array
...
-        // if multiset is enabled, redirect the heartbeat to the root game.
+        // Resolve achievement sets for multiset. This determines which sets we should
+        // query for unlocks. If empty (no hash, multiset disabled, or no sets found),
+        // we fall back to the resolved game ID.
+        $resolvedSets = collect();
+        if (config('feature.enable_multiset') && $gameHash) {
+            $resolvedSets = (new ResolveAchievementSetsAction())->execute($gameHash, $this->user);
+        }
+
+        // If multiset is enabled, redirect the heartbeat to the root game.
```
Jamiras: You should be able to roll this logic into the above if statement. If `$gameHash` is null, `ResolveRootGameFromGameAndGameHashAction` is a no-op.

PR #4288, opened 2025-12-21
feat(manage): add 'Logic' tab to achievements resource

wescopeland:
[AutoCR](https://github.com/authorblues/retroachievements/tree/main/AutoCR) is heavily leveraged by this PR as a reference implementation. Many thanks to @authorblues, @christianlegge, @CySlaytor, @joshraphael, and other AutoCR contributors for their battle-tested reference implementation.

---

This PR adds a "Logic" tab to the achievements resource in Filament. The UI is powered by `TriggerDecoderService` and several new services ported from AutoCR. The devbox on _achievementInfo.blade.php_ has been removed.

## UX Walkthrough
http://localhost:64000/manage/achievements/9/logic

Screenshot 2025-12-21 at 12 30 17 PM

There are three parts of the UI:
1. The version history (which we've been recording since Jan 2025), complete with a diff viewer.
2. The toolbar, which contains 4 controls (one of which is hidden if there's no `AddAddress` flags in the logic).
3. The detailed logic viewer.

### Version History
Each row of the version history is expandable to see who added, removed, and/or modified what conditions. Here's a good one to play with: http://localhost:64000/manage/achievements/99504/logic.

When rows are expanded, it's very easy to see who touched what:
Screenshot 2025-12-21 at 12 33 32 PM

Toggling over to "Raw" shows the raw logic for that version:
Screenshot 2025-12-21 at 12 34 26 PM

If there are more than 8 versions, older stuff gets put behind a "See more". You can test this at http://localhost:64000/manage/achievements/545113/logic:
Screenshot 2025-12-21 at 12 35 24 PM

### Toolbar and Detailed Logic Viewer
Most of this functionality is ported directly from AutoCR's source code, migrating the JS to server-side PHP. This felt like a good choice, as AutoCR is already quite battle-tested and has already teased out some interesting edge cases in its logic.

Screenshot 2025-12-21 at 12 37 05 PM

Hovering over an underlined `Mem/Val` immediately renders its corresponding code note in a tooltip:
Screenshot 2025-12-21 at 12 38 22 PM

When "Show aliases" is enabled, code notes are exhaustively parsed to try to determine what an alias for a given value is:
Screenshot 2025-12-21 at 12 40 05 PM

AutoCR also features a "Collapse AddAddress" function that I've ported over. A good demo for the "Collapse AddAddress" toggle can be found at http://localhost:64000/manage/achievements/403355/logic.

ELI5 for someone unfamiliar: in our trigger logic, `AddAddress` is a flag used to chain memory reads together. It adds the value at one address to calculate the final address for a subsequent read. This is common for stuff like pointer chains. These particular conditions can make the trigger condition table quite verbose since they're "setup conditions" rather than the core logic. Enabling "Collapse AddAddress" lets users collapse/hide them to focus on the actual comparison conditions.

**Before**
Screenshot 2025-12-21 at 12 43 20 PM

**After**
Screenshot 2025-12-21 at 12 43 23 PM

When "Copy Markdown" is pressed, content is sent to the clipboard like:
```md
### Core
  1: AddAddress Mem   32-bit BE 0x00005518 &   Value           2147483647 
  2:            Mem   32-bit BE 0x00000034 !=  Value           0          (0)
  3: AddAddress Mem   32-bit BE 0x00005518 &   Value           2147483647 
  4:            Mem   8-bit     0x0000000e !=  Value           15         (0)
  5: AddAddress Mem   32-bit BE 0x00005518 &   Value           2147483647 
  6: AddAddress Mem   32-bit BE 0x00000004 &   Value           2147483647 
  7:            Mem   8-bit     0x00000064 !=  Value           0          (0)
  8: AddAddress Mem   32-bit BE 0x00005518 &   Value           2147483647 
  9: AddAddress Mem   32-bit BE 0x00000034 &   Value           2147483647 
 10:            Mem   Bit2      0x000000b6 >   Delta Bit2      0x000000b6 (0)
```
This functionality is strictly reverse-engineered from AutoCR.

### Future work
* Port the Statistics section
* Port the Logic Analysis section

Logic feedback and writing feedback are both out of scope.


app/Platform/Services/TriggerViewerService.php
```diff
@@ -0,0 +1,653 @@
...
+        if (in_array($sourceSize, $floatTypes, true)) {
+            if ($sourceSize !== 'MBF32' && $sourceSize !== 'MBF32 LE') {
+                $alias = $this->resolveFloatValue($hexValue, $sourceSize, $noteSection);
+                if ($alias !== null) {
+                    return $alias;
+                }
+            }
+        }
+
+        return $this->resolveIntegerValue($decimalValue, $noteSection);
```
Jamiras, line 217: If this is based on AutoCR, the issue might be there as well, but this doesn't handle CSV:
http://localhost:64000/manage/achievements/60343/logic

Image

In the first line, `7` should be `game`, and in the last line, it's evaluating `0` to be the rest of the line.

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
```diff
@@ -0,0 +1,259 @@
...
+                                            @endphp
+
+                                            @if ($condition['SourceType'] === 'Recall')
+                                                <span class="{{ $sourceDisplay['cssClass'] }}">{recall}</span>
+                                            @elseif ($hasSourceNote)
+                                                <span
+                                                    x-show="showAliases"
+                                                    x-cloak
+                                                    class="text-emerald-600 dark:text-emerald-400 cursor-help"
+                                                    x-tooltip="{ content: @js('<span style=\'' . $tooltipStyle . '\'>' . e($condition['SourceTooltip']) . '</span>'), theme: $store.theme, allowHTML: true, placement: 'left' }"
```
Jamiras: Is a tooltip really desirable if the full note is already visible?

Image

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
```diff
@@ -0,0 +1,259 @@
...
+                                                    <span class="text-pink-500 dark:text-pink-400">{recall}</span>
+                                                @elseif ($condition['TargetType'] === 'Value')
+                                                    @if ($targetDisplay['valueAlias'])
+                                                        <span x-show="showAliases" x-cloak class="text-emerald-600 dark:text-emerald-400">{{ $targetDisplay['valueAlias'] }}</span>
+                                                        <span x-show="!showAliases && showDecimal" x-cloak>{{ $targetDisplay['decimalDisplay'] }}</span>
+                                                        <span x-show="!showAliases && !showDecimal" x-cloak>{{ $targetDisplay['hexDisplay'] }}</span>
+                                                    @else
+                                                        <span x-show="showDecimal" x-cloak>{{ $targetDisplay['decimalDisplay'] }}</span>
+                                                        <span x-show="!showDecimal" x-cloak>{{ $targetDisplay['hexDisplay'] }}</span>
+                                                    @endif
```
Jamiras, line 253: Minor: The old logic viewer would show the decimal value as a tooltip for hex constants.

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
Jamiras: I guess most notes include a size that's being removed, so it may not actually buy much to add that logic.

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
```diff
@@ -0,0 +1,259 @@
...
+                        <label class="inline-flex items-center gap-2 cursor-pointer select-none">
+                            <x-filament::input.checkbox x-model="collapseAddAddress" />
+                            <span class="text-sm text-neutral-600 dark:text-neutral-400">Collapse AddAddress</span>
+                        </label>
+                    @endif
+                </div>
+
+                <x-filament::button size="sm" color="gray" icon="heroicon-o-clipboard-document" @click="copyMarkdown()">
+                    Copy Markdown
+                </x-filament::button>
```
Jamiras, line 80: This always copies the original view (decimal, no aliases, no collapsed addresses). Is that intentional?

app/Platform/Services/TriggerDiffService.php
```diff
@@ -0,0 +1,274 @@
...
+     */
+    private function diffConditions(array $oldConditions, array $newConditions): array
+    {
+        // Compare conditions by position since order matters in triggers.
+        $result = [];
+        $maxLen = max(count($oldConditions), count($newConditions));
+
+        for ($i = 0; $i < $maxLen; $i++) {
+            $old = $oldConditions[$i] ?? null;
+            $new = $newConditions[$i] ?? null;
```
Jamiras: I realize this algorithm is very naive, but it shouldn't be too hard to detect when a new condition is added in the middle of stuff. In this case, I added a new condition 3. Conditions 4 and 5 are identical to the old conditions 3 and 4.

Image

RATools goes one step farther and detects any moved conditions (keeping them in the position associated to the current version), but that gets significantly more complicated when you have to consider chains (AddAddress/AddSource/etc).

resources/views/filament/resources/achievement-resource/partials/version-history.blade.php
Jamiras, line 1: I'm not sure where the extra memory is being consumed, but when I look at the logic viewer for the most-changed achievement in my DB (http://localhost:64000/manage/achievements/545237/logic), it's dangerously close to the 128MB limit.

There might be an achievement with fewer diffs but more logic that tips over. This one claims to have used 320MB: http://localhost:64000/manage/achievements/496485/logic

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
```diff
@@ -0,0 +1,259 @@
...
+
+                    <label class="inline-flex items-center gap-2 cursor-pointer select-none">
+                        <x-filament::input.checkbox x-model="showAliases" />
+                        <span class="text-sm text-neutral-600 dark:text-neutral-400">Show aliases</span>
+                    </label>
+
+                    @if ($hasAddAddress)
+                        <label class="inline-flex items-center gap-2 cursor-pointer select-none">
+                            <x-filament::input.checkbox x-model="collapseAddAddress" />
+                            <span class="text-sm text-neutral-600 dark:text-neutral-400">Collapse AddAddress</span>
```
Jamiras, line 73: I can't help but feel it would be useful to have some sort of indicator on which rows are at the end of a hidden AddAddress chain.

For example, rows 11 and 13 were removed here:

Image

But without actually staring at the barely visible row numbers, that's not obvious.

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
wescopeland: It's intentional. AutoCR does the same - I assume so whatever is copied is easily reproducible.

authorblues: The Copy Markdown button predates a lot of the other features. It exists to be able to provide a reference in code review, and is meant to represent what Jrs might expect to see in the GUI. Decimal vs hex might be the only thing I'd consider maintaining in the copy on our end.

app/Platform/Services/TriggerViewerService.php
wescopeland: Yep, [it has the same issue](https://authorblues.github.io/retroachievements/AutoCR/#!/game/10392/achievement/60343).

Screenshot 2026-01-02 at 10 30 52 AM

I've added some basic CSV parsing in latest for decimal values:
Screenshot 2026-01-02 at 10 37 33 AM

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
wescopeland: It's kinda complicated to do this in Blade and keep Alpine.js happy, but here's what latest looks like:
Screenshot 2026-01-02 at 10 46 09 AM

Things that have a dotted underline are given a tooltip. The dotted underline is also consistent with AutoCR:
Screenshot 2026-01-02 at 10 46 44 AM

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
wescopeland: Non-trivial (>= 10) hex constants in the target Mem/Val column will now have decimal tooltips:
Screenshot 2026-01-02 at 10 53 16 AM

app/Platform/Services/TriggerDiffService.php
wescopeland: Here's what it looks like in latest:
Screenshot 2026-01-02 at 11 22 03 AM

resources/views/filament/resources/achievement-resource/partials/version-history.blade.php
wescopeland: The memory is coming from tens of thousands of DOM nodes being rendered for those achievements. I agree, that really isn't going to work.

My gut feeling is this is probably the exception, not the rule.

It took some doing, but in latest, there's now logic for conditional lazy loading for achievements with triggers that meet a certain size threshold in kilobytes. This results in 496485 using ~64MB of server RAM, albeit the page still takes a long time to load and the tab uses something like 2GB of memory on my machine. Given the achievement has thousands of conditions, there probably isn't a good way around that.

resources/views/filament/resources/achievement-resource/pages/logic.blade.php
wescopeland: Agreed. I've made two tweaks in latest to help with this:
1. There's now a purple left border on the rows.
2. The row numbers are colorized and include an informational tooltip.

Screenshot 2026-01-02 at 12 25 17 PM

Screenshot 2026-01-02 at 12 25 21 PM

PR #4289, opened 2025-12-21
feat(multiset): add set to hash resolver action, show multiset hashes on ticket create

wescopeland:
This PR:
* Adds `ResolveAchievementSetGameHashesAction`, which when given an achievement set, returns all possible hashes (`GameHash` entities) that can be used to play that set.
* Allows users to select multiset-driven hashes when opening a ticket.

Last week, a player attempted to open a ticket for Zelda 1's bonus set while using a core set hash. In the ticket create form, they couldn't select the actual hash they used. Now, they'll be able to.

Rather than pull in every possible base set hash, the controller checks for which hashes returned by `ResolveAchievementSetGameHashesAction` the player had any sessions with. These hashes are then inserted into the list.

The new action will be reusable in other contexts later, such as the Supported Game Files page.

Screenshot 2025-12-21 at 3 54 26 PM


tests/Feature/Platform/Controllers/TriggerTicketControllerTest.php
Jamiras, line 1: Related: The ability to create a ticket against an achievement is tied to having a `player_games` record for the game associated to the achievement. If a player hasn't earned any achievements from the subset yet, they won't have a `player_games` record for the subset, so they can't create a ticket for the achievement.

app/Platform/Actions/BuildTicketCreationDataAction.php
Jamiras, line 1: The base game ID needs to be transferred to the query on line 125 as well. That will allow the page to auto-select the hash the player was using.

tests/Feature/Platform/Controllers/TriggerTicketControllerTest.php
wescopeland: That's a good point. I've added `Achievement::getRelatedGameIds()` which returns game IDs respecting multiset boundaries (we'll need to rethink this approach later so it's by achievement set IDs, but I'm not sure we're there yet).

`TriggerTicketPolicy::createAchievementTicket()` now uses `$user->hasPlayedGameForAchievement($achievement)` which checks if the user has a `player_games` record for any related "game".

I've also added controller tests to verify the behavior.

app/Platform/Actions/BuildTicketCreationDataAction.php
wescopeland: That makes sense, and I think there are 3 other places in the action that were similarly affected by this feedback.

The action now uses `$achievement->getRelatedGameIds()` for all the session queries. I've also added a controller test to verify that when a player has only a base game session, creating a ticket for a bonus achievement correctly auto-selects the base game hash and populates emulator info.

PR #4290, opened 2025-12-22
feat(memnotes): add link to subset notes from base set; add merge functionality to subset notes page

Jamiras:
https://discord.com/channels/310192285306454017/310195377993416714/1450477207301652571

Adds a link to the memory notes for subsets of a game from the game's memory notes page.
image

When viewing memory notes for a subset, if the user has Developer permissions, they'll be given the option to select whether to keep the base set memory note or the subset memory note.

image

Once all subset notes have been merged, the subset notes page link will disappear from the base set memory notes page, and trying to hit the subset notes page will redirect back to the base set memory notes page.

**Automatic deduplication command**:
```
sail artisan ra:platform:prune-duplicate-subset-notes
```

### DIFF
--- a/.hooks/pre-push
+++ b/.hooks/pre-push
@@ -49,12 +49,12 @@ if [ "$is_php_changed" = true ]; then
     # Check the OS. Windows does not support the --parallel flag.
     if [ "$(uname)" = "Linux" ] || [ "$(uname)" = "Darwin" ]; then
         # Use --parallel if macOS or Linux are detected.
-        printf "\n⏳ composer test -- --parallel\n"
-        vendor/bin/paratest
+        printf "\n⏳ vendor/bin/pest --parallel\n"
+        vendor/bin/pest --parallel
     else
         # If neither of those are detected, don't use --parallel.
-        printf "\n⏳ composer test\n"
-        vendor/bin/phpunit
+        printf "\n⏳ vendor/bin/pest\n"
+        vendor/bin/pest
     fi
 fi
 


public/request/game/merge-code-note.php
wescopeland, line 1: There's a very high likelihood this will need to eventually be refactored to a controller. I recommend either putting this in a `MemoryNotesController::merge()` or creating some kind of reusable `MergeMemoryNote` action class that can be ported to a controller later.

app/Helpers/render/code-note.php
wescopeland, line 1: UX: I recommend sticking all the new buttons into the "Dev" column. This keeps their placement consistent rather than dynamically shifting based on the note content.

Image

app/Helpers/render/code-note.php
```diff
@@ -45,7 +54,25 @@ function RenderCodeNotes(array $codeNotes, ?User $editingUser = null, ?int $edit
...
-        echo "<td data-address='$addr' style='width: 25%;'>";
+        if (array_key_exists('SubsetUser', $nextCodeNote)) {
+            $subsetMemNote = $nextCodeNote['SubsetNote'] ?? '';
+            sanitize_outputs($subsetMemNote);
+            $subsetMemNote = nl2br($subsetMemNote);
+
+            $keepButtonExtra = '';
+            $keepText = 'Keep Base Set Note';
+            if (empty(trim($nextCodeNote['Note'])) || $nextCodeNote['Note'] == "''") {
+                $keepText = 'Discard Subset Note';
```
wescopeland, line 65: UX, minor: I was confused by this button label.

If the buttons are all put in the dev column as suggested in another feedback comment, we may be able to make this less ambiguous. Given there are two actions available when this button appears, maybe they are:
* "Import to Base"
* "Delete"

app/Helpers/render/code-note.php
wescopeland, line 1: UX: We're effectively building a merge tool. Right now, it's difficult to scan the list of notes and see at a glance which ones need to be merged. I came up with this concept:

Image

How do you feel about something like this? The classes are `border-l-4 border-amber-500 bg-amber-500/10` and are applied to the `tr` element.

wescopeland, line 1: UX, minor: It may be useful to show the count of notes remaining to be merged.

Image

wescopeland, line 1: UX: Technically, these are all destructive actions. Should we add any more `confirm()`s?

wescopeland (commented):
Verified functionality. I have a few UX recommendations. Those designated with "minor" I consider nice-to-haves but non-blocking.

app/Helpers/render/code-note.php
Jamiras: I had this originally, and didn't like the disassociation.

1) It's easy to click on the wrong one (which might be remedied with a confirmation dialog)
2) It's much easier to know which button is associated to which note if they're right next to the note.

The whole idea behind this UI is for the user to select one of the two notes to keep. I feel that having the buttons next to the text makes it easier for the user to say "I choose _this_ one".

As for the "dynamically shifting content", you mean the fact that the buttons and text are center aligned? I'm not terribly happy about that either, but I don't know that forcing the button to be bottom aligned is any better.

Jamiras: Between the buttons and the fact that there's extra columns with text/icons in them, I think it would be sufficiently obvious which items need merging.

If we were going to go with colors, I'd probably want the left text to have a red background and the right text to have a green background to mimic github (or maybe just color the text itself). I'm not sure how that (or even this) would work with all our selectable color schemes.

Jamiras: Agreed. Done.

Jamiras: I thought about that, but I felt that popping up a confirm for every merge choice would just get annoying.

app/Helpers/render/code-note.php
Jamiras: If there is no base set note, saying "keep base set note" feels odd, so I change it to "discard subset note". 

The button is effectively always "discard subset note", but if a base set note exists, it's better to have a positive action: "keep base set note".

public/request/game/merge-code-note.php
Jamiras: I was assuming this functionality wouldn't be needed long-term.

app/Helpers/render/code-note.php
wescopeland: > As for the "dynamically shifting content", you mean the fact that the buttons and text are center aligned?

I mean the placement of the button is based on how long the note in its column is:
Screenshot 2025-12-23 at 3 41 30 PM

> It's much easier to know which button is associated to which note if they're right next to the note.

I agree with this and didn't think of it this way. Given that, it seems fine.

app/Helpers/render/code-note.php
wescopeland: That reasoning makes sense. I can't help it - I still find this confusing:
Screenshot 2025-12-23 at 3 43 54 PM

The button labels seem ambiguous to me.

"Discard Subset Note" sounds like it might delete the note from the subset entirely, when it really means "skip merging this note".

"Keep Subset Note" sounds like "leave this where it is" when it actually means "copy this to the base set".

Here are some possibly clearer alternatives:
* "Discard Subset Note" -> "Skip", "Don't Merge", "Leave Base Empty"
* "Keep Subset Note" -> "Copy to Base Set", "Merge Note", "Use Subset Note"

public/request/game/merge-code-note.php
wescopeland: My perception is the duration we need this tool may be difficult to predict (I would be happy to be wrong about this). With hundreds of subsets and folks probably not incentivized to do these merges until something goes wrong with their achievement logic, we may have these "merge conflicts" lingering for quite some time.

app/Helpers/render/code-note.php
Jamiras: > "Discard Subset Note" sounds like it might delete the note from the subset entirely, when it really means "skip merging this note".

No, it does delete the note from the subset entirely. The "resolution" is to keep the non-existant note for the base set and discard the data in the subset.

Similarly, "keep base set note" keeps the existing note for the base set and discards the data in the subset.

> "Keep Subset Note" sounds like "leave this where it is" when it actually means "copy this to the base set".

"Keep subset note" promotes the subset note data to the base set and then discards the data in the subset.

Once all the data has been discarded from the subset, the page will redirect to the notes page for the base set and the link to the subset notes will disappear.

At its most-simplest, the buttons are "keep left" and "keep right". But they both result in the data being kept in the base set and removed from the subset.

public/request/game/merge-code-note.php
Jamiras: There are 253 subsets currently with ~32000 notes.

```
select game_id, count(*) from memory_notes where game_id in (select game_id from game_achievement_sets where achievement_set_id in (select achievement_set_id from game_achievement_sets where type='bonus' or type='will_be_bonus') and type='core') group by 1 order by 2 desc;
```

```
select count(*) from memory_notes where game_id in (select game_id from game_achievement_sets where achievement_set_id in (select achievement_set_id from game_achievement_sets where type='bonus' or type='will_be_bonus') and type='core');
```

Suikoden has the most subset notes, and 4527/4754 are identical. We could (should?) automate eliminating the duplicates.

wescopeland (approved):

public/request/game/merge-code-note.php
Jamiras: added command.
```
$ sail artisan ra:platform:prune-duplicate-subset-notes
Pruning subset notes copied from base sets...
 31397/31397 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
Pruned 21903 notes.
```

Only one subset was entirely deduplicated. After the command ran, there's 252 subsets with ~10000 notes.

PR #4291, opened 2025-12-22
feat(mail): allow unsubscribing from daily digest emails

wescopeland:
https://discord.com/channels/476211979464343552/1026595325038833725/1452747878668832870

Adds a new inverted pref to support this: `EmailOff_DailyDigest`.

If the bit is set, the user won't receive daily digest emails anymore. These emails now also support RFC 8058 unsubscribe headers.

Screenshot 2025-12-22 at 5 33 49 PM


app/Mail/Services/UnsubscribeService.php
```diff
@@ -142,7 +142,15 @@ public function processUnsubscribe(string $token, bool $shouldGenerateUndoToken
...
             $currentPrefs = $user->websitePrefs;
-            $newPrefs = $currentPrefs & ~(1 << $data->preference);
+
+            // For "*Off_*" preferences, unsubscribing means setting the bit.
+            // For "*On_*" preferences, unsubscribing means clearing the bit.
+            if ($data->preference === UserPreference::EmailOff_DailyDigest) {
+                $newPrefs = $currentPrefs | (1 << $data->preference);
+            } else {
+                $newPrefs = $currentPrefs & ~(1 << $data->preference);
+            }
```
Jamiras, line 156: This might make more sense as a switch statement:
```
switch ($data->preference) {
    case UserPreference::EmailOff_DailyDigest:
        // For "*Off_*" preferences, unsubscribing means setting the bit.
        $newPrefs = $currentPrefs | (1 << $data->preference);
        break;
    default:
        // For "*On_*" preferences, unsubscribing means clearing the bit.
        $newPrefs = $currentPrefs & ~(1 << $data->preference);
        break;
}
```

app/Mail/Services/UnsubscribeService.php
```diff
@@ -308,6 +324,9 @@ private function getCategoryDescription(int $preference): array
             case UserPreference::EmailOn_TicketActivity:
                 return ['key' => 'unsubscribeSuccess-allTicketActivity'];
 
+            case UserPreference::EmailOff_DailyDigest:
+                return ['key' => 'unsubscribeSuccess-dailyDigest'];
```
Jamiras, line 336: Is there a missing translation key entry for this somewhere?

Image

app/Mail/Services/UnsubscribeService.php
wescopeland: Good idea - I wasn't sure what a good pattern might be for this at the time so went with a naive `if/else`. Updated to a `switch` in latest (two spots needed to be changed).

app/Mail/Services/UnsubscribeService.php
wescopeland: Yep - I accidentally discarded the change before pushing. Updated in latest.

Jamiras (approved):

PR #4292, opened 2025-12-23
fix(mail): automatically update first_update_id when comments are deleted

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1452819012617637999.

**Root Cause**
When a comment gets deleted, `first_update_id` isn't being updated accordingly.

This PR creates a `CommentObserver`, which on delete will find any pending delayed subscription entities and either update them to the next comment down or delete them altogether.


app/Observers/CommentObserver.php
Jamiras, line 1: I haven't tested these changes, but I was wondering if it would make more sense to just do the check when building the email?

Update to `CommentDelayedSubscriptionHandler::getUpdatesSince` (also untested):
```
    public function getUpdatesSince(UserDelayedSubscription &$delayedSubscription): int
    {
        if (!Comment::where('ID', $delayedSubscription->first_update_id)->exists()) {
            $nextComment = Comment::where('ArticleType', $this->articleType)
                ->where('ArticleID', $delayedSubscription->subject_id)
                ->where('ID' > $delayedSubscription->first_update_id)
                ->where('user_id', '!=', $delayedSubscription->user_id)
                ->orderBy('ID')
                ->first();

            if (!$nextComment) {
                return 0;
            }

            $delayedSubscription->first_update_id = $nextComment->id;
        }

        return Comment::query()
            ->where('ArticleType', $this->articleType)
            ->where('ArticleID', $delayedSubscription->subject_id)
            ->where('ID', '>=', $delayedSubscription->first_update_id)
            ->where('user_id', '!=', $delayedSubscription->user_id)
            ->count();
    }
```

The parameter is changed to pass by reference, so the `first_update_id` can be updated in place.

wescopeland: Yeah, I like that idea more.

Updated in latest to do this lazy send-time checking. Also added a few test cases.

Jamiras (approved):

PR #4296, opened 2025-12-23
feat(multiset): add 'Enable multiset' toggle for devs to enable multiset for games

wescopeland:
```sql
UPDATE game_achievement_sets
SET type = 'exclusive'
WHERE type = 'will_be_exclusive';
```

---

Should only be merged when it's time to end the multiset beta.

This PR adds an "Enable multiset" toggle button in the game "Sets" tab for users with the Developer role to turn on multiset for a game.

When the button is pressed and the action is confirmed, `will_be_bonus` and `will_be_specialty` are flipped to `bonus` and `specialty` respectively.

`will_be_exclusive` is unaffected - this type should be removed altogether for redundancy; devs should be able to directly set a type as `exclusive` instead.

Pressing the toggle logs an entry on the game's audit log.

Screenshot 2025-12-23 at 2 13 00 PM

Screenshot 2025-12-23 at 2 18 58 PM

Screenshot 2025-12-23 at 2 13 09 PM


app/Platform/Enums/AchievementSetType.php
```diff
@@ -32,20 +32,14 @@ enum AchievementSetType: string
...
-     * 3- When it's time to make multiset generally-available to all players, all will_be_*
-     *    values can transition to the real thing and these enum values can be removed.
+     * These values are set by developers using AchievementSetsRelationManager.
+     * Any will_be_* value will logically function the same as "core" until
+     * multiset is enabled for the game.
      */
     case WillBeBonus = "will_be_bonus";
     case WillBeSpecialty = "will_be_specialty";
+
+    // TODO this is redundant. remove it and only have "exclusive", maybe with a note explaining why.
```
Jamiras: What is the path forward for this? Should we just run a query on the server and clean this up as part of this PR?

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
```diff
@@ -133,6 +131,24 @@ public function table(Table $table): Table
...
+                Actions\Action::make('toggleMultiset')
+                    ->visible(fn () => $user->can('toggleMultiset', GameAchievementSet::class)
+                        && $game->gameAchievementSets()->whereIn('type', [
+                            AchievementSetType::Bonus->value,
+                            AchievementSetType::WillBeBonus->value,
+                            AchievementSetType::Specialty->value,
+                            AchievementSetType::WillBeSpecialty->value,
+                        ])->exists())
+                    ->label(fn () => $this->hasWillBeTypes($game) ? 'Enable Multiset' : 'Disable Multiset')
+                    ->icon(fn () => $this->hasWillBeTypes($game) ? 'heroicon-o-play' : 'heroicon-o-pause')
```
Jamiras: The pause icon doesn't really look like much of an icon. I wonder if this button would be better without an icon?

Image

wescopeland: Removed it in latest:
Screenshot 2026-01-02 at 4 11 45 PM

app/Platform/Enums/AchievementSetType.php
wescopeland: Now is probably a good opportunity to get rid of `will_be_exclusive`.

I've attached a SQL query to the PR description and I've removed it from the codebase in latest.

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
Jamiras: Do we still want the play button on the Enable Mulitset?

wescopeland: I think it's a nice-to-have. When quickly visually scanning the actions, it gives a subtle visual distinction between itself and the attach button.

PR #4298, opened 2025-12-23
perf(AchievementsRelationManager): reduce query count and optimize loading

wescopeland:
`AchievementsRelationManager` currently has an N+1 query problem, causing Livewire to kick off potentially hundreds of SQL queries for every re-render. Because the table itself is lazy-loaded, the queries are only observable by mounting logging explicitly into the relation manager's lifecycle hooks with `DB::enableQueryLog()`. I only started digging into this once I noticed the table felt slow when it loads hundreds of achievements.

This PR makes several optimizations to make the table load faster. I tested with http://localhost:64000/manage/games/20255. The optimizations are:

1. The `developer` relationship is now eager loaded. Before this change, each row was kicking off a separate query to figure out who the developer of the achievement was.
2. The core achievement set is now cached. The new achievement group column and several actions all need the core achievement set - without caching all these things were querying for it multiple times.
3. Achievement group assignments are now fetched in a single query. This was another N+1 problem - each row was fetching for its group.
4. `checkIfRecordIsSelectableUsing()` was changed to `selectable()`. `selectable()` is evaluated only once on mount - not per row.
5. I added permission caching with `once()`. Each row is individually resolving its perm checks. This optimization eliminates thousands of function calls that were causing overhead in Livewire's renderer.

**How to test this PR**
Load the achievements table for a game (such as the one linked above) and try to perform some normal actions as a smoke test. Things should continue functioning normally. It still takes a noticeable second to mount the table, but it's faster on my local machine.


app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -343,9 +344,9 @@ public function table(Table $table): Table
                     DeleteAction::make(),
                 ])
                     ->visible(fn (Achievement $record): bool => $this->canReorderAchievements()
-                        || $user->can('assignMaintainer', Achievement::class)
-                        || $user->can('manage', AchievementGroup::class)
-                        || $user->can('delete', $record)
+                        || $this->canAssignMaintainer()
+                        || $this->canManageAchievementGroups()
+                        || request()->user()->can('delete', $record)
```
Jamiras, line 349: `$user` is `Auth::user()`, which I believe is the same as `request()->user()`, so why change it?

Jamiras (approved):

app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
wescopeland: PHPStan was giving me a hassle here, and the only way to fix it while using `Auth::user()` was to make all this much more verbose.

PR #4299, opened 2025-12-23
feat(api): add V2 JSON:API endpoint for the `AchievementSet` resource

wescopeland:
This PR adds a V2 API endpoint for the `AchievementSet` resource. There are several nuances with this resource that should be noted:

1. It's a `show` endpoint only. There is no index/list endpoint for achievement sets. Consumers should use `GET /games/{game}?include=achievementSets` to retrieve achievement sets for a specific game.
2. The `type` attribute is contextual. When an achievement set is accessed via a game's `?include=achievementSets`, the `type` is populated from the pivot. When accessed directly via `GET /achievement-sets/{achievementSet}`, it always returns `null`.
3. Subset backing games are forcibly excluded. Both the `gameIds` attribute and `games` relationship exclude these games so API consumers don't somehow find themselves reliant on them making it harder for us to remove them.

Achievements are still out of scope. That will probably come in the next PR.

**How to test this PR**
```bash
# Get a specific achievement set
curl --request GET \
  --url http://localhost:64000/api/v2/achievement-sets/9534 \

# Get an achievement set with its linked games included
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievement-sets/9534?include=games' \

# Get a game with its achievement sets included (the recommended approach)
curl --request GET \
  --url 'http://localhost:64000/api/v2/games/1?include=achievementSets' \
```


app/Api/V2/Games/GameResource.php
```diff
@@ -58,6 +58,7 @@ public function attributes($request): iterable
     public function relationships($request): iterable
     {
         return [
+            'achievementSets' => $this->relation('achievementSets')->withoutLinks(),
```
Jamiras, line 61: Should this be `achievement-sets` to be consistent with the read endpoint?

app/Api/V2/AchievementSets/AchievementSetResource.php
```diff
@@ -0,0 +1,145 @@
...
+            'achievementsPublished' => $this->resource->achievements_published,
+            'achievementsUnpublished' => $this->resource->achievements_unpublished,
+
+            'badgeUrl' => $this->resource->image_asset_path_url,
+
+            'achievementsFirstPublishedAt' => $this->resource->achievements_first_published_at,
+            'createdAt' => $this->resource->created_at,
+            'updatedAt' => $this->resource->updated_at,
+
+            'type' => $this->getTypeFromPivot(),
```
Jamiras: > The type attribute is contextual. When an achievement set is accessed via a game's ?include=achievementSets, the type is populated from the pivot. When accessed directly via GET /achievement-sets/{achievementSet}, it always returns null.

I find it confusing that it always returns null when accessing the set directly. This information should be available. The only reason I could see not providing it is if it somehow could have multiple values from being associated to multiple games.

app/Api/V2/AchievementSets/AchievementSetSchema.php
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public function fields(): array
+    {
+        return [
+            ID::make(),
+
+            Str::make('title')->readOnly(),
+
+            Number::make('pointsTotal', 'points_total')->readOnly()->sortable(),
+            Number::make('pointsWeighted', 'points_weighted')->readOnly()->sortable(),
```
Jamiras: What does it mean to allow these be sorted if there's no index?

app/Api/V2/Games/GameResource.php
wescopeland: [It shouldn't.](https://jsonapi.org/recommendations/#naming)

> Member names SHOULD be camel-cased (i.e., wordWordWord)

app/Api/V2/AchievementSets/AchievementSetResource.php
wescopeland: Tricky - here's what I've done in latest:
* `gameIds` been removed.
* `type` has been renamed to `types` and returns an array of gameId and type pairs. "Subset games" are excluded.

app/Api/V2/AchievementSets/AchievementSetSchema.php
wescopeland: This is cruft from this branch originally having an index that I later removed. Fixed.

app/Api/V2/Games/GameResource.php
Jamiras: It still seems counter-intuitive to me:
```
      "relationships":{
         "achievementSets":{
            "data":[
               {
                  "type":"achievement-sets",
                  "id":"576"
               },
```

But I would probably expect the achievement sets to just be inlined there too instead of returned as separate records later in the response.

```
   "included":[
      {
         "type":"achievement-sets",
         "id":"576",
         "attributes":{
            "title":null,
            "pointsTotal":901,
            "pointsWeighted":2534,
```

Jamiras (approved):

PR #4300, opened 2025-12-23
fix(user): remediate stable username leak on Completion Progress page

wescopeland:
http://localhost:64000/user/Darky/progress?filter%5Bsystem%5D=37&page%5Bnumber%5D=1&filter%5Bstatus%5D=unawarded

**Before**
Screenshot 2025-12-23 at 6 29 59 PM

**After**
Screenshot 2025-12-23 at 6 29 53 PM


resources/views/pages/user/[User]/progress.blade.php
Comment (Jamiras) [inline resources/views/pages/user/[User]/progress.blade.php:1]: The "reset" button still uses the stable username.

Image

The avatar tooltip is also keyed on the stable username, though it's not visible without viewing source (which also leaks the stable username via the avatar icon name).

Image

https://github.com/RetroAchievements/RAWeb/blob/28d94253f5083b2be03a408cfc106384f9ec7792/app/Helpers/render/user.php#L66

Comment (wescopeland) [reply resources/views/pages/user/[User]/progress.blade.php:1]: Thanks, should be fixed now.

PR #4302, opened 2025-12-24
chore(composer): add passport key generation to setup block

drisc:
Adds a couple of lines to `composer.json` to generate the OAuth keys required for tests and client development.

- A comment explaining that the keys are being generated along with an assurance about if it errors.
- `php artisan passport:keys` without the `--force` option so that it does not break existing OAuth applications that the user may be testing in dev mode and does not overwrite production keys if run in error on production servers.


wescopeland:
Thanks for iterating on this. The output is definitely cleaner now!

To clarify my earlier comment: when I mentioned matching the established pattern, I meant using the same inline `file_exists()` approach that the other setup lines use.

Creating a dedicated Artisan command makes sense for things that are run independently, but this is "check if files exist, and if not, run another command". What the Artisan command buys is slightly nicer INFO-styled output messages. What it costs is 30 more lines of code to maintain and indirection from the _composer.json_ file.

The simpler solution that follows the existing pattern would be a single line in _composer.json_. I think that's the direction we should go.

drisc:
Gotcha, I've removed the new command and worked to make a clean one line implementation following the existing pattern.

PR #4303, opened 2025-12-24
chore: modernize the LeaderboardDef table schema

wescopeland:
This PR modernizes the `LeaderboardDef` table by renaming it to `leaderboards` and updating all legacy column names:
```
ID -> id
GameID -> game_id
Title -> title
Description -> description
Format -> format
LowerIsBetter -> rank_asc
DisplayOrder -> order_column
Created -> created_at
Updated -> updated_at
Mem -> trigger_definition
```

I've updated all affected:
* Model files.
* Action classes.
* Filament resources.
* API endpoints.
* Legacy helper files.
* Blade templates.
* DTO classes.
* Policies.
* The factory.
* Tests.

`Mem` (now `trigger_definition`) is deliberately not dropped because it's effectively a denormalized value for the latest trigger version. I think it's beneficial to keep, especially for games like rhythm games that have lots of leaderboards.

There will be a brief blip in prod for about 5 seconds where server errors may be encountered. That's probably okay.


Jamiras (commented):
Only did basic testing. Seems okay.

This will have conflicts with #4210. Suggest resolving them here, rather than pushing that burden onto drisc.

wescopeland:
Conflicts should all now be resolved.

As these get merged, they're going to inevitably cause conflicts with each other. I'll clean them up as I go.

PR #4304, opened 2025-12-24
chore: modernize the EmailConfirmations table schema

wescopeland:
This PR modernizes the EmailConfirmations table by renaming it to `email_confirmations` and updated all legacy column names:

```
EmailCookie -> email_cookie
Expires -> expires_at (also converted from date to datetime)
```
`User` is dropped from the table.

Note: **the migration writes data** (copying existing `Expires` values to `expires_at` before dropping the old column) which is typically undesirable. However, I feel it's warranted here to preserve existing expiration dates without needing a complicated SQL command song and dance or multiple deploys to fix the table.

### DIFF
--- a/app/Support/Redirector/LegacyRedirector.php
+++ b/app/Support/Redirector/LegacyRedirector.php
@@ -70,7 +70,7 @@ class LegacyRedirector implements Redirector
             // remove remaining, unused markers
             $parsedRedirectUrl = Url::fromString($redirectUrl);
             foreach ($parsedRedirectUrl->getAllQueryParameters() as $queryParameter => $queryParameterValue) {
-                if (str_starts_with($queryParameterValue, '{')) {
+                if (is_string($queryParameterValue) && str_starts_with($queryParameterValue, '{')) {
                     $parsedRedirectUrl = $parsedRedirectUrl->withoutQueryParameter($queryParameter);
                 }
             }


database/migrations/2025_12_31_000000_update_email_confirmations_table.php
```diff
@@ -0,0 +1,55 @@
...
+
+        Schema::table('EmailConfirmations', function (Blueprint $table) {
+            $table->dropColumn('Expires');
+        });
+
+        Schema::table('EmailConfirmations', function (Blueprint $table) {
+            $table->dateTime('expires_at')->nullable(false)->change();
+        });
+
+        Schema::rename('EmailConfirmations', 'email_confirmations');
```
Jamiras, line 32: No strong opinion from me, but should this be renamed to `user_email_confirmations` or something else?

database/migrations/2025_12_31_000000_update_email_confirmations_table.php
```diff
@@ -0,0 +1,55 @@
...
+        DB::statement('UPDATE EmailConfirmations SET Expires = DATE(expires_at)');
+
+        Schema::table('EmailConfirmations', function (Blueprint $table) {
+            $table->date('Expires')->nullable(false)->change();
+        });
+
+        Schema::table('EmailConfirmations', function (Blueprint $table) {
+            $table->dropColumn('expires_at');
+            $table->renameColumn('email_cookie', 'EmailCookie');
+            $table->string('User', 20)->after('id');
```
Jamiras, line 52: rollback doesn't populate this column, effectively breaking the rollback state. Not that I expect this to be rolled back.

wescopeland: Yea, I agree it's not a perfect rollback. This column is effectively dead in prod anyway.

database/migrations/2025_12_31_000000_update_email_confirmations_table.php
wescopeland: If you don't feel terribly strong, I do lean more towards `email_confirmations` for this one. That seems to match Laravel's existing patterns for auth-related stuff (`personal_access_tokens`, `sessions`, etc - none are prefixed with `user_*`).

Jamiras: It may be more similar to Laravel's implementation, but we seem to prefix child tables with the parent record.

Like I said, it's not a strong opinion, more of an observation.

PR #4305, opened 2025-12-24
chore: modernize the Console table schema

wescopeland:
This PR modernizes the Console table by renaming it to `systems` and updating all legacy column names:
```
ID -> id
Name -> name
Created -> created_at
Updated -> updated_at
```

Note: you will need to run `sail artisan cache:clear` after the migration.


database/migrations/2025_12_24_000002_update_console_table.php
```diff
@@ -0,0 +1,85 @@
...
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        // SQLite doesn't support dropping foreign keys by name.
+        $isSqlite = DB::connection()->getDriverName() === 'sqlite';
+
+        // 1. Drop all FK constraints.
```
Jamiras: Out of curiousity, why is this necessary? Can you not rename the table if there's a foreign key reference to it? Wouldn't the rename be smart enough to update that link?

wescopeland: Think I got a little too defensive with these. It'd be necessary for an earlier 10.x release of MariaDB, but I didn't realize the version we use in prod should handle this automagically. I've significantly simplified this.

PR #4307, opened 2025-12-25
test(game): add GameControllerShowTest

wescopeland:
This PR adds comprehensive coverage for `GameController::show()`.

I decided to leverage Pest syntax for the new _GameControllerShowTest.php_ test suite. This had the benefit of letting me use `describe` blocks for logical test case grouping, more verbose test case descriptions, and convenience helpers like `actingAs()` and `get()`.

`BuildGameShowPagePropsAction` should be refactored, but I'd prefer to have it guarded by a test suite before undertaking that work.

### DIFF
--- a/tests/Feature/Platform/Controllers/GameControllerShowTest.php
+++ b/tests/Feature/Platform/Controllers/GameControllerShowTest.php
@@ -33,6 +33,7 @@ use App\Platform\Enums\AchievementFlag;
 use App\Platform\Enums\AchievementSetType;
 use App\Platform\Enums\GameSetRolePermission;
 use App\Platform\Enums\GameSetType;
+use App\Platform\Enums\LeaderboardState;
 use App\Platform\Services\EventHubIdCacheService;
 use Database\Seeders\RolesTableSeeder;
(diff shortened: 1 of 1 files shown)


wescopeland:
I agree the `RefreshDatabase` overhead can be frustrating. 

As best as I can tell, `describe()` _may_ support scoped database refreshes, but this may require more investigation. Similar to Vitest, Pest allows putting lifecycle hooks like `beforeAll()` and `afterAll()` in a describe block. Maybe this means we can migrate at the block-level if we roll a little bit of custom code to pull this off (instead of leaning on `RefreshDatabase`).

Although one thing is curious: as best as I understand it, `RefreshDatabase` should only be running `migrate:fresh` [once per test suite](https://github.com/guiwoda/laravel-framework/blob/3c888b6c7b89c3d3f90e9024ffbebed3ee80bd23/src/Illuminate/Foundation/Testing/RefreshDatabase.php#L52), not per test case. During test case execution, work is done in a transaction, which is then rolled back at the end of the test case.

I suspect our bottleneck may be elsewhere, but I haven't dug into it yet to be sure.

PR #4308, opened 2025-12-25
docs: fix translations guide link in CONTRIBUTING.md

rammba:
Currently, opening translation link using contributing file from the root (https://github.com/RetroAchievements/RAWeb?tab=contributing-ov-file) results in non-existing page (https://github.com/RetroAchievements/RAWeb/blob/master/TRANSLATIONS.md).

This PR fixes wrong behavior using absolute path to the file, similar to how `README.md` file is referenced:
```
**PHP CS Fixer** comes as a dev dependency. Use it before you commit. See the [README](https://github.com/RetroAchievements/RAWeb/blob/master/README.md).
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -100,4 +100,4 @@ Route keys for route model binding should be validated in the respective RouteSe
 
 ## Translating the Application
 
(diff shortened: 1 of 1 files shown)


wescopeland (approved):
Thanks!

PR #4311, opened 2025-12-26
fix(claims): when a mod reactivates a claim, don't return an Unauthorized error

wescopeland:
Discovered while implementing/testing https://github.com/RetroAchievements/RAWeb/pull/4310.

When managing claim history, if a claim is Completed, mods/admins cannot currently set the claim back to being "Active" due to a logic error in the controller. The wrong policy (`review`) is currently being selected in `master`.

Related: https://discord.com/channels/310192285306454017/1068321249702445126/1453274199450718321


app/Community/Controllers/AchievementSetClaimController.php
Jamiras, line 1: > When managing claim history, if a claim is Completed or Dropped, mods/admins cannot currently set the claim back to being "Active" due to a logic error in the controller. 

As an admin, if the field is Dropped, it's disabled and cannot be changed.

If it's Completed, I can set it back to Active, but still get an error trying to change it back to Completed.

wescopeland: > As an admin, if the field is Dropped, it's disabled and cannot be changed.

Let's keep this as-is. I've updated the description accordingly.

> If it's Completed, I can set it back to Active, but still get an error trying to change it back to Completed.

Good catch. I've fixed this and added another test case to verify the behavior.

Jamiras: Dropped does work, I just can't reactivate my own dropped claims.

PR #4312, opened 2025-12-26
fix(user): remediate issue with 'Hide completed games' toggle

### LINKED ISSUE #4309: Profile completion sidebar is now empty in some cases it shouldn't be
Introduced in #4279.

For users with more than 200 mastered/completed games and accounts with the "hide completed games" preference on, this section of the sidebar will appear totally empty - [example](https://retroachievements.org/user/scatter). This happens because the API call populating this data now grabs 200 games sorted by completion progress and is filtered after the fact when you toggle the setting.

Image

Not sure if this tradeoff was discussed internally or just unnoticed, but it feels like a solvable problem to me. My suggestion would be to do two limited API calls on page load, one with mastered games and one without, save their results in page state. Switch between the lists when the checkbox is toggled. This would be an extra call on load, but it's already doing an extra API call to get the full list so maybe this wouldn't be too much of a drawback? (Could also just filter the full list clientside but I assume that's a perf hiccup on large enough lists or else there wouldn't be two API calls in the first place.) There are a few other solutions/workarounds I can think of but they have what I would consider worse tradeoffs.

wescopeland:
Resolves #4309.

http://localhost:64000/user/Kinglink

On `master`, enabling the "Hide completed games" toggle causes 0 rows to appear, because the current set of 200 rows returned represents a complete list of fully completed games.

Now, we fetch two lists on page load: the current list of 200 games, and a list of up to 200 incomplete games. Based on the toggle status, we toggle between which list of rows is being displayed.

The memory bump on the page is modest - I'm at ~45MB per hit locally. I was originally set on not fixing this because I don't want to introduce a new HTTP call to resolve the issue that will need to be thrown away when `user2` rolls around, but this solution feels like a nice happy medium.

Screenshot 2025-12-26 at 8 38 47 AM

### DIFF
--- a/app/Connect/Actions/BuildClientPatchDataAction.php
+++ b/app/Connect/Actions/BuildClientPatchDataAction.php
@@ -13,6 +13,7 @@ use App\Models\PlayerGame;
 use App\Models\Role;
 use App\Models\User;
 use App\Platform\Enums\AchievementFlag;
+use App\Platform\Enums\LeaderboardState;
 use App\Platform\Services\VirtualGameIdService;
 use Illuminate\Database\Eloquent\Collection;
 use InvalidArgumentException;
@@ -185,6 +186,7 @@ class BuildClientPatchDataAction
 
         // TODO detach leaderboards from games
         $leaderboards = $game->leaderboards()
+            ->where('state', LeaderboardState::Active) // only active leaderboards
             ->orderBy('DisplayOrder') // explicit display order
             ->orderBy('ID')           // tiebreaker on creation sequence
             ->get();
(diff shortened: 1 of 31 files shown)


resources/views/pages-legacy/userInfo.blade.php
```diff
@@ -59,6 +59,7 @@
 $totalSoftcoreAchievements = $progressionCounts['totalSoftcoreAchievements'];
 
 $userCompletedGamesList = getUsersCompletedGamesAndMax($userPage, limit: 200);
+$userIncompleteGamesList = getUsersCompletedGamesAndMax($userPage, limit: 200, isExcludingCompleted: true);
```
Jamiras, line 62: Possible optimization:

```
if (count($userCompletedGamesList) < 200):
    $userIncompleteGamesList = $userCompletedGamesList->filter(...)
```

Though the benefit of avoiding the second query is minimal if there's a limited number of entries already.

wescopeland: Good idea. Tested locally, but doesn't seem to make much/any of a dent in page load performance. Given that, simplicity may be desirable here - I'll hang on to the current implementation unless this comes back to bite us after a deploy.

PR #4313, opened 2025-12-26
chore: modernize the SetRequest table schema

wescopeland:
This PR modernizes the SetRequest table by renaming it to `user_game_list_entries` and updating all legacy column names:
```
GameID -> game_id
Updated -> updated_at
```
An index is also renamed:
```
setrequest_gameid_type_index -> user_game_list_entries_game_id_type_index
```

### DIFF
--- a/app/Models/UserGameListEntry.php
+++ b/app/Models/UserGameListEntry.php
@@ -7,7 +7,6 @@ namespace App\Models;
 use App\Community\Enums\UserGameListType;
 use App\Support\Database\Eloquent\BaseModel;
 use Database\Factories\UserGameListEntryFactory;
-use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Support\Carbon;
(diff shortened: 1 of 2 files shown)


app/Models/UserGameListEntry.php
```diff
@@ -118,9 +112,16 @@ public function game(): BelongsTo
...
+            ->select('user_game_list_entries.*')
+            ->leftJoin('Achievements', 'user_game_list_entries.game_id', '=', 'Achievements.GameID')
+            ->groupBy(
+                'user_game_list_entries.id',
+                'user_game_list_entries.user_id',
+                'user_game_list_entries.game_id',
+                'user_game_list_entries.type',
+                'user_game_list_entries.created_at',
+                'user_game_list_entries.updated_at'
+            )
```
wescopeland: Imagine we have rows that look something like this:

| id  | user_id | game_id | type | created_at |
|-----|---------|---------|------|------------|
| 1   | Alice   | 1   | play | Monday     |
| 2   | Bob     | 1   | dev  | Tuesday    |
| 3   | Alice   | 2   | play | Wednesday  |

The original code said "group by `game_id`, but give me _all_ the columns."

When grouping by `game_id`, rows 1 and 2 both have the same game ID, so they get combined into one row, but there's ambiguity on which `user_id` the result should show, as well as which `created_at` value.

In strict SQL mode, this should actually be an error. In non-strict mode, which I assume we're running, the database just picks arbitrarily.

This code change for the `->groupBy()` tells the database to "group by all these columns together". Now a row is only grouped with another row if everything specified matches. Since `id` is unique, each row stays separate, which is what we actually want here - we're just filtering out rows whose games have achievements, not actually combining rows.

Jamiras: As best as I can tell, this function is only used by a single unit test. Would it be better to just remove it?

wescopeland: Agreed, went ahead and removed this stuff.

PR #4315, opened 2025-12-26
chore: modernize the SiteAwards table schema

wescopeland:
This PR modernizes the SiteAwards table by renaming it to `user_awards` and updating all legacy column names:
```
AwardDate -> awarded_at
AwardType -> award_type
AwardData -> award_data
AwardDataExtra -> award_data_extra
DisplayOrder -> order_column
```
Indexes are also renamed:
```
siteawards_awardtype_index -> user_awards_award_type_index
siteawards_awarddata_awardtype_awarddate_index -> user_awards_award_data_award_type_awarded_at_index
siteawards_user_id_index -> user_awards_user_id_index
siteawards_user_id_awarddata_awardtype_awarddataextra_index -> user_awards_user_id_award_data_award_type_award_data_extra_index
```

Additionally, the `AwardType` enum has been converted from an abstract class with integer constants to a native PHP enum with string backing:
- 1 -> "mastery"
- 2 -> "achievement_unlocks_yield"
- 3 -> "achievement_points_yield"
- 6 -> "patreon_supporter"
- 7 -> "certified_legend"
- 8 -> "game_beaten"
- 9 -> "event"

The enum includes a few methods to facilitate V1 API and legacy UI backwards compatibility.

The migration **mutates table data**. Additionally, the migration deletes orphaned awards with deprecated types 4 and 5, of which there are several thousand orphaned rows in the prod DB.

### DIFF
--- a/resources/js/features/games/components/AchievementSetCredits/ClaimantsDisplay/ClaimantsDisplay.test.tsx
+++ b/resources/js/features/games/components/AchievementSetCredits/ClaimantsDisplay/ClaimantsDisplay.test.tsx
@@ -33,11 +33,11 @@ describe('Component: ClaimantsDisplay', () => {
     const achievementSetClaims = [
       createAchievementSetClaim({
         user: createUser({ displayName: 'Alice' }),
-        finishedAt: '2025-12-31T23:59:59Z', // !! future date shows "Expires"
+        finishedAt: '2099-12-31T23:59:59Z', // !! future date shows "Expires"
       }),
       createAchievementSetClaim({
         user: createUser({ displayName: 'Bob' }),
-        finishedAt: '2025-06-30T23:59:59Z', // !! future date shows "Expires"
+        finishedAt: '2099-06-30T23:59:59Z', // !! future date shows "Expires"
       }),
     ];
 


app/Models/PlayerBadge.php
```diff
@@ -16,26 +16,26 @@ class PlayerBadge extends BaseModel
...
     protected $fillable = [
         'user_id',
-        'AwardType',
-        'AwardData',
-        'AwardDataExtra',
-        'AwardDate',
-        'DisplayOrder',
+        'award_type',
+        'award_data',
+        'award_data_extra',
```
Jamiras: Would it make sense to name these: `award_type`, `award_id`, `award_extra` with the type/id acting like a morph map even though type isn't actually a model type?

Jamiras: Possibly consider `award_tier` instead of `award_extra` as it's effectively 0/1 for softcore/hardcore on Mastery and GameBeaten, and the tier earned for Event records.

wescopeland: In latest, I went with:
`AwardData` -> `award_key`
`AwardDataExtra` -> `award_tier`

I felt bearish on `award_id` because, whlie it works for Mastery/GameBeaten (via game_id) and Events (via event_id), for others like dev contribution badges, it's more akin to a tier index rather than an ID.

Jamiras (approved):

PR #4321, opened 2025-12-28
chore: modernize the Comment table schema

wescopeland:
This PR modernizes the Comment table by renaming it to `comments` and updating all legacy column names:
```
ID -> id
ArticleID -> commentable_id
Payload -> body
Submitted -> created_at
Edited -> updated_at
```
Indexes are also renamed:
```
comment_articleid_index -> comments_commentable_id_index
comment_user_id_submitted_index -> comments_user_id_created_at_index
```
Additionally, the `ArticleType` column is replaced with `commentable_type`, and the `ArticleType` abstract class has been converted to a native PHP enum with string backing called `CommentableType`:
* 1 -> "game.comment"
* 2 -> "achievement.comment"
* 3 -> "user.comment"
* 5 -> "user-activity.comment"
* 6 -> "leaderboard.comment"
* 7 -> "trigger.ticket.comment"
* 9 -> "user-moderation.comment"
* 10 -> "game-hash.comment"
* 11 -> "achievement-set-claim.comment"
* 12 -> "game-modification.comment"
Most of these string backing names already existed in the codebase from V2. They're set up this way so we can eventually support a `Commentable` morph, but that is out of scope for this PR.

The enum includes `toLegacyInteger()` and `fromLegacyInteger()` methods for backwards compatibility with legacy APIs and PHP UIs.

The migration **mutates table data**. Additionally, the migration deletes orphaned comments with deprecated or unexpected ArticleType values, such as 0.

The `NewsComment` and `NewsCommentPolicy` classes have been removed.

### DIFF
--- a/app/Platform/Actions/BuildGameShowPagePropsAction.php
+++ b/app/Platform/Actions/BuildGameShowPagePropsAction.php
@@ -755,10 +755,10 @@ class BuildGameShowPagePropsAction
             ->values();
 
         if (!$activeOnly) {
-            // Sort: Active/Unpublished first, Disabled last, then by DisplayOrder.
+            // Sort: Active/Unpublished first, Disabled last, then by order_column.
             $leaderboards = $leaderboards->sortBy([
                 fn ($leaderboard) => $leaderboard->state === LeaderboardState::Disabled ? 1 : 0,
-                fn ($a, $b) => $a->DisplayOrder <=> $b->DisplayOrder,
+                fn ($a, $b) => $a->order_column <=> $b->order_column,
             ])->values();
         }
 


database/migrations/2025_12_31_000010_update_comment_table.php
```diff
@@ -0,0 +1,145 @@
...
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->dropIndex('comment_articleid_index');
+            $table->dropIndex('comment_user_id_submitted_index');
+        });
+        if ($driver !== 'sqlite') {
+            DB::statement('SET FOREIGN_KEY_CHECKS=1');
+        }
+
+        Schema::table('Comment', function (Blueprint $table) {
+            $table->renameColumn('ArticleID', 'commentable_id');
```
Jamiras, line 53: Because the `commentable_type` column is kept and this column is renamed, `commentable_type` appears out-of-order:
```
> select * from comments where id=12345\G
              id: 12345
  commentable_id: 5017
         user_id: 11371
commentable_type: achievement.comment
            body: It's somewhat RNG-dependent.  He got really good spawns in the vertical shaft.  If you get the wrong ones, it's not happening no matter how good you are.
      created_at: 2015-09-14 20:36:41
      updated_at: NULL
      deleted_at: NULL
1 row in set (0.002 sec)
```

It should be before `commentable_id`.

app/Community/Enums/CommentableType.php
```diff
@@ -0,0 +1,96 @@
...
+use Spatie\TypeScriptTransformer\Attributes\TypeScript;
+
+// TODO eventually use commentable morph
+
+#[TypeScript]
+enum CommentableType: string
+{
+    case Achievement = 'achievement.comment';
+    case AchievementTicket = 'trigger.ticket.comment';
+    case Forum = 'forum.comment'; // Note: Forum comments use ForumTopicComment table, not comments table. This is for email notifications only.
```
Jamiras: Should this be `forum-topic-comment` or something similar to make the distinction more clear?

resources/js/common/components/CommentList/ReportCommentButton/ReportCommentButton.test.tsx
```diff
@@ -9,7 +8,7 @@ describe('Component: ReportCommentButton', () => {
     // ARRANGE
     const comment = createComment({
       id: 1,
-      commentableType: ArticleType.Game,
+      commentableType: 'game.comment',
```
Jamiras, line 11: Is the new enum not exposed?

Jamiras: There seem to be a lot of references to these values in the next few dozen files. Would it not be better to use an enum/constants?

wescopeland: I definitely see why this may be a point of confusion. It actually _is_ exposed, but `commentableType` is a type-safe string. Any given invalid value will throw an error on `pnpm tsc`. This is visible by starting to type in the field:

Screenshot 2025-12-29 at 2 13 54 PM

In _generated.d.ts_, we automatically have:
```ts
export type Comment = {
  id: number;
  commentableId: number;
  commentableType: App.Community.Enums.CommentableType;
```

We use `App.Community.Enums.CommentableType` to strongly type our route mappings, ie:
Screenshot 2025-12-29 at 2 37 31 PM


The type is generated from this being a native PHP enum that's used in the DTO:
```php
// CommentableType.php
#[TypeScript]
enum CommentableType: string

// CommentData.php
#[TypeScript('Comment')]
class CommentData extends Data
{
    public function __construct(
        public int $id,
        public int $commentableId,
        public CommentableType $commentableType,
```

Previously, the front-end code had to use an `ArticleType` object because abstract class "enums" can't neatly map to type-safe strings. The object thing is a workaround to preserve type safety in the UI for these cases.

database/migrations/2025_12_31_000010_update_comment_table.php
wescopeland: This should be fixed in the latest commit.

app/Community/Enums/CommentableType.php
wescopeland: Went with `forum-topic.comment`.

Jamiras: The `.comment` is the thing I was focusing on. Because all of these end in `.comment`, it's natural to assume they all use the Comment model.

wescopeland: Changed to `forum-topic-comment` in latest.

PR #4329, opened 2025-12-29
chore: modernize the UserAccounts table schema

wescopeland:
This PR modernizes the UserAccounts table by renaming it to `users` and updating almost all legacy column names:
```bash
ID -> id
User -> username
Password -> password
SaltedPass -> legacy_salted_password # we can't drop this column
EmailAddress -> email
email_backup -> email_original
RAPoints -> points_hardcore
RASoftcorePoints -> points
appToken -> connect_token
appTokenExpiry -> connect_token_expires_at
websitePrefs -> preferences_bitfield
LastLogin -> last_activity_at
Motto -> motto
ContribCount -> yield_unlocks
ContribYield -> yield_points
APIKey -> web_api_key
APIUses -> web_api_calls
LastGameID -> rich_presence_game_id
RichPresenceMsg -> rich_presence
RichPresenceMsgDate -> rich_presence_updated_at
UnreadMessageCount -> unread_messages
TrueRAPoints -> points_weighted
UserWallActive -> is_user_wall_active
Created -> created_at
Updated -> updated_at
DeletedRequested -> delete_requested_at
Deleted -> deleted_at
```
Indexes are also renamed:
```
useraccounts_display_name_unique -> users_display_name_unique
useraccounts_ulid_unique -> users_ulid_unique
users_apikey_index -> users_web_api_key_index
users_apptoken_index -> users_connect_token_index
useraccounts_lastlogin_deleted_index -> users_last_activity_at_deleted_at_index
useraccounts_visible_role_id_index -> users_visible_role_id_index
```

The following columns are intentionally left unchanged as their usages will be migrated to other existing columns in a future PR (past the maintenance window):
* `Permissions` - this is a heavy lift to get us fully over to RBAC
* `Untracked` - we need to swap everything over to `unranked_at`
* `ManuallyVerified` - we need to swap everything over to `forum_verified_at`, and I suspect there may be data integrity issues to solve along the way

### DIFF
--- a/app/Helpers/util/mail.php
+++ b/app/Helpers/util/mail.php
@@ -284,7 +284,7 @@ function informAllSubscribersAboutActivity(
     }
 
     $subscriptionService = new SubscriptionService();
-    $subscribers = $subscriptionService->getSegmentedSubscriberIds($subscriptionSubjectType, $articleID, $subjectAuthor?->ID);
+    $subscribers = $subscriptionService->getSegmentedSubscriberIds($subscriptionSubjectType, $articleID, $subjectAuthor?->id);
 
     $notificationService = new SubscriptionNotificationService();
     $notificationService->queueNotifications($subscribers['implicitlySubscribedNotifyLater'],
--- a/public/request/comment/create.php
+++ b/public/request/comment/create.php
@@ -34,7 +34,7 @@ $commentable = match ($articleType) {
     default => null,
 };
 
-$userModel = User::find($userDetails['ID']);
+$userModel = User::find($userDetails['id']);
 if (!$userModel->can('create', [Comment::class, $commentable, $articleType])) {
     return back()->withErrors(__('legacy.error.error'));
 }
--- a/public/request/comment/delete.php
+++ b/public/request/comment/delete.php
@@ -15,7 +15,7 @@ $input = Validator::validate(Arr::wrap(request()->post()), [
 ]);
 
 $comment = Comment::findOrFail((int) $input['comment']);
-$user = User::find($userDetails['ID']);
+$user = User::find($userDetails['id']);
 
 if (!$user->can('delete', $comment)) {
     abort(400);
--- a/public/request/user-game-list/toggle.php
+++ b/public/request/user-game-list/toggle.php
@@ -28,7 +28,7 @@ $typeEnum = UserGameListType::from($typeString);
 $command = '';
 
 /** @var User $user */
-$user = User::findOrFail($userDetails['ID']);
+$user = User::findOrFail($userDetails['id']);
 if ($user->gameListEntries($typeEnum)->where('game_id', $gameId)->exists()) {
     $action = new RemoveGameFromListAction();
     $success = $action->execute($user, $game, $typeEnum);
--- a/public/request/user/update-site-awards.php
+++ b/public/request/user/update-site-awards.php
@@ -52,7 +52,7 @@ $sortedAwards = $parseCsv($sortedCsv);
 
 $awards = array_merge($hiddenAwards, $sortedAwards);
 
-$userId = $userDetails['ID'];
+$userId = $userDetails['id'];
 
 foreach ($awards as $award) {
     $awardType = $award['type'];
--- a/public/request/user/update-subscription.php
+++ b/public/request/user/update-subscription.php
@@ -27,7 +27,7 @@ if (!authenticateFromCookie($user, $permissions, $userDetails, $requiredPermissi
 
 $service = new SubscriptionService();
 $service->updateSubscription(
-    User::find($userDetails['ID']),
+    User::find($userDetails['id']),
     SubscriptionSubjectType::from($subjectType),
     $subjectID,
     $input['operation'] === "subscribe");


database/migrations/2025_12_31_000005_update_useraccounts_table.php
wescopeland, line 1: This is the migration file.

database/migrations/2025_12_29_000000_update_useraccounts_table.php
```diff
@@ -0,0 +1,277 @@
...
+            $table->renameColumn('appToken', 'connect_token');
+            $table->renameColumn('appTokenExpiry', 'connect_token_expires_at');
+            $table->renameColumn('websitePrefs', 'preferences_bitfield');
+            $table->renameColumn('LastLogin', 'last_activity_at');
+            $table->renameColumn('Motto', 'motto');
+            $table->renameColumn('ContribCount', 'yield_unlocks');
+            $table->renameColumn('ContribYield', 'yield_points');
+            $table->renameColumn('APIKey', 'web_api_key');
+            $table->renameColumn('APIUses', 'web_api_calls');
+            $table->renameColumn('LastGameID', 'last_game_id');
```
Jamiras: I wonder if this should be renamed `rich_presence_game_id`?

It's more closely tied to the `rich_presence` than the `last_activity_at`

database/migrations/2025_12_29_000000_update_useraccounts_table.php
```diff
@@ -0,0 +1,277 @@
...
+                    MODIFY COLUMN `two_factor_secret` text DEFAULT NULL AFTER `password`,
+                    MODIFY COLUMN `two_factor_recovery_codes` text DEFAULT NULL AFTER `two_factor_secret`,
+                    MODIFY COLUMN `two_factor_confirmed_at` timestamp NULL DEFAULT NULL AFTER `two_factor_recovery_codes`,
+                    MODIFY COLUMN `legacy_salted_password` varchar(32) NOT NULL AFTER `two_factor_confirmed_at`,
+                    MODIFY COLUMN `remember_token` varchar(100) DEFAULT NULL AFTER `legacy_salted_password`,
+
+                    -- Email
+                    MODIFY COLUMN `email` varchar(64) NOT NULL AFTER `remember_token`,
+                    MODIFY COLUMN `email_verified_at` timestamp NULL DEFAULT NULL AFTER `email`,
+                    MODIFY COLUMN `email_backup` varchar(255) DEFAULT NULL AFTER `email_verified_at`,
```
Jamiras: I wonder if this should be renamed to `email_original` or something similar.

database/migrations/2025_12_31_000005_update_useraccounts_table.php
```diff
@@ -0,0 +1,277 @@
...
+                    MODIFY COLUMN `remember_token` varchar(100) DEFAULT NULL AFTER `legacy_salted_password`,
+
+                    -- Email
+                    MODIFY COLUMN `email` varchar(64) NOT NULL AFTER `remember_token`,
+                    MODIFY COLUMN `email_verified_at` timestamp NULL DEFAULT NULL AFTER `email`,
+                    MODIFY COLUMN `email_backup` varchar(255) DEFAULT NULL AFTER `email_verified_at`,
+
+                    -- Authorization
+                    MODIFY COLUMN `visible_role_id` bigint(20) unsigned DEFAULT NULL AFTER `email_backup`,
+                    MODIFY COLUMN `Permissions` tinyint(4) NOT NULL COMMENT '-2=spam, -1=banned, 0=unconfirmed, 1=confirmed, 2=jr-developer, 3=developer, 4=moderator' AFTER `visible_role_id`,
```
Jamiras, line 100: Was this not renamed/recased because we expect to drop it soon?

database/migrations/2025_12_29_000000_update_useraccounts_table.php
```diff
@@ -0,0 +1,277 @@
...
+
+        // Rename columns.
+        Schema::table('UserAccounts', function (Blueprint $table) {
+            $table->renameColumn('ID', 'id');
+            $table->renameColumn('User', 'username');
+            $table->renameColumn('Password', 'password');
+            $table->renameColumn('SaltedPass', 'legacy_salted_password');
+            $table->renameColumn('EmailAddress', 'email');
+            $table->renameColumn('RAPoints', 'points');
+            $table->renameColumn('RASoftcorePoints', 'points_softcore');
```
Jamiras: I know it a bunch of additional work, but would you consider making these `points_hardcore` and `points` since you're already changing them?

https://github.com/RetroAchievements/RAWeb/pull/4271#discussion_r2641270841

database/migrations/2025_12_29_000000_update_useraccounts_table.php
```diff
@@ -0,0 +1,277 @@
...
+                    MODIFY COLUMN `Permissions` tinyint(4) NOT NULL COMMENT '-2=spam, -1=banned, 0=unconfirmed, 1=confirmed, 2=jr-developer, 3=developer, 4=moderator' AFTER `visible_role_id`,
+
+                    -- Player stats
+                    MODIFY COLUMN `achievements_unlocked` int(10) unsigned DEFAULT NULL AFTER `Permissions`,
+                    MODIFY COLUMN `achievements_unlocked_hardcore` int(10) unsigned DEFAULT NULL AFTER `achievements_unlocked`,
+                    MODIFY COLUMN `completion_percentage_average` decimal(10,9) DEFAULT NULL AFTER `achievements_unlocked_hardcore`,
+                    MODIFY COLUMN `completion_percentage_average_hardcore` decimal(10,9) DEFAULT NULL AFTER `completion_percentage_average`,
+                    MODIFY COLUMN `points` int(11) NOT NULL AFTER `completion_percentage_average_hardcore`,
+                    MODIFY COLUMN `points_softcore` int(11) DEFAULT 0 AFTER `points`,
+                    MODIFY COLUMN `points_weighted` int(10) unsigned DEFAULT NULL AFTER `points_softcore`,
```
Jamiras: Recommend ordering these as `points_softcore`, `points`, and `points_weighted` so the weighted points are next to the hardcore points. It's also consistent with softcore/hardcore ordering on other fields like `achievements_unlocked`.

database/migrations/2025_12_29_000000_update_useraccounts_table.php
wescopeland: How strongly do you feel about this? Philosophically I do agree they should be changed, and the work doesn't scare me much.

What I'm very anxious about with these particular fields is the risk of breakage in changing their names. Because of how they're currently named, it's challenging to search for all possible usages and feel confident when making this particular swap.

Jamiras: I'd say I'm strongly in favor of the change. It's the only place in the database where we have a direct reference to 'softcore', excepting the denormalized value in `player_games`.

If we're going to do it, it seems like now would be the best time.

Ironically, it probably would have been easier to do with the old names.

database/migrations/2025_12_29_000000_update_useraccounts_table.php
wescopeland: Renamed in https://github.com/RetroAchievements/RAWeb/pull/4329/commits/d65ed7520d21078306364c5239efa5848a21b57a to `rich_presence_game_id`. I also adjusted the position of the column.

database/migrations/2025_12_29_000000_update_useraccounts_table.php
wescopeland: Renamed in https://github.com/RetroAchievements/RAWeb/pull/4329/commits/d97334724d1a097ab632bf17bac086d43167df3f to `email_original`.

database/migrations/2025_12_31_000005_update_useraccounts_table.php
wescopeland: Yep, same goes for the other two that weren't renamed. I can accelerate dropping those after the maintenance window - renaming them didn't feel like it was worth the effort.

database/migrations/2025_12_29_000000_update_useraccounts_table.php
wescopeland: Tackled this in https://github.com/RetroAchievements/RAWeb/pull/4329/commits/1e51a98b1594693b28aaac76848677b3b0633a51.

database/migrations/2025_12_29_000000_update_useraccounts_table.php
wescopeland: I believe this should be implicitly fixed by the transition to `points_hardcore` and `points`.

Screenshot 2025-12-30 at 4 56 10 PM

app/Helpers/database/forum.php
```diff
@@ -305,7 +305,7 @@ function generateGameForumTopic(User $user, int $gameId): ?ForumTopicComment
     $wikipediaURL = "https://www.google.com/search?q=site:en.wikipedia.org+$urlSafeGameTitle";
 
     $topicPayload = "Official Topic Post for discussion about [game=$gameId]\n" .
-        "Created " . date("j M, Y H:i") . " by [user={$user->User}]\n\n" .
+        "Created " . date("j M, Y H:i") . " by [user={$user->username}]\n\n" .
```
Jamiras: This should probably be `$user->display_name`.

app/Models/User.php
```diff
@@ -90,81 +90,50 @@ class User extends Authenticatable implements CommunityMember, Developer, HasLoc
...
+        'locale_time',
         'ManuallyVerified',
-        'Motto',
+        'motto',
         'muted_until',
         'password', // fillable for registration
         'PasswordResetToken', // fillable for when users are banned
         'Permissions',
+        'points_hardcore',
+        'points',
```
Jamiras, line 121: minor: these aren't sorted alphabetically

Jamiras: also applies to `$visible` and `$casts`.

public/API/API_GetUserRankAndScore.php
```diff
@@ -25,8 +25,8 @@
 $softcorePoints = 0;
 
 if ($user) {
-    $points = $user?->points ?? 0;
-    $softcorePoints = $user?->points_softcore ?? 0;
```
Jamiras, line 29: I didn't realize we already had helpers/aliases for these. I can see how that would make the conversion more complicated. Thanks for doing it.

wescopeland: No worries - it actually turned out to be not as bad as I thought it was going to be.

app/Models/User.php
wescopeland: Fixed in latest.

app/Helpers/database/forum.php
wescopeland: Fixed in latest.

PR #4333, opened 2025-12-30
chore: modernize the Ticket table schema

wescopeland:
This PR modernizes the Ticket table by renaming it to `tickets` and updating all legacy column names:
```
ID -> id
ticketable_model -> ticketable_type
ReportType -> type
ReportState -> state
ReportNotes -> body
ReportedAt -> created_at
ResolvedAt -> resolved_at
Updated -> updated_at
Hardcore -> hardcore
```

The `AchievementID` column is dropped. Its data is migrated to the existing `ticketable_type`/`ticketable_id` morph columns.

Additionally, the `TicketState` and `TicketType` abstract class "enums" have been converted to native PHP enums with string backing.

The migration mutates table data. Additionally, the migration deletes tickets with invalid type values.


database/migrations/2025_12_29_000001_update_ticket_table.php
```diff
@@ -0,0 +1,211 @@
...
+            $table->dropIndex('ticket_achievementid_reportstate_deleted_at_index');
+        });
+
+        // Drop the AchievementID column.
+        Schema::table('Ticket', function (Blueprint $table) {
+            $table->dropColumn('AchievementID');
+        });
+
+        // Rename the table.
+        Schema::rename('Ticket', 'trigger_tickets');
```
Jamiras: I understand the reasoning behind this, but is it really necessary? It feels exceptionally verbose and possibly restricting. It would be like renaming `Comment` to `article_comments`.

Will we ever have tickets for things that aren't trigger-based? Things like "Unwelcome concept", "Incorrect type", or "Grammatical error" are ticket types, but would still apply to triggerable objects. Besides, it looks like the schema is already ready if we do: `ticketable_type` is not `trigger_type`.

wescopeland: Updated in latest to use `tickets` as the table name and `Ticket*` for class/enum names. I've changed all `TriggerTicket` references to use `Ticket` instead.

app/Models/Achievement.php
```diff
@@ -564,7 +564,7 @@ public function visibleComments(?User $user = null): HasMany
      */
     public function tickets()
     {
-        return $this->hasMany(Ticket::class, 'AchievementID');
+        return $this->hasMany(Ticket::class, 'ticketable_id');
```
Jamiras: This should probably also filter on ticketable_type=achievement.

app/Models/Game.php
```diff
@@ -901,7 +901,7 @@ public function visibleLeaderboards(): HasMany
      */
     public function tickets(): HasManyThrough
     {
-        return $this->hasManyThrough(Ticket::class, Achievement::class, 'GameID', 'AchievementID', 'ID', 'ID');
+        return $this->hasManyThrough(Ticket::class, Achievement::class, 'GameID', 'ticketable_id', 'ID', 'ID');
```
Jamiras: Does this need a ticketable_type fiter?

database/migrations/2025_12_29_000001_update_ticket_table.php
Jamiras: Thanks. I didn't realize we already had that many things labelled as `TriggerTicket`.

app/Models/Game.php
wescopeland: I think we should hold off on adding one here. Returning all ticketable types seems accurate when referencing at the `$game->tickets()` level.

app/Models/Achievement.php
wescopeland: Agreed. Added in https://github.com/RetroAchievements/RAWeb/pull/4333/commits/a8c95e41fca905187329b554d86d9b80de7be79e.

PR #4340, opened 2025-12-31
fix: force Date parsing in UTC on history best day graph

### LINKED ISSUE #2402: History Best Days graph sending to the wrong date

Chew:
Fixes the data parsing issue by always parsing into UTC.

Unfortunately, this is a pretty hacky fix. Ideally you would just pass in the date a little more intelligently, but to avoid more fluff before a proper rewrite into React, this does the job well enough for now. I can try and do this if we want, but this works fine for now.

This might be a bit difficult to test because of time zones, I had to mess with my timezone to reproduce (then fix) it. You can do this by either changing your computer timezone or going to the "Sensors" tool on Chrome and changing the timezone ID there


wescopeland:
No worries at all, I'm fine with a hacky fix for now on a legacy page. I'll get a review in after I merge all the schema PRs.

wescopeland (approved):
Thanks!

PR #4357, opened 2026-01-03
perf: add author_yield_unlocks to achievements table

wescopeland:
`UpdateDeveloperContributionYieldJob` runs ~400 times per day taking 11 seconds or longer on average. For one achievement on my local env, I was able to hit over 30 seconds for this job.

The bottleneck is this query:
```sql
SELECT SUM(a.points), COUNT(*)
FROM achievements a
JOIN player_achievements pa ON pa.achievement_id = a.id
WHERE a.user_id = :user_id
    AND a.is_promoted = 1
    AND pa.user_id != :user_id2
    AND NOT EXISTS (
        SELECT 1 FROM achievement_maintainer_unlocks amu
        WHERE amu.player_achievement_id = pa.id
    )
```
This can result in many millions of rows needing to be scanned.

This PR:
* Adds an `author_yield_unlocks` column to the `achievements` table.
* Adds a compound `user_id` + `is_promoted` index to the `achievements` table.
* Adds a backfill command for the new column.

The end result is the job goes from taking anywhere from 10-30 seconds to taking ~100ms or less.
```
sail artisan migrate
sail artisan ra:platform:achievement:backfill-author-yield-unlocks
```


app/Platform/Commands/BackfillAuthorYieldUnlocks.php
Jamiras, line 1: I realize it matches the current behavior, but the query is not filtering untracked users, which leads to situations like this:
```
             unlocks_total: 64
          unlocks_hardcore: 54
         unlock_percentage: 0.876712329
unlock_hardcore_percentage: 0.931034483
      author_yield_unlocks: 65
```

In this case, there's 66 `player_achievement` entries. Two are untracked, which leads to the 64 `unlocks_total`, and one is the author, which leads to the 65 `author_yield_unlocks`.

It would be nice if we could just naively use `unlocks_total` instead of having a separate column to maintain.

wescopeland: Unfortunately, I don't think we can use `unlocks_total` for two reasons:
1. `unlocks_total` includes the author's own unlock, which we don't want to count.
2. We still have to subtract `achievement_maintainer_unlocks` entries.

It seems like `author_yield_unlocks` should filter out untracked users for consistency with `unlocks_total`. I've updated the implementation to do this, and added a listener on `PlayerRankedStatusChanged` to keep it in sync when users become untracked/retracked.

Jamiras: I'm definitely seeing the adjustment in the `author_yield_unlocks` column:
```
             unlocks_total: 64
          unlocks_hardcore: 54
         unlock_percentage: 0.876712329
unlock_hardcore_percentage: 0.931034483
      author_yield_unlocks: 63
```
But my total hasn't changed:
```
MariaDB [retroachievements-web]> select yield_points from users where id=43495\G
yield_points: 1856279
1 row in set (0.000 sec)

MariaDB [retroachievements-web]> select sum(points * author_yield_unlocks) from achievements where user_id=43495;
+------------------------------------+
| sum(points * author_yield_unlocks) |
+------------------------------------+
|                            1800216 |
+------------------------------------+
1 row in set (0.003 sec)
```
It seems like the command isn't doing an on-demand update. 

Additionally, while changing a user's tracked status does update the `author_yield_unlocks`, it doesn't trigger any `UpdateDeveloperContributionYield` jobs.

Simiarly, manually unlocking an achievement (and later resetting it) does increment/decrement both the `author_yield_unlocks` and `yield_unlocks`, but does not fully recalculate it.

It looks like the only things that trigger `UpdateDeveloperContributionYield` are a full player reset, or a scheduled job every tuesday.

PR #4358, opened 2026-01-03
feat(seed): Seed less (but better) data

Jamiras:
Updates the seeder to generate about half as many accounts and a quarter as many achievements. This results in a about a 85% faster seeding.

Additionally, provides more gamer-like usernames (PascalCase with numeric suffixes) and non-lorem-ipsem game titles. Achievement titles and descriptions are still lorem-ipsem.

Before:
image

After:
image

Before:
image

After:
image

### DIFF
--- a/app/Community/Actions/GenerateAnnualRecapAction.php
+++ b/app/Community/Actions/GenerateAnnualRecapAction.php
@@ -85,8 +85,8 @@ class GenerateAnnualRecapAction
     {
         $games = PlayerSession::where('user_id', $user->id)
             ->where('duration', '>=', 5)
-            ->where('created_at', '>=', $startDate)
-            ->where('created_at', '<', $endDate)
+            ->where('player_sessions.created_at', '>=', $startDate)
+            ->where('player_sessions.created_at', '<', $endDate)
             ->join('games', 'games.id', '=', 'game_id')
             ->whereNotIn('games.system_id', System::getNonGameSystems())
             ->groupBy('game_id')
@@ -184,6 +184,8 @@ class GenerateAnnualRecapAction
     private function identifyAndMergeSubsets(array &$gameData): array
     {
         $gameIds = array_keys($gameData);
+
+        // these subsets have explicit player_session entries
         $achievementSets = GameAchievementSet::whereIn('game_id', $gameIds)
             ->select(['game_id', 'achievement_set_id'])
             ->where('type', AchievementSetType::Core)
@@ -219,6 +221,24 @@ class GenerateAnnualRecapAction
                     unset($gameData[$subsetGameId]);
                 }
             }
+        }
+
+        // also get any bonus subsets for games that the user played as they might have been played through multiset
+        $bonusAchievementSetIds = GameAchievementSet::whereIn('game_id', $gameIds)
+            ->whereIn('type', [AchievementSetType::Bonus])
+            ->pluck('achievement_set_id')
+            ->toArray();
+
+        // remove any items that we've already processed
+        $bonusAchievementSetIds = array_diff($bonusAchievementSetIds, array_keys($achievementSets));
+
+        if (!empty($bonusAchievementSetIds)) {
+            $bonusCoreSetIds = GameAchievementSet::whereIn('achievement_set_id', $bonusAchievementSetIds)
+                ->where('type', '=', AchievementSetType::Core)
+                ->pluck('game_id')
+                ->toArray();
+
+            $subsetGameIds = array_merge($subsetGameIds, $bonusCoreSetIds);
         }
 
         return $subsetGameIds;
--- a/resources/views/mail/community/annual-recap.blade.php
+++ b/resources/views/mail/community/annual-recap.blade.php
@@ -94,7 +94,7 @@ You spent {{ $recapData['totalPlaytime'] }} playing games {{ $playedSystemsClaus
     width="64"
     height="64"
 >
-    Your rarest achievement earned was <a href="{{ route('achievement.show', $recapData['rarestHardcoreAchievement']) }}">{{ $recapData['rarestHardcoreAchievement']->title }}</a> from {{ $recapData['rarestHardcoreAchievement']->game->Title }}, which has only been earned in hardcore by {{ $recapData['rarestHardcoreAchievementEarnRate'] }}% of players.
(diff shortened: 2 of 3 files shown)


config/filesystems.php
```diff
@@ -122,6 +122,8 @@
          */
         storage_path('app/media/UserPic/_User.png') => public_path('assets/images/user/avatar-safe.png'),
         storage_path('app/media/Images/000001.png') => public_path('assets/images/game/icon-safe.png'),
+        storage_path('app/media/Badge/00000.png') => public_path('assets/images/achievement/badge.png'),
+        storage_path('app/media/Badge/00000_lock.png') => public_path('assets/images/achievement/badge-locked.png'),
```
Jamiras, line 126: Provides a default image for achievements

Image

database/factories/AchievementFactory.php
```diff
@@ -46,7 +46,7 @@ public function definition(): array
             'type' => null,
             'points' => fake()->randomElement($pointValues),
             'points_weighted' => rand(1, 1000),
-            'image_name' => '00001',
+            'image_name' => '00000',
```
Jamiras, line 49: This is the placeholder badge.
https://github.com/RetroAchievements/RAWeb/blob/04f3dc5b90fc9cfd57a878417af1fe78fb7f65f8/app/Connect/Actions/GetBadgeIdRangeAction.php#L38-L40

database/seeders/Concerns/SeedsUsers.php
```diff
@@ -29,6 +32,11 @@ protected function seedUserByUsername(string $username = 'verified', array $attr
         // set the connect token expiry back a few minutes so we can detect if it gets
         // updated to now + expiry_delay
         $user->connect_token_expires_at = $user->connect_token_expires_at->subMinutes(5);
+
+        $faker = Faker::create();
+        $user->created_at = Carbon::parse($faker->dateTimeBetween('-3 years', '-2 hours')->format(DateTime::ATOM));
+        $user->timestamps = false;
```
Jamiras, line 38: Allows users to have a Member Since that predates the seeding.

database/seeders/AchievementSetClaimSeeder.php
```diff
@@ -17,11 +20,48 @@ public function run(): void
...
+            ->limit(60)
+            ->get()->each(function (AchievementSet $set) {
+                AchievementSetClaim::factory()->create([
+                    'game_id' => $set->games()->first()->id,
+                    'user_id' => $set->achievements()->first()->user_id,
+                    'created_at' => $set->achievements_first_published_at->clone()->subDays(rand(2, 50))->subMinutes(rand(1, 600)),
+                    'finished_at' => $set->achievements_first_published_at->clone()->addMinutes(rand(1, 100)),
+                    'status' => ClaimStatus::Complete,
+                ]);
+            });
```
Jamiras, line 36: fixes claims being finished in the future, and for games without achievements.

database/seeders/GameSetsSeeder.php
```diff
@@ -129,7 +129,7 @@ public function run(): void
         $publisherCount = count($publisherHubIds);
         $developerCount = count($developerHubIds);
         $genreCount = count($genreHubIds);
-        foreach (Game::all() as $game) {
+        foreach (Game::orderBy('released_at')->get() as $game) {
```
Jamiras, line 132: Generating titles in release order ensures sequels are dated sequentially.

database/seeders/SystemsTableSeeder.php
```diff
@@ -12,11 +12,32 @@ class SystemsTableSeeder extends Seeder
...
+            13, // Atari Lynx
+            14, // NeoGeo Pocket
+            15, // GameGear
+            16, // GameCube
+            21, // PS2
+            25, // Atari 2600
+            41, // PSP
+            45, // Intellivision
+            57, // Channel F
+        ];
```
Jamiras, line 34: Only set a handful of systems as active. This helps minimize the number of achievements/unlocks generated.

database/seeders/UsersTableSeeder.php
```diff
@@ -26,26 +28,108 @@ public function run(): void
         $this->seedUserByUsername('banned', ['banned_at' => Carbon::now(), 'Permissions' => Permissions::Banned]);
         $this->seedUserByUsername('spammer', ['banned_at' => Carbon::now(), 'Permissions' => Permissions::Spam]);
 
+        // set the password for all users to their username
+        $salt = config('app.legacy_password_salt');
+        User::all()->each(function (User $user) use ($salt) {
+            $user->legacy_salted_password = md5($user->username . $salt);
+            $user->saveQuietly();
+        });
```
Jamiras, line 35: provides a default password for all accounts that's just the username - so you can log in as `CodeReviewer`/`CodeReviewer`.

database/seeders/AchievementSetClaimSeeder.php
```diff
@@ -17,11 +20,48 @@ public function run(): void
...
+            ->get()->each(function (AchievementSet $set) use ($count) {
+                $game = $set->games()->first();
+                if (!isValidConsoleId($game->system_id)) {
+                    return;
+                }
+
+                if ($count === 0) {
+                    return;
+                }
+                $count--;
```
wescopeland, line 52: Bug: `$count === 0` will never be true. Each iteration of `->each()` is going to have its own copy of `$count` with the original 10-25 value. This line is decrementing a local copy. 

I think the intention is for `$count` to be passed by reference.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -21,7 +21,7 @@
 
 class UpdatePlayerGameMetricsAction
 {
-    public function execute(PlayerGame $playerGame, bool $silent = false): void
+    public function execute(PlayerGame $playerGame, bool $silent = false, bool $seeding = false): void
```
wescopeland: DX: I don't think we should add seeding-specific behavior to production code.

It seems like the intent here is to prevent job dispatches. Instead of passing a `$seeding` flag, we can do something like using Laravel's `Bus::fake()` at the start of the seeder to handle this, eg:
```php
use Illuminate\Support\Facades\Bus;

public function run(): void
{
    Bus::fake(); // prevent job dispatches
```

This keeps the action clean and moves the "don't dispatch stuff" concern to the seeder itself.

database/seeders/AchievementSetClaimSeeder.php
Jamiras: fixed.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
Jamiras: done. I had to use `Queue::fakeFor` as there doesn't appear to be an easy way to reverse `Bus::fake`, and we do want some dispatched events to go through.

PR #4359, opened 2026-01-03
perf: put player_stat_rankings calculations back on a schedule

wescopeland:
The real-time approach we ultimately [switched to](https://github.com/RetroAchievements/RAWeb/pull/4189#discussion_r2596764214) created unforeseen scaling issues in production. Shortly after the production release containing this change, database errors began flooding the prod error logs. I was ultimately required to serialize all jobs through a global lock to prevent InnoDB auto-increment corruption. The corruption occurs because each recalc DELETEs and INSERTs the entire table by design (it's a materialized view), and if multiple jobs are doing this simultaneously then there's ambiguity on where the auto-increment should fall.

We're still seeing deadlocks and slow throughput:

Screenshot 2026-01-03 at 8 16 47 AM
Screenshot 2026-01-03 at 8 17 04 AM

This PR puts the recalc back behind a scheduled update. A 5 minute delay in accurate standings is acceptable for this page.

### DIFF
--- a/app/Community/Actions/GenerateAnnualRecapAction.php
+++ b/app/Community/Actions/GenerateAnnualRecapAction.php
@@ -85,8 +85,8 @@ class GenerateAnnualRecapAction
     {
         $games = PlayerSession::where('user_id', $user->id)
             ->where('duration', '>=', 5)
-            ->where('created_at', '>=', $startDate)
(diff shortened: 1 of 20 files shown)


Jamiras (approved):
Untested. Looks reasonable

PR #4360, opened 2026-01-03
docs: update README.md and CONTRIBUTING.md

wescopeland:
These docs now include more details about the current tech direction and contributing expectations that have calcified into being tribal knowledge.

Also adds a _CODE_FREEZE.md_ file in the V1 API directory. We will no longer accept changes to these endpoints unless they are to fix production issues.


README.md
```diff
@@ -135,6 +131,32 @@ sail pnpm install
...
+Before submitting a pull request, verify your changes pass all checks:
+
+```shell
+# Frontend
+pnpm verify  # Runs linting, TypeScript checks, and Vitest tests
+
+# Backend
+composer fix      # Fix code style issues
+composer analyse  # Run PHPStan static analysis
+composer test -- --parallel  # Run Pest tests
```
drisc: I'd recommend adding how to run specific backend tests as you have with the react components, checking one feature via it's test file rather than running the entire suite is much faster even with `--parallel`.

PR #4361, opened 2026-01-03
feat(api): add V2 JSON:API endpoints for the `Achievement` resource

wescopeland:
This PR adds V2 API endpoints for the `Achievement` resource. There are several nuances with this resource that should be noted:

1. The `isPromoted` filter defaults to `true`. Consumers can optionally set `filter[isPromoted]=false` for unpromoted only, or `filter[isPromoted]=all` to include both.
2. For forward-facing compatibility, the `gameId` filter queries through achievement sets (`achievement_set_achievements` -> `game_achievement_sets`), not through `achievements.game_id`.
3. The `orderColumn` attribute comes from the `achievement_set_achievements` pivot table, not the deprecated `achievements.order_column` field.
4. Achievements for Hub games and Event games are excluded from the index.
5. The `developer` relationship exposes the `User` resource that created the achievement. Deleted developers include `deletedAt` in their response and have `joinedAt` set to null.
6. The `games` relationship uses `staudenmeir/eloquent-has-many-deep` to traverse the double-pivot path (achievement -> achievement_set_achievements -> achievement_sets -> game_achievement_sets -> games).
7. The `Achievement::achievementSets()` relationship has been deleted and a singular `achievementSet` relationship has been added to the `Achievement` model, as achievements can only belong to a single set at a time.

**Other changes**
* Added more TODO comments for future relationship endpoints across the V2 schemas.
* Added a test suite for `SyncAchievementSetOrderColumnsFromDisplayOrdersAction` to verify its behavior, since it now uses `Achievement::achievementSet()`.

**How to test this PR**
```bash
# Get a paginated list of achievements (promoted only, 50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/achievements \

# Get a specific achievement
curl --request GET \
  --url http://localhost:64000/api/v2/achievements/9 \

# Get an achievement with its achievement set, games, and developer included
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievements/9?include=achievementSet,games,developer' \

# Filter achievements by game ID
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievements?filter[gameId]=1' \

# Filter achievements by type
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievements?filter[type]=progression' \

# Get all achievements, including unpromoted
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievements?filter[isPromoted]=all' \

# Sort all achievements by most RetroPoints and include their game
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievements?sort=-pointsWeighted&include=games' \
  -- header 'X-API-Key: {redacted}'
```


app/Api/V2/Achievements/AchievementSchema.php
```diff
@@ -0,0 +1,146 @@
...
+
+            Number::make('unlocksTotal', 'unlocks_total')->sortable()->readOnly(),
+            Number::make('unlocksHardcore', 'unlocks_hardcore')->sortable()->readOnly(),
+            Number::make('unlockPercentage', 'unlock_percentage')->sortable()->readOnly(),
+            Number::make('unlockHardcorePercentage', 'unlock_hardcore_percentage')->sortable()->readOnly(),
+
+            DateTime::make('createdAt', 'created_at')->readOnly(),
+            DateTime::make('modifiedAt', 'modified_at')->readOnly(),
+
+            BelongsTo::make('developer')->type('users')->readOnly(),
```
Jamiras, line 84: Do we really need to dump the entire user record for cases like this? Would it be practical to have two data types for users?
* A user has a displayName and avatarUrl (maybe lastActivityAt and deletedAt)
* A player has the full record including rich presence, points earned, motto, roles, etc.

Or maybe we could inline "developer" (the displayName) as a base attribute of the achievement so the related record wouldn't need to be requested, but could if the additional detail is desirable. I'm not sure if "developerAvatarUrl" would also be desirable as a base attribute.

Maybe I'm just being over-critical. It's just weird to see the developer's recent rich presence information when asking for an achievement.

app/Api/V2/Achievements/AchievementSchema.php
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Get the resource filters.
+     */
+    public function filters(): array
+    {
+        return [
+            WhereIdIn::make($this),
+            Scope::make('isPromoted', 'withPromotedStatus'),
+            Scope::make('gameId', 'forGame'),
+            Where::make('type'),
```
Jamiras: How do I filter on multiple types?
```
"localhost:64000/api/v2/achievements?filter\[gameId\]=2605&filter\[type\]=progression,win_condition"
```
didn't return any

```
"localhost:64000/api/v2/achievements?filter\[gameId\]=2605&filter\[type\]=progression&filter\[type\]=win_condition"
```
only returned win conditions.

app/Api/V2/Achievements/AchievementSchema.php
wescopeland: I agree it's a lot. A couple of things worth noting:
1. The full record only comes back if the consumer explicitly opts in for it with `?include=developer`. Without that, they just get the developer ID in the relationship data.
2. JSON:API really wants consumers to narrow the relationship by using a sparse fieldset: `?include=developer&fields[users]=displayName,avatarUrl`.

We could provide a default sparse fieldset, but then there's no clean way for consumers to say "give me everything" (`fields` doesn't support a wildcard). I think we're probably better off in the long run not setting a default and documenting the pattern instead.

app/Api/V2/Achievements/AchievementSchema.php
wescopeland: This should now work properly in latest.

PR #4364, opened 2026-01-04
fix: change tableFilters to just filters for leaderboards

Chew:
Seems like the Filament v4 upgrade had a lingering issue. Fixes the "Leaderboard Management for [game]" link to properly filter again. Seems that `?tableFilters` changed to just `?filters` right under our noses.


wescopeland (approved):
Thank you!

PR #4370, opened 2026-01-05
(refactor) add GetAchievementSetsAction

Jamiras:
* converts `BuildClientPatchDataV2Action` into  `GetAchievementSetsAction`, which is now an ApiAction directly callable from dorequest.
* eliminates `BuildClientPatchDataAction`. `LegacyGetPatchAction` inherits from `GetAchievementSetsAction` and just translates the response.
* eliminates `InjectPatchClientSupportLevelDataAction` as I presume it only existed to be shared by the two BuildClientPatchData actions. That code now lives in `GetAchievementSetsAction`

With the exception of supporting a null user (removed by #4186), I believe I've migrated all of the `BuildClientPatchDataV2ActionTests` into the `AchievementSetsTests` file. However, I have stuck with the multiple subtests per test paradigm to minimize data repopulation.


tests/Feature/Connect/AchievementSetsTest.php
wescopeland, line 1: I think there may be one scenario missing from this new test suite.

The old test case `testItIncludesLeaderboardData` explicitly verified that leaderboards with `state: LeaderboardState::Disabled` are excluded from the response. This test suite doesn't create any disabled leaderboards or verify they're excluded - we should probably preserve verification of this behavior.

app/Connect/Actions/GetAchievementSetsAction.php
```diff
@@ -14,47 +16,100 @@
...
-        if (!$gameHash && !$game) {
-            throw new InvalidArgumentException('Either gameHash or game must be provided to build achievementsets data.');
+    protected int $gameId;
+    protected ?string $gameHashMd5;
+    protected ?bool $isPromoted;
+    protected ClientSupportLevel $clientSupportLevel;
+
+    public function execute(User $user, int $gameId = 0, ?string $gameHash = null, ?bool $isPromoted = true): array
+    {
+        $this->user = $this->user;
```
wescopeland: Intellectual curiosity: I'm confused by this line. What is happening here?

app/Connect/Actions/LegacyGetPatchAction.php
```diff
@@ -0,0 +1,99 @@
...
+
+        // The legacy API returns the GameID as ParentID for all supported hashes (multiset preparation before creating a new API).
+        // RichPresencePatch may not be returned for unsupported hashes.
+        // So if RichPresencePatch is not being returned, also don't return ParentID.
+        if (array_key_exists('RichPresencePatch', $response)) {
+            $legacyResponse['PatchData']['RichPresencePatch'] = $response['RichPresencePatch'];
+            $legacyResponse['PatchData']['ParentID'] = $response['GameId'];
+        }
+
+        // Propogate any warning that may have been returned.
```
wescopeland: Typo: "Propagate"

app/Connect/Actions/LegacyGetPatchAction.php
```diff
@@ -0,0 +1,99 @@
...
+        // Build the response
+        $response = parent::process();
+
+        // If an error occurred, return it unmodified
+        if ($response['Success'] !== true) {
+            return $response;
+        }
+
+        // We have to relookup the game record for the image_icon_asset_path
+        [$resolvedGameId, $compatibility] = (new VirtualGameIdService())->decodeVirtualGameId($response['GameId']);
```
wescopeland: Perf, nit: `decodeVirtualGameId` is static. We probably don't need to fully instantiate the service here.

app/Connect/Actions/GetAchievementSetsAction.php
Jamiras: Nothing, apparently. :P

It should be assigning the parameter to the field.

The point of the `execute` functions is the ability to run these actions like any other action (from any context within the application). When used via the API, they'll always come in through the `initialize` function, which reads the parameters from the request.

app/Connect/Actions/LegacyGetPatchAction.php
Jamiras: Yep. the whole class is a collection of static methods. Fixed.

app/Connect/Actions/LegacyGetPatchAction.php
Jamiras: Fixed.

tests/Feature/Connect/AchievementSetsTest.php
Jamiras: Added to the tests that look at published achievements. I added an `Unpublished` one as well, but there's no way currently for the API to return it.

PR #4380, opened 2026-01-07
fix exception viewing dev interest page

Jamiras:
fixes 
> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'User' in 'ORDER BY' (Connection: mysql, SQL: select * from `users` where `users`.`id` in (45168, 172083) and `users`.`deleted_at` is null order by `User` asc)

when trying to view dev interest for a game

### DIFF
--- a/app/Platform/Actions/ResetPlayerProgressAction.php
+++ b/app/Platform/Actions/ResetPlayerProgressAction.php
@@ -53,7 +53,7 @@ class ResetPlayerProgressAction
         }
 
         $maintainers = DB::select("
-            SELECT DISTINCT COALESCE(ua.display_name, ua.username) AS Username
+            SELECT DISTINCT COALESCE(ua.display_name, ua.username) AS username
             FROM player_achievements pa
             INNER JOIN achievements ach ON ach.id = pa.achievement_id
             INNER JOIN achievement_maintainers m ON m.achievement_id = ach.id


app/Platform/Actions/BuildGameInterestedDevelopersDataAction.php
```diff
@@ -21,13 +21,12 @@ public function execute(Game $game): Collection
         $users = UserGameListEntry::whereType(UserGameListType::Develop)
             ->where('game_id', $game->id)
             ->with(['user' => function ($query) {
-                $query->orderBy('User');
+                $query->orderBy('username');
```
Jamiras, line 24: this is the actual fix for the exception

app/Platform/Actions/BuildGameInterestedDevelopersDataAction.php
```diff
@@ -21,13 +21,12 @@ public function execute(Game $game): Collection
...
             }])
+            ->whereHas('user.roles', function ($query) {
+                $query->whereIn('name', [Role::DEVELOPER, Role::DEVELOPER_JUNIOR]);
+            })
```
Jamiras, line 28: this avoids an n+1 query problem checking roles after building the list.

app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -842,8 +842,16 @@ private function getInterestedDevelopersCount(Game $game, ?User $user): ?int
...
+        return User::query()
+            ->whereIn('id', function ($query) use ($game) {
+                $query->select('user_id')
+                    ->from('user_game_list_entries')
+                    ->where('game_id', $game->id)
+                    ->where('type', UserGameListType::Develop);
+            })
+            ->whereHas('roles', function ($query) {
+                $query->whereIn('name', [Role::DEVELOPER, Role::DEVELOPER_JUNIOR]);
+            })
```
Jamiras, line 854: This adds the roles filter to counter on the game page. If there's a cleaner way to do this without a subquery, let me know.

http://localhost:64000/game/14134 was reporting three interests, but only showing two when clicked. The third user is no longer a developer.

wescopeland (approved):
Thanks for hunting this stuff down.

PR #4382, opened 2026-01-07
fix(alerts): ignore FFXI

wescopeland:
https://discord.com/channels/476211979464343552/1458253117891149844/1458493450138947678

### DIFF
--- a/app/Helpers/render/ticket.php
+++ b/app/Helpers/render/ticket.php
@@ -26,7 +26,6 @@ function ticketAvatar(
     $ticketStateClass = match ($safeTicket->state) {
         TicketState::Open, TicketState::Request => 'open',
         TicketState::Closed, TicketState::Resolved => 'closed',
-        default => '',
     };
 
     return avatar(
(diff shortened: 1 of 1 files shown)


app/Support/Alerts/Listeners/TriggerSuspiciousBeatTimeAlert.php
```diff
@@ -34,6 +34,12 @@ public function handle(PlayerGameBeaten $event): void
 
         $game = $event->game;
 
+        // Final Fantasy XI supports retroactive unlocks.
+        // We use a regex so we don't match against XII or XIII.
+        if (preg_match('/^Final Fantasy XI($|[:\s\-])/', $game->title)) {
+            return;
+        }
```
Jamiras, line 41: should this just check `getIsStandalone()`?

https://github.com/RetroAchievements/RAWeb/blob/cb28e2671f4eab946da852f9cb1e73d9a3fd8862/app/Models/Game.php#L555-L558

wescopeland: Considered that, but I don't think so. I would guess the majority of future standalones will not have retroactive unlocks in this way, and they are especially prone to cheating.

PR #4383, opened 2026-01-08
feat(manage/achievements): add achievement navigator component to heading

wescopeland:
Based on user feedback. This PR adds a lightweight Alpine.js-powered navigator component to achievement view, edit, and logic pages.

Things to note:
* Only achievements for the currently-viewed mode display. In other words, if you're viewing a promoted achievement, you won't see unpromoted achievements in the list.
* The search is client-side.
* Pagination buttons are always visible, but disabled when appropriate.

https://github.com/user-attachments/assets/8d24a828-8b9d-4cd3-8a79-a7f525031521


resources/views/filament/resources/achievement-resource/partials/achievement-navigator.blade.php
Jamiras, line 1: Editing an achievement and changing it's Promoted state updates the achievement count, but not the achievement list:

Image

wescopeland: Fixed in latest. I've also made a slight UX adjustment: both promoted and unpromoted are now always shown in groups in the selector. The label reflects which group you're currently in. This should make it less cumbersome if someone is promoting/demoting individual achievements while using this navigation control.

Screenshot 2026-01-10 at 2 32 32 PM

app/Filament/Resources/AchievementResource/Concerns/HasAchievementSetNavigation.php
```diff
@@ -0,0 +1,74 @@
...
+     * }|null
+     */
+    public function getAchievementSetNavigationData(): ?array
+    {
+        $current = $this->record;
+        if (!$current instanceof Achievement) {
+            return null;
+        }
+
+        $achievementSet = $current->achievementSets()->first();
```
Jamiras: > Call to undefined method App\Models\Achievement::achievementSets()

trying to view an achievement in filament:
http://localhost:64000/manage/achievements/1/edit

wescopeland: Fixed in latest.

Jamiras (approved):

PR #4391, opened 2026-01-08
fix(alerts): suppress Discord automatic embeds, filter out unranked users

wescopeland:
Discord is automatically popping multiple embeds on these messages. TIL you can use angle brackets to suppress them: https://support.discord.com/hc/en-us/articles/206342858--How-do-I-disable-auto-embed

Additionally, this PR adds some conditional logic to filter out alerts from unranked users.


app/Support/Alerts/SuspiciousBeatTimeAlert.php
```diff
@@ -19,7 +19,7 @@ public function __construct(
     }
 
     /**
-     * "[Scott](https://retroachievements.org/user/Scott) beat [Sonic the Hedgehog](https://retroachievements.org/game/1) in 2m 5s (3.3% of 1h median)"
+     * "[Scott](<https://retroachievements.org/user/Scott>) beat [Sonic the Hedgehog](<https://retroachievements.org/game/1>) in 2m 5s (3.3% of 1h median)"
```
Jamiras, line 22: > TIL you can use angle brackets to suppress them

I use that frequently when I link to code changes (no reason to have the github embed).

I didn't think it was necessary when using the `[]()` syntax.

PR #4394, opened 2026-01-09
fix: purge user avatar from Cloudflare on new avatar upload

wescopeland:
This PR fixes an issue where users may only see their new avatar if they wipe their browser's cache. This is because we cache user avatars in Cloudflare indefinitely, and we don't have any convenient means to cachebust them because the filenames are static.

Now, on avatar upload, an HTTP call is made to Cloudflare to purge the avatar from the cache.

An alternative I considered was adding an `avatar_uploaded_at` column to the `users` table, however after chasing down multiple places in the code we reference `/UserPic/`, this seemed fraught with peril.

I'm not sure how to test this change locally because it's prod specific, explicitly targeting `https://media.retroachievements.org`. I made sure the actual call works by making a curl call to the target endpoint with the given params.


resources/js/features/settings/hooks/useResetNavbarUserPic.ts
Jamiras, line 1: There's still instructions on the profile page to Ctrl+F5. Are they still necessary? It probably doesn't hurt to leave them in.

Image

wescopeland: Probably doesn't hurt to leave it in. I bet some browsers will try to hang on to the asset for as long as possible even with the Cloudflare invalidation happening behind the scenes.

PR #4396, opened 2026-01-09
perf: enable useScriptElementForInitialPage inertia setting

wescopeland:
https://laravel.com/docs/changelog#:~:text=Optimize%20Page%20Data%20Size%20and%20Parsing

> Opt-in support was added for loading the initial Inertia page data from a <script type="application/json"> element instead of the body tag. This results in a greatly reduced initial page data size, faster parsing, and makes the initial page data a breeze to inspect with your browser's dev tools.

Screenshot 2026-01-09 at 5 59 15 PM

### DIFF
--- a/resources/js/app.tsx
+++ b/resources/js/app.tsx
@@ -35,6 +35,12 @@ Sentry.init({
 });
 
 createInertiaApp({
+  defaults: {
+    future: {
+      useScriptElementForInitialPage: true,
+    },
+  },
+
   title: (title) => (title && title !== appName ? `${title} · ${appName}` : appName),
 
   resolve: (name) =>
--- a/resources/js/ssr.tsx
+++ b/resources/js/ssr.tsx
@@ -38,6 +38,12 @@ createServer(
 
       render: ReactDOMServer.renderToString,
 
+      defaults: {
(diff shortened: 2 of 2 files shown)


resources/js/app.tsx
Jamiras, line 1: Is there something I need to do beyond the `.hooks/post-merge` to activate these changes? I also tried `sail down`/`sail up`.

Every page I hit spits out this error and doesn't render content:
```
index.esm.js:138 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'component')
    at tv (index.esm.js:138:34)
    at app.tsx:37:1
```

wescopeland: I only tested with SSR running. I see the same problem when SSR is not running.

Looking over how the Inertia team implemented this, it assumes an SSR daemon is in place to pass off the page props to React. We could fix the setting by doing:
```ts
useScriptElementForInitialPage: !import.meta.env.DEV,
```
But, I think this solution and implementation by the Inertia team is really dangerous. It means in prod if the Inertia daemon ever goes down, the site is inaccessible.

Given this finding, IMO we shouldn't move forward with this change.

PR #4398, opened 2026-01-10
perf(game): resolve N+1 query problem with user leaderboard entries

wescopeland:
Resolves [RAWEB-1MV in Sentry](https://retroachievementsorg.sentry.io/issues/86947313/?environment=production&project=4508694685417552&referrer=issue-list&statsPeriod=1h).

On game pages, leaderboard entries for the authenticated user are being queried in a loop, which can result in hundreds of queries being fired off for things like rhythm games.

Now, a user's leaderboard entries are all fetched in a single large batch query.

**How to test this PR**
Go to a game page where you have multiple leaderboard entries submitted. Switch to the leaderboard view. Your entries and ranks should all still appear correctly.


app/Platform/Actions/BuildGameShowPagePropsAction.php
```diff
@@ -894,4 +887,79 @@ private function getAchievementSetPlayerCounts(int $achievementSetId): array
...
+
+        if (empty($subqueries)) {
+            return [];
+        }
+
+        // Combine all the subqueries with UNION ALL to execute as a single final query.
+        $combinedQuery = array_shift($subqueries);
+        foreach ($subqueries as $subquery) {
+            $combinedQuery = $combinedQuery->unionAll($subquery);
+        }
```
Jamiras, line 947: This feels very kludgy. You're still running N queries, you're just doing it in a single request.

However, due to how the user's rank is calculated, I don't see any better solutions.

wescopeland: I can't dispute that. At the very least, there's still tangible benefits in doing this sometimes-huge union. During my local testing I was able to see a nearly ~1s difference on query time for a user I was testing against.

PR #4399, opened 2026-01-10
perf(CommonPlayersStrategy): resolve N+1 query problem

wescopeland:
Resolves [RAWEB-15B](https://retroachievementsorg.sentry.io/issues/68645240/?environment=production&project=4508694685417552&referrer=issue-list&statsPeriod=1h) and [RAWEB-H3](https://retroachievementsorg.sentry.io/issues/46463870/?environment=production&project=4508694685417552&referrer=issue-list&statsPeriod=1h) in Sentry.

The game suggestion engine selects strategies randomly in a loop. Each time `CommonPlayersStrategy::select()` is called, it executes `getMasterUserIds()` which queries the database. However, for game-specific suggestions, each subsequent call is redundant.

This PR caches the call results at the instance level, so repeated calls to the same strategy instance only query the database once.


app/Platform/Services/GameSuggestions/Strategies/CommonPlayersStrategy.php
```diff
@@ -76,14 +76,14 @@ public function reasonContext(): ?GameSuggestionContextData
...
-            ->withTrashed()
-            ->whereRaw('id % 100 < 5')
-            ->limit(5)
-            ->pluck('user_id');
+        return once(function () {
+            return PlayerGame::where('game_id', $this->sourceGame->id)
+                ->where('user_id', '!=', $this->user->id)
+                ->whereAllAchievementsUnlocked()
+                ->withTrashed()
+                ->whereRaw('id % 100 < 5')
```
Jamiras: I realize this isn't new to this PR, but why are we doing this? A 5% chance of selecting a player_game record really limits the data for sets without hundreds of users. And it could even completely filter out results for a set with thousands of users if things don't line up right.

I don't imagine this filter could be applied until after all the other filters were applied, so it's not going to downsample early. Since there's a limit, wouldn't it make more sense to just use `inRandomOrder()`?

wescopeland: I don't remember why it was built this way. Looking at it today with a fresh set of eyes, I agree with you. I've changed this to `inRandomOrder()` in https://github.com/RetroAchievements/RAWeb/pull/4399/commits/371d1ab1f171efbb94f59e2c025a6b8ae134bf89.

PR #4400, opened 2026-01-10
feat: strengthen password rules, check for compromised passwords

wescopeland:
https://blog.codinghorror.com/password-rules-are-bullshit/

This PR:
* Changes the minimum password length from 8 to 10.
* Enforces that a user's password cannot equal their username or email address.
* Enforces that new passwords cannot have been part of a data breach.
* Adds a basic entropy check for passwords: they must contain at least 5 unique characters.
* Detects compromised passwords on login and alerts the user via a [session variable](https://laravel.com/docs/12.x/session#storing-data).

Screenshot 2026-01-10 at 4 58 55 PM


lang/en_US.json
```diff
@@ -203,7 +203,7 @@
     "Most of the website is still untranslated. If you'd like to help with translations, you can learn more about how to contribute <1>here</1>.": "Most of the website is still untranslated. If you'd like to help with translations, you can learn more about how to contribute <1>here</1>.",
     "Most recent game beaten": "Most recent game beaten",
     "Most recent set mastered": "Most recent set mastered",
-    "Must be at least 8 characters.": "Must be at least 8 characters.",
+    "Must be at least 10 characters.": "Must be at least 10 characters.",
```
Jamiras: I have a feeling we might use this on other fields. Should the value be bindable?

app/Support/Rules/PasswordRules.php
```diff
@@ -0,0 +1,30 @@
...
+            new MinimumUniqueCharacters(5),
+        ];
+
+        if ($requireConfirmation) {
+            $rules[] = 'confirmed';
+        }
+
+        if ($checkAgainstEmail) {
+            $rules[] = 'different:email';
+        }
```
Jamiras: Why is this conditional? It looks like it's only currently validated on creation, but I was quite easily able to change a user's password in this way.

lang/en_US.json
wescopeland: Yea, the same thought occurred to me this morning. I agree. Fixed in latest.

app/Support/Rules/PasswordRules.php
wescopeland: It shouldn't be - this is a bug. Fixed in latest.

lang/en/validation.php
```diff
@@ -100,7 +100,13 @@
     'not_in' => 'The selected :attribute is invalid.',
     'not_regex' => 'The :attribute format is invalid.',
     'numeric' => 'The :attribute must be a number.',
-    'password' => 'The password is incorrect.',
+    'password' => [
+        'letters' => 'The :attribute must contain at least one letter.',
+        'mixed' => 'The :attribute must contain at least one uppercase and one lowercase letter.',
+        'numbers' => 'The :attribute must contain at least one number.',
+        'symbols' => 'The :attribute must contain at least one symbol.',
```
Jamiras, line 107: Are these strings used? I had no issues creating an all lowercase password, or an all numeric password.

Jamiras (approved):

lang/en/validation.php
wescopeland: They're Laravel defaults - if we don't include them then we blow them away. I wasn't sure if that was desirable or not - it may cause a difficult-to-trace bug in the future when something like Fortify gets more involved in the picture.

PR #4402, opened 2026-01-11
perf: cache last_activity_at writes to reduce DB contention

wescopeland:
This PR replaces synchronous DB writes for `users.last_activity_at` with Redis, flushing to the DB every minute. This eliminates a lock contention issue we currently have in the `users` table.

Around 98% of `last_activity_at` writes in prod take <10ms. The other 2%, which amounts to 24,000(!) requests in prod per day, take 500ms - 7s. This is because `users.last_activity_at` values are constantly being written to 24/7, every second of the day. Every single write forces multiple indexes to rebalance. Rows sometimes get locked altogether from concurrent writes. The end result is slow writes to `users.last_activity_at` causing delayed page loads during high-traffic periods. Our current high-traffic periods are very likely to become the new non-peak normal sometime during 2026.

A new `UserLastActivityService` caches activity timestamps in Redis. Direct writes to `users.last_activity_at` are replaced with calls to `UserLastActivityService::touch()` which writes to Redis (<1ms) instead of the DB. A scheduled command runs every minute to flush the cached values to the DB using a single bulk update. This turns thousands of competing single-row updates into one efficient batch operation.

The `User` model's `last_activity_at` accessor reads from Redis first, falling back to the DB-stored value when needed.


app/Platform/Services/UserLastActivityService.php
```diff
@@ -0,0 +1,142 @@
...
+        );
+
+        // Track this user ID as needing to be flushed to DB.
+        // The lock prevents race conditions on the read-modify-write cycle for the pending list.
+        // Actual lock hold time is ~1-2ms. Under extreme load (1000+ req/s), requests that can't
+        // acquire the lock within LOCK_WAIT_SECONDS will silently skip tracking.
+        Cache::lock(self::PENDING_KEY . ':lock', self::LOCK_TTL_SECONDS)->block(self::LOCK_WAIT_SECONDS, function () use ($userId) {
+            $pending = Cache::get(self::PENDING_KEY, []);
+            $pending[$userId] = true;
+            Cache::put(self::PENDING_KEY, $pending, now()->addHours(self::CACHE_TTL_HOURS));
```
Jamiras: Would it be worth avoiding the write back to Cache if the user was already in the list?
```
if (!array_key_exists($userId, $pending)) {
    $pending[$userId] = true;
    Cache::put(...);
}
```

That could also eliminate the need to use an associative array:
```
if (!in_array($userId, $pending)) {
    $pending[] = $userId;
    Cache::put(...);
}
```

app/Platform/Services/UserLastActivityService.php
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Count all users active in the last N minutes.
+     */
+    public function countOnline(int $withinMinutes = 10): int
+    {
+        // In local dev, flush first since the scheduler may not be running.
+        if (app()->isLocal()) {
+            $this->flushToDatabase();
+        }
```
Jamiras, line 76: There's a small window where a user has just become active where they'll be in the Cache, but not yet flushed to disk. As the window is larger than the flush, and the user will likely remain active, I suspect this is okay, but it might be worth noting via a comment.

tests/Feature/Platform/Actions/UnlockPlayerAchievementActionTest.php
```diff
@@ -64,6 +65,9 @@ public function testManualUnlockDoesntUpdateLastActivityAt(): void
         $this->assertNotNull($playerAchievement);
         $this->assertNull($playerAchievement->unlocker_id);
 
+        // flush cached activity timestamps to the DB before checking them
+        app(UserLastActivityService::class)->flushToDatabase();
```
Jamiras, line 69: With the obvious exception of testing the service itself, wouldn't it be easier to disable the functionality in the tests?

app/Platform/Services/UserLastActivityService.php
wescopeland: > Would it be worth avoiding the write back to Cache if the user was already in the list?

Yep, good idea. I've made this change in latest.

> That could also eliminate the need to use an associative array:

I'm more bearish on this, as it causes a subtle perf hit. The proposal is O(n) whereas the current associative array implementation is O(1).

app/Platform/Services/UserLastActivityService.php
wescopeland: Updated the PHPDoc comment for the method in latest.

tests/Feature/Platform/Actions/UnlockPlayerAchievementActionTest.php
wescopeland: I think it actually might be more complicated to disable the functionality in tests. The only ideas I can think of are do a conditional in the service itself to detect if we're in tests, which I'd prefer not to do, or mock the service in test suites, which is much more verbose.

Jamiras (approved):

PR #4403, opened 2026-01-11
chore: create and write to users_online_counts table

wescopeland:
```shell
sail artisan migrate
sail artisan ra:site:user:backfill-online-counts
```

This PR begins a transition from _playersonline.log_ to a `users_online_counts` table in the database. A subsequent PR (which should also go out in a subsequent deploy) will remove reads/writes from the log file altogether.

The table is named `users_online_counts` (plural) instead of `user_online_counts` (singular) because the data isn't associated with any one specific user.

The backfill command spits out the all-time high date and value just as a sanity check for data integrity.

Additionally, the online count is now updated even while the site is in maintenance mode. It will also backfill in the event of an app or DB server outage as soon as service is restored.


app/Console/Commands/BackfillUsersOnlineCounts.php
```diff
@@ -0,0 +1,79 @@
...
+        $firstEntryTime = $lastModified->copy()->subMinutes(self::LOG_INTERVAL_MINUTES * ($lineCount - 1));
+
+        $this->info("Found {$lineCount} entries.");
+        $this->info("Calculated time range: {$firstEntryTime} to {$lastModified}");
+
+        $records = [];
+        $runningMax = 0;
+        $allTimeHighIndex = 0;
+
+        foreach ($lines as $index => $line) {
```
Jamiras, line 53: I don't know how far back we want to backfill this data. There's ~45k lines in the file on the server, which is about 2 1/2 years of backfill data. However, since that data is not timestamped, any line that's missing due to an outage or upgrade will skew all data prior to it.

wescopeland: I think it's probably more valuable to backfill it all and accept that there may be a few hours of drift than to backfill it with a cutoff date. If we had a cutoff date, I wouldn't have been able to make an analysis such as https://discord.com/channels/476211979464343552/1002693810037477377/1459954387932938445.

PR #4404, opened 2026-01-11
chore: switch reads from playersonline.log to users_online_counts

wescopeland:
Follow-up to https://github.com/RetroAchievements/RAWeb/pull/4403. Should be deployed separately.


app/Http/Actions/BuildCurrentlyOnlineDataAction.php
```diff
@@ -6,73 +6,44 @@
...
-                ->all()
-        );
+        return UsersOnlineCount::query()
+            ->orderByDesc('created_at')
+            ->take(48)
+            ->pluck('online_count')
+            ->reverse()
+            ->pad(48, 0)
+            ->values()
+            ->all();
```
Jamiras: This logic is naively assuming the data is being written every 30 minutes. If I delete the last few rows, or even just backdate them, the graph still shows the most recent row as being at "now".

It would be more accurate to fetch all rows in the last 48 hours, forward fill them to now, then backfill to 48 entries.

wescopeland: Updated in latest to use an array with 48 slots. This should produce results that are more accurate.

PR #4408, opened 2026-01-11
feat(events): support comments

wescopeland:
This PR adds comment support to event pages via a new commentable type: `event.comment`.

Screenshot 2026-01-11 at 4 19 29 PM

---

The implementation should be exhaustive:
* The comments are included in daily digest emails.
* These comments have `/comment/{comment}` redirect route support.
* There's an `/event/{event}/comments` page to see the paginated comments list.
* The unsubscribe service supports handling these things.


resources/js/features/events/components/+show/EventShowMainRoot.tsx
```diff
@@ -52,6 +53,8 @@ export const EventShowMainRoot: FC = () => {
       </div>
 
       <EventAchievementSetContainer event={event} />
+
+      <EventCommentList />
```
Jamiras, line 57: Should this widget be hidden on events that have been ended for some period of time (6 months)?

I don't think there would be much point in commenting on an event that [ended in 2019](http://localhost:64000/event/38-retro-clean-up-2019), and if there aren't any other comments on the event, no one would even know.

People can still post in the official forum topic, which leads me to question if we need this functionality at all. When would a user post to the event page instead of the official forum topic? I really don't want to see a wall of "Mastered!" on event pages.

I think this would be better handled by having [per-game/event forums](https://github.com/RetroAchievements/RAWeb/discussions/661#discussioncomment-1127130), though I understand that's a much larger undertaking.

resources/js/features/comments/EventCommentsMainRoot/EventCommentsMainRoot.tsx
```diff
@@ -0,0 +1,70 @@
...
+          paginatedData={paginatedComments}
+        />
+
+        {auth ? (
+          <SubscribeToggleButton
+            subjectId={event.id}
+            subjectType="EventWall"
+            hasExistingSubscription={isSubscribed}
+          />
+        ) : null}
```
Jamiras, line 50: This doesn't seem to be working. The Subscribe button still appears when not logged in.

Image

resources/js/features/events/components/+show/EventShowMainRoot.tsx
wescopeland: This is more just for ephemeral-esque chatter, similar to game pages. I don't think there's any harm in it. What I'm mostly considering is we have a lot of that kind of banter on Discord, but we have plenty of people participating in events that aren't on the Discord. This gives them a social avenue.

Jamiras: There's also a lot of hints on how to do certain things in the discord channels, and that sort of information would be quickly lost if these are used for banter. Having a forum for the event would allow separate topics for each featured game, as well as plenty of places for banter.

wescopeland: Makes sense. It'll be some time before I have the opportunity to improve the forum, but if that's the direction that makes more sense, it'd probably be better to close this PR.

PR #4410, opened 2026-01-13
fix(manage): remediate URL issue on most reported games page

wescopeland:
http://localhost:64000/manage/most-reported-games

Resolves game IDs in URLs being incorrect on this page. This is caused by column ambiguity in the query used to construct the table.


resources/views/livewire/administrative-tools/most-reported-games-table.blade.php
```diff
@@ -94,8 +94,8 @@ private function buildMostTicketedSetsQuery(): Builder
                     $join->on('tickets.ticketable_id', '=', 'newest_tickets.ticketable_id');
                 })
                 ->select(
-                    DB::raw('MIN(tickets.id) as id'),
-                    'games.id as ID',
+                    DB::raw('MIN(tickets.id) as min_ticket_id'),
```
Jamiras: If renaming this didn't require any other changes, it doesn't seem like it's being used and could be eliminated to make the query marginally simpler.

Jamiras (approved):

resources/views/livewire/administrative-tools/most-reported-games-table.blade.php
wescopeland: Yep, removed.

PR #4411, opened 2026-01-13
feat(manage): improve global search UX

wescopeland:
This PR makes significant adjustments to the global search UX in Filament. Notably:

* The global focus hotkey is now visibly displayed.
* The result groups are prioritized using the same logic as the player-facing global search groups.
* More metadata is displayed regarding some results, such as the game attached to an achievement or leaderboard.
* Multi-row secondary information is now truncated to one line.
* The search results box has an increased max-width and max-height.

I explored being able to render badges/icons in the results too, but this turned out to be very difficult so I scrapped that idea.

**Search Box Before**
Screenshot 2026-01-13 at 6 26 50 PM

**Search Box After**
Screenshot 2026-01-13 at 6 26 56 PM

**"super" Results Before**
Screenshot 2026-01-13 at 6 27 07 PM

**"super" Results After**
Screenshot 2026-01-13 at 6 27 19 PM

**Achievement Result Before**
Screenshot 2026-01-13 at 6 27 32 PM

**Achievement Result After**
Screenshot 2026-01-13 at 6 27 42 PM


app/Filament/FilamentServiceProvider.php
```diff
@@ -45,7 +46,9 @@ public function panel(\Filament\Panel $panel): \Filament\Panel
                 'success' => \Filament\Support\Colors\Color::Green,
                 'warning' => \Filament\Support\Colors\Color::Amber,
             ])
+            ->globalSearch(RelevanceBasedGlobalSearchProvider::class)
             ->globalSearchKeyBindings(['command+k', 'ctrl+k'])
+            ->globalSearchFieldKeyBindingSuffix()
```
Jamiras: On Windows, this says "META+K", which isn't really a key.

Image

app/Http/Controllers/Api/SearchApiController.php
```diff
@@ -63,18 +69,22 @@ public function index(Request $request): JsonResponse
...
-            'forum_comments' => fn () => $this->searchForumComments($keyword, $shouldUsePagination ? $page : null, $shouldUsePagination ? $perPage : null),
-            'comments' => fn () => $this->searchComments($keyword, $request->user(), $shouldUsePagination ? $page : null, $shouldUsePagination ? $perPage : null),
+            'users' => fn () => $this->searchUsers($keyword, $paginatedPage, $paginatedPerPage),
+            'games' => fn () => $this->searchGames($keyword, $paginatedPage, $paginatedPerPage),
+            'hubs' => fn () => $this->searchHubs($keyword, $paginatedPage, $paginatedPerPage),
+            'events' => fn () => $this->searchEvents($keyword, $paginatedPage, $paginatedPerPage),
+            'achievements' => fn () => $this->searchAchievements($keyword, $paginatedPage, $paginatedPerPage),
+            'forum_comments' => fn () => $this->searchForumComments($keyword, $paginatedPage, $paginatedPerPage),
+            'comments' => fn () => $this->searchComments($keyword, $request->user(), $paginatedPage, $paginatedPerPage),
```
Jamiras, line 87: > The result groups are prioritized using the same logic as the player-facing global search groups.

If I search on "thirteen", I get "users, leaderboards, achievements", but if I search on "tele", I get "events, games, hubs, leaderboards, users, achievements", and "snow" gives me "games, leaderboards, news, users, hubs, achievements".

Users does seem to always appear before achievements, but leaderboards and hubs don't seem to have a fixed order.

app/Filament/FilamentServiceProvider.php
wescopeland: That's a bummer. This is a bug in Filament - there's a typo in their source code. We could fix it ourselves by passing a function to `globalSearchFieldKeyBindingSuffix`, but it doesn't seem worth it.

I've removed this - it can be revisited once Filament fixes their internal bug.

app/Http/Controllers/Api/SearchApiController.php
wescopeland: Fixed a bug and added some tiebreaker logic here. Hopefully it's a little bit better now.

PR #4413, opened 2026-01-14
feat(manage): add consistent 'View on Site' button

wescopeland:
This PR adds a consistent navigation experience to most highly-touched resources in Filament.

Previously, we had Permalink fields only on the Details page.

Now, those fields have been removed in favor of a "View on Site" button in the header actions area that appears on both Details and Edit.

**Before**
Screenshot 2026-01-13 at 7 00 20 PM

Screenshot 2026-01-13 at 7 00 24 PM

**After**
Screenshot 2026-01-13 at 7 00 30 PM

Screenshot 2026-01-13 at 7 00 34 PM


app/Filament/Actions/ViewOnSiteAction.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->label('View on Site')
+            ->icon('heroicon-m-arrow-top-right-on-square')
+            ->color('gray')
+            ->url(fn (HasPermalink $record): string => $record->getPermalinkAttribute())
+            ->openUrlInNewTab();
```
Jamiras: Does this need to open in a new tab? When I'm ready to return to the site, I don't need the filament page anymore.

I realize it's consistent with the previous permalink behavior, but it seems like this change is meant to remove clicks for users.

Jamiras: Also, if I want it to open in a new tab, I can middle click on the button. There's no way (that I know of) to force it to open in the same tab.

wescopeland: Will no longer open as a new tab.

PR #4422, opened 2026-01-15
fix(alerts): exclude free points hub games

wescopeland:
https://discord.com/channels/476211979464343552/1458253117891149844/1459187927153250464


app/Support/Alerts/Listeners/TriggerSuspiciousBeatTimeAlert.php
```diff
@@ -44,6 +45,11 @@ public function handle(PlayerGameBeaten $event): void
             return;
         }
 
+        // It's not unusual for games in the Free Points hub to be beaten quickly.
+        if ($game->hubs()->where('game_sets.id', GameSet::FreePointsHubId)->exists()) {
+            return;
+        }
```
Jamiras: Consider moving this below the median time check so the query can be avoided for most checks.

Although, this probably doesn't get called enough to worry about that.

Jamiras (approved):

app/Support/Alerts/Listeners/TriggerSuspiciousBeatTimeAlert.php
wescopeland: Done.

PR #4426, opened 2026-01-16
env(horizon): reduce processes used for local horizon

Jamiras:
Reduces the memory footprint of running a local RAWeb instance from 6.9GB to 3.5GB in my VM.

By reducing the number of horizon processes from 14 to 1. Each horizon process was using about 250MB of RAM. The queue will still grow dynamically to 4 processes under heavy load, but this is hugely beneficial when idle - which will be the normal case for most development work.


config/horizon.php
Jamiras, line 1: Easiest to view with whitespace off: https://github.com/RetroAchievements/RAWeb/pull/4426/changes?w=1

config/horizon.php
```diff
@@ -186,176 +186,213 @@
...
-                'default',
-                'developer-metrics',
-                'game-metrics',
-                'player-achievements',
-                'player-beaten-games-stats',
-                'player-game-metrics',
-                'player-metrics',
-                'player-points-stats',
+    'environments' => [
+        'local' => [
```
wescopeland:
```suggestion
        'local' => env('HORIZON_USE_FULL_SUPERVISORS', false) ? [] : [
```
I have plenty of RAM to spare and like running the full prod setup. This would allow me to use an env var so I can keep doing that without needing to track a local change in this file.

config/horizon.php
```diff
@@ -186,176 +186,213 @@
...
+                'maxTime' => 0,
+                'maxJobs' => 0,
+                'memory' => 128,
+                'tries' => 1,
+                'timeout' => 300,
+                'nice' => 0,
+
+                // these settings are unique to the local environment
+                'balance' => 'false', // prevents horizon from spinning up one process per queue
+                'maxProcesses' => 4,
```
wescopeland, line 209:
```suggestion
                'maxProcesses' => (int) env('HORIZON_LOCAL_MAX_PROCESSES', 4),
```
Not sure if this is worth doing, but something to consider.

config/horizon.php
```diff
@@ -186,176 +186,213 @@
...
-            'supervisor-1' => [
-            ],
-        ],
-
-        'local' => [
-            'supervisor-1' => [
-                'maxProcesses' => 3,
+            /**
+             * Daily email supervisor - handles mass email distribution.
+             * Isolated to allow non-time-critial emails to be sent without affecting other queues.
```
wescopeland: Found a typo here, held over from the previous change where we missed it

```suggestion
             * Isolated to allow non-time-critical emails to be sent without affecting other queues.
```

config/horizon.php
Jamiras: This isn't going to work that way. If your variable is true, `local` will be defined as an empty array, which will still prevent it from reaching the '*' path, and horizon will only spin up a single `default` queue.

I can trick it with a little PHP magic, but this feels dirty. What do you think?
```
    'environments' => [
        env('HORIZON_USE_FULL_SUPERVISORS', false) ? 'unused' : 'local' => [
            'supervisor-1' => [ // locally, use single supervisor and fix it at 4 processes
```

config/horizon.php
Jamiras: The only time I hit this limit in my testing was reindexing the search stuff. I suspect other mass-update commands would behave similarly, but under normal use, it rarely even went up to 2.

On that note, I could definitely raise it higher as the normal use would keep it low. The problem was all the idle processes created by the supervisors. Each supervisor had a minimum of one process, and several had fixed sizes. When idle with the default settings, there wer 14 processes just polling the queue.

Alternately, the toggle to use the production config might be enough.

config/horizon.php
Jamiras: Also, I think this should be `HORIZON_LOCAL_USE_FULL_SUPERVISORS` (or something similar to indicate it's only for local development.

wescopeland: It feels dirty but seems pragmatic to me. We could just leave a comment or something:
```php
// When true, key becomes "unused" so your local env falls through to '*' (the prod config).
```

wescopeland (approved):

PR #4432, opened 2026-01-16
fix(GameResource): add DisallowAnimatedImageRule, block animated banners

wescopeland:
Testable with [this image](https://cdn2.steamgriddb.com/hero/0e2a88b513f0734fd19ed6c89a68658c.webp).

I generated the test fixtures using the command line. They're not real images, but they contain real image metadata.

Screenshot 2026-01-16 at 5 44 39 PM


app/Rules/DisallowAnimatedImageRule.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * APNG files contain an 'acTL' (animation control) chunk that static PNGs lack.
+     *
+     * @see https://wiki.mozilla.org/APNG_Specification#.60acTL.60:_The_Animation_Control_Chunk
+     */
+    private function isAnimatedPng(string $content): bool
+    {
+        return str_contains($content, 'acTL');
```
Jamiras, line 51: I don't really have a better solution, but the chances of this occuring in the compressed image data is non-zero.

PR #4439, opened 2026-01-17
fix(AchievementAuthorPolicy): let developers and maintainers delete credit on their achievements

wescopeland:
https://discord.com/channels/310192285306454017/310195377993416714/1461892190560321546

An achievement author/maintainer should be able to delete credit on the achievements they've authored/currently maintain.

### DIFF
--- a/app/Providers/FortifyServiceProvider.php
+++ b/app/Providers/FortifyServiceProvider.php
@@ -314,11 +314,15 @@ class FortifyServiceProvider extends ServiceProvider
      */
     private function checkForCompromisedPassword(string $password): void
     {
+        if (app()->environment('local')) {
+            return;
+        }
+
         $verifier = app(UncompromisedVerifier::class);
 
         $isSafe = $verifier->verify([
             'value' => $password,
-            'threshold' => 0,
+            'threshold' => 5,
         ]);
 
         if (!$isSafe) {


app/Policies/AchievementAuthorPolicy.php
```diff
@@ -55,6 +55,16 @@ public function delete(User $user, AchievementAuthor $achievementAuthor): bool
             return true;
         }
 
+        // Achievement developers and maintainers can delete credit on their achievements.
```
Jamiras, line 58: I thought this was intentional. Only members with specific team roles can delete stuff.

Should Jr. Devs be excluded from this priveledge?

Should Devs also be allowed to restore deleted credit on their achievements? Is that even possible currently? I didn't see a button anywhere - but there's code below this function.

Jamiras: Also, in regards to the originally linked issue: Shouldn't that already be handled by case on line 54? If he credited himself, he should be able to delete that same credit.

wescopeland: The scenario I'd like to handle is let's say a developer mistakenly credits someone for the wrong thing. Without this change, they have no mechanism to fix the mistake other than going to someone who has one of those specific team roles - kinda cumbersome.

Jr. Devs should be excluded from this - I've made that change in latest.

The restore function is misleading. For credit, it's more complexity than I'd like to support. It should just be re-added if it's accidentally deleted. I've set `restore()` to return `false`.

PR #4457, opened 2026-01-19
feat: use a new Trending Games algorithm

wescopeland:
This PR replaces the current popularity-based trending algorithm with a true trending algorithm that surfaces games experiencing unusual activity spikes. The current behavior is retained as a "Popular" tab in the same area on the home page with a toggle UI.

Screenshot 2026-01-19 at 2 36 15 PM

---

**Why?**
The current "Trending" section just shows games with the most active players. This means consistently popular games like Final Fantasy XI dominate the list in perpetuity. That's "popular", not "trending".

True trending represents a _rate of change_. These are games experiencing unusual activity relative to their ordinary baseline. A game with 400 new players when it normally gets 25/day (16x spike) is genuinely trending. A game with 500 active players when it always has 500 players is just popular.

**How to test this PR**
```
sail artisan migrate

# note - using a date flag inserts fake games for popularity
# without a date flag, real data is used with no functional changes to how we determine what's popular
sail artisan ra:update-game-activity-snapshots --date=2025-12-23
```

**Algorithm Details**
```
* Core Formula:
*   trend = (new_players_48h + k) * 7 / (new_players_baseline_14d + k)
*   score = trend * sqrt(new_players_48h) * evergreen_penalty
*
* Parameters:
* - k=10 Bayesian smoothing: Prevents small-sample noise by adding a prior to both numerator
*   and denominator. For small samples (baseline=2, recent=6), the ratio shifts from 21x to
*   9x. For large samples (baseline=174, recent=407), k barely matters: 16.4x becomes 15.9x.
* - Minimum 10 new players in 48h: Filters out statistical noise from tiny games.
* - 1.5x trend threshold: Must be 50% above baseline to qualify as "trending."
* - Evergreen penalty: Games with huge player bases get dampened because spikes are
*   less noteworthy (>30k: 0.3x, >15k: 0.5x, >5k: 0.75x).
```

I landed on this algorithm after backtesting numerous alternative algorithms. In non-statistician terms, here's how it works:
1. We compare each game's new players in the last 48 hours vs the prior 14 days.
2. We calculate how much higher than normal that is (aka: the "trend multiplier").
3. We require at least 1.5x normal activity and 10+ new players so as to filter out noise.
4. We then add a dampening filter to games with already-large player counts.
5. Finally, we rank by a score that balances the spike size with actual player volume.

Additionally, we store a reason why the game is trending:
1. New set completed 2-14 days ago: "New set driving interest"
6. Revision completed within the last 14 days: "Recently revised set"
7. Total players < 500 with an activity spike: "Rising in popularity"
8. Total players > 20,000 with an activity spike: "Classic seeing renewed interest"
9. Trend multiplier >= 10: "Many more players than usual"
10. Default fallback: "More players than usual"

Trending and popular games are recalculated every 15 minutes and stored in a new `game_activity_snapshots` table. This has the added benefit of eliminating our need for a `Cache::flexible()` on the home page. The model uses `MassPrunable`:
```php
public function prunable(): Builder
{
    // TODO eventually lower this to 7 days. it's 30 right now because i want to analyze trends in prod
    // for more potential improvements / fix anything that potentially goes wrong.
    return $this->where('created_at', '<', now()->subDays(30));
}
```

With this SQL query, you can test the new algorithm's live results in the prod DB. The query takes approximately 3s.

Screenshot 2026-01-19 at 2 51 48 PM

```sql
SELECT
    metrics.game_id,
    g.title,
    g.system_id,
    g.players_total,
    metrics.new_recent,
    metrics.new_baseline,
    ROUND((metrics.new_recent + 10) * 7.0 / (metrics.new_baseline + 10), 2) as trend,
    ROUND(
        ((metrics.new_recent + 10) * 7.0 / (metrics.new_baseline + 10))
        * SQRT(metrics.new_recent)
        * CASE
            WHEN g.players_total > 30000 THEN 0.3
            WHEN g.players_total > 15000 THEN 0.5
            WHEN g.players_total > 5000 THEN 0.75
            ELSE 1.0
          END,
        2
    ) as score,
    (SELECT MAX(claims.finished_at)
      FROM achievement_set_claims claims
      WHERE claims.game_id = metrics.game_id
        AND claims.status = 'complete'
        AND claims.set_type = 'new_set'
        AND claims.finished_at BETWEEN NOW() - INTERVAL 14 DAY AND NOW() - INTERVAL 48 HOUR
    ) as new_set_date,
    (SELECT MAX(claims.finished_at)
      FROM achievement_set_claims claims
      WHERE claims.game_id = metrics.game_id
        AND claims.status = 'complete'
        AND claims.set_type = 'revision'
        AND claims.finished_at >= NOW() - INTERVAL 14 DAY
    ) as revision_date
FROM (
    SELECT
        game_id,
        SUM(CASE WHEN created_at >= NOW() - INTERVAL 48 HOUR THEN 1 ELSE 0 END) as new_recent,
        SUM(CASE WHEN created_at >= NOW() - INTERVAL 16 DAY
                  AND created_at < NOW() - INTERVAL 48 HOUR THEN 1 ELSE 0 END) as new_baseline
    FROM player_games
    WHERE created_at >= NOW() - INTERVAL 16 DAY
      AND created_at <= NOW()
    GROUP BY game_id
    HAVING new_recent >= 10
) metrics
JOIN games g ON g.id = metrics.game_id
WHERE g.system_id != 101
  AND g.title NOT LIKE '%[Subset%'
  AND g.deleted_at IS NULL
  AND g.achievements_published > 0
  AND EXISTS (
      SELECT 1 FROM achievement_set_claims claims
      WHERE claims.game_id = metrics.game_id
        AND claims.status = 'complete'
        AND claims.set_type = 'new_set'
        AND claims.finished_at <= NOW() - INTERVAL 48 HOUR
  )
  AND NOT EXISTS (
      SELECT 1 FROM game_set_games gsg
      JOIN game_sets gs ON gs.id = gsg.game_set_id
      WHERE gsg.game_id = metrics.game_id
        AND gs.has_mature_content = 1
  )
HAVING trend >= 1.5
ORDER BY score DESC
LIMIT 4;
```


app/Console/Commands/UpdateGameActivitySnapshots.php
```diff
@@ -0,0 +1,284 @@
...
+
+    private function updatePopularGames(): void
+    {
+        $this->info('Calculating popular games...');
+
+        $gameCounts = $this->getCurrentPopularGames();
+
+        arsort($gameCounts);
+
+        $topGames = array_slice($gameCounts, 0, 10, true);
```
Jamiras: Is there some reason to capture 10 if only 4 are displayed? I can see capturing 5 or 6 in case something gets filtered for being mature (shouldn't that happen here anyway unless you're planning to tie the filtering to the user preferrences?)

wescopeland: Capped this at 6 and lifted the mature content filtering up into the command.

Jamiras: If there's no longer any post-processing, is there any reason to store more than 4 for each capture?

PR #4459, opened 2026-01-19
fix(multiset): track subset playtime correctly

wescopeland:
```php
$resolvedSets = (new ResolveAchievementSetsAction())->execute($gameHash, $playerGame->user);
foreach ($resolvedSets as $resolvedSet) {
    $activeAchievementSets[] = $resolvedSet->id;  // this is GameAchievementSet.id
}
```
But then:
```php
->whereIn('achievement_set_id', $activeAchievementSets)  // this expects AchievementSet.id
```
Every multiset player has incorrect `time_taken` on their bonus/specialty sets. Core sets work by coincidence when `GameAchievementSet.id` happens to equal `achievement_set_id`.

There are about 30,000-40,000 records on prod that need to be fixed as a result of this bug. This PR includes a command we can run on the app server to do that:
```bash
# this will kick off about 40,000 jobs!
php artisan ra:platform:recalculate-affected-player-achievement-set-metrics

# manually spin up workers, do this in ~10-16 terminal sessions
php artisan queue:work --queue=player-game-metrics

# finally, fix denormalized aggregate metrics
php artisan ra:platform:game:update-beaten-metrics "$(php artisan tinker --execute="echo \App\Models\GameAchievementSet::whereIn('type', ['bonus', 'specialty'])->pluck('game_id')->unique()->implode(',')")"
```

### DIFF
--- a/app/Platform/Actions/UpdatePlayerGameMetricsAction.php
+++ b/app/Platform/Actions/UpdatePlayerGameMetricsAction.php
@@ -30,6 +30,7 @@ class UpdatePlayerGameMetricsAction
             return;
         }
 
+        // get total time spent playing the game (and its subsets)
         $activityService = new PlayerGameActivityService();
         $activityService->initialize($playerGame->user, $playerGame->game, withSubsets: true);
 
--- a/app/Platform/Services/PlayerGameActivityService.php
+++ b/app/Platform/Services/PlayerGameActivityService.php
@@ -18,6 +18,7 @@ use App\Platform\Enums\AchievementSetType;
 use App\Platform\Enums\PlayerProgressResetType;
 use App\Platform\Enums\UnlockMode;
 use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Collection;
 
 class PlayerGameActivityService
 {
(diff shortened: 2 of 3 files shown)


app/Platform/Commands/RecalculateAffectedPlayerAchievementSetMetrics.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @param array<int, int> $nonCoreSetIds
+     * @return Builder<PlayerAchievementSet>
+     */
+    private function buildAffectedRecordsQuery(array $nonCoreSetIds): Builder
+    {
+        return PlayerAchievementSet::query()
+            ->whereIn('player_achievement_sets.achievement_set_id', $nonCoreSetIds)
+            ->where('player_achievement_sets.updated_at', '>=', self::MULTISET_ENABLED_DATE)
```
Jamiras: If I go back `2025-05-21`, I find another 2000 affected records. That corresponds with deployment of the code that is being fixed.

wescopeland: I see! In my original run, I typo'd the year to 2024, so I didn't catch this. Changed it in latest to `2025-05-01`.

PR #4460, opened 2026-01-19
feat(api): add V2 JSON:API relationship endpoint for `LeaderboardEntry` resources

wescopeland:
This PR adds a V2 API relationship endpoint for `LeaderboardEntry` resources, accessible via `GET /leaderboards/{leaderboard}/entries`. Things worth noting:

1. Entries are accessed exclusively via the relationship route. There is no standalone leaderboard entries index, as entries only make sense in the context of a specific leaderboard.
2. Rank is computed using `RANK()` window functions, correctly handling ties where players with the same score share the same rank.
3. Entries from hidden leaderboards are excluded.
4. Results are default sorted by rank asc.
5. The `user` relationship can be included via `?include=user`.
6. The `filter[user]` param accepts a ULID, username, or display name to filter entries for a specific user.

**How to test this PR**
```bash
# Get paginated entries for a leaderboard (50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/leaderboards/104370/entries \

# Get entries with user data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/leaderboards/104370/entries?include=user' \

# Filter entries by display name
curl --request GET \
  --url 'http://localhost:64000/api/v2/leaderboards/104370/entries?filter[user]=marc720' \

# Filter entries by ULID
curl --request GET \
  --url 'http://localhost:64000/api/v2/leaderboards/104370/entries?filter[user]=01FFG60MCHZ5NFBKRDSDRPAVY8' \

# Custom pagination (page 2, 25 per page)
curl --request GET \
  --url 'http://localhost:64000/api/v2/leaderboards/104370/entries?page[number]=2&page[size]=25' \
```

### DIFF
--- a/app/Community/Actions/SendDailyDigestAction.php
+++ b/app/Community/Actions/SendDailyDigestAction.php
@@ -261,6 +261,7 @@ class CommentDelayedSubscriptionHandler extends BaseDelayedSubscriptionHandler
                 ->where('commentable_id', $delayedSubscription->subject_id)
                 ->where('id', '>', $delayedSubscription->first_update_id)
                 ->where('user_id', '!=', $delayedSubscription->user_id)
+                ->where('user_id', '!=', Comment::SYSTEM_USER_ID)
                 ->orderBy('id')
                 ->first();
 
@@ -277,6 +278,7 @@ class CommentDelayedSubscriptionHandler extends BaseDelayedSubscriptionHandler
             ->where('commentable_id', $delayedSubscription->subject_id)
             ->where('id', '>=', $delayedSubscription->first_update_id)
             ->where('user_id', '!=', $delayedSubscription->user_id)
+            ->where('user_id', '!=', Comment::SYSTEM_USER_ID)
             ->count();
     }
 
--- a/app/Connect/Actions/GetGamesListAction.php
+++ b/app/Connect/Actions/GetGamesListAction.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Connect\Actions;
+
+use App\Connect\Support\BaseApiAction;
+use App\Models\Game;
+use App\Models\System;
+use Illuminate\Http\Request;
+use stdClass;
+
+class GetGamesListAction extends BaseApiAction
+{
+    protected int $consoleId;
+
+    public function execute(?int $consoleId): array
+    {
+        $this->consoleId = $consoleId ?? 0;
+
+        return $this->process();
+    }
+
+    protected function initialize(Request $request): ?array
+    {
+        if (!request()->has('c')) {
+            return $this->missingParameters();
+        }
+
+        $this->consoleId = request()->integer('c', 0);
+
+        return null;
+    }
+
+    protected function process(): array
+    {
+        if (!System::where('id', $this->consoleId)->exists()) {
+            return $this->resourceNotFound('system');
+        }
+
+        $games = Game::query()
+            ->where('system_id', $this->consoleId)
+            ->orderBy('sort_title')
+            ->select(['id', 'title'])
+            ->pluck('title', 'id') // return mapping of id => title
+            ->toArray();
+
+        if (empty($games)) {
+            // replace empty array with empty object so json_encode sends {} to client
+            $games = new stdClass();
+        }
+
+        return [
+            'Success' => true,
+            'Response' => $games,
+        ];
+    }
+}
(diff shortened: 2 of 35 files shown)


app/Api/V2/LeaderboardEntries/LeaderboardEntryResource.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Get the resource's links.
+     *
+     * @param Request|null $request
+     */
+    public function links($request): Links
+    {
+        return new Links(
+            $this->selfLink(),
+        );
```
Jamiras: If these aren't a standalone entry, why do they have a self link?
```
         "links":{
            "self":"http:\/\/localhost:64000\/api\/v2\/leaderboard-entries\/3876778"
         }
```

app/Api/V2/LeaderboardEntries/LeaderboardEntryResource.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param Request|null $request
+     */
+    public function attributes($request): iterable
+    {
+        $leaderboard = $this->resource->leaderboard;
+
+        return [
+            'score' => $this->resource->score,
+            'formattedScore' => ValueFormat::format($this->resource->score, $leaderboard->format),
+            'rank' => $this->resource->rank,
```
Jamiras, line 30: Rank appears to be the index within the results, not the player's actual rank.
```
curl --header 'X-API-Key: [redacted]' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/leaderboards/22481/entries?include=user&filter\[user\]=01BJY8JSBTA7NVT0NYSVDCK5TY" && echo
```
```
        "attributes":{
            "score":592,
            "formattedScore":"0:09.86",
            "rank":1,
            "createdAt":"2021-05-20T17:06:30.000000Z",
            "updatedAt":"2021-05-20T17:09:02.000000Z"
         },
```

Image

app/Api/V2/LeaderboardEntries/LeaderboardEntryResource.php
wescopeland: Good point - went ahead and removed this.

app/Api/V2/LeaderboardEntries/LeaderboardEntryResource.php
wescopeland: Fixed in latest with a rank subquery and added a test to cover this behavior.

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
```diff
@@ -0,0 +1,147 @@
...
+        $isFromExcludedSystem = in_array($leaderboard->game->system_id, [System::Hubs, System::Events], true);
+
+        if ($isHiddenLeaderboard || $isFromExcludedSystem) {
+            $query->whereRaw('1 = 0');
+
+            return $query;
+        }
+
+        $rankSubquery = $this->buildRankSubquery($leaderboard);
+        $query->selectRaw("*, ({$rankSubquery}) as `rank`");
```
Jamiras: It seems very inefficient to run that subquery for every record returned. You should only need to run it once to get the rank of the first entry, then you can calculate the rank of the remaining entries as either the last rank (if score doesn't change), or offset (if score does change).

See logic [here](https://github.com/RetroAchievements/RAWeb/blob/master/app/Platform/Actions/GetRankedLeaderboardEntriesAction.php). Also consider just using the `sortedEntries` helper as seen in that code. I'd recommend using the code piecemeal, but you don't need the `with('user')` clause.

app/Api/V2/LeaderboardEntries/LeaderboardEntryResource.php
Jamiras: I'm still seeing this in the response.
```
"links":{"self":"http:\/\/localhost:64000\/api\/v2\/leaderboard-entries\/576359"}}]
```

wescopeland: Fixed

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
wescopeland: Updated in latest using a `readRelatedEntries` relationship hook. [docs](https://laraveljsonapi.io/5.x/routing/controllers.html#relationship-hooks)

PR #4464, opened 2026-01-20
perf(stats): only recalculate total beats/masteries every four hours

Jamiras:
https://discord.com/channels/476211979464343552/1026595325038833725/1462551723460661411

* last_game_beaten/mastered is updated immediately when the user receives the award (it already was).
* num_beaten/mastered re-synced every four hours to account for resets and users being unranked
* when num_beaten/mastered are re-synced, we join to the `unranked_users` table instead of filtering on the `users` join. This makes the re-sync process about 60% faster (it's also another 20% faster for not having to recalculate the last beaten/mastered data).

We currently award roughly 1000 badges per day - roughly one every minute and a half. As such, while the re-syncing may cause the counts to go down (or up), it's unlikely that a user will even notice.

### DIFF
--- a/app/Connect/Actions/ResolveAchievementSetsAction.php
+++ b/app/Connect/Actions/ResolveAchievementSetsAction.php
@@ -198,13 +198,16 @@ class ResolveAchievementSetsAction
     private function sortSets(Collection $sets): Collection
     {
         return $sets->sortBy(function (GameAchievementSet $set) {
-            return match ($set->type) {
+            $typePriority = match ($set->type) {
                 AchievementSetType::Exclusive => 0,
                 AchievementSetType::Core => 1,
                 AchievementSetType::Specialty => 2,
                 AchievementSetType::Bonus => 3,
                 default => 4,
             };
+
+            // Sort primarily by type priority, then by order_column.
+            return ($typePriority * 100) + ($set->order_column ?? 0);
         });
     }
 
--- a/database/migrations/2026_01_19_000000_update_game_achievement_sets_table.php
+++ b/database/migrations/2026_01_19_000000_update_game_achievement_sets_table.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+use App\Platform\Enums\AchievementSetType;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::table('game_achievement_sets', function (Blueprint $table) {
+            $table->string('type')->default(AchievementSetType::Core->value)->change();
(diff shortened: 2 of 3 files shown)


app/Platform/Actions/RevalidateAchievementSetBadgeEligibilityAction.php
```diff
@@ -162,8 +170,15 @@ private function revalidateCompletionBadgeEligibility(PlayerGame $playerGame): b
...
+            $lastMasteredAt = StaticData::first()?->value('last_game_hardcore_mastered_at');
+            if (!$lastMasteredAt || $playerGame->completed_hardcore_at->gte($lastMasteredAt)) {
+                StaticData::query()->increment('num_hardcore_mastery_awards', 1, [
+                    'last_game_hardcore_mastered_game_id' => $playerGame->game->id,
+                    'last_game_hardcore_mastered_user_id' => $playerGame->user->id,
+                    'last_game_hardcore_mastered_at' => $playerGame->completed_hardcore_at,
+                ]);
+            } else {
+                StaticData::query()->increment('num_hardcore_mastery_awards');
             }
```
wescopeland: Should we also check for `$playerGame->user->is_unranked` here before doing any mutations to static data?

app/Platform/Actions/RevalidateAchievementSetBadgeEligibilityAction.php
```diff
@@ -88,8 +89,15 @@ private function revalidateBeatenBadgeEligibility(PlayerGame $playerGame): bool
             PlayerBadgeAwarded::dispatch($badge);
             PlayerGameBeaten::dispatch($playerGame->user, $playerGame->game, true);
 
-            if ($playerGame->beaten_hardcore_at->gte(Carbon::now()->subMinutes(10))) {
-                static_addnewhardcoregamebeaten($playerGame->game->id, $playerGame->user->username);
+            $lastBeatenAt = StaticData::first()?->value('last_game_hardcore_beaten_at');
```
wescopeland: I'm surprised this works - it kicks off two SQL queries (`value()` is a query builder method).

Image

This will execute a single query:
```php
$lastBeatenAt = StaticData::query()->value('last_game_hardcore_beaten_at');
```

app/Platform/Actions/RevalidateAchievementSetBadgeEligibilityAction.php
Jamiras: good idea.

app/Platform/Actions/RevalidateAchievementSetBadgeEligibilityAction.php
Jamiras: Thanks. It's awkward working with data in a table that's a single row. `StaticData::increment` works too, but phpstan told me I was calling a non-static function :P

PR #4465, opened 2026-01-20
perf: support GET calls on card.php with query params

wescopeland:
_card.php_ is currently called ~250,000 times per day. However, most of the calls do not contain dynamic content. Because the call uses POST, it can't be cached by Cloudflare.

This PR switches it to GET with query params (POST is still supported for backwards compatibility on the deploy). We also signal caching behavior with a `Cache-Control` header.

Using GET requires a `X-Requested-With` header to prevent casual scraping.


app/Providers/RouteServiceProvider.php
```diff
@@ -46,7 +46,10 @@ public function map(): void
     protected function mapWebRoutes(): void
     {
         Route::middleware(['web'])->group(function () {
-            // prohibit GET form requests in request/
+            // Allow GET requests to card.php for CDN caching of tooltip content.
```
Jamiras, line 49: How does the CDN caching work? I expect the user card to be the least static data supported. If the user is active, their points will routinely change. If they're not active, their last activity info will routinely change as it transitions from "X seconds" to "Y seconds" to "Z minutes", etc.

public/request/card.php
```diff
@@ -4,21 +4,37 @@
...
-        'ticket' => renderTicketCard((int) $input['id']),
-        'user' => renderUserCard($input['id']),
-        default => '?',
-    },
-]);
+$html = match ($type) {
+    'achievement' => renderAchievementCard($input['id'], $context),
+    'game' => renderGameCard((int) $input['id'], $context),
+    'hub' => renderHubCard((int) $input['id']),
+    'ticket' => renderTicketCard((int) $input['id']),
```
Jamiras, line 24: Ticket cards appear to be broken (not because of this PR): http://localhost:64000/forums/topic/33579?comment=348204#348204

>   "exception": "[object] (Illuminate\\Database\\Eloquent\\MissingAttributeException(code: 0): The attribute [badgeName] either does not exist or was not retrieved for model [App\\Models\\Achievement]. at /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php:515)
[stacktrace]
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php(495): Illuminate\\Database\\Eloquent\\Model->throwMissingAttributeExceptionIfApplicable()
#1 /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(2423): Illuminate\\Database\\Eloquent\\Model->getAttribute()
#2 /var/www/html/app/Helpers/render/ticket.php(61): Illuminate\\Database\\Eloquent\\Model->__get()
#3 /var/www/html/public/request/card.php(24): renderTicketCard()

app/Providers/RouteServiceProvider.php
wescopeland: We'll make no attempt to cache the user cards:
```php
$isUserSpecific = $type === 'user' || $context !== null;
$isCacheable = request()->isMethod('GET') && !$isUserSpecific;

$cacheControl = $isCacheable
    ? 'public, max-age=300' // 5 minutes
    : 'private, no-cache';
```

public/request/card.php
wescopeland: Fixed in latest.

PR #4474, opened 2026-01-22
feat: introduce RetroPoints 2.0

wescopeland:
This PR implements a new RetroPoints formula that addresses several issues with the current algorithm.

## The Problem
The current RetroPoints formula rewards "rare" achievements. If only 10% of players unlock some achievement, it's worth more RetroPoints than one which 90% of players have unlocked. This works well for games with normal player distributions, but breaks down in several scenarios:

* **Obscure/difficult games are undervalued.** When a game is unpopular or infamously hard, our data suggests that only hardcore fans who already know the game tend to attempt it. This makes achievements look "easy" (through relatively high unlock rates) even when/if they're genuinely difficult. _Casual players never showed up to fail._
* **Popular game subsets have uncapped inflation.** A subset of Super Mario World with 40,000+ players and an achievement with a single unlock creates extreme multipliers (40,000x) that tends to distort leaderboards and user perceptions.

### ELI5 Version

Imagine you're grading a test. If 8 out of 10 students pass, you're not sure if the test is easy or if you just happened to randomly get a bunch of smart students. But if 8,000 out of 10,000 students pass, you can be fairly confident it's a pretty easy test.

The new formula gives a "benefit of the doubt" bonus to games with few players. This bonus shrinks as more players attempt the game, because our confidence in the RetroPoints values increases with the more data we have.

If an achievement is so rare that fewer than 1 in 1,000 ranked players have it, we start applying diminishing returns. Going from 1-in-1000 to 1-in-10000 still is more rewarding, but it's not 10x more rewarding.

### Statistician Version

The existing RetroPoints calculation is:
```
RP = p × r + p × (t/a) × (1 - r)
where: p = base points, a = achievers, t = total players, r = 0.6
```

The new formula introduces two multipliers to the existing calculation.

```
RP = p × r + p × T_A_EFFECTIVE × (1 - r) × m
```

`m` is a low-player multiplier. It is calculated by:
```
T_RATIO = (T_TOTAL × 0.01) / 9
m = MAX(1, 1 / log₁₀(1 + t/T_RATIO))
```

This is a classic Bayesian-style sample correction. The multiplier:
- Equals ~103x at 1 player
- Equals ~11x at 10 players
- Equals ~2.2x at 100 players
- Stabilizes at 1x at ~1% of all ranked users (~1,200 players as of this writing)

Screenshot 2026-01-21 at 8 13 35 PM

`T_A_EFFECTIVE` is an ultra-rare dampener. It is calculated by:
```
T_A_THRESHOLD = T_TOTAL × 0.002  (0.2% of ranked users)
T_A_EFFECTIVE =
  • t/a                                          when t/a ≤ threshold
  • threshold × (1 + log₁₀((t/a) / threshold))   when t/a > threshold
```

This applies log compression to extreme rarity rates. An achievement with 1 unlock out of 40,000 possible players get valued closer to 10,000x rather than 40,000x.

Screenshot 2026-01-21 at 8 15 31 PM

---

**How to test this PR**
```
sail artisan horizon

# recalculate stuff for one game
sail artisan ra:platform:achievement:recalculate-weighted-points --game=1

# recalculate stuff for all games (takes ~10s)
sail artisan ra:platform:achievement:recalculate-weighted-points

# dispatches ~800 jobs, may take an hour or two unless you use 10+ processes like prod
sail artisan ra:platform:player:crawl-weighted-points
```

Note that `CrawlPlayerWeightedPoints` is now **significantly** more efficient. It can now cycle the entire userbase in _less than ten minutes_. Thus, rather than progressively crawling the playerbase 24/7, it'll now do burst corrections once per week.


app/Platform/Actions/CalculateAchievementWeightedPointsAction.php
```diff
@@ -0,0 +1,108 @@
...
+     * The low-player boost stabilizes to 1.0 when a game reaches this fraction of total ranked
+     * players. This makes the threshold dynamic - it scales with site growth rather than being
+     * a fixed number. A game with 1% of the site's ranked players is considered to have enough
+     * data for its rarity values to be trustworthy without adjustment.
+     */
+    public const STABLE_PLAYER_RATIO = 0.01;
+
+    /**
+     * The ultra-rare dampener activates when rarity exceeds this fraction of total players.
+     * At 0.2% (~1 in 500 unlock rate), the achievement is considered "ultra-rare" and the
```
Lafungo: I'm not sure what the intended unlock rate threshold is, but there's a mismatch here between the values cited. At ~125,000 ranked players currently, 0.2% represents a ~1 in 250 unlock rate rather than a ~1 in 500 unlock rate.

wescopeland: Yep, this comment is mixing up two different concepts. Fixed.

PR #4479, opened 2026-01-23
fix(PlayerGameActivityService): adjust reset handling

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1463460521817805000/1463460521817805000.

```
php artisan ra:platform:recalculate-multiset-game-metrics-for-resets
```

**Before**
Screenshot 2026-01-22 at 7 34 35 PM

**After**
Screenshot 2026-01-22 at 7 48 59 PM

**Root Cause**
The service is storing a single global reset date across all related "games". If a user reset one bonus subset game, that reset date was incorrectly used as the start time when calculating the core game's time to mastery.


app/Platform/Commands/RecalculateParentGameMetricsForSubsetResets.php
```diff
@@ -0,0 +1,78 @@
...
+use App\Models\PlayerGame;
+use App\Models\PlayerProgressReset;
+use App\Platform\Enums\AchievementSetType;
+use App\Platform\Enums\PlayerProgressResetType;
+use App\Platform\Jobs\UpdatePlayerGameMetricsJob;
+use Illuminate\Console\Command;
+
+class RecalculateParentGameMetricsForSubsetResets extends Command
+{
+    protected $signature = 'ra:platform:recalculate-parent-game-metrics-for-subset-resets';
```
Jamiras: Do we also need to do the inverse? i.e. recalculate subset metrics if the user has only reset the parent game?

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -537,13 +514,26 @@ public function getAchievementSetMetrics(AchievementSet $achievementSet): array
...
+                            );
+                    })
+                    ->orderByDesc('created_at')
+                    ->first();
 
-            $playerGame = $this->playerGames->where('achievement_set_id', $achievementSet->id)->first();
+                if ($gameReset?->created_at?->gt($startTime)) {
+                    $startTime = $gameReset->created_at;
+                }
+            }
```
Jamiras: This logic (and probably the next if too) should be duplicated in `getBeatProgressMetrics`. It's not looking at resets at all currently.

Recommend adding a `getAchievementEarningStartTime($achievementSet)` method (better name if you have one).

app/Platform/Commands/RecalculateParentGameMetricsForSubsetResets.php
wescopeland: Yea, I think so.

Renamed the command to:
```
php artisan ra:platform:recalculate-multiset-game-metrics-for-resets
```
And updated the behavior in latest.

app/Platform/Services/PlayerGameActivityService.php
wescopeland: Went with `getAchievementEarningStartTime`, and made this change in latest.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -605,6 +565,62 @@ private function getFilteredAchievementSetMetrics(AchievementSet $achievementSet
...
     }
 
+    /**
+     * Determines the effective start time for calculating achievement-related playtime.
+     * Takes into account the achievements published date, any reset dates, and
+     * optionally the PlayerGame creation time.
+     */
+    private function getAchievementEarningStartTime(
+        AchievementSet $achievementSet,
+        bool $includeSessionAdjustment = false,
```
Jamiras: Why would this be conditional? We want the same "start" time for both beaten and mastery metrics.

wescopeland: Adjusted this in latest and also added another test case to cover the behavior.

PR #4481, opened 2026-01-23
fix(connect): unpromoted achievements should have 0% rarity

Jamiras:
Since a user can't actually unlock unpromoted achievements, their rarity is contrived and may only be non-zero if the achievement was demoted after some number of unlocks occurred.

To minimize confusion, force unpromoted achievements to have 0% rarity. Some clients (like RetroArch) treat this as a signal to not show rarity. A better solution (for newer versions) would be to inform the user that they unlocked an unpromoted achievement. I'll look at doing that when rolling out the Unofficial -> Unpromoted changes.

### DIFF
--- a/app/Console/Commands/UpdateGameActivitySnapshots.php
+++ b/app/Console/Commands/UpdateGameActivitySnapshots.php
@@ -56,15 +56,17 @@ class UpdateGameActivitySnapshots extends Command
      * is genuinely trending. A game with 500 active players when it always has 500 is just popular.
      *
      * Core Formula:
-     *   trend = (new_players_48h + k) * 7 / (new_players_baseline_14d + k)
-     *   score = trend * sqrt(new_players_48h) * evergreen_penalty
+     *   trend = (new_players_48h * 7 + k) / (new_players_baseline_14d + k)
+     *   score = min(trend, 6) * sqrt(new_players_48h) * evergreen_penalty
      *
      * Parameters:
-     * - k=10 Bayesian smoothing: Prevents small-sample noise by adding a prior to both numerator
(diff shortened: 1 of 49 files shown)


tests/Feature/Connect/AchievementSetsTest.php
```diff
@@ -1451,6 +1458,50 @@ function createGameHash(Game $game, GameHashCompatibility $compatibility = GameH
...
+                        'ImageIconUrl' => media_asset($game->image_icon_asset_path),
+                        'Achievements' => [
+                            getAchievementPatchData($data['achievements'][0], 100.0, 100.0), // DisplayOrder: 1
+                            getAchievementPatchData($data['achievements'][2], 100.0, 100.0), // DisplayOrder: 2
+                            getAchievementPatchData($data['achievements'][1], 100.0, 100.0), // DisplayOrder: 3
+                            getAchievementPatchData($data['achievements'][6], 100.0, 100.0), // DisplayOrder: 4
+                            getAchievementPatchData($data['achievements'][3], 100.0, 100.0), // DisplayOrder: 5
+                            getAchievementPatchData($data['achievements'][4], 0.0, 0.0), // DisplayOrder: 6 (unpublished)
+                            getAchievementPatchData($data['achievements'][5], 100.0, 100.0), // DisplayOrder: 7
+                            getAchievementPatchData($data['achievements'][7], 0.0, 0.0), // DisplayOrder: 8 (unpublished)
```
wescopeland: nit: wondering if we should use "unpromoted" instead of "unpublished" in these comments

Jamiras: done. also got all the other references in the file.

PR #4482, opened 2026-01-23
refactor(connect): gameslist and officialgameslist

Jamiras:
Sadly, `officialgameslist` is being used by external clients. There 329 requests logged yesterday. I didn't actually see any requests for `gameslist`, but I know that's used by RAIntegration.dll for the Unknown Games dialog. I'm not sure what causes either to be logged.

I also ran into an issue with the new pest framework where helper functions defined in one file conflict with helper functions in another file if they have the same function name. To get around this (it's a generic PHP problem where functions defined at the top-level scope are globally available to all scripts), I've created static classes in the pest files to contain the helper functions.

### DIFF
--- a/app/Community/Actions/SendDailyDigestAction.php
+++ b/app/Community/Actions/SendDailyDigestAction.php
@@ -261,6 +261,7 @@ class CommentDelayedSubscriptionHandler extends BaseDelayedSubscriptionHandler
                 ->where('commentable_id', $delayedSubscription->subject_id)
                 ->where('id', '>', $delayedSubscription->first_update_id)
                 ->where('user_id', '!=', $delayedSubscription->user_id)
+                ->where('user_id', '!=', Comment::SYSTEM_USER_ID)
                 ->orderBy('id')
                 ->first();
 
@@ -277,6 +278,7 @@ class CommentDelayedSubscriptionHandler extends BaseDelayedSubscriptionHandler
             ->where('commentable_id', $delayedSubscription->subject_id)
             ->where('id', '>=', $delayedSubscription->first_update_id)
             ->where('user_id', '!=', $delayedSubscription->user_id)
+            ->where('user_id', '!=', Comment::SYSTEM_USER_ID)
             ->count();
     }
 
--- a/app/Platform/Actions/CalculateAchievementWeightedPointsAction.php
+++ b/app/Platform/Actions/CalculateAchievementWeightedPointsAction.php
@@ -1,0 +1,109 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Platform\Actions;
+
+/**
+ * Calculates weighted points (RetroPoints) for an achievement based on its rarity and game popularity.
+ *
+ * RetroPoints reward players for earning rare achievements. The core idea is simple: blend base
+ * points (60%) with a rarity multiplier (40%), where rarity = gamePlayers / unlocks. If 1000
+ * people played a game but only 10 unlocked an achievement, that's a 100x rarity multiplier.
+ *
+ * The original formula (RP = points * 0.6 + points * (gamePlayers / unlocks) * 0.4) had two
+ * problems this implementation addresses:
+ *
+ * 1. OBSCURE GAMES WERE UNDERVALUED: For unpopular or notoriously difficult games, often only
+ *    hardcore fans who already know the game well will play them. This makes achievements look
+ *    "easy" (high unlock rate) even when they're genuinely difficult, because casual players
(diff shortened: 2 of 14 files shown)


app/Connect/Actions/GetOfficialGamesListAction.php
```diff
@@ -0,0 +1,62 @@
...
+
+        if (empty($games)) {
+            // replace empty array with empty object so json_encode sends {} to client
+            $games = new stdClass();
+        }
+
+        return [
+            'Success' => true,
+            'Response' => $games,
+        ];
```
wescopeland, line 63: How would you feel about us always inserting a deprecation message string in this response? I'm wondering if we should make it abundantly clear to any _new_ consumers this is a deprecated routine.

Jamiras: Done. But I'm more concerned about new consumers using any dorequest call, and there's probably not an easy way to prevent that.

app/Connect/Actions/GetOfficialGamesListAction.php
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * This action provides support for an internal API function that was only used for a short period
+ * of time. It should be deprecated and eliminated, but other clients have latched onto it and
+ * until they migrate to public APIs, we have to continue supporting it.
+ *
+ * This endpoint must be maintained indefinitely for backwards compatibility with:
+ * - Batocera-based emulationstation: https://github.com/batocera-linux/batocera-emulationstation/blob/7ca0e7151b2764c73ff26f0d4689ab32d9b7abd2/es-app/src/RetroAchievements.cpp#L505
+ *   which is used by other front-ends like RetroBat.
+ */
```
wescopeland, line 22: nit: Recommend also adding a `@deprecated` annotation to this PHPDoc, eg:
```php
/**
 * @deprecated
 *
 * This action provides support...
```

Jamiras: Done. Also added to `LegacyLoginAction`.

app/Connect/Actions/GetOfficialGamesListAction.php
wescopeland: Yea, for this sort of the thing, the only way around it will probably be to have a web API that's easy/intuitive enough to use that consumers are incentivized to use it over the Connect API.

I think for these deprecated calls we really should start turning off the faucets once the V2 API rolls out. There's no reason we should feel forced to maintain these things indefinitely, especially for non-emulator consumers.

I'd propose the following deprecation strategy:
1. Update the deprecation message string to contain a link to some doc on how to migrate this specific call to the V2 Web API.
2. After this has been in the endpoint response for at least 30 days, force 1/1000 calls to fail. Failures should still contain the doc link, as well as an explanation stating the failure is random and 1/N calls are going to fail.
3. Every 60 days, double the number of deliberate failures: 1/1000 -> 1/500 -> 1/250 ... until it's eventually 1/1 and the code can be removed. We could do this programmatically with some kind of middleware that's calendar-driven.

wescopeland: I just did the math on the strategy. Starting at a 1/1000 failure rate results in ~600 days to get down to 1/1.

We could either start more aggressively (1/100) or shorten the interval to 30 days, and that obviously changes the math quite a bit.

1/100 with 60 day intervals gets us to 1/1 in about 15 months.

1/1000 with 30 day intervals gets us to 1/1 in about 10 months.

Jamiras: As much as I would like to actually deprecate this function and lock down the internal API, I don't think random failures is going to get any attention. [The code](https://github.com/batocera-linux/batocera-emulationstation/blob/7ca0e7151b2764c73ff26f0d4689ab32d9b7abd2/es-app/src/RetroAchievements.cpp#L512-L516) just returns an empty dictionary if anything goes wrong with the parsing, so the error won't be reported, and the user will just refresh and things will work and they'll assume it's a network error.

There does seem to be [some logic](https://github.com/batocera-linux/batocera-emulationstation/blob/7ca0e7151b2764c73ff26f0d4689ab32d9b7abd2/es-app/src/RetroAchievements.cpp#L529-L530) for non-200 response codes, but that just captures the descriptive name of the status code (i.e. 404 returns "NOT FOUND"), so maybe we could leverage that. But users would just report something like "Error while accessing retroachievements official games list : GONE" and the developer would not be able to reproduce due to the randomness.

PR #4483, opened 2026-01-23
fix(ResolveSubsetGameRedirectAction): handle multiple parent games

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1464267499465081038.

Now, rather than the action bailing, we redirect to the first set by creation date.


app/Platform/Actions/ResolveSubsetGameRedirectAction.php
Jamiras, line 1: I wonder if this command can be replaced with a call to `$game->getParentGameIdAttribute()` (or vice-versa)

wescopeland: Not fully replaced, but can be heavily simplified.

PR #4486, opened 2026-01-24
feat(activity): add reset events to activity timeline

Jamiras:
Only on the moderation activity timeline. Ticket history is not affected.

Events are inserted into existing sessions when possible. New sessions are created when necessary.

image

image

image

### DIFF
--- a/app/Platform/Services/PlayerGameActivityService.php
+++ b/app/Platform/Services/PlayerGameActivityService.php
@@ -25,7 +25,6 @@ class PlayerGameActivityService
     public array $sessions = [];
     public int $achievementsUnlocked = 0;
     private int $sessionAdjustment = 0;
-    private ?Carbon $lastResetCreatedAt = null;
 
     /** @var Collection<int, PlayerGame> */
     private Collection $playerGames;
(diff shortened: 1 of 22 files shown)


app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -303,6 +303,62 @@ public function addCustomEvent(
...
+            $this->addResetEvent($reset->created_at, "Reset full game");
+        }
+
+        $achievementIds = $game->achievements()->promoted()->pluck('id');
+        $achievementResets = PlayerProgressReset::where('user_id', $user->id)
+            ->where('type', PlayerProgressResetType::Achievement)
+            ->whereIn('type_id', $achievementIds)
+            ->get();
+        foreach ($achievementResets as $reset) {
+            $achievement = $game->achievements()->where('id', $reset->type_id)->first();
```
wescopeland: If a user has reset 100 individual achievements, this kicks off 100 individual queries.

Suggest eager loading or fetching all relevant achievements up front, eg:
```php
// untested pseudocode

$achievements = $game->achievements()->get()->keyBy('id');
// ...
foreach ($achievementResets as $reset) {
    $achievement = $achievements[$reset->type_id] ?? null;
}
```

app/Platform/Services/PlayerGameActivityService.php
wescopeland, line 707: Dumb question: should this also exclude `PlayerGameActivitySessionType::Reset`?

app/Platform/Services/PlayerGameActivityService.php
Jamiras: I thought the `achievements` subcollection would only be queried once, and this was just filtering on it. I didn't consider it would just be returning a Builder.

Changed.

app/Platform/Services/PlayerGameActivityService.php
Jamiras: The later code will ignore it if it doesn't find any Unlock events, but it's not a bad idea to skip it.

Changed.

PR #4492, opened 2026-01-25
fix(TicketListService): exclude tickets where the achievement is hard deleted

wescopeland:
I'm not sure how we ended up with this data to begin with - I don't recall hard deleting any achievements recently.

Resolves these errors:
Screenshot 2026-01-25 at 9 50 34 AM


app/Platform/Services/TicketListService.php
Jamiras, line 1: Need a similar fix on the ticket page itself: http://localhost:64000/ticket/26961

wescopeland: Thanks, added `abort_if(!$ticket->achievement, 404);` to latest:
Screenshot 2026-01-27 at 6 53 53 PM

Jamiras (approved):

PR #4494, opened 2026-01-25
chore: remove the /moderation-comments page, soft delete redundant comments

wescopeland:
This PR:
* Deletes the `/moderation-comments` page. We now fetch all these comments inline on user profile pages - there aren't many of them, certainly not enough to justify the existence of this page.
* Adds a **migration** which soft deletes all but the first and last pairs of account deletion type comments. This is a fire-and-forget, I've warmed up to using migrations for this sort of thing so it's a little more turnkey on deploys.
* Adds some code to `RequestAccountDeletionAccount` to automatically clean up the non-first pair of these comments when the user requests account deletion again.

User ID 1360978 is a good one to test with.

Related: https://discord.com/channels/476211979464343552/1002689485005406249/1465015638379724890

### DIFF
--- a/tests/Feature/Platform/Actions/RequestAccountDeletionActionTest.php
+++ b/tests/Feature/Platform/Actions/RequestAccountDeletionActionTest.php
@@ -117,7 +117,7 @@ class RequestAccountDeletionActionTest extends TestCase
     public function testCleansUpOldDeletionCommentsOnSubsequentRequest(): void
     {
         $this->addServerUser();
-        Mail::fake();
+        Notification::fake();
 
         /** @var User $user */
         $user = User::factory()->create(['Permissions' => Permissions::Registered]);


app/Platform/Actions/RequestAccountDeletionAction.php
```diff
@@ -42,4 +45,39 @@ public function execute(User $user): bool
...
+            ->where('body', 'like', '%canceled account deletion%')
+            ->orderBy('created_at')
+            ->value('id');
+
+        $keepIds = array_filter([$firstRequestId, $firstCancelId]);
+        if (empty($keepIds)) {
+            return;
+        }
+
+        // Soft delete all account deletion comments except the first pair.
```
Jamiras: When viewing an account log, I think I'd be more interested in when they last requested deletion.

Maybe we should keep both the first and the last?

app/Platform/Actions/RequestAccountDeletionAction.php
```diff
@@ -42,4 +45,39 @@ public function execute(User $user): bool
...
+        $baseQuery()
+            ->where(function ($query) {
+                $query->where('body', 'like', '%requested account deletion%')
+                    ->orWhere('body', 'like', '%canceled account deletion%');
+            })
+            ->whereNotIn('id', $keepIds)
+            ->update(['deleted_at' => now()]);
```
Jamiras, line 60: Since these are only being soft-deleted, would it make sense to just filter them when building the view? Then you could note something like "(89 previous cancelled deletions)".

wescopeland: Draft:
Screenshot 2026-01-27 at 7 09 30 PM

Although I can't help but wonder if the value we get from this is worth it.

app/Platform/Actions/RequestAccountDeletionAction.php
wescopeland: Updated in latest so we keep both the first and last.

app/Platform/Actions/RequestAccountDeletionAction.php
Jamiras: Honestly, who looks at this information anyway?

I typically only look at it when a user comments about suddenly not being a developer any more to see if they changed their email or requested account deletion. This is why I felt the need for listing the last request - in case they requested deletion recently and possibly canceled it.

I'm just thinking since we have the information: in addition to the latest request, we might want to know about the first request, and how many total requests have been made.

wescopeland: That's fair. Here's the approach I implemented in latest:
Screenshot 2026-01-31 at 12 16 46 PM

Now we just in-line the ordinal count on the server comment.

app/Helpers/database/user-account-deletion.php
```diff
@@ -20,17 +22,23 @@ function getDeleteDate(string|Carbon $deleteRequested): string
 function cancelDeleteRequest(string $username): bool
 {
     $user = User::whereName($username)->first();
+    if (!$user) {
+        return false;
+    }
+
+    $user->update(['delete_requested_at' => null]);
```
Jamiras, line 29: Cannot actualy cancel deletion request.
```
Add fillable property [delete_requested_at] to allow mass assignment on [App\\Models\\User]. at /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:639
```

wescopeland: Weird, I am not seeing this locally. I've added the field to the fillable list though in latest.

PR #4501, opened 2026-01-27
feat(manage): log achievement credit mutations

wescopeland:
https://github.com/RetroAchievements/RAWeb/pull/4487#pullrequestreview-3701985644

Screenshot 2026-01-27 at 6 46 42 PM


app/Platform/Actions/LogAchievementCreditActivityAction.php
Jamiras, line 1: When adding a new credit, an "original credit" record gets created. This results in two audit log messages. I don't know if that's important:

Image

wescopeland: I think it probably makes sense in terms of data accuracy.

PR #4512, opened 2026-02-01
feat(game): on desktop, use color-graded banner fallbacks

wescopeland:
This PR adds a banner fallback for desktop game pages. The fallback is a CSS color gradient derived from the in-game screenshot, displayed in compact mode. The benefit of doing this is what we can have a consistent heading UX across all game pages.

The banner height during this fallback case is not toggleable, and the user setting (normal, expanded) is not respected. We could support this to be consistent, but I figured pointless consistency would probably be worse than inconsistency: there's no content to truly reveal.

**Before**
Screenshot 2026-02-01 at 4 41 28 PM

**After**
Screenshot 2026-02-01 at 4 41 14 PM


resources/js/features/games/components/GameDesktopBanner/GameDesktopBanner.tsx
Jamiras, line 1: Mobile still uses the screenshot when a banner is not provided. Should it switch to this gradient behavior for consistency?

resources/js/features/games/components/GameDesktopBanner/GameDesktopBanner.tsx
```diff
@@ -22,84 +22,117 @@ export const GameDesktopBanner: FC<GameDesktopBannerProps> = ({ banner }) => {
 
   const isViewingSubset = game.id !== backingGame.id;
 
-  const leftColor = banner.leftEdgeColor ?? '#0a0a0a';
-  const rightColor = banner.rightEdgeColor ?? '#0a0a0a';
+  const hasCustomBanner = !!banner?.desktopMdWebp;
+
+  const leftColor = banner?.leftEdgeColor ?? '#0a0a0a';
+  const rightColor = banner?.rightEdgeColor ?? '#0a0a0a';
```
Jamiras, line 28: How/where are color values calculated from the screenshot?

A large number of pages I visit generating very dark colors, often black or near black values, even when there's not much black in the screenshot (http://localhost:64000/game/1825, http://localhost:64000/game/12326, http://localhost:64000/game/2714)

Others are generating non-black for mostly black screenshots (http://localhost:64000/game/5514, http://localhost:64000/game/20056). 

And some appear to be picking odd colors (like red for http://localhost:64000/game/90 and dark blue for http://localhost:64000/game/10631).

resources/js/features/games/components/GameDesktopBanner/GameDesktopBanner.tsx
wescopeland: IMO, those screenshots work pretty well at an XS scale. I heavily lean towards not making that change.

resources/js/features/games/components/GameDesktopBanner/GameDesktopBanner.tsx
wescopeland: It's generated from here:
```tsx
<>
  <div className="absolute inset-0">
    <img
      src={game.imageIngameUrl}
      alt=""
      aria-hidden="true"
      className="h-full w-full object-cover"
      style={{
        filter: 'blur(80px) saturate(1.8)',
        transform: 'scale(4)',
      }}
      data-testid="fallback-color-source"
    />
  </div>

  <div className="absolute inset-0 bg-black/35 light:bg-black/25" />
  <div className="absolute inset-0 bg-gradient-to-t from-black/30 to-transparent" />
</>
```

A best-case scenario would be if we could calculate the dominant colors server-side and use that for the fallback gradients. Unfortunately, that'd require backfilling for every ingame image we have. The current solution uses CSS trickery: scale the ingame screenshot, blur the heck out of it, and apply some layers.

With this technique, there will always be fallbacks where it's difficult to tell where the colors are exactly coming from relative to the ingame image. In latest, I've adjusted the transforms just a little bit more, and we may want to adjust it further before we're totally happy with it.

With Mario is Missing, the grey is coming from the buildings:
Screenshot 2026-02-04 at 9 00 27 AM

With Bushi Seiryuuden: Futari no Yuusha, the greenish tint is coming from the cliff-like thing in the background:
Screenshot 2026-02-04 at 9 01 16 AM

With Who Wants to Be a Millionaire: 2nd Edition, the green is coming from the answer option:
Screenshot 2026-02-04 at 9 06 45 AM

Jamiras (approved):

PR #4514, opened 2026-02-03
refactor(connect): add SubmitLeaderboardEntryAction

Jamiras:


tests/Feature/Connect/SubmitLeaderboardEntryTest.php
```diff
@@ -53,885 +54,1069 @@ private function buildEntry(int $rank, User $user, int $score, Carbon $when): ar
...
+                    'BestScore' => $score,
                     'RankInfo' => [
                         'NumEntries' => 1,
                         'Rank' => 1,
                     ],
                     'TopEntries' => [
-                        $this->buildEntry(1, $this->user, $bestScore, $now),
-                    ],
-                    'TopEntriesFriends' => [
-                        $this->buildEntry(1, $this->user, $bestScore, $now),
```
Jamiras, line 90: NOTE: when changing this to an action, I've removed the `Success`, `ScoreFormatted`, `LBData` and `TopEntriesFriends` fields from the `Response` object. An overall `Success` is still returned at the top level. The other fields haven't been used by the client for at least five years (they were [noted as unused](https://github.com/RetroAchievements/rcheevos/blob/046e698419197aa35aab9e9784ca5a449780b1d8/src/rapi/rc_api_runtime.c#L443-L447) when the API was migrated to rcheevos).

PR #4523, opened 2026-02-06
fix timing calculations for multiset

Jamiras:


app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -170,7 +184,7 @@ public function execute(PlayerGame $playerGame, bool $silent = false): void
             // ==========================
         }
 
-        $playerAchievements = $achievementsUnlocked->pluck('pivot');
+        $playerAchievements = $achievementsUnlocked->where('game_id', $playerGame->game_id)->pluck('pivot');
```
Jamiras, line 187: This addresses https://discord.com/channels/310192285306454017/1466362963043553506/1466362963043553506

Without the filter, a subset occurring after the base set was completed would still be flagged as the `last_unlock_at`, which is used for building the recent masters list.

app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -142,6 +142,20 @@ public function execute(PlayerGame $playerGame, bool $silent = false): void
...
+                if ($playerAchievementSet->completed_at > $playerAchievementSet->last_unlock_at) {
+                    // completed_at was previously set based on a hardcore unlock - use that
+                    $playerAchievementSet->last_unlock_at = $playerAchievementSet->completed_at;
+                } elseif (!$playerAchievementSet->completed_at) {
+                    // game hasn't been completed yet, also count the hardcore unlock as a softcore unlock
+                    $playerAchievementSet->last_unlock_at = $playerAchievementSet->last_unlock_hardcore_at;
+                }
+
+                $playerAchievementSet->time_taken += (int) $playerAchievementSet->last_unlock_at->diffInSeconds($lastUnlockAt, true);
+            }
```
Jamiras, line 157: This isn't specifically related to multiset. If a user unlocks some number of softcore achievements, then finishes the set in hardcore, the softcore completion time would only be up until the last non-hardcore unlock. It's probably not something that happens all that often, but I created that scenario in the unit test.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -136,7 +136,8 @@ public function initialize(User $user, Game $game, bool $withSubsets = false): v
                 $whenAfter = $playerGame->last_played_at->clone()->addMinutes(5);
 
                 foreach ($this->sessions as &$session) {
-                    if ($session['endTime'] >= $whenBefore && $session['endTime'] <= $whenAfter) {
+                    if ($session['endTime'] >= $whenBefore && $session['endTime'] <= $whenAfter
+                        && $session['achievementSetId'] === $playerGame->achievement_set_id) {
```
Jamiras, line 140: This prevents truncating a session if the first subset unlock was within 5 minutes of starting the multiset session.

app/Platform/Services/PlayerGameActivityService.php
```diff
@@ -760,7 +762,13 @@ private function calculatePlaytime(?Carbon $startTime, ?Carbon $endTime, int $un
                 $sessionEndTime = $firstNonHardcore;
             }
 
+            if ($lastSessionEndTime > $sessionStartTime) {
+                // don't double count overlapping sessions between set and subset
+                $sessionStartTime = $lastSessionEndTime;
+            }
```
Jamiras, line 768: I'm not sure if this will actually happen in production, or if I set up something incorrectly in the unit test, but this prevents an issue where a base set session and a subset session overlap (base set = t1-t3, subset = t2-t4, overlap = t2-t3)

wescopeland (approved):

PR #4525, opened 2026-02-06
feat(gamelist): add beaten time/percentage columns

Jamiras:
Adds "Time to Beat" and "Beat %" columns to the All Games, System Games, and Want to Play Games lists. These columns are hidden by default.

Hopefully, "Beat %" can be used for players looking for difficult games. Note: Beat % reflects the number of users who have completed all of the progression achievements for a set. As such, it will always report 0% for subsets since they cannot be beaten. It will also always report 0% for games that have not yet had progression markings added as they also cannot be beaten. 

Additionally, it's currently only looking at the hardcore beat ratio. This should give players a fairly accurate guage of how difficult **the game** is - the set difficulty would be better judged by looking at the "Mastery %".

I chose not to implement "Time to Master" and "Mastery %" this PR (even though they're more requested) as that data is on the `achievement_set` so every row would have to find the core achievement set for the game to get at the data. I'm not sure how much overhead that would add, and haven't put any thought into whether or not we should denormalize that data onto the game record. I expect it would be desirable to have exposed in the V2 APIs as well without having to jump through extra hoops with include parameters.

I've also updated the "Hardest Games" link in the Games menu to use a sort by low Beat % as the new RR formula softens the modifier for ultra-rare unlocks. We may want to switch that to low Mastery % once that's available, but that would be "Hardest Sets", not "Hardest Games".


resources/views/components/menu/main.blade.php
```diff
@@ -77,7 +77,7 @@
             <x-dropdown-header>Miscellaneous</x-dropdown-header>
             <x-dropdown-item :href="route('game.index')">All Games</x-dropdown-item>
             {{--<x-dropdown-item href="popularGames.php">Most Played</x-dropdown-item>--}}
-            <x-dropdown-item :href="route('game.index', ['sort' => '-retroRatio'])">Hardest Games</x-dropdown-item>
+            <x-dropdown-item :href="route('game.index', ['sort' => 'beatRatio'])">Hardest Games</x-dropdown-item>
```
Jamiras, line 80: This has the weird side-effect of loading the list sorted by a non-visible column. I looked at the code in `useGameListState`, but couldn't figure out how to generically force the column associated to the current sort to be visible. In addition to this specific use case, it would address bookmarks and shared links where people are sorting by columns that aren't visible by default.

lang/en_US.json
wescopeland, line 1: Looks like we already have a "Time to beat" key, which is causing the capitalization inconsistency in the menu options.

Image

To get around this, I recommend adding a new field to _en_US.json_ where the key is a camelcased string that describes that value, ie:
```
"tableTimeToBeatColumn": "Time to Beat"
```
This probably also works just as well:
```
"Time to Beat": "Time to Beat"
```

resources/js/features/game-list/components/AllGamesMainRoot/AllGamesMainRoot.test.tsx
```diff
@@ -780,4 +780,142 @@ describe('Component: AllGamesMainRoot', () => {
...
+
+    // ACT
+    await userEvent.click(screen.getByRole('button', { name: /columns/i }));
+    await userEvent.click(screen.getByRole('menuitemcheckbox', { name: /beat %/i }));
+    await userEvent.click(screen.getByRole('menuitemcheckbox', { name: /time to beat/i }));
+    await userEvent.keyboard('{escape}');
+
+    // ASSERT
+    expect(screen.queryByRole('columnheader', { name: /beat %/i })).toBeVisible();
+    expect(screen.queryByRole('columnheader', { name: /time to beat/i })).toBeVisible();
```
wescopeland: Testing, nit: recommend swapping these `queryBy*` calls with `getBy*`. (also applies to the other new test cases)

`queryBy` is most useful when you want to check if an element is _not_ present. For example:
```ts
expect(screen.queryByRole('columnheader', { name: /points/i })).not.toBeInTheDocument();
```
For asserting something is present/visible, `getBy` is the right choice.

If a test fails because an element isn't found, `getBy` will dump a very descriptive error message which can be useful for debugging. `queryBy` returns null (it doesn't throw) when an element isn't found, which is why it's used for negative asserts.

resources/js/features/game-list/utils/column-definitions/buildBeatRatioColumnDef.tsx
```diff
@@ -0,0 +1,53 @@
...
+    cell: ({ row }) => {
+      // eslint-disable-next-line react-hooks/rules-of-hooks -- the cell component is a FC. using this hook doesn't break the rules of hooks.
+      const { formatPercentage } = useFormatPercentage();
+
+      const playersTotal = row.original.game?.playersTotal ?? 0;
+      const beatRatio =
+        playersTotal > 0 ? (row.original.game?.timesBeatenHardcore ?? 0) / playersTotal : 0.0;
+
+      return (
+        <p className={playersTotal === 0 ? 'text-muted' : ''}>{formatPercentage(beatRatio)}</p>
```
wescopeland: UX, minor: My gut feeling is we don't need the precision of two decimal places for this value. The game page uses one decimal place:

Image

We can achieve this with a few args to `formatPercentage`:
```ts
{formatPercentage(beatRatio, { maximumFractionDigits: 1, minimumFractionDigits: 1 })}
```

resources/views/components/menu/main.blade.php
wescopeland: I think it's okay - it doesn't bother me at all.

wescopeland (approved):

PR #4528, opened 2026-02-07
feat(api): add V2 JSON:API relationship endpoint for `PlayerGame` resources

wescopeland:
This PR adds a V2 API relationship endpoint for `PlayerGame` resources, accessible via `GET /users/{user}/player-games`.

This is the first relationship endpoint on the `User` resource.

Things worth noting:
1. Player games are accessed exclusively via a relationship route. There is no standalone index, as a global list of these records has no practical value.
2. Hub and event "games" are excluded from the results.
3. Results are default sorted by `-lastPlayedAt` (most recently played first).
4. The resource exposes only game-level data: timestamps, beat milestones, and time tracking. Per-set progress is deliberately omitted, as that data belongs at the `PlayerAchievementSet` level.
5. The `game` and `achievementSets` relationships can be included via `?include=game` or `?include=achievementSets`.
6. The `filter[gameId]` param filters to a specific game.

**How to test this PR**
```bash
# Get a paginated list of player games for a user (50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/users/Scott/player-games \

# Get player games with game data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-games?include=game' \

# Get player games with achievement sets included
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-games?include=achievementSets' \

# Filter to a specific game
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Jamiras/player-games?filter[gameId]=11289' \

# Apply custom pagination
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-games?page[number]=2&page[size]=25' \
```


app/Api/V2/PlayerGames/PlayerGameResource.php
```diff
@@ -0,0 +1,88 @@
...
+class PlayerGameResource extends BaseJsonApiResource
+{
+    /**
+     * Get the resource's attributes.
+     *
+     * @param Request|null $request
+     */
+    public function attributes($request): iterable
+    {
+        return [
```
Jamiras, line 24: Each record is also returning it's type/id. Without direct access, should those be exposed?
```
"data":[
      {
         "type":"player-games",
         "id":"648041",
         "attributes":{
```

app/Api/V2/PlayerGames/PlayerGameResource.php
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param Request|null $request
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'game' => $this->relation('game')->withoutLinks(),
+            'user' => $this->relation('user')->withoutLinks(),
+        ];
+    }
```
Jamiras, line 53: The response is pretty useless without the `?include=game` parameter. Can we default it?

I guess it's not important when querying against a specific game: `?filter[gameId]=X`, but otherwise, the data has no context.

app/Api/V2/PlayerGames/PlayerGameResource.php
```diff
@@ -0,0 +1,88 @@
...
+        return [
+            // Core set fields.
+            'coreAchievementsTotal' => $this->resource->achievements_total,
+            'coreAchievementsUnlocked' => $this->resource->achievements_unlocked,
+            'coreAchievementsUnlockedHardcore' => $this->resource->achievements_unlocked_hardcore,
+            'coreAchievementsUnlockedSoftcore' => $this->resource->achievements_unlocked_softcore,
+            'corePointsTotal' => $this->resource->points_total,
+            'corePoints' => $this->resource->points,
+            'corePointsHardcore' => $this->resource->points_hardcore,
+            'corePointsWeighted' => $this->resource->points_weighted,
```
Jamiras: Do we still want to call the base set "core"?

Jamiras: Also note that these fields are deprecated and should really be read from the player_achievement_set for the base game.

https://github.com/RetroAchievements/RAWeb/blob/c25aaec516294bd8179d705cfd85fbd74ad28392/app/Platform/Actions/UpdatePlayerGameMetricsAction.php#L170-L184

app/Api/V2/PlayerGames/PlayerGameResource.php
```diff
@@ -0,0 +1,88 @@
...
+
+            // All sets fields.
+            'achievementsTotal' => $this->resource->all_achievements_total,
+            'achievementsUnlocked' => $this->resource->all_achievements_unlocked,
+            'achievementsUnlockedHardcore' => $this->resource->all_achievements_unlocked_hardcore,
+            'pointsTotal' => $this->resource->all_points_total,
+            'points' => $this->resource->all_points,
+            'pointsHardcore' => $this->resource->all_points_hardcore,
+            'pointsWeighted' => $this->resource->all_points_weighted,
```
Jamiras: This implies the presence of subsets, but I don't see any way to query them. `include=achievementSets` returns an error.

Jamiras: Also note that these are marked as deprecated.

https://github.com/RetroAchievements/RAWeb/blob/c25aaec516294bd8179d705cfd85fbd74ad28392/app/Platform/Actions/UpdatePlayerGameMetricsAction.php#L83-L92

app/Api/V2/PlayerGames/PlayerGameResource.php
```diff
@@ -0,0 +1,88 @@
...
+            'beatenAt' => $this->resource->beaten_at,
+            'beatenHardcoreAt' => $this->resource->beaten_hardcore_at,
+            'coreCompletedAt' => $this->resource->completed_at,
+            'coreCompletedHardcoreAt' => $this->resource->completed_hardcore_at,
+
+            // Time tracking.
+            'playtimeTotal' => $this->resource->playtime_total,
+            'timeToBeat' => $this->resource->time_to_beat,
+            'timeToBeatHardcore' => $this->resource->time_to_beat_hardcore,
+            'timeTaken' => $this->resource->time_taken,
```
Jamiras: `time_taken` data was moved to the achievement sets in #3496 (https://github.com/RetroAchievements/RAWeb/pull/3496/changes#r2076632558). 

> time_taken and time_taken_hardcore are being tracked at the achievement_set level now. The game only tracks playtime_total.

> NOTE: I'm not removing time_taken at this time because there is code using it. This PR eliminates the dependency, but leaving the column here allows doing work outside this PR without having to undo/redo the migrations in this PR.

We should probably drop the column now.

app/Api/V2/PlayerGames/PlayerGameResource.php
Jamiras, line 1: I think we need to rethink exposing player_games as anything more than just the "beat" metrics (which are core set specific) and the "timestamps" and "time tracking" subsections. Everything else is better handled at the player_achievement_set level.

app/Api/V2/PlayerGames/PlayerGameResource.php
wescopeland: I was initially laser-focused on removing these when I was drafting the code for this branch.

However, these two specific fields are [mandatory](https://jsonapi.org/format/#document-resource-objects) on every JSON:API resource object:
> A resource object MUST contain at least the following top-level members:
> * id
> * type

app/Api/V2/PlayerGames/PlayerGameResource.php
wescopeland: Unfortunately, defaulting it would go against the spec convention.

Reasoning: **in isolation**, you're right that this response _is_ pretty useless and it'd make sense to default that `include`. However, it's also possible the client already has the game data cached/stored elsewhere from a previous call and doesn't need (or want) this additional overhead.

app/Api/V2/PlayerGames/PlayerGameResource.php
wescopeland: Per https://github.com/RetroAchievements/RAWeb/pull/4528#discussion_r2799420445, in latest I've removed these fields entirely.

Additionally, `?include=achievementSets` is now supported. Per-set player progress on this endpoint isn't available yet, but I've left a TODO indicating it'll be added later.

app/Api/V2/PlayerGames/PlayerGameResource.php
wescopeland: In latest, I've removed this field from the response.

app/Api/V2/PlayerGames/PlayerGameResource.php
wescopeland: That's a fair point. In latest, I've stripped all this down to the game-specific fields as you've suggested. Pet-set player progress will come via a future `?include=playerAchievementSets` relationship which I've added a TODO for.

PR #4535, opened 2026-02-07
feat: add support for emulator core policies

wescopeland:
This PR adds support for flagging specific emulator cores as either "warned", softcore-only, or blocked entirely.

Tickets opened from these cores are given a new ticket state: Quarantined. A quarantined ticket is not assigned to the achievement author/maintainer, and it does not increment the Open ticket count. It sits indefinitely in a "Pending Review" state so it may one day be passed on to the developer responsible for the emulator core.

The motivation for creating this PR is specific cores have been flagged during a data analysis for having 8x higher ticket rates than supported alternatives. This is taking up a lot of precious volunteer time to manage.

**How it works**
A migration is included in this PR.

http://localhost:64000/manage/emulators/12?relation=3
A new "Core Policies" tab is available in the management app for users who can manage our emulators.
Screenshot 2026-02-07 at 5 08 38 PM

When "Add core policy" is clicked, the following dialog appears:
Screenshot 2026-02-07 at 5 08 58 PM

Two policy levels can be chosen: Unsupported and Blocked. If unsupported is used, the player can still unlock achievements, but they'll be forced to softcore mode. If blocked is chosen, a warning achievement will still be sent to the client, but all other achievements and leaderboards are removed from the response.
Screenshot 2026-02-07 at 5 10 06 PM

The optional "Recommendation" text will be appended to the warning achievement description the user will see in their emulator when using the unsupported core.

When a policy is created, it's added to the table:
Screenshot 2026-02-07 at 5 11 27 PM

When a user makes a call to `patch` or `achievementsets` from the unsupported core, a warning achievement appears in the response:
Screenshot 2026-02-07 at 5 12 36 PM

If the user's most recent session with the game used an unsupported core and they create a ticket, the ticket is automatically given a Quarantined status. The breadcrumb still says "Open Tickets", but the ticket is not really open. Notifications aren't sent out regarding the ticket.
Screenshot 2026-02-07 at 5 15 19 PM


app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorCorePoliciesRelationManager.php
```diff
@@ -0,0 +1,119 @@
...
+    public function form(Schema $schema): Schema
+    {
+        return $schema
+            ->components([
+                Forms\Components\TextInput::make('core_name')
+                    ->label('Core Name')
+                    ->required()
+                    ->maxLength(80)
+                    ->placeholder('dolphin')
+                    ->helperText('Lowercase core identifier prefix as it appears before "_libretro" in user agent strings. This is a prefix match. "doublecherry" will also match "doublecherrygb".'),
```
Jamiras: This isn't going to work. We want to allow `snes9x`, but not `snes9x2010`.

app/Models/Emulator.php
```diff
@@ -236,6 +236,14 @@ public function userAgents(): HasMany
         return $this->hasMany(EmulatorUserAgent::class);
     }
 
+    /**
+     * @return HasMany<EmulatorCorePolicy, $this>
+     */
+    public function corePolicies(): HasMany
+    {
+        return $this->hasMany(EmulatorCorePolicy::class);
+    }
```
Jamiras: Tying these to individual emulators means we have to duplicate the restriction in every frontend: RetroArch, RAlibretro, Firelight, etc.

Also note that it doesn't appear as if Firelight or Bizhawk currently embed core information in their user agents.

app/Connect/Actions/GetAchievementSetsAction.php
```diff
@@ -54,7 +55,12 @@ protected function initialize(Request $request): ?array
         $this->userAgentService = new UserAgentService();
         $this->clientSupportLevel = $this->userAgentService->getSupportLevel(request()->header('User-Agent'));
         if ($this->clientSupportLevel === ClientSupportLevel::Blocked) {
-            return $this->unsupportedClient();
+            // Core-specific blocks show warnings but still allow game loading.
+            // Only emulator-level blocks should fully reject the request.
+            $corePolicy = $this->userAgentService->getCorePolicyForUserAgent(request()->header('User-Agent'));
+            if (!$corePolicy) {
+                return $this->unsupportedClient();
+            }
```
Jamiras: I'm not sure I understand the desire for this logic. If the emulator is fully blocked, why would a core in the emulator be allowed through if it's just restricted?

app/Connect/Actions/GetAchievementSetsAction.php
```diff
@@ -465,18 +475,40 @@ private function buildAllSetsOptedOutPatchData(
...
-            $warningAchievement = (new CreateWarningAchievementAction())->execute(
-                title: match ($this->clientSupportLevel) {
+            // Core-specific policies use a different warning title and description
+            // than emulator-level warnings.
+            $corePolicy = $this->userAgentService?->getCorePolicyForUserAgent(request()->header('User-Agent'));
+
+            if ($corePolicy) {
+                $title = 'Warning: Unsupported Core';
+
+                // Warned cores still allow hardcore, so the description should
```
Jamiras: I think this should read:
> Warned cores _can_ still allow hardcore...

app/Connect/Actions/GetAchievementSetsAction.php
```diff
@@ -488,6 +520,31 @@ private function injectClientSupportWarning(array &$response): void
         }
     }
 
+    /**
+     * When a core-specific policy blocks the client, replace all achievements
+     * with a single warning achievement and remove all leaderboards.
+     */
+    private function replaceWithBlockedCoreWarning(array &$response): void
+    {
+        $corePolicy = $this->userAgentService?->getCorePolicyForUserAgent(request()->header('User-Agent'));
```
Jamiras: You make this call many times. Consider storing it in a variable or class field.

app/Connect/Actions/GetAchievementSetsAction.php
Jamiras: Edit: I see. The core restriction is also validated in the first call. Consider making a second function on UserAgentService that could be used like this:
```
[$this->clientSupportLevel, $corePolicy] = $this->userAgentService->getSupportLevelAndCorePolicy(...)
```

app/Filament/Resources/EmulatorResource/RelationManagers/EmulatorCorePoliciesRelationManager.php
wescopeland: Updated in latest so this is now an exact match.

app/Models/Emulator.php
wescopeland: In latest, the migration has been significantly adjusted and the table is now named `emulator_core_restrictions`. While "emulator_" is the table prefixed, there is no FK to `emulators`. The relation manager has been removed in favor of a new resource:

Screenshot 2026-02-12 at 8 57 05 PM

The resource can be managed by users who can manage emulators.

The restrictions now apply regardless of which emulator loads the core.

It may be desirable to request Firelight or BizHawk to include metadata in their user agents regarding what core is being used.

app/Connect/Actions/GetAchievementSetsAction.php
wescopeland: Refactored to this suggestion in latest.

app/Connect/Actions/GetAchievementSetsAction.php
wescopeland: Updated in latest.

app/Connect/Actions/GetAchievementSetsAction.php
wescopeland: The core restriction is now resolved a single time via `getSupportLevelAndCoreRestriction()` and stored as a class field.

app/Filament/Resources/EmulatorCoreRestrictionResource.php
```diff
@@ -0,0 +1,125 @@
...
+    public static function form(Schema $schema): Schema
+    {
+        return $schema
+            ->components([
+                Forms\Components\TextInput::make('core_name')
+                    ->label('Core Name')
+                    ->required()
+                    ->maxLength(80)
+                    ->placeholder('dolphin')
+                    ->helperText('Lowercase core identifier as it appears before "_libretro" in user agent strings. This is an exact match.'),
```
Jamiras: Since we're decoupling cores from emulators and using exact matches, we should probably also require the `_libretro` as part of the name, that way non-libretro cores with the same prefix can be filtered separately.

app/Filament/Resources/EmulatorCoreRestrictionResource.php
```diff
@@ -0,0 +1,125 @@
...
+                    ])
+                    ->helperText('Use Warned for minor issues, Unsupported for cores that should not earn hardcore, and Blocked for cores with major debilitating issues.'),
+
+                Forms\Components\TextInput::make('recommendation')
+                    ->label('Recommendation')
+                    ->maxLength(255)
+                    ->placeholder('Please use standalone Dolphin instead.')
+                    ->helperText('If set, this text is appended to the warning message shown to the player in their emulator.'),
+
+                Forms\Components\Textarea::make('notes')
```
Jamiras, line 60: recommend making this required.

app/Filament/Resources/EmulatorCoreRestrictionResource.php
Jamiras: Also note there might need to be multiple variations for the same core (I don't know if this is a linux thing, or just something that changed in the libretro build process and one of these users hasn't updated their cores in a while):
```
RetroArch/1.21.0 (Linux 6.12) genesisplusgx_libretro/v1.7.4-302fe82
RetroArch/1.21.0 (Windows 11 Pro x64 25H2 (10.0.26100.7623) 10.0) genesis_plus_gx_libretro/v1.7.4_7c5819b
```

resources/views/pages/ticket/[Ticket].blade.php
Comment (Jamiras) [inline resources/views/pages/ticket/[Ticket].blade.php:1]: To unquarantine a ticket, you have to "reopen this ticket". Maybe that label should be changed to "Approve this ticket" or something similar?

app/Filament/Resources/EmulatorCoreRestrictionResource.php
wescopeland: Updated in latest so this is now an exact match. We will come across a few cores where we need multiple rows (like your example above - Dolphin will also require at least 2 rows to cover Android):
Screenshot 2026-02-13 at 7 12 41 PM

app/Filament/Resources/EmulatorCoreRestrictionResource.php
wescopeland: Updated in latest so the field is required and the column is no longer nullable:
Screenshot 2026-02-13 at 7 12 35 PM

resources/views/pages/ticket/[Ticket].blade.php
Comment (wescopeland) [reply resources/views/pages/ticket/[Ticket].blade.php:1]: Updated in latest:
Screenshot 2026-02-13 at 7 12 26 PM

Comment (Jamiras) [reply resources/views/pages/ticket/[Ticket].blade.php:1]: Should the audit comment also be changed?

Image

wescopeland:
Good catch - it should, taken care of in https://github.com/RetroAchievements/RAWeb/pull/4535/commits/e1d88681c93b33cce132af5a801fd0f99e825f7a.

PR #4540, opened 2026-02-08
fix(UserGameListApiController): validate request input

wescopeland:
Resolves this error from a user calling the endpoint manually:
Screenshot 2026-02-08 at 12 55 00 PM


Jamiras (approved):
While I agree with this change, I wonder why it's necessary. We should fix the source, not the destination. 

Do you have any idea how an invalid enum would get passed to that **internal** endpoint? I couldn't even force the error by modifying any of the existing user game page list URLs - I just got back empty lists or 404s.

PR #4543, opened 2026-02-09
feat(AchievementResource): allow editing the embed_url field again

wescopeland:
https://discord.com/channels/310192285306454017/310195377993416714/1470413361693986837


app/Filament/Resources/AchievementResource.php
```diff
@@ -266,24 +265,15 @@ public static function form(Schema $schema): Schema
...
-                                    // Only allow clearing the field, not changing to a different value.
-                                    if (!empty($value) && $value !== $record?->embed_url) {
-                                        $fail('This field is deprecated. You can only clear it, not change it to a different URL.');
-                                    }
-                                },
-                            ]),
-                    ])
-                    ->visible(fn (?Achievement $record): bool => !empty($record?->embed_url)),
+                            ->disabled(!$user->can('updateField', [$schema->model, 'embed_url'])),
+                    ]),
```
Jamiras, line 313: This field should have some sort of validation. At a minimum, URL validation. Should submissions also be restricted to youtube.com?

app/Filament/Resources/AchievementResource.php
Jamiras, line 1: Should this field also be visible on the View page?

wescopeland: It should be - fixed in latest.

app/Filament/Resources/AchievementResource.php
wescopeland: Agreed on both counts. I've added this validation in latest.

PR #4544, opened 2026-02-10
feat: add achievement2 foundation

wescopeland:
This PR lays the foundation for the React/Inertia achievement details page. The new page is currently only accessible to privileged users - this is very WIP and is not a beta feature yet.

> [!NOTE]
> These screenshots should be taken with a **huge** grain of salt. There is a lot of functionality missing, and of the functionality there, most of it hasn't had a proper UX pass.
> 
> **The purpose of this PR is to get a foundation into our codebase I can iterate on.**

http://localhost:64000/achievement2/9

Screenshot 2026-02-09 at 6 57 35 PM

Screenshot 2026-02-09 at 6 57 46 PM


resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
```diff
@@ -0,0 +1,193 @@
...
+    </div>
+  );
+};
+
+interface PointsLabelsProps {
+  points: number | undefined;
+  pointsWeighted: number | undefined;
+}
+
+const PointsLabels: FC<PointsLabelsProps> = ({ points, pointsWeighted }) => {
```
Jamiras: shouldn't this be in a separate file?

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+            <div className="flex flex-col gap-3">
+              <p>{achievement.description}</p>
+
+              <div className="hidden md:block">
+                <PointsLabels
+                  points={achievement.points}
+                  pointsWeighted={achievement.pointsWeighted}
+                />
+              </div>
```
Jamiras, line 73: The image isn't quite large enough to support a title, description and points values.

Image

Consider putting them near the Progression marker similar to mobile:

Image

Image

Jamiras: I guess it doesn't really matter once the descriptions start getting longer.

Image

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+              <span className="hidden whitespace-nowrap md:block">
+                {t('{{percentage}} unlock rate', {
+                  percentage: formattedUnlockPercentage,
+                })}
+              </span>
+            </p>
+          </div>
+
+          <div className="hidden text-xs md:block">
```
Jamiras, line 152: i can't find this information at SM level.

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
```diff
@@ -0,0 +1,193 @@
...
+            <p className="flex gap-1">
+              <span>{t('{{val, number}} softcore', { val: unlocksSoftcoreTotal })}</span>
+              <span className="text-neutral-700 light:text-neutral-300">{'·'}</span>
+              <span>{t('{{val, number}} hardcore', { val: unlocksHardcoreTotal })}</span>
+              <span className="text-neutral-700 light:text-neutral-300">{'·'}</span>
+              <span>{t('playerCount', { val: playersTotal, count: playersTotal })}</span>
```
Jamiras, line 158: Image

I assume the "0 players" is because the achievement is demoted, but I don't see any mention of unpromoted here or in #4549.

I imagine the whole progress bar and stats widget should be hidden if the achievement isn't promoted. The old page just shows an unitialized widget:

Image

resources/js/common/components/AchievementBreadcrumbs/AchievementBreadcrumbs.tsx
Jamiras, line 1: Would it make sense to eliminate the breadcrumb and use chips similar to the game page?

Image
I'm not sure where the best place to put the description would be.

resources/js/features/achievements/components/+show-sidebar/AchievementShowSidebarRoot.tsx
```diff
@@ -0,0 +1,20 @@
+import type { FC } from 'react';
+
+import { AchievementGamePanel } from '../AchievementGamePanel';
+
+export const AchievementShowSidebarRoot: FC = () => {
+  return (
+    <div data-testid="sidebar" className="flex flex-col gap-6">
+      <AchievementGamePanel />
```
Jamiras: It's kind of weird having the game panel in the sidebar (particularly on smaller sizes where it ends up below the comments).

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
```diff
@@ -0,0 +1,193 @@
...
+              />
+            </div>
+          </div>
+
+          {achievement.unlockedHardcoreAt || achievement.unlockedAt ? (
+            <div className="flex w-full items-center gap-2 px-1.5">
+              <div
+                className={cn(
+                  'flex items-center gap-1.5',
+                  achievement.unlockedHardcoreAt ? 'text-[gold] light:text-amber-500' : null,
```
Jamiras: Should the softcore unlock be gray?

Image

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
wescopeland: It isn't currently there at XS or SM (I would be happy to add it back - I recognize it's a behavior change from the current Blade implementation).

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
wescopeland: Agreed - this metadata should be hidden for unpromoted achievements. Let's punt that change to a separate PR. I'll make a note to tackle this.

resources/js/common/components/AchievementBreadcrumbs/AchievementBreadcrumbs.tsx
wescopeland: This is a tough change to accommodate for mobile - it's almost guaranteed that game chip will wrap.

If you feel strongly about this, I don't mind exploring it in a subsequent PR, but there's a lot of UX tuning that'll be involved to get this idea right.

I also kinda like the idea of that heading style being reserved for the game page, purely from a data hierarchy POV. The game page is the root container for everything, thus gets this "super heading" style.

resources/js/features/achievements/components/+show-sidebar/AchievementShowSidebarRoot.tsx
wescopeland: It's above the comments on smaller sizes:
Screenshot 2026-02-12 at 12 30 35 PM

The sidebar element needs to be hidden at those sizes though. Done in latest.

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
wescopeland: With a longer description (ie: "World Warp I"), I definitely am a fan of the current label positioning on MD+.

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
wescopeland: Changed to gray in latest.

resources/js/features/achievements/components/AchievementHero/AchievementHero.tsx
wescopeland: If a component is render-only with no internal state/hooks, sometimes I'll lean into locality of behavior and keep it in the only place it's used (`PointsLabels` isn't referenced externally anywhere and is private due to not being exported).

That explanation aside, I've moved this to a separate file in latest.

resources/js/common/components/AchievementBreadcrumbs/AchievementBreadcrumbs.tsx
Jamiras: Everything wraps on mobile. You could always ellipsis it like the game panel does

image

My inclination towards using a chip over a breadcrumb is mostly just for UI consistency. It started with moving the points and progression marker down to mimic the summary bar on the game page

image

Then I noticed the breadcrumb was no longer present and remembered that was because of the chips, so I mocked it all up.

wescopeland: I still can't say I'm convinced we should take this approach, solely due to:
> I also kinda like the idea of that heading style being reserved for the game page, purely from a data hierarchy POV. The game page is the root container for everything, thus gets this "super heading" style.

Some pages should indeed have this heading style, I just don't know if achievements are one of them. Achievements themselves will never have dedicated banners - it'll always be the fallback case.

resources/js/features/achievements/components/+show-sidebar/AchievementShowSidebarRoot.tsx
Jamiras: It's still below the comments on MD.

wescopeland: Bummer, I missed that one. Fixed in https://github.com/RetroAchievements/RAWeb/pull/4544/commits/e12919b6c26e675cd0e283e60b9711dfe69868ec.

PR #4545, opened 2026-02-10
bugfix(connect): fix unlocks not being returned for compatibility tester using patch API

Jamiras:
Fixes an issue where a compatibility tester loading a game using the `patch` API (instead of the newer `achievementsets` API) would not register any previous unlocks.

Also fixes an issue where a non-compatibility test loading a game using the `achievementsets` API would register the previous unlocks, but since the achievements weren't included in the `achievementsets` API, it's a moot point.


app/Connect/Actions/StartSessionAction.php
```diff
@@ -43,8 +44,17 @@ protected function initialize(Request $request): ?array
...
+            $isTester = GameHash::where('game_id', $realGameId)
+                ->where('compatibility', $compatibility)
+                ->where('compatibility_tester_id', $this->user->id)
+                ->exists();
+            if (!$isTester) {
+                // don't create sessions for incompatible hashes unless the player is the tester
+                return $this->emptyResponse();
+            }
+
+            $gameId = $realGameId;
```
Jamiras, line 57: this allows the `patch` API to return the unlocks for the compatibility tester.

clarification: the `patch` API returns a virtual id. this uses that virtual id to determine if the user is a tester.

app/Connect/Actions/StartSessionAction.php
```diff
@@ -79,6 +89,12 @@ protected function process(): array
         // resolved game ID.
         $resolvedSets = collect();
         if ($gameHash) {
+            if ($gameHash->compatibility !== GameHashCompatibility::Compatible
+                && $gameHash->compatibility_tester_id != $this->user->id) {
+                // don't create sessions for incompatible hashes unless the player is the tester
+                return $this->emptyResponse();
+            }
```
Jamiras, line 96: this prevents the `achievementsets` API from returning unlocks for a non-compatibility tester using a non-compatible hash.

clarification: the `achievementsets` API returns the real game id. this performs a secondary check to not process incompatible hashes.

tests/Feature/Connect/StartSessionTest.php
Jamiras, line 1: I also took the opportunity to convert this over to pest.

tests/Feature/Connect/StartSessionTest.php
```diff
@@ -4,235 +4,201 @@
...
-class StartSessionTest extends TestCase
-{
-    use BootstrapsConnect;
-    use RefreshDatabase;
-    use TestsEmulatorUserAgent;
-    use TestsPlayerAchievements;
-
-    private UpsertGameCoreAchievementSetFromLegacyFlagsAction $upsertGameCoreSetAction;
-    private AssociateAchievementSetToGameAction $associateAchievementSetToGameAction;
+uses(LazilyRefreshDatabase::class);
```
wescopeland, line 24: Is there any reason we wouldn't want to use this over `RefreshDatabase`? I'm wondering if we should migrate all our test suites.

Jamiras: What I've found from Google is: because it uses transactions, anything that relies on precise ID values may be affected.  So as long as we're not asserting record IDs, we should be fine. The other thing it can't automatically do is rollback database modifications. There's [at least one](https://github.com/RetroAchievements/RAWeb/blob/21e33f575b2ea29678a3c2222a00ddd6cd2c4cf9/tests/Feature/Connect/GetFriendListTest.php#L221-L224) of those to force an error that could potentially leak SQL.

PR #4546, opened 2026-02-10
cleanup(menu): remove top-level achievements menu

Jamiras:
https://discord.com/channels/310192285306454017/1437983579941376154/1469428027820736683

The top-level Achievements menu has three menu items:
* **All Achievements** - this is still accessible from the Achievements panel on the main page. While I feel that this list is too unwieldy to actually be of use, it doesn't hurt anything to leave it in place until we can do more analysis of how people are using he page.
* **Easy Achievements** - this shows all achievements worth 1 RetroPoint. There's nearly 20,000, so it's pretty pointless to have a "page" highlighting them.
* **Hardest Achievements** - this shows achievements with the highest RetroPoints. Because of how RetroPoints are calculated, this effectively becomes a list of 100 point achievements ordered by fewest players. You have to go through 9 pages before you find the first 50 pointer, and 45 pages before you find the first 25 pointer.

It's also important to note that RetroPoints reflect rarity more than difficulty, so calling them Easiest or Hardest is disingenuous. The only reason the sort works is because difficulty can be estimated by a combination of achievement point values and rarity.

These lists probably made sense in the early days of the site, but now they're just diluted to the point of worthlessness.

NOTE: This PR only removes the menu. You can still get to the "easy" and "hardest" achievements list by sorting on RetroPoints after opening the full list from the Achievements panel on the main page. The "My Unlocked Achievements" page (which also uses achievementList.php) is also still available in the same location as before (the user menu).


wescopeland (approved):
While we're making this change, do you think it'd be worth adjusting our RetroPoints label?:
> A measurement of rarity and estimated difficulty. Derived from points, number of achievers, and number of players.

PR #4547, opened 2026-02-10
chore: filter out hydration errors caused by google translate

wescopeland:
Nothing to really test here - Google Translate is causing Sentry events to get spammed due to hydration errors. I can't separate legit hydration errors from red herrings caused by Google Translate.

This PR detects Google Translate in the Sentry reporting code, and if present (and it's a hydration error), bails before sending an error event.


resources/js/app.tsx
```diff
@@ -32,6 +43,14 @@ Sentry.init({
   tracePropagationTargets: ['localhost', /^https:\/\/retroachievements\.org\/internal-api/],
   replaysSessionSampleRate: import.meta.env.SENTRY_REPLAYS_SESSION_SAMPLE_RATE,
   replaysOnErrorSampleRate: 1.0,
+
+  beforeSend(event) {
+    if (isGoogleTranslateActive() && event.message?.includes('Hydration')) {
```
Jamiras: minor: would it be better to swap the order these? it seems like checking the event for Hydration would be easier than scanning the document for class tags.

this probably isn't called frequently enough to matter.

Jamiras (approved):

resources/js/app.tsx
wescopeland: Good call - it's definitely cheaper to swap the order of these because `isGoogleTranslateActive()` does a full DOM scan. Updated in latest.

PR #4559, opened 2026-02-14
feat(game): make the pixelated filter on zoomable images conditional

wescopeland:
This PR makes a change the zoomable image on game pages.

In the modal, we currently are applying a `pixelated` filter on every image to make them look better at such a large resolution. This PR makes this filtering conditional based on the system ID. Systems with standard 3D-capable hardware now are excluded from this filter.

To see the change, just click on one of the screenshots to pop the zoomable image modal.

Eventually, medialibrary support should solve this problem. This is more of a low-tech band-aid for now that should cover most of the egregious cases where this issue pops up.

**Before**
Screenshot 2026-02-14 at 8 09 07 AM

**After**
Screenshot 2026-02-14 at 8 06 06 AM

### DIFF
--- a/app/Community/Enums/TicketState.php
+++ b/app/Community/Enums/TicketState.php
@@ -14,6 +14,7 @@ enum TicketState: string
     case Open = 'open';
     case Resolved = 'resolved';
     case Request = 'request';
+    case Quarantined = 'quarantined';
 
     public const REASON_DEMOTED = 'Demoted';
 
@@ -24,6 +25,7 @@ enum TicketState: string
             self::Open => 'Open',
             self::Resolved => 'Resolved',
             self::Request => 'Request',
+            self::Quarantined => 'Pending Review',
         };
     }
 
@@ -47,6 +49,7 @@ enum TicketState: string
             self::Open => 1,
             self::Resolved => 2,
             self::Request => 3,
+            self::Quarantined => 4,
         };
     }
 
@@ -61,6 +64,7 @@ enum TicketState: string
             1 => self::Open,
             2 => self::Resolved,
             3 => self::Request,
+            4 => self::Quarantined,
             default => throw new InvalidArgumentException("Invalid legacy TicketState value: {$value}"),
         };
     }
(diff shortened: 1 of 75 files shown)


resources/js/common/utils/getIsSystemPixelated.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Systems with 3D-capable hardware produce screenshots that look worse
+ * with nearest-neighbor upscaling. For these systems, we skip the
+ * pixelated image-rendering filter.
+ */
+const nonPixelatedSystemIds = new Set([
+  2, // Nintendo 64
```
Jamiras, line 7: It's interesting that you chose this system for your example images. I'd argue that it should still be pixelated as the Nintendo 64's native resolution is only 320x240 (SNES is 256x224).

http://localhost:64000/game/10137 is a good example of "sharp" pixels.

Similarly, while PSX does support 640x480, it usually runs at 320x240 due to performance/memory limitations.

DS and DSi are 256x192 per screen, and could easily be pixelated prefered as well.

Saturn is 320x224.

PSP is 480x272

CD-i is 384x240

3D0 is 320x240

3DS is 400x240 for the top screen and 320x240 for the bottom screen

I'd argue that anything less than 640x480 should retain the pixelated filter. 

But as most of the emulators for truly 3D systems support running at resolutions higher than native resolution, I could see people submitting higher resolution screenshots. 

Additionally, playing these games on a CRT would have introduced a bit of pixel blending, so I could go either way on this.

wescopeland: Tricky! I could also see it both ways, but I'm getting hung up on what you wrote here:
> ... as most of the emulators for truly 3D systems support running at resolutions higher than native resolution, I could see people submitting higher resolution screenshots.

I think a reasonable criteria for not enabling this CSS filter is "the system has 3D-capable hardware where emulators commonly upscale beyond system-native resolution".

If we're okay with that as a criteria, then all these "borderline" systems should remain in `nonPixelatedSystemIds` because it's common for emulator upscaling to be desirable and active, and in some cases (like DuckStation and ares), the games are upscaled aggressively:
* N64
* PlayStation
* GameCube + Wii
* Nintendo DS + DSi
* PSP
* Saturn
* Dreamcast

By the same logic, the following systems should _not_ be in this list and I've removed them in latest:
* CD-i
* 3DO
* Zeebo

Jamiras: GameCube, Wii, and Dreamcast are all 640x480, so i have no concerns about not pixelating them.

I'm still waffling on the others. I'm conflicted on keeping them pixel-perfect like images on the 2D systems and allowing a slight blur like a psuedo-CRT filter.

PR #4562, opened 2026-02-14
feat(api): add V2 JSON:API relationship endpoint for PlayerAchievementSet resources

wescopeland:
This PR adds a V2 API relationship endpoint for `PlayerAchievementSet` resources. This is a follow-up to #4528.

These resources are accessible by two means:
1. Directly by `GET /users/{user}/player-achievement-sets`.
2. Sideways through `GET /users/{user}/player-games?include=playerAchievementSets`, which allows per-set breakdowns of the retrieved `PlayerGame` records.

Things worth noting:
1. There is no standalone index for `PlayerAchievementSet` resources. These records only make sense in the context of other related or parent records.
2. Results are default sorted by `-lastUnlockAt` (most recent activity first), mirroring `PlayerGameSchema`'s `-lastPlayedAt` pattern.
3. The `achievementSet` and `game` relationships can be included via `?include=achievementSet` or `?include=game`.
4. The `filter[gameId]` param filters to achievement sets for a specific game (it joins through `game_achievement_sets`).
5. A custom `UserScopedHasManyThrough` relationship class was needed for the sideways `player-games?include=playerAchievementSets`.
6. Virtually every field in schema is sortable.

**How to test this PR**
```bash
# Get a paginated list of player achievement sets for a user (50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/users/Scott/player-achievement-sets \

# Get player achievement sets with game data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievement-sets?include=game' \

# Get player achievement sets with achievement set data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievement-sets?include=achievementSet' \

# Filter to a specific game
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Jamiras/player-achievement-sets?filter[gameId]=11289&include=game' \

# Include per-set breakdowns on player games
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-games?include=playerAchievementSets' \

# Apply custom pagination
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievement-sets?page[number]=2&page[size]=25' \
```

### DIFF
--- a/app/Connect/Actions/SubmitLeaderboardEntryAction.php
+++ b/app/Connect/Actions/SubmitLeaderboardEntryAction.php
@@ -19,7 +19,7 @@ class SubmitLeaderboardEntryAction extends BaseAuthenticatedApiAction
 {
     protected Leaderboard $leaderboard;
     protected int $score;
-    protected ?GameHash $gameHash;
+    protected ?GameHash $gameHash = null;
     protected Carbon $when;
     protected ClientSupportLevel $clientSupportLevel;
 
--- a/app/Filament/Resources/GameHashResource/Pages/AuditLog.php
+++ b/app/Filament/Resources/GameHashResource/Pages/AuditLog.php
@@ -4,8 +4,33 @@ namespace App\Filament\Resources\GameHashResource\Pages;
 
 use App\Filament\Pages\ResourceAuditLog;
 use App\Filament\Resources\GameHashResource;
+use App\Models\GameHash;
+use Illuminate\Contracts\Support\Htmlable;
 
 class AuditLog extends ResourceAuditLog
 {
     protected static string $resource = GameHashResource::class;
+
+    public function getTitle(): string|Htmlable
+    {
+        /** @var GameHash $gameHash */
+        $gameHash = $this->getRecord();
+
+        return "{$gameHash->name} - Audit Log";
+    }
+
+    public function getBreadcrumbs(): array
+    {
+        /** @var GameHash $gameHash */
+        $gameHash = $this->record;
+        $game = $gameHash->game;
+
+        return [
+            route('filament.admin.resources.games.index') => 'Games',
+            route('filament.admin.resources.games.view', $game) => $game->title,
+            route('filament.admin.resources.games.hashes', $game) => 'Hashes',
+            $gameHash->name,
+            'Audit Log',
+        ];
+    }
 }
--- a/app/Platform/Actions/BuildGameListAction.php
+++ b/app/Platform/Actions/BuildGameListAction.php
@@ -242,6 +242,7 @@ class BuildGameListAction
                     'medianTimeToBeatHardcore',
                     'numRequests',
                     'numVisibleLeaderboards',
+                    'playersHardcore',
                     'playersTotal',
                     'pointsTotal',
                     'pointsWeighted',
--- a/app/Platform/Actions/UpdateGameBeatenMetricsAction.php
+++ b/app/Platform/Actions/UpdateGameBeatenMetricsAction.php
@@ -18,15 +18,15 @@ class UpdateGameBeatenMetricsAction
         $query = PlayerGame::where('player_games.game_id', $game->id)
             ->leftJoin('unranked_users', 'player_games.user_id', '=', 'unranked_users.user_id')
             ->whereNull('unranked_users.id')
-            ->whereNotNull('player_games.time_to_beat')
-            ->whereNull('player_games.time_to_beat_hardcore');
+            ->whereNotNull('player_games.beaten_at')
+            ->whereNull('player_games.beaten_hardcore_at');
         [$game->times_beaten, $game->median_time_to_beat] = $this->getMedian($query, 'player_games.time_to_beat');
 
         // Get median time to beat (hardcore).
         $query = PlayerGame::where('player_games.game_id', $game->id)
             ->leftJoin('unranked_users', 'player_games.user_id', '=', 'unranked_users.user_id')
             ->whereNull('unranked_users.id')
-            ->whereNotNull('player_games.time_to_beat_hardcore');
+            ->whereNotNull('player_games.beaten_hardcore_at');
         [$game->times_beaten_hardcore, $game->median_time_to_beat_hardcore] =
              $this->getMedian($query, 'player_games.time_to_beat_hardcore');
 
--- a/app/Platform/Concerns/BuildsGameListQueries.php
+++ b/app/Platform/Concerns/BuildsGameListQueries.php
@@ -346,7 +346,7 @@ trait BuildsGameListQueries
                     break;
 
                 /*
-                 * median time to beat the game in hardcore / total_players
+                 * percentage of hardcore players for the game that have beaten the game
                  */
                 case GameListSortField::BeatRatio->value:
                     $query
@@ -356,7 +356,7 @@ trait BuildsGameListQueries
                                 ELSE 0
                             END AS can_be_beaten"
                         )
-                        ->selectRaw('times_beaten_hardcore / players_total as beat_ratio')
+                        ->selectRaw('times_beaten_hardcore / players_hardcore as beat_ratio')
                         ->orderByDesc('can_be_beaten')
                         ->orderBy('beat_ratio', $sortDirection)
                         ->orderByDesc('players_total');
--- a/resources/js/common/layouts/AppLayout/AppLayout.tsx
+++ b/resources/js/common/layouts/AppLayout/AppLayout.tsx
@@ -25,7 +25,7 @@ interface AppLayoutMainProps {
 
 const AppLayoutMain: FC<AppLayoutMainProps> = ({ children, className }) => {
   return (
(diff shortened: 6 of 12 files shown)


app/Api/V2/PlayerAchievementSets/PlayerAchievementSetResource.php
```diff
@@ -0,0 +1,70 @@
...
+            'lastUnlockHardcoreAt' => $this->resource->last_unlock_hardcore_at,
+            'completedAt' => $this->resource->completed_at,
+            'completedHardcoreAt' => $this->resource->completed_hardcore_at,
+
+            'timeTaken' => $this->resource->time_taken,
+            'timeTakenHardcore' => $this->resource->time_taken_hardcore,
+
+            'createdAt' => $this->resource->created_at,
+            'updatedAt' => $this->resource->updated_at,
+        ];
```
Jamiras, line 45: This data is very much dependent on the game context. Can we somehow embed it here like is done on the achievementSet record?
```
            "types":[
               {
                  "gameId":697,
                  "type":"bonus"
               }
            ]
```
That way callers don't have to fetch the entire game record just to get the related ID.

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetResource.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Get the resource's relationships.
+     *
+     * @param Request|null $request
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'achievementSet' => $this->relation('achievementSet')->withoutLinks(),
+            'game' => $this->relation('game')->withoutLinks(),
```
Jamiras, line 57: When asking for a game for a subset, should the subset game be returned?

```
curl --header 'X-API-Key: ...' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/users/Jamiras/player-achievement-sets?filter\[gameId\]=697&include=game" && echo
```
I'm asking for game 697, but getting back game 697 and game 20026 (the subset). If that's what I wanted, I should have used `&include=achievementSet` instead.

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetSchema.php
```diff
@@ -0,0 +1,107 @@
...
+     * Get the resource fields.
+     */
+    public function fields(): array
+    {
+        return [
+            ID::make(),
+
+            Number::make('achievementsUnlocked', 'achievements_unlocked')->readOnly(),
+            Number::make('achievementsUnlockedHardcore', 'achievements_unlocked_hardcore')->readOnly(),
+            Number::make('achievementsUnlockedSoftcore', 'achievements_unlocked_softcore')->readOnly(),
```
Jamiras: This is a denormalized column that was added for query performance (it's just unlocked-hardcore_unlocked). Do we want to expose it? If so, should we also generate a `pointsSoftcore` value?

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetSchema.php
```diff
@@ -0,0 +1,107 @@
...
+            Number::make('completionPercentage', 'completion_percentage')->sortable()->readOnly(),
+            Number::make('completionPercentageHardcore', 'completion_percentage_hardcore')->sortable()->readOnly(),
+
+            DateTime::make('lastUnlockAt', 'last_unlock_at')->sortable()->readOnly(),
+            DateTime::make('lastUnlockHardcoreAt', 'last_unlock_hardcore_at')->sortable()->readOnly(),
+            DateTime::make('completedAt', 'completed_at')->sortable()->readOnly(),
+            DateTime::make('completedHardcoreAt', 'completed_hardcore_at')->sortable()->readOnly(),
+
+            Number::make('timeTaken', 'time_taken')->sortable()->readOnly(),
+            Number::make('timeTakenHardcore', 'time_taken_hardcore')->sortable()->readOnly(),
```
Jamiras: Would it be prudent to change the label on these to reflect units? i.e. `timeTakenMinutes` or maybe even completely rename them for consumability? i.e. `minutesEarningAchievements`

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetSchema.php
```diff
@@ -0,0 +1,107 @@
...
+
+            DateTime::make('createdAt', 'created_at')->sortable()->readOnly(),
+            DateTime::make('updatedAt', 'updated_at')->readOnly(),
```
Jamiras: Will consumers ever need these?

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetResource.php
Jamiras: Also, this data becomes unusable when queried through a player game without any way to associate the playerAchievementSets to achievementSets.
```
curl --header 'X-API-Key: ...' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/users/Jamiras/player-games?include=game,achievementSets,playerAchievementSets&filter\[gameId\]=697" && echo
```

One is a subset and one is the base set, and the only way to tell them apart currently is to look at the achievement count, which is completely dependent on how many the user has unlocked, and may be the same between the base set and subset.
```
      {
         "type":"player-achievement-sets",
         "id":"8271271",
         "attributes":{
            "achievementsUnlocked":138,
            "achievementsUnlockedHardcore":138,
            ...
         }
      },
      {
         "type":"player-achievement-sets",
         "id":"8271274",
         "attributes":{
            "achievementsUnlocked":166,
            "achievementsUnlockedHardcore":166,
            ...
```

wescopeland: Fair. In latest, I've added a `setContext` attribute (akin to the `types` example) to `PlayerAchievementSet` resources that embeds the game+type pairs:
Screenshot 2026-02-18 at 5 40 12 PM

I think this should resolve both concerns. Callers now shouldn't need to fetch the full game record just for the related ID, and the resources are distinguishable when included on player games.

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetResource.php
wescopeland: Good catch. Fixed in latest:
Screenshot 2026-02-18 at 5 49 23 PM

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetSchema.php
wescopeland: Yep, I intentionally added this here because it's very normal for front-ends to show mixed hardcore+softcore progress for a user's games.

To address this feedback, I originally started going down the path of adding a `pointsSoftcore` field, but then I realized it wouldn't be sortable because it's computed.

Given that, I've removed `achievementsUnlockedSoftcore` in latest.

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetSchema.php
wescopeland: It does seem like a good idea for the field names to communicate their units. I've added units to the player game and player achievement set responses (`*Seconds` suffixes).

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetSchema.php
wescopeland: I could think of some contrived use cases when I added them, but it's probably better to just remove them for now. Done in latest.

app/Api/V2/PlayerAchievementSets/PlayerAchievementSetResource.php
Jamiras: This is not descriptive enough. If there are multiple bonus sets, they still can't be told apart. 

```
curl --header 'X-API-Key: ...' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/users/Jamiras/player-games?include=game,achievementSets,playerAchievementSets&filter\[gameId\]=9093" && echo
```

```
      {
         "type":"player-achievement-sets",
         "id":"24205073",
         "attributes":{
            "achievementsUnlocked":87,
            "achievementsUnlockedHardcore":87,
            ...
            "setContext":[
               {
                  "gameId":9093,
                  "type":"bonus"
               }
            ]
         }
      },
      {
         "type":"player-achievement-sets",
         "id":"24205042",
         "attributes":{
            "achievementsUnlocked":211,
            "achievementsUnlockedHardcore":211,
            ...
            "setContext":[
               {
                  "gameId":9093,
                  "type":"bonus"
               }
            ]
         }
      }
```
Recommend adding an `achievementSetId` value to `setContext`.

wescopeland: Added this field in latest.

PR #4565, opened 2026-02-16
feat(achievement): build AchievementChangelog

wescopeland:
This PR builds out the changelog UI/UX for `achievement2`.

The changelog is presented as a timeline, where the top-most item is the most recent change.
Screenshot 2026-02-16 at 12 01 26 PM

Initial development activities that occurred between creation and the first promotion are tucked away into an "Initial development" collapse:
Screenshot 2026-02-16 at 12 02 25 PM

The changelog is synthesized through 3 "eras" of audit history that have existed in RAWeb.

**Era 1 (before April 2022)**
We just know something was edited, but have no field-level details about those edits.
Screenshot_2026-02-15_at_4 21 49_PM

**Era 2 (April 2022 - February 2024)**
We know _what_ was edited, but not _how_. We may know a specific field on an achievement changed, but we don't know what the field was before or after the change happened.
Screenshot_2026-02-15_at_4 22 35_PM

**Era 3 (February 2024+)**
We have a complete picture of what was edited and precisely how, with full before/after diffs.
Screenshot_2026-02-15_at_4 23 00_PM

Here's an example of an achievement with changes recorded through all 3 eras:
Screenshot 2026-02-16 at 12 03 17 PM

Event circles receive different colors based on the type of event:
Screenshot 2026-02-16 at 12 11 03 PM

Some achievements have no initial promotion event that is recorded anywhere in our audit log. For these, the Created circle is green instead of blue:
Screenshot 2026-02-16 at 12 12 06 PM

If the same type of edit occurs numerous times around a similar timestamp, we collapse this all into a single event:
Screenshot 2026-02-16 at 12 13 25 PM

If the same type of edit ultimately results in a net-zero change (ie: points changed from 4 to 5 and then back to 4), we don't show the event at all.

Good URLs to test with:
http://localhost:64000/achievement2/1 - achievement with changes in all 3 eras, created-as-promoted edge case, has a banned+deleted user who made changes
http://localhost:64000/achievement2/9 - similar to ID 1
http://localhost:64000/achievement2/69990 - many collapsed era 1 edits
http://localhost:64000/achievement2/233062 - promoted, demoted, and promoted again
http://localhost:64000/achievement2/145613 - promoted, demoted, promoted, and many changes
http://localhost:64000/achievement2/431345 - straightforward created, promoted, demoted, promoted diff with no changes recorded
http://localhost:64000/achievement2/561681 - never promoted (at least in my local DB), so no "Initial development" collapse in the UI
http://localhost:64000/achievement2/413233 - lots of initial development diffs


resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    case 'badge-updated':
+      return t('Badge updated');
+
+    case 'embed-url-updated':
+      return t('Embed URL updated');
+
+    case 'logic-updated':
+      return t('Logic updated');
```
Jamiras, line 180: The management page for achievement logic has historical diffs. Consider providing some way for developers to jump to the diff associated to the most recent logic updated event in the collapsed list.

Jamiras: Related note: I'm sure it's probably somewhere on your backlog, but there's no Manage button on the achievement2 page yet.

app/Platform/Actions/BuildAchievementChangelogAction.php
```diff
@@ -0,0 +1,656 @@
...
+            } else {
+                $collapsed[] = $current;
+                $current = $next;
+            }
+        }
+
+        $collapsed[] = $current;
+
+        // Remove entries where collapsing ultimately produced a net-zero change.
+        return array_values(array_filter($collapsed, fn ($e) => $e->count !== $REMOVAL_SENTINEL));
```
Jamiras: I'm assuming the problem is in this code somewhere, but http://localhost:64000/achievement2/70798?tab=changelog is not reporting multiple logic edits (the change on 20 Aug appears to be ignored)

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+      return fromGame && toGame
+        ? t('Transferred from {{fromGame}} to {{toGame}}', { fromGame, toGame })
+        : t('Transferred to a different achievement set');
+    }
+
+    case 'type-set': {
+      const typeName = entry.fieldChanges[0]?.newValue;
+
+      return typeName ? t('Type set to {{typeName}}', { typeName }) : t('Type set');
```
Jamiras: Why does "Type removed" show like a diff, but "Type set" is just a statement?

Image

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
```diff
@@ -0,0 +1,197 @@
...
+    case 'title-updated':
+      return t('Title updated');
+
+    case 'points-changed': {
+      const oldVal = entry.fieldChanges[0]?.oldValue;
+      const newVal = entry.fieldChanges[0]?.newValue;
+
+      return oldVal && newVal
+        ? t('Points changed from {{oldVal}} to {{newVal}}', { oldVal, newVal })
+        : t('Points changed');
```
Jamiras: Why is this not shown using diff-style semantics?

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    case 'badge-updated':
+      return t('Badge updated');
```
Jamiras, line 174: The audit log shows the before and after images. Should this?

If there's concerns about inappropriate content, that could apply to titles and descriptions too.

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
```diff
@@ -0,0 +1,197 @@
...
+                ) : null}
+              </div>
+            ))}
+          </div>
+        ) : null}
+
+        <div className="flex items-center gap-1.5 text-2xs">
+          {entry.user ? <UserAvatar {...entry.user} size={16} showLabel={true} /> : null}
+
+          <span className="text-neutral-500">{formatDate(entry.createdAt, 'll')}</span>
```
Jamiras: It's unimportant for most users, but would it be desirable to include the time in a tooltip?

Or maybe always show the time if it's in the last couple days?

Jamiras: Are there any thoughts of having bulk edits share a dot on the timeline?

Image

Image

What if multiple edits occur on the same day but at different times? Without the time portion of the timestamp, it's unclear when things happened as bulk edits or individual ones.

I guess that's all taken care of by the filament audit log (there should probably be a direct link to that somewhere on this tab). The player-facing changelog doesn't have to be as granular.

resources/js/features/achievements/components/+show/AchievementShowRoot.tsx
Jamiras, line 1: After selecting the Changelog tab and refreshing, the Comments tab becomes selected, but still shows the Changelog data

Image

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
wescopeland: > The management page for achievement logic has historical diffs. Consider providing some way for developers to jump to the diff associated to the most recent logic updated event in the collapsed list.

I think this is a good idea. I'll add this to my TODO list - the complexity is high enough that it probably justifies being in its own separate PR.

> Related note: I'm sure it's probably somewhere on your backlog, but there's no Manage button on the achievement2 page yet.

Yep, this is on my TODO list. Still quite a bit to work through before this page is ready for any sort of beta.

app/Platform/Actions/BuildAchievementChangelogAction.php
wescopeland: Fixed in latest:
Screenshot 2026-02-17 at 5 50 19 PM

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
wescopeland: Yeah, that doesn't flow very well. Updated in latest to be consistent:
Screenshot 2026-02-17 at 5 55 39 PM

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
wescopeland: I was afraid it might be overkill for this field, but every other field change uses the diff-style, so we should probably be consistent. Updated in latest:

Screenshot 2026-02-17 at 5 58 21 PM

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
wescopeland: Added this functionality in latest:
Screenshot 2026-02-17 at 6 13 57 PM

Screenshot 2026-02-17 at 6 14 21 PM

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
wescopeland: I'm okay with a tooltip that shows the time, but I think it should be done using a `title` attribute rather than `BaseTooltip` due to the subtle nuance that `BaseTooltip` creates tap targets, which would be undesirable here.

We could alternatively show it always, but I think for most users this would only be noise, especially given it's in the server's timezone.

Here's what it looks like in latest:
Screenshot 2026-02-17 at 6 17 04 PM

wescopeland: Only vibes, but I think the current approach of separate timeline nodes/events is better for this particular view. For the privileged users who want that deeper context, they should definitely dive into the richer audit log in Filament.

resources/js/features/achievements/components/+show/AchievementShowRoot.tsx
wescopeland: Yep, known issue, and I probably should've mentioned it as such. I'll fix this in a separate PR - it'll need to be fixed by a change in both the Laravel controller and React code.

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
Jamiras: It's actually less consistent now - especially given than you've changed points to use diff mechanics. I would expect removed to have just the red lined out value and set to have a green non-lined-out value.

wescopeland: Updated in latest:
Screenshot 2026-02-19 at 5 26 07 PM

Jamiras: It looks like there's an empty red box on the Type Set item.
image

wescopeland: Thanks - didn't notice it on my monitor. Fixed in latest.

PR #4578, opened 2026-02-20
docs(terms): add section for Achievement Data

wescopeland:
We've gotten a few questions about this recently (ie: https://discord.com/channels/476211979464343552/1474331490216054826).

I've added a new section in our terms to clarify:
Screenshot 2026-02-20 at 7 47 40 AM


resources/js/pages/terms.tsx
Jamiras, line 1: From your screenshot, I assumed the first paragraph was new too. I wonder if we should add a note about media.

> The company names, product names, service names, logotypes, **boxart, and screenshots** on this web site are for identification purposes only.

wescopeland: Yea, this is a good idea - we probably should do this. I went ahead and added that copy:
Screenshot 2026-02-23 at 5 33 58 PM

PR #4584, opened 2026-02-21
fix: remove avif fallback banners

wescopeland:
Resolves https://discord.com/channels/310192285306454017/758865736072167474/1474516509563949097.

It turns out AVIF is probably the least optimal compression algorithm for the specific image used as our banner fallback.

**Uncompressed Source**
Screenshot 2026-02-21 at 8 33 13 AM

**WebP**
Screenshot 2026-02-21 at 8 33 20 AM

**AVIF**
Screenshot 2026-02-21 at 8 33 25 AM

AVIF is good at compressing photos (ie: most conventional banners), but is not good at compressing this sort of fine detail spread all over a single continuous image.

This PR makes several changes:
* Swaps image generation from using `Spatie\Image\Image` to using raw `Imagick`. On my local, I accidentally generated fallbacks using the GD driver, which compresses significantly worse than Imagick.
* Drops the AVIF fallbacks entirely in favor of WebP.


app/Platform/Data/PageBannerData.php
wescopeland, line 1: I'm not a fan of shoving WebP files into fields clearly labeled as AVIF, but I couldn't think of a better idea that also didn't involve a lot of pain over something that's probably superficial.

Jamiras: is there some way to just get the code to prefer the webp over the avif? why are both needed?

Jamiras (approved):

app/Platform/Data/PageBannerData.php
wescopeland: I just realized we can set the avif references here to `null` and the front-end will fall back to the webp assets (made this change in latest).

Generally, avif compresses 20-30% better than WebP and for large photographic content like banner images its lossy compression isn't noticeable. We use both because WebP has significantly broader browser support.

PR #4585, opened 2026-02-22
feat(game): add screenshot media library infrastructure

wescopeland:
Games currently support a single ingame screenshot and a single title screenshot, stored as varchar columns on the `games` table. Uploads are capped at 1MB and forcibly downscaled to 320x240 via GD. Additionally, the original file is destroyed, meaning we are deliberately incurring data loss.

This PR creates the infrastructure for multiple screenshots per game using Spatie MediaLibrary, adds a dedicated Media page for `GameResource` in Filament, and preserves backward compatibility for all existing consumers of the current image assets in the `games` table.

### `game_screenshots` table + `GameScreenshot` model
[See migration](https://github.com/wescopeland/RAWeb/blob/da52d35be0b195800e9b3340da797582e8fc1206/database/migrations/2026_02_20_000000_create_game_screenshots_table.php)

Screenshot business logic gets its own dedicated table. Each row has a type (title / ingame / completion), primary flag, approval status, and ordering via `order_column`.

`completion` is a new type that is meant to serve as the opposite of a title screen - it's a screen showing "The End" that may eventually be used for `user2` beaten showcases.

The primary flag is used to determine what the "canonical" ingame image is when multiple uploads are supported in a subsequent PR.

The approval status is used to support an approval queue, facilitating player uploads of content, which will also be handled in a subsequent PR.

### System resolution metadata
A JSON `screenshot_resolutions` column on `systems` stores valid native output resolutions per system (eg: SNES = 256x224, Saturn = 14 valid resolutions). A boolean `supports_resolution_scaling` column flags systems whose emulators support internal resolution scaling (PSP, PS1, PS2, N64, GameCube, Wii, Dreamcast, DS, DSi, and 3DS). Upload validation checks exact resolution matches and accepts 2x/3x integer multiples for scaling-enabled systems. Systems will null resolutions (computers, arcade) allow any dimensions.

### Upload flow
All uploads go through a new `AddGameScreenshotAction`, which validates file constraints (4MB max, 64x64 min, 1920x1080 max, PNG/JPEG/WebP only, no animation), checks SHA1 for duplicates and known placeholders, enforces per-type caps (20 ingame, 1 title, 1 completion), and validates resolution against the system.

For this piece of work, every Filament upload sets the new screenshot as primary for the type, which matches current production behavior where each upload replaces the previous one. "Demoted" screenshots are set to `pending` status to be reviewed later when multiple files are supported in a subsequent PR.

### Backwards compatibility
`GameScreenshotObserver` automatically syncs the primary screenshot's legacy path back to the `image_ingame_asset_path` column on the `games` table. All existing consumers (V1 API, game pages, etc) continue working without changes.

### Conversions
Original files are now stored on S3. Spatie generates sm (320px), md (640px), and lg (1280px) conversions in both WebP and AVIF. Small images are never upscaled due to the code using `Fit::Max`. These conversions aren't rendered in the React front-end yet.

A subsequent PR will include a command to backfill all existing legacy images to match these conversions, and flag the ones that fail to meet system-level resolution constraints.

### Media Page
As mentioned above, a new "Media" tab on `GameResource` consolidates all game media management: badge, box art, title screenshot, banner, and ingame screenshots. The media section has been removed from the Edit page.

## How to test this PR
```
# adds new game_screenshots table
# updates systems table with new columns
# seeds resolution metadata for dozens of systems
sail artisan migrate

# restart horizon
sail artisan horizon
```

Similar to the banners PR, ensure these are in your _.env_:
```
MEDIA_URL=http://localhost:${APP_PORT}/media

AWS_BUCKET=local
AWS_DEFAULT_REGION=eu-west-1
```

Navigate to any game in Filament and click the Media tab, ie: http://localhost:64000/manage/games/4874/media
Screenshot 2026-02-22 at 7 44 34 AM

In the In-Game Screenshot section, upload a file and **click the Save Changes button**:
Screenshot 2026-02-22 at 7 46 11 AM

Screenshot 2026-02-22 at 7 46 18 AM

You should now be able to verify:
1. A `game_screenshots` row is created with `is_primary = 1` and `status = approved`.
2. A `media` row is created with `collection_name = screenshots` and `custom_properties` containing `sha1` and `legacy_path`.
3. The game's `image_ingame_asset_path` column is updated to the new `/Images/NNNNNN.png` path.
4. In Minio, the original + sm/md/lg WebP & AVIF conversions are present under _local/game/screenshots_. Note, an LG image is only created if the source image is large enough to support it - we never scale up, only down.

Remember, Minio is accessed at http://localhost:64041/buckets and the credentials are `retroachievements / retroachievements`.

Screenshot 2026-02-22 at 7 48 39 AM

Visit the game page for the game. The ingame screenshot should render correctly (everything else being broken is expected given our env vars changes):
Screenshot 2026-02-22 at 7 49 18 AM

Upload a second in-game screenshot for the same game. Verify the first screenshot is demoted via `is_primary = 0` and `status = pending`. The game page should now show the new screenshot:
Screenshot 2026-02-22 at 7 50 26 AM

## What's next

This is the first of several phases for this work.

Tentatively, I envision the following phases looking like this:

The **second** phase will be for backfilling all existing legacy screenshots into the new system, migrating the game title screenshots to Media Library, and adding gallery management in Filament (view existing screenshots, set screenshots to primary, delete stuff, reorder stuff).

The **third** phase will be for viewing multiple ingame screenshots on the game page.

The **fourth** phase will be for supporting player uploads, which go into a Filament-powered review queue (really need to think through this).


database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
wescopeland, line 1: Honestly, most of my time working on this PR was spent here in this migration.

I've done a ton of research and referenced many different sources. Still, as much as I've tried, I'm sure I've made a mistake somewhere. However, given this lives in the DB, we can always change these configuration settings without needing to do a code deploy.

app/Platform/Actions/AddGameScreenshotAction.php
```diff
@@ -0,0 +1,215 @@
...
+            ->map(fn (array $r) => "{$r['width']}x{$r['height']}")
+            ->join(', ');
+
+        $scalingNote = $system->supports_resolution_scaling
+            ? " (or 2x/3x integer multiples)"
+            : '';
+
+        throw ValidationException::withMessages([
+            'screenshot' => "This screenshot's dimensions ({$width}x{$height}) don't match the expected resolutions for {$system->name}: {$formatted}{$scalingNote}.",
+        ]);
```
Jamiras: I have strong opinions against doing exact dimension checking. Every non-handheld system before 2000 (and many after) should expect images where height = width * .75 (4:3 aspect ratio). It doesn't matter what the raw dimensions are - we should aim for the intended dimensions.

Using NES as an example:
* The native dimensions is 256x240 (NTSC), which is cropped to 256x224 and then stretched to fit a 4:3 TV screen
* At 1x scaling and Preserve Aspect Ratio selected, RALibretro renders Super Mario Bros at 292x224, whereas the raw pixel buffer captured by the screenshot button is 256x224.
* 298x224 would be true 4:3, so we can't even just verify the height value is 3/4 of the width.
* At 3x scaling and Preserve Aspect Ratio selected, the window contents are 876x670 (not 876x672 as I would have expected, so there's some rounding going on in the aspect ratio reported by the core). 

When I add screenshots to games, I typically use the largest integer scaling supported by RAlibretro that fits on my screen (using these enlarged images better hides the extra pixels added to stretch the image). I then screenshot the whole window (Alt+Printscreen) to preserve the aspect ratio, then crop out the image from within the window.

app/Platform/Actions/AddGameScreenshotAction.php
```diff
@@ -0,0 +1,215 @@
...
+     */
+    public function execute(
+        Game $game,
+        UploadedFile $file,
+        ScreenshotType $type,
+        ?string $description = null,
+        bool $isPrimary = false,
+    ): GameScreenshot {
+        $this->validateFile($file);
+        $this->validateResolution($file, $game);
```
Jamiras, line 32: FWIW, I did not receive an error trying to upload a 584x448 image to an NES game.

Jamiras: Or a 1024x1536 jpg.

wescopeland: This should've been firing on in-game images, but title images probably needs this validation as well. It's now on both in latest.

Screenshot 2026-02-25 at 6 44 43 PM

app/Platform/Actions/AddGameScreenshotAction.php
wescopeland: Per this feedback and our follow-up discussion, I've extensively reworked the resolution validation to follow an approach similar to [MobyGames](https://www.mobygames.com/info/screenshots/):

* LCD handhelds (GB, GBA, WonderSwan, etc) require exact dimensions + 2x/3x integer multiples.
* Systems with analog TV output (NES, Genesis, PlayStation, etc) require exact dimensions + 2x/3x integer multiples + SMPTE 601 capture resolutions.
* Game-dependent systems (Arcade, DOS, etc) have no dimension validation.

Resolutions are primarily sourced from [our screenshot dimension guidelines](https://docs.retroachievements.org/guidelines/content/game-info-and-hub-guidelines.html#screenshot-dimensions), MobyGames, the libretro docs, and libretro core source code.

`supports_resolution_scaling` has been replaced with `has_analog_tv_output`.

Integer multiples are accepted for all systems now universally rather than this being gated by a system-driven flag.

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
```diff
@@ -0,0 +1,409 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::table('systems', function (Blueprint $table) {
+            $table->json('screenshot_resolutions')->nullable()->after('name_short');
```
Jamiras, line 14: rollback doesn't remove this column

app/Platform/Actions/AddGameScreenshotAction.php
Jamiras: I thought the discussion included stretching the image to 4:3 on the game page/popup. That doesn't appear to be happening.

app/Filament/Resources/GameResource/Pages/Media.php
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    protected function mutateFormDataBeforeSave(array $data): array
+    {
+        $this->authorizeFields($this->record, $data);
+
+        $action = new ApplyUploadedImageToDataAction();
+        $action->execute($data, 'image_icon_asset_path', ImageUploadType::GameBadge);
+        $action->execute($data, 'image_title_asset_path', ImageUploadType::GameTitle);
+        $action->execute($data, 'image_box_art_asset_path', ImageUploadType::GameBoxArt);
```
Jamiras, line 183: When trying to attach both a title image and screenshot in the same Save, only the screenshot gets saved.

Jamiras: The title image also gets set back to 000002 when attaching the ingame image after refreshing.
* Upload title
* Save
* Refresh
* Upload in game image
* Save
* View page/db. Note: title is 000002.

app/Platform/Actions/AddGameScreenshotAction.php
```diff
@@ -0,0 +1,171 @@
...
+            throw ValidationException::withMessages([
+                'screenshot' => 'This image is a known placeholder and cannot be uploaded.',
+            ]);
+        }
+
+        // Reject duplicates based on SHA1 within this game's screenshots collection.
+        $isDuplicate = $game->media()
+            ->where('collection_name', 'screenshots')
+            ->where('custom_properties->sha1', $hash)
+            ->exists();
```
Jamiras, line 122: It's unclear to me how the collection is attached, but does this prevent users from uploading the title image as an in-game image?

app/Rules/ValidScreenshotResolutionRule.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        foreach ($resolutions as $resolution) {
+            $baseW = $resolution['width'];
+            $baseH = $resolution['height'];
+
+            if ($width === $baseW && $height === $baseH) {
+                return true;
+            }
+
+            // Both axes must scale by the same integer factor.
```
Jamiras, line 96: If we're resizing/stretching the images for the game page, shouldn't we always just require the images be unscaled when uploaded?

Jamiras, line 96: If we're resizing/stretching the images for the game page, shouldn't we always just require the images be unscaled when uploaded?

app/Support/Media/CreateLegacyScreenshotPngAction.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace App\Support\Media;
+
+class CreateLegacyScreenshotPngAction
+{
+    private const MAX_WIDTH = 320;
+    private const MAX_HEIGHT = 240;
+
+    /**
+     * Resizes a source image to fit within 320x240 (preserving aspect ratio),
```
Jamiras, line 13: Perhaps this is where the 4:3 stretching should occur?

config/media.php
```diff
@@ -70,8 +70,12 @@
             ],
         ],
 
-        // Future game media types can go here:
-        // 'screenshot' => [...],
+        'screenshot' => [
+            'sm' => ['width' => 320],
+            'md' => ['width' => 640],
+            'lg' => ['width' => 1280],
```
Jamiras, line 76: Will small screenshots (i.e. 256x224) have `md` and `lg` versions, or will they be scaled on the client?

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
```diff
@@ -0,0 +1,409 @@
...
+            // Sony consoles
+            12,  // PlayStation
+            21,  // PlayStation 2
+
+            // Microsoft consoles
+            22,  // Xbox
+
+            // Atari consoles
+            50,  // Atari 5200
+            51,  // Atari 7800
```
Jamiras, line 366: Jaguar/Jaguar CD?

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
```diff
@@ -0,0 +1,409 @@
...
+
+            // NEC consoles
+            8,   // PC Engine/TurboGrafx-16
+            49,  // PC-FX
```
Jamiras, line 373: PC Engine CD?

app/Platform/Actions/AddGameScreenshotAction.php
wescopeland: Added this CSS scaling in latest (to both the thumbnail and zoomable image):
Screenshot 2026-02-27 at 6 05 51 PM

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
wescopeland: I'm not seeing the same locally. I'm confused how this can be - it's explicitly listed in `down()`:
```php
Schema::table('systems', function (Blueprint $table) {
    $table->dropColumn(['screenshot_resolutions', 'has_analog_tv_output']);
});
```

app/Filament/Resources/GameResource/Pages/Media.php
wescopeland: Thanks, fixed both issues in latest:

Screenshot 2026-02-27 at 5 27 48 PM


Screenshot 2026-02-27 at 5 27 22 PM

app/Platform/Actions/AddGameScreenshotAction.php
wescopeland: It doesn't prevent it yet, but it may in the next phase. Currently, title images don't have `media` records created for them, so we're also not writing SHA1 hashes for those files. This is intentional to keep the scope of this PR under control:

> The second phase will be for backfilling all existing legacy screenshots into the new system, migrating the game title screenshots to Media Library, and adding gallery management in Filament (view existing screenshots, set screenshots to primary, delete stuff, reorder stuff).

app/Rules/ValidScreenshotResolutionRule.php
wescopeland: Great question, I can see how this is puzzling. For 2D systems, the scaling is just nearest-neighbor upscaling, so requiring stuff to be unscaled would be fine.

3D systems like PS1, PS2, N64, etc obviously support **internal** resolution scaling, so a 2x/3x capture genuinely will have more detail than a 1x capture that's stretched up. The systems table currently doesn't distinguish between 2D/3D, and some systems like PS1/N64 had games of both styles.

Previously, this PR did have a `supports_resolution_scaling` flag added to `systems`. We replaced it with `has_analog_tv_output`. I think a flag like this is probably what we need to drive the logic you're mentioning in this validation rule. We could bring the flag back if it's desirable - but should we? I'm not 100% sure if this permissiveness will be a problem on those low-res 2D systems.

wescopeland: https://github.com/RetroAchievements/RAWeb/pull/4585#discussion_r2866578801

app/Support/Media/CreateLegacyScreenshotPngAction.php
wescopeland: For now, I'm doing it client-side with CSS aspect ratio rules (https://github.com/RetroAchievements/RAWeb/pull/4585#discussion_r2866574138).

We may want to move it to the server later, but right now I can't think of a good use case to do it server-side instead of client-side.

config/media.php
wescopeland: Small screenshots will **not** have scaled `md` and `lg` versions (the versions technically exist, but they're no-ops). Any upscaling should be done client-side.

These conversions should be considered ceilings rather than targets. If we do the upscaling on the server, we'd effectively be doing nearest-neighbor scaling to produce a larger file that looks identical to what CSS will do for free.

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
wescopeland: Thanks, added in latest. We were also missing Atari 2600 here ☹️

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
wescopeland: Added in latest.

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
Jamiras: Ya, I see it there. Not sure what happened. All I know is I did a rollback/migrate and it complained about the column already existing. Cannot reproduce now.

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
Jamiras: I thought 2600 was explicitly ignored.
```
            // Atari 2600 (ID 25) - TIA is 160px wide but vertical scanline
            // count varies per game (~160 to ~230+). Stella core is dynamic.
```

I knew it had double wide pixels, I've never thought about the vertical resolution.

wescopeland: I added it because technically it did have analog TV out natively (by default).

But resolution output also does vary per game! I figured it may be very surprising to see a `0` on that field in the DB and someone might think it's actually a mistake.

Here's how I handled this case in the front-end (with respect to scaling):
```ts
// CRT systems displayed non-square pixels stretched to 4:3. Only apply
// aspect ratio scaling when we also have known resolutions, otherwise we
// can't predict the image dimensions (eg: Atari 2600, Arcade).
const aspectRatio = hasAnalogTvOutput && expectedWidth ? 4 / 3 : undefined;
```

So Atari 2600 won't get the scaling due to its dynamic resolutions, but we can still flag it accurately in the DB for this analog-out field.

app/Filament/Resources/GameResource/Pages/Media.php
Jamiras: Now, when uploading both, there's a weird issue where the image name is captured for the in-game image, but not for the title.

Image

I also see this when just uploading the in-game image. I guess it's because the field is a list (even though I can't upload more)?

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
```diff
@@ -0,0 +1,414 @@
...
+
+            // Sony consoles
+            12,  // PlayStation
+            21,  // PlayStation 2
+
+            // Microsoft consoles
+            22,  // Xbox
+
+            // Atari consoles
+            25,  // Atari 2600
```
Jamiras, line 364: If I upload this image to an Atari 2600 game,

Image

The popup seems to be upscaling a downscaled image (note the subpixel artifacts)

Image

vs

Image

I'm not sure if that's due to the css scaling, or some processing applied during the upload.

Jamiras: Additionally, an image captured at native resolution 

Image

is not being stretched to 4:3.

Image

It is, however, being scaled for the popup without any subpixel artifacts.

app/Filament/Resources/GameResource/Pages/Media.php
wescopeland: This is a side effect of one asset using the media library whereas the other isn't. This is a temporary issue that should be fixed in phase 2.

database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
wescopeland: The subpixel artifacts were caused by using the wrong scaling algorithm in the back-end. Should now be fixed in latest:
Screenshot 2026-03-02 at 5 35 09 PM

The Atari 2600 4:3 issue should now be fixed as well:
Screenshot 2026-03-02 at 5 45 14 PM

Jamiras: The native resolution screenshot (left) is 160x228 and isn't being scaled to fit the widget. The captured image (right) is 822x864. Both are sized to 1024x768 correctly in the popup.

image

wescopeland: My understanding is this is existing behavior on prod, ie https://retroachievements.org/game/3818:
Screenshot 2026-03-05 at 5 31 53 PM

These both render as 160x144. On prod today, small native-res images render at their natural size. I'm not sure we should change this. I could also be misunderstanding the issue.

Jamiras: I know this happens on prod currently. Since we're changing the behavior such that uploading native resolution images is the preferred format, I think we should try to do something so images better fill the widget. I think that implies scaling the images in the widget so they have some minimum size.

My comment is mostly reporting that having two differently sized images seems wrong. The smaller should be enlarged to match the larger.

wescopeland: I agree with you.

I'm feeling a bit stuck here because CSS genuinely cannot conditionally size one element based on a sibling's content dimensions. There is no CSS mechanism for "be the same width as your neighbor."

Our constraint is we want images at natural sizes when they're the same dimensions, smaller images should match the larger size when the sizes don't match, and we want a max width of 320px on these thumbnails. We also don't want to do this with JS.

Every fix approach I've tried (normalize widths through a grid layout, force equal 320px widths, etc) breaks something: non-integer scaling ruins pixel art, portrait systems like DS have their images warped, TATE arcade games have their images warped, etc.

This is very likely actually fixable with relative ease during Phase 3. A huge challenge right now is only one of the screenshot types is stored in MediaLibrary for this PR. By Phase 3, both will be, and we'll have backfilled all our legacy stuff. MediaLibrary stores the width and height on every media record at upload/create time, so the front-end theoretically can know both images' actual dimensions and could set a shared width without relying on CSS sorcery.

Jamiras: I'm willing to delay this functionality as it's unlikely a user will capture images at different resolutions for initial uploads, and I'm pretty sure we already have cases where this is occurring.

PR #4587, opened 2026-02-22
fix(UpdatePlayerGameMetricsAction): use firstOrCreate to prevent race condition duplicate inserts

wescopeland:
Resolves:
Screenshot 2026-02-22 at 11 44 09 AM

**Root Cause**
Two concurrent `UpdatePlayerGameMetricsJob` instances for the same user+game can both query for an existing `PlayerAchievementSet`. Both find nothing, both try to INSERT, and the second one hits a unique constraint.

We're basically hand-rolling a `first()` + `new` + `save()` pattern. This PR replaces that pattern with a `firstOrCreate()` which should handle the operation atomically.


app/Platform/Actions/UpdatePlayerGameMetricsAction.php
```diff
@@ -81,33 +81,35 @@ public function execute(PlayerGame $playerGame, bool $silent = false): void
...
+                    ->where('achievement_set_id', $achievementSet->id)
+                    ->exists();
 
-            $setAchievementsUnlocked = $achievementsUnlocked->filter(fn (Achievement $achievement) => in_array($achievement->id, $setAchievementIds));
-            if ($setAchievementsUnlocked->count() === 0 && $achievementSet !== $coreAchievementSet && !$playerAchievementSet->exists) {
-                // no unlocks for the subset and no existing metrics, don't create a player_achievement_set
-                continue;
+                if (!$hasExistingMetrics) {
+                    continue;
+                }
```
Jamiras, line 100: Suggest just fetching this here to avoid an extra lookup in a few lines:
```
if ($isUntouchedSubset) {
    $playerAchievementSet = PlayerAchievementSet...first();
    if (!$playerAcheivementSet) {
        continue;
    }
} else {
    $playerAchievementSet = PlayerAchievementSet::firstOrCreate(...);
}

wescopeland: Made this change in latest.

PR #4589, opened 2026-02-22
feat(achievement): add Recent Unlocks component

wescopeland:
This PR populates the "Recent Unlocks" tab in `achievement2`. The implementation is fairly straightforward and almost identical to the existing PHP UI with a few subtle changes:
* If it's a _very_ recent unlock (within the day), a DiffTimestamp is shown.
* For performance, the rows themselves are deferred to a client-side fetch rather than being eagerly put into server-side props.

http://localhost:64000/achievement2/9?tab=unlocks

Screenshot 2026-02-22 at 4 30 06 PM

Screenshot 2026-02-22 at 4 30 18 PM


resources/js/features/achievements/components/AchievementRecentUnlocks/AchievementRecentUnlocks.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useFormatDate } from '@/common/hooks/useFormatDate';
+import { usePageProps } from '@/common/hooks/usePageProps';
+
+export const AchievementRecentUnlocks: FC = () => {
+  const { achievement, recentUnlocks } = usePageProps<App.Platform.Data.AchievementShowPageProps>();
+
+  const { t } = useTranslation();
+  const { formatDate } = useFormatDate();
+
+  const placeholderRowCount = Math.min(achievement.unlocksTotal ?? 0, 50);
```
Jamiras, line 24: I wonder if it would make sense to restrict this to a smaller number (maybe 10?). I'm not sure how players would use this information other than to ask a recent earner how they did it.

Conversely, it would be nice to have a "more..." link.

wescopeland: I had a similar thought - I kept it at 50 because I wasn't sure if there was some downstream consumer of this data (ie: events) that we're not aware of.

Some sort of more link does make sense (or maybe pagination of the existing list). I'll add this to my notes.

PR #4592, opened 2026-02-24
chore: suppress doupload.php GD image driver warnings

wescopeland:
Screenshot 2026-02-23 at 7 51 56 PM

These error logs are actually caused by network issues between the server and client. The GD image driver is emitting PHP errors when it tries to process corrupt image data, even though our code gracefully handles that already.


app/Helpers/util/media.php
Jamiras, line 1: Won't this just change the error being logged? If we suppress the actual error, the code should proceed to the `throw new Exception('Cannot create image')`, which seems less useful, and just as likely to fill the logs.

PR #4593, opened 2026-02-24
feat(api): add V2 JSON:API relationship endpoint for GameHash resources

wescopeland:
This PR adds a V2 API relationship endpoint for `GameHash` resources, accessible via `GET /games/{game}/hashes`. This provides the functionality of V1's `API_GetGameHashes` endpoint. Things worth noting:

* Hashes are accessed exclusively via the relationship route. There is (currently) no standalone index, but one may be desirable later.
* The `labels` attribute is parsed from its stored comma-separated string into a JSON array.
* All compatibility states are returned by default. The `filter[compatibility]` param supports comma-separated values to filter by one or more states (eg: compatible, incompatible, untested, patch-required).
* Soft-deleted hashes are automatically excluded.
* The `game` relationship can be included via `?include=game`.

**How to test this PR**
```bash
# Get paginated hashes for a game (50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/games/1/hashes \

# Get only compatible hashes (matches V1 API behavior)
curl --request GET \
  --url 'http://localhost:64000/api/v2/games/1/hashes?filter[compatibility]=compatible' \

# Get compatible and incompatible hashes
curl --request GET \
  --url 'http://localhost:64000/api/v2/games/1/hashes?filter[compatibility]=compatible,incompatible' \

# Get hashes with game data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/games/1/hashes?include=game' \

# Custom pagination (page 2, 25 per page)
curl --request GET \
  --url 'http://localhost:64000/api/v2/games/1/hashes?page[number]=2&page[size]=25' \
```


app/Api/V2/GameHashes/GameHashResource.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    /**
+     * Get the resource's attributes.
+     *
+     * @param Request|null $request
+     */
+    public function attributes($request): iterable
+    {
+        return [
+            'md5' => $this->resource->md5,
```
Jamiras: I wonder if we should rename this to avoid confusion against the fact that many systems aren't full-file md5s. And we may want this field available in the future if we ever populate and expose `file_md5`.

app/Api/V2/GameHashes/GameHashSchema.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Get the resource filters.
+     */
+    public function filters(): array
+    {
+        return [
+            WhereIdIn::make($this),
+            WhereIn::make('compatibility', 'compatibility')->delimiter(','),
```
Jamiras, line 66: Minor: this doesn't error if an invalid value is provided.
```
curl --header 'X-API-Key: ...' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/games/1446/hashes?filter\[compatibility\]=all"
```

app/Api/V2/GameHashes/GameHashResource.php
wescopeland: The best thing I can come up with is `raMd5`. I went with that in latest.

app/Api/V2/GameHashes/GameHashSchema.php
wescopeland: Will now error in latest. Needed to be fixed for achievement `filter[type]` as well. Added a few new test cases to verify the behavior.

app/Api/V2/GameHashes/GameHashResource.php
Jamiras: I'd probably eliminate all references to "MD5". Maybe `raHash` or just `hash`.

wescopeland: I considered that, but it's duplicative with our existing `hash` column in the table. It also may paint us into a corner if we ever support more than md5 as a hashing algorithm.

Jamiras: `hash` is currently `NULL` for all entries in the table. I'm confused what that field would be used for if not the RA hash. I'm also uncertain what `type` and `file_name_md5` would be used for. Similarly, what's the difference between `crc` and `file_crc` or `sha1` and `file_sha1`?

To me, it seems like the `XXX` and `file_XXX` columns are redundant, and `md5` should be moved into `hash`.

wescopeland: I agree on the confusion on `file_name_md5`. I'm also unsure about `file_crc`, `file_sha1`, and `type`.

Where I'm stuck is: what if we have multiple hashes for a single file at some point? We already know `md5` has performance issues on larger files. I'm not sure we'd want to store SHA1 or CRC hashes for files, but I can  imagine a far-off future where we store xxh128 or Blake-3 hashes, especially for systems like PS2 (or PS3).

Jamiras: I think the `file_` hashes are for users to validate their dumps without relying on our partial hashing logic. We wouldn't be generating them every time a user loads a game.

wescopeland: Agreed - I'm still puzzled by `file_name_md5` though 🤔 

The only thing I can imagine is it's an MD5 hash of a standardized filename string itself? Maybe from a No-Intro import?

wescopeland: I'm not sure how to move forward.

`hash` doesn't seem like a good field name because it may imply we only will ever support one hash or hashing mechanism.

Jamiras: While I would prefer something more generic (like `raHash` - as in the values generated by `RAHasher`), I'm not against `raMD5`. I just want to ensure we distance ourselves as much from the naive whole-file hashing that users may try to match against.

And we should ensure that whatever is picked is something we're happy with going forward. We want this to be as future-proof as possible.

PR #4595, opened 2026-02-25
fix(user): remediate pluralization issue with event award tooltips

wescopeland:
Resolves this pluralization issue:
Screenshot 2026-02-25 at 8 39 43 AM


app/Helpers/render/site-award.php
```diff
@@ -280,7 +280,8 @@ function RenderAward(
                     $tooltipTitle = "{$event->title} - {$tierLabel}";
                 }
 
-                $tooltipDescription = "Awarded for earning at least {$actualEventAward->points_required} points";
+                $pointsLabel = $actualEventAward->points_required === 1 ? 'point' : 'points';
```
Jamiras: Wouldn't it be better/safer to use `Str::plural`?

Jamiras (approved):

PR #4611, opened 2026-03-01
feat(achievement): add embed url support

wescopeland:
http://localhost:64000/achievement2/24?tab=video

This PR adds `embed_url` support to achievement2. The implementation is fairly straightforward. If an achievement has an `embed_url` value, a new "Video" tab appears:

Screenshot 2026-03-01 at 10 04 52 AM

### DIFF
--- a/app/Platform/Enums/AchievementPageTab.php
+++ b/app/Platform/Enums/AchievementPageTab.php
@@ -13,7 +13,4 @@ enum AchievementPageTab: string
     case Comments = 'comments';
     case Tips = 'tips';
     case Unlocks = 'unlocks';
-
-    /** TODO this will eventually be rolled into 'Tips' */
-    case Video = 'video';
 }
--- a/resources/js/features/achievements/components/+show/AchievementShowRoot.tsx
+++ b/resources/js/features/achievements/components/+show/AchievementShowRoot.tsx
@@ -24,7 +24,7 @@ export const AchievementShowRoot: FC = () => {
 
   const tabConfigs: TabConfig[] = [
     { value: 'comments', label: t('Comments') },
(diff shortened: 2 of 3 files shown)


resources/js/features/achievements/components/+show/AchievementShowRoot.tsx
```diff
@@ -56,6 +63,12 @@ export const AchievementShowRoot: FC = () => {
               <AchievementCommentList />
             </BaseTabsContent>
 
+            {achievement.embedUrl ? (
+              <BaseTabsContent value="video">
+                <VideoEmbed src={achievement.embedUrl} />
+              </BaseTabsContent>
+            ) : null}
```
Jamiras, line 71: The manage page restricts URLs now, but we have a decent amount of image/twitch links, neither of which work with this implemention.

```
select id, embed_url from achievements where embed_url is not null and embed_url != '' and embed_url not like '%youtu%';
```

The old page correctly embeds the images, and creates links to the twitch videos.

app/Platform/Enums/AchievementPageTab.php
```diff
@@ -13,4 +13,7 @@ enum AchievementPageTab: string
     case Comments = 'comments';
     case Tips = 'tips';
     case Unlocks = 'unlocks';
+
+    /** TODO this will eventually be rolled into 'Tips' */
+    case Video = 'video';
```
Jamiras: Should this be changed to Media as well? It's a bit disjointed to see `?tab=video` in the URL.

wescopeland: I think I'm going to drop `case Video` altogether and replace it with `case Tips`. It's still a bit disjointed, but given that's the direction this is likely headed, it'll be a bit cleaner later.

PR #4613, opened 2026-03-01
ci: use faster node.js tooling

wescopeland:
I like fast tools.

The Vite ecosystem is rapidly moving towards a world where developer tools are written in Rust or Go instead of JavaScript. This PR:
* Swaps `eslint-prettier-plugin` for `oxfmt`, a Rust equivalent made by the Vite team that's ~20x faster.
* Swaps `eslint` itself for `oxlint`, a Rust equivalent made by the Vite team that's ~20x faster with most plugins we use natively supported as part of the package itself. This has the positive effect of reducing a large number of direct dependencies in our _package.json_ file.
* Swaps `typescript` with `tsgo`, a TypeScript toolchain developed by Microsoft which is written in Go. It's still in preview, but is stable enough to use with even more thorough type analysis than the latest version of TypeScript (TS 6). This toolchain will eventually become TypeScript 7. There's no reason for us not to adopt this now.
* Sets all lint warnings to instead be errors. Warnings are noise that often require a subjective decision to be made. There will only be errors now.
* Includes a _.vscode/extensions.json_ and _.vscode/settings.json_ with opinionated sane defaults for VSCode.

In a few months, Vite itself will be switching to Rust. This will improve `pnpm build` and `pnpm test` runtimes. The Rust version, Vite 8, is currently in beta, but it's not yet stable enough for us to make the move.

Here's one sample benchmark I ran locally when TypeScript 5 and TypeScript 7 were both installed simultaneously:
```
❯ hyperfine --warmup 1 'pnpm tsc' 'pnpm tsgo'
Benchmark 1: pnpm tsc
  Time (mean ± σ):      8.130 s ±  0.199 s    [User: 12.379 s, System: 0.644 s]
  Range (min … max):    7.886 s …  8.462 s    10 runs

Benchmark 2: pnpm tsgo
  Time (mean ± σ):      1.211 s ±  0.015 s    [User: 5.271 s, System: 0.919 s]
  Range (min … max):    1.191 s …  1.243 s    10 runs

Summary
  pnpm tsgo ran
    6.71 ± 0.18 times faster than pnpm tsc
```

**How to test this PR**
```bash
sail pnpm install # or pnpm install

sail pnpm format:check
sail pnpm lint
sail pnpm tsc
```

There's also a new `pnpm fix` command that will auto-format and auto-fix any fixable linting issues.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -223,6 +223,9 @@ importers:
       autoprefixer:
         specifier: ^10.4.14
         version: 10.4.20(postcss@8.5.4)
+      babel-plugin-react-compiler:
+        specifier: ^1.0.0
+        version: 1.0.0
       chokidar:
         specifier: ^4.0.1
         version: 4.0.1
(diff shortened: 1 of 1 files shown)


Jamiras (approved):
Had an error the first time I ran `pnpm run build`:
```
Error: Loading PostCSS Plugin failed: Cannot find module 'postcss-import'
Require stack:
- /home/jamiras/source/RAWeb/postcss.config.js
```
Could not reproduce on subsequent runs.

PR #4616, opened 2026-03-01
perf(api): add SWR caching to the V1 API

wescopeland:
The V1 API handles ~1.66M requests per day and generates ~101M DB queries per week.

This PR adds `Cache::flexible()` to several endpoints. Several of the endpoints have common enough data to share the same caches. The following endpoints are affected:
* API_GetConsoleIDs - standalone cache, 4 possible keys
* API_GetGame - uses a shared `base-data` cache
* API_GetGameExtended - uses a shared `base-data` cache + `achievements` caches
* API_GetGameInfoAndUserProgress - uses a shared `base-data` cache + `achievements` caches
* API_GetGameList - standalone per-console cache, filtering and pagination are now applied in PHP

`GetGame`, `GetGameExtended`, and `GetGameInfoAndUserProgress` all share the same per-game `base-data` cache key, so warming it from any endpoint benefits the others.

`GetGameExtended` and `GetGameInfoAndUserProgress` also share the `achievements:promoted` cache.

User-specific data is still fetched live.

### DIFF
--- a/public/API/API_GetConsoleIDs.php
+++ b/public/API/API_GetConsoleIDs.php
@@ -18,8 +18,8 @@ use Illuminate\Support\Facades\Cache;
  *    bool      IsGameSystem        indicates if the system is a game system (not Events, Hubs, etc.)
  */
 
-$onlyActive = (int) (bool) request()->query('a', '0');
-$onlyGameConsoles = (int) (bool) request()->query('g', '0');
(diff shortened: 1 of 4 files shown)


wescopeland:
After doing a bunch of benchmarking, I don't think this will actually help much.

PR #4619, opened 2026-03-04
Add clarification to UWC report alert

Hexadigital:
DevCompliance gets a lot of Unwelcome Concept reports from players who don't think it is an unwelcome concept, but would like to see the achievement award more points, or be made easier.

I'd like to adjust the alert that appears when making an UWC report to clarify that those changes should be suggested in the forum thread, rather than to DevComp, as we can only reject those.

### DIFF
--- a/resources/views/components/message/template-kind-alert.blade.php
+++ b/resources/views/components/message/template-kind-alert.blade.php
@@ -24,6 +24,10 @@
                             consult the docs.
                         </a>
                     </p>
+
+                    <p>
+                        Please note that DevCompliance does not rescore or rebalance achievements. If the concept is fine but the difficulty or reward is not, please suggest a change in the forum thread instead.
+                    </p>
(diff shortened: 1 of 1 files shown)


resources/views/components/message/template-kind-alert.blade.php
wescopeland, line 1: It actually appears that this is dead code I forgot to clean up.

_TemplateKindAlert.tsx_ is where this content is surfaced.

Hexadigital: Thank you! I got side tracked, but I'll take a look at putting this in the proper place tomorrow.

Hexadigital: Nevermind, I don't feel comfortable editing that file - it's not a big deal though, I'll just close this out.

PR #4620, opened 2026-03-05
fix: always allow artists to edit their fields

Chew:
This changes the check for updating a game from globally blocking anyone who is a Jr. Dev from editing the game page at all, to still allowing Artists the ability to edit the things they can edit regardless.

Paths checked:

- User is only a Jr. Dev - Cannot view edit page at all
- User is a Jr. Dev and Artist - Can view the edit page, but can only edit banner and icon
- User is a Jr. Dev with a claim - Can edit the game page as expected

I'm not super familiar with the Filament area and its permissions, but I got this far and the three checks above worked. Let me know if I have overlooked anything important. As is important with permissions, of course.


app/Policies/GamePolicy.php
wescopeland, line 1: DX: Recommend matching the coding style in `AchievementPolicy::updateField()` exactly. Ideally, the two policies should look identical in structure. No need for any abstraction yet because we only do this in two places, but we can probably remove the stray line of whitespace on line 139, the comment on line 140, and match the comment on line 145.

Chew: I can reorganize a bit, some stuff is specific to the achievement page, though.

For the permissions, Dev can do everything the Jr. Dev can do, plus edit the title. Should I compress it like this?
```php
$developerBasePermissions = [
    'banner',
    'legacy_guide_url',
    'developer',
    'publisher',
    'genre',
    'image_icon_asset_path',
    'image_box_art_asset_path',
    'image_title_asset_path',
    'image_ingame_asset_path',
    'released_at',
    'released_at_granularity',
    'trigger_definition',
];

$roleFieldPermissions = [
    Role::DEVELOPER_JUNIOR => $developerBasePermissions,

    Role::DEVELOPER => [
        ...$developerBasePermissions,
        'title',
    ],

    Role::ARTIST => [
        'banner',
        'image_icon_asset_path',
    ],
];
```

Or is it better with them being listed out individually?

This could also happen on Achievement but there are a lot less permissions there

wescopeland: Good question. I kinda like having these listed individually, but it's a soft preference.

app/Policies/GamePolicy.php
```diff
@@ -131,22 +126,32 @@ public function updateField(User $user, Game $game, string $fieldName): bool
...
         $allowedFieldsForUser = collect($roleFieldPermissions)
-            ->filter(function ($fields, $role) use ($userRoles) {
-                return $userRoles->contains($role);
+            ->filter(function ($fields, $role) use ($userRoles, $user, $game) {
+                if (!$userRoles->contains($role)) {
+                    return false;
+                }
+
+                // Junior Developers have additional specific criteria that must be satisfied
+                // before they are allowed to edit achievement fields.
```
wescopeland: nit:

```suggestion
                // before they are allowed to edit game fields.
```

wescopeland (approved):
Thanks!

Jamiras (approved):

PR #4629, opened 2026-03-07
feat(achievement): add management functionality

wescopeland:
This PR adds management functionality to achievement2.

On desktop, this is facilitated by a new "Contribute" area in the sidebar:
Screenshot 2026-03-07 at 3 03 01 PM

Three buttons are available:
* Quick Edit
* Manage
* Logic

Manage and Logic are fairly self-explanatory: they link to those pages of the `AchievementResource` in Filament.

Quick Edit is where most of the meat in this PR is. Though I am generally opposed to having management functionality in the player-facing site, I've made an exception here due to this being such a common user journey and very popular request.

When Quick Edit mode is enabled, the user can modify title, description, points, and type in-place. This is permission aware, so Developers, Junior Developers, and Writers should only be able to touch what they're permissioned to touch. Additionally, a button appears for Demote / Promote, based on the achievement's current promoted status.

This is where a firm line is drawn. More advanced management functionality for achievements, such as badge uploading and embed URL setting, remains in the management app.

https://github.com/user-attachments/assets/6e4ad5e1-9fc5-4fef-9c03-37f0266223c2

### DIFF
--- a/app/Http/Middleware/EncryptCookies.php
+++ b/app/Http/Middleware/EncryptCookies.php
@@ -32,6 +32,7 @@ class EncryptCookies extends Middleware
         'prefers_hidden_user_profile_stats',
         'prefers_seeing_saved_hidden_rows_when_reordering',
         'progression_status_widths_preference',
+        'prefers_expanded_rich_presence',
         'retroachievements_authenticated',
     ];
 }
--- a/app/Platform/Actions/BuildGameShowPagePropsAction.php
+++ b/app/Platform/Actions/BuildGameShowPagePropsAction.php
@@ -325,6 +325,7 @@ class BuildGameShowPagePropsAction
             isSubscribedToTickets: $user ? $subscriptionService->isSubscribed($user, SubscriptionSubjectType::GameTickets, $backingGame->id) : false,
             isLockedOnlyFilterEnabled: $isLockedOnlyFilterEnabled,
             isMissableOnlyFilterEnabled: $isMissableOnlyFilterEnabled,
+            isRichPresenceExpanded: Cookie::get('prefers_expanded_rich_presence') === 'true',
             isViewingPublishedAchievements: $isPromoted,
             followedPlayerCompletions: $this->buildFollowedPlayerCompletionAction->execute($user, $backingGame),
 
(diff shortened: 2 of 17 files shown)


resources/js/features/achievements/components/AchievementInlineActions/AchievementInlineActions.tsx
```diff
@@ -37,17 +49,70 @@ export const AchievementInlineActions: FC = () => {
...
+              <LuX className="size-3" />
+              {t('Cancel')}
+            </BaseButton>
+
+            <BaseButton onClick={handleSave} size="xs" className="gap-1" disabled={isSaving}>
+              <LuCheck className="size-3" />
+              {t('Save')}
+            </BaseButton>
+          </div>
+        ) : null}
```
Jamiras, line 94: Can we get a Cancel in here too? I usually look for a Cancel next to the Save button.

resources/js/features/achievements/components/AchievementInlineActions/AchievementInlineActions.tsx
```diff
@@ -37,17 +49,70 @@ export const AchievementInlineActions: FC = () => {
...
+                href={`/manage/achievements/${achievement.id}/logic`}
+                target="_blank"
+                className="flex items-center gap-1"
+              >
+                <LuCode className="size-3" />
+                {t('Logic')}
+              </a>
+            ) : null}
+
+            <BaseButton onClick={() => setIsEditMode(true)} size="xs" className="gap-1">
```
Jamiras, line 75: I find the switch to edit mode too subtle. Would it be reasonable to show the borders immediately instead of on hover of each field?

Or maybe change the background of the entire main section to some other color?

app/Data/UserPermissionsData.php
Jamiras, line 1: I forget which roles are required to view the achievement2 page, so was unable to test field access

resources/js/features/achievements/components/AchievementInlineActions/AchievementInlineActions.tsx
wescopeland: Not a problem - added it in latest.

resources/js/features/achievements/components/AchievementInlineActions/AchievementInlineActions.tsx
wescopeland: Yea, if you find it too subtle then it's an issue. The design should be understandable.

In latest, I've made the borders persistent instead of only showing them on hover.

PR #4633, opened 2026-03-08
feat(api): add V2 API support for PlayerAchievement resources

wescopeland:
This PR adds V2 API support for `PlayerAchievement` resources through several relationship endpoints.

The resources are accessible by two means:
1. From the user side, ie: `GET /users/{user}/player-achievements`.
2. From the achievement side, ie: `GET /achievements/{achievement}/player-achievements`.

Things worth noting:
1. Player achievements are accessed exclusively via relationship routes. There is no standalone index.
2. Hub and event game achievements are excluded from the results via `System::getNonGameSystems()`.
3. Results are default sorted by `-unlockedAt` (most recent unlocks first).
4. When accessed from the achievement side, unranked users are excluded from the results. This filtering is not applied from the user side.
5. The `achievement`, `game`, and `user` relationships can be included via `?include=`.
6. Several filters are supported: `filter[gameId]`, `filter[achievementSetId]`, `filter[unlockedFrom]`, and `filter[unlockedTo]`.

**How to test this PR**
```bash
# Get a paginated list of player achievements for a user (50 per page)
curl --request GET \
  --url http://localhost:64000/api/v2/users/Scott/player-achievements \

# Get a user's most recent achievement unlocks with achievement data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievements?include=achievement' \

# Get a user's most recent achievement unlocks with game data included
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievements?include=game' \

# Get a user's list of achievement unlocks by game ID
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievements?filter[gameId]=1' \

# Get a user's list of achievement unlocks by achievement set ID
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievements?filter[achievementSetId]=1' \

# Filter a user's list of achievement unlocks by an inclusive date range
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievements?filter[unlockedFrom]=2024-01-01&filter[unlockedTo]=2024-12-31' \

# Get players who unlocked a specific achievement
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievements/9/player-achievements?include=user&fields[users]=displayName' \

# Apply custom pagination
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/player-achievements?page[number]=2&page[size]=25' \
```

### DIFF
--- a/app/Console/Commands/UpdateGameActivitySnapshots.php
+++ b/app/Console/Commands/UpdateGameActivitySnapshots.php
@@ -135,11 +135,10 @@ class UpdateGameActivitySnapshots extends Command
               AND g.achievements_published > 0
               AND g.players_total >= 15
               AND EXISTS (
-                  SELECT 1 FROM achievement_set_claims claims
-                  WHERE claims.game_id = metrics.game_id
-                    AND claims.status = 'complete'
-                    AND claims.set_type = 'new_set'
-                    AND claims.finished_at <= '{$targetDateStr}' - INTERVAL 48 HOUR
+                  SELECT 1 FROM game_achievement_sets gas
+                  JOIN achievement_sets asets ON asets.id = gas.achievement_set_id
+                  WHERE gas.game_id = metrics.game_id
+                    AND asets.achievements_first_published_at <= '{$targetDateStr}' - INTERVAL 48 HOUR
               )
               AND NOT EXISTS (
                   SELECT 1 FROM game_set_games gsg
--- a/docs/UBUNTU24_DESKTOP_VM_SETUP.md
+++ b/docs/UBUNTU24_DESKTOP_VM_SETUP.md
@@ -236,9 +236,12 @@ You should be able to access the site now via (http://localhost:64000)
 - If the port forwarding is set up correctly, you should also be able to access the site using the same URL from the host machine.
 
 #### Note about running tests
-Something about the way sail/docker manages the RAWeb instance causes the log files to be owned by sail and any tests that try to write to the log file will fail with a permissions error. To fix permissions for the tests, run the following command:
+Something about the way sail/docker manages the RAWeb instance causes the log files to be owned by sail and any tests that try to write to the log file will fail with a permissions error. 
+A similar thing also happens with temporary image uploads.
+To fix permissions for the tests, run the following command:
 ```
 sudo chmod 666 ~/source/RAWeb/storage/logs/*.log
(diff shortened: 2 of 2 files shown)


app/Api/V2/PlayerAchievements/PlayerAchievementResource.php
```diff
@@ -0,0 +1,52 @@
...
+     * Get the resource's relationships.
+     *
+     * @param Request|null $request
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'achievement' => $this->relation('achievement')->withoutLinks(),
+            'game' => $this->relation('game')->withoutLinks(),
+            'user' => $this->relation('user')->withoutLinks(),
```
Jamiras: Unless specifically asking for the related achievements, there's no way to tell which achievement an unlock is associated to:
```
      {
         "type":"player-achievements",
         "id":"71484928",
         "attributes":{
            "unlockedAt":"2024-10-14T00:00:38.000000Z",
            "unlockedHardcoreAt":"2024-10-14T00:00:38.000000Z"
         }
      }
```

See [related](https://github.com/RetroAchievements/RAWeb/pull/4562#discussion_r2817899421).

Jamiras: The opposite is true when querying the data through the achievement, where the associated user is unknown unless the relationship is explicitly requested, so I'm not sure there's a clean solution to address both.

wescopeland: Thanks - I've updated latest so these relationships are included in both responses. It looks something like this:

Screenshot 2026-03-25 at 5 59 28 PM

Jamiras: This is definitely preferable to not having the data, but is there some way to only show the secondary relationship? When querying a user's unlock, the user field is going to be the same on every record. Similarly, when querying an achievement's unlock, the achievement field is going to be the same on every record.

wescopeland: Yea, it's fairly straightforward and still spec-compliant. I've made this change in latest.

PR #4639, opened 2026-03-10
feat(game): add screenshot backfill command, move title image to media

wescopeland:
This PR kicks off Phase 2 of the game screenshot to MediaLibrary migration. There will be one or two more PRs in Phase 2 before it's complete.

Two things are accomplished here:
1. Title screenshot uploads are now routed through `AddGameScreenshotAction` in Filament, just like ingame screenshots already are as a result of #4585.
2. A backfill command has been added to migrate all existing legacy screenshots into MediaLibrary, closing the loop on the double-writes.

The backfill is idempotent - it's safe to re-run without creating duplicate assets.

The React front-end is unaffected by this PR. That is still forthcoming.

**How to test this PR:**
Title image uploads should still work in the Media tab for `GameResource` in Filament, but you should now also see `game_screenshots` and `media` rows being created for them.

For the backfill command, you'll need local screenshot media in _storage/app/media/Images/_ associated with a game's record. Then:
```
sail artisan migrate

# restart horizon first
sail artisan horizon

sail artisan ra:platform:game:backfill-screenshots 1 # this is the id
```
Without an ID, the command queues jobs on the `default` queue for every non-backfilled game. This is CPU-intensive due to image conversions that are necessary. I've done a full local sync from all images in prod just to test it out. The server won't have any trouble with it, but I don't recommend trying it locally if you're compute-bottlenecked.

### DIFF
--- a/resources/js/features/achievements/components/+show/AchievementShowRoot.test.tsx
+++ b/resources/js/features/achievements/components/+show/AchievementShowRoot.test.tsx
@@ -11,7 +11,12 @@ import { AchievementShowRoot } from './AchievementShowRoot';
 describe('Component: AchievementShowRoot', () => {
   beforeEach(() => {
     vi.clearAllMocks();
-    vi.spyOn(router, 'reload').mockImplementation(vi.fn());
+
+    vi.spyOn(router, 'reload').mockImplementation((options?: Record<string, unknown>) => {
+      if (options && typeof options.onFinish === 'function') {
+        (options.onFinish as () => void)();
+      }
+    });
 
     window.HTMLElement.prototype.hasPointerCapture = vi.fn();
     window.HTMLElement.prototype.scrollIntoView = vi.fn();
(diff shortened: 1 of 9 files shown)


database/migrations/2026_03_12_000000_update_game_screenshots_table_add_dimensions.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::table('game_screenshots', function (Blueprint $table) {
+            $table->unsignedSmallInteger('width')->nullable()->after('media_id');
+            $table->unsignedSmallInteger('height')->nullable()->after('width');
```
wescopeland, line 14: These both need to be stored as columns on the table, or Filament queries for game_screenshots (coming in a subsequent PR) slow to a crawl.

app/Platform/Actions/BackfillGameScreenshotsAction.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        $hash = sha1($fileContents);
+
+        // Check idempotency, and bail if this type already has this exact image.
+        // Scoped by type so identical content used for both title and ingame still creates both records.
+        $alreadyExists = $game->gameScreenshots()
+            ->ofType($type)
+            ->whereHas('media', fn ($q) => $q->where('custom_properties->sha1', $hash))
+            ->exists();
```
Jamiras: Probably not an issue for prod, but I had a game with screenshots from the previous PR and ran the backfill command on that game and it generated a duplicate row:
```
+---------+----------+--------+------------+----------+
| game_id | media_id | type   | is_primary | status   |
+---------+----------+--------+------------+----------+
|   25623 |     1626 | ingame |          1 | approved |
|   25623 |     1630 | ingame |          1 | approved |
|   25623 |     1631 | title  |          1 | approved |
+---------+----------+--------+------------+----------+
```

wescopeland: No harm in addressing this now. The SHA1 check is probably too narrow. In latest, the idempotency check will look for existing screenshots of that type instead of matching on the hash. This should still achieve the desired outcome.

PR #4645, opened 2026-03-13
chore: prepare for FrankenPHP + Laravel Octane

wescopeland:
https://discord.com/channels/476211979464343552/718091467893243937/1482131064334848051

I've already verified this is working on stage. On stage, the entire Connect API is being served by Octane with a php-fpm fallback, whereas the rest of the site is still being served through php-fpm. This is handled elegantly through nginx request routing.

**How to test this PR**
```
# rebuild your sail image
sail build --no-cache

sail composer setup
sail up

# should return 'Octane'
curl http://localhost:64008/octane

# now hit the Connect API through port 64008
curl -H "User-Agent: RetroArch/1.22.1" "http://localhost:64008/dorequest.php?r=ping&..."

# you should still be able to navigate the site on localhost:64000
# localhost:64008 does serve the site, but that can be ignored.
```

**php-fpm ?r=ping**
Screenshot 2026-03-13 at 7 20 10 PM

**Octane ?r=ping**
Screenshot 2026-03-13 at 7 23 10 PM


.gitignore
Jamiras, line 1: This got added to my .gitignore by some step of the process:
```
+
+**/caddy
+frankenphp-worker.php
```

wescopeland: Looks like `octane:install` auto-appends those if it's running fresh, which is why I didn't see it.

These gitignores are overly-broad, and we've already manually committed everything else `octane:install` does. I've removed it since it's doing more harm than good.

PR #4647, opened 2026-03-14
chore: update systems screenshots resolutions migration

wescopeland:
Hasn't run in prod yet. These changes align with: https://github.com/RetroAchievements/docs/pull/327/changes


database/migrations/2026_02_20_000001_update_systems_table_add_screenshot_resolutions.php
```diff
@@ -258,11 +258,11 @@ public function up(): void
             // Atari ST - RA docs: 320x200.
             36 => [[320, 200]],
 
-            // Amstrad CPC - RA docs: 320x226.
-            37 => [[320, 226]],
+            // Amstrad CPC - RA docs: 192x272.
+            37 => [[192, 272]],
```
Jamiras: [Wikipedia](https://en.wikipedia.org/wiki/Amstrad_CPC#Video) says 160x200, 320x200, or 640x200 depending on the color depth.

It looks like the core provides extra borders by default (see `crop_screen_borders`: https://github.com/libretro/libretro-cap32?tab=readme-ov-file#features-not-covered-in-docs)

I believe the border is present on the original hardware, so we should probably expect to find it in our screenshots. I'm not sure how many 4-bpp or 2-bpp games exist, but we should probably support those resolutions too. Based on doubling the width that would be 384x272 and 768x272.

wescopeland: Thanks - updated in latest.

PR #4649, opened 2026-03-14
feat(achievement): support event achievements on achievement2

wescopeland:
This PR adds event achievement support to achievement2. As far as I know, this is the last piece of missing functionality on the page.

Good URLs to test with:
http://localhost:64000/achievement2/490222
http://localhost:64000/achievement2/493464

Functionality on this page variant is limited. There are no tabs - unlockers are optimistically always shown. Quick Edit and Logic views are not supported. Softcore stats are not displayed. 

Screenshot 2026-03-14 at 1 50 43 PM

Screenshot 2026-03-14 at 1 49 16 PM

Screenshot 2026-03-14 at 1 47 55 PM


resources/js/features/achievements/components/AchievementEventInfo/AchievementEventInfo.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  return (
+    <div className="rounded-lg bg-embed p-1 light:border light:border-neutral-200 light:bg-white">
+      <BaseTable className="overflow-hidden rounded-lg text-2xs">
+        <BaseTableBody>
+          <BaseTableRow>
+            <BaseTableHead scope="row" className="h-auto text-right align-middle text-text">
+              {t('From')}
+            </BaseTableHead>
```
Jamiras, line 39: It took me a while to find this in the sidebar, but I think it probably makes sense there.

One thing I would like to see is either a link directly to the source achievement, or some way to view the comments from the source achievement on this page. 

It might also be worth pulling in "Report an issue" and the "Open tickets" count. Players are really interacting with the source achievement. From their perspective, the only thing that really changes is the unlock state/progress bar and breadcrumb links.

resources/js/features/achievements/components/ProximityAchievements/ProximityAchievements.tsx
```diff
@@ -64,7 +66,9 @@ export const ProximityAchievements: FC = () => {
 
   return (
     <div>
-      <h2 className="mb-0 border-0 text-lg font-semibold">{t('More from this set')}</h2>
+      <h2 className="mb-0 border-0 text-lg font-semibold">
+        {isEventGame ? t('More from this event') : t('More from this set')}
+      </h2>
```
Jamiras, line 71: Recommend prioritizing active achievements here - it seems to just list the first achievements from the event (even if they're no longer active).

Possibly, even limit it just to the active achievements. There's probably very little impetus to look at non-active event achievements outside the context of the main event page.

resources/js/features/achievements/components/AchievementEventInfo/AchievementEventInfo.tsx
wescopeland: Made a bunch of little tweaks in latest:
Screenshot 2026-03-18 at 7 58 31 PM

* With the exception of the Media tab, the tabs have returned. All tabs are pulled from the source except recent unlocks.
* Comment interactions are associated with the source.
* Report an issue and tickets are shown.

resources/js/features/achievements/components/ProximityAchievements/ProximityAchievements.tsx
wescopeland: I've added active achievement filtering in this component. It's handled in the back-end. The filtering will not apply if all achievements are inactive.

Jamiras: The "View all" link should still indicate the set has 60+ achievements:
image

resources/js/features/achievements/components/AchievementEventInfo/AchievementEventInfo.tsx
Jamiras: The Recent Unlocks tab should be explicitly limited to the event achievement. I'm seeing unlocks from before and after the active period, implying that it's showing the source unlocks. If the event achievement is manually unlocked (due to using a flikki) or of the source achievement is reset, the user should still appear in the recent unlocks list.

resources/js/features/achievements/components/AchievementTabsList/AchievementTabsList.test.tsx
Jamiras, line 1: I thought you fixed the rounded corners on the subtabs?

Image

app/Platform/Controllers/AchievementController.php
Jamiras, line 1: Get error accessing non-event achievement2: http://localhost:64000/achievement2/173076

> Attempted to lazy load [eventData] on model [App\Models\Achievement] but lazy loading is disabled.

resources/js/features/achievements/components/AchievementEventInfo/AchievementEventInfo.tsx
wescopeland: I'm very confused by this. It already is explicitly limited to the event achievement. This is clearly happening in production too: https://retroachievements.org/achievement/490222

Screenshot 2026-03-21 at 10 44 48 AM

When I flip the unlock order for debugging purposes, I see unlocks within the time range I'd expect:
Screenshot 2026-03-21 at 10 47 44 AM

Can you give me an event achievement ID where you're seeing unexpected unlocks from before the active period?

Jamiras: I guess it is. The data is just weird:
image

The [Action Man achievement](http://localhost:64000/achievement2/490222?tab=unlocks) you provided as a sample link is particularly bad. All 50 entries are after the achievement end date.

resources/js/features/achievements/components/AchievementTabsList/AchievementTabsList.test.tsx
wescopeland: This UX is unchanged since https://github.com/RetroAchievements/RAWeb/pull/4548. These rounded corners are intentional - do you think this should be changed?

app/Platform/Controllers/AchievementController.php
wescopeland: Thanks, fixed this in latest:
Screenshot 2026-03-21 at 11 00 43 AM

resources/js/features/achievements/components/ProximityAchievements/ProximityAchievements.tsx
wescopeland: Fixed in latest:
Screenshot 2026-03-21 at 11 00 17 AM

resources/js/features/achievements/components/AchievementEventInfo/AchievementEventInfo.tsx
wescopeland: It looks like those are all from manual unlocks ☹️

Screenshot 2026-03-21 at 11 04 57 AM

I don't have a strong sense of what the right answer for this is. It doesn't seem like we should filter those out. On the other hand, it may seem non-sensical unless a user knows what's going on.

Jamiras: > Can you give me an event achievement ID where you're seeing unexpected unlocks from before the active period?

I think I was misreading the year. I agree that the current behavior matches the old achievement page.

I'm going to say things are fine as they are. Most people won't look at achievement unlock history for inactive event achievements, which is where all the confusion came from.

resources/js/features/achievements/components/AchievementTabsList/AchievementTabsList.test.tsx
Jamiras: I might have been thinking of #4607. I typically assume tabs connect to their container, and these buttons behave more like tabs than buttons.

PR #4652, opened 2026-03-15
feat(game): support media restrictions

wescopeland:
https://discord.com/channels/310192285306454017/310195377993416714/1482452813777731765

```
sail artisan migrate
```

Games with media restrictions cannot show any screenshots, box art, or a banner. The filtering has to be done in the back-end so the values still don't leak into props for hydration.

Screenshot_2026-03-14_at_3 15 38_PM


app/Models/Game.php
```diff
@@ -535,17 +537,17 @@ public function getCurrentBannerMediaAttribute(): ?Media
...
     public function getImageBoxArtUrlAttribute(): string
     {
-        return media_asset($this->image_box_art_asset_path);
+        return media_asset($this->is_media_restricted ? self::PLACEHOLDER_IMAGE_PATH : $this->image_box_art_asset_path);
     }
 
     public function getImageTitleUrlAttribute(): string
     {
-        return media_asset($this->image_title_asset_path);
+        return media_asset($this->is_media_restricted ? self::PLACEHOLDER_IMAGE_PATH : $this->image_title_asset_path);
```
Jamiras: This doesn't affect stuff that reads the fields directly (such as API_GetGameExtended)

wescopeland: Made an update so these legacy endpoints will use the placeholder images when a game is under a media restriction. Also added associated tests for the logic.

PR #4656, opened 2026-03-18
feat(seed): populate AotW and dummy site event

Jamiras:
Adds an AotW event and DevJam event to simulate an automatically tracked event with active periods and a manual site event.

image

image

image


database/seeders/EventsSeeder.php
```diff
@@ -0,0 +1,241 @@
...
+                    ->where('created_at', '<', $date)
+                    ->whereNull('unranked_at')
+                    ->whereNotIn('id', array_column($aotwPlayers, 'id'))
+                    ->inRandomOrder()
+                    ->limit($newPlayers)
+                    ->get();
+                foreach ($users as $user) {
+                    if (!in_array($user, $aotwPlayers)) {
+                        $aotwPlayers[] = $user;
+                    }
```
wescopeland: nit: I think this `!in_array($user, $aotwPlayers)` may be dead code. The query is already doing `whereNotIn('id', array_column($aotwPlayers, 'id'))`.

Maybe you could get away with just:
```php
$aotwPlayers = array_merge($aotwPlayers, $users->all());
```

database/seeders/EventsSeeder.php
```diff
@@ -0,0 +1,241 @@
...
+        $eventGame->achievements_published = 0;
+        $eventGame->achievements_unpublished = 0;
+        $eventGame->save();
+
+        $event = Event::create([
+            'legacy_game_id' => $eventGame->id,
+            'image_asset_path' => '/Images/000001.png',
+        ]);
+
+        $achievementCount = 0;
```
wescopeland: nit: I think this line can probably be removed. The `for ($achievementCount = 0; ...)` should already be doing the same thing.

database/seeders/EventsSeeder.php
```diff
@@ -0,0 +1,241 @@
...
+        for ($achievementCount = 0; $achievementCount < $numAchievements; $achievementCount++) {
+            $achievement = Achievement::create([
+                'title' => $prefix ? "$prefix $achievementCount" : "Placeholder",
```
wescopeland: Bug, minor: I suspect this will produce titles like "Week 0" for Achievement of the Week.

You may want something like `$prefix ? "$prefix {$achievementCount + 1}"`.

database/seeders/EventsSeeder.php
```diff
@@ -0,0 +1,241 @@
...
+            if ($date < Carbon::now()) {
+                // pick a random achievement with at least 8 unlocks so there's a decent chance the simulate play will unlock it.
+                $sourceAchievement = Achievement::promoted()->where('unlocks_hardcore', '>', 8)->inRandomOrder()->first();
+                $achievement->title = $sourceAchievement->title;
+                $achievement->image_name = $sourceAchievement->image_name;
+                $achievement->save();
+
+                $achievement->eventData->source_achievement_id = $sourceAchievement->id;
+
+                // between 15 and 40 players will join the first week. each subsequent week, an additional 6-12 will join.
```
wescopeland, line 69: Docs, minor: The sum of `rand(3,8)*4 + rand(2,6)` gives a range of 14-38, not 15-40. I believe this comment may be inaccurate. Or it may be a signal the bell curve being produced by the code isn't the one you're going for. Wanted to flag it either way just in case.

database/seeders/EventsSeeder.php
Jamiras: Yep. I had something like that and misplaced it when I tried to make things more generic.

database/seeders/EventsSeeder.php
Jamiras: done. another leftover from a previous iteration.

database/seeders/EventsSeeder.php
Jamiras: done.

database/seeders/EventsSeeder.php
Jamiras: I was using `rand(2,6)*5` before, but didn't feel like it was generating enough data. I guess I missed one when I updated them.

PR #4696, opened 2026-03-28
feat: add Manual Unlocker role and port admin unlock tool to Filament

Tsearo:
```
sail artisan migrate
```

- Adds a new Manual Unlocker role
- Allows users with the Manual Unlocker role to award achievements similar to what mods/admins can do currently with the Unlock Achievement tool
- Allows users with the Manual Unlocker role view-only access to game and achievement pages in order to review logic when necessary.
- Ports the Unlock Achievement tool to Filament in a new  Unlock Tools page along with some additional validation checks
- Allows for unpromoted achievements to have hardcore unlocks awarded

Adds a new role for Manual Unlocker to provide access to a ported and upgraded version of the Unlock Achievement Admin Tool.

The Role mainly provides access to the new Unlock Tools page under Tools in Filament but also gives read access to the Game and Achievement pages in order for the team to view achievement logic which is at times helpful for review.
image

The page has an Unlock Achievements form and a copy of the Get Game Achievement IDs which is an existing helpful tool under Admin tools that unlockers will benefit from. The main unlock form allows for csv and search fields for both users and achievements depending on preference.
image

All fields are required in order to submit and all fields are validated before processing. 

If everything is proper then a modal window will appear giving an overview of the users, number of achievements, total points and total number of unlock jobs to be queued along with a list of all achievements and their points/typing grouped by set/subset.
image

The following validations are placed on the field level which will prevent the confirmation modal from apearing:
- Invalid users
- Invalid achievements
- Achievements from different sets (or exclusive subsets), base and subsets are allowed

image

Gemini Code Assist used to help with initial development of unlock-achievements.blade.php as well as some testing/bugfixes.


database/migrations/2026_03_23_000000_create_manual_unlocker_role.php
```diff
@@ -0,0 +1,61 @@
...
+
+return new class extends Migration {
+    public function up(): void
+    {
+        // In test environments the seeder creates all roles from config, also don't create if database is not seeded
+        if (app()->environment('testing') || !DB::table('auth_roles')->exists()) {
+            return;
+        }
+
+        $roleId = DB::table('auth_roles')->insertGetId([
```
wescopeland: nit: `$roleId` is unused. Recommend changing `->insertGetId()` to `->insert()`.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
```diff
@@ -0,0 +1,294 @@
...
+                <x-filament::button type="submit">Submit</x-filament::button>
+            </div>
+        </div>
+    </form>
+
+    <x-filament::modal id="confirm-unlock-modal" width="2xl">
+        <x-slot name="heading">Confirm Unlocks</x-slot>
+
+        @php
+            $achievements = !empty($this->validAchievementIds) ? \App\Models\Achievement::with(['game.system'])->whereIn('id', $this->validAchievementIds)->get() : collect();
```
wescopeland: Perf: This re-queries the DB each time the modal renders. `submit()` already fetches these near line 45, but the collection gets thrown away.

I'm wondering if we can do something like storing the loaded achievements (with `game.system` eagerly loaded) as a component property during `submit()` and reference that in the template to reduce DB round-trips.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
```diff
@@ -0,0 +1,294 @@
...
+        $usernames = array_filter(array_unique(array_map('trim', preg_split('/[\s,]+/', $this->usernamesCsv))));
+        $users = User::whereIn('username', $usernames)->get();
+        $foundUsernamesLower = $users->pluck('username')->map(fn($u) => strtolower((string) $u))->toArray();
+
+        $this->missingUsernames = array_values(array_filter($usernames, fn($u) => !in_array(strtolower((string) $u), $foundUsernamesLower)));
+        $this->validUserIds = $users->pluck('id')->toArray();
+        $this->validUsernames = $users->pluck('username')->toArray();
+
+        // Parse Achievements
+        $csvIds = array_filter(array_unique(array_map('trim', preg_split('/[\s,]+/', $this->achievementIdsCsv))), fn($id) => is_numeric($id) && (int) $id > 0);
```
wescopeland: DX: `ParseIdsFromCsvAction` may be reusable here.

app/Community/Enums/ManualUnlockRequestItemState.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function getLabel(): string
+    {
+        return match ($this) {
+            ManualUnlockRequestItemState::Approved => 'Approved',
+            ManualUnlockRequestItemState::Denied => 'Denied',
+            ManualUnlockRequestItemState::Pending => 'Pending',
+            ManualUnlockRequestItemState::Request => 'Request',
+        };
+    }
```
wescopeland: DX, minor: can probably be eliminated in favor of just `$this->name`.

app/Community/Enums/ManualUnlockRequestState.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Used for partially approved requests
+     */
+    case Partial = "partial";
+
+    /**
+     * Used for requests requesting more details
+     */
+    case Requesting = "requesting";
+
```
wescopeland: nit:
```suggestion
```

app/Community/Enums/ManualUnlockRequestState.php
wescopeland, line 1: DX, nit: The PHPDoc on Approved, Denied, and Pending restates the case name - I'd remove those three doc comments.

I'd consider renaming Partial and Requesting so they're self-evident and removing their doc comments too:
```php
// just ideas for names, ideally we can read the code and understand what these
// do without referencing any associated PHPDoc comments

case PartiallyApproved = "partially-approved";
case NeedsMoreDetails = "needs-more-details";
```

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland, line 1: Perf: Recommend adding some validation here. At the moment, there's no cap on how many usernames or achievement IDs can be entered in a single pass. The nested loop in `dispatchJobs()` dispatches `users * achievements` individual jobs. A textarea with 500 usernames and 500 achievement IDs = 250k dispatched jobs.

It may be worth adding a `maxLength` or custom validation rule to the form fields to bound this.

app/Filament/Pages/UnlockTools.php
```diff
@@ -0,0 +1,41 @@
...
+
+class UnlockTools extends Page
+{
+    protected static ?int $navigationSort = 1;
+
+    protected static string|BackedEnum|null $navigationIcon = 'heroicon-s-lock-open';
+
+    protected static string|UnitEnum|null $navigationGroup = 'Tools';
+
+    protected static ?string $title = 'Unlock Tools';
```
wescopeland: UX, minor:

"Tools" is slightly redundant with the category heading. I'd consider renaming this to something like "Manual Unlocks".

Image

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland, line 1: UX: Mixing error reporting and confirmation into one modal creates ambiguity.

My sense is the error state probably shouldn't pop a modal. The data is bad, and the modal is telling us it's bad, but we need to close the modal and go fix the data, then submit again.

I think we should separate error handling from confirmation, and the field itself should strictly validate its inputs. If a given achievement ID or user is invalid, I think we should tell the user at the field(s) and not pop the modal until their data is fixed. This is more than a warning: it's a signal that something is about to go horribly awry. Mistakes are not easily reversible; my sense is the UX should strictly guide the user towards not making a mistake. For the confirmation UX, the modal does an excellent job with that. It's mostly the error handling I'm fixated on.

Image

wescopeland, line 1: UX: I like this multi-game confirmation UI a lot, but I'm pressed to believe having multiple games given in a single list of achievement IDs should be considered a validation error. My sense is, more often than not, this implies user error, and given how hard these things are to undo, we should guard against it instead.

Image

resources/views/filament/pages/manual-unlocks.blade.php
wescopeland, line 1: UX, suggestion (non-blocking): It would probably make sense for the two panels to talk to each other. The "Get Game Achievement IDs" tool is a valid companion to the unlock tool, and it justifies its presence on the page.

If I look up IDs on the right, It'd be nice if I could send them to the left with one click. Otherwise, they are neighbors who don't speak.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland, line 1: UX: The confirm modal has no max height constraint. When pasting a game's set, the modal content extends beyond the monitor's viewport:

Image

I think `<x-filament::modal>` supports a `maxHeight` prop.

wescopeland, line 1: UX: It seems like the modal content can truncate these Progression and Win Condition labels.

Image

app/Policies/GamePolicy.php
wescopeland, line 1: UX, suggestion: On the game resource, recommend not loading the Set Credits relation manager for users with only this new role:

Image

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland, line 1: UX, suggestion: It may be worthwhile in the summary section at the top of the dialog to also indicate how many points total are about to be granted to each user in the list (the sum of points for all achievements).

Image

app/Filament/Pages/ManualUnlocks.php
wescopeland, line 1: DX: Now that we're moving this to Filament, we should probably rip out the legacy functionality in _admin.blade.php_ and/or _request/admin.php_.

wescopeland (commented):
Thanks for opening this PR! This is a digestible chunk, which is very appreciated. I've reviewed thoroughly and noted the things that caught my eye. Please let me know if any of this doesn't make sense.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland: Please disregard - this is my fault. I forgot to run `pnpm build`.

app/Filament/Pages/ManualUnlocks.php
Tsearo: Makes sense, I wasn't sure if I should wait until after this PR is merged to give other admin/mods time to adapt or if it should be done immediately. Will make sure to make a small announcement in mods chat showing where the tool moved to.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: I agree mostly that it's not common to have multiple games in a single request so can definitely add some validation to block that. The one exception I can think of being in regards to subsets. It's not too uncommon for some requests to contain some combination of core/subset achievements. Would it make sense to block completely separate games but still display the existing multi-game confirmation if all of the sets are part of the same game?

Tsearo: This makes a lot more sense after thinking about it, it doesn't make sense to show any achievements if there are invalid inputs. Switching the validations and errors to the fields feels much better overall.

Tsearo: This is very reasonable I mainly was not sure what would make for a good limit. Adding the total number of jobs seems like a good start and then giving a notice when over a certain amount and blocking if over a unreasonable amount. It's not very common to award a large amount of both users or achievements or individual. I'll start with limiting each field to 100 entries and displaying the total number of jobs that will be created in the modal popup.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: Oh nice catch I didn't notice we had that action saved already. It seems to work well for the achievement Ids

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: That is a good catch as that is extra info we can confirm before proceeding. Added total points along with total jobs that will be queued to the summary section.

app/Policies/GamePolicy.php
Tsearo: Updated this as well as the achievement resource to remove unnecessary sections and audit log. Tried to follow your example for the Playtest Manager change but was not fully sure how these changes should be made with two open PRs touching similar sections in different ways.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: This is probably my biggest weak point, I set a min-w-[40rem] to fix this but honestly am not very familiar with best practices in regards to how the modal resizing/flex/etc should work ideally.

resources/views/filament/pages/manual-unlocks.blade.php
Tsearo: I do really like this idea and it would be an improvement, given how uncommon of a scenario it is though I will hold off on figuring out how to implement this for now but will plan to come back to it for improvemenet.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: I believe I fixed this properly but am not sure a good way to verify. Is there an easy way to see all db requests I do not see these specific ones in the laravel debugbar.

database/migrations/2026_03_23_000001_create_manual_unlock_requests_table.php
Jamiras, line 1: These tables don't seem related to the PR.

> Additional database prep for upcoming Manual Unlock Request feature

Recommend removing them from this PR in case things need to change when the feature is reviewed.

app/Community/Enums/ManualUnlockRequestItemState.php
Jamiras, line 1: Similarly, these files are related to the secondary feature. Recommend keeing this PR strictly to the changes required for migrating the page.

Alternately, will the admin page even be necessary once the user-driven feature is converted? I expect that to have the functionality directly in the request page.

app/Filament/Resources/AchievementResource/RelationManagers/AuthorshipCreditsRelationManager.php
Jamiras: What is this change for? How are authorship credits related to manual unlocks?

app/Filament/Resources/GameResource.php
```diff
@@ -648,6 +665,23 @@ public static function getPages(): array
         ];
     }
 
+    /**
+     * Users with only view achievement roles (Manual Unlocker) don't
+     * need to see game metadata, metrics, or rich presence on the detail page.
+     */
+    private static function userHasOnlyViewAchievementRoles(User $user): bool
+    {
+        $acchievementOnlyRoles = [Role::MANUAL_UNLOCKER];
```
Jamiras: `achievement` is misspelled.

app/Filament/Resources/GameResource.php
Jamiras, line 1: These changes seem to imply that a user with only Manual-Unlocker permissions should only be able to view the achievement list on the game page, but the badge and banner image remain. Is that intentional?

I imagine this is just so the user can get the achievement IDs, but I think it would make more sense to have the manual unlock tool be searchable - like the associate event achievemnts dialog:

Image

app/Filament/Resources/AchievementResource.php
Jamiras, line 1: If you're going to filter the visible elements of the game page, you should also filter the visible elements for the achievement page.

app/Policies/AchievementPolicy.php
```diff
@@ -39,6 +39,11 @@ public function manage(User $user): bool
              * writers may update achievement title and description if the respective achievements are open for editing
              */
             Role::WRITER,
+
+            /*
+             * unlockers can view achievement logic as part of their review
```
Jamiras, line 47: Would someone "viewing achievement logic" not have some level of Developer role? Could we just leverage that and make Achievements not visible for someone only with "Manual-Unlocker"?

app/Policies/TicketPolicy.php
```diff
@@ -23,6 +23,7 @@ public function manage(User $user): bool
             Role::TICKET_MANAGER,
             Role::DEVELOPER,
             Role::DEVELOPER_JUNIOR,
+            Role::MANUAL_UNLOCKER,
```
Jamiras: Why does a Manual-Unlocker need Manage Ticket?

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
```diff
@@ -0,0 +1,280 @@
...
+                    ->rows(2)
+                    ->required(),
+
+                Forms\Components\Textarea::make('achievementIdsCsv')
+                    ->label('Achievement IDs')
+                    ->placeholder("9, 17, 25 or 9 17 25")
+                    ->helperText("Paste a comma or space-separated list of Achievement IDs.")
+                    ->rows(2)
+                    ->required()
+                    ->rules(['regex:/^\d+([\s,]+\d+)*$/']),
```
Jamiras: The Get Game Achievement IDs widget supports searching. The username/achievement IDs widgets probably should too.

Image

database/migrations/2026_03_23_000001_create_manual_unlock_requests_table.php
Tsearo: I was trying to do some prep-work early but that makes sense to review as part of the whole change as it will make more sense at that time. Removing this from the current PR.

app/Community/Enums/ManualUnlockRequestItemState.php
Tsearo: Removing these files as well as the above.

I think it will still be useful for some time as the request unlock functionality will take some time to build and expand to the point where it will be able to completely handle most scenarios and there still may be some edge cases where it is beneficial to keep the unlock tool.

At some point I would also like to port over the Copy Unlocks legacy admin page tool but still likely keep that visible to just mods/admins so we can move closer to not needing the legacy page at all.

app/Filament/Resources/GameResource.php
Tsearo: It was intentional yes but can easily be removed if felt unnecessary. I mainly left it as game icons are usually an easy visual identifier for me but I can see how it isn't really necessary.

My initial implementation of the unlock tool did have a searchable user and achievement selection but in reality the majority of the work we do is copy and paste and it is not uncommon to have a list of users/ids to paste in rapidly or at once.

The main reason for giving access to the achievements is just to view the logic as that is sometimes helpful for the unlock team to confirm requirements. This was the simplest method I could think of to give access to that information but I'm sure there is a better alternative.

app/Policies/AchievementPolicy.php
Tsearo: We have a few members on the unlock team that do not have a developer (or jr) role which is the main reason for the changes to allow seeing the logic. Two are past developers so they are at least familiar with it but no longer have the role.

app/Policies/TicketPolicy.php
Tsearo: This was solely to give access to viewing the achievement logic dropdown. As far as I could tell in the [Ticket].blade.php file the only thing manage really does by itself is give access to the logic dropdown and a link to message the reporter. I can remove the role from manage and create a separate function solely for this purpose though if preferred to eliminate confusion.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: Copying same comment from above:
My initial implementation of the unlock tool did have a searchable user and achievement selection but in reality the majority of the work we do is copy and paste and it is not uncommon to have a list of users/ids to paste in rapidly or at once.

Tsearo: Initially I tried allowing for both but did not easily find a way to support both searchable fields while also allowing to paste in a csv (or space separate) list. I may try searching more to see if there is a solution that allows for both as that would ultimately be ideal and would essentially build some of the validation checks in making it a bit cleaner.

app/Filament/Resources/AchievementResource.php
Tsearo: I do limit the Credits section and Audit Log page as these don't seem relevant to the team but everything else felt like valuable information to keep. I suppose Metrics could also be useful on the game page to keep as well.

app/Filament/Resources/GameResource.php
Jamiras: The Add Hubs dialog supports both. Maybe that would be an option?
image

app/Policies/TicketPolicy.php
Jamiras: We probably need to break out a separate permission. Manage ticket implies alot more than just viewing user play history.

app/Filament/Resources/GameResource.php
Tsearo: Oh nice I didn't think about splitting it out into two fields, I'll review this page to see how it handles combining the inputs and should be able to replicate the same behavior. Thanks for pointing that out.

wescopeland: I think we've got the right feature and are really homing in on the right execution. The confirmation modal is thoughtful - someone will look at it and feel confident about what they're about to do.

This form though makes me have to stop and think:
Screenshot 2026-04-04 at 7 51 42 AM

Hubs works because it's two fields and a button. This form is a lot more involved, even though the real workflow is simple: pick users, pick achievements, pick mode, go.

I think the CSV + select pairs should be collapsed into a single visual group. The lowest-code way to solve this is to probably use [Filament Tabs](https://filamentphp.com/docs/5.x/schemas/tabs):
```php
Tabs::make('users')
    ->label('Users')
    ->tabs([
        Tab::make('Paste Users')
            ->icon('heroicon-o-clipboard')
            ->schema([
                Forms\Components\Textarea::make('usernamesCsv')
                    ->label('Users')
                    ->placeholder("User1, User2, User3")
                    ->helperText("Paste a comma or space-separated list of usernames.")
                    ->rows(2),
            ]),
        Tab::make('Search Users')
            ->icon('heroicon-o-magnifying-glass')
            ->schema([
                Forms\Components\Select::make('usernamesSelect')
                    ->label('Users')
                    ->multiple()
                    ->getOptionLabelsUsing(function (array $values): array {
                        return User::whereIn('id', $values)->pluck('username', 'id')->toArray();
                    })
                    ->searchable()
                    ->getSearchResultsUsing(function (string $search): array {
                        return User::search($search)
                            ->withTrashed()
                            ->take(50)
                            ->get()
                            ->pluck('username', 'id')
                            ->toArray();
                    })
                    ->helperText('Search and select users to add.'),
            ]),
    ]),

Tabs::make('achievements')
    ->label('Achievements')
    ->tabs([
        Tab::make('Paste Achievements')
            ->icon('heroicon-o-clipboard')
            ->schema([
                Forms\Components\Textarea::make('achievementIdsCsv')
                    ->label('Achievement IDs')
                    ->placeholder("9, 17, 25")
                    ->helperText("Paste a comma or space-separated list of achievement IDs.")
                    ->rows(2)
                    ->rules(['regex:/^\d+([\s,]+\d+)*$/']),
            ]),
        Tab::make('Search Achievements')
            ->icon('heroicon-o-magnifying-glass')
            ->schema([
                Forms\Components\Select::make('achievementIdsSelect')
                    ->label('Achievements')
                    ->multiple()
                    ->searchable()
                    ->getSearchResultsUsing(function (string $search): array {
                        return Achievement::with('game')
                            ->where('title', 'like', "%{$search}%")
                            ->orWhere('id', 'like', "%{$search}%")
                            ->limit(50)
                            ->get()
                            ->mapWithKeys(function ($achievement) {
                                return [$achievement->id => "ID: {$achievement->id} - Title: {$achievement->title} - Game: {$achievement->game->title}"];
                            })
                            ->toArray();
                    })
                    ->getOptionLabelsUsing(function (array $values): array {
                        return Achievement::with('game')
                            ->whereIn('id', $values)
                            ->get()
                            ->mapWithKeys(function ($achievement) {
                                return [$achievement->id => "ID: {$achievement->id} - Title: {$achievement->title} - Game: {$achievement->game->title}"];
                            })
                            ->toArray();
                    })
                    ->helperText('Search and select achievements to add.'),
            ]),
    ]),
```

This implementation gives us something like:
Screenshot 2026-04-04 at 8 03 18 AM

This is good. Three concepts, three groups, and the eye moves top to bottom without any detours.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland: Yep, it's fixed!

Verifying Livewire performance is really tricky because it tries to do everything async. Here's an easy(ish) way to verify - in _unlock-achievements.blade.php_, we can tap into the component's lifecycle events:
```php
public function rendering(): void
{
    \DB::enableQueryLog();

    if ($this->loadedAchievements && $this->loadedAchievements->isNotEmpty()) {
        $this->loadedAchievements->loadMissing(['game.system']);
    }
}

public function rendered(): void
{
    logger('Modal render queries:', \DB::getQueryLog());
    \DB::disableQueryLog();
}
```

Then, navigate to http://localhost:64000/log-viewer. Without the fix applied, I see:
Screenshot 2026-04-04 at 8 16 32 AM

With the fix applied, I see:
Screenshot 2026-04-04 at 8 16 36 AM

app/Filament/Pages/UnlockTools.php
wescopeland: Still need a rename from "Unlock Tools" to "Manual Unlocks".

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland: Definitely in a lot better place in latest. Only remaining nit is I'm wondering if `$hasUnpromoted`, `$isInvalidHardcore`, `$hasNoValidData`, `$hasMissingData`, and `$hasBlockingWarning` are dead code in the modal.

wescopeland: The current implementation effectively addresses my concern. My only remaining nit is UX-related:
Screenshot 2026-04-04 at 8 25 13 AM

The message says "Warning" even though it's a hard validation failure. I think we should remove the "Warning:" from the string.

app/Filament/Pages/ManualUnlocks.php
wescopeland: Looks like we cleared it up in _admin.blade.php_, but there's still some leftover code in _public/request/admin.php_ for handling the logic that was previously on the page:
```php
if ($action === 'manual-unlock') {
    $awardAchievementID = requestInputSanitized('a');
    $awardAchievementUser = requestInputSanitized('u');
```

resources/js/types/generated.d.ts
wescopeland, line 1: nit: Recommend running `composer types` - 'playtest-manager' is a new valid value.

I think at the moment Playtest Managers are locked out of game set credits entirely. They should still be able to grant set credit.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
wescopeland, line 1: Bug: Unpromoted achievements in softcore mode are allowed to be submitted, but the back-end silently drops them.

In `UnlockPlayerAchievementAction`, we have:
```php
// commit the unlock
if ($achievement->is_promoted) {
    $unlock->save();
```

app/Policies/AchievementPolicy.php
wescopeland, line 1: Bug: Users with only the Artist role now have access to 'Quick Edit' and 'Logic' on the achievement show page:

Image

`viewAchievementLogic` now returns true for any role that can manage achievements, but I think the perms should be far more restrictive. I think it should only be allowed for three roles: devs, jrdevs, and manual unlockers.

app/Filament/Resources/GameResource.php
Tsearo: I love this! I wasn't super happy with the clutter but my knowledge on options wasn't great, the tabs look like they are a perfect fit for this

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: Oh good catch, I believe the current admin.php page allows for them (in softcore even if hardcore is selected) but didn't notice the action behaves differently. 
Imo it's always been rather an annoyance when doing unlocks that we cannot award unpromoted achievements, is there a good reason for this or would it be fine to allow specifically for manual unlocks to still award while unpromoted? Not sure if it relates deeper into things like metrics/jobs/etc?

I see the following options for how to proceed:

- Allow manual unlocks for unpromoted achievements - softcore and hardcore
   Personally preferred option, would be under the assumption that there are no other reasons for blocking it and it is exclusively only able when being manually awarded

- Allow manual unlocks for unpromoted achievements - softcore only
    Matches what the current admin.php tool does, softcore is much more lenient in general so I don't see much reason to block it and can add that to the the UnlockPlayerAchievementAction easy enough

- Validation/Block if any unpromoted achievements
    Simplifies validation checks and matches current action

wescopeland: I'm personally fine with it being allowed, as long as it's only for manual unlocks (and the code should enforce this). It makes sense from a manual unlock workflow POV.

app/Policies/AchievementPolicy.php
Tsearo: This makes sense and can be done by removing the manage role from viewAchievementLogic. I'll need to think of how to best ensure that other manage roles only show the Quick Edit when appropriate. I believe Game_Hash_Manager and Artist is the only applicable one that doesn't need it at all and Writer will still want to be able to make text edits. Will add some further validations to the Contribute panel along with the changes I'm doing for the unlockers to be able to see the logic button.

app/Policies/AchievementPolicy.php
```diff
@@ -41,8 +41,8 @@ public function manage(User $user): bool
             Role::WRITER,
 
             /*
-             * unlockers can view achievement logic as part of their review
-             */
+            * unlockers can view achievement logic as part of their review
+            */
```
wescopeland: nit: seem to be missing some whitespace here for consistency

app/Data/UserPermissionsData.php
```diff
@@ -49,6 +50,7 @@ public function __construct(
         public Lazy|bool $updateAchievementDescription,
         public Lazy|bool $updateAchievementIsPromoted,
         public Lazy|bool $updateAchievementPoints,
+        public Lazy|bool $updateAchievementQuickEdit,
```
wescopeland: nit: Let's rename this to `$quickEditAchievement`. There will probably be a few other quick edit things in the future, and this will help with grouping them together in this file.

resources/js/features/achievements/components/AchievementContributePanel/AchievementContributePanel.tsx
```diff
@@ -36,24 +36,27 @@ export const AchievementContributePanel: FC = () => {
...
+            ) : null}
             <div className="grid grid-cols-2 gap-1">
-              <PlayableSidebarButton
-                href={`/manage/achievements/${achievement.id}`}
-                target="_blank"
-                IconComponent={LuWrench}
-              >
-                {t('Manage')}
-              </PlayableSidebarButton>
+              {can.manageAchievements ? (
```
wescopeland, line 52: nit: recommend adding a line of whitespace for readability:
```tsx
) : null}

<div className="grid grid-cols-2 gap-1">
```

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
```diff
@@ -0,0 +1,412 @@
...
+            $baseGameIds = collect();
+            foreach ($setIds as $setId) {
+                $setLinks = $links->where('achievement_set_id', $setId);
+                $coreLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Core->value);
+                $bonusLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Bonus->value);
+                $specialtyLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Specialty->value);
+                $exclusiveLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Exclusive->value);
+
+                // Specialty and Bonus sets fall back to the base game ID.
+                // Exclusive sets remain isolated by falling back to their own core link's game ID.
```
wescopeland: Bug: Unfortunately, I think this `else` block's logic doesn't match the multiset rules at the moment. It's loading `game_achievement_sets` rows and reducing each selected achievement set to one `baseGameId` with this logic:
- use the `bonus` link if it's present
- otherwise use the `specialty` link
- otherwise use the `core` link

We also have `will_be_bonus` and `will_be_specialty` cases. As of https://github.com/RetroAchievements/RAWeb/pull/4723, `challenge` and `will_be_challenge` also exist.

Let's say we have this setup for a game:
- Base game: Dragon Quest III, game_id = 100
- Challenge subset game: Dragon Quest III [Subset - Challenge], game_id = 200

For that challenge set, `game_achievement_sets` can look like:
- (game_id=200, type=core) for the "subset game" itself
- (game_id=100, type=challenge) for the base game link

The current validator ignores `challenge`, so it picks 200.
The real multiset resolver uses `challenge`, so it picks 100.

If a manual unlocker selects one normal core achievement from game 100 and one challenge achievement from the same family, the form can incorrectly say "Achievement IDs belong to different (base) games".

A minimal fix probably looks something like:
```php
// untested pseudocode

foreach ($setIds as $setId) {
    $setLinks = $links->where('achievement_set_id', $setId);
    $coreLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Core->value);
    $bonusLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Bonus->value);
    $challengeLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Challenge->value);
    $specialtyLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::Specialty->value);
    $willBeBonusLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::WillBeBonus->value);
    $willBeChallengeLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::WillBeChallenge->value);
    $willBeSpecialtyLink = $setLinks->firstWhere('type', App\Platform\Enums\AchievementSetType::WillBeSpecialty->value);

    // Subsets that load with a base game fall back to that game's ID.
    // Specialty and exclusive sets remain isolated by falling back to their own core link's game ID.
    $baseGameId = $bonusLink?->game_id
        ?? $challengeLink?->game_id
        ?? $specialtyLink?->game_id
        ?? $willBeBonusLink?->game_id
        ?? $willBeChallengeLink?->game_id
        ?? $willBeSpecialtyLink?->game_id
        ?? $coreLink?->game_id;

    if ($baseGameId) {
        $baseGameIds->push($baseGameId);
    }
}
```

wescopeland (commented):
Almost there!

app/Policies/TicketPolicy.php
Tsearo: Removed MANUAL_UNLOCKER from Manage and split permissions out into viewLogic and viewHistory functions which are now reference in [Ticket].blade.php for their respective sections visibility.

app/Filament/Resources/AchievementResource.php
Tsearo: I went back and forth a bit but ended up just leaving the game and achievement pages displaying everything and only removing the contribution and audit views.

resources/views/livewire/administrative-tools/unlock-achievements.blade.php
Tsearo: Updated Usernames and Achievement sections to use Filament Tabs per Wes' suggestion so now you can select either csv or search for each field depending on preference.

wescopeland (approved):

Jamiras (approved):

PR #4724, opened 2026-04-03
feat(media): add thumbnail transform and regenerate command

wescopeland:
In prod, run:
```
php artisan ra:platform:game:regenerate-screenshot-conversions # kicks off ~1500 jobs
```

---

This PR:
* Adds a new `thumbnail` transform to game screenshots.
* Adds a dedicated regenerate command for game screenshot media.

Thumbnails are height constrained to 240px, with no width constraint. This allows them to retain their aspect ratio, and also solves a layout shift caused by uncached images.

While MediaLibrary does have its own built-in regenerate command, in prod it'll pull directly from S3, which is undesirable. We should pull from the CDN first, and only pull from S3 if it's somehow a CDN miss. We should also skip any work that isn't needed.

**How to test this PR**
In your _.env_:
```
AWS_URL=https://s3.retroachievements.org
```

Then:
```
sail artisan ra:platform:game:regenerate-screenshot-conversions 1
```

Then in your _.env_:
```
AWS_URL=http://localhost:${FORWARD_MINIO_PORT}/local
```

Now visit http://localhost:64000/game/1. You should see the thumbnail variants loaded for the images:
Screenshot 2026-04-03 at 6 01 14 PM

### Before
Screenshot 2026-04-03 at 5 14 54 PM

Screenshot 2026-04-03 at 5 17 10 PM

### After
Screenshot 2026-04-03 at 5 15 38 PM

Screenshot 2026-04-03 at 5 17 13 PM


app/Platform/Actions/RegenerateGameScreenshotConversionsAction.php
Jamiras, line 1: This doesn't seem to have any effect if images are smaller than then 240px:

http://localhost:64000/game/549

Image

Is there some way we can automatically double image sizes if the system's resolution is less than 150px (GameBoy and MegaDuck are 192x144, GameGear is 160x144, Lynx is 160x102, Arduboy is 128x64, WonderSwan is 224x144, PokeMini is 96x64)? Maybe even NeoGeo Pocket at 160x152, Supervision and WASM-4 at 160x160, and GBA at 240x160?

Should we just scale all thumbnails to 240px if they aren't +/-10% (216-266)?

At a minimum, it would be ideal if all screenshots for a single game end up at the same resolution - whatever that ends up being.

wescopeland: > This doesn't seem to have any effect if images are smaller than then 240px

This is intentional. The current `thumbnail` transform implementation is downscale-only.

I think there may be two separate questions here:
* Should the preview area enforce a consistent visual height?, and
* Should low-res systems be upscaled to fill that height?

For most of the systems you've listed, the second option is where things get kinda messy. A clean upscale needs nearest-neighbor at an integer multiple, and many of these systems do not land cleanly at a max height of 240px. For example, Game Boy @ 2x is 288px.

I think we have three coherent technical options:
1. Keep the 240px cap and only scale down. This preserves asset quality and avoids introducing fractional upscales, which we've already seen folks get upset about. Some low-res systems will still appear smaller.
2. Allow integer-scaled previews to exceed 240px. This gives more visual consistency, but I'm not convinced this added height is desirable.
3. Force non-integer scaling to 240px. We've taken a similar approach elsewhere, and received feedback that this was undesirable.

We also have a social option: we could let the community sort these instances out when player uploads are a thing. We could even introduce some kind of "bounty board" that specifically surfaces + incentivizes these pages for players to fix. Thanks to the MediaLibrary migration, we already know exactly which game pages suffer from this problem.

wescopeland:
If there are no objections, I'd like to merge this PR sometime this week.

Jamiras (approved):
That last comment isn't resolved. My concern is that both images end up the same size. Ideally, ultra-small images would be enlarged to better fill the area, but I understand not wanting to oversize them. If the solution is simply to capture new images at consistent sizes, sobeit.

While I don't necessarily want to block merging this until a solution is decided upon, it does seem pertintent to the feature.

However, given this constraint:
> This is intentional. The current thumbnail transform implementation is downscale-only.

I've approved it.

PR #4752, opened 2026-04-08
feat(event): add ConvertGameToEvent command

Jamiras:
https://discord.com/channels/476211979464343552/1491494198103511243

This isn't the first game I've had to convert to an event, and I'm sure it won't be the last. As such, rather than trying to remember all the manual steps, I've written a command to do it.


app/Platform/Commands/ConvertGameToEvent.php
```diff
@@ -0,0 +1,90 @@
...
+                'award_key' => $event->id,
+            ]);
+
+        // update metrics
+        UpdateGamePlayerCountJob::dispatch($game->id);
+        app()->make(UpdateTotalGamesCountAction::class)->execute();
+
+        foreach (PlayerStatRankingKind::beatenCases() as $kind) {
+            UpdateBeatenGamesLeaderboardJob::dispatch($systemId, $kind)->onQueue('game-beaten-metrics');
+        }
```
Jamiras: Any other denormalized data that should be updated?

app/Platform/Commands/ConvertGameToEvent.php
```diff
@@ -0,0 +1,90 @@
...
+            ->where('award_key', $game->id)
+            ->delete();
+
+        // move mastery badges
+        PlayerBadge::where('award_type', AwardType::Mastery)
+            ->where('award_key', $game->id)
+            ->update([
+                'award_type' => AwardType::Event,
+                'award_key' => $event->id,
+            ]);
```
Jamiras, line 52: This will effectively create hardcore event badges for softcore completion badges. Would it be better to delete the softcore completion badges?

wescopeland: I think so, yes.

app/Platform/Commands/ConvertGameToEvent.php
wescopeland: None that I can think of.

app/Platform/Commands/ConvertGameToEvent.php
wescopeland, line 1: Should any of this be wrapped in a transaction? Lines 67-85 in particular seem unsafe to retry or partially fail. I think if anything throws after the wrapper rows are created, the DB can be left with a half-converted game.

wescopeland (approved):

Jamiras:
Made the following changes:
* Wrapped the whole thing in a transaction.
* Softcore unlocks are discarded
* Softcore badges are discarded
* Game hashes are discarded

wescopeland (approved):

PR #4757, opened 2026-04-09
perf(user): avoid loading extra models into memory just for beaten game percentage calculation

Jamiras:
Replaces a few queries fetching large amounts of data with a couple of aggregation queries.

Before:
image

After:
image

Extension of #4749

### DIFF
--- a/app/Community/Components/UserProfileMeta.php
+++ b/app/Community/Components/UserProfileMeta.php
@@ -4,10 +4,13 @@ declare(strict_types=1);
 
 namespace App\Community\Components;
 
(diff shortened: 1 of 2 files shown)


app/Community/Components/UserProfileMeta.php
Jamiras, line 1: Note: this does have an impact on the actual beaten games percentage calculation.

The previous algorithm would count a mastered game twice (once for the beaten award and once for the mastered award).

Jamiras:
crap. just saw #4750

PR #4763, opened 2026-04-11
feat: add a softcore_only emulator flag

wescopeland:
https://discord.com/channels/476211979464343552/1492213658258903240

When the flag is set, the warning achievement will stop asking users to update their emulator.

http://localhost:64000/manage/emulators/24

Screenshot 2026-04-11 at 7 28 11 AM

Screenshot 2026-04-11 at 7 29 00 AM


app/Connect/Actions/GetAchievementSetsAction.php
```diff
@@ -54,7 +55,8 @@ protected function initialize(Request $request): ?array
         $this->isPromoted = Achievement::isPromotedFromLegacyFlags($flag);
 
         $userAgentService = new UserAgentService();
-        [$this->clientSupportLevel, $this->coreRestriction] = $userAgentService->getSupportLevelAndCoreRestriction(request()->header('User-Agent'));
+        [$this->clientSupportLevel, $this->coreRestriction, $this->isSoftcoreOnlyEmulator]
```
Jamiras: I would rather introduce a new `ClientSupportLevel::SoftcoreOnly` than pass back an extra flag that only applies to one of the existing levels.

app/Platform/Services/UserAgentService.php
```diff
@@ -338,10 +341,18 @@ public function getSupportLevelAndCoreRestriction(string|array|null $userAgent):
              * special case: Dolphin/e5d32f273f must still be allowed as it's the most stable development build
              */
             if (str_starts_with($userAgent, 'Dolphin/e5d32f273f ')) {
-                return [ClientSupportLevel::Outdated, null];
+                return [ClientSupportLevel::Outdated, null, $isSoftcoreOnly];
```
Jamiras: Outdated already implies softcore only.
https://github.com/RetroAchievements/RAWeb/blob/0a20b38e3ba18287227c23219e638e46f9f662ec/app/Enums/ClientSupportLevel.php#L12-L13

app/Connect/Actions/GetAchievementSetsAction.php
wescopeland: Introduced this in latest and removed that extra flag.

app/Platform/Services/UserAgentService.php
wescopeland: https://github.com/RetroAchievements/RAWeb/pull/4763#discussion_r3089906374

Additionally, the flag has some documentation to hopefully bridge the distinction:
```php
// client is recognized, but it's known that hardcore is unsupported for it.
case SoftcoreOnly = 6;
```

app/Platform/Services/UserAgentService.php
```diff
@@ -338,12 +340,20 @@ public function getSupportLevelAndCoreRestriction(string|array|null $userAgent):
              * special case: Dolphin/e5d32f273f must still be allowed as it's the most stable development build
              */
             if (str_starts_with($userAgent, 'Dolphin/e5d32f273f ')) {
-                return [ClientSupportLevel::Outdated, null];
+                return [$isSoftcoreOnly ? ClientSupportLevel::SoftcoreOnly : ClientSupportLevel::Outdated, null];
```
Jamiras: We can probably remove support this version of dolphin. There's only one user who has used it in the last three weeks.
```
MariaDB [retroachievements]> select user_id, count(*) from player_sessions where id > 87113599 and user_agent like '%e5d32f273f%' group by 1;
+---------+----------+
| user_id | count(*) |
+---------+----------+
|  153229 |       10 |
+---------+----------+
1 row in set (1.800 sec)
```

Jamiras (approved):

app/Enums/ClientSupportLevel.php
```diff
@@ -25,6 +25,9 @@ enum ClientSupportLevel: int
     // but hardcore unlocks and leaderboard submissions are still allowed.
     case Warned = 5;
 
+    // client is recognized, but it's known that hardcore is unsupported for it.
```
Jamiras: I feel like this could be worded better. Maybe:
> client is recognized, but doesn't meet our hardcore requirements

app/Platform/Services/UserAgentService.php
wescopeland: Removed.

app/Enums/ClientSupportLevel.php
wescopeland: Updated to this exact copy.

PR #4799, opened 2026-04-21
feat(manage/game): hide Rejected screenshots by default

wescopeland:
Users are surprised to see unpublished screenshots in the media tab's screenshots relation manager. For auditing purposes, I'm not yet certain we should soft delete these.

The Status filter now defaults to "Published + Pending". Selectable filter values show counts.

Screenshot 2026-04-21 at 5 52 13 PM

Screenshot 2026-04-21 at 5 52 18 PM


app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
Jamiras, line 1: What about Replaced images?

Image

They currently show up on the default filter (Pending + Approved) and can't be filtered on directly.

wescopeland: Thanks, addressed this in latest. They're now excluded from the default along with Rejected and have their own filter option.

Jamiras (approved):

PR #4800, opened 2026-04-21
fix(home): prevent in game status bar from overlapping with nav header

angelol-git:
fix for #4755 

Between widths 640px to 1023px (sm-lg breakpoints). Extra margin top value was applied to the in game status bar. 

**Before**
before

**After**
after


wescopeland (approved):
Thanks!

Jamiras (approved):

PR #4813, opened 2026-04-24
fix(game): recalc denormalized stats on set version changes

wescopeland:
Should hopefully resolve https://discord.com/channels/476211979464343552/1026595325038833725/1497348396456087765


Jamiras (approved):
I think the reported issue is just another instance fixed by #4740 and the metrics just hadn't been recalcuated yet, which is what this PR does.

PR #4820, opened 2026-04-25
fix(GameScreenshotModerationResource): scope the table query

wescopeland:
This URL should currently only show two rows: https://retroachievements.org/manage/game-screenshot-moderations?filters[status][value]=rejected

However, it's showing _all_ screenshots with a rejected status. That isn't the intent of the tool. The table should only show screenshots that reached a status due to human review, likely through this very tool.

Now, for anything non-pending, we require a non-null `reviewed_by_user_id` value for the row to be displayed.


app/Filament/Resources/GameScreenshotModerationResource.php
Jamiras, line 1: > The table should only show screenshots that reached a status due to human review, likely through this very tool.

How else would they end up rejected? Even if its from a someone Unpublishing an image, it was still human reviewed. And most of those seem to be people replacing images with something of an acceptable resolution. Maybe there should be a way to flag an in-game image as Replaced (how does that happen anyway?).

wescopeland: The intention here is to have the tool only show images rejected through the tool.

Rejections that happen through the game resource media tab are likely through standard dev activity, not a result of moderation review.

> Maybe there should be a way to flag an in-game image as Replaced (how does that happen anyway?).

"Replaced" is an automatic side effect of approving some images. Given an image is a primary screenshot but has an incorrect resolution, approving an image that has a correct resolution will set the old primary to Replaced.

Jamiras (approved):

PR #4822, opened 2026-04-25
fix(game-checklist): show softcore progress label

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1497366307484598292.

Screenshot 2026-04-25 at 6 23 28 PM


resources/js/common/components/GamesListItem/GamesListItem.tsx
Jamiras, line 1: There's still something wrong with the "of X points" counts (unrelated to this change, but should be fixed either as part of this PR, or separately):

Image

In this case, [I do have](http://localhost:64000/user/Jamiras/game-checklist?list=Example%3A1505) 75 points of hardcore and 75 points of softcore, but the "of 0" should be "of 320".

Jamiras (approved):

resources/js/common/components/GamesListItem/GamesListItem.tsx
wescopeland: Thanks, fixed in latest.

PR #4824, opened 2026-04-26
feat(achsets): capture achievement set versions

Jamiras:
Achievement sets will now be versioned immediately upon claim completion and periodically if changes to the number of published achievements or published points are detected.

This could be used to drive the #core-set-log lists, and serves as data to derive "missing points" history (see  https://discord.com/channels/310192285306454017/1370440692874088622/1370440692874088622 and https://discord.com/channels/310192285306454017/1387686580415758509/1387686580415758509)

### DIFF
--- a/app/Platform/Actions/CheckForAchievementSetChangesAction.php
+++ b/app/Platform/Actions/CheckForAchievementSetChangesAction.php
@@ -10,7 +10,7 @@ class CheckForAchievementSetChangesAction
 {
     public function execute(AchievementSet $achievementSet): void
     {
-        $latestVersion = $achievementSet->versions()->latest()->first();
+        $latestVersion = $achievementSet->versions()->orderByDesc('version')->first();
 
         if ($latestVersion) {
             // update the volatile metrics so they're accurate at the point where a new version is created
--- a/app/Platform/Actions/CheckForAchievementSetChangesAction.php
+++ b/app/Platform/Actions/CheckForAchievementSetChangesAction.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Platform\Actions;
+
+use App\Models\AchievementSet;
+
+class CheckForAchievementSetChangesAction
+{
+    public function execute(AchievementSet $achievementSet): void
+    {
+        $latestVersion = $achievementSet->versions()->latest()->first();
+
+        if ($latestVersion) {
+            // update the volatile metrics so they're accurate at the point where a new version is created
+            $latestVersion->players_total = $achievementSet->players_total ?? 0;
+            $latestVersion->players_hardcore = $achievementSet->players_hardcore ?? 0;
+            $latestVersion->achievements_unpublished = $achievementSet->achievements_unpublished ?? 0;
+
+            if ($latestVersion->points_total === $achievementSet->points_total) {
+                // if the points_total changed, the points_weighted value will have changed too,
+                // and we don't want that reflected in the existing version - only in the new one.
+                $latestVersion->points_weighted = $achievementSet->points_weighted ?? 0;
+            }
+
+            $latestVersion->save();
+
+            if ($latestVersion->achievements_published === $achievementSet->achievements_published
+                && $latestVersion->points_total === $achievementSet->points_total) {
+                // no change to number of published achievements or points, just update the other fields.
+                return;
+            }
+
+            // change detected, create a new version
+        } elseif (!$achievementSet->achievements_published) {
+            // don't bother versioning anything without published achievements
+            return;
+        }
+
+        $achievementSet->versions()->create([
+            'version' => $latestVersion ? $latestVersion->version + 1 : 1,
+            'parent_id' => $latestVersion?->id,
+            'players_total' => $achievementSet->players_total ?? 0,
+            'players_hardcore' => $achievementSet->players_hardcore ?? 0,
+            'achievements_published' => $achievementSet->achievements_published ?? 0,
+            'achievements_unpublished' => $achievementSet->achievements_unpublished ?? 0,
+            'points_total' => $achievementSet->points_total ?? 0,
+            'points_weighted' => $achievementSet->points_weighted ?? 0,
+            'created_at' => $achievementSet->updated_at,
+        ]);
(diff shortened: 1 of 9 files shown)


app/Platform/Actions/CheckForAchievementSetChangesAction.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace App\Platform\Actions;
+
+use App\Models\AchievementSet;
+
+class CheckForAchievementSetChangesAction
+{
+    public function execute(AchievementSet $achievementSet): void
+    {
+        $latestVersion = $achievementSet->versions()->latest()->first();
```
wescopeland: Minor: `latest()` is shorthand for `orderByDesc('created_at')`, but `created_at` may not be a good proxy for "latest" in this case.

The schema unfortunately allows for null `created_at` values, so `created_at` is theoretically not monotonic with the `version` field.

Practically, I don't think there's much of a concern - it's more about correctness. Given bad data, `latest()` can plausibly return the wrong row, causing the next insert to compute `version + 1` and `parent_id` from the wrong record, which can fork the version chain or hit the `(achievement_set_id, version)` unique key on subsequent runs.

Either `created_at` should be adjusted in the migration to be non-nullable, or we should order by a monotonic field, ie:
```php
$latestVersion = $achievementSet->version()->orderByDesc('version')->first();
```

app/Platform/Commands/CheckForAchievementSetChanges.php
```diff
@@ -0,0 +1,65 @@
...
+                ->first();
+            if ($gameAchievementSet) {
+                $this->checkForChangesAction->execute($gameAchievementSet->achievementSet);
+                $this->info('Done');
+            } else {
+                $this->error("Unknown game");
+            }
+        } else {
+            if ($this->option('full')) {
+                $achievementSets = AchievementSet::where('achievements_published', '>', 0)->get();
```
wescopeland: Consider:
```php
$achievementSets = AchievementSet::where('achievements_published', '>', 0)
    ->orWhereHas('versions')
    ->get();
```
In the event the daily sweep fails, this will capture previously-versioned sets that are fully demoted so they also get the demoted version recorded during a `--full` reconciliation/backfill.

database/factories/AchievementSetVersionFactory.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+class AchievementSetVersionFactory extends Factory
+{
+    protected $model = AchievementSetVersion::class;
+
+    public function definition(): array
+    {
+        $achievementSet = AchievementSet::factory()->create();
+        $playersTotal = rand(0, $achievementSet->players_total);
+        $playersHardcore = min(rand(0, $achievementSet->playersHardcore), $playersTotal);
```
wescopeland: nit: Recommend using the snake_case field name for consistency.

```suggestion
        $playersHardcore = min(rand(0, $achievementSet->players_hardcore), $playersTotal);
```

app/Platform/Actions/CheckForAchievementSetChangesAction.php
wescopeland: Thinking more on this, `orderByDesc('version')` is probably the right call, no matter where `created_at` ends up. If two versions somehow have identical `created_at` values due to some future-as-of-yet-unwritten side effect writing to the table, we'd end up with the same forking/uniqueness problem.

app/Platform/Actions/CheckForAchievementSetChangesAction.php
```diff
@@ -0,0 +1,53 @@
...
+            if ($latestVersion->points_total === $achievementSet->points_total) {
+                // if the points_total changed, the points_weighted value will have changed too,
+                // and we don't want that reflected in the existing version - only in the new one.
+                $latestVersion->points_weighted = $achievementSet->points_weighted ?? 0;
+            }
+
+            $latestVersion->save();
+
+            if ($latestVersion->achievements_published === $achievementSet->achievements_published
+                && $latestVersion->points_total === $achievementSet->points_total) {
```
wescopeland, line 30: Intellectual curiosity: Should we also treat trigger logic changes, achievement type changes (progression/win only), and/or explicit achievement point value changes (swaps like Achievement A going from 10 -> 5 and Achievement B going from 5 -> 10 are invisible at the moment) as new versions?

I only ask because `UpdateGameMetricsAction` hashes over `id`, `trigger_definition`, `type`, and `points` for every promoted achievement. When the hash changes, the hash is stored on `games.achievement_set_version_hash` - our back-end seems to already consider these sorts of changes material enough to document as a version hash, but the version history table doesn't record that anything happened.

app/Platform/Actions/CheckForAchievementSetChangesAction.php
Jamiras: I agree. I was just looking for the opposite of `first()` and `latest()` is the only thing I found.

app/Platform/Actions/CheckForAchievementSetChangesAction.php
Jamiras: The impetus for this feature is to help players understand where they lost points. None of those changes factor into that.

I don't think changing `type` warrants tagging a new achievement set version. "total_points" and "num_achievement" are properties of the achievement set. "type" is not. But I could see tagging a new version if the beat requirements changed. I guess that kind of information could be stored in the `definition` column. I wasn't sure what it was for, so I left it alone.

Similarly, `trigger_definition` isn't really a property of the achievement set. Additionally, it's already versioned by itself.

app/Platform/Commands/CheckForAchievementSetChanges.php
Jamiras: Makes sense. The `--full` option was a last minute addition. Originally, I just had this as an initial populating step via `if (!AchievementSetVersions::exist())`, but I figured there was a non-zero chance of a claim being completed between deployment and running the command.

database/factories/AchievementSetVersionFactory.php
Jamiras: done.

wescopeland (approved):

PR #4825, opened 2026-04-26
use intersection of players for bonus subsets, only subset player count for other subsets

Jamiras:
Any subset that requires additional effort from the user (opt-in or unique hash) should only count players who have earned achievements in that subset (and not include players from the base set). Bonus sets, which are loaded without additional effort from the user should continue to count players who have earned achievements in either the base set or the bonus set.

As discussed in this discord thread, and summarized here: https://discord.com/channels/310192285306454017/1488177530786680852/1488273598375530517
>  If bonus were truly "something in addition to what is expected" (as defined in the dictionary), then it should leverage the base player count. As soon as you start throwing in things that require playing abnormally, they should only count the players who chose to play in that manner.

Because players can still load bonus sets using a unique hash (at least for most existing bonus sets), it's possible to have players who have played only the bonus set and not the base set. As such, some bonus sets may appear to have more players than the base set. Similarly, games with multiple bonus sets may have different counts for each bonus set. The base set will still only count players who have earned achievements in the base set.

Additionally, as this is changing the bonus set behavior from "inherit player count from parent" to "intersect parent players with bonus set players", this will fix https://discord.com/channels/310192285306454017/1493357676376227983/1493357676376227983

### DIFF
--- a/app/Platform/Actions/UpdateGamePlayerCountAction.php
+++ b/app/Platform/Actions/UpdateGamePlayerCountAction.php
@@ -18,11 +18,12 @@ class UpdateGamePlayerCountAction
 
         $coreGameAchievementSet = $game->gameAchievementSets()->core()->first();
         if ($coreGameAchievementSet) {
-            $bonusSubsetAchievementSet = GameAchievementSet::query()
+            $bonusSubsetAchievementSets = GameAchievementSet::query()
                 ->where('achievement_set_id', $coreGameAchievementSet->achievement_set_id)
                 ->where('type', AchievementSetType::Bonus)
-                ->first();
-            if ($bonusSubsetAchievementSet) {
+                ->select('game_id')
+                ->get();
+            foreach ($bonusSubsetAchievementSets as $bonusSubsetAchievementSet) {
                 // if this is a bonus subset, also include the players of the parent game
                 $gameIds[] = $bonusSubsetAchievementSet->game_id;
             }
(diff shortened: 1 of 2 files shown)


wescopeland:
Will this require any command to run to fix prod player count data?

Jamiras:
It will self-correct over time as soon as the player count for a subset changes (either through a new player, or a player becoming unranked).

I've been testing with specific subsets using `sail artisan ra:platform:game:update-player-count N`. You can leave off the `N` to regenerate everything, but since only subsets are affected, we could manually create a CSV of all subsets and pass that in.

app/Platform/Actions/UpdateGamePlayerCountAction.php
```diff
@@ -5,43 +5,49 @@
...
-            } else {
-                $parentGame = null;
+        $gameIds = [$game->id];
+
+        $coreGameAchievementSet = $game->gameAchievementSets()->core()->first();
+        if ($coreGameAchievementSet) {
+            $bonusSubsetAchievementSet = GameAchievementSet::query()
+                ->where('achievement_set_id', $coreGameAchievementSet->achievement_set_id)
+                ->where('type', AchievementSetType::Bonus)
+                ->first();
```
wescopeland: What should we do if a bonus set is linked to multiple games? This happens with stuff like Pokemon.

Jamiras: The `Game::parent` logic (which is what was previously used), just grabs the oldest record. 
https://github.com/RetroAchievements/RAWeb/blob/bd1f8d7e925ccc52ebb784fc3380b8e410fec6e4/app/Models/Game.php#L677-L679

I think in this case, we'd probably want to intersect both parents.

Jamiras: done

wescopeland (approved):

PR #4826, opened 2026-04-26
fix(achievement): remediate stale state from delete-then-restore actions

wescopeland:
Given an achievement is soft deleted in Filament and then restored, it's never placed back into `achievement_set_achievements`, and thus is trapped in purgatory.

This PR makes a behavior change to the flow. Rather than removing the achievement from `achievement_set_achievements` on soft delete, we now preserve the row and rely on `achievements.deleted_at` to control visibility.


Jamiras (approved):
I'm not really sure how to test this, but the unit tests look good.

PR #4827, opened 2026-04-27
perf(api): optimize API_GetGameProgression

wescopeland:
This PR improves the performance of API_GetGameProgression.php, which on prod is currently taking 30s+ for some parameters. The optimizations introduced by this branch make the endpoint ~9-10x faster along the most intensive data paths.

**API_GetGameProgression.php?i=20580&h=1**
```
Benchmark 1: old
  Time (mean ± σ):      3.454 s ±  0.041 s    [User: 0.003 s, System: 0.003 s]
  Range (min … max):    3.413 s …  3.510 s    5 runs

Benchmark 2: optimized
  Time (mean ± σ):     358.5 ms ±  11.2 ms    [User: 3.1 ms, System: 3.3 ms]
  Range (min … max):   350.0 ms … 378.2 ms    5 runs

optimized ran
  9.63 ± 0.32 times faster than old
```

**API_GetGameProgression.php?i=17976**
```
Benchmark 1: old
  Time (mean ± σ):      3.755 s ±  0.068 s    [User: 0.006 s, System: 0.007 s]
  Range (min … max):    3.673 s …  3.819 s    5 runs

Benchmark 2: optimized
  Time (mean ± σ):     347.0 ms ±   6.6 ms    [User: 7.6 ms, System: 8.1 ms]
  Range (min … max):   337.4 ms … 353.5 ms    5 runs

Summary
  optimized ran
   10.82 ± 0.28 times faster than old
```


public/API/API_GetGameProgression.php
```diff
@@ -121,50 +122,86 @@
...
     ->whereIn('user_id', $recentPlayerIds)
     ->whereIn('achievement_id', $achievementIds)
     ->whereNull('unlocker_id')
-    ->orderByDesc('unlocked_effective_at');
-foreach ($allUnlocks->get() as $unlock) {
+    ->select(['user_id', 'achievement_id'])
+    ->selectRaw(sprintf($timestampExpression, 'unlocked_at') . ' as unlocked_at_timestamp')
+    ->selectRaw(sprintf($timestampExpression, 'unlocked_hardcore_at') . ' as unlocked_hardcore_at_timestamp');
+$latestUnlockAt = null;
+foreach ($allUnlocks->toBase()->get() as $unlock) {
```
wescopeland: We're using `toBase()->get()` for the large `player_achievements` and `player_sessions` reads - we only need the scalar data, so we'll skip hydrating full Eloquent models for these.

public/API/API_GetGameProgression.php
```diff
@@ -121,50 +122,86 @@
     ->where('type', PlayerProgressResetType::Game)
     ->where('type_id', $game->id)
     ->whereIn('user_id', $recentPlayerIds)
-    ->pluck('created_at', 'user_id');
+    ->pluck('created_at', 'user_id')
+    ->map(fn ($createdAt): int => $createdAt->getTimestamp());
+
+$timestampExpression = DB::connection()->getDriverName() === 'sqlite'
+    ? "strftime('%%s', %s)"
+    : 'UNIX_TIMESTAMP(%s)';
```
wescopeland: Bypass Carbon as much as possible.

public/API/API_GetGameProgression.php
```diff
@@ -121,50 +122,86 @@
...
 
 // ===== build unlock lists grouped by user (in reverse order so we can use array_pop) =====
 $unlocks = [];
+$achievementsFirstPublishedAtTimestamp = $achievementsFirstPublishedAt?->getTimestamp();
 $allUnlocks = PlayerAchievement::query()
+    ->forceIndex('player_achievements_achievement_id_user_id_unlocked_hardcore_at')
```
wescopeland, line 131: Force the planner to use the right index.

public/API/API_GetGameProgression.php
Jamiras: There's a `unixTimestampStatement` helper function for this.

Jamiras (approved):

PR #4828, opened 2026-04-27
feat: add media contribution site awards, only for non-dev community contributions

wescopeland:
This PR adds 6 award tiers for user-submitted game screenshots.

**Eligibility**
The spirit of this badge is to reward _community contribution_, not standard dev work. A screenshot only counts towards the submitter's progress when all of the following are true:
* It was approved via the `GameScreenshotModerationResource` tool in Filament.
* A different user than the submitter approved it.
* The submitter has no authored achievements for that game (promoted or unpromoted, doesn't matter).
* The submitter has no active claim on that game.

**Revalidation**
Award eligibility is recomputed (and the badge can be downgraded or revoked) when any of these conditions change:
* A screenshot is approved in the Filament review tool.
* A user creates or drops a claim.
* A user authors an achievement for the game (becoming a dev for that game).

---

Screenshot 2026-04-27 at 7 07 39 PM

Screenshot 2026-05-04 at 7 49 13 PM


app/Helpers/render/site-award.php
```diff
@@ -349,6 +349,19 @@ function RenderAward(
...
+            tooltip: "<div class='p-2 max-w-[320px] text-pretty text-menu-link flex flex-col gap-1'><p class='font-bold'>Media Contribution</p><span>Awarded for contributing {$threshold} or more approved screenshots to game galleries.</span><p class='italic'>{$awardDate}</p></div>",
+            iconUrl: asset("/assets/images/badge/mediaContrib-$awardData.png"),
+            iconSize: $imageSize,
+            iconClass: 'goldimage',
+            context: $ownerUsername,
+            altText: 'Media Contribution',
+            hasLink: false,
+        );
+
+        return;
```
Jamiras, line 369: It would be nice to have a progress counter somewhere so users know how close they are to the next tier. Maybe under Developer Stats on the profile page?

app/Models/PlayerBadge.php
```diff
@@ -79,11 +79,21 @@ class PlayerBadge extends BaseModel
         50_000_000,
     ];
 
+    private const MEDIA_CONTRIBUTION_BOUNDARIES = [
+        2,
+        10,
+        100,
+        500,
+        1000,
+        5000,
```
Jamiras: 2 seems like a very low bar for the first badge. And the jumps are very non-linear.

2->10 = x5
10 -> 100 = x10
100 -> 500 = x5
500 -> 1000 = x2
1000 -> 5000 = x5

The developer badge starts at 100 achievements, and roughly doubles at each step. https://github.com/RetroAchievements/RAWeb/blob/eac9e3c88f92d1d7a68388118c90e193020dc41a/app/Models/PlayerBadge.php#L46-L62

I'd probably go with 10 as a starting point, and x5 for the multiplier: 10, 50, 250, 1000, 5000. But that eliminates an entire tier. Maybe a 2500 could be inserted between the last two steps - it is quite a jump.

Doubling does feel like a more achievable progression, but greatly lowers the last tier: 10, 25, 50, 100, 250, 500.

Jamiras (approved):

app/Models/PlayerBadge.php
wescopeland: I was thinking 2 because that would serve as a natural incentive for filling out 1 game with primary images. What do you think?

Jamiras: That still feels like too little effort. Especially given that we support up to 20 images per game.

When I add media to random game pages, it takes 5-20 minutes (depending how far into the game I play for the in-game screenshot) to get the two primary images, box art, and a basic badge.

5 minutes for a badge would probably draw a lot of users, but how many of them would bother going for the next tier?

wescopeland: I went with a middle option in latest:
10 -> 30 -> 100 -> 300 -> 1000 -> 3000

app/Helpers/render/site-award.php
wescopeland: I agree. I'm wary about putting it in Developer Stats - it might muddy the meaning of that section.

In latest, I've put it in the badge instead and simplified the badge copy (it's self-evident):

Screenshot 2026-05-02 at 12 26 20 PM

---

Screenshot 2026-05-02 at 12 40 02 PM

Jamiras: I'm not fond of this solution. The badges should report what the user has achieved, not what they're working towards achieving. Particularly with the date in there. That's the date they hit the last tier, but reads as if it's the date the tally was last updated.

This also hides the actual tiers. "12 of 30" doesn't tell the user they've hit the "10" tier.

Maybe you could combine the old and new?
> Awarded for contributing 10 approved screenshots to game gallerys. 18 more to next tier.
> 05 Apr 2026

wescopeland: That works for me. Updated in latest.

Screenshot 2026-05-03 at 2 38 01 PM

---

Screenshot 2026-05-03 at 2 40 21 PM

Jamiras (approved):

PR #4829, opened 2026-04-28
fix(game): don't allow community screenshot uploads when there's an active claim

wescopeland:
Given dev work is ongoing, have `GameScreenshotPolicy::create()` always return false. This means players won't be able to see or use the "Upload Screenshot" button in the game page sidebar, as it could be disruptive while a dev is working on the game/set.


app/Policies/GameScreenshotPolicy.php
```diff
@@ -106,4 +117,20 @@ private function canDeveloperReviewGame(User $user, int $gameId): bool
...
+                fn ($claim) => in_array($claim->status, [ClaimStatus::Active, ClaimStatus::InReview], true)
+            );
+        }
+
+        return $game->achievementSetClaims()->activeOrInReview()->exists();
+    }
+
+    private function isSubsetGame(Game $game): bool
+    {
+        return str_contains($game->title, '[Subset') || $game->is_subset_game;
```
Jamiras: Is there a reason to do the string comparison? If so, it should probably be moved into the `Game::getIsSubsetGameAttribute` helper so other parts of the code can leverage it.

app/Policies/GameScreenshotPolicy.php
```diff
@@ -106,4 +117,20 @@ private function canDeveloperReviewGame(User $user, int $gameId): bool
             ->where('is_promoted', true)
             ->exists();
     }
+
+    private function hasActiveOrInReviewClaim(Game $game): bool
```
Jamiras: Is there a helper for this? I couldn't find the implementation.
https://github.com/RetroAchievements/RAWeb/blob/eac9e3c88f92d1d7a68388118c90e193020dc41a/app/Platform/Data/GameData.php#L72

wescopeland: My commit missed a change ☹️ 

I've reworked this in latest so it's now an attribute on the model.

app/Policies/GameScreenshotPolicy.php
wescopeland: Yep, I did this intentionally to short-circuit the query in `->is_subset_game` (`getParentGameIdAttribute()`).

In latest, I've moved this short-circuit to the model. We can probably leverage it for a while until we need to ultimately rip it out.

Jamiras (approved):

PR #4832, opened 2026-04-29
fix(set-requests): preserve selected system default labels

### LINKED ISSUE #4335: Set requests list: impossible to show "All Systems" again after changing
**Decription**
On the Set Requests page of a user, it successfully shows all systems by default, but changing the system filter and going back doesn't work.

**To Reproduce**
1. Go to https://retroachievements.org/games/requests/Mindhral > the filter is on All Systems, and all systems are shown, including unsupported ones (such as Wii or DOS to this date)
2. Select any other system, or "Only Supported Systems" > the list is updated correctly
3. Select "All Systems" again > the list is updated with all supported systems, but doesn't include DOS and Wii

**Expected behavior**
To get the games for unsupported systems again without the need to reload the whole page.

**Screenshots**
On page load:
Image
Selecting "Only Supported Systems":
Image
Selecting "All Systems":
Image

 - OS: Windows 10
 - Browser: Firefox 146.0.1

**Additional context**
Watching the requests sent using Firefox dev tools, we can see that the filter[system] argument is missing on the last call. If I resend the request adding filter[system]=all, the response contains all the expected games count.
It also shows on the filter widget, where you can only see "System" on the 3rd screenshot above instead of "System | All Systems" on the first one.

angelol-git:
fix for #4335 

For the backend, User Set Requests API would only return the "supported" value for system filter. 
It can now return "all" for /games/requests/[user].

For the frontend, Set Requests now reflects the original default value when reselected.
Previously, reselecting it would case the filter label to be empty. 
Effects ui for /games/requests/[user] and /games/requests pages.

**Before** 

https://github.com/user-attachments/assets/3b11d0bd-d59b-45db-823a-01ae491a67fe

**After**

https://github.com/user-attachments/assets/fd3a748b-b03b-4051-8b6b-7dd8f1fcbe96


resources/js/features/game-list/components/DataTableFacetedFilter/DataTableFacetedFilter.tsx
```diff
@@ -204,8 +205,8 @@ function FacetedFilterContent<TData, TValue>({
 
   const handleOptionToggle = (option: FacetedFilterOption) => {
     if (isSingleSelect) {
-      if (option.isDefaultOption) {
-        // Clear the filter when the default option is selected.
+      if (option.isDefaultOption && option.keepsExplicitFilterValue !== true) {
+        // Clear the filter when the default option is selected and ui does not reflect a specific selected value
```
wescopeland: DX: As best as I can tell, we're trying to patch this bug by adding a flag that turns off the behavior of another flag. That's a code smell.

`isDefaultOption` is already doing two unrelated things:
1. "Show me as selected when no filter is set."
2. "When clicked, clear the filter."

The bug is purely number 2. Clicking "All systems" clears the filter, and the trigger label is gated on `selectedValues.size > 0`, so it disappears.

`keepsExplicitFilterValue` papers over that. It doesn't address the root cause. The cleaner fix is to either decouple those two responsibilities, or have the trigger label fall back to rendering the default option when nothing is selected. Then the clearing branch can go away entirely.

angelol-git: Thanks, for the insight.

I split `isDefaultOption` into `clearsFilterOnSelect` for filter state behaviour and `selectedWhenEmpty` for the ui selection state.

app/Community/Controllers/Api/UserSetRequestListApiController.php
```diff
@@ -87,12 +87,12 @@ private function buildResponse(GameListRequest $request, ?User $targetUser = nul
...
     }
 
-    private function buildFilters(GameListRequest $request): array
+    private function buildFilters(GameListRequest $request, ?User $targetUser = null): array
     {
         $filters = $request->getFilters(defaultAchievementsPublishedFilter: 'none');
 
         if (!isset($filters['system'])) {
-            $filters['system'] = ['supported'];
+            $filters['system'] = $targetUser ? ['all'] : ['supported'];
```
wescopeland: nit: You're branching on the array instead of the value, which marginally muddies the intent.

Branch on the value:
```suggestion
            $filters['system'] = [$targetUser ? 'all' : 'supported'];
```

wescopeland (approved):
Thanks!

PR #4843, opened 2026-05-02
fix(PasswordRules): enforce a maximum length

wescopeland:
See: https://discord.com/channels/476211979464343552/1026595325038833725/1500148632433852606

Local Sail is unaffected due to how it's configured.


app/Support/Rules/MinimumUniqueCharacters.php
```diff
@@ -40,4 +39,36 @@ public function validate(string $attribute, mixed $value, Closure $fail): void
...
+     * characters seen so far and stop as soon as that minimum is met.
+     */
+    private function hasMinimumUniqueCharacters(string $value): bool
+    {
+        $uniqueCharacters = [];
+        $offset = 0;
+        $byteLength = strlen($value);
+
+        while ($offset < $byteLength) {
+            if (preg_match('/\G./us', $value, $matches, 0, $offset) === 1) {
```
Jamiras: I'm a bit concerned about doing this for every character, but this isn't high-traffic code, and it should short circuit after a couple handfuls of charaters.

From the code, it looks like you're trying to match multibyte characters. In that case, you could just do a much cheaper check for ASCII first:
```
$character = $value[$offset];
if (ord($character) >= 0x80 && preg_match(...) === 1) {
```

Also, there aren't any unit tests for multi-byte. I recommend having one where the number of unique bytes is greater than the minimum length, but the number of characters is not.

wescopeland: Thanks, addressed both points in latest.

Jamiras (approved):

PR #4846, opened 2026-05-02
feat(sentry): Move connect warnings from file to database

Jamiras:
This PR moves the logic for submitlbentry and the page for displaying the results. unlockachievement will be handled when it gets converted (which I plan to do after this is merged since it will share a lot of the functionality implemented for submitlbentry). The redundant write-to-file logic will be removed after unlockachievement is converted.


resources/views/pages-legacy/sentry.blade.php
```diff
@@ -0,0 +1,280 @@
...
+                        <?php endif ?>
+                    </summary>
+                    <?php
+                    unset($sniff['achievement']);
+                    unset($sniff['game']);
+                    unset($sniff['leaderboard']);
+                    unset($sniff['link']);
+                    unset($sniff['userinfo']);
+                    ?>
+                    <pre><?= json_encode($sniff, JSON_PRETTY_PRINT) ?></pre>
```
wescopeland: Careful - this may be worth escaping.

As written, someone malicious can put JS into their user agent, which this page will then execute. I successfully triggered XSS with a user agent of:
```
<img src=foo onerror=alert(1)>
```

```suggestion
                    <pre><?= e(json_encode($sniff, JSON_PRETTY_PRINT)) ?></pre>
```

app/Connect/Actions/SubmitLeaderboardEntryAction.php
```diff
@@ -41,37 +42,48 @@ protected function initialize(Request $request): ?array
...
 
-            // hash failed - ignore offset
-            $offset = 0;
+        if ($this->connectWarning) {
+            $this->connectWarning->related_type = 'leaderboard';
+            $this->connectWarning->related_id = $this->leaderboard->id;
+            $this->connectWarning->hardcore = 1; // leaderboard submissions are only allowed in hardcore
+            $this->connectWarning->offset = $request->has('o') ? (int) $request->input('o') : null;
+            $this->connectWarning->extra = $this->score;
+            $this->connectWarning->validation_hash = $request->input('v', ''); // capture unnormalized parameter
```
wescopeland: If a client sends a `v` longer than 32 chars, will this cause a server error?

Jamiras: Testing suggests it just gets silently truncated, but I do remember seeing errors in that regard. Maybe its a prod setting?

In any case, silently truncating to 32 characters could be undesirable for other reasons. I've increased the field to 40 characters, and added an `mb_strimwidth` call.

resources/views/pages-legacy/sentry.blade.php
Jamiras: I thought the `<pre>` effectively made it a code block and prevented execution. Also, this is unchanged from the previous file which hasn't encountered that issue in the time it's been around. 

I've made the suggested change.

wescopeland: Yep, understood it's unchanged. Only suggested now because the code will be publicly readable.

The XSS happens even with `<pre>`:
Screenshot 2026-05-02 at 5 54 55 PM

wescopeland (approved):

PR #4848, opened 2026-05-03
feat(manage/media): let some roles wipe all game screenshots

wescopeland:
```bash
sail artisan migrate
```

http://localhost:64000/manage/games/44/media

This PR lets admins, game editors, and media editors invoke an action to clear/reset all screenshots for a game. This effectively soft-deletes all the `game_screenshot` entities associated with the game and restores the original placeholder images.

This is a destructive action buried in a new header menu:
 
Screenshot 2026-05-03 at 3 32 24 PM

---

Screenshot 2026-05-03 at 3 32 30 PM

---

Screenshot 2026-05-03 at 3 32 36 PM

---

Screenshot 2026-05-03 at 3 32 11 PM


app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
```diff
@@ -223,6 +261,11 @@ public function table(Table $table): Table
                     ->placeholder('Published + Pending')
                     ->options(fn (): array => $this->getStatusFilterOptions()),
 
+                Tables\Filters\TrashedFilter::make()
+                    ->label('Deleted screenshots')
+                    ->placeholder('Current screenshots')
+                    ->trueLabel('All screenshots')
+                    ->falseLabel('Deleted screenshots'),
```
Jamiras: This filter feels wrong.

Image

I'm not viewing "Currently Deleted screenshots", I'm viewing non-deleted screenshots.

And I'm not sure that I like the way it combines with the other filter:

Image

I feel like Deleted should just be its own Status. What does it mean to have a Deleted Published screenshot?

app/Platform/Services/GameScreenshotValidationService.php
Jamiras, line 1: A screenshot cannot be re-uploaded if it's identical to a deleted one.

Image

app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
```diff
@@ -344,14 +388,15 @@ public function table(Table $table): Table
                         ->requiresConfirmation()
                         ->modalDescription(fn (GameScreenshot $record): string => $record->is_primary
                             ? 'This is a primary screenshot. The next published screenshot of this type will be promoted automatically, or the placeholder will be restored.'
-                            : 'Are you sure you want to delete this screenshot?')
+                            : 'Are you sure you want to permanently delete this screenshot?')
```
Jamiras: How do I get this button? I can't delete individual screenshots.

Image

app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
wescopeland: I wrestled with this too when figuring out what solution I wanted to build for this functionality. My original idea was to have an "Archived" status, but that felt awfully ambiguous with "Rejected" also in the mix.

A deleted screenshot having a status of published means, on restore, the screenshot will be published again. That makes cleanly undoing the delete possible.

Here's where I've landed in latest - no more trashed entity filter. It's disguised as a Status to hopefully be more intuitive even though the underlying model is unchanged:
Screenshot 2026-05-06 at 7 44 35 PM

app/Platform/Services/GameScreenshotValidationService.php
wescopeland: You're right. This is intentional, especially through the player-facing UI. Reasoning: if some back office action deleted the image, it shouldn't be reuploaded by a player again for back office review.

app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
wescopeland: In the previous diff, it authorized implicitly against the deletion policy (original uploader + still Pending). I think we should remove it altogether TBH, which I've done in latest.

Jamiras: I still can't delete individual screenshots.

Oh, you removed the action, not the policy restrictions.

Why shouldn't we be able to delete individual screenshots? Especially as we replace them with resolution-accurate images?

wescopeland: I guess what I'm trying to avoid is confusion between "Unpublish" and "Delete" being in that menu. Functionally to a developer, they both do the same thing.

"Rejected" is currently a status that has a valid role - the review tooling needs to be able to reject stuff, and that doesn't necessarily mean it's a trashed record.

Maybe what goes in here is just Delete and not "Unpublish".

Jamiras: Would it make sense to only be able to Delete stuff that's already been Rejected? It's an extra step that most people wouldn't do as Rejecting something already takes it off the main list.

I guess I'm struggling to define the separation between Rejected and Deleted if both can be brought back via Approve/Restore. Clearly, Rejected applies to the review tool, but what point does Deleted serve? If it's just for the "Clear Screenshots" functionality, that can just as easily mark everything as Rejected.

wescopeland: I think you're probably right.

In latest, I've:
* Removed the migration and any corresponding notion of soft deletes from the `GameScreenshot` model.
* Adjusted `ClearGameScreenshotsFromGamePageAction` so it finds all approved/pending screenshots for the game and marks them as Rejected.
* Re-added per-row `DeleteAction`, but only for rejected images.

Thanks for this feedback, I think this is probably a more intuitive mental model.

app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
```diff
@@ -340,15 +367,17 @@ public function table(Table $table): Table
...
-                        ->using(function (GameScreenshot $record) use ($game): void {
+                        ->modalHeading('Permanently delete screenshot?')
+                        ->modalDescription('This cannot be undone. The image and its history will be removed.')
+                        ->visible(fn (GameScreenshot $record): bool => $record->status === GameScreenshotStatus::Rejected)
+                        ->action(function (GameScreenshot $record) use ($game): void {
                             $screenshotUrl = $record->media?->getUrl();
                             $type = $record->type->label();
-                            $wasPrimary = $record->is_primary;
 
                             $record->media?->delete();
```
Jamiras, line 382: This deletes the `media` record, but does it actually delete the uploaded file?

Jamiras (approved):

app/Filament/Resources/GameResource/RelationManagers/GameScreenshotsRelationManager.php
wescopeland: It doesn't. Unless we intervene, the uploaded file stays in our S3 bucket forever.

We should address this, but I'd like to do so separately. MediaLibrary has an Artisan command for cleaning these up. What I'd like is to combine that with our RegenerateGameScreenshotConversations command, maybe on a weekly schedule.

PR #4849, opened 2026-05-03
feat(messages): update policy for fresh accounts

wescopeland:
Fresh accounts may no longer open multiple threads to the same team account.

**How to test this PR**
1. Create a brand new account. Verify the email address.
2. Navigate to http://localhost:64000/user/RAdmin.
3. Click the envelope and send a DM.
4. Try to navigate to RAdmin's profile, click the envelope, and send another DM. You should see that access is denied.


app/Policies/MessageThreadPolicy.php
Jamiras, line 1: When trying to create a mail from the link on the Contact Us page, the error page gets embedded in a popup?

Image

Jamiras (approved):
Functional. Slightly weird behavior that can be fixed separately if it's too complicated to slip in.

app/Policies/MessageThreadPolicy.php
wescopeland: This is an unfortunate Inertia quirk. Any time an Inertia page (Contact Us) needs to render a Blade page (errors), it bails to a dialog.

The behavior is undesirable, and it can be disabled. I've experimented with disabling it, but I remember backing off on the change (though I don't remember why - I assume I discovered some even more undesirable side effect of disabling the dialog).

PR #4850, opened 2026-05-04
feat(connect): convert awardachievement to action

Jamiras:
and moves its generated warnings to the database


app/Connect/Actions/AwardAchievementAction.php
```diff
@@ -0,0 +1,273 @@
...
+            'AchievementID' => $this->achievement->id,
+            'AchievementsRemaining' => $achievementsRemaining,
+            'Score' => $this->user->points_hardcore,
+            'SoftcoreScore' => $this->user->points,
+        ];
+
+        if ($alreadyAwarded) {
+            if ($this->hardcore && $this->user->isRanked() && $this->achievement->eventAchievements()->active()->exists()) {
+                // if event achievements are active, assume they still need to be unlocked and indicate success.
+            } else {
```
wescopeland: As best as I can tell, this drops the replay path that active event achievements depend on. Given (a) a player already has a source achievement unlocked in hardcore mode, and (b) a linked event achievement becomes active later, then this returns `Success: true` but no longer dispatches `UnlockPlayerAchievementJob`.  

This test case is green on `master` but red on this branch:
```php
public function testRepeatedHardcoreUnlockDispatchesJobWhenEventBecomesActive(): void
{
    $now = Carbon::now()->clone()->subMinutes(5)->startOfSecond();
    Carbon::setTestNow($now);

    $game = $this->seedGame(withHash: false);
    $sourceAchievement = $this->seedAchievement($game);

    $this->seedEmulatorUserAgents();
    $this->addHardcoreUnlock($this->user, $sourceAchievement, $now->clone()->subMinutes(65));

    $eventActiveFrom = $now->clone()->addDay()->startOfDay();
    System::factory()->create(['id' => System::Events]);
    $eventGame = Game::factory()->create(['system_id' => System::Events]);
    $eventAchievement = Achievement::factory()->promoted()->create(['game_id' => $eventGame->id]);

    EventAchievement::create([
        'achievement_id' => $eventAchievement->id,
        'source_achievement_id' => $sourceAchievement->id,
        'active_from' => $eventActiveFrom,
        'active_until' => $eventActiveFrom->clone()->addDay(),
    ]);

    Carbon::setTestNow($eventActiveFrom->clone()->addHour());
    Queue::fake();

    $validationHash = $this->buildValidationHash($sourceAchievement, $this->user, 1);

    $this->withHeaders(['User-Agent' => $this->userAgentValid])
        ->get($this->apiUrl('awardachievement', [
            'a' => $sourceAchievement->id,
            'h' => 1,
            'v' => $validationHash,
        ]))
        ->assertSuccessful()
        ->assertJsonPath('Success', true);

    Queue::assertPushedOn(
        'player-achievements',
        UnlockPlayerAchievementJob::class,
        function (UnlockPlayerAchievementJob $job) use ($sourceAchievement): bool {
            $tags = $job->tags();

            return
                in_array(Achievement::class . ':' . $sourceAchievement->id, $tags, true)
                && in_array(User::class . ':' . $this->user->id, $tags, true);
        },
    );
}
```

public/dorequest.php
```diff
@@ -399,14 +261,23 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
+                }
+            } else {
+                $targetUser->points += $achievement->points;
+            }
 
-            if (!isset($unlockAchievementResult['Error'])) {
-                dispatch(new UnlockPlayerAchievementJob($targetUser->id, $achievement->id, $hardcore))
-                    ->onQueue('player-achievements');
+            dispatch(new UnlockPlayerAchievementJob($targetUser->id, $achievement->id, $hardcore))
+                ->onQueue('player-achievements');
```
wescopeland, line 286: This new direct dispatch path reports unpromoted standalone achievements in the `SuccessfulIDs` field + inflates the response score.

Previously, `?r=awardachievements` invoked `unlockAchievement()`, which rejected `is_promoted=false` achievements before dispatch.

This test case is green on `master` but red on this branch:
```php
// AwardAchievementsTest.php

public function testUnpromotedAchievementIsNotAwarded(): void
{
    /** @var System $standalonesSystem */
    $standalonesSystem = System::factory()->create(['id' => System::Standalones]);
    /** @var Game $game */
    $game = Game::factory()->create(['system_id' => $standalonesSystem->id]);

    /** @var User $integrationUser */
    $integrationUser = User::factory()->create(['Permissions' => Permissions::Registered, 'connect_token' => Str::random(16)]);
    /** @var User $delegatedUser */
    $delegatedUser = User::factory()->create(['username' => 'Username', 'Permissions' => Permissions::Registered, 'connect_token' => Str::random(16)]);

    /** @var Achievement $achievement1 */
    $achievement1 = Achievement::factory()->create(['id' => 1, 'game_id' => $game->id, 'user_id' => $integrationUser->id]);

    $scoreBefore = $delegatedUser->points_hardcore;
    $softcoreScoreBefore = $delegatedUser->points;

    $params = [
        'u' => $integrationUser->username,
        't' => $integrationUser->connect_token,
        'r' => 'awardachievements',
        'k' => $delegatedUser->username,
    ];
    $payload = [
        'a' => (string) $achievement1->id,
        'h' => 1,
        'v' => md5($achievement1->id . $delegatedUser->username . '1'),
    ];

    $requestUrl = sprintf('dorequest.php?%s', http_build_query($params));
    $this->post($requestUrl, $payload)
        ->assertExactJson([
            "Success" => true,
            "Score" => $scoreBefore,
            "SoftcoreScore" => $softcoreScoreBefore,
            "ExistingIDs" => [],
            "SuccessfulIDs" => [],
        ]);
}
```

public/dorequest.php
```diff
@@ -181,144 +176,11 @@ function DoRequestError(string $error, ?int $status = 200, ?string $code = null)
...
-        if (empty($response['Score'])) {
-            $response['Score'] = $user->points_hardcore;
-            $response['SoftcoreScore'] = $user->points;
-        }
-
-        $response['AchievementID'] = $achIDToAward;
-        break;
-
     // This is only currently supported for "Standalone" integrations.
     case "awardachievements":
```
wescopeland, line 187: An unfortunate side effect of `unlockAchievement()` was it also updated `StaticData` for every successful delegated unlock. We should probably continue mutating these global stats for batch unlocks in this routine.

app/Connect/Actions/AwardAchievementAction.php
```diff
@@ -0,0 +1,273 @@
...
+            // NOTE: delegated validationStr has an extra copy of the achievement ID.
+            $validationStr = $this->achievement->id . $request->input('k', '') . ($this->hardcore ? '1' : '0') . $this->achievement->id;
+            if ($offset !== 0) {
+                $validationStr .= $offset;
+            }
+            if ($validationHash !== md5($validationStr)) {
+                return $this->accessDenied();
+            }
+        } elseif (empty($validationHash)) {
+            $this->addSmell($request, 'no_validation');
```
wescopeland: If `$validationHash` is invalid, is it desirable to retain the given value of `o`?

app/Connect/Actions/AwardAchievementAction.php
wescopeland, line 1: DX: I think it's desirable to still have some kind of centralized `unlockValidationHash()` / `submitValidationHash()` logic. This is reimplemented in several places in the diff.

Jamiras: It should be implemented in two places: In the action and in the test. I specifically wanted separate implementations to ensure the tests fail if the active one gets accidentally changed.

I actually moved it from the Achievement/Leaderboard objects to the action specifically because it's specific to the Connect API call. Maybe you're seeing the removed references as well?

app/Connect/Actions/AwardAchievementAction.php
Jamiras: No. Probably not. Updated and added unit test.

app/Connect/Actions/AwardAchievementAction.php
Jamiras: Yes, I agree that this change appears to skip pushing the job, yet there is test showing the achievement _is_ being unlocked.
```
    test('repeated hardcore unlock still unlocks event achievement', function () {
        ...
        $this->assertHasHardcoreUnlock($this->user, $eventAchievement);
        ...
    });
```

public/dorequest.php
Jamiras: done.

public/dorequest.php
Jamiras: Good call. Added test and made appropriate changes.

app/Connect/Actions/AwardAchievementAction.php
Jamiras: It's a problem in the test. Because the prior achievement unlock was in the active period of the event, the event achievement was immediately unlocked before the API call was made.

Test corrected, code updated. Thanks for spotting this.

app/Connect/Actions/AwardAchievementAction.php
wescopeland: > I actually moved it from the Achievement/Leaderboard objects to the action specifically because it's specific to the Connect API call.

Locality of behavior is a fair reasoning. I can get on board with that. If we re-use these elsewhere at some point, maybe that's the time to consider extraction.

wescopeland (approved):

PR #4854, opened 2026-05-05
fix(connect/sentry): escape a few more user-injectable strings

wescopeland:
Patching a few more of these from the page now that the code will be publicly readable.


resources/views/pages-legacy/sentry.blade.php
```diff
@@ -226,7 +226,7 @@ function toggle(S, x) {
                 $classes = [
                     'smell',
                     'user-' . key_to_class($sniff['user']),
-                    'method-' . $sniff['method'],
+                    'method-' . key_to_class($sniff['method']),
```
Jamiras, line 229: `method` should be safe, as these entries won't be generated unless the method gets mapped to an action that supports generating them.

It doesn't hurt to be cautious though.

resources/views/pages-legacy/sentry.blade.php
```diff
@@ -235,10 +235,10 @@ function toggle(S, x) {
                 ?>
                 <details class="<?= implode(' ', $classes) ?>">
                     <summary class="cursor-pointer mb-2">
-                        <code><?= $sniff['date'] ?></code>
-                        <code style="padding:1px 5px;border:1px solid steelblue;background:royalblue;color:white"><?= $sniff['method'] ?></code>
+                        <code><?= e($sniff['date']) ?></code>
```
Jamiras: How would someone inject stuff into a timestamp field?

https://github.com/RetroAchievements/RAWeb/blob/701893db29ae342d1006fee6006d1c2df0790e9c/app/Connect/Actions/BuildConnectSniffsAction.php#L35

resources/views/pages-legacy/sentry.blade.php
```diff
@@ -235,10 +235,10 @@ function toggle(S, x) {
...
+                        <code style="padding:1px 5px;border:1px solid steelblue;background:royalblue;color:white"><?= e($sniff['method']) ?></code>
                         <?php foreach ($sniff['smells'] ?? [] as $smell): ?>
-                            <code style="padding:1px 5px;border:1px solid <?= $colors[$smell]['border'] ?? 'gray' ?>;background:<?= $colors[$smell]['background'] ?? 'gray' ?>;color:white"><?= $smell ?></code>
+                            <code style="padding:1px 5px;border:1px solid <?= e($colors[$smell]['border'] ?? 'gray') ?>;background:<?= e($colors[$smell]['background'] ?? 'gray') ?>;color:white"><?= e($smell) ?></code>
```
Jamiras: The `colors` array is hard-coded and not user-injectable.

resources/views/pages-legacy/sentry.blade.php
```diff
@@ -13,8 +13,8 @@
 function RenderFilterButton(string $text, string $filter, int $count): void
 {
     if ($count > 0) {
-        echo "<button class='btn' type='button' data-filter='$filter'>";
-        echo "$text ($count)";
+        echo "<button class='btn' type='button' data-filter='" . e($filter) . "'>";
```
Jamiras: `$filter` will have already been sanitized by `key_to_class`. This may cause a double-escape that would prevent things from being matched by the `toggleFilter` function.

resources/views/pages-legacy/sentry.blade.php
```diff
@@ -13,8 +13,8 @@
...
+        echo e($text) . " ($count)";
```
Jamiras, line 17: IMO, this is the only thing that really needs to be escaped.

resources/views/pages-legacy/sentry.blade.php
```diff
@@ -25,7 +25,7 @@ function key_to_class(?string $key): string
         return '_null_';
     }
 
-    return str_replace([' ', '.'], '-', str_replace('%', '_pct_', $key));
+    return preg_replace('/[^A-Za-z0-9_-]/', '-', str_replace('%', '_pct_', $key)) ?? '';
```
wescopeland, line 28: I hardened this a little bit in latest - it's technically possible to break out of the current implementation and inject custom attributes.

resources/views/pages-legacy/sentry.blade.php
wescopeland: Too broad - removed this.

resources/views/pages-legacy/sentry.blade.php
wescopeland: Too broad - removed this.

Jamiras (approved):

PR #4860, opened 2026-05-06
perf: skip achievement metric recounts for live player count updates

wescopeland:
Addresses [RAWEB-1QM](https://retroachievementsorg.sentry.io/issues/89269143/?environment=production&project=4508694685417552&referrer=issue-list&statsPeriod=90d).

When a player's unlock count for a game becomes non-zero, we do two different jobs worth of work:
1. We update the game's player count. "The game now has 80,001 players instead of 80,000."
2. We recount the stats for every achievement on the game's set from scratch. "For achievement X, count all unlocks. For achievement Y, count all unlocks. For achievement Z... etc".

That second step is the slow `player_achievements GROUP BY achievement_id` query from RAWEB-1QM.

This PR makes a minor adjustment: for these player count driven cascades, we'll no longer recount every achievement's unlock totals. Instead, we'll use the unlock totals already cached on each achievement's row and just recompute the unlock percentage, unlock hardcore percentage, and weighted points.

In other words, this PR introduces a "fast track" path for when only the player count changes. The "slow track" path (current behavior) is still present for:
- achievements promoted / demoted
- achievement transfers
- resets
- players getting untracked / retracked


app/Platform/Actions/UpdateAchievementMetricsAction.php
```diff
@@ -12,6 +12,8 @@
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 
+// TODO either convert to a service, or refactor so there's only one publicly-exposed function
```
Jamiras, line 15: The second call is here: https://github.com/RetroAchievements/RAWeb/blob/e636b693b2a01cc039923d9faca48194199b635a/app/Platform/Actions/UpdateGameAchievementsMetricsAction.php#L22

The easiest way to eliminate the second public method would be to have that class inherit from this one and make `update` protected. I think that would still work with the additional changes you've made there.

Conceptually, this class works on an array of achievements for a single game such that we can do bulk updates, but the primary interface accepts a single achievement (which is called through the `DispatchUpdateAchievementsMetricsJob` triggered by unlocking and resetting achievements and point changes).

app/Platform/Jobs/UpdateGamePlayerCountJob.php
```diff
@@ -24,16 +24,22 @@ class UpdateGamePlayerCountJob implements ShouldQueue, ShouldBeUniqueUntilProces
...
     }
 
     public int $uniqueFor = 3600;
 
     public function uniqueId(): string
     {
-        return config('queue.default') === 'sync' ? '' : $this->gameId;
+        return config('queue.default') === 'sync'
+            ? ''
+            : $this->gameId . ':' . ($this->shouldRecalculateAchievementUnlockCounts ? 'recount' : 'stored-counts');
```
Jamiras, line 42: This change implies a similar change should be made to the other jobs in the chain. Yes?

app/Platform/Listeners/DispatchUpdatePlayerGameMetricsJob.php
```diff
@@ -15,15 +15,20 @@ public function handle(object $event): void
...
         $user = null;
         $game = null;
-        // TODO forward hardcore flag
-        $hardcore = null;
+
+        /**
+         * Keep full recounts unless another job is already refreshing the changed unlock total.
+         * Live unlocks dispatch UpdateAchievementMetricsJob separately, so their player count
+         * cascade only needs to recompute percentages from stored unlock counts.
+         */
```
Jamiras, line 30: Are you sure about this? The only thing I see in `UnlockPlayerAchievementAction` is a `PlayerAchievementUnlocked` notification - which is what is being behaviorally modified here.

app/Platform/Actions/UpdateAchievementMetricsAction.php
wescopeland: Truthfully, I'm a bit anxious about touching this in this PR.

I added this comment because my change adds a 3rd public method in an attempt to be minimally disruptive - it isn't ideal. A refactor to slim down to one public method may be risky as part of this change. I'm hesitant to roll it into a perf-focused branch.

app/Platform/Jobs/UpdateGamePlayerCountJob.php
wescopeland: Which jobs specifically are you thinking of?

The uniqueness split is already on the two jobs that carry the fast/slow track behavior: `UpdatePlayerGameMetricsJob` and `UpdateGamePlayerCountJob`.

There's a sibling chain through `DispatchUpdateAchievementMetricsJob` & `UpdateAchievementMetricsJob`, but that doesn't have a fast/slow track, so I'm not sure it benefits from the same uniqueness split.

app/Platform/Listeners/DispatchUpdatePlayerGameMetricsJob.php
wescopeland: Tricky to get the wording on this one dialed in. Updated in latest to:
```php
/**
 * PlayerAchievementUnlocked kicks off several separate metrics updates.
 *
 * UpdateAchievementMetricsJob recounts the unlocked achievement's
 * denormalized totals from player_achievements. This player game metrics
 * path may also update the game's player count, which means aggregate
 * achievement percentages and RetroPoints (weighted points) need to be
 * refreshed for the game.
 *
 * Since the achievement job owns the expensive unlock recount, this path
 * can reuse the stored unlock totals and avoid running that same recount again.
 */
$shouldRecalculateAchievementUnlockCounts = true;
```

app/Platform/Jobs/UpdateGamePlayerCountJob.php
Jamiras: I might be getting Jobs and Actions conflated. I just saw the parameter get added to a whole bunch of actions, but the uniqueId was not update in their corresponding jobs. As long as the actions continue to call actions, the job probably doesn't need to be updated.

app/Platform/Listeners/DispatchUpdatePlayerGameMetricsJob.php
Jamiras: > Since the achievement job owns the expensive unlock recount, this path can reuse the stored unlock totals and avoid running that same recount again.

Can you be certain the expensive calculations will be finished before this path tries to use the results?

wescopeland: Nope. You're right to question this assumption.

In latest, I've changed this path so it re-reads the stored counts before actually using them. From there, we guard the update on those same counts still being present. If the recount wins the race and changes the stored counts first, then this path skips that row and falls back to doing a normal full recount for it.

Jamiras (approved):

PR #4868, opened 2026-05-09
feat: add developer inactivity alerts

wescopeland:
https://discord.com/channels/476211979464343552/1002688331827658774/1498772708413149294

This PR adds support for surfacing developer inactivity alerts into a Discord feed channel. Users with a junior developer, moderator, or admin role are excluded from this check.

The alert checks for:
* overall site activity (`users.last_updated_at`)
* publishing new memory notes, achievements, or leaderboards
* updating any kind of trigger logic
* creating a new claim of any kind
* commenting on a ticket
* resolving a ticket

If the developer is site-inactive for 3 months, an alert will surface. If the developer doesn't have any of the above activities recorded for 6 months, an alert will surface.

The alerts are driven by a nightly scheduled command. Results are cached in Redis so the same alert doesn't pop multiple times.


app/Platform/AppServiceProvider.php
```diff
@@ -143,19 +145,21 @@ public function boot(): void
...
-                $schedule->command(UpdateBeatenGamesLeaderboard::class)->everyFiveMinutes();
-                $schedule->command(UpdatePlayerPointsStats::class, ['--existing-only'])->hourly();
-                $schedule->command(ProcessExpiringClaims::class)->hourly();
-                $schedule->command(UpdateDeveloperContributionYield::class)->weeklyOn(2, '10:00'); // Tuesdays at 10AM UTC
-                $schedule->command(CrawlPlayerWeightedPoints::class)->weeklyOn(3, '10:00'); // Wednesdays at 10AM UTC
-                $schedule->command(CheckForAchievementSetChanges::class)->daily();
-            }
+            $schedule->command(DeleteStalePlayerPointsStatsEntries::class)->weekly();
+            $schedule->command(UpdateDeveloperContributionYield::class)->weeklyOn(2, '10:00'); // Tuesdays at 10AM UTC
+            $schedule->command(CrawlPlayerWeightedPoints::class)->weeklyOn(3, '10:00'); // Wednesdays at 10AM UTC
```
wescopeland, line 162: Sorted these around and removed the production conditional on several of them (no longer needed now that we have a dedicated stage server).

app/Platform/Actions/DetectInactiveDevelopersAction.php
```diff
@@ -0,0 +1,205 @@
...
+        $alertEntries = [];
+
+        $developers = User::query()
+            ->whereNull('banned_at')
+            ->whereHas('roles', fn (Builder $query) => $query->where('name', Role::DEVELOPER))
+            ->whereDoesntHave('roles', fn (Builder $query) => $query->whereIn('name', [
+                Role::ADMINISTRATOR,
+                Role::DEVELOPER_JUNIOR,
+                Role::MODERATOR,
+            ]))
```
Jamiras: If we have Roles set up right, then this secondary check shouldn't be necessary. A moderator/developer can still be a moderator without the developer role.

app/Platform/Actions/DetectInactiveDevelopersAction.php
```diff
@@ -0,0 +1,205 @@
...
+            ),
+
+            // resolving a ticket
+            $this->maxTimestamp(
+                Ticket::query()
+                    ->where('resolver_id', $developer->id)
+                    ->whereNotNull('resolved_at'),
+                'resolved_at',
+            ),
+        ]);
```
Jamiras, line 202: Ideally, this would short-circuit if it found a value more recent than the threshold. I imagine most of these are not cheap queries.

app/Platform/AppServiceProvider.php
Jamiras: I assumed the environment check was so they don't run in the local environment. I hadn't really considered not wanting to run them in stage.

wescopeland: Yep, purely here as a guard against stage.

These shouldn't ever run locally. On our app server, we have a cron that runs every 60 seconds:
```
php artisan schedule:run
```
This Laravel built-in command checks if there are any scheduled commands that are due for execution and executes what's ready to go.

app/Platform/Actions/DetectInactiveDevelopersAction.php
wescopeland: Added short-circuiting in latest.

app/Platform/Actions/DetectInactiveDevelopersAction.php
wescopeland: I'm an inactive developer. But it's reasonable for me to have the role in production for as long as I'm a member of technical staff.

In latest, I've narrowed the secondary check to look for root instead.

Jamiras (approved):

PR #4874, opened 2026-05-10
feat(connect): move rate limiting to Laravel

wescopeland:
https://discord.com/channels/310192285306454017/310195377993416714/1503024264465350786

This PR sets up rate limiting for the Connect API in the Laravel layer. Right now, it's managed in nginx, which is insufficient for our needs. We need per-routine limiting, which can't work in nginx when the routine often lives in a `POST` body.

There are now three limits:
* Publishing (600/m)
* Default (180/m)
* Login (5/m)

180/m for the default limit sounds high, but it's actually lower than what nginx has been set to nearly this whole time.

Once this is merged and deployed, we can remove the nginx limit on _dorequest.php_ from the app server.


app/Connect/RouteServiceProvider.php
```diff
@@ -5,15 +5,40 @@
...
+use Symfony\Component\HttpFoundation\Response;
 
 class RouteServiceProvider extends ServiceProvider
 {
     use HandlesPublicFileRequests;
 
+    private const array DEVELOPER_PUBLISH_REQUEST_TYPES = [
+        'submitcodenote',
+        'submitgametitle',
+        'submitrichpresence',
```
Jamiras: `submitgametitle` and `submitrichpresence` can probably use the default rate limiting logic. 

`submitgametitle` is only used to associate unknown hashes to games, and with the rare exception of a multi-disc game, probably won't be called more than once every few weeks by a single developer. And even with a multi-disc game, the developer has to load each disk separately, select the game to associate and then submit the association. I don't see it as possible to do that quickly enough to trigger the rate limiter.

`submitrichpresence` updates the rich presence for a game. A game can only have one rich presence, so it shouldn't be possible to bulk publish, and won't likely run into any sort of rate limiting issues.

app/Connect/RouteServiceProvider.php
```diff
@@ -5,15 +5,40 @@
...
+
+    private const array LOGIN_REQUEST_TYPES = [
+        'login',
+        'login2',
+    ];
+
+    private const int PUBLISH_PER_MINUTE = 600;
+    private const int DEFAULT_PER_MINUTE = 180;
+    private const int LOGIN_PER_IP = 30;
+    private const int LOGIN_PER_USER_IP = 5;
```
Jamiras, line 43: I might be affected by this. While I'm iterating on DLL changes, I may relaunch several times in a short period, with each performing a login. But I'm not sure if it would be frequently enough to trigger this limit.

app/Connect/RouteServiceProvider.php
wescopeland: Removed those two routines from the list in latest.

app/Connect/RouteServiceProvider.php
wescopeland: In latest, successful logins will no longer count against the rate limit.

wescopeland:
Additional change made in latest: delegated Connect API calls (eg: HorizonXI) now have a generous bucket.

Jamiras (approved):

PR #4878, opened 2026-05-12
feat(forum): add support for comment_role_id field

wescopeland:
https://discord.com/channels/310192285306454017/1499468217654051137/1503387011774550116

```
sail artisan migrate
```

This PR adds a nullable `comment_role_id` field to `forum_topics`. When it's set:
* Normal (non-staff) users may no longer comment on the forum topic - it's effectively locked for them.
* Users with the role may reply as themselves.
* Users with the role may reply as any team account they have access to that also has the role.
* Moderators bypass the role gate and can still always reply.

At the moment, this can only be set on a topic by manually doing it in the database. It may be desirable in the future to make this editable in the front-end moderation controls. This is a QoL concern that I think that would more than double the size of this PR, so I decided to punt it for now.

**As a QA member when the topic has `comment_role_id=25`**
Screenshot 2026-05-12 at 7 23 01 PM

**As a regular user when the topic has `comment_role_id=25`**
Screenshot 2026-05-12 at 7 23 47 PM


app/Helpers/database/forum.php
Jamiras, line 1: Should users still be allowed to edit posts if this setting is on and they don't have the desired roles? Actually locked threads prevent that.

wescopeland: That's a good catch - for this use case it's probably fine to keep it permissive as-is. What's actually going to happen is only users with these roles are going to be able to post in the thread from inception.

Jamiras (approved):

PR #4882, opened 2026-05-13
provide discord hook for connect warnings

Jamiras:
I'm assuming this will point to the same place as the other warning. I did not test the actual discord integration, but the code is largely copied from the existing warning and the unit tests suggest everything should behave the same.


app/Connect/Support/GeneratesConnectWarnings.php
```diff
@@ -22,6 +24,7 @@ public function handleRequest(Request $request): JsonResponse
 
         if ($this->connectWarning !== null) {
             $this->finalizeWarning();
+            $this->sendNotifications();
             $this->connectWarning->save();
```
wescopeland, line 28: DX, nit: recommend saving the warning before queuing the alert.

app/Support/Alerts/SuspiciousConnectWarningAlert.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public function toDiscordMessage(): string
+    {
+        $user = User::whereName($this->warning->username)->firstOrFail();
+
+        $gameId = match ($this->warning->method) {
+            'submitlbentry' => Leaderboard::findOrFail($this->warning->related_id)->value('game_id'),
+            default => Achievement::findOrFail($this->warning->related_id)->value('game_id'),
+        };
```
wescopeland, line 27: Bug: calling `->value('game_id')` on a model instance doesn't read the attribute. This fires a second query like `SELECT game_id FROM achievements LIMIT 1` with no constraints and returns `game_id` of whatever achievement happens to be first in the table, not the one we've just loaded.

The fix is trivial:
```php
$gameId = match ($this->warning->method) {
    'submitlbentry' => Leaderboard::find($this->warning->related_id)?->game_id,
    default => Achievement::find($this->warning->related_id)?->game_id,
};
```

app/Connect/Support/GeneratesConnectWarnings.php
```diff
@@ -87,4 +90,19 @@ private function finalizeWarning(): void
...
+        if (str_contains($this->connectWarning->smells, 'repeated_validation')
+            || str_contains($this->connectWarning->smells, 'wrong_client')) {
+
+            // only send one notification per user per day
+            $key = 'user:' . strtolower($this->connectWarning->username) . ':connect_warning_notification';
+            if (!Cache::has($key)) {
+                Cache::put($key, '1', 24 * 60 * 60);
+
+                (new SuspiciousConnectWarningAlert($this->connectWarning))->send();
+            }
```
wescopeland: Recommend using `Cache::add()`, which combines `has()` and `put()` in a single operation to avoid a race if two events fire simultaneously.
```php
if (Cache::add($key, true, now()->addDay())) {
    (new SuspiciousConnectWarningAlert($this->connectWarning))->send();
}
```

app/Connect/Actions/AwardAchievementAction.php
```diff
@@ -167,6 +168,19 @@ protected function process(): array
...
         }
 
+        $playerSession = app()->make(ResumePlayerSessionAction::class)->execute(
+            $this->user,
+            $this->achievement->game,
+            ($this->gameHash && !$this->gameHash->isMultiDiscGameHash()) ? $this->gameHash : null,
+            timestamp: $this->when,
+            userAgent: $this->userAgent,
+            ipAddress: $this->ipAddress,
+        );
```
wescopeland, line 178: Perf: For normal unlocks are we now resuming sessions twice? Should this be within `if ($this->connectWarning)`?

app/Connect/Support/GeneratesConnectWarnings.php
Jamiras: done

app/Connect/Support/GeneratesConnectWarnings.php
Jamiras: done

app/Support/Alerts/SuspiciousConnectWarningAlert.php
Jamiras: I was trying to avoid loading the whole model, but `findOrFail` probably does that anyway.

Updated.

app/Connect/Actions/AwardAchievementAction.php
wescopeland: Sorry to nag, still confused about this one. It seems like we may be resuming sessions twice (which may be desirable, I'd just like to understand it better).

Jamiras: It has to be resumed at this point to attach to the connect warning, but could be moved inside the if.
Calling it here also provides more context to the session that's not being passed through the UnlockAchievementJob (but probably should be).

wescopeland (approved):

PR #4883, opened 2026-05-13
still use core set player count for subset retroratio calculations

Jamiras:
All subsets now report player counts and unlock rates specific to the subset.

Subset RetroRatio calculations will now use the player count from the parent set.

Note: this re-introduces the bug where a subset with more players than the parent set looks to have unlock ratios above 100%. Since it's buried inside the RetroRatio calculation, I don't know how pressing it would be to fix that. I don't feel like it's important enough to run a query to get the deduplicated count.


app/Platform/Actions/UpdateGamePlayerCountAction.php
```diff
@@ -5,39 +5,19 @@
...
-        // so we COUNT(DISTINCT user_id) there instead of SUM-ing rows directly
-        $isMultiGame = count($gameIds) > 1;
-        $countExpr = fn (string $column): string => $isMultiGame
-            ? "COUNT(DISTINCT CASE WHEN player_games.{$column} > 0 THEN player_games.user_id END)"
-            : "SUM(CASE WHEN player_games.{$column} > 0 THEN 1 ELSE 0 END)";
+        $countExpr = fn (string $column): string => "SUM(CASE WHEN player_games.{$column} > 0 THEN 1 ELSE 0 END)";
 
         $row = PlayerGame::query()
-            ->whereIn('player_games.game_id', $gameIds)
+            ->where('player_games.game_id', $game->id)
```
wescopeland, line 20: It looks like a few tests may actually be failing even though they're not reported by CI:
```
❯ composer test tests/Feature/Platform/Actions/UpdateGamePlayerCountActionTests.php

   FAIL  Tests\Feature\Platform\Actions\UpdateGamePlayerCountActionTests
  ✓ core set → constructs player count from player games records                                                                                  0.29s
  ✓ core set → ignores unranked users                                                                                                             0.07s
  ✓ core set → ignores bonus subset players                                                                                                       0.08s
  ⨯ subset → constructs player count from game and bonus subset records                                                                           0.08s
  ✓ subset → constructs player count from game and challenge subset records                                                                       0.08s
  ✓ subset → constructs player count from game and specialty subset records                                                                       0.08s
  ✓ subset → constructs player count from game and exclusive subset records                                                                       0.08s
  ⨯ subset → constructs player count from game and bonus subset records ignoring duplicates                                                       0.08s
  ⨯ subset → constructs player count from multiple parent games and bonus subset records                                                          0.09s
```

`composer test -- --parallel` isn't picking up the test file. I can only get it to run when I invoke it by path. I think it's because the file name is `*Tests.php` instead of `*Test.php`.

Jamiras: well, that's unfortunate. good catch. renamed and updated.

wescopeland (approved):

PR #4891, opened 2026-05-15
perf: denormalize games.parent_game_id

wescopeland:
Follow-up to hotfix from https://github.com/RetroAchievements/RAWeb/pull/4887.

This PR adds a `games.parent_game_id` denormalized field and largely deletes the hotfix code. The field is written in-place during the migration itself. Observers keep the field synced.

It's a known limitation that the field is lossy. In other words, only one game ID is supported in this field, so subsets like Pokemon R/B's won't necessarily be accurate. I think the trade-off is fine for the time being.


app/Models/Game.php
```diff
@@ -738,7 +683,7 @@ public function getIsSubsetGameAttribute(): bool
             return true;
         }
 
-        return $this->parentGameId !== null;
+        return $this->parent_game_id !== null;
```
Jamiras, line 681: The short-circuit can be removed now.

app/Models/Game.php
```diff
@@ -951,6 +894,14 @@ public function visibleModificationsComments(?User $user = null): HasMany
         return $this->modificationsComments()->visibleTo($currentUser);
     }
 
+    /**
+     * @return BelongsTo<Game, $this>
+     */
+    public function parent(): BelongsTo
+    {
+        return $this->belongsTo(Game::class, 'parent_game_id');
+    }
```
Jamiras: Is there a reason to add this instead of updating `parentGame`? As best as I can tell, there's only two references to `parentGame()`, and those can be updated by just removing the parentheses.

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
Jamiras, line 1: I'm not sure if this isn't working, or if I'm missing something.

* Open game 4: http://localhost:64000/manage/games/4?relation=1
* Associate existing subset 25811
* View game page: http://localhost:64000/game/4 (subset visible)
* Query database:
```
MariaDB [retroachievements-web]> select parent_game_id from games where id=25811;
+----------------+
| parent_game_id |
+----------------+
|          25807 |
+----------------+
1 row in set (0.001 sec)
```
I think it should return 4.

app/Platform/Actions/UpdateAchievementMetricsAction.php
```diff
@@ -42,8 +42,8 @@ public function update(Game $game, Collection $achievements): void
 
         // if game has a parent game, fetch the parent game's players metrics
         $retroRatioPlayerCount = $playersHardcore;
-        if ($game->parentGameId) {
-            $retroRatioPlayerCount = Game::find($game->parentGameId)->players_hardcore ?? 0;
+        if ($game->parent_game_id) {
+            $retroRatioPlayerCount = Game::find($game->parent_game_id)->players_hardcore ?? 0;
         }
```
Jamiras: Might be better to use relationship in case it's already loaded?
```
$game->loadMissing('parentGame');
if ($game->parentGame) {
    $retroRatioPlayerCount = $game->parentGame->players_hardcore ?? 0;
}
```

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
wescopeland: This is confusing, and your expectation is reasonable, but I think what's happening right now that you're seeing is the correct behavior.

> It's a known limitation that the field is lossy. In other words, only one game ID is supported in this field, so subsets like Pokemon R/B's won't necessarily be accurate. I think the trade-off is fine for the time being.

In this case, game 25807 (the real parent game where this subset was first attached) is the earliest non-core link as the parent. Therefore, it "wins" over the new one to game 4. This preserves existing prod behavior.

app/Models/Game.php
wescopeland: Done in latest.

app/Models/Game.php
wescopeland: Updating `parentGame` is better. Done in latest.

app/Platform/Actions/UpdateAchievementMetricsAction.php
wescopeland: Updated in latest.

app/Filament/Resources/GameResource/RelationManagers/AchievementSetsRelationManager.php
Jamiras: Ah, I was expecting the behavior to move the subset. Detaching from the other game did work.

Jamiras (approved):

PR #4894, opened 2026-05-15
fix: add horizontal scrolling for legacy page tables

angelol-git:
Some legacy pages had tables that would cause horizontal overflow on smaller width screens. Notably in `globalRanking.blade.php` and `leaderboardinfo.blade.php`.

Used an existing `table-wrapper` class as used in `achievementList.blade.php` and other blades.

**Before**

https://github.com/user-attachments/assets/b946a736-3dbc-4356-a884-5622d6fb7579

**After**

https://github.com/user-attachments/assets/2e88dbbd-45ee-4a9c-83b4-0458bf2d00d2


wescopeland (approved):
Thank you!

Issue #4900, opened 2026-05-18
User Progression Status displaying the wrong "Most Recent" System


angelol-git:
Originally From: https://discord.com/channels/310192285306454017/1081973869994320002/threads/1504361895552090182

**Describe the bug**
When a users last played game is from a system with no unlocked achievements. It can fall back to a different unexpected system.

**To Reproduce**
1. Add a game to a users play history with the following:
- is the last played game 
- from a system the user has not yet played
- has an achievement set but no earned achievements 

**Screenshots**

<img width="1706" height="474" alt="Image" src="https://github.com/user-attachments/assets/341790c4-098c-430c-a75b-6fd5e85e3aeb" />

<img width="1746" height="1190" alt="Image" src="https://github.com/user-attachments/assets/76db58d5-c2cc-4fa5-88bd-d531053363ac" />

**Expected behavior**
Last played system should either be `N64` or `PSP` depending on criteria . 

**Findings**
Digging around I believe the problem starts here:
https://github.com/RetroAchievements/RAWeb/blob/d40e4a74da497de153cff7354d18f2125ddb89cb/resources/views/pages-legacy/userInfo.blade.php#L46-L50

The `recentlyPlayedSystemId` is collected if the user's last played game (by default looks at last 5 games) is from a valid system + has an achievement set. It doesn't matter if they have earned any achievements on that game.

`recentlyPlayedSystemId` gets passed to:
https://github.com/RetroAchievements/RAWeb/blob/d40e4a74da497de153cff7354d18f2125ddb89cb/app/Platform/Actions/GetUserProgressionStatusCountsAction.php#L129-L131

where `recentSystemId` is checked in `systemProgress`. But `systemProgress` only contains the users last played games with **earned** achievements. 

Otherwise it falls back to `array_key_first($systemProgress)` which can give unexpected results. `systemProgress` default order is `systemId` from lowest to greatest. So `$topSystemId ` equals the lowest `systemId` the user has earned achievements on. Which is often the Genesis/Mega Drive

**Solution**
I can implement a solution but just needed clarification if the "Most Recent" system should be from a game with earned achievements or **no** earned achievements.

wescopeland:
Hi!

For a solution, "Most Recent" should fall back to the most recent system among games that have earned achievements. In the case of the screenshots, it should be PSP rather than N64.

PR #4903, opened 2026-05-18
fix(user): resolve Most Recent System progression

angelol-git:
fix for: #4900 

Added `lastPlayedAt` key to `systemProgress`. 

So recently played `topSystemId` now has a single source of truth from `systemProgress`.

**Before**
Screenshot 2026-05-18 at 5 32 57 PM

**After**
Screenshot 2026-05-18 at 5 33 57 PM


app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -126,9 +130,13 @@ public function execute(User $user, ?int $recentSystemId = null): array
...
-        $topSystemId = ($recentSystemId !== null && isset($systemProgress[$recentSystemId]))
-            ? $recentSystemId
-            : array_key_first($systemProgress);
+        // topSystem is the most recent valid system where the user has earned achievements.
+        $topSystemId = !empty($systemProgress)
+            ? collect($systemProgress)
+                ->sortByDesc(fn ($progress) => strtotime($progress['lastPlayedAt'] ?? '') ?: 0)
+                ->keys()
+                ->first()
+            : null;
```
wescopeland: Perf: This is working harder than needed at O(n log n) complexity.

If we only need the single newest item, we don't need to fully order the whole list. To optimize, try walking the list once and remembering the best candidate seen so far.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -126,9 +130,13 @@ public function execute(User $user, ?int $recentSystemId = null): array
...
         $avgCompletionPercentage = $this->calculateAvgCompletionExcludingSubsets($user);
 
-        $topSystemId = ($recentSystemId !== null && isset($systemProgress[$recentSystemId]))
-            ? $recentSystemId
-            : array_key_first($systemProgress);
+        // topSystem is the most recent valid system where the user has earned achievements.
+        $topSystemId = !empty($systemProgress)
+            ? collect($systemProgress)
+                ->sortByDesc(fn ($progress) => strtotime($progress['lastPlayedAt'] ?? '') ?: 0)
+                ->keys()
```
wescopeland: Intellectual curiosity: Does this take into account data tracked by `addOrphanBadgeCounts()`?

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -18,14 +18,15 @@ class GetUserProgressionStatusCountsAction
     /**
      * @return array{
      *     avgCompletionPercentage: float,
-     *     systemProgress: array<int, array{unfinishedCount: int, beatenSoftcoreCount: int, beatenHardcoreCount: int, completedCount: int, masteredCount: int, systemId: int}>,
+     *     systemProgress: array<int, array{unfinishedCount: int, beatenSoftcoreCount: int, beatenHardcoreCount: int, completedCount: int, masteredCount: int, lastPlayedAt: string|null, systemId: int}>,
+     *
```
wescopeland:
```suggestion
```

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -100,13 +102,15 @@ public function execute(User $user, ?int $recentSystemId = null): array
             $beatenHc = (int) $row->beaten_hc_count;
             $beatenSc = (int) $row->beaten_sc_count;
             $unfinished = (int) $row->unfinished_count;
+            $lastPlayedAt = (string) $row->last_played_at;
```
wescopeland: Docs: The docblock states `lastPlayedAt: string|null`, but `(string) null` is `""`. I don't believe `lastPlayedAt` will ever be `null`.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
angelol-git: Yes. With the current logic, orphaned systems sort below any system with a real lastPlayedAt, but they can still technically end up as `$topSystemId`.

Should an orphaned system be eligible as a topSystem? I am assuming not.

wescopeland: They shouldn't be eligible - I agree.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
angelol-git: Updated in latest. Changed to a O(n) approach.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
angelol-git: Updated in latest. Switched this to a null safe date string conversion so it stays consistent with the docblock and the rest of the codebase.

app/Platform/Actions/GetUserProgressionStatusCountsAction.php
```diff
@@ -126,9 +130,22 @@ public function execute(User $user, ?int $recentSystemId = null): array
...
 
-        $topSystemId = ($recentSystemId !== null && isset($systemProgress[$recentSystemId]))
-            ? $recentSystemId
-            : array_key_first($systemProgress);
+        // topSystem is the most recent valid system where the user has earned achievements.
+        $topSystemId = null;
+        $latestTimestamp = null;
+
+        foreach ($systemProgress as $systemId => $progress) {
+            if ($progress['isOrphan'] || $progress['lastPlayedAt'] === null) {
```
wescopeland: DX: My sense is `isOrphan` exists only to gate this one check, but it's now a part of the documented return shape, so we're leaking an implementation detail.

Could we compute `$topSystemId` before `addOrphanBadgeCounts()`? At that point, `$systemProgress` only contains rows from the qualifying `player_games` query, so `lastPlayedAt === null` should be enough to exclude entries that cannot be "Most Recent".

This way, `addOrphanBadgeCounts()` can stay as a counts-only supplement, and we can drop `isOrphan` from the return shape/docblock entirely.

angelol-git: Updated in latest. `$topSystemId` is now computed before `addOrphanBadgeCounts()`, and `isOrphan` was removed.

wescopeland (approved):

PR #4908, opened 2026-05-20
fix: prevent self-referential EventAchievement infinite loops

wescopeland:
https://discord.com/channels/476211979464343552/1026595325038833725/1506645623003156500


app/Filament/Resources/EventAchievementResource.php
```diff
@@ -167,6 +168,17 @@ public static function form(Schema $schema): Schema
...
+                            ->rule(function (?EventAchievement $record): Closure {
+                                return function (string $attribute, mixed $value, Closure $fail) use ($record): void {
+                                    if (
+                                        $record !== null
+                                        && $value !== null
+                                        && (int) $value === (int) $record->achievement_id
+                                    ) {
+                                        $fail('Source achievement cannot be the same as the event achievement.');
+                                    }
+                                };
```
Jamiras, line 181: Would we ever want to allow any event achievement to depend on another event achievement? Should the restriction just be "source achievement cannot be event achievement"? 

Otherwise, we could still end up with an infinite loop if any achievement in the chain pointed back to the start node: A -> B -> C -> A

wescopeland: Good point.

I've reverted most of the original implementation in latest (let's just treat it as a data issue if it happens in prod somehow) and tightened this rule.

Jamiras (approved):

PR #4909, opened 2026-05-20
chore: adjust game screenshot limits

wescopeland:
This PR makes a few minor adjustments to game screenshot caps and demotion logic:
1. The in-game screenshot limit has been lowered from 20 to 10.
2. Demoted title and completion images through the relation manager now have their statuses set to `Replaced` instead of `Pending`, which better aligns with the behavior of the moderation tool.

10 games already have more than 10 screenshots. Those will continue to behave as they do today, but they won't support more screenshots being added to them.


app/Filament/Resources/GameScreenshotModerationResource.php
```diff
@@ -395,11 +395,12 @@ private static function buildApproveIngameDescription(
...
         $countLabel = $approvedCount === 1 ? 'screenshot' : 'screenshots';
         $mediaPageUrl = GameResource::getUrl('media', ['record' => $record->game]);
+        $cap = ScreenshotType::Ingame->approvedCap();
 
-        if ($approvedCount >= 20) {
+        if ($approvedCount >= $cap) {
             return new HtmlString(
                 $subjectLine
-                . self::buildApproveExplanation('This game already has 20 in-game screenshots approved.&nbsp;(20&nbsp;max)')
+                . self::buildApproveExplanation("This game already has {$cap} in-game screenshots approved.&nbsp;({$cap}&nbsp;max)")
```
Jamiras: Should the first placeholder use `$approvedCount`?

app/Platform/Enums/ScreenshotType.php
```diff
@@ -18,4 +18,12 @@ public function label(): string
             self::Completion => 'Completion',
         };
     }
+
+    public function approvedCap(): int
+    {
+        return match ($this) {
+            self::Title, self::Completion => 1,
+            self::Ingame => 10,
```
Jamiras, line 26: Unrelated: Why wasn't this `InGame`?

Jamiras (approved):

app/Platform/Enums/ScreenshotType.php
wescopeland: Mirrors the schema for consistency (but not a hard opinion, we can change it):

Image

app/Filament/Resources/GameScreenshotModerationResource.php
wescopeland: Good catch, I think it should. Updated.

PR #4912, opened 2026-05-22
perf(api): improve performance of API_GetRecentGameAwards

wescopeland:
Addresses [RAWEB-25T](https://retroachievementsorg.sentry.io/issues/109000503/?environment=production&project=4508694685417552&referrer=issue-list&statsPeriod=90d) and [RAWEB-1P4](https://retroachievementsorg.sentry.io/issues/88025910/?environment=production&project=4508694685417552&referrer=issue-list).

Performance is improved by addressing two issues:
1. `->whereDate()` wraps `awarded_at` in `DATE(...)`, so MariaDB can't use any index on `awarded_at`.
2. `user_awards` doesn't have any index to support the `award_type` filter + `awarded_at` sort this endpoint is doing.

On my local, the changes in this branch drop the endpoint response time from ~650ms to less than 1ms.


public/API/API_GetRecentGameAwards.php
```diff
@@ -68,7 +69,9 @@
 
 // If the consumer is trying to filter by a start date, add that filtering to the query.
 if ($targetDate !== null) {
-    $baseQuery->whereDate('awarded_at', '<=', $targetDate);
+    // whereDate() forces a full scan + filesort on user_awards, so
+    // we'll use where() with an explicit end-of-day bound instead.
+    $baseQuery->where('awarded_at', '<=', Carbon::parse($targetDate)->endOfDay());
```
Jamiras, line 74: A similar change can be made here:
https://github.com/RetroAchievements/RAWeb/blob/f4c096f62be50cb9d4fc372ae8a551362ec74b75/app/Community/Actions/BuildDeveloperFeedDataAction.php#L130-L131

To shave a few ms off the developer feed page for users with less than 20000 unlocks.

The only other reference I found to `whereDate` is in the connect warning stuff, so I'll address that in the next PR for that feature.

Jamiras (approved):

public/API/API_GetRecentGameAwards.php
wescopeland: Thanks, applied this change in latest (this file had two areas that were affected).

PR #4913, opened 2026-05-22
feat(ticket): make it safe to render leaderboard tickets

wescopeland:
This PR makes it safe to render leaderboard tickets in the site front-end. Explicit leaderboard ticket creation is still unsupported.

**How to test this PR**
```sql
-- manually create a leaderboard ticket
INSERT INTO
	tickets (ticketable_type, ticketable_id, ticketable_author_id, type, state, body, hardcore, reporter_id, created_at, updated_at)
VALUES
	('leaderboard', 104370, 38243, 'did_not_submit', 'open', 'aaaa bbbb cccc dddd eeee ffff gggg hhhh', 1, 500838, NOW(), NOW());
```

There are three surfaces to try out:
* http://localhost:64000/game/1/tickets?filter%5Bstatus%5D=unresolved
* http://localhost:64000/ticket/111009 (replace this with your new ticket ID)
* On any forum post, `[ticket=ID]`.

**There is still quite a bit left to do in subsequent PRs**, so I wouldn't necessarily treat any visible UI/UX at the moment as gospel.

Screenshot 2026-05-22 at 12 49 55 PM

Screenshot 2026-05-22 at 12 50 06 PM

Screenshot 2026-05-22 at 12 50 21 PM


resources/views/pages/ticket/[Ticket].blade.php
```diff
@@ -124,27 +136,50 @@
...
+            @else
+                <p role="heading" aria-level="2" class="mb-0.5 text-2xs font-bold">Leaderboard Information</p>
+                <div class="relative w-full p-2 bg-embed rounded">
+                    <x-ticket.stat-element label="Leaderboard"><a href="{{ $ticketable->getTicketableUrl() }}">{{ $ticketableTitle }}</a></x-ticket.stat-element>
+                    <x-ticket.stat-element label="Game">{!! gameAvatar($ticketableGame, iconSize: 16) !!}</x-ticket.stat-element>
+                    <x-ticket.stat-element label="Author">{!! userAvatar($ticketableAssignee ?? 'Deleted User', iconSize: 16) !!}</x-ticket.stat-element>
+                    @if ($ticketable->format)
+                        <x-ticket.stat-element label="Format">{{ ValueFormat::toString($ticketable->format) }}</x-ticket.stat-element>
                     @endif
+                    <x-ticket.stat-element label="Rank order">{{ $ticketable->rank_asc ? 'Low score wins' : 'High score wins' }}</x-ticket.stat-element>
```
Comment (Jamiras) [inline resources/views/pages/ticket/[Ticket].blade.php]: This should say "Lowest time wins" or "Highest time wins" for time formats. We probably need a new helper in `ValueFormat` to classify them.

Comment (Jamiras) [reply resources/views/pages/ticket/[Ticket].blade.php]: And I'm not sure "wins" is the best verbiage to use here. Maybe something along the lines of "Low scores are better", which is more in-line with the setting:

Image

You could probably just change the output to "lower is better" / "higher is better" and avoid the whole thing about score vs. time.

app/Platform/Services/TicketListService.php
Jamiras, line 1: Leaderboards only show up under the "all" filter. They should show up under "Core".

Also "Core"/"Unofficial" should be changed to "Published"/"Unpublished", and the header should no longer be "Achievement Type".

Image

resources/views/pages/ticket/[Ticket].blade.php
```diff
@@ -124,27 +136,50 @@
...
+                    @if ($reporterLeaderboardEntry)
+                        @php
+                            $reporterRank = $ticketable->getRank($reporterLeaderboardEntry->score);
+                            $reporterScoreLabel = $ticketable->format
+                                ? ValueFormat::format($reporterLeaderboardEntry->score, $ticketable->format)
+                                : (string) $reporterLeaderboardEntry->score;
+                        @endphp
+                        <x-ticket.stat-element label="Reporter's entry">#{{ $reporterRank }} — {{ $reporterScoreLabel }} ({{ getNiceDate($reporterLeaderboardEntry->updated_at->unix()) }})</x-ticket.stat-element>
```
Comment (Jamiras) [inline resources/views/pages/ticket/[Ticket].blade.php]: Would it be beneficial to list the rank relative to the number of entries?
> <span>#</span>5 / 11 - ...

resources/views/components/ticket/ticket-list.blade.php
```diff
@@ -59,11 +60,17 @@
                                 <a href="{{ route('ticket.show', ['ticket' => $ticket]) }}">{{ $ticket->id }}</a>
                             </td>
                             <td>{{ $ticket->state->label() }}</td>
-                            <td>{!! achievementAvatar($ticket->achievement) !!}</td>
+                            <td>
+                                @if ($ticket->ticketable_type === TicketableType::Achievement->value)
+                                    {!! achievementAvatar($ticket->ticketable) !!}
+                                @else
+                                    <a href="{{ $ticket->ticketable->getTicketableUrl() }}">{{ $ticket->ticketable->getTicketableTitle() }}</a>
```
Jamiras: I know leaderboards dont' have an avatar, but it would be nice to have some sort of indicator. I imagine things will only get more complicated once we add tickets for rich presence.

Image

resources/views/pages/ticket/[Ticket].blade.php
Comment (wescopeland) [reply resources/views/pages/ticket/[Ticket].blade.php]: Applied the simpler suggestion of "Lower is better" / "Higher is better".

app/Platform/Services/TicketListService.php
wescopeland: Updated in latest:
Screenshot 2026-06-03 at 5 34 41 PM

resources/views/pages/ticket/[Ticket].blade.php
Comment (wescopeland) [reply resources/views/pages/ticket/[Ticket].blade.php]: Probably wouldn't hurt - added in latest:
Screenshot 2026-06-03 at 5 40 18 PM

resources/views/components/ticket/ticket-list.blade.php
wescopeland: This isn't glamorous, but here's where I've landed in latest:
Screenshot 2026-06-03 at 5 44 56 PM

This is temporary:
1. Blade doesn't have access to that icon, and
2. I'd like to migrate this page to React soon and give it a UX pass in the process. Trying to get some of these nuts-and-bolts things in place before I begin that work.

Jamiras (approved):

PR #4924, opened 2026-05-26
fix: prevent achievement page tabs from deleting draft comments

Loloseiz:
Resolves #4737 

The problem was that useFormDraft's debounce is cancelled when the form unmounts. Also, Inertia navigation doesn't reload drafts because React Hook Form caches defaultValues on initial mount.

Forced a synchronous save to sessionStorage during unmount cleanup, and explicitly reset the form whenever the draftKey changes. As mentioned in the Issue's discussion, it also solves the problems in the "View all comments" page. (https://retroachievements.org/achievement/243928/comments)

### DIFF
--- a/resources/js/common/components/CommentList/useSubmitCommentForm.ts
+++ b/resources/js/common/components/CommentList/useSubmitCommentForm.ts
@@ -1,5 +1,4 @@
 import { zodResolver } from '@hookform/resolvers/zod';
-import { useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { route } from 'ziggy-js';
@@ -39,20 +38,16 @@ export function useSubmitCommentForm({
   type FormValues = z.infer<typeof addCommentFormSchema>;
 
   const draftKey = `comment-${commentableType}-${commentableId}`;
+  const baseDefaults = { body: '' };
 
   const draft = loadDraft<FormValues>(draftKey);
 
   const form = useForm<FormValues>({
     resolver: zodResolver(addCommentFormSchema),
-    defaultValues: { body: draft.body ?? '' },
+    defaultValues: { ...baseDefaults, ...draft },
   });
 
-  useEffect(() => {
-    const currentDraft = loadDraft<FormValues>(draftKey);
-    form.reset({ body: currentDraft.body ?? '' });
-  }, [draftKey, form]);
-
-  const { clearDraft } = useFormDraft(draftKey, form);
+  const { clearDraft } = useFormDraft(draftKey, form, baseDefaults);
 
   const mutation = useSubmitCommentMutation();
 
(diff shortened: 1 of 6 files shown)


resources/js/common/components/CommentList/useSubmitCommentForm.ts
```diff
@@ -35,11 +38,22 @@ export function useSubmitCommentForm({
...
-    defaultValues: { body: '' },
+    defaultValues: { body: draft.body ?? '' },
   });
 
+  useEffect(() => {
+    const currentDraft = loadDraft<FormValues>(draftKey);
+    form.reset({ body: currentDraft.body ?? '' });
+  }, [draftKey, form]);
+
+  const { clearDraft } = useFormDraft(draftKey, form);
```
wescopeland: I've confirmed this works, however, the implementation suggests `useFormDraft` is broken.

If RHF `defaultValues` caching requires a reset here, then any `useFormDraft` caller with a changing draft key has / will have the same bug. Let's move the key-change rehydration into `useFormDraft` so it resets from base defaults + `loadDraft(key)`, and ensure `clearDraft()` still suppresses the final unmount write.

Loloseiz: I'll refactor `useFormDraft` accordingly as you said, all while double-checking other components using this hook since I guess it could create edge cases if the current behavior isn't the expected one.

Loloseiz: I proceeded with the requested changes, and added tests for `useFormDraft.ts`, as well as tweaked other tests that needed to integrate the `sessionStorage` refresh.

Let met know if this needs further changes :+1:

resources/js/common/hooks/useFormDraft.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        return form;
+      },
+      { initialProps: { key: 'initial-key' as string | null } },
+    );
+
+    // ACT
+    rerender({ key: null });
+
+    // ASSERT
+    expect(true).toBe(true);
```
wescopeland, line 133: https://github.com/RetroAchievements/RAWeb/blob/master/docs/AI_POLICY.md

Loloseiz: Oh, I'm so sorry it's my first time contributing to open source projects, I didn't know about AI policies..

A bit late, but I used Gemini to correct parts of my code and to complete the tests to attain max coverage, I overlooked that to push my code faster since I needed to get back to work and that was stupid of me.

I'll still try to push my manual correction tomorrow even if this doesn't make it right, I'm very sorry as I love the project, I sincerely wanted to help out and I feel bad if my attempt only made you waste your time.

Sorry about that again, next time I'll be more careful

wescopeland: No worries, thanks for your understanding and honesty.

PR #4925, opened 2026-05-26
feat(connect): add scheduled command to delete old connect warnings

Jamiras:
Also includes the change from `whereDate` mentioned [here](https://github.com/RetroAchievements/RAWeb/pull/4912#discussion_r3304626242), and the addition of `toBase` for supplemental data used by the page where models weren't needed.


app/Platform/Commands/DeleteOldConnectWarnings.php
wescopeland, line 1: Any reason to use a custom command over `MassPrunable`? ie: _GameActivitySnapshot.php_

Jamiras: 1) I didn't know about that.
2) We don't have 90 days of data yet, so it's not possible to test with the hard-coded default.

I'll make the change.

wescopeland (approved):

PR #4926, opened 2026-05-26
refactor(connect): add AwardAchievementsAction

Jamiras:


app/Connect/Actions/AwardAchievementsAction.php
wescopeland, line 1: I can't help but notice `AwardAchievementAction` and `AwardAchievementsAction` have several blocks that are nearly identical code. Is it worth having a shared helper for some of the inner-workings of `::process()`?

app/Connect/Support/CanBeDelegated.php
```diff
@@ -25,6 +25,19 @@ protected function applyDelegationForUnlock(Request $request, Achievement $achie
         });
     }
 
+    protected function applyDelegationForUnlocks(Request $request, array $achievements): ?array
+    {
+        return $this->applyDelegation($request, function () use ($achievements) {
+            foreach ($achievements as $achievement) {
+                if (!$achievement->getCanDelegateUnlocks($this->user)) {
+                    return false;
+                }
```
wescopeland, line 34: Subtle behavior change here. Though correct, it does seem worth mentioning.

The old code silently skipped per-achievement entries the caller isn't allowed to delegate. The new code short-circuits the entire request.

I have no opinion on whether we should use the old vs new, but wanted to mention the behavior change just in case it was unintentional.

wescopeland (approved):
Non-blocking feedback

app/Connect/Support/CanBeDelegated.php
Jamiras: It was intentional. It's disingenuous to return "success, but nothing was unlocked" as an error state. A similar change was made for attempting to use the API with a non-standalone system.

app/Connect/Actions/AwardAchievementsAction.php
Jamiras: I don't think it can really be broken down into helpers, but I've updated `AwardAchievementAction` to call this with an array of the single achievement being unlocked.

There was some missing functionality that I added here to compensate: event achievement unlocks are now handled, and remaining achievements for completion is now returned.

wescopeland (approved):

PR #4931, opened 2026-05-28
refactor: delete the entire mysqli / legacyDbFetch layer and all dead helpers

wescopeland:
This PR migrates every legacy SQL query to either Eloquent or query builder. The `mysqli`, `legacyDb*`, `s_mysql_query`, `sanitize_sql_inputs`, `log_sql_fail`, and `applyFoundRows` calls have all been removed.

Migrations are done low-risk & naively, with an intent to preserve 1:1 functionality. This means:
* `DB::select()` and `DB::table()` are often used, and table aliases are often preserved.
* Soft delete scopes are explicitly bypassed to keep result sets identical.
* PascalCase associative array shapes are preserved.

Additionally:
* Every orphaned/dead legacy helper has been deleted.
* `performSearch()` has been thinned out for its one caller - it only searches for users now.


app/Helpers/database/player-history.php
```diff
@@ -10,37 +10,38 @@ function getUserBestDaysList(User $user, int $offset, int $limit, int $sortBy):
...
-    } elseif ($sortBy == 13) {// Total Points earned, desc
-        $orderCond = "ORDER BY TotalPointsEarned ASC ";
+
+    [$orderColumn, $orderDirection] = match ($sortBy) {
+        1 => ['Date', 'desc'],               // Date, asc
+        2 => ['NumAwarded', 'desc'],         // Num Awarded, asc
+        3 => ['TotalPointsEarned', 'desc'],  // Total Points earned, asc
+        11 => ['Date', 'asc'],               // Date, desc
+        12 => ['NumAwarded', 'asc'],         // Num Awarded, desc
+        13 => ['TotalPointsEarned', 'asc'],  // Total Points earned, desc
```
Jamiras: I realize these comments are copied from the previous implementation, but they're redundant with the actual code (except the asc/desc is inverted - which is confusing). Can we get rid of them?

Jamiras (approved):

app/Helpers/database/player-history.php
wescopeland: Gladly - removed them in latest.

PR #4933, opened 2026-05-29
refactor(connect): add UnknownRequestAction

Jamiras:
This finalizes the dorequest conversion.


public/dorequest.php
wescopeland, line 1: Not sure if it matters, `?r=richpresencepatch` now falls into `UnknownRequestAction`.

wescopeland (approved):

public/dorequest.php
Jamiras: I grepped for it being called on three separate days without any hits. I'm not sure if it was ever actually live. The DLL has been pulling the `RichPresencePatch` data out of the game response since the code was uploaded to GitHub in 2015:

https://github.com/RetroAchievements/RAIntegration/blob/bfd94aefc11b1c46a223352b2732644d2fd9c9e8/RA_AchievementSet.cpp#L459

PR #4935, opened 2026-05-30
perf(api): denormalize unlock time medians

Jamiras:
The unlock time median queries are very non-performant for some games - particularly those with long playtimes resulting in many sessions per user.

This eliminates the runtime calculations in favor of denormalized values that are recalculated (at most) once a day.

I have made the following changes to the logic:
* Hardcore and softcore calculations are made at the same time
  - API_GetGameProgression no longer has an `h` flag to specify which query to perform.
  - The softcore clause now only samples players who have more softcore unlocks than hardcore unlocks
* If the base query doesn't find any users who have mastered the set, a separate query is made to grab the five most recent masters. This ensures that even the rarest achievements should have some unlocks for the calculation.
* Unranked users are now ignored
* Instead of looking for the 100 most recent users, we look for 60, 100, 150, or 200 depending on the current median time to master value


app/Platform/Jobs/UpdateGameAchievementUnlockMediansJob.php
wescopeland, line 1: Recommend adding:
```php
public int $uniqueFor = 86400; // 1 day
```
`ShouldBeUnique` causes the lock to hold until the job completes. If the job throws mid-run, the lock becomes orphaned and no future recalc for the game will ever dispatch. With `$uniqueFor`, the lock can self-heal.

app/Platform/Actions/UpdateGameAchievementUnlockMediansAction.php
```diff
@@ -0,0 +1,291 @@
...
+            return;
+        }
+
+        $recentSoftcorePlayerIds = $this->getRecentPlayerIds($game, false);
+        $recentHardcorePlayerIds = $this->getRecentPlayerIds($game, true);
+        $recentPlayerIds = array_unique(array_merge($recentSoftcorePlayerIds, $recentHardcorePlayerIds));
+
+        [$unlock_times, $unlock_hardcore_times] = $this->calculateUnlockTimes($game, $achievements, $recentPlayerIds);
+
+        foreach ($achievements as &$achievement) {
```
wescopeland: Minor: Assuming the `&` is unneeded, recommend removing it.

`$achievement` is an object, so mutating its properties never needs a reference.

app/Platform/Actions/UpdateGameAchievementUnlockMediansAction.php
wescopeland, line 1: Minor: `execute()` performs N achievement saves, then updates the game's `last_median_unlock_calculation_at`. If for any reason there's a failure partway through, some achievements will be left updated, the game flag remains unset, and the next trigger will recompute everything.

Recommend wrapping this in a transaction:
```php
// ... start transaction ...

foreach ($achievements as $achievement) {
  // ...
}

$game->last_median_unlock_calculation_at = Carbon::now();
$game->saveQuietly();

// ... end transaction ...
```

public/API/API_GetGameProgression.php
```diff
@@ -38,23 +37,21 @@
...
-$preferHardcore = (int) request()->query('h');
 
 $game = Game::with('system')->find($gameId);
 if (!$game) {
     return response()->json([], 404);
 }
 
+if (!$game->last_median_unlock_calculation_at) {
+    (new UpdateGameAchievementUnlockMediansAction())->execute($game);
+}
```
wescopeland: I now better understand the reasoning behind a question you asked.

I propose we make this async instead, ie:
```php
dispatch(new UpdateGameAchievementUnlockMediansJob($game->id))->onQueue('achievement-metrics');
```

Multiple benefits to doing this:
1. The unique lock on the job stops the DB from being slammed by multiple people pointing at the same unpopulated game.
2. The heavy computation no longer runs inside the web request.
3. Everything gets collapsed to a single execution path (the job).

The trade-off is the first hit of a never-computed game will return null/0 until the job finishes. I believe this trade-off is acceptable.

app/Platform/Commands/UpdateGameAchievementUnlockMedians.php
wescopeland, line 1: Request: Can we support omitting `gameIds` to dispatch a job for every game onto the achievement-metrics queue? `UpdateGameMetrics` does something similar - optional gameIds, inline execute() when gameIds are passed, and chunk(100) + dispatch to the queue when gameIds aren't given.

In prod, I can fire up Horizon on the app server in addition to the worker server and we can probably do a complete backfill relatively quickly.

app/Platform/Actions/UpdateGameAchievementUnlockMediansAction.php
Jamiras: I don't think that's a problem. Partially updated data is better than completely stale data, and by not updating the timestamp, things _will_ get recalculated to correct the problem.

Rolling back the transaction isn't going to do anything in regards to recomputing everything.

app/Platform/Actions/UpdateGameAchievementUnlockMediansAction.php
Jamiras: Is there a penalty for using the `&`? It reminds me that the object is being mutated.

I've made the change anyway.

app/Platform/Commands/UpdateGameAchievementUnlockMedians.php
Jamiras: done

app/Platform/Jobs/UpdateGameAchievementUnlockMediansJob.php
Jamiras: Google suggested that `$uniqueFor` would interfere with the delayed job and we should just let it clean up after itself. I didn't realize that it wouldn't clean up if it threw. Further googling says it's okay if it's greater than the delay + execution time, which 24 hours should be. I also saw evidence suggesting 24 hours was the default.

I've applied the suggested change.

public/API/API_GetGameProgression.php
Jamiras: This was intended to backfill on-demand. If we're planning to do a full backfill, there's no reason to even have this check.

Removed.

app/Platform/Actions/UpdateGameAchievementUnlockMediansAction.php
wescopeland: > It reminds me that the object is being mutated.

I didn't think of it that way. No penalty for it being there.

database/migrations/2026_05_30_000001_add_last_median_unlock_to_games_table.php
wescopeland, line 1: Intellectual curiosity: any benefit now to keeping this column? It's written, but never read.

wescopeland (approved):

database/migrations/2026_05_30_000001_add_last_median_unlock_to_games_table.php
Jamiras: It was meant to drive delaying the job, but it looks like laravel does that for us. Removed.

PR #4941, opened 2026-06-01
feat(manage): set default unpromoted if no promoted achievements in table

Chew:
Small QoL I've seen brought up. On games tables, "Promoted" is always the default filter. A set in development won't have any promoted achievements, so devs working on a set always have to set the filter. If there are no promoted achievements, set the default to unpromoted instead.

:+1: Still works with any promoted achievements
CleanShot 2026-06-01 at 15 05 27@2x

This set has no promoted achievements. Previously would still be promoted with no items, now shows unpromoted by default.
CleanShot 2026-06-01 at 15 05 49@2x

I've seen this asked for before, but I don't remember where or by who. If there is a better way to get the count of the published/promoted achievements, that could be used instead, but this is as close as I got that worked.


app/Filament/Resources/GameResource/RelationManagers/AchievementsRelationManager.php
```diff
@@ -183,7 +183,7 @@ public function table(Table $table): Table
                     ->placeholder('All')
                     ->trueLabel('Promoted')
                     ->falseLabel('Unpromoted')
-                    ->default(true),
+                    ->default($this->getCoreAchievementSet()->achievements_published > 0),
```
wescopeland: Careful, the return value of `getCoreAchievementSet()` is nullable. This is due to a possible race in the back-end.

Chew: Is there something more stable for getting that info, or is null coalescing good enough to handle those race conditions?

wescopeland: Null coalescing should be good enough 👍

Jamiras (approved):

wescopeland (approved):

PR #4949, opened 2026-06-04
feat(nav): redirect to manage if there's no dropdown on mobile

Chew:
https://canary.discord.com/channels/310192285306454017/1498889559998595144/1498889559998595144

This PR allows mobile users who see no dropdown a way to get to the manage area.

It does (a bunch of) checks to see if there's a visible dropdown to show, and if there is not, just redirect the user.

All of this only affects mobile users. As a result, an override of the desktopHref was needed to let us effectively make it act like it does in desktop on mobile if there's no dropdown to show.

### DIFF
--- a/resources/views/components/menu/management.blade.php
+++ b/resources/views/components/menu/management.blade.php
@@ -9,6 +9,18 @@ $user = request()->user();
 $tools = $settings['tools'] ?? [];
 $visibleTools = collect($tools)->filter(fn($tool) => $user?->can($tool['abilities']));
 
+$anyDropdownItems =
+    $visibleTools->isNotEmpty() ||
+    $user?->can('develop') ||
+    $user?->can('manage', App\Models\Ticket::class) ||
+    $user?->can('manage', App\Models\AchievementSetClaim::class) ||
+    $user?->can('manage', App\Models\GameHash::class) ||
+    $user->Permissions >= Permissions::Developer ||
+    $user?->can('manage', App\Models\News::class) ||
+    $user->can('manage', User::class) ||
+    $user->Permissions === Permissions::Moderator ||
+    $user->can('tool');
+
 ?>
 
 <x-nav-dropdown
@@ -16,6 +28,7 @@ $visibleTools = collect($tools)->filter(fn($tool) => $user?->can($tool['abilitie
     dropdown-class="dropdown-menu-right"
     :title="__('Manage')"
     :desktopHref="route('filament.admin.pages.dashboard')"
+    :force-href="!$anyDropdownItems"
 >
     <x-slot name="trigger">
         <x-fas-toolbox />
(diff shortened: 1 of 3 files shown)


wescopeland:
My eye darts between the 3 _Manage_ labels that all have a different semantic meaning:

Screenshot 2026-06-04 at 5 31 38 PM

One is a button, another is a heading, and another is a link.

> Ideally, if there is nothing in the dropdown, we just send them to the manage area anyway.

I think this is the solution we want. This should be doable with some conditional rendering in the Blade template. If there's nothing to show in the dropdown, we can use a simple link.

Chew:
Maybe I was looking in the wrong places, basically every role should have `accessManagementTools`, including Jr Devs and Manual Unlock, the code here seems to allow it to show up in `$tools` and `$visibleTools`, that just pulls from the settings file. Was this just a bug with that not showing up?

wescopeland:
Yep. I would most certainly classify this issue as a bug 🐛

Chew:
I'll take a look into that and see if we can resolve the root issue instead 😃

Chew:
This is apparently not an issue with the tool selection since it works in development. Works on my machine™ strikes again. I'll rework this to redirect if there's nothing in the dropdown instead.

resources/views/components/menu/management.blade.php
```diff
@@ -9,13 +9,26 @@
...
+    $user->can('tool');
+
 ?>
 
 <x-nav-dropdown
     :class="$class ?? ''"
     dropdown-class="dropdown-menu-right"
     :title="__('Manage')"
     :desktopHref="route('filament.admin.pages.dashboard')"
+    :force-href="!$anyDropdownItems"
```
wescopeland, line 31: nit: Better to call this something like `shouldForceHref`. At the moment, it's almost indistinguishable in purpose between `desktopHref` unless the reader looks around to see what the type of `$anyDropdownItems` is.

Might be a good idea to rename `$anyDropdownItems` too to something like `$hasAnyDropdownItems`.

resources/views/components/menu/management.blade.php
```diff
@@ -9,13 +9,26 @@
 $tools = $settings['tools'] ?? [];
 $visibleTools = collect($tools)->filter(fn($tool) => $user?->can($tool['abilities']));
 
+$anyDropdownItems =
+    $visibleTools->isNotEmpty() ||
+    $user?->can('develop') ||
+    $user?->can('manage', App\Models\Ticket::class) ||
+    $user?->can('manage', App\Models\AchievementSetClaim::class) ||
+    $user?->can('manage', App\Models\GameHash::class) ||
```
wescopeland, line 17: Minor: For users whose permissions pass these checks but are still blocked by the later nesting, `$anyDropdownItems` will technically become true even though no item is rendered.

For example, a Junior Developer can manage tickets/claims per the policy classes, but they fail the outer `@can('develop')` on line 41. Therefore, assuming I have my boolean math correct, the mobile Manage trigger still renders as a non-link button with an empty dropdown instead of forcing the Filament dashboard href.

PR #4950, opened 2026-06-04
fix(ticket): allow reassigning tickets to reporter without a new comment

### LINKED ISSUE #2542: Unable to reassign a ticket to a reporter unless you comment first
On RA Web in a ticket, if a user responds and reassigns the ticket to the author, you cannot reassign it back to the user unless you comment.

I had a user reassign a ticket to me even though the action item is currently for them. Now I can't reassign it to them unless I leave a comment first.

https://retroachievements.org/ticket/74234

firefox_qlMVABLV75

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/2542.

Reassigning without a reply is now possible, with an added JS `confirm()` speedbump before doing so.


resources/views/pages/ticket/[Ticket].blade.php
Comment (Jamiras) [inline resources/views/pages/ticket/[Ticket].blade.php:1]: Something about this change allows a ticket to be reassigned to the reporter immediately after it has been opened. The first initial reassign to reporter should still require a comment from the dev.

Comment (wescopeland) [reply resources/views/pages/ticket/[Ticket].blade.php:1]: Fixed in latest.

Jamiras (approved):

PR #4952, opened 2026-06-04
fix(user): allow viewing dev stats page even if user only has contributed code notes

### LINKED ISSUE #3091: Code notes created link takes you back to the dev's page

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/3091.

http://localhost:64000/individualdevstats.php?u=Eliline is a good page to test with.

Screenshot 2026-06-04 at 8 35 58 AM


resources/views/pages-legacy/individualdevstats.blade.php
```diff
@@ -20,10 +20,7 @@
 
 $userArchInfo = getUserAchievementInformation($devUser);
 
-// Only get stats if the user has a contribution count
-if (empty($userArchInfo)) {
-    abort_with(redirect(route('user.show', $dev)));
```
Jamiras, line 25: This PR effectively opens up the page for all users:

http://localhost:64000/individualdevstats.php?u=Sarconius

Maybe leaving this redirect, but add a Role check to it:
```
if (empty($userArchInfo) && !$devUser->hasRole([Role::DEVELOPER, Role::DEVELOPER_JUNIOR]))
```
Though that wouldn't handle a case where a user added notes then lost the role before publishing any achievements.

wescopeland: I considered that too - though nothing links to the page for those particular users.

For the reasons you've stated, I'm not fond of the role check.

This feels less like a bug and more like an empty state.

Jamiras: > nothing links to the page for those particular users.

You can go to any actual developer's page, then use the search box to jump to a non-developer.

image

> This feels less like a bug and more like an empty state.

I guess that's reasonable.

Jamiras (approved):

PR #4953, opened 2026-06-04
refactor(connect): add UploadBadgeImageAction

Jamiras:
provides the dorequest conversion treatment to doupload


app/Filament/Actions/ProcessUploadedImageAction.php
```diff
@@ -44,8 +46,14 @@ public function execute(string $tempImagePath, ImageUploadType $imageUploadType)
...
-                $file = createFileArrayFromDataUrl($dataUrl);
-                $imagePath = UploadBadgeImage($file);
+                $file = new UploadedFile(
+                    $tempImagePath,
+                    str_replace('image/', 'data-url.', $mimeType),
+                    $mimeType,
+                    test: true, // prevent HTTP validation
+                );
+
+                $imagePath = new UploadBadgeImageAction()->execute($file);
```
wescopeland: Careful, the return shape here is different from what Filament expects and it throws a server error on achievement badge upload.

Jamiras: Yea, it's definitely broken. I think I may have test game badges instead of achievement ones.

Jamiras: fixed.

app/Filament/Actions/ProcessUploadedImageAction.php
```diff
@@ -44,8 +46,18 @@ public function execute(string $tempImagePath, ImageUploadType $imageUploadType)
...
-                $file = createFileArrayFromDataUrl($dataUrl);
-                $imagePath = UploadBadgeImage($file);
+                $file = new UploadedFile(
+                    $disk->path($tempImagePath),
+                    str_replace('image/', 'data-url.', $mimeType),
+                    $mimeType,
+                    test: true, // prevent HTTP validation
+                );
+
+                $result = new UploadBadgeImageAction()->execute(auth()->user(), $file);
```
wescopeland, line 56: May still be a subtle bug here due to this code:
```php
        if (!$this->user->can('create', Achievement::class)) {
            return $this->mustBeDeveloper();
        }
```
Does this mean users without a developer role (ie: event manager, artist) would no longer be able to upload badges in Filament?

Jamiras: Most likely, yes. This is just the check that was there previously: https://github.com/RetroAchievements/RAWeb/blob/4baa7de00b0841d2e2e79e1174c98daf678f29d5/public/doupload.php#L29

I can easily change this to check the `update` policy to allow Artists and Writers to upload badges, but that would restrict Junior Developers because they are only allowed to update things associated to a game they have active claims on, and the image is not directly attached to a game. I guess I could check for either permission.

That wouldn't handle the Event Manager role, and I don't see anything referencing that role in the `AchievementPolicy` class. Are you sure they can upload achievement badges? Game badges have their own policy, and I don't see Event Manager in there either.

wescopeland: RE: Event Managers. Apologies - I was not 100% yesterday and misspoke. Event Managers can't update achievement/event_achievement badges.

I think we probably want to use `AchievementPolicy::updateField()`, with `$fieldName` set to "image_name". This has multiple benefits:
* Allows devs & artists through by default.
* Excludes writers.
* Has a path towards fixing the issue for JrDevs.

Seems like we need to add 'image_name' to this array, and then we wouldn't need to bail out of the policy anymore:
```php
Role::DEVELOPER_JUNIOR => ['title', 'description', 'type', 'points', 'order_column'],
```

Jamiras: The junior developer path is still trying to look for a claim, and we don't have an achievement to cross-reference against:
https://github.com/RetroAchievements/RAWeb/blob/4baa7de00b0841d2e2e79e1174c98daf678f29d5/app/Policies/AchievementPolicy.php#L177-L179

I've added a special-case for junior developers to the check in the action.

wescopeland (approved):

app/Connect/Actions/UploadBadgeImageAction.php
```diff
@@ -0,0 +1,135 @@
...
+
+        if (!imagepng($image, $localImagePath) || !imagepng($imageLocked, $localImagePathLocked)) {
+            return $this->internalError('Failed to write image to disk.');
+        }
+
+        Storage::disk('media')->put($imagePath, file_get_contents($localImagePath));
+        Storage::disk('media')->put($imagePathLocked, file_get_contents($localImagePathLocked));
+
+        UploadToS3($localImagePath, $imagePath);
+        UploadToS3($localImagePathLocked, $imagePathLocked);
```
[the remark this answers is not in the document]
wescopeland: I wouldn't worry about this.

wescopeland (approved):

PR #4956, opened 2026-06-05
fix(forums): use selected comment for meta description

Chew:
Very small minor QoL thing. If a user links to a specific comment, it will show that comment in the meta description.

In theory we can also make it show the author in the meta as well, I think `$comment->user` should be available to us at least. 

`collect()`ing feels inefficient, I'm not sure of a faster way, this could get expensive with a bunch of comments. Everything else seems right, though. I think we could use the map, but PHP maps are weird and I'm not sure if I can "give" that comment back outside the loop.

Can't really show a before and after since it's hard to embed local links, but here's some screenshots showing it works. Maybe a test could be useful as well, if needed.

CleanShot 2026-06-05 at 14 14 16@2x
CleanShot 2026-06-05 at 14 14 02@2x


wescopeland (approved):
Thanks!

PR #4959, opened 2026-06-06
fix(game): add screenshot max size validation, bump max size to 6MB

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1512463207351062589.

Two issues at play here:
1. The back-end max size is too small for 3x upscaled images.
2. The front-end has no size validation whatsoever.

This PR bumps the max size to 6MB and adds front-end size validation.

Screenshot 2026-06-05 at 9 06 05 PM


app/Platform/Services/GameScreenshotValidationService.php
Jamiras, line 1: The Wii's internal resolution is 640x480, so a 3x upscaled image should be 1920x1440. The images provided in the link are 2500x1368 (3.9x width, 2.85x height), so it looks like Dolphin is approximating widescreen (which the Wii did) and the user is probably playing with a maximized window that doesn't quite fill a 2560x1440 monitor because of the window border or OS UI elements.

Wii allows uploading images of any resolution. How does that play into making sure all the screenshots for a single game have the same resolution when multiple users contribute? 1368 is not a scalar multiple of 480, so how will others match it?

Also, I realize that 6MB is hardly anything these days, but if there's 10x screenshots, the gallery would be 60MB. The gallery appears to use much smaller images (which makes sense - my 1080p monitor can't even show the full image without resizing it). Do we even keep the original image around? If so, how can a user access them? If not, why allow such large images in the first place?

Jamiras (approved):

app/Platform/Services/GameScreenshotValidationService.php
wescopeland: > Wii allows uploading images of any resolution. How does that play into making sure all the screenshots for a single game have the same resolution when multiple users contribute?

The review tool adjustments in #4963 and dialog adjustments in #4967 should take care of this.

> Also, I realize that 6MB is hardly anything these days, but if there's 10x screenshots, the gallery would be 60MB. The gallery appears to use much smaller images

Yep - for these hi-res shots, the gallery should never be rendering the raw asset. It should always be rendering a transformed asset instead (WebP).

> Do we even keep the original image around?

We do - it's stored in Cloudflare R2 (formerly S3). We need to always keep the original around in case we want to change what transforms we run on these images.

> how can a user access them?

At the moment, we don't expose any way to access the raw untransformed hi-res image. This wouldn't be hard to do, I'm just not sure if it would be useful.

PR #4963, opened 2026-06-07
feat(manage/media): rewrite the screenshot review tool based on the initial test outcome

wescopeland:
This PR completely revamps the screenshot review tool - both the list/table UX and the dialog UX.

It took a lot of iteration to reach this point - initial designs can be found here: https://discord.com/channels/310192285306454017/888607146290343976/1512886134965801153

The goal of this redesign is to bake in functionality to assist reviewers as much as possible. We can't fully automate the review process, but we can bake in quite a bit of intelligence to help augment the discerning reviewer.

### UX Comparisons

**List Before**
Screenshot 2026-06-07 at 5 14 26 PM

**List After**
Screenshot 2026-06-07 at 5 13 41 PM

**Dialog Before**
Screenshot 2026-06-07 at 5 14 35 PM

**Dialog After**
Screenshot 2026-06-07 at 5 14 07 PM

### High-level UX Overview

**List**
- The 'Status' column is hidden by default.
- The approve and reject buttons have been combined into a single 'Review' button.
- Context cues are shown in-line on each row where we can make educated guesses about what has been uploaded.
- The list is now filterable by system.
- Clicking the thumbnail now pops the Review dialog.

**Dialog**
This is the most important piece of the rework.

We now have a side-by-side comparison of the Current Primary vs the Submission image on a checkerboard background to expose the image framing:

Screenshot 2026-06-07 at 5 22 18 PM

Clicking an image in the review dialog pops a simulation of what the image looks like on the game page in the media dialog. This can also help reviewers spot artifacts from manually resized pixel art images:
Screenshot 2026-06-07 at 5 23 01 PM

The UI takes an educated guess as to what the right action is. We frame this as "Suggested path" instead of "Recommendation" to hopefully not induce too much bias in the reviewer - the intent is just to help them have to think less:
Screenshot 2026-06-07 at 5 23 19 PM

Pagination buttons are included in the heading to help the reviewer tap through review items. On performing one of the three actions, the dialog automatically advances to the next item (and it respects the current filter options):
Screenshot 2026-06-07 at 5 24 02 PM

More real-world examples:
Screenshot 2026-06-07 at 5 25 03 PM

Screenshot 2026-06-07 at 5 25 13 PM

Screenshot 2026-06-07 at 5 25 26 PM

Screenshot 2026-06-07 at 5 25 36 PM

Screenshot 2026-06-07 at 5 25 49 PM

Screenshot 2026-06-07 at 5 26 03 PM

Screenshot 2026-06-07 at 5 26 23 PM

Screenshot 2026-06-07 at 5 26 43 PM


app/Filament/Resources/GameScreenshotModerationResource.php
```diff
@@ -355,358 +499,359 @@ private static function applyReviewFeedScope(Builder $query): Builder
...
-            ScreenshotType::Title->label() => 0,
-            ScreenshotType::Ingame->label() => 1,
-            ScreenshotType::Completion->label() => 2,
+        $recordKey = (string) $record->getKey();
+        $context = ScreenshotReviewContext::make($record);
+        $recommendation = $context->recommendedAction();
+        $cards = [];
+
+        $cards[] = [
+            'title' => 'Promote to primary',
```
Jamiras: Would it be beneficial to specify the type here? This sits below a "primary" list, and it's not immediately clear which image is being replaced.

Image

Jamiras: Also, there's no way to easily promote to primary without retiring the current primary. Maybe I want to keep the old primary as a secondary image.

app/Filament/Resources/GameScreenshotModerationResource.php
```diff
@@ -165,156 +210,78 @@ public static function table(Table $table): Table
...
-                            HTML);
-                    })
-                    ->action(function (GameScreenshot $record, array $data) use ($user) {
-                        (new RejectGameScreenshotAction())->execute(
-                            screenshot: $record,
-                            reviewer: $user,
-                            reason: GameScreenshotRejectionReason::from($data['rejection_reason']),
-                            notes: $data['rejection_notes'] ?? null,
-                        );
+                Action::make('review')
```
Jamiras, line 260: Instead of hiding the Status column, could you put this in that column instead of saying "Pending"? That would simplify the flow when removing the default Pending filter to more easily differentiate between approved and rejected.

It's probably fine as-is. I didn't realize the Status column was just hidden at first, and expect others to have a similar experience.

app/Filament/Resources/GameScreenshotModerationResource.php
wescopeland: > Would it be beneficial to specify the type here?

Yes, I think so:
Screenshot 2026-06-09 at 6 28 38 PM


> Also, there's no way to easily promote to primary without retiring the current primary. Maybe I want to keep the old primary as a secondary image.

Added this functionality to the confirm dialog. Note this should only impact in-game images.
Screenshot 2026-06-09 at 6 28 45 PM

app/Filament/Resources/GameScreenshotModerationResource.php
wescopeland: Agree it's probably fine as-is, but made a small change in latest so that if you filter by status, we auto-toggle the column visibility.

Jamiras (approved):

app/Filament/Resources/GameScreenshotModerationResource/Pages/Index.php
```diff
@@ -5,12 +5,236 @@
...
+
+        $this->performMountedReviewOrAdvance(
+            $recordKey,
+            $user,
+            fn (GameScreenshot $record): bool => GameScreenshotModerationResource::approve(
+                $record,
+                $user,
+                ScreenshotReviewDecision::from($kind),
+            ),
+        );
```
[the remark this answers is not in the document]
wescopeland: This is technically valid, but I don't think this is a P1. Our Filament tooling is internal & back office. The only way we hit this is for an authenticated staff member to hand-craft a Livewire payload, which is then going to just throw an error.

Regardless, I've fixed this anyway.

app/Filament/Resources/GameScreenshotModerationResource/Pages/Index.php
```diff
@@ -5,12 +5,236 @@
...
+            $recordKey,
+            $user,
+            function (GameScreenshot $record) use ($user, $reason, $notes): bool {
+                try {
+                    (new RejectGameScreenshotAction())->execute(
+                        screenshot: $record,
+                        reviewer: $user,
+                        reason: GameScreenshotRejectionReason::from($reason),
+                        notes: $notes,
+                    );
```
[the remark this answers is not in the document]
wescopeland: Same comment as I left elsewhere:

> This is technically valid, but I don't think this is a P1. Our Filament tooling is internal & back office. The only way we hit this is for an authenticated staff member to hand-craft a Livewire payload, which is then going to just throw an error.
> 
> Regardless, I've fixed this anyway.

PR #4967, opened 2026-06-08
feat(GameScreenshotUploadDialog): adjust copy based on user behavior

wescopeland:
https://discord.com/channels/310192285306454017/1512802214152900648/1513663596163371168

We'll no longer explicitly state what the supported resolutions are in the upload dialog. We just guide the user towards using the emulator's built-in screenshot tool, which should almost always be correct.

A subsequent PR will add a link to the upload guidelines in our documentation.

This also adds nudges for mismatched/orphaned resolutions.

Screenshot 2026-06-08 at 5 52 18 PM

Screenshot 2026-06-08 at 5 52 37 PM

Screenshot 2026-06-08 at 5 57 42 PM


resources/js/features/games/components/UploadForm/ScreenshotPreviewMeta.tsx
```diff
@@ -97,39 +67,30 @@ export const ScreenshotPreviewMeta: FC<ScreenshotPreviewMetaProps> = ({
...
-interface AcceptedSizesTooltipProps {
-  screenshotResolutions: Array<{ width: number; height: number }>;
-  supportsUpscaledScreenshots?: boolean;
+function buildConsistencyNudgeMessage(
+  selectedType: App.Platform.Enums.ScreenshotType | undefined,
+  t: TFunction,
+): string {
+  if (selectedType === 'title') {
+    return t(
+      'More likely to be accepted if you also submit a matching in-game screenshot at this resolution.',
```
Jamiras: My immediate reaction to this text is to click on the Title button and drag in another image. Because the first image hasn't actually been submitted, this changes the submission into a Title image submission. This behavior is consistent with the current dialog, but the text is directing me to do something that might introduce user error.

I'm not sure how best to handle this. Maybe change the text to "More likely to be accepted if you also submit a matching in-game screenshot at this resolution after submitting this one." Or maybe it's just something users will figure out themselves when they switch types without submitting.

resources/js/features/games/components/UploadForm/ScreenshotPreviewMeta.tsx
```diff
@@ -97,39 +67,30 @@ export const ScreenshotPreviewMeta: FC<ScreenshotPreviewMetaProps> = ({
...
+    );
+  }
+
+  if (selectedType === 'ingame') {
+    return t(
+      'More likely to be accepted if you also submit a matching title screenshot at this resolution.',
```
Jamiras: This text is still shown when trying to submit the title image immediately after submitting the in-game image. The suggestion doesn't go away until an approved screenshot exists at the matching resolution.

resources/js/features/games/components/UploadForm/ScreenshotPreviewMeta.tsx
```diff
@@ -1,52 +1,40 @@
...
-
   const invalidExplanation = supportsUpscaledScreenshots
-    ? t("Use your emulator's screenshot tool, ideally at 2x or 3x internal resolution.")
-    : t("Use your emulator's screenshot tool, not a desktop capture.");
+    ? t(
+        "This doesn't look like a native capture. Use your emulator's screenshot tool at native, 2x, or 3x internal resolution. Don't manually resize.",
+      )
+    : t(
+        "This doesn't look like a native capture. Use your emulator's screenshot tool at native resolution. Don't manually resize.",
+      );
```
Jamiras, line 34: I wonder if it would make sense to specify "Don't use your operating system's screenshot functionality". This is how I used to capture game images as I preferred the aspect-ratio-corrected sizes.

Oh, I see ", not a desktop capture" was removed from the wording. Should it be added back?

Jamiras: Also, I've seen mention that RetroArch screenshots are aspect corrected. I don't think that's true, but I know that on-screen widgets (i.e. the FF indicator) sometimes appear in screenshots. There's also an option that applies shaders before capturing screenshots, but that should be off by default. These things should be investigate and added to the documentation. I don't think they belong here, but there might be confusion about assuming that the "emuatlor's screenshot tool" is going to just work.

MokaStitcher: <sub>hope it's fine to interject</sub>
The RetroArch option that messes up screenshots is 'GPU Screenshot', found in the Video settings. I can confirm that having this setting turned on modifies the aspect ratio of the screenshots in some cores like snes9x and as you said also captures things external to the game like filters, notifications, FPS indicator, etc.
The setting was unfortunately enabled by default before February of this year: https://github.com/libretro/RetroArch/commit/eace37cc7666e48e990f9af43d243346b5cebe65
I don't believe updating an existing RetroArch install would turn it off, so I suspect many many RA users have it enabled like I did. Even a device I setup for the first time on March 12th still had the setting turned on.
Because of this I definitely second the need to document this somewhere and make people aware of it when beta resumes. That being said, this is purely based on my experience and the research I did when I realized most of the screenshots I had taken were invalid. By all means feel free to double check everything I've said here beforehand
<sub>thank you both for you hard work on RA, I hope this was helpful</sub>

resources/js/features/games/components/UploadForm/ScreenshotPreviewMeta.tsx
wescopeland: A copy change is likely the best fit here. Updated in latest:
Screenshot 2026-06-09 at 7 13 05 PM

resources/js/features/games/components/UploadForm/ScreenshotPreviewMeta.tsx
wescopeland: Thanks - fixed in latest.

resources/js/features/games/components/UploadForm/ScreenshotPreviewMeta.tsx
wescopeland: > Oh, I see ", not a desktop capture" was removed from the wording. Should it be added back?

Yes, I think so. Done in latest:
Screenshot 2026-06-09 at 7 12 55 PM


> shader & GPU Screenshot setting discussion

Agree this is a problem - also agree it doesn't belong here. This level of nuance probably belongs in our documentation, largely because we can't detect this failure mode without human review. cc @tomojira for visibility.

Jamiras (approved):

PR #4969, opened 2026-06-09
feature(connect): add pending loss-of-hardcore-support warning

Jamiras:
Currently, when we update the minimum hardcore version supported for an emulator, users on the previous minimum hardcore version are immediately thrown into softcore only mode without warning.

This adds the ability to schedule the minimum hardcore version change and passes that information to the client via a warning achievement.

image

image

image

image


app/Connect/Actions/GetAchievementSetsAction.php
```diff
@@ -488,6 +490,19 @@ private function injectClientSupportWarning(array &$response): void
...
+                $emulatorUserAgent = $userAgentService->getEmulatorUserAgent($this->userAgent);
+                if (!$emulatorUserAgent || !$emulatorUserAgent->pending_minimum_hardcore_version_at) {
+                    return;
+                }
+                $days = (int) ceil($emulatorUserAgent->pending_minimum_hardcore_version_at->diffInDays(Carbon::now(), true));
+                if ($days > 14) {
+                    return;
+                }
+
+                $title = 'Warning: Outdated Emulator (please update)';
```
Jamiras: Since RetroArch doesn't show the description in the popup, I'm wondering if there's better wording to use here to indicate that the emulator is about to become outdated instead of using the exact same messaging as when it has become outdated.

app/Platform/Services/UserAgentService.php
```diff
@@ -337,6 +337,14 @@ public function getSupportLevelAndCoreRestriction(string|array|null $userAgent):
             return [ClientSupportLevel::SoftcoreOnly, $coreRestriction];
         }
 
+        if ($supportLevel === ClientSupportLevel::Full
+            && $emulatorUserAgent->pending_minimum_hardcore_version
+            && $emulatorUserAgent->pending_minimum_hardcore_version_at) {
+            if (UserAgentService::versionCompare($data['clientVersion'], $emulatorUserAgent->pending_minimum_hardcore_version) < 0) {
+                $supportLevel = ClientSupportLevel::SoftcorePending;
```
wescopeland: Would it make sense to enforce pending cutovers here once their date has passed rather than exclusively relying on the scheduled command?

app/Platform/Commands/UpdateMinimumEmulatorVersions.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public function handle(): void
+    {
+        $count = EmulatorUserAgent::query()
+            ->where('pending_minimum_hardcore_version_at', '<=', Carbon::now())
+            ->update([
+                'minimum_hardcore_version' => DB::raw('pending_minimum_hardcore_version'),
+                'pending_minimum_hardcore_version' => null,
+                'pending_minimum_hardcore_version_at' => null,
+            ]);
```
wescopeland, line 31: In Filament, it's possible to accidentally set the date without setting the version. When the command runs, `minimum_hardcore_version` then gets set to null.

I think we probably want a `->whereNotNull('pending_minimum_hardcore_version')` in this query.

app/Connect/Actions/GetAchievementSetsAction.php
wescopeland: How about a countdown? Something like:

> Warning: Update Your Emulator Within 4 Days

Jamiras: The countdown is in the description. This is just meant to get their attention.

Maybe
> Warning: Soon-to-be Outdated Emulator (please update)

app/Platform/Commands/UpdateMinimumEmulatorVersions.php
Jamiras: Done.

I tried to make a rule to require both fields if either was set, but couldn't get it to work quite right. Once a field was marked as required, the submission would report an error even if the fields were cleared out.

app/Platform/Services/UserAgentService.php
Jamiras: Technically, it would probably make sense to _only_ do it here. It only really matters for enforcement. We don't display the information anywhere other than the management console where it's set.

Then we wouldn't need a scheduled command that would really only do something once every few months.

The trade-off would be that a pending version change could just idle until a user violated it, which might look awkward in the management console. However, given that we've already had almost 200 sessions this month on RetroArch versions prior to 1.8.2 (where they started providing a version), I have very little concern of a pending version not being migrated in a timely manner.
```
MariaDB [retroachievements]> select count(*) from player_sessions where id > 96865144 and user_agent like 'libretro%';
+----------+
| count(*) |
+----------+
|      197 |
+----------+
```

app/Connect/Actions/GetAchievementSetsAction.php
wescopeland: IMO, what you currently have hits harder than that idea - I like the current one more.

I still lean towards having the countdown in the title, that would hit hardest. Only vibes: I assume most of the time the description will not be read.

Just my opinion though.

wescopeland (approved):

app/Connect/Actions/GetAchievementSetsAction.php
Jamiras: I've modified the text to:

> Warning: Outdated Emulator in 4 days (please update)

app/Platform/Services/UserAgentService.php
```diff
@@ -337,6 +337,14 @@ public function getSupportLevelAndCoreRestriction(string|array|null $userAgent):
...
+            }
+        }
```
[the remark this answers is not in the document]
Jamiras: Related: https://github.com/RetroAchievements/RAWeb/pull/4969#discussion_r3384450280

I've gone ahead and updated the service to do the conversion if the date is in the past. As the command only runs once a day, it's entirely likely that the service will update the version before the command runs. It's possible the command may not even be needed any more, except for the case where an outdated version is not actively being used.

wescopeland (approved):

PR #4978, opened 2026-06-11
feat(manage/games): cap to the max number of similar games to 6 per game ID

wescopeland:
https://discord.com/channels/310192285306454017/888607146290343976/1514281892562604092

This PR caps the max number of similar games at 6 per each game ID. The cap is bidirectional.

Screenshot 2026-06-11 at 6 11 02 PM

Screenshot 2026-06-11 at 6 17 10 PM


app/Platform/Actions/LinkSimilarGamesAction.php
```diff
@@ -4,46 +4,73 @@
...
 
-            if (!$isAlreadyAttached) {
-                $similarGamesSet->games()->attach($parentGame->id);
+            $candidateCount = $candidateSet->exists ? $candidateSet->games()->count() : 0;
+            if ($candidateCount + 1 > self::MAX_SIMILAR_GAMES) {
+                throw new SimilarGamesCapExceededException($gameId, self::MAX_SIMILAR_GAMES);
             }
+
+            $targetSets[] = $candidateSet;
+        }
```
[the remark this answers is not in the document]
wescopeland: Technically true, but this will virtually never happen in practice. Our traffic volume that's messing with Similar Games at any one time is pretty low, especially on the same game ID.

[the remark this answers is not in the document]
wescopeland: No. This is case-by-case in Filament. You shouldn't remember this as a rule.

app/Exceptions/SimilarGamesCapExceededException.php
```diff
@@ -0,0 +1,17 @@
...
+
+use RuntimeException;
+
+class SimilarGamesCapExceededException extends RuntimeException
+{
+    public function __construct(
+        public readonly int $offendingGameId,
+        public readonly int $cap,
+    ) {
+        parent::__construct("Game #{$offendingGameId} would exceed the {$cap}-similar-game cap.");
```
Jamiras, line 15: > App\Exceptions\SimilarGamesCapExceededException::__construct(): Argument #1 ($offendingGameId) must be of type int, string given, called in /var/www/html/app/Platform/Actions/LinkSimilarGamesAction.php on line 49

When trying to link a game by ID where the count would exceed.

Image

app/Filament/Resources/GameResource/Pages/SimilarGames.php
```diff
@@ -181,18 +187,31 @@ public function table(Table $table): Table
...
+                        } catch (SimilarGamesCapExceededException $e) {
+                            $offending = $e->offendingGameId === $game->id
+                                ? $game
+                                : Game::find($e->offendingGameId);
+                            $title = $offending?->title ?? "Game #{$e->offendingGameId}";
+
+                            Notification::make()
+                                ->danger()
+                                ->title('Cap exceeded')
+                                ->body("Cannot add: {$title} would exceed the {$e->cap}-similar-game cap. Remove an entry first.")
```
Jamiras: If you try to add more games than would be supported (i.e. add 4 games to a list already containing 3), the error message that comes back is misleading:

Image

It seems to indicate that the "self" games cannot be added to itself.

app/Exceptions/SimilarGamesCapExceededException.php
wescopeland: Thanks, fixed in latest.

app/Filament/Resources/GameResource/Pages/SimilarGames.php
wescopeland: Updated the copy when hitting this edge in latest:
Screenshot 2026-06-15 at 4 55 03 PM

Jamiras (approved):

PR #4979, opened 2026-06-12
feat(api): add V2 JSON:API endpoints for the AchievementSetVersion resource

Chew:
This PR adds V2 JSON:API endpoints for `AchievementSetVersion` resources. Since this feature is brand new, it doesn't replace any V1 functionality.

The resource is accessible as a top-level index, as well as through achievementSet relationship routes:
* `GET /api/v2/achievement-set-versions`
* `GET /api/v2/achievement-sets/{set}/achievement-set-versions`

Things worth noting:
1. There is no `show` route. Similar to `AchievementSetClaim`, claims are list-only. No `self` link is emitted on individual records.
2. Default sort is `-createdAt`.
3. Default attributes embed minimal achievement set context, so clients can have contextual info without a forced `?include=`.
4. The full `achievementSet` relationship are still includable via `?include=achievementSet`. Relationship data is only emitted when explicitly included.

**How to test this PR:**
```bash
# Get the pre-sorted list of versions
curl --request GET \
  --url http://localhost:64000/api/v2/achievement-set-versions \

# Get all versions on a single set
curl --request GET \
  --url http://localhost:64000/api/v2/achievement-sets/1/achievement-set-versions \

# Sort by date and apply custom pagination
curl --request GET \
  --url 'http://localhost:64000/api/v2/achievement-set-versions?sort=-createdAt&page[number]=2&page[size]=25' \
```

First time doing this, scrutinize as needed please :)


wescopeland:
Looks like checks are failing - happy to review once CI is green 👍

Chew:
I'm not sure how to fix the page issue, and need advice on the policy thing. Do I just remove the empty ones or add returns like it says?

wescopeland:
For the policy thing, I recommend just adding stubs that all return false.

What page issue are you facing? I'm not seeing it in CI for some reason.

Chew:
For some reason they get cut off in the CI but locally the errors are something like:

1. It rejects invalid page number
2. It rejects invalid page lengths

Something along those lines. I'm not sure what I've done differently where these tests pass in other controllers, but fail here. These tests aren't a part of the tests I wrote

Chew:
Alright, everything has been sorted and figured out. Will address Greptile's reviews just to be sure as well.

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Get the resource's relationships.
+     *
+     * @param Request|null $request
+     * @return iterable
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'achievementSets' => $this->relation('achievementSet')
```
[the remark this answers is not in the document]
Chew: The tests obviously don't fail, so this isn't an issue

app/Api/V2/AchievementSetVersions/AchievementSetVersionSchema.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        return [
+            ID::make(),
+
+            Number::make('version')->readOnly(),
+
+            DateTime::make('createdAt')->sortable()->readOnly(),
+            DateTime::make('updatedAt')->sortable()->readOnly(),
+
+            ArrayHash::make('definition', 'definition')->readOnly(),
```
Jamiras: I realize this is more of a comment for #4942, but it's awkward seeing `version:1` in the definition, and a different `version:X` in the container. This is because the version in the definition is a schema version and the version in the container is a revision counter.
```
         "attributes":{
            "version":2,
            "createdAt":"2026-04-26T18:44:05.000000Z",
            "updatedAt":"2026-06-02T15:35:10.000000Z",
            "definition":{
               "version":1,
               "achievements":[
```

Which brings up a discussion for this functionality that I don't expect you to have the answers to.

How will future schema revisions interfere with downstream consumers if they're not strictly additive? Should the embedded definition just be dumped as-is, or should we process it so we can control the output?

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @param Request|null $request
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'achievementSets' => $this->relation('achievementSet')
+                ->withoutLinks()
+                ->showDataIfLoaded(),
+        ];
```
Jamiras, line 51: I'm not sure how we do this for other endpoints, but when requesting data through the relationship, it's redundant to include the relationship in the response:

`localhost:64000/api/v2/achievement-sets/2539/achievement-set-versions`:
```
         "type":"achievement-set-versions",
         "id":"11414",
         "attributes":{}
         "relationships":{
            "achievementSet":{
               "data":{
                  "type":"achievement-sets",
                  "id":"2539"
               }
            }
         },
```

Chew: Good call, I removed `showDataIfLoaded()` and it removed that from the response. it's covered by the `achievementSetId` anyway

app/Api/V2/AchievementSetVersions/AchievementSetVersionSchema.php
wescopeland: Originally I was thinking we should dump the whole definition and let consumers figure it out, but after thinking more about it, that probably isn't viable. Any schema change will be a breaking change for all API consumers.

We probably have no choice but to drop `definition` from the response and send processed values, ie:
```json
{
  "type": "achievement-set-versions",
  "id": "42",
  "attributes": {
    "version": 2,
    "createdAt": "2026-04-26T18:44:05Z",
    "updatedAt": "2026-06-02T15:35:10Z",
    "playersTotal": 1234,
    "playersHardcore": 567,
    "achievementsPublished": 50,
    "achievementsUnpublished": 0,
    "pointsTotal": 500,
    "achievementSnapshot": [
      { "id": 123, "points": 10, "isPromoted": true, "type": "progression" },
      { "id": 456, "points": 5,  "isPromoted": false, "type": null }
    ]
  }
}
```

Note that in this example, the inner `version` (schema version) is dropped, as it's an implementation detail.

Chew: Implemented this, probably not very well, but it's there. Made a presenter class to parse it.

Ideally, if we do need to update the schema, it's just adding to this response, not removing. Then there's no breaking changes.

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Get the resource's relationships.
+     *
+     * @param Request|null $request
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'achievementSets' => $this->relation('achievementSet')
```
wescopeland: Should this be singular? My interpretation is this implies a version can be associated to multiple sets.

app/Api/V2/AchievementSetVersions/AchievementSetVersionPresenter.php
wescopeland, line 1: Using `switch` here is clever and I like it. Recommend for `default` returning `null` -- at the moment a consumer may have trouble distinguishing between "no snapshot exists" vs "there's nothing present for this version".

app/Api/V2/AchievementSetVersions/AchievementSetVersionSchema.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return [
+            ID::make(),
+
+            Number::make('version')->readOnly(),
+
+            DateTime::make('createdAt')->sortable()->readOnly(),
+            DateTime::make('updatedAt')->sortable()->readOnly(),
+
+            ArrayHash::make('achievementSnapshot', 'achievementSnapshot')->readOnly(),
```
wescopeland: Possibly safe to remove

app/Api/V2/AchievementSetVersions/AchievementSetVersionSchema.php
```diff
@@ -0,0 +1,98 @@
...
+     * Default sort order when client doesn't provide any.
+     */
+    protected $defaultSort = '-createdAt';
+
+    /**
+     * Relationships that should always be eager loaded.
+     */
+    protected array $with = [
+        'achievementSet.gameAchievementSets',
+    ];
```
wescopeland, line 44: Possibly safe to remove

Chew: Removing this seems to give me an error when attempting to run `/api/v2/achievement-set-versions?include=achievementSet`

`Attempted to lazy load [gameAchievementSets] on model [App\\Models\\AchievementSet] but lazy loading is disabled`

I'm sure there's a better way to resolve it, though

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
Chew: Set it to singular, I did plural mostly because I saw it elsewhere, and maybe it makes more sense to `?include=achievementSets`, but you are still getting an individual achievement set, just multiple of those individuals. Could go either way, but I think singular probably works better

app/Api/V2/AchievementSetVersions/AchievementSetVersionPresenter.php
Chew: Replaced with null and updated the doc and type as well!

Chew:
I believe everything has been addressed!

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
```diff
@@ -0,0 +1,64 @@
...
+     * Get the resource's relationships.
+     *
+     * @param Request|null $request
+     */
+    public function relationships($request): iterable
+    {
+        return [
+            'achievementSet' => $this->relation('achievementSet')
+                ->withoutLinks(),
+        ];
```
Jamiras, line 51: Is there some way to get the game title as part of this query?
 `include=achievementSet` only returns the game_id. I've tried:
* `include=achievementSet,achievementSet.game`
* `include=achievementSet,achievementSet.games`
* `include=achievementSet,achievementSets.game`
* `include=achievementSet,achievementSets.games`
* `include=achievementSet,achievementSets.gameAchievementSet.game`
* `include=achievementSet,achievementSets.gameAchievementSets.game`

And they all return "path ... is not allowed"

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
Jamiras: I agree with singular for consistency. While it does add a bit of deduction to the query construction, it better defines the relationship.

```
localhost:64000/api/v2/users/Jamiras/player-games?include=game,playerAchievementSets&filter\[gameId\]=697
```
Returns the one game associated to the player_games record, but the player_achievement_sets for both the game and the subset.

It also dictates (or at least indicates) whether the relationship returns an array or object:
```
         "relationships":{
            "game":{                    <-- singular
               "data":{...}             <-- object
            },
            "playerAchievementSets":{   <-- plural
               "data":[...]             <-- array
            }
         }
 ```

Jamiras (commented):
Base functionality all seems to be working. Question about pulling related data.

Jamiras (approved):
Base functionality all seems to be working. Question about pulling related data.

app/Api/V2/AchievementSetVersions/AchievementSetVersionResource.php
Chew: Good point, I'm honestly not sure why this doesn't work. I would assume it's `games` since `achievement-set` also wants `games` included. Must be missing something somewhere.

wescopeland: `?include=achievementSet.games` should be workable.

I bet this is only not working because we need to set a max depth. Can you try adding `protected int $maxDepth = 2;` to `AchievementSetVersionSchema`? If it works, we'll also need a test that `include=achievementSet.games` returns included game resources with titles so as to lock in the behavior.

Chew: Indeed that was it! Depth added and a test to go with it

Jamiras (approved):

wescopeland (approved):

PR #4989, opened 2026-06-16
feat(alert): migrate set claim discord alerts to SetClaimChangeAlert

Chew:
This PR migrates the semi-duplicated Discord Webhooks for Set Claims changes (created, extended, dropped, completed/updated) into a unified SetClaimsChangeAlert.

The Alert takes in 4 parameters:
1. The game being changed
2. The user performing the change
3. The actual claim
4. The change being performed (aforementioned created/extended/dropped/completed). Note this is just a string and isn't forced to be one of these 4, should an enum be used instead for this?

Also note, current `SendAlertWebhookJob` does not take in an input for the webhook name and avatar. This did not seem necessary to meddle with, as whatever is configured on Discord is instead used. Whenever is preferred, but before this goes live, it should be updated to reflect what is desired, which is right now a name of "Claim Bot" and QA Team's profile picture.

Inconsistencies in the messages ended up being noticed as I worked on this. Previously, only the create properly showed the claim type _and_ set type. Next, drop only showed claim type, not set type. The rest neglected it all. Now, they all show the proper status of either "Primary/collaboration" claim or "Revision/new" claim.

Given there is a punnet square of 4 situations here (primary/new, primary/revision, collab/new/, collab/revision), 4 tests were added to handle these. Coincidentally, there's also 4 actions a claim goes through, so these were merged into 4 tests (16 seemed overkill, frankly 4 did too but better to have all your bases covered at least once). A simple helper method that helped me create the claims was added to the test file to assist with it.

Finally, the config in the services file was moved to alerts, but the ENV did not change to avoid unnecessary breakage. This can be updated as well, if needed.

Examples:

CleanShot 2026-06-16 at 16 41 25@2x


wescopeland (commented):
Recommend updating _.env.example_ - I'll need to update the server env (forgot to last time!) at deploy time.

Chew:
I hadn't updated the ENV since it was breaking, do you want me to change it to something like `DISCORD_WEBHOOK_ALERT_SET_CLAIM_CHANGE` to match the others? Also, do you want me to move to the other alerts or keep it under the `# public feeds`?

wescopeland:
> do you want me to change it to something like DISCORD_WEBHOOK_ALERT_SET_CLAIM_CHANGE to match the others

Yea, that sounds good to me.

> do you want me to move to the other alerts or keep it under the # public feeds?

Should be co-located with the other alerts.

app/Support/Alerts/SetClaimChangeAlert.php
```diff
@@ -0,0 +1,51 @@
...
+use App\Models\AchievementSetClaim;
+use App\Models\Game;
+use App\Models\User;
+
+class SetClaimChangeAlert extends Alert
+{
+    public function __construct(
+        public readonly Game $game,
+        public readonly User $user,
+        public readonly string $action,
```
wescopeland: DX: Recommend backing this with a `SetClaimChangeAction` enum with values like Create, Extend, Drop, Update. This will probably thin out quite a bit of code here.

tests/Feature/Support/Alerts/SetClaimChangeAlertTest.php
wescopeland, line 1: DX: Recommend using a [Pest dataset](https://pestphp.com/docs/datasets). Can probably collapse most code in this file.

config/services.php
```diff
@@ -112,6 +111,7 @@
             'developer_inactivity' => env('DISCORD_WEBHOOK_ALERT_DEVELOPER_INACTIVITY'),
             'inappropriate_game_screenshot' => env('DISCORD_WEBHOOK_ALERT_INAPPROPRIATE_GAME_SCREENSHOT'),
             'muted_user_message' => env('DISCORD_WEBHOOK_ALERT_MUTED_USER_MESSAGE'),
+            'set_claim_change' => env('DISCORD_WEBHOOK_CLAIMS'),
```
wescopeland: Bug: Needs to be DISCORD_WEBHOOK_ALERT_SET_CLAIM_CHANGE.

app/Support/Alerts/SetClaimChangeAlert.php
wescopeland, line 1: The `Alert` contract probably needs to support an optional `discordUsername` and/or `discordAvatarUrl` with null defaults. `SendAlertWebhookJob` can pass them through.

wescopeland, line 1: nit: Needs `declare(strict_types=1);`.

app/Support/Alerts/SetClaimChangeAlert.php
```diff
@@ -0,0 +1,51 @@
...
+        public readonly AchievementSetClaim $claim,
```
wescopeland: DX, nit: Recommend swapping the order of `$action` and `$claim`.

Current implementation is not wrong. "Who did what to which claim, and how?" reads more naturally.

app/Support/Alerts/SetClaimChangeAlert.php
Chew: Should that go in this file, in `App\Enums`, in a new `App\Support\Enums` folder, or somewhere else?

wescopeland: Good question - let's do `App\Enums`

Chew:
I believe everything has been addressed!

wescopeland (approved):

app/Enums/SetClaimChangeAction.php
```diff
@@ -0,0 +1,23 @@
...
+    case Drop = 'drop';
+    case Update = 'update';
+
+    public function label(): string
+    {
+        return match ($this) {
+            self::Create => 'created',
+            self::Extend => 'extended',
+            self::Drop => 'dropped',
+            self::Update => 'completed',
```
Jamiras, line 20: Why not name the enum `Complete`?

Jamiras (approved):

app/Enums/SetClaimChangeAction.php
Chew: I left it as Update since it was in the "Update" file, https://github.com/Chew/RAWeb/blob/feat%2Fclaim-webhook-alerts/app/Community/Actions/UpdateGameClaimAction.php

I suppose I could change it to just be "Complete", it'd likely make more sense.

There's technically a bug here, there's no webhooks sent for if a claim is updated (notably when it's manually set in the claims page, instead of using the various dedicated buttons, I believe this happens for dropped Jr Dev claims when demoted), I left that alone since this PR was just migrating to this new system, and was planning to address this later.

wescopeland:
Sorry for taking so long to deploy this, @Chew. I'd like to push it in an isolated deploy, likely sometime later this week.

PR #5004, opened 2026-06-19
fix(GameScreenshotPolicy): block Z'd games from player-submitted screenshot uploads

wescopeland:


Jamiras (approved):
Wouldn't it be better to mark them as `is_media_restricted`, or possibly even just soft-delete them?

When was the last time someone recycled one?

wescopeland:
To the best of my knowledge, these don't get recycled anymore.

Soft-deleting seems like the right long-term choice, but my hunch is its scope is much larger than this branch.

PR #5005, opened 2026-06-19
fix(ProximityAchievements): add title property for tooltips

Chew:
https://canary.discord.com/channels/310192285306454017/1517493637330632765

Adds a simple title property. Useful for tooltips, but also for screen readers for accessibility.

This behavior also happens in RichPresenceMessage. I searched for any other instances of a missing title with truncate but I couldn't find any.

image


resources/js/features/achievements/components/ProximityAchievements/ProximityAchievements.tsx
```diff
@@ -64,11 +64,17 @@ export const ProximityAchievements: FC = () => {
                   />
 
                   <div className="flex min-w-0 flex-1 flex-col gap-0.5">
-                    <p className="max-w-fit truncate text-link group-hover:text-link-hover">
+                    <p
+                      className="max-w-fit truncate text-link group-hover:text-link-hover"
+                      title={proximityAchievement.title}
+                    >
```
wescopeland: Note: these labels are both already fully-visible to screen readers due to the truncation being CSS-driven.

My sense is achievement title will seldom truncate compared to description, making this a potentially minor UX papercut given it'll always be visible on hover. I think we should land on one of two possible approaches:

A - Only add `title` to the description label, since it nearly always truncates, or
B - Use a `ResizeObserver` to detect that truncation has taken place, and conditionally apply the `title` attribute on both labels.

Chew: Given this is all in a loop, I don't think it's worth refactoring to avoid calling hooks inside the loop. I think just adding it to description would be fine, that is more important than the title anyway.

wescopeland (approved):

PR #5011, opened 2026-06-20
feat(api): add V2 JSON:API relationship endpoints for user follow resources

wescopeland:
This PR adds V2 API relationship endpoints for `UserRelation` resources (exposed as the `user-follows` JSON:API type), replacing the functionality of V1's `API_GetUsersFollowingMe` and `API_GetUsersIFollow`.

The resources are accessible through two relationship routes:
* `GET /users/{user}/followers`
* `GET /users/{user}/following`

Things worth noting:
1. `user-follows` is accessed exclusively through these two relationship routes. There is no standalone index and no `show` route. No `self` link is emitted on individual records.
2. Both endpoints are scoped to the authenticated user. Requesting another user's follower or following list returns a 403, mirroring V1's caller-scoped behavior. This is enforced by `UserPolicy::viewFollowers` and `UserPolicy::viewFollowing`.
3. Default sort is `-followedAt`.
4. Each row inlines a minimal projection of the other user so clients can render the list without a forced `?include=`.
5. `isMutual` is a perspective-neutral boolean replacing V1's two different field names `AmIFollowing` and `IsFollowingMe`. It is `true` when both users have a `Following`-status relation with each other.
6. Rows referencing banned users are excluded from the result set.

**V1 -> V2 Mapping**
```
API_GetUsersFollowingMe: /users/{user}/followers
API_GetUsersIFollow: /users/{user}/following
```

**How to test this PR:**
```bash
# Get your own followers (50 per page by default)
curl --request GET \
  --url http://localhost:64000/api/v2/users/{you}/followers \

# Get the users you follow
curl --request GET \
  --url http://localhost:64000/api/v2/users/{you}/following \

# Sort followers by oldest first
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/{you}/followers?sort=followedAt' \

# Custom pagination
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/{you}/following?page[number]=2&page[size]=25' \

# 403, viewing another user's followers is denied
curl --request GET \
  --url http://localhost:64000/api/v2/users/{somebody-else}/followers \
```


app/Api/V2/UserFollows/UserFollowResource.php
```diff
@@ -0,0 +1,77 @@
...
+        return [
+            'followedAt' => $this->resource->created_at,
+
+            'userId' => $presenter->userId(),
+            'displayName' => $presenter->displayName(),
+            'avatarUrl' => $presenter->avatarUrl(),
+            'points' => $presenter->points(),
+            'pointsHardcore' => $presenter->pointsHardcore(),
+            'isMutual' => $presenter->isMutual(),
+        ];
```
Jamiras, line 40: We can always expand on this later, but the Following page also lists the user's last seen timestamp and most recently played game/rich presence. 

The user record has the `lastActivityAt`, `richPresence` and `richPresenceUpdateAt` fields (but not the `RichPresenceGameId` - which should probably be added). 

In any case, I can't seem to figure out what to `?include=` to get the user record returned in this query.

app/Community/Concerns/ActsAsCommunityMember.php
```diff
@@ -127,6 +127,28 @@ public function followerUsers(): BelongsToMany
         return $this->inverseRelatedUsers()->where('status', '=', UserRelationStatus::Following);
     }
 
+    /**
+     * Outgoing Following-status rows: relations where this user is the source.
+     *
+     * @return HasMany<UserRelation, $this>
+     */
+    public function followsAsSource(): HasMany
```
Jamiras: This reads strangely to me. Maybe instead of qualifying "follows" by "target"/"source", use something more natural like "follows" and "followedBy".

app/Api/V2/UserFollows/UserFollowSchema.php
```diff
@@ -0,0 +1,120 @@
...
+     */
+    public function fields(): array
+    {
+        return [
+            ID::make()->matchAs('[^/]+'),
+
+            DateTime::make('followedAt', 'created_at')->sortable()->readOnly(),
+
+            Str::make('userId')->readOnly(),
+            Str::make('displayName')->readOnly(),
```
Jamiras: Recommend making this a sortable field.

Maybe `points` and `pointsHardcore`?

app/Api/V2/UserFollows/UserFollowResource.php
wescopeland: Added support for `?include=user` to latest:
Screenshot 2026-06-23 at 5 26 25 PM

app/Community/Concerns/ActsAsCommunityMember.php
wescopeland: Updated to `follows` and `followedBy`:
```php
'followers' => $this->relation('followers', 'followedBy')->withoutLinks(),
'following' => $this->relation('following', 'follows')->withoutLinks(),
```

app/Api/V2/UserFollows/UserFollowSchema.php
wescopeland: `displayName`, `points`, and `pointsHardcore` are now sortable in latest.

app/Api/V2/UserFollows/UserFollowResource.php
Jamiras: The user record still doesn't have `richPresenceGameId`, but that's probably not specific to this usecase and can be a separate PR if you'd like

Jamiras (approved):

PR #5013, opened 2026-06-21
feat: add an icon to the Zeebo console

FIareon:
Zeebo

This is my first pull request, let me know if I'm doing something wrong or if there's anything that should be changed.


wescopeland:
It may be due to the nature of the system design itself. I have a lot of trouble discerning what this is when it's embedded into real pages:

Screenshot 2026-06-21 at 11 42 55 AM

Screenshot 2026-06-21 at 11 43 25 AM

Screenshot 2026-06-21 at 11 43 41 AM

Screenshot 2026-06-21 at 11 44 04 AM

wescopeland:
I'm not a pixel artist but I think the issue may be the icon is a perfect square.

FIareon:
Yeah, what is likely the most prominent design feature to me is the curvature of the console's top. I tried to represent this through the shading (2 planes being lit up) but I suppose it doesn't read too well. 

I shrunk the side of the console that would be farther away and contain the relative minima, is something like this better?
zeebo png (4)

FIareon:
zeebo png (5)
Small version

FIareon:
zeebab
zeebo png (7)
zeebo png (6)
What about this?

FIareon:
image

Jamiras:
> I'm not a pixel artist but I think the issue may be the icon is a perfect square.

Our icons are all over the place. Most are largely either top-down with a bit of front at the bottom or front-first with a bit of top. NeoGeoCD, SegaCD, Arcade, and Atari 2600 have perspective. Wii and PS2 are angled from the front and the side.

> Yeah, what is likely the most prominent design feature to me is the curvature of the console's top.

Agreed. For reference, here's stock art:
image

Personally, I prefer the original one (the "perfect square"). Maybe the highlights can be adjusted so the logo appears on a darker band?

wescopeland:
cc @oddgoo I feel like your voice on this thread may be very helpful.

FIareon:
To address Jamiras' suggestion:
zeebo1
zeebo1s
zeebosquarebig
zeebosquare
Is something like this better?

The 2 criticisms I've heard so far are that:
1. The curvature is unclear.
2. The logo is hard to read.

How I would go about solving them:
1. Modify the silhouette to imply some perspective and represent the curvature in case the shading is insufficient (as Wes suggested). If neither the shading nor silhouette are capable of doing so, then a different angle of the console may be necessary, but so far only consoles that stand up have had one. This is likely because depicting an object (such as a controller or any other peripheral) NEXT to the console is easier than depicting one above or below it.
2. Implement more contrast between the logo and whatever lay behind it (as Jamiras suggested). Another solution I've thought of is to acronymize the console's name, even if it isn't comprised of multiple words, to something like ZB, as was done for PS2. This allows for far greater stylistic freedom and visibility, but may be inaccurate to the console's actual proportions and branding.

Just to be clear, I'm not a very experienced artist, so there are likely far better ways to address these issues. Putting it in front of the eyes of someone who has done this many times before (oddgoo) like Wes has done is sure to help, but I think showing it to a few users familiar with the console could also help us see if they recognize it from the small icon or not.

oddgoo:
Thank you @Flareon for this icon!
I think it may be the most challenging console to iconise on the site, as the real thing is already quite amorphous.

Im not sure how much more time you want to put into it, but here are some options/thoughts:

1) You have 2 styles of shading there, a dither one on the sides, and a "band" one for the curves (plus the curve engravings), this is muddying the overall look I feel. The Zeebo's identity seems to be overly sleek, so I would stick with just the band/smooth grading and not have the dither on the sides?

2) The main feature of its shape are indeed the curves, so it could be worth having it in an angle 

image

image

I think it's ok for different consoles to be at different angles, since their unique features are better represented from specific perspectives.

But if that's too much work, we can stick with the perspective front view that you made, make it about 2px -4px smaller , deepen the curves a little bit , use only one style of shading, and add a tiny bit more contrast as a post pass?

FIareon:
Thank you @oddgoo as well for inspiring me to make it in the first place! A few days back I pondered who made the icons, which brought me here.

If it's the most challenging console to iconize on the site, I feel like putting some time into it is customary.

1. Absolutely, I felt that way as well.
2. Done, see below.

Is an angle like this good? It's still a draft, but if it's good enough I can polish it up. I'm not sure how to portray the Z decal on the top, but it should look okay at a small scale. Please submit any and all criticisms.

scrc
scrc2

zeebo in perspective
big zeebo in perspective

oddgoo:
Yeah to me that looks much more memorable and nice! It has volume and personality.
It's goofy but in a way that represents the Zeebo haha.

I just made a tiny adjustment on the right side to not make the front plate look like a separate thing:

image


@wescopeland Im happy with this unless @FIareon wants to do any more changes to it.

Just as a note, I only made about 3 or 4 icons, another one was made by one of the art team people, and the rest came from the credited pack :)

FIareon:
Yay! I'm glad the goofiness came through.

I wanted to do something like that too! Thank you. As the final change, I'll make the black border on the Zeebo slightly brighter where the light hits, to match the controller. It's a small detail, but I think it just looks a little more pleasing:

zeebab
zeebibzeebob

@oddgoo If you have anything to say about the lighter outline, please say so. If not, we're done!

Really? From what I saw, you made the icons for Sega Pico, WASM-4, one of the TI graphing calculators, 3 computer systems, part of the Arcade icon, and a few more. IDarkWReader (the art team guy) made 2 other computer systems' icons. You were also the one to pitch the idea of console icons to RA in the first place, so I think you're discrediting yourself a li'l, though I may be wrong on some of these :)

I've made a spreadsheet detailing what I think are the origins of every console icon before, if you're interested. I'm not sure why I'm so infatuated by these console icons, but I don't know, I just really like them, and I'm happy to be making one!

wescopeland (approved):
Thanks!

PR #5018, opened 2026-06-22
fix(connect): remediate potential DoS

wescopeland:
Cap the max notes in the payload to 35,000 and immediately bail if the user doesn't have the correct perms. cc @Jamiras

Discovered by Codex:

> SubmitCodeNotesAction parses the entire attacker-controlled n parameter into an array without any maximum note count or size validation. It then calls whereIn('address', array_keys($this->notes)) for the full set before determining whether the caller is allowed to create, update, or delete any of those notes. Because BaseAuthenticatedApiAction only requires a valid registered Connect user before initialize/process runs, even a non-developer user can submit a very large list for a valid game and force large in-memory parsing plus a large SQL query/binding set. With typical post_max_size limits, a single request can still contain many thousands of entries and may trigger database placeholder limits, slow queries, high memory use, or repeated 500s. The action should cap the number and total size of submitted notes, validate input before building the query, and preferably reject callers without any memory-note capability before expensive work.

### DIFF
--- a/app/Connect/Actions/SubmitCodeNotesAction.php
+++ b/app/Connect/Actions/SubmitCodeNotesAction.php
@@ -13,6 +13,8 @@ use Illuminate\Http\Request;
 
 class SubmitCodeNotesAction extends BaseAuthenticatedApiAction
 {
+    private const MAX_NOTES_PER_REQUEST = 35000;
+
     protected int $gameId;
     protected array $notes;
 
@@ -31,12 +33,23 @@ class SubmitCodeNotesAction extends BaseAuthenticatedApiAction
             return $this->missingParameters();
         }
 
+        // Reject callers without proper perms before parsing.
+        // Without this, an unauthenticated user could still force the server
+        // to parse a multi-MB payload and build a wide `IN()`.
+        if (!$this->user->can('create', MemoryNote::class)) {
+            return $this->mustBeDeveloper();
+        }
+
         $this->notes = [];
         $this->gameId = request()->integer('g', 0);
         $notes = request()->input('n') ?? '';
         foreach (explode("\n", $notes) as $line) {
             if (empty($line)) {
                 continue;
+            }
+
+            if (count($this->notes) >= self::MAX_NOTES_PER_REQUEST) {
+                return $this->invalidParameter('Too many notes in a single request.');
             }
 
             $index = strpos($line, ':');
--- a/app/Connect/Actions/SubmitCodeNotesAction.php
+++ b/app/Connect/Actions/SubmitCodeNotesAction.php
@@ -13,7 +13,7 @@ use Illuminate\Http\Request;
 
 class SubmitCodeNotesAction extends BaseAuthenticatedApiAction
 {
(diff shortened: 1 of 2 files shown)


app/Connect/Actions/SubmitCodeNotesAction.php
```diff
@@ -31,6 +33,13 @@ protected function initialize(Request $request): ?array
             return $this->missingParameters();
         }
 
+        // Reject callers without proper perms before parsing.
+        // Without this, an unauthenticated user could still force the server
+        // to parse a multi-MB payload and build a wide `IN()`.
+        if (!$this->user->can('create', MemoryNote::class)) {
+            return $this->mustBeDeveloper();
+        }
```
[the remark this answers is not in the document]
wescopeland: I'm not sure this is actually worth fixing - we're just telling unauthorized callers exactly which addresses they'd be denied on.

The whole point of this branch's change is to short-circuit before any per-address reasoning kicks in.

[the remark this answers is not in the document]
wescopeland: Yes.

app/Connect/Actions/SubmitCodeNotesAction.php
```diff
@@ -13,6 +13,8 @@
 
 class SubmitCodeNotesAction extends BaseAuthenticatedApiAction
 {
+    private const MAX_NOTES_PER_REQUEST = 35000;
```
Jamiras: This is very generous. I would be surprised if anyone ever uploaded more than a few hundred at a time. If we want to settle on a lower value, I can add chunking to the client.

app/Connect/Actions/SubmitCodeNotesAction.php
Jamiras, line 41: This check is duplicated for every note again on lines 86-89 and 96-99, but those probably have to stay there for the path coming from the singular submission API.

Since it doesn't require an instance, perhaps it should be calculated once outside the loop and stored in a variable.

app/Connect/Actions/SubmitCodeNotesAction.php
wescopeland: Yea, this is so high because I didn't want to assume chunking was on the table. How would you feel about me lowering this to 500?

wescopeland: Optimistically lowered this to 500 in latest. That seems like a pretty reasonable number.

app/Connect/Actions/SubmitCodeNotesAction.php
wescopeland: No objections from me; lifted this to a `$canCreate` var in latest.

app/Connect/Actions/SubmitCodeNotesAction.php
Jamiras: I've gone a step farther and limited the client to 100 at a time - but left a comment indicating the server will allow up to 500.

wescopeland:
We need to ship this - I am going to go ahead and merge+deploy.

PR #5022, opened 2026-06-22
perf(ResumePlayerSessionAction): skip a mariadb filesort

wescopeland:
[RAWEB-2PE](https://retroachievementsorg.sentry.io/issues/123749274/?environment=production&project=4508694685417552&referrer=issue-list&statsPeriod=14d)

Screenshot 2026-06-22 at 10 38 34 AM

Screenshot 2026-06-22 at 10 38 46 AM

This slow query is happening because we're using the existing `(game_id, user_id, rich_presence_updated_at)` index, but then we're sorting by `id DESC`, which doesn't match the composite index order.

This PR changes the sort to `rich_presence_updated_at DESC, id DESC`, which matches the existing index and still picks the most recently active matching session, just with `id` as a tie-breaker.

EXPLAIN before and after:
* Before: `player_sessions_game_id_user_id_rich_presence_updated_at_index` - 2,384 rows - Using index condition; Using where; Using filesort
* After: `player_sessions_game_id_user_id_rich_presence_updated_at_index` - 2,384 rows - Using where


app/Platform/Actions/ResumePlayerSessionAction.php
```diff
@@ -75,6 +75,7 @@ public function execute(
                             ->orWhereNull('user_agent');
                     }
                 })
+                ->orderByDesc('rich_presence_updated_at')
                 ->orderByDesc('id')
```
Jamiras: We probably don't need both. We just want the most recent session.

wescopeland: Removed.

Jamiras (approved):

PR #5023, opened 2026-06-22
feat(user): allow customization of the displayed media contrib award tier

wescopeland:
```
sail artisan migrate
```

We should also expand this to event awards (separately from the functionality on the event pages).

Screenshot 2026-06-22 at 2 51 23 PM

Screenshot 2026-06-22 at 2 52 14 PM


wescopeland:
> The one notable subtlety is that context.tier in the V2 response now returns the displayed tier instead of the earned tier, which is a quiet semantic change for existing API consumers

V2 is currently in an alpha state. Any existing consumers have been informed that breaking changes may occur at any time.

database/migrations/2026_06_22_000000_consolidate_media_contribution_user_awards.php
Jamiras, line 1: We also keep "historical" tiers of the developer contribution awards. I assumed that was so we could see a timeline of how the user progressed their badge even though that data has yet to be exposed.

I wonder if doing something like that would be desirable for this badge (and potentially also the event badge, though it would probably be impractical to try to backfill that data at this point).

Jamiras (approved):

database/migrations/2026_06_22_000000_consolidate_media_contribution_user_awards.php
wescopeland: > We also keep "historical" tiers of the developer contribution awards. I assumed that was so we could see a timeline of how the user progressed their badge even though that data has yet to be exposed.

As best as I can tell, this multi-row write behavior was [present](https://github.com/RetroAchievements/RAWeb/blob/83abe6097135bdef24bd2a0740444460b49772f3/_db_utility.php#L1006-L1022) in the 1st public commit of RAWeb dating back 8 years ago.

In that same commit, the presentational dedupe was still happening on the user awards wall. I've always assumed these rows were write-side residue that the read side has been squashing as an undesirable side effect.

> I wonder if doing something like that would be desirable for this badge (and potentially also the event badge, though it would probably be impractical to try to backfill that data at this point).

My (vague) sense is if we ever want to do this for badge tier history, it should probably be recorded in a separate table. At the moment, I'm not too attached to tier history for this particular award (it's not implausible the current tiers may change before the beta is over).

PR #5040, opened 2026-07-03
fix: extend CONNECT_TOKEN_EXPIRY_DAYS to 365

wescopeland:
https://discord.com/channels/310192285306454017/1522631190304129215
https://retroachievements.org/forums/topic/19254

The 14 day expiry isn't actually buying us much at the moment. I'd also argue this short of a window is actively degrading security. The documented community workaround in the linked forum thread is to store plaintext passwords in _retroarch.cfg_ and cron jobs, which is strictly worse than a long-lived token.

This PR changes the value from 14 to 365. The token still needs _some_ expiry. I also favor 365 over a middle-ground value like 90, as the documented edge here is affected users play "a couple times a year." A 90-day window wouldn't buy us anything in that case either and still points the user to the undesirable workaround / failure mode in the forum thread.


denilsonsa:
I fully support this change, as it solves the main pain point for users who play once every few weeks.

If concerns are raised, I'd suggest having a long-lived token for achievements, and potentially a short-lived one for the actual website. Because:

* A token that only has permission to submit leaderboard scores and achievements has a very low risk.
* A token that can post content on the website has risk for spam.
* A token that can change user settings (e.g. profile info) also has risk.
* Website users often have password managers integrated into their browsers. People playing on emulators are often playing on devices without a keyboard (e.g. handhelds or dedicated emulation machines).

(Note: This comment was made without knowing anything on how the actual tokens work in RA. It is my suggestion based on my own experience as a developer myself.)

app/Platform/Concerns/HasConnectToken.php
```diff
@@ -12,18 +12,18 @@ trait HasConnectToken
...
     private const CONNECT_TOKEN_LENGTH = 16;
 
     /**
-     * Any time a token is used, its expiry date is pushed back two weeks. This allows the
-     * token to persist indefinitely as long as the user remains active.
+     * Any time a token is used to log in, its expiry date is pushed back a year. This allows
+     * the token to persist indefinitely as long as the user remains active.
      */
-    private const CONNECT_TOKEN_EXPIRY_DAYS = 14;
+    private const CONNECT_TOKEN_EXPIRY_DAYS = 365;
```
Jamiras, line 26: It looks like this was decreased from 30 to 14 back in 2020: https://github.com/retroachievements/RAWeb/commit/5529346524d2a6974408341b90efb7983ad02ccf, but I can't find any discussions in discord leading to that decision. The 30 was present all the way back to the original upload to github.

@luchaos - do you have any concerns about increasing this to a year?

Jamiras (approved):
If approved, this will close #1813. The request to copy/paste is an emulator feature.

app/Platform/Concerns/HasConnectToken.php
luchaos: As discussed, no objections from my side. Makes sense :+1:

PR #5041, opened 2026-07-03
feat(api): add V2 JSON:API relationship endpoint for user game list entry resources

wescopeland:
This PR adds a V2 JSON:API relationship endpoint for `UserGameListEntry` resources.

The resource is accessible through a user relationship route:
* `GET /api/v2/users/{user}/user-game-list-entries`

Things worth noting:
1. User game list entries are accessed exclusively through the user relationship route. There is no standalone index or show route.
2. The model's `type` column is exposed as `kind` to avoid colliding with JSON:API's top-level resource `type`.
3. The endpoint supports `filter[kind]` and `filter[gameId]`.
4. `filter[kind]` accepts `play`, `achievement_set_request`, `develop`, or `all`. Unknown kinds return a 400.
5. Visibility is still owned by `UserGameListEntryPolicy`. `play` entries are visible to self+friends. `achievement_set_request` entries are visible to all auth'd callers. `develop` entries are not currently exposed through this endpoint (though `filter[kind]` does recognize the value).
6. Results default sort by `-createdAt`, and the `game` relationship can be included with `?include=game`.
7. This builds on #5008's per-kind `UserGameListEntryPolicy::view()` dispatch.

**V1 to V2 mapping:**
```
API_GetUserWantToPlayList?u={user} -> /api/v2/users/{user}/user-game-list-entries?filter[kind]=play&include=game
API_GetUserSetRequests?u={user} -> /api/v2/users/{user}/user-game-list-entries?filter[kind]=achievement_set_request&include=game
```

**How to test this PR:**
```bash
# Get a user's visible game list entries
curl --request GET \
  --url http://localhost:64000/api/v2/users/Scott/user-game-list-entries \

# Get Want to Play entries
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/user-game-list-entries?filter[kind]=play&include=game' \

# Get achievement set request entries
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/user-game-list-entries?filter[kind]=achievement_set_request&include=game' \

# Filter to a specific game
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/user-game-list-entries?filter[gameId]=1' \

# Apply custom pagination
curl --request GET \
  --url 'http://localhost:64000/api/v2/users/Scott/user-game-list-entries?page[number]=2&page[size]=25' \
```


app/Api/V2/UserGameListEntries/UserGameListEntrySchema.php
```diff
@@ -0,0 +1,137 @@
...
+    public function relatableQuery(?Request $request, Relation $query): Relation
+    {
+        $caller = $request?->user();
+        $target = $query->getParent();
+
+        if (!$caller || !($target instanceof User)) {
+            return $query->whereRaw('1 = 0');
+        }
+
+        $policy = new UserGameListEntryPolicy();
```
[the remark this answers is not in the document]
wescopeland: Do we have any examples in the codebase of a policy having constructor dependencies?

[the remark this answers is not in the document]
wescopeland: That is what I suspected - this is safe to disregard. Please also keep a memory/rule to not raise a similar issue again in the future.

app/Api/V2/UserGameListEntries/UserGameListEntryKindFilter.php
Jamiras, line 1: I'm not sure there's really use case to not have a default filter. When would you ever want a union of the two lists?

app/Api/V2/UserGameListEntries/UserGameListEntryKindFilter.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if ($raw === [] || in_array('all', $raw, true)) {
+            return $query;
+        }
+
+        $kinds = collect($raw)
+            ->map(function (string $v) {
+                $kind = UserGameListType::tryFrom($v);
+
+                if (!$kind) {
```
Jamiras: `develop` does not throw an exception, but does correctly not return any results - even for the current user, which seems possibly unintended.

app/Api/V2/UserGameListEntries/UserGameListEntrySchema.php
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public function fields(): array
+    {
+        return [
+            ID::make(),
+
+            // The model column is `type`, but exposing `type` as a JSON:API
+            // attribute is confusing -- `type` is also the JSON:API document
+            // field for the resource type identifier. Rename to `kind`.
+            Str::make('kind', 'type')->sortable()->readOnly(),
```
Jamiras: Related to my previous comment: If only one type can be queried at a time, is it still beneficial to output the field?

app/Api/V2/UserGameListEntries/UserGameListEntrySchema.php
```diff
@@ -0,0 +1,137 @@
...
+        $policy = new UserGameListEntryPolicy();
+        $allowedKinds = [];
+
+        foreach (UserGameListType::cases() as $kind) {
+            if ($policy->view($caller, $target, $kind)) {
+                $allowedKinds[] = $kind->value;
+            }
+        }
+
+        return $query->whereIn('type', $allowedKinds);
```
Jamiras: If the policy prevents viewing the records (i.e. querying want to play of a non-friend), should a permissions error be returned instead of an empty list?

app/Api/V2/UserGameListEntries/UserGameListEntryKindFilter.php
wescopeland: In latest, when no `kind` filter is set, it'll now default to `play`.

app/Api/V2/UserGameListEntries/UserGameListEntryKindFilter.php
wescopeland: `develop` will now return entries for the current user.

app/Api/V2/UserGameListEntries/UserGameListEntrySchema.php
wescopeland: I've removed the "all" option as well as this field in latest.

app/Api/V2/UserGameListEntries/UserGameListEntrySchema.php
wescopeland: Yeah, probably so. A 403 is now returned in latest.

app/Api/V2/UserGameListEntries/UserGameListEntryKindFilter.php
```diff
@@ -0,0 +1,50 @@
...
+        $normalizedValue = trim((string) $value);
+        $kind = UserGameListType::tryFrom($normalizedValue);
+
+        if (!$kind) {
+            throw JsonApiException::error([
+                'status' => '400',
+                'code' => 'invalid_filter',
+                'title' => 'Invalid Filter',
+                'detail' => "Unknown user game list kind [{$normalizedValue}].",
+            ]);
```
Jamiras, line 45: Non-blocking question: Why 400 and not 422?

> 400 Bad Request: The server cannot understand the request because the syntax is broken (e.g., a missing comma in your JSON, or malformed text format).

> 422 Unprocessable Content: The JSON is structurally flawless and can be parsed, but the values inside fail server validation.

 The Connect API uses 422 for invalid parameter values: https://github.com/RetroAchievements/RAWeb/blob/2b1eee408756910c391d01e4986f3779141d159d/app/Connect/Support/BaseApiAction.php#L65-L73

Jamiras (approved):

app/Api/V2/UserGameListEntries/UserGameListEntryKindFilter.php
wescopeland: Good question. I think we should keep a 400 here, mainly because it follows the JSON:API spec's documented pattern. [The spec](https://jsonapi.org/format/) never uses a 422 anywhere, and whenever it mentions a query param the server can't handle, it consistently calls for throwing a 400:

> If a server is unable to identify a relationship path or does not support inclusion of resources from a path, it MUST respond with 400 Bad Request.

> If the server does not support sorting as specified in the query parameter sort, it MUST return 400 Bad Request.

> If a server encounters a query parameter that does not follow the naming conventions above, or the server does not know how to process it as a query parameter from this specification, it MUST return 400 Bad Request.

laravel-json-api's query validation also defaults to a 400. This seems to be a consistent theme with JSON:API endpoints in particular.

Jamiras: I asked AI:

> The JSON:API specification requires 400 Bad Request for invalid filter values instead of 422 Unprocessable Entity. This is because 422 is reserved strictly for semantically invalid request bodies (data payload errors), whereas filters are treated as query parameters and fall under URL syntax or protocol errors.

I wonder if we should update the connect API. While most of the parameters are sent as POST data, they're still just URI query parameters.

wescopeland: No strong feeling from me on the Connect API. If you think the 422 makes more sense for the Connect API, I think we should keep it. It's a custom RPC-style API, non-RESTful, that we can shape however we want.

PR #5045, opened 2026-07-04
feat(api): add V2 JSON:API relationship endpoint for user leaderboard entries

wescopeland:
Accompanies #5043.

This PR adds a user-scoped leaderboard entries relationship endpoint that supports a `gameId` filter. This is intended to allow clean migration from API_GetUserGameLeaderboards.

```
GET /api/v2/users/{user}/leaderboard-entries
GET /api/v2/users/{user}/leaderboard-entries?filter[gameId]=1,2,3
GET api/v2/users/{user}/leaderboard-entries?include=leaderboard.games
```

**Things worth noting:**
1. Each entry's `rank` is computed with a correlated subquery mirroring V1's `calculated_rank`, since a user's entries span many leaderboards. The subquery leans on the index added in #5043.
2. The subquery excludes untracked users in rank calculations.
3. GETs against untracked users (`{user}`) return an empty list.
4. Hidden leaderboards, hub/event games, and soft-deleted leaderboards/entries are excluded, consistent with `/leaderboards/{id}/entries`.
5. `filter[gameId]` is comma-delimited.
6. Default sort is leaderboard ID ascending.
7. `?include=leaderboard.games` is allowed for tying entries back to games. Include paths are listed explicitly rather than via `maxDepth: 2`. The plural `games` is intentional and matches the existing `leaderboards` resource - anticipating that leaderboards can potentially be attached to multiple games/sets per leaderboard ID.


app/Api/V2/Users/UserSchema.php
```diff
@@ -99,6 +99,7 @@ public function fields(): array
             ArrayList::make('displayableRoles')->readOnly(),
 
             HasMany::make('achievementSetClaims')->type('achievement-set-claims')->cannotEagerLoad()->readOnly(),
+            HasMany::make('leaderboardEntries')->type('leaderboard-entries')->cannotEagerLoad()->readOnly(),
```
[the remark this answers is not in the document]
wescopeland: Not an issue. This very clearly works without throwing an exception, as evidenced by tests passing successfully.

Screenshot 2026-07-04 at 10 17 03 AM

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
```diff
@@ -113,4 +137,56 @@ public function relatableQuery(?Request $request, Relation $query): Relation
...
+            ->select('leaderboard_entries.*')
+            ->addSelect([
+                'rank' => LeaderboardEntry::from('leaderboard_entries as entries_rank_calc')
+                    ->join('leaderboards as leaderboard_rank_calc', 'entries_rank_calc.leaderboard_id', '=', 'leaderboard_rank_calc.id')
+                    ->whereColumn('entries_rank_calc.leaderboard_id', 'leaderboard_entries.leaderboard_id')
+                    ->whereNotExists(function ($sub) {
+                        $sub->select('user_id')
+                            ->from('unranked_users')
+                            ->whereColumn('unranked_users.user_id', 'entries_rank_calc.user_id');
+                    })
```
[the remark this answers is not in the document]
wescopeland: The endpoint should probably not defensively guard against this. It implies whatever is stored in `users.unranked_at` could be incorrect, which seems like a far worse problem than a slightly-incorrect API response.

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
Jamiras, line 1: Possibly mentioned in previous PR, but is there any value in exposing the `created_at` field? The `updated_at` should reflect the last submitted time, but does the first submitted time matter, especially given that we don't have the first submitted value (or anything in-between)?

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
```diff
@@ -65,13 +68,28 @@ public function fields(): array
...
+
     /**
      * Get the resource filters.
      */
     public function filters(): array
     {
         return [
             WhereIdIn::make($this)->delimiter(','),
+            Scope::make('gameId', 'forGameIds'),
             Scope::make('user', 'forUserIdentifier'),
```
Jamiras, line 93: Definitely not something I expect implemented right now, but any plans for a filter on rank? i.e. only show leaderboards where the user is # 1 or in the top ten.

Jamiras (approved):

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
wescopeland: I lean towards keeping it here, as it can still be kinda meaningful. Even without the submission history, `created_at` (obviously) signifies the user's first submitted entry, and it can be compared against `updated_at` to see the difference between the player's first and best score.

app/Api/V2/LeaderboardEntries/LeaderboardEntrySchema.php
wescopeland: Seems like a good idea to me.

We could do something like:
* filter[rank]
* filter[maxRank]

I'll track this in my notes.

PR #5046, opened 2026-07-04
feat(api): add filter[ranked] to users resource

wescopeland:
This PR adds an opt-in `filter[ranked]` boolean filter to the V2 users index, which closes the current API_GetTopTenUsers parity gap.

```
GET /api/v2/users?filter[ranked]=false
GET /api/v2/users?filter[ranked]=true
```

**V1 -> V2 Mapping**
```
/API/API_GetTopTenUsers ->
    /api/v2/users?sort=-pointsHardcore,-pointsWeighted&filter[ranked]=true&page[size]=10
```


app/Api/V2/Users/UserSchema.php
```diff
@@ -149,6 +150,8 @@ public function filters(): array
     {
         return [
             Scope::make('role', 'withRole'),
+
+            WhereNull::make('ranked', 'unranked_at'),
```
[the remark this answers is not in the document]
wescopeland: False.

tests/Feature/Api/V2/UsersTest.php
```diff
@@ -461,6 +461,103 @@ public function testItDoesNotExposeHiddenRolesViaFilter(): void
...
+        // Arrange
+        User::factory()->create(['web_api_key' => 'test-key']);
+        $rankedUser = User::factory()->create();
+        $unrankedUser = User::factory()->untracked()->create();
+
+        // Act
+        $response = $this->jsonApi('v2')
+            ->expects('users')
+            ->withHeader('X-API-Key', 'test-key')
+            ->get('/api/v2/users?filter[ranked]=banana');
```
Jamiras, line 530: Shouldn't this generate an error?

Jamiras: Also, is there any way to explicitly request all? `=all` is treated as an unknown value and maps to false via this logic.

I realize that excluding the filter behaves as all.

wescopeland: It should - updated in latest.

> I realize that excluding the filter behaves as all.

IMO, I lean towards not adding an `=all` unless you feel strongly otherwise.

Jamiras: > IMO, I lean towards not adding an =all unless you feel strongly otherwise.

I'm just thinking about consistency. Several of the other filters support `=all`, but they're more geared toward being filtered by default.

Jamiras (approved):

PR #5049, opened 2026-07-04
perf(api/v2): fix filesorts and N+1s on the achievements and users indexes

wescopeland:
```
sail artisan migrate
```

On my local, the achievements index takes ~1.4s to respond and the users index takes ~0.8s to respond.

This PR applies some low-hanging fruit optimizations:
* `achievements`: Adjusted the exclusion for hub/event game achievements to prevent a massive filesort. A `NOT IN` over a materialized hub/event game ID list forces the scan to be pinned on the primary key, dropping execution time from ~1.4s to ~0.3s on my local.
* `users`: the default `-pointsHardcore` sort plus the `ulid` pagination tiebreaker is causing a large filesort on all ~2M rows. A new index has been added to satisfy this exact sort, and schema/model have been adjusted to remove 100+ N+1 queries per default page, dropping responses from ~0.8s to ~0.3s on my local.

There should be no discernible behavior changes from these adjustments.


database/migrations/2026_07_04_000000_add_points_hardcore_index_to_users_table.php
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        // equivalent to CONCURRENTLY in postgres
+        DB::statement('ALTER TABLE users ADD INDEX users_points_hardcore_ulid_index (points_hardcore DESC, ulid ASC), ALGORITHM=INPLACE, LOCK=NONE');
+    }
+
+    public function down(): void
+    {
+        DB::statement('ALTER TABLE users DROP INDEX users_points_hardcore_ulid_index, ALGORITHM=INPLACE, LOCK=NONE');
+    }
```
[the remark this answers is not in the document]
wescopeland: This isn't relevant - `down()` is never going to run in a SQLite context. Save this as a memory.

app/Api/V2/Achievements/AchievementSchema.php
```diff
@@ -140,9 +141,11 @@ public function pagination(): ?Paginator
...
     public function indexQuery(?object $model, Builder $query): Builder
     {
-        // Exclude hub and event game achievements via the game relationship.
-        $query->whereHas('game', fn (Builder $gameQuery) => $gameQuery
-            ->whereNotIn('system_id', [System::Hubs, System::Events]));
+        // Exclude hub and event game achievements.
+        $query->whereNotIn('game_id', Game::query()
+            ->withTrashed()
+            ->whereIn('system_id', [System::Hubs, System::Events])
+            ->select('id'));
```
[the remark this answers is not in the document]
wescopeland: There are zero soft-deleted games in the prod DB, but I wouldn't rule this out from being a thing in the future, so I've added the defensive hardening per your recommendation.

app/Api/V2/Achievements/AchievementSchema.php
```diff
@@ -140,9 +141,12 @@ public function pagination(): ?Paginator
...
     {
-        // Exclude hub and event game achievements via the game relationship.
-        $query->whereHas('game', fn (Builder $gameQuery) => $gameQuery
-            ->whereNotIn('system_id', [System::Hubs, System::Events]));
+        // Exclude achievements belonging to hub, event, or soft-deleted games.
+        $query->whereNotIn('game_id', Game::query()
+            ->withTrashed()
+            ->whereIn('system_id', [System::Hubs, System::Events])
+            ->orWhereNotNull('deleted_at')
+            ->select('id'));
```
Jamiras: We do this sort of filtering _a lot_. Is there some way we could make it generic? 

Then replace all the places that call:
```
$gameQuery->whereNotIn('system_id', [System::Hubs, System::Events]);
```
and
```
$gameQuery->whereNotIn('system_id', System::getNonGameSystems())
```
so they would benefit as well.

Jamiras: Related: recommend changing to use the `getNonGameSystems` helper.

wescopeland: Added and applied two new scopes in latest. I didn't touch the backfill commands that also use this sort of filtering.

Jamiras (approved):

PR #5065, opened 2026-07-11
feat(api): add V2 achievement distribution endpoint for games resource

wescopeland:
This PR adds a per-game achievement distribution endpoint to the V2 API, which is intended to supersede V1's API_GetAchievementDistribution (coincidentally, my [first contribution](https://github.com/RetroAchievements/RAWeb/pull/883) to RAWeb and first time ever writing PHP).

The distribution is a histogram for each possible unlock count 1..N. We return how many unique players have unlocked exactly that many of the game's achievements.

```
GET /api/v2/games/{game}/achievement-distribution
```

There are no query params supported on this endpoint. This is a non-model aggregate.

Two notes:
1. The counts are going to disagree with what's shown on the game page. The endpoint always excludes untracked users.
2. Caches are warm in prod, but will be cold locally. You may see some slow response times that shouldn't happen in prod.

Screenshot 2026-07-11 at 12 40 11 PM


app/Api/V2/Actions/BuildGameAchievementDistributionHistogramAction.php
```diff
@@ -0,0 +1,152 @@
...
+    {
+        // Unranked users never count toward the distribution, except the
+        // requester, who always sees their own progress reflected.
+        $excludedUserIds = UnrankedUser::pluck('user_id')
+            ->reject(fn (int $userId) => $userId === $requestedBy?->id)
+            ->all();
+
+        return [
+            'promoted' => $this->buildGroup($game, $excludedUserIds, isPromoted: true),
+            'unpromoted' => $this->buildGroup($game, $excludedUserIds, isPromoted: false),
```
Jamiras, line 32: I realize this data is accessible via a query parameter in v1, but is it actually useful? Unpromoted achievements should not be earnable -  the only way is if they were earned and then demoted. And when that happens, we don't actually show the unlocks on the game or achievement pages.

Here's an example achievement that was earned by two users before it was demoted: http://localhost:64000/achievement/37546?tab=unlocks

wescopeland: [Hyrum's Law](https://www.hyrumslaw.com/)

 If we remove it, my sense is we won't have a clear V1 -> V2 migration path for the data. We can let consumers decide how useful it is.

Jamiras (approved):

PR #5066, opened 2026-07-11
feat(settings): switch to a tab-driven layout

wescopeland:
This PR revamps the page layout of the settings page such that it's now tab-driven. This is primarily done as preparation for upcoming OAuth (V2 API) UI/UX and improved notifications+subscriptions UI support.

Note: Third parties commonly point users to the settings page to grab their API key. This now lives under the "Applications" tab (whose naming will make much more sense when OAuth is hooked up there). I've opted _not_ to include any copy in the Profile (default) tab informing users where to look for their API key for three reasons:

1. The V2 Web API beta is imminent - we've almost reached functionality parity with V1.
2. OAuth should be the preferred authentication mechanism for all V2 API consumers, and we should go well out of our way to make API keys cumbersome to use.
3. The settings page shouldn't apologize for itself by telling users where to go to get the Web API key when this auth method will very soon fall out of favor.

If and when the V2 API supports webhooks for consuming various resources (ie: achievement unlocks, game awards), I will block API keys from being used for those endpoints altogether.

**Before**
Screenshot 2026-07-11 at 6 55 22 PM

**After**
Screenshot 2026-07-11 at 6 55 13 PM

Screenshot 2026-07-11 at 6 55 37 PM


resources/js/features/settings/components/SiteAwardsSection/SiteAwardsSection.tsx
```diff
@@ -2,20 +2,21 @@ import type { FC } from 'react';
...
 
-      <BaseButton size="sm" asChild>
-        <a href="/reorderSiteAwards.php">{t('Reorder Site Awards')}</a>
-      </BaseButton>
-    </div>
+        <BaseButton className="self-start" size="sm" asChild>
+          <a href="/reorderSiteAwards.php">{t('Reorder Site Awards')}</a>
+        </BaseButton>
+      </div>
+    </SectionStandardCard>
```
Jamiras: Do we event want to keep this section? There's a separate menu item next to Settings in the user menu for this.

resources/js/features/settings/components/AccountTabPanel/AccountTabPanel.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <div className="flex flex-col gap-4">
+      {!auth?.user.isMuted && auth?.user.isEmailVerified ? <ChangeUsernameSectionCard /> : null}
+
+      <ChangeEmailAddressSectionCard />
+      <ChangePasswordSectionCard />
+
+      <div className="flex flex-col gap-4 border-t border-neutral-700 pt-6 light:border-neutral-300">
+        <ResetGameProgressSectionCard />
```
Jamiras: I think we should remove this section too. It's much more intuitive for a user to reset a game from the game page and individual achievements from the individual achievement pages.

resources/js/features/settings/components/KeysSectionCard/KeysSectionCard.tsx
Jamiras, line 1: Should "Emulator Sessions" be given its own section?

Jamiras (approved):

resources/js/features/settings/components/SiteAwardsSection/SiteAwardsSection.tsx
wescopeland: Happy to remove this - done in latest.

resources/js/features/settings/components/AccountTabPanel/AccountTabPanel.tsx
wescopeland: Also happy to remove this one, done in latest.

resources/js/features/settings/components/KeysSectionCard/KeysSectionCard.tsx
wescopeland: I'm unsure. Gonna hang on to this as-is for now, primarily because we don't have a good way to discriminate those sessions (the section would effectively be one button).

PR #5067, opened 2026-07-12
fix(messages): use team inbox owner name for delegated inbox participant selection

angelol-git:
Fixes: https://discord.com/channels/310192285306454017/1525605055946821804

Delegated inboxes on mobile displayed the incorrect `otherParticipant.` 

**To Reproduce**

1. Sign in as a team member such as, `QualityAssurance`.
2. Create or use an existing message thread initiated by `QATeam` to another user.
3. Open the Messages inbox.
4. Change to the `QATeam` inbox.
5. View the inbox on mobile/card layout.

**Fix**
Use `senderUserDisplayName` to filter a valid  `otherParticipant` . As used in the desktop view `MessagesTableRow.tsx`.

**Desktop Reference**
desktop

**Before**
before

**After**
after


resources/js/features/messages/components/MessagesCardList/MessagesCard.tsx
```diff
@@ -13,13 +13,14 @@ interface MessagesCardProps {
...
 
 export const MessagesCard: FC<MessagesCardProps> = ({ messageThread }) => {
-  const { auth } = usePageProps();
+  const { auth, senderUserDisplayName } =
+    usePageProps<App.Community.Data.MessageThreadIndexPageProps>();
 
   const { t } = useTranslation();
 
   const otherParticipant =
     (messageThread.participants?.find(
```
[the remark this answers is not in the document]
angelol-git: I believe in production `display_name` will always be unique. 

In the case a user sends a message to themselves, the fallback to the owner would be correct.

resources/js/features/messages/components/MessagesCardList/MessagesCard.tsx
```diff
@@ -13,13 +13,14 @@ interface MessagesCardProps {
...
-      (p) => p.displayName !== auth?.user.displayName,
+      (p) => p.displayName !== senderUserDisplayName,
     ) as App.Data.User) ?? messageThread.participants?.[0];
```
wescopeland: DX: We're doing this calculation for "the other participant" with the identical `find + ?? foo.participants?.[0]` fallback in several places. 

Recommend extracting a small shared util, something like `getOtherParticipant(messageThread, senderUserDisplayName)` into the feature's utils folder.

wescopeland (commented):
Thanks - verified the fix, have one DX comment so we can prevent the bug from surfacing again.

resources/js/features/messages/components/MessagesCardList/MessagesCard.tsx
angelol-git: Yup good idea, updated in latest.

Jamiras (approved):
> Create or use an existing message thread between QATeam and another user.

I could only reproduce by creating a new thread from the team account. I assume threads created by a user are unaffected.

wescopeland (approved):

PR #5068, opened 2026-07-12
chore: overhaul dev setup and remove dead tooling

wescopeland:
This PR significantly overhauls the dev setup docs and removes quite a bit of vestigial tooling cruft that has built up over the years from V1-V3.

* The README has been simplified. **Sail is now the only documented setup path in the README.**
* XAMPP/Valet/hybrid setup instructions have been moved to _docs/guides/alternative-environments.md_.
* _.env.example_ has been reorganized into commented sections. ~35 dead environment variables have been removed.
* Removed all Pusher/websockets scaffolding.
* Removed the legacy `mysqli` singleton and the `ext-mysqli` dependency.
* Removed Rector. It's unused and its config was only current to PHP 8.0 and Laravel 9.
* Removed `worksome/envy`. It can't see docker-compose or Vite-only vars, so it fights our env file layout.
* Removed phpMyAdmin from _docker-compose.yml_. It's unconventional. Use any DB client via forwarded port 64010.
* Removed the PHP 8.2 and 8.3 dockerfiles, the unused _docker/pgsql_ folder, and the stale root-level `sail` script copy.
* Removed orphaned config files: _feed.php_, _translatable.php_, _blade-country-flags.php_.
* Added missing `php8.4-gmp` to the Sail image.
* oxfmt now also formats _README.md_ and _docs/**/*.md_.


config/broadcasting.php
```diff
@@ -30,33 +30,6 @@
```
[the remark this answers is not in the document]
wescopeland: Does this matter if there are no broadcasts in the app?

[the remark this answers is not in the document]
wescopeland: A simple grep shows no events implement this.

.env.example
```diff
@@ -55,136 +97,128 @@ OCTANE_HTTPS=false
...
-SCOUT_PREFIX=
-SCOUT_QUEUE=true
-MEILISEARCH_NO_ANALYTICS=true
-MEILISEARCH_HOST=http://meilisearch:7700
-MEILISEARCH_KEY=c0f9L4BrnVg9CoWXzy70olfWgrsct888
+# ------------------------------------------------------------------
+# Debugging and monitoring
+# ------------------------------------------------------------------
+
+#DEBUGBAR_ENABLED=false
```
Jamiras: I wonder if we should enable this by default. The example file isn't really used outside of local development.

.env.example
```diff
@@ -11,42 +14,81 @@ FORWARD_MINIO_CONSOLE_PORT=64041
...
+# ------------------------------------------------------------------
 # Application
+# ------------------------------------------------------------------
 
 APP_NAME=RetroAchievements
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
-APP_TIMEZONE=UTC
 #APP_VERSION=DEV
```
Jamiras: This should probably be uncommented too.
```
WARN[0000] The "APP_VERSION" variable is not set. Defaulting to a blank string. 
```

Or maybe `composer setup` should do these?

From the Ubuntu doc:
https://github.com/RetroAchievements/RAWeb/blame/e6b458d515a8ee0e6cb66938ccd5e4b8ec2d5318/docs/UBUNTU24_DESKTOP_VM_SETUP.md#L178-L182

.env.example
```diff
@@ -11,42 +14,81 @@ FORWARD_MINIO_CONSOLE_PORT=64041
...
+#REDIS_DB=0
+#REDIS_CACHE_DB=1
+#REDIS_QUEUE_DB=2
+
+# ------------------------------------------------------------------
+# Cache, queue, session, filesystem
+# ------------------------------------------------------------------
+
+#CACHE_DRIVER=redis
+#QUEUE_CONNECTION=database
```
Jamiras: This is the other thing I change every time I set up a new test environment:
https://github.com/RetroAchievements/RAWeb/blame/e6b458d515a8ee0e6cb66938ccd5e4b8ec2d5318/docs/UBUNTU24_DESKTOP_VM_SETUP.md#L178-L182

Jamiras (approved):

.env.example
wescopeland: This one surprises me and we should chat about.

I always use:
```
QUEUE_CONNECTION=redis
```

This seems like a better default if the README is going to direct users to sail. `sync` makes more sense with your documented setup.

In .env.example, I think we should change this:
```
#QUEUE_CONNECTION=database
#QUEUE_CONNECTION=redis
```
to this:
```
QUEUE_CONNECTION=redis
```

Any objections to this change?

.env.example
wescopeland: Uncommenting is probably best - done in latest.

.env.example
wescopeland: I think that's a good instinct - uncommented in latest.

.env.example
Jamiras: I thought `sync` was preferred for local environments as it surfaces job errors to the webpage.

I'm fine with having the default be `redis` as it better mimics production.

If this is the preferred direction, please remove the note about changing it from the ubuntu directions. Even though I'm running in a VM, I'm still using sail.

PR #5073, opened 2026-07-13
feat(api/UserResource): add lastGame relationship

wescopeland:
https://github.com/RetroAchievements/RAWeb/pull/5011#discussion_r3475169385

```
GET http://localhost:64000/api/v2/users/Scott?include=lastGame
```

This is expressed through a relationship - there is no top-level field on the user resource.

Screenshot 2026-07-12 at 7 59 40 PM


app/Api/V2/Users/UserResource.php
Jamiras, line 1: This should also be exposed for the `user/following` endpoint.
```
curl --header 'X-API-Key: ...' --header 'Accept: application/vnd.api+json' "localhost:64000/api/v2/users/Jamiras/following?include=user.lastGame"
```

wescopeland: Done in latest

Jamiras (approved):

PR #5074, opened 2026-07-13
feat(settings): add OAuth management UI

wescopeland:
Most of the diff is tests.

This PR builds on the settings tab-driven layout to flesh out the Applications tab, adding OAuth application management for users and app developers. App developers can register their apps and get their client ID + secret. Users can manage what apps they've connected to their account.

The functionality is behind a feature flag.

### 1. Enable the feature locally
```
sail artisan migrate
```
```
# in your .env
FEATURE_OAUTH=true
OAUTH_MAX_APPLICATIONS_PER_USER=5
```

### 2. Verify panel visibility

Navigate to http://localhost:64000/settings?tab=applications:
Screenshot 2026-07-13 at 7 44 47 PM

Set FEATURE_OAUTH=false - the "For Developers" panel should disappear.

Set FEATURE_OAUTH=true again.

### 3. Register an application

Click "Register application" and use these values:
* Name: Local QA Client
* Redirect URI: http://localhost:64000/auth/callback
* Public client: Not checked

Screenshot 2026-07-13 at 7 47 45 PM

On submit, the UI will show you some credentials. Copy these somewhere.

Screenshot 2026-07-13 at 7 48 22 PM

Once you've confirmed this dialog, it will close and you'll see your new application in the list:

Screenshot 2026-07-13 at 7 49 31 PM

### 4. As another user, authorize the application

In an incognito browser window, sign in as a different user. Then visit this URL (replace YOUR_CLIENT_ID with the client ID you copied):

http://localhost:64000/oauth/authorize?client_id=YOUR_CLIENT_ID&redirect_uri=http%3A%2F%2Flocalhost%3A64000%2Fauth%2Fcallback&response_type=code&scope=data%3Aread&state=qa

You'll land on this screen that we built in #4164:
Screenshot 2026-07-13 at 7 52 05 PM

Click the Authorize button. You should see some JSON dumped to the screen - this is expected (this is a demo route for local testing purposes). Copy the `code` value - it should be a very long string.

We need to exchange the code for access and refresh tokens. You can do this with a curl call:
```bash
curl -X POST 'http://localhost:64000/oauth/token' \
  -H 'Accept: application/json' \
  -d 'grant_type=authorization_code' \
  -d 'client_id=YOUR_CLIENT_ID' \
  -d 'client_secret=YOUR_CLIENT_SECRET' \
  -d 'redirect_uri=http://localhost:64000/auth/callback' \
  -d 'code=CODE'
```

You'll get a JSON response that contains an `access_token` value. Copy it.

## 5. Verify the connected application

Still as the 2nd user, go to http://localhost:64000/settings?tab=applications. You'll see a new card showing that Local QA Client is connected to their account:

Screenshot 2026-07-13 at 7 56 40 PM

You can also verify the access token works (note: it expires after 15 minutes -- real applications will refresh it using the refresh token):
```bash
curl 'http://localhost:64000/api/v2/systems' \
  -H 'Accept: application/vnd.api+json' \
  -H 'Authorization: Bearer YOUR_ACCESS_TOKEN'
```


wescopeland:
@greptileai Exhaustively review this PR with a security-oriented posture. I am specifically interested in any security issues you may find.

app/Community/Actions/CreateOAuthClientAction.php
```diff
@@ -0,0 +1,50 @@
...
+    public function execute(
+        User $user,
+        string $name,
+        array $redirectUris,
+        bool $isConfidential,
+        bool $enableDeviceFlow = false,
+    ): OAuthClient {
+        $maxApplications = max(1, (int) config('oauth.max_applications_per_user'));
+
+        if (OAuthClient::query()->active()->ownedBy($user)->count() >= $maxApplications) {
```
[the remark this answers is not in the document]
wescopeland: Declined - this race condition exists, but the attacker and victim are both the same person lol

Limiting this wouldn't really enforce a legitimate production security boundary. It's high complexity for near-zero benefit.

wescopeland:
@greptileai Once again, exhaustively review this PR with a security-oriented adversarial posture. I am specifically interested in any security issues you may find.

If you find no security issues, respond to this message stating so.

resources/js/features/settings/components/OAuthApplicationManagementDialogTrigger/OAuthApplicationManagementDialogTrigger.tsx
```diff
@@ -0,0 +1,142 @@
...
+        </BaseDialogHeader>
+
+        <OAuthApplicationForm application={application} onUpdated={() => setIsOpen(false)} />
+
+        {application.confidential ? (
+          <div className="flex flex-col gap-3 border-t border-neutral-700 pt-4 light:border-neutral-200">
+            <div>
+              <h3 className="font-semibold">{t('Client secret')}</h3>
+              <p className="text-sm text-neutral-400">
+                {t('Rotating the secret immediately stops the old secret from working.')}
```
Jamiras: I believe "rotating" implies having the old token still available for some period of time before it's deactivated. Maybe this should say "Regenerating the secret..."

app/Policies/OAuthClientPolicy.php
Jamiras, line 1: I'm very hesitant to have this exposed for every user. Without some sort of approval process (and no way currently to view all registered clients), literally anything could use our credentials for their website.

I can't think of a reasonable way to restrict it though. Many integrations create a unique user for the integration so we can't rely on the user having a minimum number of points. Maybe we could rely on a minimum account age.

This is probably a minor issue as long as the application remains read only. But it might be a design decision we regret in the future.

For public data, the caller should use an API token. The OAuth logic is really only necessary for those APIs restricted to the current user.

app/Enums/OAuthScope.php
Jamiras, line 1: We probably want this to be more granular - if nothing else, for future expansion.

`profile.read` -> `View information about your RetroAchievements user`
`player_activity.read` -> `View information about games you've played and achievements you've unlocked`
`followers.read` -> `View information about users you are following`

etc.

app/Enums/OAuthScope.php
```diff
@@ -0,0 +1,21 @@
...
+    case Read = 'data:read';
+
+    /**
+     * Used by Passport::tokensCan() to document each scope.
+     * These strings are not exposed to the UI.
+     */
+    public function description(): string
+    {
+        return match ($this) {
+            self::Read => 'View your RetroAchievements data',
```
Jamiras: Because this effectively opens up the entire V2 API, it's not just granting access to _your_ data.

Jamiras: Things that aren't user specific, like the list of systems or recent players for a game probably shouldn't be gated behind an oauth scope and therefore wouldn't need a specific permission prompt. 

For uses like that, are we expecting clients to use the `X-API-Key` interface and not require their users to OAuth?

Not providing a scope on the URL does still show the request for "your RetroAchievements data". Passing an unknown scope generates an error. Does the system default to this scope if one is not provided?

resources/js/features/settings/components/OAuthApplicationManagementDialogTrigger/OAuthApplicationForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+          render={({ field }) => (
+            <BaseFormItem className="flex flex-col gap-1">
+              <BaseFormLabel>{t('Redirect URI')}</BaseFormLabel>
+              <BaseFormControl>
+                <BaseInput
+                  placeholder={t('https://example.com/oauth/callback', {
+                    nsSeparator: null,
+                  })}
+                  required
+                  {...field}
```
Jamiras, line 60: Why is this field required if all the URLs pass it in?
> localhost:64000/oauth/authorize?client_id=YOUR_CLIENT_ID&**redirect_uri=http%3A%2F%2Flocalhost%3A64000%2Fauth%2Fcallback**&response_type=code&scope=data%3Aread&state=qa

```
curl -X POST 'http://localhost:64000/oauth/token' \
  ...
  -d 'redirect_uri=http://localhost:64000/auth/callback' \
  ...
```

resources/js/features/settings/components/OAuthApplicationManagementDialogTrigger/OAuthApplicationManagementDialogTrigger.tsx
wescopeland: I thought "rotate" meant the secret was regenerated instantly, but I was wrong. I've changed everywhere affected in latest to use the "regenerate" terminology instead.

app/Policies/OAuthClientPolicy.php
wescopeland: > Maybe we could rely on a minimum account age.

Agreed, I've implemented this in latest.

I considered several of the points in your message when planning this branch. There are numerous mitigations built into this branch's code:

1. We enforce an OAuth scope on all grants of `data:read`. This scope system caps what the "anything" you're describing can do. An OAuth token can only read data the user could already see, so a malicious app can't post, modify a profile, or take any action on the user's behalf. When we add more scopes later (ie: writes, etc), existing grants won't silently inherit them. Every app will have to send the user back through the consent screen, and the user will see exactly what new power is being requested by that app.
2. Tokens expire in 15 days, and refresh tokens expire in 30 days. This bounds the lifetime of any credential the user hands over. If a token leaks, or if a user forgets they authorized some app, the access doesn't persist indefinitely on its own like our API keys do. And the user has the ability to revoke the grant on their own at any time. Stale or abandoned grants decay.
3. There's a per-user quota of 5 applications. This limits the scale of abuse per account.
4. We have rate limiting on the token management and consent approval endpoints as another layer of defense. This won't stop a patient human, but it puts a speedbump on automation-driven abuse.
5. We have an `ra:community:oauth:deactivate-client` Artisan command that kills a given client. This command terminates the client, severs every user's grant, and invalidates every access+refresh token it was ever issued, for all users, with no way for that app to reauthorize. We may not be able to pre-screen, but we can execute a total irreversible takedown the moment we notice something bad is happening.

> For public data, the caller should use an API token.

While I agree that purely public data doesn't need user auth at all, API keys are single full-access never-expiring forever credentials, and every third-party app users paste it into today is a liability this system is designed to replace. For reasons mentioned above, I believe we should try to strongly dissuade people from using API keys once this system rolls out.

> Without some sort of approval process (and no way currently to view all registered clients), literally anything could use our credentials for their website.

I agree we need a Filament resource of OAuth clients. Will add this to my TODO list.

app/Enums/OAuthScope.php
wescopeland: After giving this considerable thought, I believe the V2 API's data breaks into three tiers:
1. Global stuff that isn't user-specific, like systems, games, hashes, achievements, sets/versions, events, hubs, leaderboards, claims, etc.
2. User data that's public, like profiles, player games, player achievements, awards, and recent games.
3. Genuinely caller-private stuff, like followers / following and Want to Play lists.

My sense is the neatest solution at the moment is:
* Narrow `data:read` to "publicly visible RA data" (tier 1 and tier 2).
* Reword the consent copy to "View publicly visible RetroAchievements data".

I've implemented these changes in latest. The tier 3 endpoints have a naive `DenyOAuthTokens` middleware applied to them which blocks their access for the time being.

Adding more scopes later for access to the tier 3 data is safe. Apps are not automatically granted new scopes. Apps have to request consent, and users must provide that consent and will see exactly what it is they're forking over.

`data:read` is the default via `Passport::defaultScopes()`, which is why an authorize URL without a scope still shows the prompt.

Ideally, clients do not continue using `X-API-Key` over the long-term. I believe OAuth is the more desirable path. Maybe we should even make that Tier 3 data _require_ OAuth.

resources/js/features/settings/components/OAuthApplicationManagementDialogTrigger/OAuthApplicationForm.tsx
wescopeland: The registered URI is an allowlist. When the app passes `redirect_uri` in a request, the server checks it against that allowlist and rejects anything that doesn't match. Without it, anyone could grab your `client_id` (it's public) and swap in their own `redirect_uri` to steal auth codes. Every client must have one registered.

app/Enums/OAuthScope.php
wescopeland: See: https://github.com/RetroAchievements/RAWeb/pull/5074#discussion_r3606911027

app/Enums/OAuthScope.php
Jamiras: Despite the fact that it is public data, I feel uneasy grouping public user data together with general site data.

I imagine most users want to know exactly what parts of their user data is needed by an integrating application - even if it is public data.

resources/js/features/settings/components/OAuthApplicationManagementDialogTrigger/OAuthApplicationForm.tsx
Jamiras: That makes sense, but then why not just always use the one in the database? Isn't the URL field redundant? Or is it intentionally there to server as a secondary validation?

app/Policies/OAuthClientPolicy.php
Jamiras: > While I agree that purely public data doesn't need user auth at all, API keys are single full-access never-expiring forever credentials, and every third-party app users paste it into today is a liability this system is designed to replace. 

It just feels weird that a client would have to ask the user to log in just to view purely public information.

I guess it makes sense in the context of rate limiting where each client should not be impacted by the overall use of the client.

Jamiras (approved):
.

app/Api/RouteServiceProvider.php
```diff
@@ -194,14 +198,14 @@ private function apiRoutes(): void
...
-                                    $relationships->hasMany('following')->readOnly();
+                                    $relationships->hasMany('followers')->readOnly()->middleware(DenyOAuthTokens::class);
+                                    $relationships->hasMany('following')->readOnly()->middleware(DenyOAuthTokens::class);
                                     $relationships->hasMany('leaderboardEntries')->readOnly();
                                     $relationships->hasMany('playerAchievements')->readOnly();
                                     $relationships->hasMany('playerAchievementSets')->readOnly();
                                     $relationships->hasMany('playerGames')->readOnly();
                                     $relationships->hasMany('tickets')->readOnly();
-                                    $relationships->hasMany('userGameListEntries')->readOnly();
+                                    $relationships->hasMany('userGameListEntries')->readOnly()->middleware(DenyOAuthTokens::class);
```
Jamiras, line 212: Just a reminder: We'll want to document which scopes each API requires in the documentation. I'm not sure if there's an easy way to annotate things such that it would be automatic.

wescopeland: Big agree. See also: https://github.com/RetroAchievements/api-docs/pull/100#issuecomment-5023387264

I'm going to explore how we can emit an [OpenAPI schema](https://www.reddit.com/r/eli5_programming/comments/188ycwc/eli5_whats_openapi_schema/) automatically. This will land as a PR in this repo.

resources/js/features/settings/components/OAuthApplicationManagementDialogTrigger/OAuthApplicationForm.tsx
wescopeland: It's intentional, but it isn't revalidation. It's more subtle and aligns us with the OAuth spec:

1. [A client can register more than one redirect URI.](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.3)

> If multiple redirection URIs have been registered ... the client MUST include a redirection URI with the authorization request using the "redirect_uri" request parameter.

2. [It's a binding across two hops of the OAuth flow rather than something that serves as re-validation.](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3)

> redirect_uri
> REQUIRED, if the "redirect_uri" parameter was included in the authorization request as described in Section 4.1.1, and their values MUST be identical.

See also: [the attack this query param exists to stop](https://datatracker.ietf.org/doc/html/rfc6749#section-10.6).

This is implemented by Passport itself. Off-the-shelf OAuth packages send `redirect_uri` unconditionally.

PR #5082, opened 2026-07-18
feat(user): add support for another Patreon tier

wescopeland:
This PR adds support for another Patreon tier. In the user moderation panel, there are no two buttons for the Patreon Supporter row:

Screenshot 2026-07-18 at 1 34 23 PM

When someone is given the $1 award, things mostly behave as they do today, with one minor change. The tooltip for the badge is no longer a native tooltip:

Screenshot 2026-07-18 at 1 41 51 PM

When someone is given the $2 award, several things change. First, they get a holo-badge:

https://github.com/user-attachments/assets/56e95087-d39e-4b23-be97-573cef08ffc6

They also get a "Supporter" role to show off on their user profile, card, and forum posts:
Screenshot 2026-07-18 at 1 38 22 PM


database/migrations/2026_07_18_000000_create_supporter_role.php
```diff
@@ -0,0 +1,46 @@
...
+        ]);
+    }
+
+    public function down(): void
+    {
+        $role = DB::table('auth_roles')->where('name', Role::SUPPORTER)->first();
+
+        if ($role) {
+            DB::table('users')
+                ->where('visible_role_id', $role->id)
```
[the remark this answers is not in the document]
wescopeland: Not worth stressing about.

app/Helpers/render/site-award.php
```diff
@@ -338,23 +339,39 @@ function RenderAward(
...
+        $description = $isSupporterTier
+            ? '$2 Patreon supporter. Thank you so much for your support!'
+            : 'Thank you so much for your support!';
+
+        echo avatar('patreonSupporterAward', $awardData,
+            link: route('patreon-supporter.index'),
+            tooltip: "<div class='p-2 max-w-[320px] text-pretty text-menu-link flex flex-col gap-1'><p class='font-bold'>Patreon Supporter</p><span>{$description}</span><p class='italic'>Supporting RA since {$awardDate}</p></div>",
+            iconUrl: asset('/assets/images/badge/patreon.png'),
+            iconSize: $imageSize,
+            iconClass: $isSupporterTier ? 'holoimage' : 'goldimage',
```
Jamiras, line 364: I'm hesitant to introduce a "holoframe" for this.

1) It's something new and shiny that people are going to want everywhere. The most likely place I could see it being used is for the topmost tier of tiered events.

or

2) People have scripts to disable the gold borders and either this won't appear, or it will stand out even moreso.

and

3) What happens when we inevitably add another tier a few years from now?

Can't we just get a new image like we do for other badge tiers?

app/Helpers/render/site-award.php
```diff
@@ -338,23 +339,39 @@ function RenderAward(
...
+        // derived from whether the user holds the publicly-visible Supporter role.
+        $ownerUser = User::whereName($ownerUsername)->first();
+        $isSupporterTier = $ownerUser && $ownerUser->hasRole(Role::SUPPORTER);
+
+        $description = $isSupporterTier
+            ? '$2 Patreon supporter. Thank you so much for your support!'
+            : 'Thank you so much for your support!';
+
+        echo avatar('patreonSupporterAward', $awardData,
+            link: route('patreon-supporter.index'),
```
Jamiras, line 360: The supporter list doesn't highlight the second tier of supporters. Should it?

public/request/user/update.php
```diff
@@ -88,14 +89,54 @@
...
+    Validator::validate(['value' => $value], [
+        'value' => ['required', 'integer', Rule::in([1, 2])],
+    ]);
+
+    $comment = DB::transaction(function () use ($foundTargetUser, $value) {
+        User::whereKey($foundTargetUser->id)->lockForUpdate()->first();
+
+        $hasBadge = HasPatreonBadge($foundTargetUser);
+        $hasSupporterRole = $foundTargetUser->hasRole(Role::SUPPORTER);
+        $currentTier = $hasBadge ? ($hasSupporterRole ? 2 : 1) : 0;
```
Jamiras: Relying on the role to differentiate the tiers doesn't allow for future tiers. Recommend just using the badge tier field.

app/Helpers/render/site-award.php
wescopeland: 1. That's a feature, not a bug. For the time being we should keep it only to this and reject any requests to put it elsewhere until this has had some time in the oven.

2. I'm not worried about how this impacts people using userscripts.

3. We could add juice elsewhere besides the badge if this happens.

app/Helpers/render/site-award.php
wescopeland: I wasn't sure whether to include it or not, but it's a thin enough change I've made this tweak in latest:

Screenshot 2026-07-22 at 6 56 51 PM

public/request/user/update.php
wescopeland: Done in latest.

app/Helpers/render/site-award.php
Jamiras: I was imagining a tiered list. Newest could remain unchanged, then there would a $2 tier list, then a $1 tier list. This would make bigger donators more prominent. I know I've seen something like that somewhere.

app/Helpers/render/site-award.php
Jamiras: I still have reservations about this, but I'm willing to see how it goes.

app/Helpers/render/site-award.php
wescopeland: Ah, that makes sense. Done in latest:
Screenshot 2026-07-23 at 5 20 11 PM

public/request/user/update.php
```diff
@@ -88,14 +89,38 @@
...
+    Validator::validate(['value' => $value], [
+        'value' => ['required', 'integer', Rule::in([1, 2])],
+    ]);
+
+    $comment = DB::transaction(function () use ($foundTargetUser, $value) {
+        User::whereKey($foundTargetUser->id)->lockForUpdate()->first();
+
+        $currentTier = GetPatreonSupporterTier($foundTargetUser);
+
+        if ($value === $currentTier) {
```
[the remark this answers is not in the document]
wescopeland: You already gave this feedback in https://github.com/RetroAchievements/RAWeb/pull/5082#discussion_r3608860694

I thought we fixed this with a DB transaction?

Jamiras (approved):

PR #5085, opened 2026-07-19
feat(api/v2): add maxRank filter for leaderboard entries

wescopeland:
This PR adds a `filter[maxRank]` to V2's `LeaderboardEntry` resource:

```
GET /api/v2/users/{user}/leaderboard-entries?filter[maxRank]=10
GET /api/v2/leaderboards/{leaderboard}/entries?filter[maxRank]=10
```


Jamiras (approved):
Makes sense for `user/x/leaderboard_entries`, but for `leaderboard/x/entries`, it's basically redundant with `page[size]` unless the stopping point is a shared rank.

PR #5087, opened 2026-07-21
perf: add a materialized view for player ranks

wescopeland:
```
sail artisan migrate

sail artisan horizon # restart
sail artisan ra:platform:player:update-global-rankings
```

This PR adds a `player_global_rankings` materialized view holding every ranked user's aggregates and rank for the daily, weekly, and all-time ranking windows in both hardcore and casual modes.

Until now, site rank has been computed on demand in `getUserRank()`. This helper function ran a `COUNT(*)` over the entire `users` table and cached the result in Redis for 15 minutes. That cache only helps with repeat reads from the same user. Every cold lookup pays the full cost of the `COUNT(*)` query. Meanwhile, the Global Rankings page has very little caching and is mostly making hot DB reads that cause tremendous strain on DB server CPU, especially when crawlers blindly iterate through every page in the table.

Now, a scheduled job rebuilds the table every 15 minutes using the same pattern as the global beaten games leaderboard.

Everywhere rank is exposed (user profiles, cards, the V1 API, the Global Ranking page) now read from the view. The Redis rank cache has been removed entirely.

User rank is now also exposed on the V2 API:
```
GET /api/v2/users/{user}
```
Screenshot 2026-07-21 at 6 35 23 PM

Additionally, `/api/v2/users` now includes some meta with counts for the number of ranked users:
Screenshot 2026-07-21 at 6 36 07 PM

A few other notes:
- Soft deleted users no longer count toward `TotalRanked`.
- The date filter on the Global Rankings page (which seems to actually be broken) has been removed altogether.


app/Api/V2/Controllers/UserController.php
```diff
@@ -140,6 +157,23 @@ private function stashReciprocalUserIds(Collection $ids): void
...
+     * @return array{rankedUsers: array{hardcore: int, casual: int}}
+     */
+    private function rankedUserMeta(): array
+    {
+        $counts = PlayerGlobalRankingTotal::query()
+            ->whereIn('rank_type', [RankType::Hardcore, RankType::Casual])
+            ->pluck('total', 'rank_type');
+
+        return [
+            'rankedUsers' => [
```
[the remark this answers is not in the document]
wescopeland: Declining. This query is about as cheap as MariaDB queries get. It's a `whereIn` on a 3 row table. I don't believe Redis w/ `Cache::remember()` would be meaningfully faster.

database/migrations/2026_07_20_000000_create_player_global_rankings_tables.php
```diff
@@ -0,0 +1,49 @@
...
+            $table->index(['window', 'mode', 'points_weighted'], 'player_global_rankings_weighted_index'); // efficient sorting
+            $table->index(['window', 'mode', 'awards_count'], 'player_global_rankings_awards_index'); // efficient sorting
+
+        });
+
+        // stores precomputed all-time population counts for each rank type.
+        // this lets us do a constant-time lookup for counts instead of repeatedly
+        // counting the number of ranking rows.
+        Schema::create('player_global_ranking_totals', function (Blueprint $table) {
+            $table->unsignedTinyInteger('rank_type')->primary(); // 1 = hardcore, 2 = casual, 3 = weighted (hardcore)
```
Jamiras: Why is this an enum instead of a string? I thought we were moving away from integer constants in the DB.

I see it's mapped from an existing enum. As best as I can tell, you're touching all the code using that enum, so it would be a good time to change it.

Jamiras (approved):

database/migrations/2026_07_20_000000_create_player_global_rankings_tables.php
wescopeland: Addressed in latest

PR #5111, opened 2026-07-24
chore: audit our gh workflows, add a labeler workflow

wescopeland:


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,33 @@
...
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  label:
+    runs-on: ubuntu-22.04
```
Chew: Something to consider, for jobs like this, there is a [ubuntu-slim](https://github.blog/changelog/2026-01-22-1-vcpu-linux-runner-now-generally-available-in-github-actions/) optimized for quick tasks like these.

wescopeland: Thanks for the suggestion. Several jobs needed to have their Ubuntu versions changed to either 24 or slim (which is derived from 24). I've updated 9 of these in the latest commit.

.github/workflows/ci.yml
Jamiras, line 1: Curious why this was necessary, and why not go to `ubuntu-26.04` or just `ubuntu-latest`?

Jamiras (approved):

.github/workflows/ci.yml
wescopeland: Most of the changes in this file are from Greptile's flagged security issue here: https://github.com/RetroAchievements/RAWeb/pull/5111#discussion_r3645402731

For the ubuntu version, I have no preference. I didn't realize Ubuntu 26 was already in LTS.

PR #5117, opened 2026-07-27
feat(settings): add confirm dialog for username change requests

### LINKED ISSUE #5115: Add another layer of confirmation for changing a username
### User Story

As a user changing my username,
I want to explicitly confirm the exact name I'm about to change to,
so that I can't accidentally commit to a username that I never meant to type.

### Why it matters

Several users recently tried to change their password and didn't notice they had filled out their new password into the Change Username fields. Their password then sat publicly exposed as their username on the site and on third-party sites that mirror our data, and username changes are rate-limited to one request per 30 days, so they couldn't fix the issue. A user with root DB access had to intervene manually.

We do show a confirmation today, but it's a native `window.confirm()` that only warns about the 30-day cooldown. It never shows the user the name they're about to request, which is exactly the info that would have caught this before it happened. Browsers also visually deemphasize native confirms, and use click through them reflexively.

### Acceptance criteria

- Submitting the Change Username form opens a proper dialog instead of a `window.confirm()`.
- The dialog displays the requested name verbatim and requires checking an acknowledgement like "I confirm I want to change my username to [new username]" before the confirm button becomes enabled.
- Cancelling or dismissing the dialog submits nothing and preserves the form state.
- The 30-day cooldown warning remains part of the dialog copy.
- Case-only changes keep their current behavior of applying instantly without review, but still go through the confirmation dialog.
- The acknowledgement checkbox resets every time the dialog opens.
- The username fields block browser password autofill as a second layer of defense.

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5115.

Also sprinkles in some DOM attributes to try to keep these fields out of password managers.

Screenshot 2026-07-27 at 4 55 21 PM


Jamiras (approved):

resources/js/features/settings/components/ChangeUsernameSectionCard/ChangeUsernameConfirmDialog.tsx
Jamiras, line 1: When errors occur, the dialog is not closed. Is this intentional?

Image

wescopeland: Have an idea to improve this UX. Going to address this a different way in an immediate follow-up.

PR #5118, opened 2026-07-28
bugfix(metrics): improve playtime calculations when sessions overlap

Jamiras:
Fixes https://discord.com/channels/310192285306454017/1511806236050460733/1511832640683245768

Should also fix https://discord.com/channels/310192285306454017/1530794157260406834/1530794157260406834 and https://discord.com/channels/310192285306454017/1519494550060662814

NOTE: existing data has to be regenerated. I came up with the following queries to hit most of the affected data, but they're very slow.

Direct strike (only update the records that appear to be affected):
```
select concat('sail artisan ra:platform:player:update-game-metrics ', user_id, ' ', game_id)
  from player_games where time_to_beat > playtime_total;
```
This generates over 8000 affected user/game combinations.

Broader strike (update all records for games where at least one user is affected):
```
select concat('sail artisan ra:platform:game:update-player-games ', sub.game_id)
  from (select distinct game_id from player_games where time_to_beat > playtime_total) as sub;
```
This generates ~1500 affected games, most of which have thousands of users, but ensures the same logic is applied to everyone.

A third option is to update all records for any game with subsets, though there are several users who have managed to create overlapping sessions without subsets.

It might even be worth running against all games with published achievements (~12000).

Additional note: some records will still be returned by the query after updating the data. As best as I can tell, these fall into two categories: 1) short games where rounding causes discrepancies when comparing seconds to minutes and 2) games with generated sessions (unlocks prior to 2023).


wescopeland (approved):
Let's go with the broader strike.

Instructions for future-me:

---

On deploy, start Horizon on the app server, then in a tinker session connected to prod run:
```php
use App\Models\PlayerGame;
use App\Platform\Jobs\UpdatePlayerGameMetricsChunkJob;

$dryRun = true;

$chunkSize = 250;

$affected = PlayerGame::query()
    ->whereColumn('time_to_beat', '>', 'playtime_total')
    ->get(['user_id', 'game_id']);

$byGame = $affected->groupBy('game_id');

echo $affected->count() . " affected player_games across " . $byGame->count() . " games\n";

$jobCount = 0;
foreach ($byGame as $gameId => $rows) {
    foreach ($rows->pluck('user_id')->unique()->chunk($chunkSize) as $userIds) {
        $jobCount++;

        if (!$dryRun) {
            UpdatePlayerGameMetricsChunkJob::dispatch((int) $gameId, null, $userIds->values()->all())
                ->onQueue('player-game-metrics-batch');
        }
    }
}

echo ($dryRun ? 'Would dispatch ' : 'Dispatched ') . $jobCount . " chunk jobs\n";
```
Assuming the echo looks okay, flip `$dryRun` and execute.

Verified locally:
Image

After queues are flushed, run:
```php
use App\Models\GameAchievementSet;
use App\Platform\Jobs\UpdateGameBeatenMetricsJob;

$gameIds = Game::query()
  ->whereHasPublishedAchievements()
  ->pluck("id");

foreach ($gameIds as $gameId) {
  UpdateGameBeatenMetricsJob::dispatch($gameId)->onQueue('game-beaten-metrics');
}
```

Jamiras:
FYI, that's still the direct strike. You're only updating the records where the discrepancy exists, but you're doing it in a more efficient manner by grouping them by game and only doing the game recalculation once.

wescopeland:
Yep, was tired and got my wires crossed.

> It might even be worth running against all games with published achievements (~12000).

Looking at this with a fresh pair of eyes, I think we should do this too. I've appended this script to my message above:
```php
use App\Models\GameAchievementSet;
use App\Platform\Jobs\UpdateGameBeatenMetricsJob;

$gameIds = Game::query()
  ->whereHasPublishedAchievements()
  ->pluck("id");

foreach ($gameIds as $gameId) {
  UpdateGameBeatenMetricsJob::dispatch($gameId)->onQueue('game-beaten-metrics');
}
```

PR #5119, opened 2026-07-28
feat(manage/games): cap the number of attached genre hubs

### LINKED ISSUE #5099: Cap games at three genres
### User Story

As a user who curates hubs,
I want a game to carry at most three genre hubs,
so that game metadata stays readable and people stop enumerating every applicable subgenre instead of using the appropriate parent hub.

### Why it matters

A game was recently found that lists these genres:
* Extreme Sports - Motocross
* Extreme Sports - Snowboarding
* Kart Racing
* Sports
* Sports - Basketball
* Sports - Football | Soccer
* Sports - Volleyball

The metadata row is a reader-facing summary. Seven entries does not summarize anything, and they dilute the hubs themselves. If every sports compilation is attached to every individual sport, the individual sport hubs stop meaning "this game is a basketball game".

### Acceptance criteria

* Attaching a genre hub to a game that already has three is rejected, with a message explaining the cap.
* The cap is enforced from both the game side and the hub side, including the CSV bulk add path on the game's Hubs page and any bulk attach from a hub's games list.
* Games already exceeding the cap are not broken. Adding another is blocked.
* Genre and subgenre hubs are both tallied in the cap count.

### Open questions

* Should any role be able to override the cap for genuine edge cases, or is it absolute?

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5099.

Games are now capped to a maximum of 3 genre hub attachments, using a similar mechanism that we recently used to cap the count of Similar Games. Genre and subgenre hubs both count towards the tally.

Screenshot 2026-07-27 at 10 27 42 PM


app/Models/GameSet.php
```diff
@@ -284,6 +284,11 @@ public static function boot()
     public const DeveloperEventsHubId = 5;
     public const FreePointsHubId = 3796;
 
+    /**
+     * @var string[]
+     */
+    public const GenreHubTitlePrefixes = ['[Genre - ', '[Subgenre - '];
```
Jamiras, line 290: Are there any plans to clean up the games that already have more than 3?
```
select game_id, count(*) from game_set_games where game_set_id in (select id from game_sets where title like '[Genre -%' or title like '[Subgenre - %') group by 1 order by 2 desc limit 20;
```

Jamiras (approved):

app/Models/GameSet.php
wescopeland: We should provide a list to the site cleanup team (I did something similar for the Similar Games stuff), but I don't think we should do this ourselves.

Haven't run this on prod - on my local I only see 7 games affected.

Jamiras: That's what I had on my local too.

PR #5127, opened 2026-07-30
fix: add more deadlock resiliency

wescopeland:
We're still seeing quite a few deadlocks in prod largely due to how much write contention there is on the `users` table.

This PR attempts a cheap fix we've used elsewhere of sorted chunking + retries for the hot paths that are throwing server errors.

If this still doesn't work, a more expensive fix will be to start breaking out some of the denormalized columns on `users` into their own tables. This is less than ideal and a lot of work over numerous deploys, so hopefully this change works.


app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
Jamiras, line 1: These are only touching a single row, and aren't doing any aggregation. How does wrapping it in a transaction help?

Jamiras (approved):

app/Platform/Actions/IncrementDeveloperContributionYieldAction.php
wescopeland: These writes are deadlocking in production. They aren't the cause of the deadlock, but they are indeed collateral damage from something further upstream. InnoDB automagically picks a deadlock victim by what it believes is doing the least amount of work, so the write that only touches a single row is what it's deciding to kill.

Laravel only supports deadlock retries in `DB::transaction()`. It isn't possible to retry without a transaction (short of retrying the entire job).

In latest I've fixed the stale comment, and I've wrapped the same pattern in `UpdatePlayerMetricsAction`, which is showing up in the same production deadlock log.

PR #5128, opened 2026-07-31
fix(ui): move displayed badge dialog back to center

Chew:
Small regression from the Tailwind 4 update. Adds some classes to force it to the center of the screen like before.

Before:
CleanShot 2026-07-30 at 19 45 38@2x

After:
CleanShot 2026-07-30 at 19 45 13@2x


wescopeland (approved):
Thanks!

PR #5130, opened 2026-07-31
fix(comments): keep unsubmitted comment drafts

mariobob:
Resolves https://github.com/RetroAchievements/RAWeb/issues/4737

Writing a comment on an achievement page and then switching to Recent Unlocks or Changelog threw the comment away. The tabs are Radix `Tabs.Content`, which unmounts whatever isn't active, and the comment form didn't persist anything.

Comment drafts now go into `sessionStorage` through the existing `useFormDraft` hook, keyed by the commentable and its author, so a draft survives a tab switch and is shared with the wall's "all comments" page without carrying into anyone else's session. Submitting clears it.

Two `useFormDraft` fixes were needed first: flush the debounced write when the form unmounts, and rehydrate when the draft key changes. Both live in the hook, so every caller benefits.

comment-draft-fix

## AI disclosure
Claude Code was used for this change: it wrote the hook changes and the tests. I reviewed every line, verified the tests fail without the fix, and manually reproduced the issue and the fix.


mariobob:
@greptile-apps Both fixed.

Worth noting the same account-boundary issue applies to the existing forum/message draft keys from https://github.com/RetroAchievements/RAWeb/pull/4667, which aren't scoped to a user either. I've left those alone to keep this PR focused.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
```diff
@@ -35,11 +39,16 @@ export function useSubmitCommentForm({
   });
   type FormValues = z.infer<typeof addCommentFormSchema>;
 
+  const draftKey = `comment-${commentableType}-${commentableId}-${auth?.user.id ?? 'guest'}`;
```
mariobob: Keyed on the commentable so the wall's tab and its "all comments" page share a draft. Scoped to the author so a draft can't survive into another user's session in the same tab.

resources/js/common/components/CommentList/CommentList.test.tsx
```diff
@@ -3,12 +3,16 @@ import userEvent from '@testing-library/user-event';
...
-import { render, screen } from '@/test';
+import { render, screen, waitFor } from '@/test';
 import { createComment, createUser } from '@/test/factories';
 
 import { CommentList } from './CommentList';
 
 describe('Component: CommentList', () => {
+  beforeEach(() => {
+    sessionStorage.clear();
+  });
```
mariobob, line 14: Prevents leaking comment drafts between tests.

resources/js/features/achievements/components/+show/AchievementShowRoot.test.tsx
```diff
@@ -19,6 +20,8 @@ describe('Component: AchievementShowRoot', () => {
   beforeEach(() => {
     vi.clearAllMocks();
 
+    sessionStorage.clear();
```
mariobob, line 23: Prevents leaking comment drafts between tests.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
```diff
@@ -55,8 +64,10 @@ export function useSubmitCommentForm({
       {
         loading: t('Submitting...'),
         success: () => {
+          clearDraft();
           onSubmitSuccess?.();
-          form.reset();
+
+          form.reset(draftDefaultValues);
```
mariobob: The form's defaultValues hold the restored draft, so on submission the draft needs to be reset explicitly.

wescopeland:
Hi! Thanks for opening this PR into RAWeb. I'll review it later, but before I do, I have one request.

Please write a more concise PR description. Claude has a tendency to output long prose that is easy to gloss over. https://github.com/ayghri/i-have-adhd and this prompt may be useful in this regard:
```md
## Writing Style: ASD-STE100 Simplified Technical English

ASD-STE100 Simplified Technical English is a controlled writing standard. Aerospace and defense groups made it. It helps people write clear technical text.

Key rules:
- Use approved words only. The standard gives a word list. Each word has one meaning.
- Use one word for one idea. Do not use two words for the same thing.
- Write short sentences. Use 20 words or less for instructions.
- Use active voice. Write "Turn the switch", not "The switch must be turned".
- Write short paragraphs. Keep one topic in each paragraph.

The goal is easy reading. Many readers are not native English speakers. Clear text helps them do the work in a safe and correct way.
```

resources/js/common/hooks/useFormDraft.ts
```diff
@@ -1,34 +1,84 @@
...
+    return () => {
+      clearTimeout(timer);
+    };
   }, [key, values]);
 
+  useEffect(() => {
+    return () => {
+      pendingWriteRef.current?.();
+    };
+  }, []);
```
mariobob: Unmounting cancels the debounce, so this writes out whatever it was still holding.

resources/js/common/hooks/useFormDraft.ts
```diff
@@ -1,34 +1,84 @@
...
+    previousKeyRef.current = key;
+
+    if (!key || key === previousKey) {
+      return;
     }
 
-    return () => {
-      if (timerRef.current) {
-        clearTimeout(timerRef.current);
+    pendingWriteRef.current?.();
```
mariobob: This flushes the old draft, then rehydrates by hand (useForm() only reads defaultValues on mount).

mariobob:
Hi @wescopeland! Fair enough, I collapsed the detail behind `<details>` so it's short at a glance and reads better now.

For what it's worth, the length was deliberate rather than generated filler. This codebase defaults to no inline comments, so I moved the reasoning into the description instead of losing it (I'd rather it live somewhere than nowhere).

mariobob (commented):
Posting the detail here rather than in the description, since a squash merge etches the description into the commit message permanently. Comments stay on GitHub where this kind of thing belongs.

<details>
<summary><b>Two <code>useFormDraft</code> bugs had to be fixed first</b></summary>

Otherwise this wouldn't have worked at all:

- **The debounce was lost on unmount.** The cleanup cancelled the pending `setTimeout`, so up to 500ms of typing was dropped exactly when the form went away. It now flushes the pending write instead of discarding it. This is the part that makes tab switching work at all.
- **Drafts didn't reload when the key changed.** `useForm()` only reads `defaultValues` on mount, so a form that stays mounted across an Inertia navigation kept the previous draft.

Per @wescopeland's review on #4924, the key-change rehydration lives in `useFormDraft` rather than at the call site, so every caller with a changing key benefits instead of just this one. The hook resets from base defaults + `loadDraft(key)`, which is why callers now pass their base defaults as a third argument. `clearDraft()` cancels the pending write too, so a comment that was just submitted can't get written back out on unmount.

</details>

<details>
<summary><b>Testing</b></summary>

- New `useFormDraft` unit tests covering the debounce, the unmount flush, `clearDraft()` suppression, key-change rehydration, and unavailable storage.
- New `CommentList` tests for draft restore across unmount, no restore after submit, no bleed between different commentables, and no bleed between accounts.
- New `AchievementShowRoot` test walking the issue's exact repro: type a comment → switch to Unlocks → switch back → the text is still there. It fails on `master` and passes here.
- New cross-account tests for the forum and message forms: type a draft as one user, unmount, render as another, assert nothing is restored.
- Full JS suite passes (`sail pnpm test:run`, 4741 tests), along with `pnpm lint`, `pnpm tsc`, and `pnpm format`. Changed files are at 100% coverage.

Manually verified against a local Sail instance, signed in on an achievement page:

- Typing a comment then switching to Recent Unlocks and back restores the text, and Submit is immediately enabled (RHF re-validates the restored draft).
- The draft carries between the achievement page's Comments tab and `/achievement/{id}/comments`, so both cases from the issue discussion are covered.
- Submitting posts the comment and leaves the box empty on the next visit.

</details>

<details>
<summary><b>Why an empty entry stays in <code>sessionStorage</code> after submit</b></summary>

After a successful submit, `sessionStorage` holds `{"body":""}` rather than nothing. `clearDraft()` does remove the key and does suppress the pending write, so the submitted text is never written back. The empty entry comes from the `form.reset()` that follows, which persists the now-pristine form exactly as it would if a user had cleared the box by hand. It restores as an empty textarea, so I left it alone rather than adding pristine-value detection to the hook.

</details>

<details>
<summary><b>Review follow-ups</b></summary>

- All draft keys are scoped to the signed-in user. `sessionStorage` survives a logout and login in the same tab, so an account-agnostic key handed the previous user's unsubmitted content to whoever signed in next. This covers the pre-existing forum and message keys from #4667 as well, folded in here rather than left as a follow-up.
- `sessionStorage` writes and removals are wrapped, matching the read already guarded in `loadDraft`. `setItem` throws when storage is full or blocked, which took down the debounce timer and threw from the unmount cleanup while React was tearing the form down.

</details>

mariobob:
Also, [Greptile's second pass](https://github.com/RetroAchievements/RAWeb/pull/5130#discussion_r3690174413) flagged the same account-boundary issue against the existing forum and message draft keys from https://github.com/RetroAchievements/RAWeb/pull/4667, since those aren't scoped to a user either.

The root cause predates this PR, though my unmount flush does widen when drafts get written. I've got the fix ready on a branch off this one: https://github.com/mariobob/RAWeb/pull/1 (but unfortunately can't stack it across forked repo)

Want it folded in here, or as a separate PR once this merges?

wescopeland:
Want to try a [stacked PR](https://github.github.com/gh-stack/)?

mariobob:
I absolutely wanted to, but can't for two reasons:
1. [Docs](https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests): _Stacked pull requests require all branches to be in the same repository. Cross-fork stacks are not supported._
2. Permissions: I have `pull: true, push: false` on this repo, so my branches can't live here, only in my forked repo.

wescopeland:
Bummer. Let's add it to this branch's changes then rather than a follow-up.

mariobob:
Added, in https://github.com/RetroAchievements/RAWeb/pull/5130/changes/f51262a92dd5122468f9c23736f3b27d32feed02

resources/js/common/hooks/useFormDraft.ts
```diff
@@ -1,34 +1,84 @@
...
+    const timer = setTimeout(writeDraft, 500);
+    timerRef.current = timer;
+
+    return () => {
+      clearTimeout(timer);
+    };
   }, [key, values]);
 
+  useEffect(() => {
+    return () => {
```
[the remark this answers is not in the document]
mariobob: Done in f51262a92dd5122468f9c23736f3b27d32feed02.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
wescopeland: `?? 'guest'` feels disingenuous. Guests cannot comment.

resources/js/common/hooks/useFormDraft.ts
wescopeland, line 1: DX, major: `useFormDraft` has two hydration paths. The call sites own one of them.

All five call sites write the same three lines:
```ts
const draftKey = `...`;
const draftDefaultValues = { body: '' };
const form = useForm({ defaultValues: {...draftDefaultValues, ...loadDraft<FormValues>(draftKey) } });
const { clearDraft } = useFormDraft(draftKey, form, draftDefaultValues);
```

The hook can already hydrate the form via `form.reset({ ...baseDefaultValues, ...loadDraft<T>(key) })`. The merge thus exists twice - once at mount, and again on key change.

I think we should collapse the two paths into one hook and give that new hook the form.

```ts
export function useDraftForm<T extends FieldValues>(
  key: string | null,
  options: UseFormProps<T> & { defaultValues: DefaultValues<T> },
) {
  const scopedKey = useUserScopedDraftKey(key); // this is new
  const form = useForm<T>({
    ...options,
    defaultValues: { ...options.defaultValues, ...(scopedKey ? loadDraft<T>(scopedKey) : {}) },
});

// ... persistence ...

return { form, clearDraft };
```

This pattern gives us three benefits:
- `loadDraft` is no longer necessary in the five call sites.
- The call site gives the defaults one time rather than twice.
- Only one hydration rule needs to stay correct.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
wescopeland: DX, major: `auth?.user.id ?? 'guest'` is written in 5 files.

Scoping is a property of the draft key. It is not a property of the form.

Let's put this in one place, ie:
```ts
function useUserScopedDraftKey(key: ...): string | null {
  const { auth } = usePageProps();

  return key ? `${key}-${auth?.user.id ?? 'guest'}` : null; // i still don't feel confident about 'guest'
}
```

resources/js/common/hooks/useFormDraft.ts
```diff
@@ -1,34 +1,84 @@
...
-export function useFormDraft<T extends FieldValues>(key: string | null, form: UseFormReturn<T>) {
+export function useFormDraft<T extends FieldValues>(
+  key: string | null,
+  form: UseFormReturn<T>,
+  baseDefaultValues: DefaultValues<T>,
+) {
   const values = useWatch({ control: form.control });
+
   const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const pendingWriteRef = useRef<(() => void) | null>(null);
```
wescopeland: DX, minor: `pendingWriteRef` lets a cleanup function read the most recent values. A ref that holds the values can do this directly. React already runs all cleanup functions before any setup function in a commit, and the flush should always occur before a rehydrate reset. Knowing this, we can remove the manual `pendingWriteRef.current?.()` call and we can remove one effect entirely.

```ts
const values = useWatch({ control: form.control });
const latestRef = useRef(values);
latestRef.current = values;

useEffect(() => {
  if (!key) return;
  const timer = setTimeout(() => writeDraft(key), 500);

  return () => clearTimeout(timer);
}, [key, values]);

useEffect(() => {
  if (!key) return;

  return () => writeDraft(key);
}, [key]);
```

This should be logically equivalent with two less refs and one less effect. We allow React to control the "flush before reset" logic instead of rolling it ourselves.

Alternatively, we could keep `timerRef` and clear it in `clearDraft`. That still leaves us with just one ref and eliminating an effect.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
wescopeland: Bug, major: `clearDraft` does not actually durably clear this draft. The failure is caused by the immediate invocation of `form.reset(draftDefaultValues)`.

1. `clearDraft()` clears the timer, nulls `pendingWriteRef`, and removes the key.
2. `form.reset(draftDefaultValues)` changes the watched values.
3. The `[key, values]` effect sees new values and schedules a new 500ms write.
4. The write puts `{"body", ""}` back under the key which step 1 already removed.

The user does not see any incorrect text as a result, but we are indeed bleeding one leftover storage entry for each item the user comments on.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
mariobob: Agreed, I was in a dilemma over this, it was papering over a case that can't happen.

`useUserScopedDraftKey` returns `null` when nobody is signed in, and `null` already means "don't persist" in this hook. So a signed-out visitor stores nothing at all, rather than sharing a `guest` bucket.

resources/js/common/hooks/useFormDraft.ts
mariobob: Done, built as you sketched it. `useDraftForm` owns `useForm`, so the merge exists once and the call sites no longer touch `loadDraft`:

```ts
const { form, clearDraft } = useDraftForm<FormValues>(`create-topic-${forum.id}`, {
  resolver: zodResolver(formSchema),
  defaultValues: { title: '', body: '', postAsUserId: 'self' },
});
```

Options are `UseFormProps<T>` plus a required `defaultValues`, so anything `useForm` takes still passes straight through.

One wrinkle your sketch didn't cover: `useCreateMessageThreadForm` did `delete draft.recipient` before hydrating, so a stored recipient never lands back in the form. With the hook owning `useForm` the call site can't intercept that any more, so I added an `excludeFromDraft` option for it:

```ts
excludeFromDraft: ['recipient'],
```

It's one option for one caller, if you'd rather drop it and let the recipient restore like any other field I can do it. But that would change existing behaviour and one existing test, which is why I kept it.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
mariobob: Gone from all five. Scoping now happens inside `useUserScopedDraftKey`, called by `useDraftForm`, so call sites pass the plain key and never think about the user.

Combined with https://github.com/RetroAchievements/RAWeb/pull/5130/changes#r3699691443, `'guest'` doesn't exist anymore either.

resources/js/common/hooks/useFormDraft.ts
mariobob: Adopted, thank you. Both `pendingWriteRef` and `timerRef` are gone, and the manual `pendingWriteRef.current?.()` call with them.

I went one step further than your sketch: the flush effect and the rehydrate effect are both keyed on `[draftKey]`, so they merge into one. React runs the cleanup before the next setup, which is exactly the "flush before reset" ordering, so the merged effect gets it for free:

```ts
useEffect(() => {
  if (!draftKey) return;

  if (previousKeyRef.current !== draftKey) {
    form.reset({ ...defaultValues, ...readDraft<T>(draftKey, excludeFromDraft) });
  }
  previousKeyRef.current = draftKey;

  return () => writeDraft(draftKey);
}, [draftKey]);
```

That leaves 2 refs and 2 effects, down from 3 and 3.

`timerRef` disappeared because of the fix in https://github.com/RetroAchievements/RAWeb/pull/5130/changes#r3699784634: a stray timer firing after a clear now removes the entry instead of writing one, so there's nothing left to cancel.

resources/js/common/components/CommentList/useSubmitCommentForm.ts
mariobob: Confirmed and fixed, thank you.

`clearDraft` now lives with the form it resets, so the two aren't split across a hook boundary any more. More importantly, a write whose values match the defaults removes the entry rather than storing it:

```ts
if (isPristine(latestValuesRef.current, defaultValues)) {
  sessionStorage.removeItem(target);
} else {
  sessionStorage.setItem(target, JSON.stringify(latestValuesRef.current));
}
```

That makes the reset in step 2 harmless whenever it lands, so the ordering stops mattering. It also fixes the same leak from the other direction: a user who types into a form and then deletes it all no longer leaves an entry behind.

Two tests cover it, and both fail if the pristine check is removed:

- `given clearDraft is called, removes the entry and resets the form without writing it back`
- `given the user empties the form themselves, drops the stored entry`

wescopeland (approved):
Thanks!

I've made some slight optimizations in latest (namely swapping to the newer `form.subscribe()` API which is more optimized than using a watch).

mariobob:
Brilliant!

PR #5134, opened 2026-07-31
feat(manage): add an OAuthClientResource class

### LINKED ISSUE #5094: Add a Filament resource for OAuth applications
### User Story

As a site administrator,
I want to view and manage registered OAuth applications from the management app,
so that I can audit what apps exist, investigate abuse reports, and revoke problem clients without touching the DB.

### Why it matters

#5074 gives users self-service OAuth application management, which means the public can create clients that hold creds and request access to user accounts. Right now, the only admin-side visibility into those clients is raw/direct DB access. We'll need to answer "what is this app, who owns it, how many users are connected to it" and act on bad actors.

### Acceptance criteria

* A Filament resource lists registered OAuth applications with owner, client type, redirect URIs, and creation date.
* RAdmin can view an application's details, including how many users have active grants for it.
* RAdmin can revoke an application (disabling its tokens/grants) with the action audit-logged.
* Client secrets are never displayed in the panel.

Depends on #5074.

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5094.

http://localhost:64000/manage/o-auth-clients

Screenshot 2026-07-31 at 7 56 44 PM

This new resource is accessible only to admins. Clients can be revoked, and revokes are logged.


app/Models/OAuthClient.php
Jamiras, line 1: Should name be unique? Should we reserve some names?

Image

app/Filament/Resources/OAuthClientResource/Pages/AuditLog.php
Jamiras, line 1: The audit log is not filtered to the current record.

app/Filament/Resources/OAuthClientResource.php
Jamiras, line 1: The page title is "OAuth Applications", but the URL is "o-auth-clients"

app/Models/OAuthClient.php
wescopeland: I don't think we should do either. Uniqueness is trivially bypassed ("RAToo1s").

What I've done in latest is added the owner of the app to the consent screen:

Screenshot 2026-08-06 at 8 28 13 AM

Three things to note about the owner name bit:
1. It's a link that opens a new tab. This is required for this particular page, otherwise consent requests will be destroyed.
2. There's additional padding around the link so we don't fail accessibility checks on mobile.
3. The link doesn't use our link color, as this draws the eye away from the security concerns outlined to the user on the page.

app/Filament/Resources/OAuthClientResource/Pages/AuditLog.php
wescopeland: This is a surprising bug that is nontrivial to fix.

`audit_log.subject_id` is `BIGINT UNSIGNED`, and the OAuth client primary key (from Passport's own migrations, not ours) is a UUID string. A UUID obviously does not fit into an integer column, so MariaDB is coercing the strings to ints in the queries to build the audit log.

When OAuth clients are created around the same time, they get close-ish UUIDs that coerce to the _same_ int, which is why they appear on the same page.

Side note: the only reason MariaDB is allowed to do this at all is because we're not running MariaDB in strict mode. IMO, this should really cause a SQL error.

I don't want to mess with Passport's tables or implementation too much, so I'm taking a feathered touch approach to fix this. Here's what I've done:
* Latest includes a migration that adds a `numeric_id` column to the `oauth_clients` table.
* `numeric_id` is written in an incrementing fashion on model create.
* The audit log uses `numeric_id` as the subject ID for these entities.

app/Filament/Resources/OAuthClientResource.php
wescopeland: Fixed in latest.

app/Models/OAuthClient.php
Jamiras: While I agree with the sentiment, I'm not certain about the presentation.

Firstly, "registered by" reads weird. Maybe "authored by", "owned by", or "maintained by"?

Secondly, putting it between the "wants access to your account" and "This will allow" breaks up the natural flow of language between those two elements. But my only suggestion to improve it is to make it a footnote repeating the application name:

Image

Jamiras (approved):

app/Models/OAuthClient.php
wescopeland: I'm not certain we should change this - would like to discuss more in this thread before I merge this PR. Here's my current sense:
* This is an anti-phishing indicator. I feel lukewarm about is being placed under the action buttons. On some Android handhelds, it may be pushed beyond the fold.
* "Registered by" is vocabulary consistent with what we display on the settings page for OAuth ("Register application"). "Owned", "authored", and "maintained" all claim things we can't independently verify.

Looking at this with a fresh pair of eyes, I don't like us restating the app name here. I've changed "This will allow [app] to:" to "This will let it:".

Screenshot 2026-08-07 at 9 54 36 AM

Jamiras: I think my problem with "registered" is it has little meaning to a user.

Or maybe it's just the placement. It sort of reads like your account was registered by Jamiras, or the request for access was registered by Jamiras.

Maybe:
> Owner Line Test App
> Registered by Jamiras
> Wants access to your account.
> This will let it:

wescopeland: I'm struggling to land somewhere I like. I tried the "maybe" above, but it becomes multiple bold lines of equal weight.

Here's latest:
Screenshot 2026-08-07 at 12 00 00 PM

Jamiras: That works for me.

Curious why you chose to use a full-size label with a missing icon rather than the more subdued single-liner that was in previous iterations. For most use-cases, it's the least significant information in the dialog.

wescopeland: Out of scope for now, but I'm thinking at some point it may be desirable to allow folks to upload icons for their registered apps to give them more of an identity. The gamepads are both placeholders at the moment. There wasn't much beyond that driving it though.

Chew: This might also be out of scope, but a way to "verify" certain connections might help with user trust too. _Especially_ for our stuff, but maybe not for others, cus it may seem like we're responsible, but might mitigate impersonation if that ever becomes an issue, especially if we're giving away write access.

I would maybe also replace "Register" with "Developer", e.g. "App developed by," which more likely actually means something to the user.

Otherwise, I like the format. This is laid out similarly to Discord's, with "App wants to request" header on top, then the permissions, then additional metadata.

wescopeland: > This might also be out of scope, but a way to "verify" certain connections might help with user trust too. Especially for our stuff, but maybe not for others, cus it may seem like we're responsible, but might mitigate impersonation if that ever becomes an issue, especially if we're giving away write access.

Agreed - out of scope for this PR.

> I would maybe also replace "Register" with "Developer", e.g. "App developed by," which more likely actually means something to the user.

I don't think we should change this copy. We can't independently verify the user who registered the app is the same person who developed it.

PR #5137, opened 2026-08-01
feat(tickets): prevent most untracked users from creating tickets

### LINKED ISSUE #5098: Restrict ticket creation for untracked users
### User Story

As a developer,
I want untracked users without a team-based role to be unable to open tickets,
so that I stop spending time on reports filed by accounts whose unlock history has a signal of being untrustworthy.

### Why it matters

Untracked accounts are, in practice, accounts whose achievement data is stamped as performed outside the competitive norms of our system. Letting these users file tickets means a developer has to evaluate a bug report from a source that has already signaled it is unreliable.

### Acceptance criteria

* A user with `unranked_at` set and no displayable roles (excluding "Supporter") cannot create tickets of any kind.
* Users with a team-based role who are untracked can still create tickets.
* The "Report an Issue" UI does not offer a create action to a blocked user.

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5098.

Users who are untracked and lacking a team role can no longer create tickets.


Jamiras (approved):

Alexdatadestroyer:
What about people who get voluntarily Untracked and has no role? Is there a way (like an internal role or something) to differentiate voluntary from involuntary Untrackeds?

Normally, its not expected a voluntarily Untracked user to have roles. This can even happen if the user changes their email account too.

wescopeland:
They won't be able to open tickets. This is a known gap, and the affected user count is very small.

We don't have any mechanism currently for differentiating between voluntary and involuntary untracks. If we build this mechanism into the database one day, we can revisit.

PR #5139, opened 2026-08-01
feat(achievement): add 'Logic updated' deep links

### LINKED ISSUE #5100: Deep link achievement changelog logic entries to the Filament Logic page
### User Story

As a developer reading an achievement's changelog on the achievement details page,
I want a "Logic updated" entry to link straight to that version in the management app's Logic page,
so that I can see what actually changed without leaving the page, opening the management app, finding the achievement, and hunting for the right version.

### Why it matters

The changelog tab on `achievement.show` already tells you that logic changed, when, and by whom.

It cannot tell you _what_ changed, because trigger conditions are not something we render in our React UI (nor is this currently desirable). The place that answers that question already exists: the Logic page in Filament.

Right now, these two views are totally disconnected.

### Acceptance criteria

* A "Logic updated" changelog entry renders as a link when the viewer has `viewAchievementLogic`, and as plain text otherwise.
* The link targets the achievement's Filament Logic page.
* The link identifies the specific trigger version the entry correpsonds to, and the Logic page opens with that version expanded.
* `AchievementChangelogEntryData` carries the trigger version for entries sourced from the `triggers` table.
* Logic entries that predate `TRIGGERS_CUTOFF`, which come from legacy comments and have no trigger row, still render, and either link to the Logic page without a version or render as plain text.
* A version targeted by a deep link is reachable even when it falls past the eighth version, which the version history currently collapses behind "See N more".

wescopeland:
Resolves #5100.

Screenshot 2026-08-01 at 2 16 12 PM

If we have a diff associated with a logic update changelog entry, we now deep link to the Logic page for permissioned users.

### DIFF
--- a/resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
+++ b/resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
@@ -10,6 +10,7 @@ import {
 } from '@/common/components/+vendor/BaseTooltip';
 import { UserAvatar } from '@/common/components/UserAvatar';
 import { useFormatDate } from '@/common/hooks/useFormatDate';
+import { usePageProps } from '@/common/hooks/usePageProps';
 import { cn } from '@/common/utils/cn';
 
 import { FIELD_LEVEL_TRACKING_CUTOFF } from '../../utils/fieldLevelTrackingCutoff';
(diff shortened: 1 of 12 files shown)


resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
```diff
@@ -43,7 +46,13 @@ export const AchievementChangelogEntry: FC<AchievementChangelogEntryProps> = ({
...
       <div className="flex min-w-0 flex-col gap-0.5">
         <p className="flex items-center gap-1 text-xs">
-          {header}
+          {logicHref ? (
+            <a href={logicHref} target="_blank" rel="noopener noreferrer">
+              {header}
+            </a>
+          ) : (
+            header
+          )}
```
Jamiras: Conceptually I like the feature, but I feel that turning the audit message into a link makes it harder to visually track the page.

Image

Would it be practical to leave "Logic updated" in blue to better separate the sections, and just put a link icon next to the text?

Jamiras (approved):

resources/js/features/achievements/components/AchievementChangelog/AchievementChangelogEntry.tsx
wescopeland: Yea, that works for me. Updated in latest:

Screenshot 2026-08-07 at 1 10 30 PM
Screenshot 2026-08-07 at 1 10 28 PM

PR #5141, opened 2026-08-01
fix(seo): format meta description numbers with an explicit locale

### LINKED ISSUE #5133: Frontend tests fail outside en-US locales
**Describe the bug**

Three frontend tests fail on a clean checkout for any contributor whose shell locale formats numbers with a `.` thousands separator (most of continental Europe). The assertions hard-code US-formatted numbers, but the code under test formats via i18next's `{{val, number}}` interpolation, which resolves to Node's default `Intl` locale (and that comes from the environment).

`package.json` pins the timezone for tests (`"test": "TZ=UTC vitest run ..."`) but nothing pins the locale, so the tests inherit whatever `LC_ALL`/`LANG` the contributor's terminal has.

CI never catches this because GitHub's Ubuntu runners default to `C`/en-US.

**To Reproduce**

From a clean checkout, with any locale that groups thousands with a dot:

```shell
LANG=hr_HR.UTF-8 pnpm test:run
```

```shell
FAIL  resources/js/features/game-list/utils/buildSystemGamesMetaDescription.test.ts
  > given a very large count, rounds down to the nearest hundred and adds a plus sign
  Expected: "Explore 2,300+ SNES games on RetroAchievements. ..."
  Received: "Explore 2.300+ SNES games on RetroAchievements. ..."

FAIL  resources/js/features/games/hooks/useGameMetaDescription.test.ts
  > given the user is viewing published achievements with large point values, formats stuff correctly with commas
  Expected: "There are 50 achievements worth 1,200 points. ..."
  Received: "There are 50 achievements worth 1.200 points. ..."

FAIL  resources/js/features/games/hooks/useGameMetaDescription.test.ts
  > given the user is viewing unpublished achievements with large point values, formats stuff correctly with commas
  Expected: "There are 50 achievements worth 1,500 points. ..."
  Received: "There are 50 achievements worth 1.500 points. ..."

Test Files  2 failed | 569 passed (571)
     Tests  3 failed | 4731 passed (4734)
```

Forcing the locale makes them pass:

```shell
LC_ALL=en_US.UTF-8 pnpm test:run   # 4734 passed
```

The failure isn't limited to dot-separator locales. Locales differ in whether they group 4-digit numbers at all (CLDR `minimumGroupingDigits`), so `sl_SI` and `it_IT` fail the same tests with an *ungrouped* `2300`:

```shell
Received: "Explore 2300+ SNES games on RetroAchievements. ..."
```

`hr_HR`, `de_DE` and `nl_NL` produce the `2.300` shown above. Only en-US-style locales pass.

**Expected behavior**

`pnpm test:run` passes on a clean checkout regardless of the contributor's system locale, the same way it already does regardless of their timezone.

This also blocks pushing, since `.hooks/pre-push` runs the suite.

**Additional context**

Two possible fixes:

1. Pin the locale next to the timezone in `package.json`, mirroring what's already done for `TZ`:
   ```json
   "test": "TZ=UTC LC_ALL=en_US.UTF-8 vitest run --coverage --coverage.reporters=text --reporter=dot",
   "test:run": "TZ=UTC LC_ALL=en_US.UTF-8 vitest --run",
   ```
   Smallest change, and consistent with the existing `TZ` precedent. Worth applying to `.hooks/pre-push` too if it doesn't go through these scripts.

2. Make the three assertions locale-independent, or have the tests pin the i18next/`Intl` locale explicitly rather than relying on the ambient default.

Option 1 is probably preferable since it fixes the whole class rather than the three tests that happen to trip on it today, since any future `{{val, number}}` assertion has the same exposure.

Happy to open a PR for whichever approach you'd prefer.

**Desktop**
 - OS: macOS 15 (Darwin 24.6.0), reproduced with system locale `en_HR` and shell `LANG=hr_HR.UTF-8`
 - Node 24 / pnpm 10.17, matching the CI pins

mariobob:
Follow-on to #5140, which pinned the locale for test runs and closed #5133. This fixes the call sites those tests were pointing at, so it's complementary rather than an alternative. Rebased on top of #5140; no overlap with it.

## Summary
Five meta descriptions format their numbers with a bare `.toLocaleString()`, which resolves against whatever locale the runtime happens to have. `formatNumber()` is what the rest of the codebase already uses for this (29 files call it directly and 27 more go through `useFormatNumber()`) and it defaults to `en-US`. These five were the only bare calls left.

Output is unchanged for anyone already running in an en-US-style locale, including CI.

<details>
<summary><b>Why it's worth doing on top of the test pin</b></summary>

These strings are built inside components that render through Inertia SSR, where `.toLocaleString()` reads the **SSR process's** locale, so the formatting depends on how the server happens to be provisioned rather than on anything deterministic. `useFormatNumber()`'s docblock names this exact hazard: it exists to prevent "SSR hydration mismatches from locale leaking between concurrent requests".

To be clear about the severity: this is latent rather than visibly broken. A server running with `LANG` unset or `C` resolves to `en-US`, so production output is almost certainly correct today; it just isn't guaranteed by anything except the deployment environment. The change makes it explicit, and brings the last five stragglers in line with the convention the other 56 call sites already follow.

</details>

<details>
<summary><b>Testing</b></summary>

The suite passes under every locale that reproduced the original #5133 failures, with nothing pinned:

| `LANG` | Result |
| --- | --- |
| `hr_HR.UTF-8` | 4720 passed |
| `de_DE.UTF-8` | 4720 passed |
| `sl_SI.UTF-8` | 4720 passed |
| `it_IT.UTF-8` | 4720 passed |

`hr_HR`, `de_DE` and `nl_NL` group thousands with a `.`; `sl_SI` and `it_IT` don't group 4-digit numbers at all (CLDR `minimumGroupingDigits`), so they failed with an ungrouped `2300`. Both classes are fixed at the source.

Note that since #5140 the npm scripts set `LC_ALL` themselves, so reproducing the matrix means invoking vitest directly:

```shell
env -u LC_ALL LANG=hr_HR.UTF-8 TZ=UTC ./node_modules/.bin/vitest --run
```

`pnpm lint`, `pnpm tsc`, `pnpm format` and the coverage thresholds all pass.

</details>

## AI disclosure
Claude Code was used for this change: it traced the root cause, made the edits, and ran the locale matrix above. I reviewed every line and can explain the implementation.

### DIFF
--- a/resources/js/features/game-list/hooks/useHubPageMetaDescription.ts
+++ b/resources/js/features/game-list/hooks/useHubPageMetaDescription.ts
@@ -2,6 +2,7 @@ import { usePage } from '@inertiajs/react';
 
 import { usePageProps } from '@/common/hooks/usePageProps';
 import { cleanHubTitle } from '@/common/utils/cleanHubTitle';
+import { formatNumber } from '@/common/utils/l10n/formatNumber';
 
 export function useHubPageMetaDescription() {
   const page = usePage();
@@ -21,7 +22,7 @@ export function useHubPageMetaDescription() {
   }
 
   if (paginatedGameListEntries.total) {
-    return `Explore a collection of ${paginatedGameListEntries.total.toLocaleString()} ${paginatedGameListEntries.total === 1 ? 'classic game' : 'classic games'} in the ${cleanedTitle} hub.`;
+    return `Explore a collection of ${formatNumber(paginatedGameListEntries.total)} ${paginatedGameListEntries.total === 1 ? 'classic game' : 'classic games'} in the ${cleanedTitle} hub.`;
   }
 
   // This only happens with orphaned hubs, which probably generate no SEO juice anyway.
--- a/resources/js/features/game-list/utils/buildSystemGamesMetaDescription.ts
+++ b/resources/js/features/game-list/utils/buildSystemGamesMetaDescription.ts
@@ -1,3 +1,5 @@
+import { formatNumber } from '@/common/utils/l10n/formatNumber';
+
 export function buildSystemGamesMetaDescription(totalGames: number, systemName: string): string {
   let firstPhrase;
 
@@ -11,7 +13,7 @@ export function buildSystemGamesMetaDescription(totalGames: number, systemName: 
   } else {
     // For a games count of 100+, round down to nearest hundred and add a plus sign.
     const roundedGames = Math.floor(totalGames / 100) * 100;
-    firstPhrase = `Explore ${roundedGames.toLocaleString()}+ ${systemName} games on RetroAchievements.`;
+    firstPhrase = `Explore ${formatNumber(roundedGames)}+ ${systemName} games on RetroAchievements.`;
   }
 
   const secondPhrase = 'Track your progress as you beat and master each title.';
--- a/resources/js/features/games/hooks/useGameMetaDescription.ts
+++ b/resources/js/features/games/hooks/useGameMetaDescription.ts
@@ -1,4 +1,5 @@
 import { usePageProps } from '@/common/hooks/usePageProps';
+import { formatNumber } from '@/common/utils/l10n/formatNumber';
 
 import { getAchievementSetPointsStats } from '../utils/getAchievementSetPointsStats';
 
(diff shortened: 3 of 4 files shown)


wescopeland (approved):
Thanks!

PR #5145, opened 2026-08-02
feat: support watched term alerts for the mod team

### LINKED ISSUE #5091: Alert moderators when comments/posts match a configurable term watchlist
### User Story

As a moderator,
I want an automatic alert when a new comment matches a configurable list of watched terms,
so that I can act on these comments quickly instead of discovering them through manual eyeballing and user reports.

### Why it matters

Comments promoting or teaching cheating methods show up nearly every day, and the current workflow is the mod team manually patrolling reports and deleting these comments one at a time. An alert in the moderation feed on Discord removes the patrol work. Making the term list configurable keeps specific tool names out of the codebase and lets the team adjust the list as new terms show up without requiring a code change.

### Acceptance criteria

- Staff can manage the list of watched terms without a redeploy.
- When a new comment or forum post matches a watched term, an alert with a link to the comment is posted to `#mod-feed`.
- Matching is case-insensitive.
- Users on the site see no change in behavior and receive no indication their comment/post matched.

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5091.

https://discord.com/channels/476211979464343552/1488680676021768414/1533221430324629614

Watched terms can now be configured by users with `Role::MODERATOR` attached. When a given term is written on a forum post or comment, it will surface in the moderator feed.

Screenshot 2026-08-02 at 11 51 29 AM


app/Support/Alerts/WatchedTermAlert.php
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public function toDiscordMessage(): string
+    {
+        $userUrl = route('user.show', ['user' => $this->user]);
+
+        $termLabel = count($this->matchedTerms) === 1 ? 'watched term' : 'watched terms';
+        $terms = implode(', ', array_map(
+            fn (string $term): string => sprintf('`%s`', $term),
+            $this->matchedTerms,
+        ));
```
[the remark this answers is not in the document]
wescopeland: Not a P1 - this assumes a user probably knows the list ahead of time to exploit the flaw, and the list itself is private. This is very unlikely to ever occur in practice.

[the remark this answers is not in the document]
wescopeland: Yes

app/Community/Enums/CommentableType.php
```diff
@@ -80,6 +80,16 @@ public static function fromLegacyInteger(int $value): ?self
         };
     }
 
+    public function isManagementComment(): bool
```
Jamiras: Would `isUmoderatedComment` be better?

app/Models/ModerationWatchlistTerm.php
Jamiras, line 1: Is there any reason to have a `created_by` field or some form of audit log? There's nothing indicating who added a word (or changed one to something else).

tests/Feature/Community/Services/WatchedTermMatcherTest.php
```diff
@@ -0,0 +1,66 @@
...
+    // Assert
+    expect($matches)->toEqual(['watchedtool']);
+})->with([
+    'exact' => 'you should try watchedtool for this',
+    'different case' => 'you should try WatchedTool for this',
+    'inside a longer word' => 'grab the watchedtoolkit build',
+    'inside a URL' => 'see https://example.com/watchedtool/downloads',
+    'split by bold tags' => 'grab watched[b][/b]tool here',
+    'split by italic tags' => 'grab watched[i][/i]tool here',
+    'split by a url tag' => 'grab watched[url=https://example.com][/url]tool here',
```
Jamiras, line 27: What about in a url tag?
```
grab it [url=https://watchedtool.com]here[/url]
```

Jamiras: On a related note: Since the minimum term length is four characters, are we at all concerned about falsely matching a substring in a hash like YouTube videos and bit.ly links generate?

app/Community/Enums/CommentableType.php
wescopeland: I think it may be equally problematic, as similar to "management" it's describing something akin to a workflow. 

I've renamed this in latest to `isPublicComment`, flipped the things within it, and changes the usages to use the `!` operator (to avoid a double negative that's difficult to reason about).

app/Models/ModerationWatchlistTerm.php
wescopeland: Added an audit log in latest:
Screenshot 2026-08-07 at 1 21 46 PM

Screenshot 2026-08-07 at 1 21 50 PM

tests/Feature/Community/Services/WatchedTermMatcherTest.php
wescopeland: > What about in a url tag?

Already works, but definitely worth locking in via a test. Added these cases in latest.

> Since the minimum term length is four characters, are we at all concerned about falsely matching a substring in a hash like YouTube videos and bit.ly links generate?

It's worth noting, but I don't think we should fret too much over it. YouTube video IDs are 11 chars. A four-letter term matching has roughly 1 in 260,000 per link odds, per term. Given this is a signal meant for someone to manually review, it's probably not worth too much complexity to protect against stray/false hits.

Jamiras (approved):

PR #5147, opened 2026-08-02
fix: restore deferred props lost to the bfcache

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1523653400158277642/1523653400158277642.

The root cause is similar to something we've repaired before with the game page. It's caused by some parts of the UI being in React and others being in Inertia. A full page nav cancels any deferred props request that's still in-flight. This leads to lost deferred props staying in a loading state indefinitely. Screenshots are loaded via deferred props, which is what's causing the reported behavior.

`useReloadDeferredOnBackForward` solved this problem for a few other game page props. Rather than sprinkling this hook around even more, I've deleted it and lifted the behavior up to a global in our client-side _app.tsx_:
```ts
window.addEventListener('pageshow', restoreDeferredPropsOnPageShow);
```
This will keep us from forgetting to re-apply it anytime we use deferred props.


Jamiras (approved):
Could not reproduce locally using steps in discord and a random game. Did not introduce any new issues, so willing to give it a pass.

PR #5148, opened 2026-08-02
fix(game): sort achievements by the selected key only

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1509946937892278344/1509946937892278344.

Four sort orders group unlocked achievements first before applying the selected sort key. The original PHP game page did not exhibit this behavior - it's something the React page adopted, and I believe it's undesirable because it makes our UI dishonest. This PR removes that grouping from everything except "Unlocked first".

**Before**
Screenshot 2026-08-02 at 3 11 04 PM

**After**
Screenshot 2026-08-02 at 3 10 54 PM


resources/js/common/utils/sortAchievements.ts
```diff
@@ -177,13 +117,6 @@ export function sortAchievements(
...
 
       return sortedAchievements.sort((a, b) => {
-        // First, prioritize unlocked achievements.
-        const unlockedResult = compareUnlockStatus(a, b);
-        if (unlockedResult !== 0) {
-          return unlockedResult;
-        }
-
-        // Then, sort by unlocksHardcore within each group (unlocked and not unlocked).
         const unlocksDiff = (a.unlocksHardcore as number) - (b.unlocksHardcore as number);
```
Jamiras: Unrelated: If `unlocksHardcore` is the same, it should also check non-hardcore unlocks:

http://localhost:64000/game/2858?sort=wonBy

22 > 21 > 21 > 17, but the second 21 has more total unlocks than the first.

Image

Jamiras (approved):

resources/js/common/utils/sortAchievements.ts
wescopeland: Thanks, I've fixed this in latest.

PR #5149, opened 2026-08-02
fix(manage/logic): normalize decoded hex so equivalent logic stops showing as modified

wescopeland:
Resolves https://discord.com/channels/310192285306454017/1505528547559866428/1505528547559866428.

http://localhost:64000/manage/achievements/340159/logic

**Before**
Screenshot 2026-08-02 at 4 34 55 PM

**After**
Screenshot 2026-08-02 at 4 34 47 PM

### DIFF
--- a/app/Platform/Services/TriggerDecoderService.php
+++ b/app/Platform/Services/TriggerDecoderService.php
@@ -11,6 +11,13 @@ use Illuminate\Support\Str;
 
 class TriggerDecoderService
 {
+    private function formatHex(string $hex): string
+    {
+        $trimmed = ltrim(strtolower($hex), '0');
+
+        return '0x' . str_pad($trimmed, 6, '0', STR_PAD_LEFT);
+    }
+
     private function parseOperand(string $mem): array
     {
         $end = strlen($mem);
@@ -112,7 +119,7 @@ class TriggerDecoderService
                 $count++;
             }
 
-            $value = '0x' . str_pad(substr($mem, 0, $count), 6, '0', STR_PAD_LEFT);
+            $value = $this->formatHex(substr($mem, 0, $count));
             $mem = substr($mem, $count);
 
             return [$type, $size, $value, $mem];
@@ -179,7 +186,7 @@ class TriggerDecoderService
             $count++;
         }
 
-        $address = '0x' . str_pad(substr($mem, 0, $count), 6, '0', STR_PAD_LEFT);
+        $address = $this->formatHex(substr($mem, 0, $count));
         $mem = substr($mem, $count);
 
         return [$type, $size, $address, $mem];
--- a/app/Platform/Services/TriggerDecoderService.php
+++ b/app/Platform/Services/TriggerDecoderService.php
@@ -11,11 +11,17 @@ use Illuminate\Support\Str;
 
 class TriggerDecoderService
 {
+    public function formatAddress(int $address): string
+    {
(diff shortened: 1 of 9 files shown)


app/Platform/Services/TriggerDecoderService.php
```diff
@@ -11,6 +11,19 @@
...
+    {
+        return $this->formatHex(dechex($address));
+    }
+
+    private function formatHex(string $hex): string
+    {
+        $trimmed = ltrim(strtolower($hex), '0');
+        $width = strlen($trimmed) > 6 ? 8 : 6;
+
+        return '0x' . str_pad($trimmed, $width, '0', STR_PAD_LEFT);
```
Jamiras, line 84: This solution really only addresses larger addresses where the default padding to 6-digits was not sufficient. Ideally, we'd select an appropriate length for formatting based on the needs of the achievement.

The [DLL logic](https://github.com/RetroAchievements/RAIntegration/blob/29bacaad41e40b6e929e102f4262a103380d27ab/src/devkit/context/impl/EmulatorMemoryContext.cpp#L90-L95) is a bit more complicated as it looks at the length of the maximum address for the system and rounds that up to the nearest multiple of two:
```
    if (m_nTotalMemorySize <= 0x10000)
        m_fFormatAddress = FormatAddressSmall;
    else if (m_nTotalMemorySize <= 0x1000000)
        m_fFormatAddress = FormatAddressMedium;
    else
        m_fFormatAddress = FormatAddressLarge;
```
And RATools takes it a step farther and formats things based on the largest address used in the set, allowing sets only using the base system memory to not have extra padding for extra memory (Apple II) or banked memory (GBC).

As historically, all addresses were formatted to 6 digits, having some be 4 digits now causes some no-op changes to get logged when submitting unrelated changes. In this case, I updated the achievement badge:
```
select * from triggers where triggerable_type='achievement' and triggerable_id=83383;
```
And the addresses went from 6 digits (old default) to 4 digits (RATools optimized).

This already shows as "no changes" in [the existing UI](http://localhost:64000/manage/achievements/83383/logic) because the addresses are padded to 6-digits. At some point, it might be nice to detect this during upload and not create a new version.

Jamiras: I guess that's just a long-winded way to say that the width should be uniform for the achievement instead of calculated for each value. Otherwise some rows might use 6 digits and others use 8. And if we could reasonably decide to use 4 for systems with less than 64KB of RAM that would be even better.

Jamiras (approved):

app/Platform/Services/TriggerDecoderService.php
wescopeland: Alright, I've done a rework.

Decode now finds the largest mem address in the trigger code, rounds its digit count up to 4 / 6 / 8, and pads every address to that width. Since we don't know the system's memory size deterministically in RAWeb, we key off the largest address used, similar to RATools, so logic that occupies a smaller memory space renders at 4 digits.

wescopeland:
@greptile-apps review

wescopeland:
> The remaining eight-digit address-width loss should be fixed before merging.
> Width calculation strips leading zeroes, so an explicit address such as 0x00123456 still renders as 0x123456.

@greptile-apps That is the intended behavior and not a bug. I am not going to change anything. Leading zeroes are not semantic in these trigger strings - treating them as significant is actually what caused the bug this PR is trying to fix.

PR #5154, opened 2026-08-07
feat(game): add marker for event achievements to achievements list on game page

Jamiras:
Adds a marker to the achievements list on the game page indicating that the achievement is active in at least one event.

~~Marked as draft because I'm seeking help with UI issues.~~


resources/js/common/components/ActiveEventsIndicator/ActiveEventsIndicator.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+interface ActiveEventsIndicatorProps {
+  activeEvents: NonNullable<App.Platform.Data.ActiveEventAchievement['type']>;
+}
+
+export const ActiveEventsIndicator: FC<ActiveEventsIndicatorProps> = ({
+  activeEvents,
+}) => {
+  const link = activeEvents.link;
+  const label = activeEvents.summary.replace('\n', '<br/>');
```
Jamiras: I can't get this to actually be multiline.

Image

resources/js/common/components/RaEvent/RaEvent.tsx
```diff
@@ -0,0 +1,18 @@
+import type { IconType } from 'react-icons/lib';
+
+export const RaEvent: IconType = (props) => {
+  return (
+    <svg
+      {...props}
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      fill="currentColor"
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 36 36"
```
Jamiras: This needs to be replaced with a 24x24 SVG.

resources/js/common/components/AchievementsListItem/AchievementsListItem.tsx
```diff
@@ -168,6 +184,12 @@ export const AchievementsListItem: FC<AchievementsListItemProps> = ({
         {playersTotal !== null ? (
           <div className="mt-1 md:col-span-2 md:flex md:flex-col-reverse md:justify-end md:gap-y-1 md:pt-1">
             {/* Meta chips (Desktop) */}
+            {activeEventAchievement ? (
+              <div className="hidden items-center justify-end gap-x-1 md:flex">
+                <ActiveEventsIndicator activeEvents={activeEventAchievement} />
+              </div>
+            ) : null}
+
```
Jamiras: When both an event indicator and a type indicator are present, they align vertically.

Image

resources/js/common/components/ActiveEventsIndicator/ActiveEventsIndicator.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <div
+            data-testid='type-active_event'
+            className={cn(
+              'type-ind group',
+              'cursor-pointer',
+              'flex items-center gap-1',
+            )}
+            >
+            <div aria-label='active_event'>
+              <RaEvent className="size-4.5" stroke={stroke} border="none" />
```
Jamiras: border="none" doesn't make the border go away

app/Platform/Actions/BuildGameShowPagePropsAction.php
wescopeland, line 1: Unrelated to this file, I think we should also revisit https://github.com/RetroAchievements/RAWeb/pull/3260.

It would really be nice to indicate these somehow in emulators. For events like AOTW and AOTM, I like to just fire up the game and not even look at my computer, but I nearly always forget what the target achievement is and I suspect I'm not the only one.

wescopeland (commented):
Will carve out some time to assist tomorrow. Wanted to leave 1 meta discussion comment in the meantime.

app/Platform/Actions/BuildGameShowPagePropsAction.php
Jamiras: > I nearly always forget what the target achievement is and I suspect I'm not the only one.

That's half of the motivation for this. When I open the page for the AotW game, it would be nice to be able to quickly find the AotW achievement.

The other half is for when I'm not playing the AotW. I might accidentally select a game that's associated to an evergreen event. Having a marker would be a signal that I could make progress on an evergreen event, and might make a difference on how long I decide to stick with the game if it's not quite grabbing my attention.

As for #3260, I'd be fine with having some sort of marker in the client indicating an achievement is associated to an event, but I don't think I want to go so far as to indicate which event(s). I also would limit the indicator to the list and not include it in the popup.

Also note that RetroArch doesn't really have any way to add custom indicators to the achievement lists. We still use `[m]` for missable there
image
Other clients might be similarly restricted.

wescopeland:
I've made numerous updates in https://github.com/wescopeland/RAWeb/commit/f84205c53c3d5848372d1b1f101d22a88c583f36. You're welcome to cherry-pick that commit onto this branch if you'd like.

Here's the summary of changes in the commit:
* The back-end DTO has been reshaped. `link` became `eventId`, and `summary` became `eventTitle` + `activeUntil`. This lets the UI completely own the presentation layer. With `eventId`, we can also use `<InertiaLink />` for faster client-side routing.
* `buildActiveEventAchievements` was greatly simplified as a result of these DTO changes (the complexity was moved to the front-end).
* The front-end has new copy indicating how long an event achievement is active for.
* `RaEvent` has been removed, replaced by `LuCalendar` / `LuCalendarCheck`.
* Various CSS adjustments to get the alignment and styling right.

Screenshot 2026-08-08 at 10 50 06 AM

Screenshot 2026-08-08 at 10 50 18 AM

Screenshot 2026-08-08 at 10 42 57 AM

Screenshot 2026-08-08 at 11 18 40 AM

---

I did make one more behavior change that probably represents the sharpest opinion of the commit: **evergreen achievements no longer get the indicator.**

Without this filter, 1,332 games show an achievement with this new indicator. That is roughly 1 in 10 of every playable game on the site, and this compounds over time as evergreen events keep accumulating.

It also skews toward exactly the games people look at. The median players on a game covered under evergreen is 973. Without is 249.

It's true that these achievements are part of an event, and someone may want to know, but I don't think this is the right place to surface that. My sense is the better UX is to have this indicator surface urgency, which it now does.

Jamiras:
> I did make one more behavior change that probably represents the sharpest opinion of the commit: **evergreen achievements no longer get the indicator.**

> Without this filter, 1,332 games show an achievement with this new indicator. That is roughly 1 in 10 of every playable game on the site, and this compounds over time as evergreen events keep accumulating.

I didn't realize it was quite that high, but that's half of why I wanted to implement this feature. I want to know that I'm actually done with a game when I decide to move on to something else. If there's _any_ achievement I could earn with just a little more effort, I want to know. If it's in an evergreen event where I already have progress, I might even be willing to put forth extra effort to get it. Otherwise I have to revisit the game later when I realize that I missed something.

While the time-limited events could be prioritized, people seeking those will be going specifically to the associated games knowing the event credit is there. I also want something to inform users when they visit a random game that there could be event credit gained from playing it.

If you feel that an extra indicator is that intrusive, perhaps it could be associated to a profile setting, or some sort of opt-in associated to the event?

wescopeland:
> perhaps it could be associated to a profile setting

I like this idea. I'll have a commit in my branch for this shortly.

Jamiras:
I've cherry-picked your changes (thanks).

The only additional change I've made at this time is to change "Last day" to "Less than an hour left" as there was already logic for "3 hours left".

wescopeland:
Played with a bunch of ideas, but I think the setting is probably the best approach. I finally landed on the setting implementation in https://github.com/wescopeland/RAWeb/commit/327c9171a7cc79dc016d17e4f44e4cb15b2e34b7.

Screenshot 2026-08-08 at 3 33 59 PM

Screenshot 2026-08-08 at 3 34 05 PM

Jamiras:
Pulled in the profile changes. At this point, you've written more code than me. I'm not sure how you want to handle approving. From my perspective, it seems to be functional, and meets my original intent.

I did notice that the evergreen events are mostly labelled as Evergreen in the event titles, so saying the event is evergreen is redundant. Maybe instead of saying Evergreen for the expiration it should say "No time limit" or something similar?
image

wescopeland:
👍  I'll fix CI and then we ask the robot what it thinks.

wescopeland:
@greptile-apps review

wescopeland:
Looks like as a co-author, GitHub hides the Approve button from me 😅

I'm happy with this implementation. If you're happy with this implementation, let's merge it.

Jamiras:
> I'm happy with this implementation. If you're happy with this implementation, let's merge it.

I'm happy too. 

My only comments are the thing about Evergreen being redundant in the tooltip (mentioned above) and the tooltip isn't accessible in mobile, so you have to click through to see which event it's associated to. It's not a showstopper, and the only solution I can think of is to pop a dialog that links to the events instead of having the indicator link to the event. If that's desirable, it should probably be a separate PR.

PR #5157, opened 2026-08-07
chore(api/v2): adjust CORS

### LINKED ISSUE #5156: V2 API cant be called from browsers: CORS preflight is never answered on the api.retroachievements.org subdomain
**Describe the bug**
I was testing and preparing some things on my site that could make use of OAuth when it
releases (RetroAchievements-based web widgets, mainly for iiSU). I was not able to call the
V2 API tho, the request dies with a CORS error. The CheevoPresence app I have uses the same
endpoint with the same header auth without problems, but its a native desktop app so I think it only affects browser based clients.

I used Claude to diagnose the error since I thought it was maybe on my end, but it
turned out the [CORS config file](https://github.com/RetroAchievements/RAWeb/blob/master/config/cors.php) only matches the `api/*` path. Since the V2 API is on a subdomain, its path after the domain is `v2/...` and results into the CORS error

**To Reproduce** (using example.com):

In a browser console:
```js
// the call to the "api." subdomain with preflight failing:
fetch('https://api.retroachievements.org/v2/users/Denzi', { headers: { Authorization: 'Bearer x' } })
```

Image

```js
// same request against a lowercase api/* path and preflight answers 204:
fetch('https://retroachievements.org/api/v2/users/x', { headers: { Authorization: 'Bearer x' } })
```
Image

So my proposed fix would be, in [cors.php](https://github.com/RetroAchievements/RAWeb/blob/master/config/cors.php):
```php
'paths' => ['api/*', 'v2/*', 'oauth/*', 'sanctum/csrf-cookie'],
```

Including `oauth/*` can't hurt (I think) because browser-based public clients will need to read the
`/oauth/token` response once OAuth launches. With this change, browser-based cross-origin
requests to V2 should not run into the CORS error anymore!

wescopeland:
Closes https://github.com/RetroAchievements/RAWeb/issues/5156.


Jamiras (approved):
Untested (not really familiar with the functionality). Trusting your knowledge and the unit test coverage.

PR #5158, opened 2026-08-08
feat(api/v2): generate and publish an OpenAPI spec

### LINKED ISSUE #5092: Generate an OpenAPI spec for the V2 API
### User Story

As a maintainer of the V2 API and its documentation,
I want an OpenAPI spec generated from RAWeb's actual V2 server definitions,
so that docs and typed clients stay accurate without anyone hand-maintaining a 20,000-line spec that silently drifts from the code as it changes/evolves over time.

### Why it matters

https://github.com/RetroAchievements/api-docs/pull/100 is building V2 docs on top of a hand-authored OpenAPI spec. This is prone to drift, and no one will notice until a consumer hits the drift. RAWeb's schema classes must be the source of truth for resources, attributes, filters, includes, and pagination, so the spec should be emitted from them.

A generated spec also unlocks static type generation for TS/Kotlin/etc consumers and docs rendering via something like Scalar. This gives us one place to document auth requirements (also: see discussion on #5074 about documenting which OAuth scopes each endpoint requires).

### Acceptance criteria

* An _openapi.json_ for the V2 API is generated from RAWeb's V2 server definitions via a single Artisan command, covering every registered resource's endpoints, attributes, filters, includes, sorts, and pagination.
* The spec documents authentication for each endpoint, including required OAuth scopes where they apply.
* Regenerating the spec is deterministic, so drift between the code and the spec is mechanically detectable.
* The spec is published somewhere consumers can reach it. Where docs render from it (a Scalar page hosted by RAWeb, the api-docs site, or elsewhere) is intentionally left open and will be decided when the spec exists.

wescopeland:
Resolves #5092.

This branch has an outrageous number of line additions, but that count is misleading. It almost entirely comes from the generated spec file itself, which needs to be committed to the repository and deliberately not excluded from diffs. This is a 2MB file that compresses to about 25KB, so I have no concerns checking it in.

This PR allows RAWeb to emit an OpenAPI spec for the V2 API. The spec can then be consumed by various documentation tools. A new CI task detects drift in the spec. If there is drift, the task will fail unless the developer who opened the PR runs this command:
```
sail artisan ra:api:generate-openapi-spec
```
After running the command, _public/openapi/v2.json_ gets updated, and the diff is viewable in the PR itself.

The spec file should _not_ be a deploy-time artifact because the spec is a contract and not a build output. Comitting it keeps the generator out of production dependencies and forces every contract change to show up in a PR where someone can see it.

There is no tooling provided in this PR for actually viewing or interacting with the spec. This branch is completely unopinionated in that regard.

There is an issue with the spec currently in that every attribute currently reads as non-nullable. Integer counts are also typed as `number` rather than `integer`. Both of these problems will be fixed in a follow-up (which will also surface the diff in the spec).


public/openapi/v2.json
Jamiras, line 1: Is `{serverUrl}` the expected value for `url`? Should there be more variables other than the default?
```
    "servers": [
        {
            "url": "{serverUrl}",
            "variables": {
                "serverUrl": {
                    "default": "https://api.retroachievements.org/api/v2"
                }
            }
        }
    ],
```

Jamiras (approved):

public/openapi/v2.json
wescopeland: This is coming from the spec generator itself. This is valid but kinda confusing. I've updated the spec generator so this isn't coming from a variable anymore.

PR #5161, opened 2026-08-10
fix(forums): track edited_at separately from updated_at

### LINKED ISSUE #5096: Authorizing a user's forum posts falsely marks them as edited
### User Story

As a new user whose first forum post was held for verification,
I want my posts to appear as originally written once a moderator authorizes me,
so that they aren't publicly labeled "last edited" for something I never did.

### Why it matters

When a moderator verifies a forum post for an account, we flip the account's whole batch of `is_authorized` and `authorized_at` values, and this write touches `updated_at`.

The forum UI infers "edited" purely from the gap between `createdAt` and `updatedAt`. Unverified posts often sit in the queue for hours or days, so verification basically always clears that threshold. Every post the user made before verification is then seen as edited at the moment of approval.

Readers see an edit that never happened.

### Acceptance criteria

- Authorizing a user's posts does not change `updated_at` on those posts.
- A post that was never edited shows no "last edited" label after its author is authorized, regardless of how long it sat as unauthorized.
- A post that was genuinely edited before authorization keeps its real `updated_at` and still shows the edited label.
- Existing posts whose `updated_at` was corrupted by a past authorization are corrected.
- Authorization still updates `is_authorized` and `authorized_at`.

### Open questions

- Is a "was this really edited" signal better sourced from something explicit than from the `updated_at` gap heuristic? `edited_by_id` already exists - maybe we should fix this by adding an `edited_at` or by leveraging `edited_by_id`.

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5096.

Adds and backfills a `forum_topic_comments.edited_at` column. `updated_at` is no longer given an implied/overloaded meaning as "edited". Something having been edited is tracked in the new column.


database/migrations/2026_08_10_000000_add_edited_at_to_forum_topic_comments_table.php
```diff
@@ -0,0 +1,107 @@
...
+            ->pluck('forum_verified_at', 'id')
+            ->all();
+    }
+
+    private function looksEdited(object $comment, ?string $forumVerifiedAt): bool
+    {
+        $createdAt = Carbon::parse($comment->created_at);
+        $updatedAt = Carbon::parse($comment->updated_at);
+
+        if ($updatedAt->diffInSeconds($createdAt, absolute: true) < 120) {
```
Jamiras: Are you sure about this logic? About 25% of my edits wouldn't make it past this threshold. I have a tendency to notice typos or things that don't read well immediately after posting and quickly fix them. This happens less with the Preview functionality, but it still happens.

Jamiras: I see that it's intentional:
https://github.com/RetroAchievements/RAWeb/blob/294dbbf2a51114d46154741bbf2a7a7781335ffa/resources/js/features/forums/components/ForumPostCard/ForumPostCardMeta/ForumPostCardTimestamps.tsx#L41-L48

But since the UI is still filtering out these values, should we try to fill them?

Jamiras (approved):

database/migrations/2026_08_10_000000_add_edited_at_to_forum_topic_comments_table.php
wescopeland: Went ahead and dropped the threshold in the migration. I also found ~28k rows that were bumped by a previously-executed Artisan command, and those are now excluded too.

PR #5162, opened 2026-08-11
feat(forums): add quote/citation support to forum posts

Ke-abr:
Adds a "Quote" button to each forum post. Clicking it drops a citation into the reply:
- a link back to the original post
- the author's name
- the quoted body
 
Built entirely from the existing `[url]`/`[b]`/`[quote]` shortcodes.

Quotes chains can balloon fast, so nested [quote] blocks collapse behind a toggle by default, leaving only the most recent quote expanded.

Quote :
Capture d&#39;écran 2026-08-11 173600

Quote collapsed :

Capture d&#39;écran 2026-08-11 173729

Quote expanded : 

Capture d&#39;écran 2026-08-11 173753

If you click on the username from the quote it links to the quoted post.

## AI disclosure
Claude Code was used for this change: it wrote the component, hook wiring, and tests. 
I reviewed every line, tested the flow manually in the browser (quoting a post, quoting a reply that itself contains a quote to check the collapse behavior, and quoting with an in-progress draft to confirm it appends instead of replacing), and ran the full test suite before pushing.

### DIFF
--- a/lang/en_US.json
+++ b/lang/en_US.json
@@ -729,6 +729,7 @@
     "<1>last</1> edited": "<1>last</1> edited",
     "Joined {{joinDate}}": "Joined {{joinDate}}",
     "Edit": "Edit",
+    "Quote post": "Quote post",
     "Copy post link": "Copy post link",
     "Copied!": "Copied!",
     "Authorize": "Authorize",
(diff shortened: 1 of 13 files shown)


wescopeland:
Thank you for the AI disclosure. This works, but the implementation has major issues. [I don't think I'm prepared to accept Claude's implementation of this feature](https://github.com/RetroAchievements/RAWeb/blob/master/docs/AI_POLICY.md#enforcement), and unfortunately I need to close this PR.

These are a few of the things I noticed:
1. The quote-injection plumbing is modeling an event as state, threads it as a prop, and then fights the consequences in a `useEffect`.
2. `canQuote` + `onQuote` is a two-prop contract for a single capability.
3. `ShortcodeQuote` and `ShortcodeSpoiler` contain blocks of code that are near verbatim clones.

PR #5167, opened 2026-08-14
feat(forums): hide posts from blocked users

### LINKED ISSUE #5165: Expand blocked user functionality to the site forum
### User Story

As someone who reads the forum,
I want posts from users I've blocked to be hidden from me,
so that I can read threads without seeing any content from people I want no interaction with.

### Why it matters

The forum has no per-user muting. If a user finds another user's posts to be hostile or exhausting, their only options are to stop reading the thread or file a moderation report. Reports are only really right for stuff that breaks the User CoC. Managing micro-aggressions between two users is not the mod team's job. A block function will let people solve it themselves.

This impacts the forum only. This issue does not impact comments.

### Acceptance criteria

* Posts by a blocked user are collapsed or hidden in forum threads. We should also include something like "Show post from [user]" so context is recoverable.
* Blocked users' threads and posts do not appear in the blocker's Recent Forum Activity anywhere on the site.
* Team members' official posts are not maskable by a block.
* Blocking does not affect what moderators see.

### Open questions

* What happens with quotes and replies referencing a blocked user's post?

wescopeland:
Resolves https://github.com/RetroAchievements/RAWeb/issues/5165.

This PR brings more intelligence to the user blocking feature to the forum. To mitigate scope, comments are unaffected, but should also be encompassed by this change in a subsequent PR.

Given User A has blocked User B, they will no longer see User B's forum posts. They will instead see a marker specifying that something has been hidden, but nothing else. The codebase considers these hidden posts to be "masked". If User A is a moderator, they see a "Show post" button on the row, allowing them to see the post content.

Screenshot 2026-08-14 at 8 13 08 AM

Behavior changes can be observed on five surfaces: the forum topic page, the forum index, the recent posts page, the home page recent posts component, and the forum sidebar.

Blocking a team account does not mask their posts anywhere. If User B opens a topic, User A will not see their topic anywhere on the site unless they land sideways on the topic's page via a direct URL.

Unblocked users may quote content from a blocked user's posts. This is a known gap, and we make no effort to conceal this.


app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
```diff
@@ -63,22 +85,116 @@ public function execute(
...
+            ->select(['forum_topic_comments.id', 'forum_topic_comments.forum_topic_id'])
+            ->join('forum_topics', 'forum_topics.id', '=', 'forum_topic_comments.forum_topic_id')
+            ->where('forum_topic_comments.is_authorized', 1)
+            ->whereNotIn('forum_topic_comments.author_id', $maskedAuthorIds)
+            ->whereNotIn('forum_topics.author_id', $maskedAuthorIds)
+            ->where('forum_topics.required_permissions', '<=', $permissions)
+            ->whereNull('forum_topics.deleted_at')
+            ->orderByDesc('forum_topic_comments.created_at')
+            ->orderByDesc('forum_topic_comments.id')
+            ->limit(self::MASKED_SCAN_WINDOW)
```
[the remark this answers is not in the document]
wescopeland: Are you sure about your claim of broken pagination? The total and items come from the same deduped map, so the paginator should be internally consistent. This seems like a P3 at best, and my sense is it's working as designed.

app/Community/Actions/BuildThinRecentForumPostsDataAction.php
```diff
@@ -12,22 +12,28 @@
...
     ): Collection {
         $userClause = $this->buildUserClause($fromAuthorId, $permissions);
 
         $subQuery = DB::table('forum_topic_comments as ftc')
             ->select('*')
             ->whereRaw($userClause)
+            ->whereNotIn('ftc.author_id', $maskedAuthorIds)
             ->orderBy('ftc.created_at', 'desc')
-            ->limit($limit + 20); // cater for spam messages
+            ->limit($limit + 20 + (empty($maskedAuthorIds) ? 0 : self::MASKED_AUTHOR_SCAN_BUFFER)); // cater for spam messages
```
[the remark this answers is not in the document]
wescopeland: Probably also not a P1. I'm pretty sure this means the underfill would need something like 270 of the newest posts to sit inside topics started by users this one particular person blocked.

wescopeland:
> When a blocked user writes the newest reply, these clauses order and paginate by that hidden activity before its timestamp is replaced, displacing genuinely recent topics across pages.

My sense is this is overstated at best. A topic bumped by a masked reply will keep its bumped position, and the displayed latest comment is going to show an older date. The sort may look slightly out of order for that person, but this is just a cosmetic inconsistency. Worse, if we resort stuff, I'm pretty sure that could truly desync pagination. I don't think this is a P1, and I think the trade-off we're intentionally making here is measured.

app/Community/Actions/BuildAggregateRecentForumPostsDataAction.php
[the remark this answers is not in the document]
wescopeland: No thank you.

Chew:
I would personally argue that a "Show post" button should always show. Especially since you do not know who sent the post. This behavior matches other platforms (e.g. Twitter, Discord), where you just see "You blocked this person" and still have a way to show it. Though, this behavior is only really necessary on threads and not anywhere else.

wescopeland:
I sense that will be contentious. We have users specifically asking us _not_ to have that button because psychologically they'll always click through it. I want to see what kind of feedback we get by not including the button. It's a lot easier to add it later than remove it.

wescopeland:
@greptile-apps

> The PR is not safe to merge while hidden replies still control forum-topic ordering.
> 
> Forum topics are ordered and paginated using blocked users’ latest replies before displayed metadata is replaced, so hidden activity can displace visibly newer topics.

You likely missed my response to this feedback item:

My sense is this is overstated at best. A topic bumped by a masked reply will keep its bumped position, and the displayed latest comment is going to show an older date. The sort may look slightly out of order for that person, but this is just a cosmetic inconsistency. Worse, if we resort stuff, I'm pretty sure that could truly desync pagination. I don't think this is a P1, and I think the trade-off we're intentionally making here is measured.

app/Community/Components/ForumRecentActivity.php
Jamiras, line 1: Given a forum post with three recent posts and one from a blocked user, the recent forum posts page shows "3 recent posts:

Image

Should the count ignore the blocked posts?

Jamiras: Similarly, if a forum post has one recent post from a blocked user, the topic does not show up in the recent list at all.

Since the forum itself shows "Blocked post", should the link still exist?

resources/views/pages-legacy/viewforum.blade.php
Jamiras, line 1: The main forums index shows blocked users as the most recent poster to individual forums:

Image

wescopeland: Thanks, fixed in latest.

app/Community/Components/ForumRecentActivity.php
wescopeland: > Should the count ignore the blocked posts?

Yep, it should. Done in latest.

> Since the forum itself shows "Blocked post", should the link still exist?

This issue only happens when every post in the thread is from a blocked user, or if the blocked user is the person who started the topic. That's intentional, as the block should mitigate the effect of the user bumping topics into the feed you see.

resources/views/pages-legacy/viewforum.blade.php
Jamiras: I'm still seeing this.
image

But only as an admin (no moderator role). As a generic registered user it doesn't appear.

Maybe it's intentional? The user/post is being filtered out in other places such as the Forum Activity list on the right (also shown in screenshot).

wescopeland: It was intentional:
```php
$forumIndexMaskedAuthorIds = $userModel && (new ForumTopicCommentPolicy())->manage($userModel)
    ? [] // mods see everything
    : (new GetMaskedForumAuthorIdsAction())->execute($userModel);
```

But it's also inconsistent with the rest of the branch. I've removed the admin/mod visibility exemptions in latest.

Jamiras (approved):

PR #5168, opened 2026-08-15
fix(api/v2): set attribute nullability in the schema

wescopeland:
Follow-up from https://github.com/RetroAchievements/RAWeb/pull/5158.

Currently in the spec in `master`, every attribute is treated as non-nullable. We have to explicitly set the nullability of each attribute, which this PR does.


Jamiras (approved):
Spot checked a couple data types. Did not thoroughly review.

PR #5169, opened 2026-08-15
fix(api/v2): discriminate between 'integer' and 'number'

wescopeland:
Stacked on https://github.com/RetroAchievements/RAWeb/pull/5168.

---

In `master`, whole number fields are currently being typed as "number", which is float-compatible. These should be "integer" instead. This PR swaps 71 of 75 `Number::make()` fields to `Integer::make()`. Additionally, a few casts are set on `PlayerAchievementSet` so it emits percentage types correctly.


Jamiras (approved):
Spot checked several objects. Did not thoroughly review.

PR #5180, opened 2026-08-19
perf(profile): eliminate unnecessary model hydration on profile page

Jamiras:
Attempt to fix out-of-memory errors seen in log

> Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes)

Reduces number of models loaded by 90%, and improves page load time by 40%:
image

image

For another user: models down by 75%, page load time down by 30%:
image

image

Reduction in page load times might be exaggerated. I tried to load each page and refresh it on both branches to eliminate query caching.

The debugbar memory usage did not seem to noticeably change. I'm not sure exactly what it's capturing.


app/Helpers/database/player-game.php
```diff
@@ -121,7 +131,14 @@ function getUserProgress(User $user, array $gameIDs, int $numRecentAchievements
...
             ->select(
-                'achievements.*',
+                'achievements.id',
+                'achievements.title',
+                'achievements.description',
+                'achievements.points',
+                'achievements.type',
+                'achievements.image_name',
+                'achievements.game_id',
+                'achievements.order_column',
```
Jamiras, line 149: I believe this to be the most important change.

Fetching `achievements.*` pulls in the trigger definitions, and thousands of definitions would take an unnecessarily large amount of memory - particularly if the achievements are complicated.

app/Helpers/database/player-game.php
```diff
@@ -59,12 +61,20 @@ function getUserProgress(User $user, array $gameIDs, int $numRecentAchievements
     $unlockedAchievements = [];
     $lockedAchievements = [];
 
-    $games = Game::with('system')->whereIn('id', $gameIDs)->get()->keyBy('id');
+    $games = Game::whereIn('id', $gameIDs)->toBase()->get()->keyBy('id');
```
wescopeland, line 64: I wouldn't be surprised if this change causes media-restricted games to leak their attached media through the V1 Web API, specifically `LastGame` in API_GetUserSummary.

`toBase()` returns `stdClass` entities, and accessors are a feature of hydrated models.

I lean towards us patching this in the API endpoint rather than gutting the optimization here, something like:
```php
'ImageTitle' => $game->is_media_restricted
    ? Game::PLACEHOLDER_IMAGE_PATH
    : ($game->image_title_asset_path ?? Game::PLACEHOLDER_IMAGE_PATH),
// ... etc ...
```

Jamiras: Indeed it does.

I've added a unit test and modified the code to handle it. I've put the logic as close to the query as possible to ensure it's probably handled for all consumers.

app/Helpers/database/player-game.php
Jamiras: Something similar could be done for games as the rich presence may be large, but there's a lot less of those records being loaded and there's around 15 fields that would need to be specified.

wescopeland (approved):

PR #5181, opened 2026-08-19
feat(tickets): add a basic (barren) React tickets list page

wescopeland:
This PR builds the initial React tickets index page. There is a _lot_ missing from this page - intentionally - to narrow this branch's scope. Notably, there are no UI controls of any kind on the page. Mobile is also lacking. This branch's changes are foundational work for the UI, just to get something workable on the screen that can be built on top of.

http://localhost:64000/tickets2
http://localhost:64000/tickets2?filter[status]=all
http://localhost:64000/tickets2?page[number]=2

The UI here already has slight differences from our conventional datatables, even though it very much is a datatable:
* Rows themselves are clickable, leading to the ticket details page.
* Only the ticket ID has a link color, even though the other entities (achievement, leaderboard, game, users) are clickable too. This is to maintain the data hierarchy on the page - if everything gets a link color the hierarchy gets extremely muddied.
* Based on the screen width, text from one column may bleed into text on another column. This is a known issue and will be fixed in a subsequent PR.

Status is represented with an a11y-aware Linear-like glyph:

Screenshot 2026-08-19 at 5 35 43 PM
Screenshot 2026-08-19 at 5 36 10 PM
Screenshot 2026-08-19 at 5 34 59 PM
Screenshot 2026-08-19 at 5 35 26 PM

Eventually, I'd like to see these statuses not only apply to tickets which a dev needs to jump on, but also general "issues" like UWC reports, etc. Eventually, the "Ticket Manager" may become an "Issue Manager", and users will have an "inbox" that shows them the current status of all their issues.

Screenshot 2026-08-19 at 5 29 43 PM


app/Platform/Controllers/TicketController.php
```diff
@@ -25,12 +28,21 @@ public function resourceName(): string
...
-    public function index(): View
+    public function index(TicketListRequest $request): InertiaResponse
     {
-        $this->authorize('viewAny', $this->resourceClass());
+        $this->authorize('viewAny', Ticket::class);
 
-        return view('resource.index')
-            ->with('resource', $this->resourceName());
+        $action = new BuildTicketListAction();
+        $result = $action->execute(TicketListScope::All, null, $request);
```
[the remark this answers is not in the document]
wescopeland: Even though system is nullable, it shouldn't be. Every game has a system. This is a non-issue - commit this to memory.

resources/js/features/tickets/utils/getTicketStateLabel.ts
Jamiras, line 1: Suggest using a different color for open - none of the other symbols use the link color.

Image

Maybe green? Possibly changing to orange if it's been open for more than some threshold?

Jamiras (approved):
It's hard to tell whether functionality is intentionally or unintentionally missing (like tooltips).

The basic grid functionality and links all seem to work. I couldn't figure out how to make the status grouping headers appear.

Did not look at code.

resources/js/features/tickets/utils/getTicketStateLabel.ts
wescopeland: Green is tough because that's the color of the Resolved glyph. I went with `neutral-400` for Open and `neutral-600` for Closed in latest. I think this probably makes sense because the most common color should lean towards being something that's quiet. The less-common things can get a splash of color so they stand out.

Screenshot 2026-08-20 at 5 28 12 PM

Jamiras (approved):

PR #5183, opened 2026-08-22
feat(tickets): add desktop filter controls to tickets2

wescopeland:
This PR adds a Linear-like filters bar to the React ticket manager. Mobile not playing nicely with the filters bar is a known issue and will be addressed in a subsequent PR.

A single button opens a menu of filter options, each with its own list of possible filter values. Once a filter is set, it appears in the filter bar. Every option (except emulators) shows a computed value of how many tickets it will match.

Also includes some performance optimizations to bring down the main query from ~200ms to ~50ms.

https://github.com/user-attachments/assets/5597cbb7-17e5-448f-a47c-34b66199035f


resources/js/features/tickets/components/TicketListFilterControl/TicketListFilterControl.tsx
Jamiras, line 1: > Every option (except emulators) shows a computed value of how many tickets it will match.

It looks like that is also showing a computed value.

resources/js/features/tickets/components/TicketListFilterControl/TicketListFilterControl.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <BaseButton
+          size="sm"
+          aria-label={t('Filter')}
+          className={cn(
+            'border-dashed bg-neutral-950 light:bg-white',
+            buildTrackingClassNames('Click Ticket Filter'),
+          )}
+          data-testid="add-filter"
+        >
+          <RxPlusCircled className={cn('size-4', isLabelHidden ? null : 'mr-2')} />
```
Jamiras, line 61: It seems a little unintuitive that existing filters show up under the "+" icon. But the behavior is correct - if a different filter value is selected, the existing filter value is replaced

image

Jamiras (commented):
Seems functional. Did not review code.

resources/js/features/tickets/components/TicketListFilterControl/TicketListFilterControl.tsx
wescopeland: Yea, this is exactly what Linear does. I _believe_ the intention is to mitigate any visual shift in the menu, so the user can build/use muscle memory regardless of whatever options they have set.

I've taken a creative liberty and added a subtle "active dot" in latest:
Screenshot 2026-08-26 at 7 15 12 PM

resources/js/features/tickets/components/TicketListFilterControl/TicketListFilterControl.tsx
wescopeland: So it is :)

I drafted the PR description before squeezing that in. My original implementation of the emulator ticket counts resulted in a pretty expensive query, but I was able to optimize the implementation and forgot to update the PR description accordingly.

Jamiras (approved):

PR #5184, opened 2026-08-23
feat(tickets): add display panel control to tickets2

wescopeland:
This PR adds a Linear-like "display panel" control to the tickets2 page:

Screenshot 2026-08-23 at 3 48 10 PM
Screenshot 2026-08-23 at 3 48 15 PM

This single button allows the user to control which columns are visible, as well as what the active sort order is (and whether it's ascending or descending). Additionally, the user's view preference is now persisted like what we do in our other datatables.


resources/js/features/tickets/hooks/useTicketListColumnDefinitions.ts
```diff
@@ -1,19 +1,55 @@
...
+  const hardcoreLabel = t('Hardcore');
+  const casualLabel = t('Casual');
+
   return [
     buildIdColumnDef({ t_label: t('ID') }),
     buildTicketableColumnDef({ t_label: t('Issue with') }),
+    buildReportColumnDef({
+      t_label: t('Report'),
+      ticketTypeLabels,
+    }),
```
Jamiras: I'm not sure how useful it will be to have this. It's free-form data, and most of them are similar to "I did [achievement requirements] and didn't get the unlock"

resources/js/features/tickets/hooks/useTicketListColumnDefinitions.ts
```diff
@@ -25,6 +61,34 @@ export function useTicketListColumnDefinitions(): ColumnDef<App.Platform.Data.Ti
...
+      id: 'core',
+      t_label: t('Core'),
+      getText: (entry) => entry.emulatorCore,
+    }),
+    buildTicketMetadataColumnDef({
+      id: 'hash',
+      t_label: t('Hash'),
+      getText: (entry) => entry.gameHash?.name ?? entry.gameHash?.md5 ?? null,
+      widthClassName: 'min-w-[8em] flex-[2_1_0]',
+    }),
```
Jamiras: I like the idea of being able to see which version of the game the player was using, but in practicality, unless all the other fields are hidden, the differentiator gets truncated. "Game Title: Subtitle (U)" and "Game Title: Subtitle (J)" are indistinguishable if all you see if "Game Title: Su...".

Jamiras (approved):
Seems functional. A couple of minor comments and the practicality of couple of fields.

Did not review code.

resources/js/features/tickets/components/TicketListDisplayPanel/TicketListDisplayPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  let directionLabel = isAscending ? t('Oldest first') : t('Newest first');
+  if (sortField === 'state') {
+    directionLabel = isAscending ? t('Ascending') : t('Descending');
+  }
+
+  const sortFieldLabels = {
+    createdAt: t('Created'),
+    state: t('Status'),
+    resolvedAt: t('Resolution date'),
```
Jamiras: "Resolution date" wraps when selected

Image

resources/js/features/tickets/hooks/useTicketListColumnDefinitions.ts
wescopeland: You've worked on far more tickets than I have; I trust your judgment. I assumed this may be helpful for cases where someone culling tickets (ie: for a specific set of niche filter options) may want to scan the list and see if ticket descriptions were actually useful. Do you think we should keep or cut this column?

resources/js/features/tickets/hooks/useTicketListColumnDefinitions.ts
wescopeland: Good point, I'm laser-focused on this here:
> I like the idea of being able to see which version of the game the player was using

Being bold: in latest, we now attempt to strip the leading title from the hash label. If the title has no trailing tags, we show the first 8 characters of the hash. We also now pop a tooltip.

Screenshot 2026-08-26 at 7 23 05 PM

resources/js/features/tickets/components/TicketListDisplayPanel/TicketListDisplayPanel.tsx
wescopeland: Changed this string in latest:
Screenshot 2026-08-26 at 7 27 51 PM

resources/js/features/tickets/hooks/useTicketListColumnDefinitions.ts
Jamiras: In my opinion, most culling will occur based on other attributes (like the emulator used, or even the user who reported the issue). I don't think there will be any decisions made from this information:

Image

wescopeland: I went ahead and removed the 'Report' column in latest.

Jamiras (approved):

PR #5193, opened 2026-08-30
feat(connect): support leaderboard.state

Jamiras:
Allows the DLL to promote and demote leaderboards.

For backwards compatibility, if a state is not sent, new leaderboards will be created as active.

NOTE: this re-enables the logic that requires a claim to create new leaderboards.
https://discord.com/channels/310192285306454017/310195377993416714/1448435888433529005


app/Platform/Listeners/EnsureTriggerVersionedOnPromotion.php
```diff
@@ -7,21 +7,33 @@
...
 use RuntimeException;
 use Spatie\Activitylog\CauserResolver;
 
 class EnsureTriggerVersionedOnPromotion
 {
-    public function handle(AchievementPromoted $event): void
+    public function handle(AchievementPromoted|LeaderboardPromoted $event): void
     {
-        $achievement = $event->achievement;
+        /* Model $model */
```
wescopeland: nit: This comment is benign and can be removed.

app/Models/Leaderboard.php
```diff
@@ -103,6 +105,14 @@ protected static function boot(): void
                     $service->clearForGameId((int) $originalGameId);
                 }
             }
+
+            if ($leaderboard->wasChanged('state')) {
+                if ($leaderboard->state === LeaderboardState::Unpromoted) {
+                    LeaderboardUnpromoted::dispatch($leaderboard);
+                } elseif ($leaderboard->getOriginal('state') === LeaderboardState::Unpromoted) {
+                    LeaderboardPromoted::dispatch($leaderboard);
```
wescopeland, line 113: `LeaderboardPromoted` will dispatch on a transition from Unpromoted to Disabled.

wescopeland: Maybe that's actually okay. Technically, I guess it _is_ promoted, from a certain point of view.

wescopeland (approved):

app/Platform/Listeners/EnsureTriggerVersionedOnPromotion.php
Jamiras: Leftover attempt at trying to appease a `composer analyse` error.

app/Models/Leaderboard.php
Jamiras: Intended. If a leaderboard goes straight from unpromoted to disabled, we still want to version the trigger.

PR #5199, opened 2026-09-04
fix: correct game forum resource URL encoding (#3339)

### LINKED ISSUE #3339: Auto generated resources URL in game forum dropping part of the search if there's an apostrophe

adamjohnny5:
Fixes URL issues in official game forum topics involving titles with apostrophes or other reserved characters. The changes replace manual escaping with proper query encoding, add regression coverage for special characters, and ensure external search URLs preserve the full game title.

Testing included targeted backend coverage for the affected flow, `composer fix`, `composer analyse`, and manual review of the generated resource URLs.


app/Helpers/database/forum.php
wescopeland: Docs: This comment was referencing the bug addressed by this changeset and is now safe to remove.

app/Helpers/database/forum.php
```diff
@@ -386,13 +386,18 @@ function generateGameForumTopic(User $user, int $gameId): ?ForumTopicComment
...
-    $wikipediaURL = "https://www.google.com/search?q=site:en.wikipedia.org+$urlSafeGameTitle";
+    $gameFAQsURL = 'https://www.google.com/search?' . http_build_query([
+        'q' => "site:www.gamefaqs.com $searchTitle",
+    ]);
+    $longplaysURL = 'https://www.google.com/search?' . http_build_query([
+        'q' => "site:www.youtube.com longplay $searchTitle",
+    ]);
+    $wikipediaURL = 'https://www.google.com/search?' . http_build_query([
+        'q' => "site:en.wikipedia.org $searchTitle",
+    ]);
```
wescopeland: DX, minor: a closure may be more DRY:

```php
$searchUrl = fn (string $query): string => 'https://www.google.com/search?' . http_build_query(['q' => $query]);

$hashesUrl = ...;
$gameFaqsUrl = $searchUrl("site:www.gamefaqs.com $searchTitle");
$longplaysUrl = $searchUrl(...);
$wikipediaUrl = $searchUrl(...);
```

tests/Feature/Community/Controllers/AchievementSetClaimControllerTest.php
```diff
@@ -32,6 +33,38 @@ class AchievementSetClaimControllerTest extends TestCase
...
+        );
+        $this->assertStringContainsString(
+            '[url=https://www.google.com/search?q=site%3Awww.youtube.com+longplay+Kirby%27s+Adventure+%26+More%3F+NES]Longplay[/url]',
+            $comment->body,
+        );
+        $this->assertStringContainsString(
+            '[url=https://www.google.com/search?q=site%3Aen.wikipedia.org+Kirby%27s+Adventure+%26+More%3F+NES]Wikipedia[/url]',
+            $comment->body,
+        );
+    }
```
wescopeland: Testing: This is a rare circumstance where I think we should delete this test case. Two reasons behind my thinking:

1. We're testing a legacy helper rather than the controller this suite belongs to.
2. This test case has a hidden coupling with the `$forumId` matcher in _app/Helpers/database/forum.php_, and this is undocumented in the test case itself. I don't think it's worth documenting or remediating, because the matcher itself should also be eliminated eventually.

app/Helpers/database/forum.php
adamjohnny5: Forgot about that line. Its gone.

app/Helpers/database/forum.php
adamjohnny5: Done as requested,.

tests/Feature/Community/Controllers/AchievementSetClaimControllerTest.php
adamjohnny5: Test has been removed.

app/Helpers/database/forum.php
```diff
@@ -386,19 +386,19 @@ function generateGameForumTopic(User $user, int $gameId): ?ForumTopicComment
...
+
+    $searchUrl = fn (string $query): string => 'https://www.google.com/search?' . http_build_query(['q' => $query]);
 
     $hashesURL = route('game.hashes.index', ['game' => $gameId]);
-    $gameFAQsURL = "https://www.google.com/search?q=site:www.gamefaqs.com+$urlSafeGameTitle";
-    $longplaysURL = "https://www.google.com/search?q=site:www.youtube.com+longplay+$urlSafeGameTitle";
-    $wikipediaURL = "https://www.google.com/search?q=site:en.wikipedia.org+$urlSafeGameTitle";
+    $gameFAQsURL = $searchUrl("site:www.gamefaqs.com $searchTitle");
+    $longplaysURL = $searchUrl("site:www.youtube.com longplay $searchTitle");
+    $wikipediaURL = $searchUrl("site:en.wikipedia.org $searchTitle");
```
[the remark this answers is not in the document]
wescopeland: That was intentional.

wescopeland (approved):
Thanks!

PR #5203, opened 2026-09-08
feat(notes): add paginator to code notes page

Jamiras:
Fixes out of memory error accessing http://localhost:64000/codenotes.php?g=2621

https://retroachievements.org/log-viewer?file=5a6da50a-laravel-2026-09-04.log&query=log-index%3A33


resources/views/pages-legacy/codenotes.blade.php
```diff
@@ -16,13 +16,24 @@
...
+{
+    return MemoryNote::query()
+        ->where('game_id', $gameId)
+        ->whereNot('body', '')
+        ->whereNot('body', "''")
+        ->select(['address', 'body', 'user_id'])
+        ->orderBy('address');
+};
+
+$codeNoteCount = $baseMemoryNoteQuery($gameID)->count();
```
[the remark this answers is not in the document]
Jamiras: $codeNoteCount is now updated in the subset branch.

resources/views/pages-legacy/codenotes.blade.php
```diff
@@ -403,8 +416,136 @@ function saveCodeNote(rowIndex, isDeleting = false) {
...
+        echo "<tbody>";
+
+        $subsetNote = null;
+        $rowIndex = 0;
+        foreach ($codeNotes as $nextCodeNote) {
+            if ($hasSubsetNotes) {
+                $subsetNote = $subsetNotes->where('address', $nextCodeNote->address)->first();
+            }
+
+            $trimmedNote = trim($nextCodeNote->body);
```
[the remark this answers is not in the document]
Jamiras: I would argue that this is a good thing and we should do it when submitting new notes as well.

resources/views/pages-legacy/codenotes.blade.php
```diff
@@ -40,48 +51,47 @@
...
+        ->offset($offset)
+        ->get();
+
+    $subsetNotes = $baseMemoryNoteQuery($gameID)
+        ->toBase()
+        ->get();
+
+    // make sure dummy notes exist for all addresses defined in the subset
+    foreach ($subsetNotes as $note) {
+        $baseNote = $codeNotes->where('address', $note->address)->first();
```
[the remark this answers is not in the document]
Jamiras: This was attempting to handle an edge case where a subset note exists between pages of the base set by loading all the notes for the current page, but that will create the illusion of the off-page notes not existing in the base set. I've modified the logic to filter to match the current page, assuming the likelihood of losing the subset note is minimal.

resources/views/pages-legacy/codenotes.blade.php
```diff
@@ -40,48 +51,47 @@
...
+        if (!$baseNote) {
+            $codeNotes->push((object) ['address' => $note->address, 'body' => '', 'user_id' => $note->user_id]);
```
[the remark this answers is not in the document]
Jamiras: I needed something as a placeholder. I've updated the logic to use 0 and hide the avatar.

resources/views/pages-legacy/codenotes.blade.php
```diff
@@ -40,48 +52,59 @@
...
-
-                $found = true;
-                break;
+    $codeNoteCount = $baseMemoryNoteQuery($baseGameId)->count();
+    $codeNotes = $baseMemoryNoteQuery($baseGameId)
+        ->limit($perPage)
+        ->offset($offset)
+        ->get();
+
+    if ($codeNotes->empty()) {
```
wescopeland: This disables all subset merging.

http://localhost:64000/codenotes.php?g=25606

Prod:

Image

Local:

Image

resources/views/pages-legacy/codenotes.blade.php
wescopeland, line 1: Something strange is happening on http://localhost:64000/codenotes.php?g=17923. The page says there are 5 code notes, but the table has over 200 rows.

resources/views/pages-legacy/codenotes.blade.php
Jamiras: Changed to `empty($codeNotes)`. Why do they differ?

resources/views/pages-legacy/codenotes.blade.php
Jamiras: There are only 5 notes for the base game (http://localhost:64000/codenotes.php?g=11202). This is a result of me recalculating `$codeNodeCount` per https://github.com/RetroAchievements/RAWeb/pull/5203#discussion_r3959136347 so the pagination happens appropriately.

While not intuitive, the message is correct. The game has 5 notes and there are 200+ subsets notes that need merging.

resources/views/pages-legacy/codenotes.blade.php
Jamiras: Related: I noticed on this branch (and also tested on master). When clicking the keep buttons, they don't get hidden, while the note text and avatar do.

https://github.com/RetroAchievements/RAWeb/blob/5fda24c34e12243335917d55e08f182f060b0b0d/resources/views/pages-legacy/codenotes.blade.php#L193-L197

wescopeland: `collect([1, 2, 3])->empty()` creates and returns an empty collection. In this sequence, `->empty()` is a factory.

`empty(collect([1, 2, 3]))` checks if a given var is falsy. `collect([1, 2, 3])` will never be falsy. This causes the current code to unfortunately break in a different way. Check http://localhost:64000/codenotes.php?g=25606&o=5000.

To confuse matters even more, the fix we probably want to reach for is yet another expression of emptiness:
```php
if ($codeNotes->isEmpty()) {
```

Jamiras: Updated. Also moved in a level as there's potentially a case where only subset notes exist (and no core notes).

wescopeland (approved):

PR #5205, opened 2026-09-08
feat(events): auto-sync AOTW Discord winner roles

wescopeland:
This PR allows the web server to sync the AOTW Winner role by reusing Discord API mechanisms we already have working for managing roles for muted users and handling display name changes.

A migration is included in this PR, which creates a `event_winner_discord_role_grants` table. This table is agnostic to the particular event in question, so it could in theory be reused for different events one day.

The new table has an `expires_at` timestamp field. When that date is surpassed, the role is wiped from all associated users. We cannot make a call to Discord's API that naively says "wipe Role X from all users in our server". This is part of the server intents that Discord no longer allows us to have. Instead, we need to use a lookup table and keep things in sync.

When this change goes live, we need to manually remove the AOTW Winner role from everyone on the server so the lookup table can be properly synchronized. We also need to strongly encourage people not to manually assign the role, because it won't be auto-removed when appropriate (as it wont exist in our lookup table).

I've updated the server environments so they have the role ID ready to go.


app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
```diff
@@ -0,0 +1,111 @@
...
+
+        foreach ($winners as $user) {
+            // A failed revoke must finish before this user can receive another grant.
+            if ($grants->contains('user_id', $user->id)) {
+                continue;
+            }
+
+            try {
+                $member = $this->findDiscordMemberAction->execute($user->display_name);
+                $discordUserId = $member['user']['id'] ?? null;
```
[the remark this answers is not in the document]
wescopeland: Unverified users can't change their nicknames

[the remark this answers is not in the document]
wescopeland: The net impact of this is zero. Someone who shouldn't have the role gets the role, and then they're either banned from the server or cast a vote they shouldn't have.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        $current = EventAchievement::currentAchievementOfTheWeek()->first();
+        $winners = User::query()
+            ->whereNull('banned_at')
+            ->whereNull('unranked_at')
+            ->where(fn (Builder $query) => $query->whereNull('muted_until')->orWhere('muted_until', '<=', now()))
+            ->whereHas('playerAchievements', fn (Builder $query) => $query
+                ->where('achievement_id', $current?->achievement_id)
+                ->whereNotNull('unlocked_hardcore_at'))
```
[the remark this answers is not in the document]
wescopeland: Not something I'm worried about for AOTW, but it's a small fix and could potentially apply to other events, so I've fixed this in latest.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
```diff
@@ -0,0 +1,111 @@
...
+            ->get()
+            ->keyBy('id');
+
+        $grants = EventWinnerDiscordRoleGrant::with('user')->where('discord_role_id', $roleId)->get();
+
+        // Revoke expired grants before assigning roles for the new week.
+        foreach ($grants as $key => $grant) {
+            if ($current && $grant->expires_at->equalTo($current->active_until) && $winners->has($grant->user_id)) {
```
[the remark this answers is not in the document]
wescopeland: Valid - I'd prefer to mitigate unnecessary calls to Discord. Fixed in latest.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
```diff
@@ -0,0 +1,115 @@
...
+
+        $grants = EventWinnerDiscordRoleGrant::with('user')->where('discord_role_id', $roleId)->get();
+
+        // Revoke expired grants before assigning roles for the new week.
+        foreach ($grants as $key => $grant) {
+            if ($current && $grant->expires_at->isFuture() && $winners->has($grant->user_id)) {
+                $grant->update(['expires_at' => $current->active_until]);
+
+                continue;
+            }
```
Jamiras, line 69: What is this trying to do? I assumed it was trying to extend the grant if the user earned the new AotW within the first few minutes of the last one expiring, but the `expires_at->isFuture()` would be false for that case.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
```diff
@@ -0,0 +1,115 @@
...
+
+            try {
+                if ($grant->discord_user_id !== null) {
+                    $wasRevoked = $this->setDiscordMemberRoleAction->execute(
+                        $grant->discord_user_id,
+                        $roleId,
+                        shouldHaveRole: false,
+                        reason: "AOTW winner removed: {$grant->user->display_name}",
+                    );
```
Jamiras, line 78: If the user does earn the new AotW within a few minutes of the old one expiring, the role will be revoked here, then immediately restored when the `$winners` loop is processed.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
```diff
@@ -0,0 +1,115 @@
...
+                        continue;
+                    }
+                }
+
+                $grants->push(EventWinnerDiscordRoleGrant::create([
+                    'user_id' => $user->id,
+                    'discord_role_id' => $roleId,
+                    'discord_user_id' => $discordUserId,
+                    'expires_at' => $current->active_until,
+                ]));
```
Jamiras, line 128: The old bot would also send a message to the AotW channel:

Image

app/Community/AppServiceProvider.php
```diff
@@ -61,6 +63,7 @@ public function boot(): void
             /** @var Schedule $schedule */
             $schedule = $this->app->make(Schedule::class);
 
+            $schedule->command(SyncAotwWinnerDiscordRoles::class)->everyFifteenMinutes()->withoutOverlapping();
```
Jamiras, line 66: With this directly integrated in the website, I was expecting the unlock to immediately try to award the role. Could `UnlockPlayerAchievementAction` also kick off this command, or a similar command specific to the user who got the unlock?

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
wescopeland: The `$grant->update()` is there to handle the off-chance that an end date for an AOTW event achievement changes. This may be overly-defensive.

The `continue` within the conditional is there so we don't constantly remove and re-add roles a user should have during each sync attempt.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
wescopeland: That's true. I wasn't sure if that was worth fixing - it seems unlikely to happen often and I wasn't sure if it was worth the complexity trade-off to address.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
wescopeland: I mistakenly believed we couldn't do this without a privileged intent that was revoked from us, but it turns out that's not true. I've added this behavior in latest.

app/Community/AppServiceProvider.php
wescopeland: We should definitely do this. To keep it simple, I added an optional `$user` arg to the action. Updated in latest.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
Jamiras: The AotW is _supposed_ to take about 2 hours to earn, but there seems to be people who always start playing before the achievement is announced, so they get the AotW shortly after it is announced. This week it only took about 20 minutes:
image

I agree that it's probably not worth the effort if that's not what the other block was trying to address. Besides, sending the notification would further complicate any attempts to be smart about recycling the row.

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
Jamiras: I got the general idea that it's an update and ignore. I just wasn't sure why.

I think the only time the AotW has ever been extended was due to a server outage, but I don't see any reason to remove the code since you've already written it.

app/Community/Listeners/SyncAotwWinnerDiscordRole.php
Jamiras, line 1: I'm guessing I'm having trouble with some service not being restarted correctly, but I can't get a manual unlock to create a row in the new table.

I've verified that the code is calling into the `handle()` function below by generating an exception that got logged, and if I modify the command to take a User as input, that works too. Going to chalk it up as a me-problem. The code looks right.

Jamiras (approved):

app/Community/Actions/SyncAotwWinnerDiscordRolesAction.php
Chew: Yeah, I've seen it happen usually at least once a week and they to ping someone to re-add it, but this is because the bot adds it before it gets around to removing it.

Another concern was people joining the server after getting the achievement, and thus don't get the role since they missed the assign window, that's a separate issue to this, of course, but unsure if that was also addressed.

app/Community/Listeners/SyncAotwWinnerDiscordRole.php
wescopeland: Do you have both new env vars in your _.env_ file? Horizon workers never resync their env vars after booting - make sure they're both there and restart Horizon.

database/migrations/2026_09_08_000000_create_event_winner_discord_role_grants_table.php
```diff
@@ -0,0 +1,37 @@
...
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    public function up(): void
+    {
+        Schema::create('event_winner_discord_role_grants', function (Blueprint $table) {
```
Jamiras: I wonder if this should just be renamed to `discord_role_grants`. I could see this being used for things other than event winners.

app/Community/Listeners/SyncAotwWinnerDiscordRole.php
Jamiras: After fully rebooting the VM, it seems to work. Not sure why killing and restarting Horizon wasn't enough.

database/migrations/2026_09_08_000000_create_event_winner_discord_role_grants_table.php
wescopeland: It does seem generic enough. Renamed it in latest.

Jamiras (approved):

PR #5213, opened 2026-09-09
chore: adjust how search exposes a field

wescopeland:


wescopeland:
This plugged a security hole. I've hotfixed this change to production.

PR #5214, opened 2026-09-09
fix(api/v2): enforce a cap on number of leaderboard entries that can be returned

wescopeland:
Currently, `?filter[maxRank]` has no maximum allowed value. This PR sets a cap at 100.

### DIFF
--- a/app/Api/V2/LeaderboardEntries/LeaderboardEntryCollectionQuery.php
+++ b/app/Api/V2/LeaderboardEntries/LeaderboardEntryCollectionQuery.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Api\V2\LeaderboardEntries;
+
+use App\Api\V2\DefaultCollectionQuery;
+
+class LeaderboardEntryCollectionQuery extends DefaultCollectionQuery
+{
+    /**
+     * Get the validation rules for the request.
+     */
+    public function rules(): array
+    {
+        return array_merge(parent::rules(), [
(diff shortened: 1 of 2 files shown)


app/Api/V2/LeaderboardEntries/LeaderboardEntryCollectionQuery.php
```diff
@@ -0,0 +1,20 @@
...
+
+class LeaderboardEntryCollectionQuery extends DefaultCollectionQuery
+{
+    /**
+     * Get the validation rules for the request.
+     */
+    public function rules(): array
+    {
+        return array_merge(parent::rules(), [
+            'filter.maxRank' => ['integer', 'min:1', 'max:100'],
```
Jamiras, line 17: Is this really what you want to be restricting? If I query all my leaderboard-entries, there are quite a few where my rank is over 100:
```
      "attributes": {
        "score": 7250,
        "formattedScore": "1:12.50",
        "rank": 370,
        "createdAt": "2019-03-02T16:30:10.000000Z",
        "updatedAt": "2019-03-02T16:30:10.000000Z"
      },
```

If you want to limit the number of returned values, you should restrict `page[size]`, which (at least for this endpoint) defaults to 50.

wescopeland: I think this is actually fine. This PR was actually intended to plug a way to pin the DB, but something was wrong with my test script and I don't think it can actually happen in practice.

PR #5215, opened 2026-09-09
feat(tickets): improve the filtered empty state

wescopeland:
This PR improves the empty state for filtered ticket list pages. Now, there's a button to remove your current filter selection. It prefetches on hover.

**Before**
Screenshot 2026-09-09 at 7 21 20 PM

**After**
Screenshot 2026-09-09 at 7 21 14 PM

### DIFF
--- a/resources/js/features/tickets/components/TicketListEmptyState/TicketListEmptyState.tsx
+++ b/resources/js/features/tickets/components/TicketListEmptyState/TicketListEmptyState.tsx
@@ -1,12 +1,52 @@
 import type { FC } from 'react';
 import { useTranslation } from 'react-i18next';
 
-export const TicketListEmptyState: FC = () => {
+import { BaseButton } from '@/common/components/+vendor/BaseButton';
+import { EmptyState } from '@/common/components/EmptyState';
+
+interface TicketListEmptyStateProps {
+  onPrefetchViewAll?: () => void;
+  onViewAll?: () => void;
+  scope?: App.Platform.Enums.TicketListScope;
+  unfilteredTotal?: number | null;
+}
+
+export const TicketListEmptyState: FC<TicketListEmptyStateProps> = ({
+  onPrefetchViewAll,
+  onViewAll,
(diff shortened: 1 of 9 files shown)


resources/js/features/tickets/components/TicketListEmptyState/TicketListEmptyState.tsx
```diff
@@ -1,12 +1,52 @@
...
+        {hasTicketHistory && onViewAll ? (
+          <BaseButton
+            variant="ghost"
+            size="sm"
+            className="mt-3 text-link"
+            onClick={onViewAll}
+            onMouseEnter={onPrefetchViewAll}
+          >
+            {t('View all tickets')}
+          </BaseButton>
```
Jamiras, line 47: > Now, there's a button to revert your filter selection to the server default options. It prefetches on hover.

This doesn't revert the filter selection, it completely clears it. For example, if I go to Manage > Tickets, the default filter is Unresolved. If I then filter it down to no results, pressing Reset goes back to Unresolved, but pressing View All Tickets goes to no filter.

This is consistent with the verbiage, but inconsistent with the described behavior.

wescopeland: I've updated the PR description to be accurate, and I've pushed a change to update a behavior oddity of the Resolved By scope.

Jamiras: Having "View All Tickets" on pre-filtered pages is still awkward.

I would expect the "no results found" message to be tied to a "reset filters" option, rather than a "remove filters" option.

wescopeland: That makes sense. Updated in latest:
Screenshot 2026-09-16 at 7 56 10 AM

Jamiras (approved):

PR #5217, opened 2026-09-10
fix(tickets): change status sort labels

wescopeland:
Changes a few tooltip labels that in hindsight I'm not too happy with.

**Before**
Screenshot 2026-09-10 at 5 34 28 PM
Screenshot 2026-09-10 at 5 34 22 PM

**After**
Screenshot 2026-09-10 at 5 34 00 PM
Screenshot 2026-09-10 at 5 34 07 PM

### DIFF
--- a/resources/js/features/tickets/components/TicketListDisplayPanel/TicketListDisplayPanel.tsx
+++ b/resources/js/features/tickets/components/TicketListDisplayPanel/TicketListDisplayPanel.tsx
@@ -60,7 +60,7 @@ export const TicketListDisplayPanel: FC<TicketListDisplayPanelProps> = ({
 
   let directionLabel = isAscending ? t('Oldest first') : t('Newest first');
   if (sortField === 'state') {
-    directionLabel = isAscending ? t('Ascending') : t('Descending');
+    directionLabel = isAscending ? t('Open first') : t('Closed first');
   }
 
   const sortFieldLabels = {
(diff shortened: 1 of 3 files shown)


resources/js/features/tickets/components/TicketListDisplayPanel/TicketListDisplayPanel.tsx
Jamiras, line 1: Sorting by Resolved ascending without an explicit filter on "status=resolved" appears to be "Newest unresolved first", not "Oldest first".

wescopeland: Thanks. I've made two adjustments in https://github.com/RetroAchievements/RAWeb/pull/5217/changes/dfbf846edeb2190b73cf1b7e3ff70d4a22f376b5:

1. When the Resolved sort is active, we now do:
```php
case TicketListSortField::ResolvedAt:
    $query->orderByRaw('resolved_at IS NULL'); // !! <-------- new
    $query->orderBy('resolved_at', $sort['direction']);
    $query->orderByDesc('created_at');
    break;
```
This is specifically to handle when status is unfiltered (set to "All"). This keeps every unresolved ticket from sorting to the top when the user picks "Oldest first".

2. When the user is filtering on a status that cannot have any resolved tickets and they sort by resolved, we now change their status filter to "Resolved". If they're already on "Resolved", "Closed", or "All", we don't change this filter option.

Jamiras (approved):

PR #5221, opened 2026-09-11
perf(tickets): add more prefetches

wescopeland:
This PR brings over the prefetching pattern from other datatables to the tickets datatable. Filter options, sort settings, and reset buttons now all prefetch their destination on hover (observable in the network tab). On a slower connection, this results in considerably improved perceived performance of the page due to prewarming the React Query cache.

Will have a small conflict with #5215's changes.

### DIFF
--- a/resources/js/features/tickets/components/+index/TicketIndexRoot.test.tsx
+++ b/resources/js/features/tickets/components/+index/TicketIndexRoot.test.tsx
@@ -431,7 +431,7 @@ describe('Component: TicketIndexRoot', () => {
     });
 
     // ACT
-    await userEvent.click(screen.getByRole('button', { name: 'Go to next page' }));
+    await userEvent.hover(screen.getByRole('button', { name: 'Go to next page' }));
 
     // ASSERT
     await waitFor(() => {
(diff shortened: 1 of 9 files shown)


Jamiras (approved):
> Filter options, sort settings, and reset buttons now all prefetch their destination on hover (observable in the network tab). 

The only filter option I see this working with is the Reset button, which is probably the only one I'd want it to work on. It shouldn't be prefetching data as I mouse over options in the menu to get to the one I actually want. It already does that for the Sort menu. After opening the menu, as I mouse down to the last option, it prefetches the other unselected one. Even if it is only 60KB, it's still using server resources that the client isn't going to use.

Which brings up a more philosophical question: Do these pages really need to be that extra level of snappy? They're relatively low-traffic, aren't they?

wescopeland:
I believe [fast is a feature](https://joeblu.com/blog/slow_software_destroys_flow/). I really like fast software, even if it's through perceived performance.

But I also agree those menu options are overeager. I've made a modification in https://github.com/RetroAchievements/RAWeb/pull/5221/commits/32b2c7c6d82b2808120f524e3010412da8ea7bab so these use a hover delay, similar to Inertia's `Link` component.

resources/js/common/hooks/useHoverIntent.ts
```diff
@@ -0,0 +1,23 @@
+import { useRef } from 'react';
+
+/**
+ * Ensure the user is being intentful with their hover over a list of
+ * items, similar to Inertia Link prefetching, so we don't overheat the
+ * query cache.
+ */
+export function useHoverIntent(delayMs = 150) {
```
Jamiras, line 8: Is 150 the InertiaLink default? In 8/10 tests, I still triggered the prefetch of the middle option trying to mouse to the third.

Jamiras (approved):

resources/js/common/hooks/useHoverIntent.ts
wescopeland: Inertia's default is even more aggressive at 75ms, faster than the blink of an eye. 150 is about as high as I'd feel comfortable raising this without tweaking the intent algorithm itself.

PR #5222, opened 2026-09-11
perf(tickets): optimize several queries

wescopeland:
Several ticket list queries on my local take 200ms+ to resolve. This PR makes the following optimizations:

* Adds an indexed virtual `state_sort_order` column to the tickets table (similar to our virtual column on `player_achievements`), that mitigates the need for us to sort in-line with a `CASE` expression.
* Adds a covering index for ticket facet counts.
* Forces the query planner to avoid creation date index scans when newest-first results fit on a single page.
* Replaces separate counts for each publish / dev type option with one grouped query per facet.

Several legacy perm conditions have been migrated to role-driven conditions. This PR also improves the responsiveness of #5221.

http://localhost:64000/tickets?filter%5Bstatus%5D=all&sort=-state

**Before**
Screenshot 2026-09-11 at 12 38 47 PM

**After**
Screenshot 2026-09-11 at 12 39 16 PM

### DIFF
--- a/app/Platform/Actions/BuildTicketListAction.php
+++ b/app/Platform/Actions/BuildTicketListAction.php
@@ -62,7 +62,15 @@ class BuildTicketListAction
         }
 
         $query = $service->applyFilters(clone $renderableTickets, $filterOptions, $comparisonUser);
-        $this->applySort($query, $request->getSort());
+        $sort = $request->getSort();
+
+        // If the result set is small, filter it first instead of scanning creation dates for a full page.
+        // This makes several queries ~20x faster.
+        if ($total <= self::PER_PAGE && $sort['field'] === TicketListSortField::CreatedAt && $sort['direction'] === 'desc') {
+            $query->ignoreIndex('tickets_created_at_index');
+        }
+
+        $this->applySort($query, $sort, TicketListStatusFilter::from($filterOptions['status']));
         $this->applyTicketListEntryEagerLoads($query);
 
         $paginator = new LengthAwarePaginator(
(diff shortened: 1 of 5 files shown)


app/Platform/Actions/BuildTicketListAction.php
```diff
@@ -63,7 +62,15 @@ public function execute(
         }
 
         $query = $service->applyFilters(clone $renderableTickets, $filterOptions, $comparisonUser);
-        $this->applySort($query, $request->getSort());
+        $sort = $request->getSort();
+
+        // If the result set is small, filter it first instead of scanning creation dates for a full page.
+        // This makes several queries ~20x faster.
+        if ($total <= self::PER_PAGE && $sort['field'] === TicketListSortField::CreatedAt && $sort['direction'] === 'desc') {
```
Jamiras: Why specifically limit this to descending?

app/Platform/Actions/BuildTicketListAction.php
```diff
@@ -108,25 +115,24 @@ public function getAvailableFilters(TicketListScope $scope, ?int $systemId = nul
...
-                );
+                // Avoid scanning unrelated statuses.
+                $ranks = match ($status) {
+                    TicketListStatusFilter::All => null,
+                    TicketListStatusFilter::Unresolved => [0, 1],
+                    TicketListStatusFilter::Open => [0],
+                    TicketListStatusFilter::Request => [1],
+                    TicketListStatusFilter::Quarantined => [2],
+                    TicketListStatusFilter::Resolved => [3],
+                    TicketListStatusFilter::Closed => [4],
```
Jamiras: What happens when we add a new status that appears somewhere in the middle of the list?

database/migrations/2026_09_11_151503_add_facet_covering_index_to_tickets_table.php
```diff
@@ -0,0 +1,32 @@
...
+            $table->index([
+                'deleted_at',
+                'state',
+                'type',
+                'hardcore',
+                'emulator_id',
+                'ticketable_type',
+                'ticketable_id',
+                'ticketable_author_id',
+            ], 'tickets_facet_counts_index');
```
Jamiras, line 22: How does this massive index help if only some of the filters are set?

Jamiras (approved):

app/Platform/Actions/BuildTicketListAction.php
wescopeland: Cruft - dropped in latest.

app/Platform/Actions/BuildTicketListAction.php
wescopeland: We'll have to regenerate the virtual field and remember to update this matcher. I think the design can be improved so this is less of a landmine. I've made some refactors in latest:

* The migration now generates the virtual field automagically through a static function added to the `TicketState` enum itself, which is where the rank numbers also live.
* A new test case has been added which compares the stored rank of each state with the enum, and it will fail if the ranks change without including a new migration that also regenerates the virtual field.
* This matcher is gone, with `TicketListStatusFilter::states()` taking its place. The sort now derives ranks from `TicketState::sortOrder()`.

database/migrations/2026_09_11_151503_add_facet_covering_index_to_tickets_table.php
wescopeland: This is a [covering index](https://www.sqlservercentral.com/articles/increase-query-speed-with-covering-indexes).

`deleted_at` and `state` are the index seek, as they are always set by the query.

Extra filters on top of the seek are applied within the same index scan (verifiable via `EXPLAIN`), so the index is utilized no matter what filters the user sets.

Jamiras (approved):

